PR #5, opened 2022-05-06
Publish relevant crate names

r1viollet:
# What does this PR do?

Publish all the relevant crates to crates.io
This branch can be useful to maintain comments / remember what crates have been secured.

# Motivation

Securing the relevant crates for Datadog.

# Additional Notes

We do not need to merge this PR.

# How to test the change?

[Check the crates here](https://crates.io/teams/github:datadog:libdatadog-owners)


ivoanjo:
Added keep-open for this one; not sure exactly what we want to do with it but I don't think we should close it at this point.

PR #17, opened 2022-05-27
Try building FFI in CI, including Windows

morrisonlevi:
# What does this PR do?

Builds the FFI artifacts on all platforms, including Windows. 

# Motivation

The FFI has been broken several times in the past because this wasn't being run in CI.

# Additional Notes

I don't know much about Windows, I may need help fixing everything.

# How to test the change?

No changes to test; the CI covers everything.


gleocadie:
> I've handed this PR over to the .NET profiling people to get that windows stuff working.

Sorry for the delay, I'm taking this.

.github/workflows/test.yml
```diff
@@ -2,18 +2,19 @@ name: Test
 on: [push]
 env: 
   CARGO_TERM_COLOR: always
+  RUST_VERSION: 1.56.1
```
morrisonlevi, line 5: While this PR was open, it seems master upgraded to 1.60 and this PR brings it back down (probably not on purpose).

ffi-build.sh
```diff
@@ -32,17 +34,28 @@ case "$target" in
...
         ;;
+
     "x86_64-unknown-linux-gnu"|"aarch64-unknown-linux-gnu")
         expected_native_static_libs=" -ldl -lrt -lpthread -lgcc_s -lc -lm -lrt -lpthread -lutil -ldl -lutil"
         native_static_libs=" -ldl -lrt -lpthread -lc -lm -lrt -lpthread -lutil -ldl -lutil"
         ;;
+
+    "x86_64-pc-windows-msvc")
+        expected_native_static_libs="" # I don't know what to expect
+        native_static_libs="" # I don't know what to expect
```
morrisonlevi, line 53: Hmm. How did we not end up with test failures with these set to blank?

gleocadie: hum good question, maybe because `cargo` was not able to retrieve those informations

.github/workflows/test.yml
gleocadie: maybe updating from main is the thing to do.

ffi-build.sh
morrisonlevi: I think the libs `NtDll;UserEnv;Bcrypt;crypt32;wsock32;ws2_32;shlwapi` should be in this list, rather than hard-coding them in the CMakeLists.txt. That's how it works on other platforms. That is, if cargo is working "correctly" and knows about the deps (it should).

ffi-build.sh
```diff
@@ -118,7 +134,7 @@ if [ -n "$unexpected_native_libs" ]; then
 fi
 cd -
 
-echo "Generating the ddprof/ffi.h header..."
+echo "Generating the $destdir/ddprof/ffi.h header..."
```
morrisonlevi, line 137:
```suggestion
echo "Generating the $destdir/include/ddprof/ffi.h header..."
```

.github/actions/cache/action.yaml
```diff
@@ -5,16 +5,19 @@ inputs:
...
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
         path: |
           ~/.cargo/registry/
           ~/.cargo/git/db/
           ~/.cargo/bin/
           target/
-        key: v1-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}-${{ inputs.rust_version }}
+        key: v1-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}-${{ inputs.rust_version }}-${{ inputs.build_profile }}
```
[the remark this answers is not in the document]
morrisonlevi: I think since we commit `Cargo.lock` anyway that it doesn't matter. If we update a version, then it won't be in the cache and it will pull it.

r1viollet:
@pawelchcki we will need your approval on this one (core things)

ffi-build.sh
gleocadie: but how would you compile using the CMake if you do not specify the list. (sorry new to that)
If the CMakeList is generated why pushing it ?

pawelchcki (approved):
Windows builds! LGTM

PR #20, opened 2022-06-01
IPC framework based on tarpc 

pawelchcki:
# What does this PR do?
Implement a msgpack based RPC / IPC. Where a server is spawned in an async context, and clients can be either async or sync. In addition the RPC can be used in a 1 way style. Where messages are simply sent onto a pipe, without requiring reading of any responses. 

This way we can optimize for cases where data transmission is simply in a single direction. 

This RPC framework uses `tarpc` crate and adds to it possibility to submit RPC requests without using tokio, and also transport FileDescriptors that can be used to safely delegate processing to the sidecar process.

See example use here:
https://github.com/DataDog/libdatadog/pull/20/files#diff-b5cbb946eecfad576b88206a0a05bf1a78331d41fcd5842db73e9b902831a202R26

Some quick benchmarks
<pre>
<font color="#26A269">write only interface</font>    time:   [1.8545 µs <b>1.9404 µs</b> 2.0843 µs]                                  
                        change: [-0.9627% +12.268% +31.181%] (p = 0.13 &gt; 0.05)
                        No change in performance detected.
<font color="#A2734C">Found 4 outliers among 100 measurements (4.00%)</font>
  4 (4.00%) high severe

<font color="#26A269">two way interface</font>       time:   [18.007 µs <b>19.087 µs</b> 20.356 µs]                               
                        change: [-15.207% -4.5682% +5.7143%] (p = 0.47 &gt; 0.05)
                        No change in performance detected.
<font color="#A2734C">Found 11 outliers among 100 measurements (11.00%)</font>
  6 (6.00%) low mild
  4 (4.00%) high mild
  1 (1.00%) high severe
</pre>

# Motivation

In some cases the context of data we process needs to live longer than a os process that produces it. For these cases we can spawn a subprocess that will receive the data from producers - and later transform and send to the backend/agent.

For efficient operation we need an optimized RPC, and a method to coordinate sidecar lifecycle.


ivoanjo:
@pawelchcki is this PR still relevant, or can we close it? (It looks like #45 is a superset)

pawelchcki:
@ivoanjo sorry I missed the notification :| 
>@pawelchcki is this PR still relevant, or can we close it? (It looks like https://github.com/DataDog/libdatadog/pull/45 is a superset)

This PR is actually a superset of #45 :sweat_smile:

paullegranddc:
> This PR is actually a superset of https://github.com/DataDog/libdatadog/pull/45 😅

Would be easier for review to make the PR into #45 and then change that to master once #45 has been merged

ddtelemetry/src/worker.rs
```diff
@@ -441,6 +450,18 @@ impl TelemetryWorkerHandle {
             .register_metric_context(name, tags, metric_type, common, namespace)
     }
 
+    pub async fn send_msg(&self, msg: TelemetryActions) -> Result<()> {
+        Ok(self.sender.send(msg).await?)
```
paullegranddc, line 454: Do we want to use send here, or try_send? 
`send` will block until the receiver has space for the message, which is probably not what we want

ddtelemetry/src/worker.rs
```diff
@@ -140,10 +146,12 @@ impl TelemetryWorker {
             telemetry_worker_log!(self, DEBUG, "Handling action {:?}", action);
             match action {
                 Start => {
-                    let req = self.build_app_started();
-                    self.send_request(req).await;
-                    self.deadlines.schedule_next_heartbeat();
-                    self.data.started = true;
+                    if !self.data.started {
```
paullegranddc, line 149: 👍

ddtelemetry/src/ipc/transport/blocking.rs
```diff
@@ -0,0 +1,295 @@
...
+                }
+                None => {
+                    let n = unsafe {
+                        let dst = buf.chunk_mut();
+                        let dst = &mut *(dst as *mut _ as *mut [MaybeUninit<u8>]);
+                        let mut buf_window = tokio::io::ReadBuf::uninit(dst);
+
+                        let b = &mut *(buf_window.unfilled_mut()
+                            as *mut [std::mem::MaybeUninit<u8>]
+                            as *mut [u8]);
```
paullegranddc, line 94: Same comment for the &mut [u8]

paullegranddc (approved):

ddtelemetry/src/ipc/platform/unix/channel/async_channel.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn poll_read(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> Poll<io::Result<()>> {
+        let project = self.project();
+        let mut fds = [0; MAX_FDS];
+
+        unsafe {
+            let b = &mut *(buf.unfilled_mut() as *mut [std::mem::MaybeUninit<u8>] as *mut [u8]);
```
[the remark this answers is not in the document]
pawelchcki: Yeah, I think filling with zero's would be safer, and more secure. 
I basically followed what tokio AsyncRead was doing here. 

This is part of their optimization - since its a buffer that will be written to - then the contents of memory don't matter as much. But I'd prefer to play  on the safe side here

pawelchcki:
Thanks @paullegranddc for the review :bow:! :tada:

PR #22, opened 2022-06-07
Attempt to implement a safer process Fork utility.

pawelchcki:
# What does this PR do?

implement "safer_fork" function, that turns passed function pointer into separate process. Additional data can be passed to the process as long as its been marked as `ForkSafe`. 

Care must be taken that only the types known to behave well when forked will be marked as ForkSafe.

This should reduce the chance where e.g. a Mutex shared with another thread will be passed to forked process.

# Motivation

I've found the Rust type checker to not be very useful when safeguarding semantics around fork - this implementation  can help a little - by y making the data that goes into a forked process explicit and better vetted.


ddtelemetry/src/fork.rs
```diff
@@ -0,0 +1,185 @@
...
+
+macro_rules! impl_forksafe_for_tuple {
+    () => { impl ForkSafe for () {} };
+    ($A:ident $($I:ident)*) => {
+        impl_forksafe_for_tuple!($($I)*);
+
+        impl<$A : ForkSafe, $($I: ForkSafe),*> ForkSafe for ($A, $($I),*) {}
+    }
+}
+// Implement tuples "auto_trait" ForkSafe for *all tuples that contain only ForkSafe valuess
```
ivoanjo, line 59: Minor typo

```suggestion
// Implement tuples "auto_trait" ForkSafe for *all tuples that contain only ForkSafe values
```

ddtelemetry/src/fork.rs
```diff
@@ -0,0 +1,185 @@
...
+pub unsafe fn fork() -> anyhow::Result<Fork> {
+    let res = libc::fork();
+    match res {
+        -1 => Err(anyhow::format_err!("Can't fork child process")),
+        0 => Ok(Fork::Child),
+        res => Ok(Fork::Parent(res)),
+    }
+}
+
+/// Forks supplied funciton pointer into a new process returning Pid of the process and call exit(0) once child function pointers
```
ivoanjo, line 82: Minor typo

```suggestion
/// Forks supplied function pointer into a new process returning Pid of the process and call exit(0) once child function pointers
```

ddtelemetry/src/fork.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Function pointer is used to prevent capturing values in a closure, and all Args must be marked as ForkSafe
+///
+/// Marker ForkSafe should only be applied to types and traits that do not rely on global state
+/// especially Mutexes shared with already running threads must be avoided
+///
+/// # Errors
+/// function will return an Error if
+///
+/// This function will return an error if .
+pub fn safer_fork<Args>(args: Args, f: fn(Args) -> ()) -> anyhow::Result<libc::pid_t>
```
ivoanjo, line 94: I don't quite get why this is called "safer_fork", perhaps the documentation could be expanded to clarify it?

ddtelemetry/src/fork.rs
```diff
@@ -0,0 +1,185 @@
...
+///
+/// # Safety
+///
+/// Existing state of the process must allow safe forking, e.g. no background threads should be running
+/// as any locks held by these threads will be locked forever
+///
+pub unsafe fn fork() -> anyhow::Result<Fork> {
+    let res = libc::fork();
+    match res {
+        -1 => Err(anyhow::format_err!("Can't fork child process")),
```
ivoanjo, line 76: Is there any way we could also include the `errno` for this failure in this message? May help with debugging.

ivoanjo (commented):
This looks reasonable, but I'm missing the context for what this will be used for. Could you share a few notes on the plan here?

ddtelemetry/src/fork.rs
paullegranddc: If I had to guess, it would be because:
* `f` is a function pointer (not a closure) so it cannot capture values that are not fork-safe
* `Args` is fork-safe

So the only values the function called after fork"should be" able access are forksafe.

In practice I don't know if this is true. Globals for instance can still be accessed and be not fork safe.

For instance
```rust
lazy_static! {
    static ref ARRAY: Mutex<Vec<u8>> = Mutex::new(vec![]);
}
```

The Mutex could cause a deadlock if the fork tries to access it.

ddtelemetry/src/fork.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/// Returns PID of current process
+pub fn getpid() -> libc::pid_t {
+    unsafe { libc::getpid() }
+}
+
+/// Sets test panic handler that will ensure exit(1) is called after
+/// the original panic handler
+pub fn set_default_child_panic_handler() {
```
paullegranddc, line 114: if this is only used for test, it could be in the test module

ddtelemetry/src/fork.rs
paullegranddc: Maybe `https://doc.rust-lang.org/std/io/struct.Error.html#method.last_os_error`

ddtelemetry/src/fork.rs
pawelchcki: Exactly, also a reason why Global mutexes are bad :tm:

pawelchcki: Exactly, also a reason why Global mutexes are bad :tm:

pawelchcki:
closing in favor of using explicit unsafe whenever fork is used

PR #23, opened 2022-06-07
Fix clippy errors - including clippy::significant_drop_in_scrutinee

pawelchcki:
# What does this PR do?

Clippy found a potential unexpected deadlock related to unexpected late drops in match block.
This PR fixes it

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/ddprof-ffi/src/vec.rs
+++ b/ddprof-ffi/src/vec.rs
@@ -124,6 +124,7 @@ impl<T> Default for Vec<T> {
     }
 }
 
+#[allow(clippy::get_first)]
 #[cfg(test)]
 mod test {
     use crate::vec::*;
--- a/ddtelemetry/src/worker.rs
+++ b/ddtelemetry/src/worker.rs
@@ -256,6 +256,8 @@ impl TelemetryWorker {
         let mut series = Vec::new();
         for (context_key, extra_tags, points) in self.data.metric_buckets.flush_series() {
             let context_guard = self.data.metric_contexts.get_context(context_key);
+
+            #[allow(clippy::significant_drop_in_scrutinee)]
             let context = match context_guard.read() {
                 Some(context) => context,
                 None => {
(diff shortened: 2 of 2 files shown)


ddprof-ffi/src/vec.rs
```diff
@@ -154,7 +154,7 @@ mod test {
         assert!(ffi_vec.capacity >= 2);
 
         let slice = unsafe { ffi_vec.as_slice().as_slice() };
-        let first = slice.get(0).unwrap();
+        let first = slice.first().unwrap();
```
paullegranddc, line 157: I disagree with the lint in this specific case.

```rust
let first = slice.first().unwrap();
let second = slice.get(1).unwrap();
```

breaks the symmetry of the code compared to 
```rust
let first = slice.get(0).unwrap();
let second = slice.get(1).unwrap();
```

and makes the code not as obvious

I think we should just disable the lint for this line

ddtelemetry/src/worker.rs
```diff
@@ -256,18 +256,15 @@ impl TelemetryWorker {
         let mut series = Vec::new();
         for (context_key, extra_tags, points) in self.data.metric_buckets.flush_series() {
             let context_guard = self.data.metric_contexts.get_context(context_key);
-            let context = match context_guard.read() {
```
paullegranddc, line 259: Same here. I don't understand how the fix improves anything regarding the aim of lint.
If I understand correctly, the lint is triggered when a guard is kept to long when called in as arg of a match.

It's not the case here. `context_guard` is dropped at the end of the function, because it has to live at least as long as the ref returned by `read`.
Modifying the match to an if - else doesn't  change this in the case `context_guard.read()` returns None.

ddprof-ffi/src/vec.rs
pawelchcki: I agree 👍

ddtelemetry/src/worker.rs
pawelchcki: since its not a deadlock yet, I've just disabled the lint for this particular match

paullegranddc (approved):

paullegranddc (approved):

pawelchcki:
I see 1.56 now complains about unknown lint settings 😒 . I think this can be addressed in a separate PR/discussion.

PR #25, opened 2022-06-08
Bump Rust version to 1.60

r1viollet:
# What does this PR do?

Alpine released a new version including rust 1.60. We no longer need to build with rust 1.56.

# Motivation

Build with newer version

# Additional Notes

N/A

# How to test the change?

CI

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,7 +18,7 @@ jobs:
     name: "clippy #${{ matrix.version }}"
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


ivoanjo (approved):
👍 LGTM

Should we also add the `rust-version` to `Cargo.toml` as discussed in https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html#cargo-rust-version ? That should help avoid any "oops wrong Rust version" issues :)

morrisonlevi (approved):

PR #32, opened 2022-07-19
Drop "V3" suffix from exporter APIs

ivoanjo:
# What does this PR do?

Drop the "V3" suffix from exporter APIs. Does not otherwise change the behavior in any way.

# Motivation

As discussed in <https://github.com/DataDog/libdatadog/pull/24#discussion_r921924120> having the "V3" suffix on APIs doesn't seem that useful because:

1. We're still iterating on APIs all the time. We just broke the API again by changing the prefixes to `ddog_` so it seems a bit too early to version things. We never supported more than a version at a time!

2. Since the datadog agent just proxies profiling requests, we probably don't need to support more than one API version at a time.

3. The new intake format (internally called v2.4, whereas the old v3 was renamed to v1.3) does not need extensive API changes, so it's quite possible that we can move profilers to use v2.4 without changing the libdatadog APIs. So having v3 in the API names would force us to break the API even if not needed.

For these reasons, I decided to ride the wave of #24 already breaking all our APIs and also dropping the V3 suffix. Let me know your thoughts :)

# How to test the change?

Validate that profiles are still being reported with example apps/downstream profilers.

### DIFF
--- a/ddprof-exporter/src/lib.rs
+++ b/ddprof-exporter/src/lib.rs
@@ -29,12 +29,12 @@ pub struct Exporter {
     runtime: Runtime,
 }
 
-pub struct FieldsV3 {
+pub struct Fields {
     pub start: DateTime<Utc>,
     pub end: DateTime<Utc>,
 }
 
-pub struct ProfileExporterV3 {
+pub struct ProfileExporter {
     exporter: Exporter,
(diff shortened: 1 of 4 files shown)


paullegranddc (approved):
LGTM, but I'm not among users of these APIs

PR #33, opened 2022-07-25
Add ddprof_ffi_Profile_set_endpoint

kevingosse:
# What does this PR do?

This PR implements and exposes a `ddprof_ffi_Profile_set_endpoint` function, that is used to lazily associate endpoints to spans.

# Motivation

The goal is to help profilers support endpoint profiling.

### DIFF
--- a/ddprof-ffi/src/profiles.rs
+++ b/ddprof-ffi/src/profiles.rs
@@ -340,6 +340,18 @@ pub extern "C" fn ddprof_ffi_Profile_add(
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn ddprof_ffi_Profile_set_endpoint<'a>(
+    profile: &mut ddprof_profiles::Profile,
+    local_root_span_id: CharSlice<'a>,
+    endpoint: CharSlice<'a>
+) {
+    let local_root_span_id = local_root_span_id.to_utf8_lossy();
+    let endpoint = endpoint.to_utf8_lossy();
+
+    profile.add_endpoint(local_root_span_id, endpoint);
+}
+
 #[repr(C)]
 pub struct EncodedProfile {
     start: Timespec,
@@ -391,6 +403,7 @@ pub unsafe extern "C" fn ddprof_ffi_Profile_serialize(
         Some(x) if *x < 0 => None,
         Some(x) => Some(Duration::from_nanos((*x) as u64)),
     };
+
     match || -> Result<_, Box<dyn Error>> { Ok(profile.serialize(end_time, duration)?) }() {
         Ok(ok) => SerializeResult::Ok(ok.into()),
         Err(err) => SerializeResult::Err(err.into()),
--- a/ddprof-ffi/src/profiles.rs
+++ b/ddprof-ffi/src/profiles.rs
@@ -340,6 +340,22 @@ pub extern "C" fn ddprof_ffi_Profile_add(
     }
 }
 
+/// Associate an endpoint to a given local root span id.
+/// During the serialization of the profile, an endpoint label will be added
+/// to all samples that contain a matching local root span id label.
+///
+/// Note: calling this API causes the "trace endpoint" and "local root span id" strings
+/// to be interned, even if no matching sample is found.
+///
+/// # Arguments
+/// * `profile` - a reference to the profile that will contain the samples.
+/// * `local_root_span_id` - the value of the local root span id label to look for.
+/// * `endpoint` - the value of the endpoint label to add for matching samples.
+///
+/// # Safety
+/// The `profile` ptr must point to a valid Profile object created by this
+/// module.
+/// This call is _NOT_ thread-safe.
 #[no_mangle]
 pub unsafe extern "C" fn ddprof_ffi_Profile_set_endpoint<'a>(
     profile: &mut ddprof_profiles::Profile,
--- a/ddprof-ffi/src/profiles.rs
+++ b/ddprof-ffi/src/profiles.rs
@@ -344,7 +344,7 @@ pub extern "C" fn ddprof_ffi_Profile_add(
 pub unsafe extern "C" fn ddprof_ffi_Profile_set_endpoint<'a>(
     profile: &mut ddprof_profiles::Profile,
     local_root_span_id: CharSlice<'a>,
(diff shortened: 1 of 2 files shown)


ddprof-profiles/src/lib.rs
```diff
@@ -438,17 +478,41 @@ impl From<&Profile> for pprof::Profile {
...
+            .samples
+            .iter()
+            .map(|(sample, values)| pprof::Sample {
+                location_id: sample.locations.iter().map(Into::into).collect(),
+                value: values.to_vec(),
+                label: sample.labels.clone(),
+            })
+            .collect();
+
+        if !profile.endpoints.mappings.is_empty() {
```
r1viollet, line 491: it is nice that users without endpoints won't pay the cost for this

ddprof-profiles/src/lib.rs
```diff
@@ -763,4 +828,124 @@ mod api_test {
...
+            locations: vec![],
+            values: vec![1, 10000],
+            labels: vec![id2_label, other_label],
+        };
+
+        profile.add(sample1).expect("add to success");
+
+        profile.add(sample2).expect("add to success");
+
+        profile.add_endpoint(Cow::from("10"), Cow::from("my endpoint"));
```
r1viollet, line 884: nice, I think this goes well with a lot of the runtime requirements

ddprof-profiles/src/lib.rs
```diff
@@ -763,4 +828,124 @@ mod api_test {
...
+                .string_table
+                .get(l3.str as usize)
+                .unwrap(),
+            "my endpoint"
+        );
+
+        let s2 = serialized_profile.sample.get(1).expect("sample");
+
+        // The trace endpoint label shouldn't be added to second sample because the span id doesn't match
+        assert_eq!(s2.label.len(), 2);
```
r1viollet, line 949: so when `local root span id` does not match a known endpoint, we remove the label from the serialized data
(just checking if I understand this well)

r1viollet (approved):
I did a quick review on APIs and things look good.
A minor concern about the increase in complexity. I think a dedicated field could be better (and we would then only append to the labels).
I'm not commenting on rust side of things.
Thanks for adding this !

ddprof-profiles/src/lib.rs
```diff
@@ -438,17 +478,41 @@ impl From<&Profile> for pprof::Profile {
...
+                value: values.to_vec(),
+                label: sample.labels.clone(),
+            })
+            .collect();
+
+        if !profile.endpoints.mappings.is_empty() {
+            for sample in samples.iter_mut() {
+                let mut endpoint: Option<&i64> = None;
+
+                for label in &sample.label {
```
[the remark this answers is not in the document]
kevingosse: The lookup here is to find the span id. We could indeed add a dedicated field for the span id, but I believe the cost of the lookup to be negligible (I'd be very surprised if a profile stores more than ~10 labels on a sample). 
That said, we already iterate over all the labels once to serialize them, so I could merge the two loops.

ddprof-profiles/src/lib.rs
kevingosse: We don't touch `local root span id`, but we only add the endpoint label if `local root span id` matches a known endpoint

ivoanjo (approved):
This looks great! Ideally @morrisonlevi, @paullegranddc or @pawelchcki can comment on the small Rust-isms that we can perhaps improve on but here sir, take my 👍

ddprof-profiles/src/lib.rs
ivoanjo: This is a good point though -- I guess since we `intern(...)` the endpoint, we still ship it, even if there's no matching `local root span id` that needs it.

Which I don't think is particularly problematic (but perhaps documenting with a comment on the API?).

ddprof-profiles/src/lib.rs
```diff
@@ -219,6 +242,7 @@ impl Profile {
             strings: Default::default(),
             start_time,
             period: None,
+            endpoints: Endpoints::new(),
         };
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/33/commits/f8fd2f958c99d9a75bf0afe2bb492c6d2546ea50

ddprof-ffi/src/profiles.rs
```diff
@@ -340,6 +340,18 @@ pub extern "C" fn ddprof_ffi_Profile_add(
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn ddprof_ffi_Profile_set_endpoint<'a>(
+    profile: &mut ddprof_profiles::Profile,
+    local_root_span_id: CharSlice<'a>,
+    endpoint: CharSlice<'a>,
+) {
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/33/commits/f8fd2f958c99d9a75bf0afe2bb492c6d2546ea50

ddprof-profiles/src/lib.rs
kevingosse: Comment added in https://github.com/DataDog/libdatadog/pull/33/commits/f8fd2f958c99d9a75bf0afe2bb492c6d2546ea50

morrisonlevi (commented):
It looks good! I'm going to try using it; will report back soon.

morrisonlevi (approved):
I hit some snags trying it out from the PHP profiler because of how the profiler is written. However, I couldn't see any reason why this PR wouldn't be suitable. Have you tried out the PR in .NET? Ideally someone would send profiles, even if just a PoC, using the PR to demonstrate the real-world viability, but I'm not going to block on it.

kevingosse:
> I hit some snags trying it out from the PHP profiler because of how the profiler is written. However, I couldn't see any reason why this PR wouldn't be suitable. Have you tried out the PR in .NET? Ideally someone would send profiles, even if just a PoC, using the PR to demonstrate the real-world viability, but I'm not going to block on it.

Yep, the feature on the .NET side is already fully written and tested 🙂 https://github.com/DataDog/dd-trace-dotnet/pull/3015

PR #36, opened 2022-08-11
Cleanup socket APIs for other Rust consumers

morrisonlevi:
# Description and Motivation

We should probably do a pass on this at some point, but consumers of
the Rust API should not have to deal with a boxed error generally so
that we avoid allocations and retain information about errors. We
would still have boxed errors in a few cases, most likely:

 1. The errors on FFI boundaries will need boxed for ABI stability.
 2. Having multiple errors from different crates in the same function.
    Although it's possible to make error types to serve this case, it
    does take effort that's probably better spent elsewhere for now.

Anyway, in this case, there is only one error type and the boxing is
unnecessary for Rust callers.

# Additional Notes

I also changed some usages of `std::error::Error` to just `Error` because
it's already imported.

I also exported `socket_path_from_uri` in `exporter` to match
`socket_path_to_uri`.

# How to test the change?

Most things will Just Work, but you may need to wrap the error in
`Ok(...?)` to get a conversion to another error type.


ivoanjo (commented):
Seems reasonable 👍

PR #38, opened 2022-08-15
Use anyhow::Result instead of Result<_, Box<dyn Error>>

morrisonlevi:
# What does this PR do?

You can't put the latter into the former, and the former can move
across thread boundaries while the latter can't (at least, I _think_
that's what the giant error message said).

# Motivation

I hit some of these errors while using `anyhow` in the PHP profiler.
Since ddcommon is using the crate anyway, it seemed like we could align
here.

# Additional Notes

# How to test the change?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,6 +183,7 @@ dependencies = [
 name = "ddprof-ffi"
 version = "0.7.0-rc.1"
 dependencies = [
+ "anyhow",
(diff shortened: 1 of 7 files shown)


paullgdc (approved):

paullegranddc (approved):

ivoanjo (approved):
Looks reasonable!

PR #43, opened 2022-08-24
Add support for Windows Named Pipe

gleocadie:
# What does this PR do?

This PR adds support for the  Windows Named Pipe.

# Motivation

The .NET tracer currently runs in environment (AAS) where traces are sent to the agent through windows named pipe. The customers that have their apps in that environment are waiting for the profiler to work there too.

# Additional Notes

Not on top of my head

# How to test the change?

* I build a custom version of libdatadog, linked it to the compiler.
* Implement support for named pipe in the profiler
* Setup the datadog agent to listen on windows named pipe by setting the environment variable `DD_APM_WINDOWS_PIPE_NAME` to `\\.\pipe\datadogpipe\mypipe`
* Run the application

I check the profiler log and I saw the agent url is the named pipe:
`[2022-08-23 22:01:41.691 | info | PId: 421300 | TId: 411900] Using agent endpoint windows:\\.\pipe\datadogpipe\mypipe`

And saw in the dashboard my profiles:
image (16)

### DIFF
--- a/ddcommon/src/connector/conn_stream.rs
+++ b/ddcommon/src/connector/conn_stream.rs
@@ -38,10 +38,10 @@ pub type ConnStreamError = Box<dyn std::error::Error + Send + Sync>;
 
 use hyper::{client::HttpConnector, service::Service};
 impl ConnStream {
-    pub async fn from_uds_uri(uri: hyper::Uri) -> Result<ConnStream, ConnStreamError> {
+    pub async fn from_uds_uri(_uri: hyper::Uri) -> Result<ConnStream, ConnStreamError> {
         #[cfg(unix)]
         {
-            let path = super::uds::socket_path_from_uri(&uri)?;
+            let path = super::uds::socket_path_from_uri(&_uri)?;
             Ok(ConnStream::Udp {
                 transport: tokio::net::UnixStream::connect(path).await?,
             })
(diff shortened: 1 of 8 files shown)


ddcommon/src/connector/conn_stream.rs
```diff
@@ -26,16 +26,22 @@ pub enum ConnStream {
...
+        transport: tokio::net::windows::named_pipe::NamedPipeClient,
+    },
 }
 
 pub type ConnStreamError = Box<dyn std::error::Error + Send + Sync>;
 
 use hyper::{client::HttpConnector, service::Service};
 impl ConnStream {
-    pub async fn from_uds_uri(uri: hyper::Uri) -> Result<ConnStream, ConnStreamError> {
+    pub async fn from_uds_uri(_uri: hyper::Uri) -> Result<ConnStream, ConnStreamError> {
```
paullegranddc, line 41: Did you put `_uri` here because of unused warnings when compiling on `#[cfg(not(unix))`?

If yes, I'd rather put something like

```rust
#[cfg(not(unix))]
{
    let _ = uri;
    Err(super::errors::Error::UnixSocketUnsupported.into())
}
```

Because when I see underscored variables in function signatures, I assume the value will. be unused which is not always the case

paullegranddc (approved):
Otherwise LGTM

ddcommon/src/connector/conn_stream.rs
gleocadie: yeah exactly. Sorry for that, I'm a beginner in Rust.
Thanks, I'll fix it

gleocadie: done in https://github.com/DataDog/libdatadog/pull/43/commits/c9ee643c12d20d6c56852cf9d59f1d987f7f867d

morrisonlevi (approved):
I know nothing about Windows, but if it works for you, this looks fine to me!

gleocadie:
> I know nothing about Windows, but if it works for you, this looks fine to me!

We will put some tests in our hand with named pipe to ensure that it works accordingly (once released)

PR #46, opened 2022-08-30
Fix `build-profiling-ffi.sh` script not working with relative paths + Fix incorrect placeholder for version

ivoanjo:
# What does this PR do?

* Fix `build-profiling-ffi.sh` script not working with relative paths
* Fix incorrect placeholder for version

# Motivation

See individual commit messages for details.

# How to test the change?

* Check that `build-profiling-ffi.sh` still produces the same output as before, as well as working now with relative paths
* Check that `datadog_profiling_with_rpath.pc` gets templated correctly with the version

### DIFF
--- a/build-profiling-ffi.sh
+++ b/build-profiling-ffi.sh
@@ -63,24 +63,23 @@ case "$target" in
 esac
 
 echo "Recognized platform '${target}'. Adding libs: ${native_static_libs}"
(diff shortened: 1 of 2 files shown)


sanchda (approved):
Not bad!  Fixes the problem without adding any new tools or drastically changing the build system.  Thank you!

PR #49, opened 2022-09-05
Fix potential issue discovered by new lint suspicious_to_owned

pawelchcki:
see:
https://github.com/rust-lang/rust-clippy/pull/8984

### DIFF
--- a/profiling-ffi/src/exporter.rs
+++ b/profiling-ffi/src/exporter.rs
@@ -123,7 +123,7 @@ pub extern "C" fn profile_exporter_new(
     match || -> anyhow::Result<ProfileExporter> {
         let family = unsafe { family.to_utf8_lossy() }.into_owned();
         let converted_endpoint = unsafe { try_to_endpoint(endpoint)? };
-        let tags = tags.map(|tags| tags.iter().map(|tag| tag.clone().into_owned()).collect());
+        let tags = tags.map(|tags| tags.iter().map(|tag| tag.clone()).collect());
         ProfileExporter::new(family, tags, converted_endpoint)
     }() {
         Ok(exporter) => NewProfileExporterResult::Ok(Box::into_raw(Box::new(exporter))),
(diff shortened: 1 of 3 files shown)


profiling-ffi/src/exporter.rs
```diff
@@ -123,7 +123,7 @@ pub extern "C" fn profile_exporter_new(
     match || -> anyhow::Result<ProfileExporter> {
         let family = unsafe { family.to_utf8_lossy() }.into_owned();
         let converted_endpoint = unsafe { try_to_endpoint(endpoint)? };
-        let tags = tags.map(|tags| tags.iter().map(|tag| tag.clone().into_owned()).collect());
```
pawelchcki, line 126: As it stands right now  `into_owned()` was a noop, so I think we can remove it completely

r1viollet (approved):
LGTM

profiling-ffi/src/exporter.rs
r1viollet: I'm not sure why we were doing clone + to_owned. I might be missing something.

morrisonlevi: It's important that we get a `Cow::Owned` version because the tag can come from FFI and we don't know its real lifetime. I don't think `clone` will actually work for what we want:

https://doc.rust-lang.org/src/alloc/borrow.rs.html#194-203

That does say it just borrows if it's borrowed, right?

paullegranddc: The version before didn't do that though.

the `Tag::into_owned` was calling to `Cow::to_owned` which comes from the blanket implementation https://doc.rust-lang.org/std/borrow/enum.Cow.html#impl-ToOwned that just calls `clone`. So borrowed tags stayed borrowed before.

But I'm not sure why this is an issue, since tag have to be dropped using drop_tag which understands wether the cow is owned or borrowed?

morrisonlevi:
Thanks, Pawel. I forgot this was already open when I opened #52, which superseded this one.

PR #50, opened 2022-09-06
Add AAS Resource id

gleocadie:
# What does this PR do?

This PR add a new tag `aas.resource.id`. This tag will be added only for applications running on AAS (Azure App Services).

# Motivation

AAS resource id is like container id, this will allow billing customer when they use profilers (first .NET) in that environment.

# Additional Notes

The computation of the resource id was backported from the [.NET tracer code](https://github.com/DataDog/dd-trace-dotnet/blob/834a4b05b4ed91a819eb78761bf1ddb805969f65/tracer/src/Datadog.Trace/PlatformHelpers/AzureAppServices.cs#L215) 

# How to test the change?

🛑 Tests are currently in progress with a custom version of the profiler and libdatadog:
- Creating a VM
- Installing a .NET application
- Setting up environment variables
- Installing .NET profiler

checking in the UI if the `aas.resource.id` tag is present as expected


morrisonlevi (approved):
I don't know anything about AAS. The tests seem reasonable to me. Can you think of anything else that should be tested or done? Otherwise LGTM.

morrisonlevi:
Actually, could you merge in the main branch or rebase to ensure that the clippy nightly job looks okay? It's currently failing due to something that has since been fixed.

morrisonlevi (changes requested):
Actually, could you merge in the main branch or rebase to ensure that the clippy nightly job looks okay? It's currently failing due to something that has since been fixed.

gleocadie:
> Actually, could you merge in the main branch or rebase to ensure that the clippy nightly job looks okay? It's currently failing due to something that has since been fixed.

sure will do.
Will also update the PR (PM asked for other information)

ddcommon/Cargo.toml
```diff
@@ -18,6 +18,7 @@ futures-util = {version = "0.3.0", default-features = false}
 hex = "0.4"
 hyper = {version = "0.14", features = ["http1", "client", "tcp", "stream"], default-features = false}
 hyper-rustls = {version = "0.23", default-features = false, features = ["native-tokio", "http1", "tls12"]}
+indexmap = "1.8"
```
morrisonlevi, line 21: If this is only used for tests (I think it is?) then you would put it in the `[dev-dependencies]` section instead of the `[dependencies]` section.

ddcommon/src/azure_app_services.rs
```diff
@@ -0,0 +1,495 @@
...
+            instance_id,
+            is_relevant,
+            site_kind,
+            site_type,
+        }
+    }
+
+    pub fn is_relevant(&self) -> bool {
+        self.is_relevant
+    }
```
morrisonlevi, line 160: What is the function `is_relevant` relevant to?

ddcommon/Cargo.toml
gleocadie: done

ddcommon/src/azure_app_services.rs
gleocadie: is the AAS metadata is relevant => if we can use it

morrisonlevi (approved):
Looks good to me, although I don't understand the AAS specific bits ^_^

PR #51, opened 2022-09-13
fix nightly lints surprise build breakage

pawelchcki:
Nightly builds that pull in latest lints daily, often cause us to chase broken tests.
In a few cases it turned out that the clippy checks were later removed - while we've lost time fixing code to keep the tests up to date. 

This PR introduces adds latest "stable" clippy to be used as reference against latest nightly - and allows other builds in the matrix to finish if one of them fails.

Additionally it improves the caching on lints, and should reduce their execution time to seconds, instead of minutes

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,6 +30,7 @@ jobs:
     name: "clippy #${{ matrix.rust_version }}"
     needs: setup-clippy-matrix
     strategy:
+      fail-fast: false
       matrix:
         rust_version: ${{ fromJson(needs.setup-clippy-matrix.outputs.versions) }}
     runs-on: ubuntu-latest
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -80,8 +80,9 @@ jobs:
             ~/.cargo/registry/
             ~/.cargo/git/db/
             ~/.cargo/bin/
-          key: v0.5.0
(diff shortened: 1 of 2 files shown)


morrisonlevi (commented):
Given that we're locked to v1.60 for now, maybe we should switch to latest stable instead of nightly? Or are there specific nightly features that we use or want to test against in CI?

pawelchcki:
@morrisonlevi 

> Given that we're locked to v1.60 for now, maybe we should switch to latest stable instead of nightly? Or are there specific nightly features that we use or want to test against in CI?

Nightly has a few more lints, and I think it might be worth keeping, for now, since its track record seems somewhat good so far.
I'd suggest we revisit its removal once we hit another false positive. 

I've added stable version + improved caching.

.github/workflows/lint.yml
```diff
@@ -58,7 +61,16 @@ jobs:
...
-      - run: cargo install cargo-bundle-licenses
+      - name: Cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry/
+            ~/.cargo/git/db/
+            ~/.cargo/bin/
+            ~/.cargo/.crates.toml
+          key: "v1-0.5.0"
```
morrisonlevi, line 72: This number is magic to me; where does it come from?

morrisonlevi (approved):

PR #52, opened 2022-09-13
Change Tag to hold a String instead of Cow<str>

morrisonlevi:
# What does this PR do?

Change `Tag` to hold a `String` instead of `Cow<str>`.

I also implemented `AsRef<str>` for `Tag`. 

# Motivation

This originated from a lint on nightly which didn't like `Cow::to_owned`
being called, and instead callers should use `Cow::clone` or
`Cow::into_owned`. However, while inspecting the code with fresh eyes, we
saw that we're always holding a `Cow::Owned` variant, so just hold the
`String` instead.

Simultaneously, I got a crash report in Rust for the PHP profiler. I
don't support forking, and they forked. Anyway, for this particular
crash it's not much harder to support `fork` (well, it leaks
technically but works) than to disable everything on a fork. As part of
supporting fork, I need to fixup the `process_id` tag to have the new
pid after a fork, and `Tag::as_ref` will help with that.

# Additional Notes

I also noticed that we were fully qualifying two usages of `DateTime` and
`Utc` even though we're publicly using those, so I shortened them.

# How to test the change?

If your code is calling `Tag::to_owned` then it will break. It should be
reworked into using a `Tag::clone` or call `.to_string` on it or similar.

### DIFF
--- a/ddcommon/src/tag.rs
+++ b/ddcommon/src/tag.rs
@@ -18,6 +18,12 @@ impl Debug for Tag {
     }
 }
 
+impl AsRef<str> for Tag {
+    fn as_ref(&self) -> &str {
+        self.value.as_ref()
+    }
+}
+
 // Any type which implements Display automatically has to_string.
 impl Display for Tag {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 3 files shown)


Qard (approved):

r1viollet (approved):
LGTM

pawelchcki (approved):

PR #53, opened 2022-09-22
Report profiling data in v2.4 intake format; compress files

ivoanjo:
# What does this PR do?

This PR changes libdatadog to report profiling data using the v2.4 intake format. It also compresses the files ~including~ excluding the new `event.json` file using lz4.

This was tested with both the Ruby (agent and agentless modes) and the PHP profilers.

This also introduces two breaking changes:

1. Breaking API change: the
`ddog_ProfileExporter_new` / `ProfileExporter::new` functions now take two additional arguments -- the profiling library name and version.
2. Breaking behavior change: Files are now automatically compressed with lz4 by libdatadog so libraries mustn't do their own compression (Ruby is such an example). 

We expect that other than the API change, using the v2.4 intake format and the added compression will be transparent to the libdatadog users.

Thanks to @morrisonlevi for pairing with me on this.

Note that this does not (yet) include support for including attributes in the reporting data. I'll leave that for a separate PR.

# Motivation

Get libdatadog users to use intake v2.4, and lay the ground work for including attributes in the reported data.

Also save some bytes over the wire for files. A few stats:
    - PHP CLI: running `composer create-project symfony/symfony-demo` resulted in a 40016 byte pprof which compressed to 18139 bytes in 141661 nanoseconds. That's a compression ratio of 2.2061 and a space savings of 54.67%.
    - Ruby github relenv test app: 
    |                      | Uncompressed | Gzip (Current Ruby profiler) | LZ4 (libdatadog) |
|----------------------|-------------:|-----------------------------:|-----------------:|
| code-provenance.json |          49K |                           6K |             9.5K |
| rubyprofile.pprof    |         402K |                          48K |              76K |

# Additional Notes

(Nothing)

# How to test the change?

Validate that data can still be reported to the backend, in both agent as well as agentless modes.


ivoanjo:
@morrisonlevi I was able to test your lz4 changes successfully with the Ruby profiler as well. I had, of course, to disable my own compression -- funnily enough the intake would still accept profiles that were gzipped by Ruby + lz4'd by libdatadog but they never showed up on the profile list ;)

profiling/src/exporter/mod.rs
```diff
@@ -120,40 +124,61 @@ impl ProfileExporter {
...
 
         for file in files {
-            form.add_reader_file(
-                format!("data[{}]", file.name),
-                Cursor::new(file.bytes.to_owned()),
-                file.name,
-            )
+            let mut encoder = FrameEncoder::new(Vec::new());
+            encoder.write_all(file.bytes)?;
+            form.add_reader_file(file.name, Cursor::new(encoder.finish()?), file.name)
```
r1viollet, line 172: How is the compression format defined in the payload ?

ivoanjo: AFAIK the intake just looks for magic bits to distinguish if an uploaded file is [gzip|lz4] compressed or not. Otherwise no changes needed.

morrisonlevi:
I'm clarifying with Florian whether it's important that `name` and `filename` are like this:

```
Content-Disposition: form-data; name="time"; filename="time.pprof"
```

Or if this is fine:

```
Content-Disposition: form-data; name="time.pprof"; filename="time.pprof"
```

profiling/src/exporter/mod.rs
morrisonlevi: Ivo is right, according to Florian:

> Levi: How does it know what compression format to use? Inspect magic bytes?
> Florian: yes
> Florian: I was looking at java profiler and it uses LZ4 by default
> Florian: Jaroslav experiments has shown that it's a good space/CPU tradeoff for the profiler

It's also is why I picked lz4. I didn't test or verify these claims beyond "works for me."

morrisonlevi (commented):
I'm still waiting on Florian for a question about name vs filename. Aside from this, with fresh I eyes I think we should set profiling library name and version at the same place we set family, which is when we build the exporter, not the request, as this information is unlikely to change.

I also don't know how many people are re-using an exporter vs just making a new one so... it may be a moot point.

ivoanjo:
> I'm still waiting on Florian for a question about name vs filename. Aside from this, with fresh I eyes I think we should set profiling library name and version at the same place we set family, which is when we build the exporter, not the request, as this information is unlikely to change.

Right, that makes sense. I'll probably not have time today, but I've made and note and will come back soon-ish™️

profiling/src/exporter/mod.rs
```diff
@@ -130,30 +145,55 @@ impl ProfileExporter {
...
+            "tags_profiler": tags_profiler,
+            "start": start.format("%Y-%m-%dT%H:%M:%S%.9fZ").to_string(),
+            "end": end.format("%Y-%m-%dT%H:%M:%S%.9fZ").to_string(),
+            "family": self.family.as_ref(),
+            "version": "4",
+        })
+        .to_string();
+
+        form.add_reader_file_with_mime(
+            // intake will look for a field of this name
```
ivoanjo, line 170: Should this be file, not field?

```suggestion
            // intake will look for a file of this name
```

morrisonlevi (commented):
This looks good to me, but then again I wrote or paired for all of it, so that's expected ^_^. I'll let someone else review it.

profiling/src/exporter/mod.rs
```diff
@@ -130,30 +145,55 @@ impl ProfileExporter {
...
-            form.add_reader_file(
-                format!("data[{}]", file.name),
-                Cursor::new(file.bytes.to_owned()),
-                file.name,
-            )
+            let mut encoder = FrameEncoder::new(Vec::new());
+            encoder.write_all(file.bytes)?;
+            let encoded = encoder.finish()?;
+            // intake does not care about these name of the form field for these
+            form.add_reader_file(file.name, Cursor::new(encoded), file.name)
```
ivoanjo, line 184: I'm not quite sure this comment adds a lot -- we're sending the file name as the file name which is kinda the obvious thing to do...? (So maybe remove it?)

profiling/src/exporter/mod.rs
morrisonlevi: Interestingly, it wasn't a mistake. Intake does not look for `filename=event.json`, it looks for `name=event`.

profiling/src/exporter/mod.rs
morrisonlevi: At least currently, the code doesn't match the "examples" from confluence and the RFC, so that's why I left a comment. I will improve it.

profiling/src/exporter/mod.rs
ivoanjo: Ahhh. I guess `// intake looks for name=event for this file` or something similar would be clearer? Or even `// Intake does not look for filename=event.json, it looks for name=event`? 😄

profiling/src/exporter/mod.rs
r1viollet: Thanks. There are indeed magic values at the start of the payload that you can use. 
As a Datadog agnostic library, this is not super friendly.

profiling/src/exporter/mod.rs
```diff
@@ -106,15 +112,24 @@ impl Request {
...
+    pub fn new<F, N, V>(
+        profiling_library_name: N,
+        profiling_library_version: V,
+        family: F,
         tags: Option<Vec<Tag>>,
         endpoint: Endpoint,
-    ) -> anyhow::Result<ProfileExporter> {
+    ) -> anyhow::Result<ProfileExporter>
+    where
+        F: Into<Cow<'static, str>>,
```
r1viollet, line 123: why do we need to alias to different types ? 

Also the `Cow 'static` lifetime seems strange to me.

r1viollet (approved):
LGTM !
Happy to integrate this in native !

PR #54, opened 2022-09-23
Paullgdc/telemetry ffi again

paullegranddc:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/ddtelemetry-ffi/Cargo.toml
+++ b/ddtelemetry-ffi/Cargo.toml
@@ -9,9 +9,10 @@ edition = "2021"
 [lib]
 # LTO is ignored if "lib" is added as crate type
 # cf. https://github.com/rust-lang/rust/issues/51009
-crate-type = ["staticlib", "cdylib"]
+crate-type = ["lib", "staticlib", "cdylib"]
 
 [dependencies]
 ddtelemetry = { path = "../ddtelemetry", version = "0.8.0" }
 ddcommon-ffi = { path = "../ddcommon-ffi", version = "0.8.0" }
+ddcommon = { path = "../ddcommon", version = "0.8.0" }
 paste = "1"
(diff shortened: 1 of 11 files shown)


ddtelemetry-ffi/Cargo.toml
```diff
@@ -9,9 +9,10 @@ edition = "2021"
 [lib]
 # LTO is ignored if "lib" is added as crate type
 # cf. https://github.com/rust-lang/rust/issues/51009
-crate-type = ["staticlib", "cdylib"]
+crate-type = ["lib", "staticlib", "cdylib"]
```
morrisonlevi, line 12: Why do you need lib? Note that due to the above comment, you will lose LTO and your library size will increase.

paullegranddc: This is a just a temporary hack to hijack the current build script and re-esxport ddtelemetry-ffi inside the profiler-ffi crate, to build and the telemetry example using the artifact built by the build-profiler-ffi script

Once I figure out how to actually build the library, I'll remove this

ivoanjo (commented):
Hey Paul 👋! Since you had asked feedback on the API (sorry it took a week), I looked at this file and the generated `telemetry.h` from the POV of "I'm given this thing to integrate with on my library".

So my comments are all from that POV. In particular, when I had a question, if it wasn't on this file, or on the header, I purposely did not go one level deeper into the Rust code, to hopefully give you the POV of a (\*cof\* very \*cof\*) lazy person that needs to get the job done and what was going through my head.

API design in C is for sure awful lol.

Also, I saw that in an earlier version you were using C, and now moved to C++ (although you still list the version as "C89"). I would actually suggest going back to C, as I'd expect most libraries will either be Rust or C, so I think you'd get a more realistic example.

(I am aware the profiler example is C++ ahaha. But I wish it was C as well).

ivoanjo:
Also, have you considered doing an actual implementation for one of the languages that are using libdatadog? I think having an actual, real API consumer would help uncover some of the sharp edges and issues earlier.

PR #60, opened 2022-10-10
`build-telemetry-ffi.sh` for building a working ddtelemetry shared artifact

pawelchcki:
# What does this PR do?
Split telemetry ffi compilation into separate script.

+ additionally update github action scripts to build the script and bump version to avoid warning about deprecated github actions


build-profiling-ffi.sh
```diff
@@ -154,9 +154,6 @@ cbindgen --crate ddcommon-ffi \
 cbindgen --crate "${datadog_profiling_ffi}" \
     --config profiling-ffi/cbindgen.toml \
     --output "$destdir/include/datadog/profiling.h"
-cbindgen --crate ddtelemetry-ffi \
-    --config ddtelemetry-ffi/cbindgen.toml \
-    --output "$destdir/include/datadog/telemetry.h"
-./target/debug/dedup_headers "$destdir/include/datadog/common.h" "$destdir/include/datadog/telemetry.h" "$destdir/include/datadog/profiling.h"
+./target/debug/dedup_headers "$destdir/include/datadog/common.h" "$destdir/include/datadog/profiling.h"
```
paullegranddc, line 157: We need to generate the headers here for telemetry if we want to use both `datadog/profiling.h` and `datadog/telemetry.h` together, because otherwise the 2 build scripts are going to generate different `datadog/common.h` headers.
Maybe the header generation can be moved to a seprate scripts that we execute once per build.

ddtelemetry-ffi/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 # This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
 
 [package]
-name = "ddtelemetry-ffi"
+name = "datadog-telemetry-ffi"
```
paullegranddc, line 5: If you rename the crate, it'd be a good idea to move it to the same path as it's name

build-profiling-ffi.sh
pawelchcki: :thinking: 
Problem is the generated FFI library only has symbols for either profiling or telemetry.
So having extra header information might not be useful

build-telemetry-ffi.sh
```diff
@@ -0,0 +1,159 @@
...
+        ;;
+
+    *)
+        >&2 echo "Unknown platform '${target}'"
+        exit 1
+        ;;
+esac
+
+echo "Recognized platform '${target}'. Adding libs: ${native_static_libs}"
+sed < profiling-ffi/datadog_profiling.pc.in "s/@Datadog_VERSION@/${version}/g" \
```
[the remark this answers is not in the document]
pawelchcki: there is nothing interesting in those pkg-config files at the moment. 
I think I can actually remove them from telemetry build

build-profiling-ffi.sh
paullegranddc: `common.h` only contains function declarations which are contained in both `profiling.h` and `telemetry.h` (things like ffi::Vec C bindings for instance). So generating from both headers will not expose symbols in common.h that are not in built artefacts.

But if you don't use both headers to generate it, `profiling.h` and `telemetry.h` might contain the same definitions causing collisions if you include both headers.

paullegranddc: Right now, I don't think you could load both telemetry and profiling dynlibs anyway because we export the same symbols from the two libs, but it's not a reason to not make the headers compatible.

ivoanjo: I'm curious, do we expect there to be many situations where we only make use of profiling, but not telemetry symbols, or vice-versa?

pawelchcki: @ivoanjo at the moment its TBD - I think. 
For PHP case separate header sets will work without a problem. 
For other Languages in the future probably a single header set will be the way to go

bwoebi: Well, I'm currently anyway re-exporting everything via `pub use ddsomething::*;` to only have a single static library. I just need all my headers cooperating. (because cbindgen does not (yet) support recursive header generation across multiple crates at once)

ivoanjo: I'd be happy to target a single library for Ruby as well until we find good reasons to do otherwise :)

pawelchcki:
Hmm this PR grew unexpectedly. 
I've added a cbindgen step in Dockerfile - so that the correct headers will be built. 

@morrisonlevi I think we could use `docker buildx bake` directly in the other build repo. Or migrate entirely to bake release builds.

If we add centos to dockerfile. The release build could be something as simple as:
`docker buildx bake release` which would run both arm and amd64 builds. although at a penalty of running things with qemu.

We currently don't have any runners available anywhere that are able to run both docker and arm (its either, or). Maybe things could change in the future.

.dockerignore
```diff
@@ -1,6 +1,6 @@
 target/
 build/
 .github/
-Dockerfile.build
+build.Dockerfile
```
ivoanjo, line 4: Is this still needed with `Dockerfile.build` is getting removed?

.github/workflows/test.yml
```diff
@@ -72,11 +72,17 @@ jobs:
...
+        run: |
+           chmod +x build-profiling-ffi.sh
+           ./build-profiling-ffi.sh ${OUTPUT_FOLDER}/profiling
+
+      - name: "Generate Telemetry FFI"
         shell: bash
         run: |
            chmod +x build-profiling-ffi.sh
-           ./build-profiling-ffi.sh ${OUTPUT_FOLDER}
+           ./build-profiling-ffi.sh ${OUTPUT_FOLDER}/telemetry
```
ivoanjo, line 85: Should this be `build-telemetry-ffi`?

build-profiling-ffi.sh
```diff
@@ -35,7 +35,7 @@ case "$target" in
         remove_rpath=1
         ;;
 
-    "x86_64-apple-darwin")
+    "x86_64-apple-darwin"|"aarch64-apple-darwin")
         expected_native_static_libs=" -framework Security -framework CoreFoundation -liconv -lSystem -lresolv -lc -lm -liconv"
```
ivoanjo, line 39: This is interesting... Were we missing support for arm macs?

docker-bake.hcl
```diff
@@ -1,11 +1,11 @@
 target "alpine-base" {
-  dockerfile = "tools/docker/Dockerfile.build"
+  dockerfile = "tools/docker/build.Dockerfile"
   tags = ["ghcr.io/datadog/libdatadog-build:alpine-base"]
   target = "alpine_builder"
 }
 
 target "alpine-build" {
-  dockerfile = "tools/docker/Dockerfile.build"
+  dockerfile = "tools/docker/build.Dockerfile"
```
ivoanjo, line 8: Wait, are we missing a file from this PR? I initially thought this file may just be for testing, but I see it referenced here 🤨

build-telemetry-ffi.sh
```diff
@@ -0,0 +1,131 @@
...
+
+# Unless explicitly stated otherwise all files in this repository are licensed
+# under the Apache License Version 2.0. This product includes software developed
+# at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+set -eu
+
+destdir="$1"
+
+mkdir -v -p "$destdir/include/datadog" "$destdir/lib/pkgconfig" "$destdir/cmake"
```
ivoanjo, line 11: ⬇️  This is quite the big bash blob to duplicate, and it's so similar to the profiling one. May be worth taking a stab at having a single one for both telemetry and profiling.

ivoanjo (commented):
I'm still a Rust noob, but I took at stab at reviewing this and left a few notes :)

.dockerignore
pawelchcki: Fixed the gitignore rules, and readded the file :)

build-profiling-ffi.sh
pawelchcki: No one so far tried compiling the lib on a mac

docker-bake.hcl
pawelchcki: Re-added it, sorry for the confusion

build-telemetry-ffi.sh
pawelchcki: I think we should too, but want to sync with you folks first before making more changes to the build process. Lets handle this in a separate PR

.github/workflows/test.yml
pawelchcki: Ohhh thats why the CI worked when the script was broken :facepalm:

build-telemetry-ffi.sh
ivoanjo: Sounds extremely reasonable 👍

r1viollet:
What is the status on this PR ? Do you need a review ?

r1viollet:
@paullegranddc we would need FFIs for telemetry, is this the PR we need to resume ?

PR #62, opened 2022-10-12
Profiling Breakdown API

morrisonlevi:
# What does this PR do?

Adds timestamp information to profiling samples for both the Rust and C APIs. The pprof alterations come from this PR: felixge/pprof#3. Note that when encoded, it's possible for the `ticks` values to be negative as samples can be collected from different sources asynchronously and end up happening before the start time of the profile. I am unsure if this is a problem.

Updates the `profiles.c` and `profiles.rs` examples to be more interesting in general, but also to use breakdowns. They both write a mostly hard-coded profile to stdout, which you can feed into protoc to explore.

# Motivation

I want to move the timeline UI project forward, and my recent R&D week was spent in these parts of libdatadog, so I could quickly make these changes.

# Additional Notes

This particularly PR would force everyone to send a timestamp for the sample. Maybe it would be prudent to make the timestamp optional. Can you think of any reason to do so?

# How to test the change?

In both the C and Rust APIs, you need to send the timestamp of the sample, which in C is the number of nanoseconds since the UNIX epoch and in Rust this is a `SystemTime`. These will encoded in the pprof as offsets since the start time of the profile.

### DIFF
--- a/examples/ffi/exporter.cpp
+++ b/examples/ffi/exporter.cpp
@@ -2,6 +2,7 @@ extern "C" {
 #include <datadog/common.h>
 #include <datadog/profiling.h>
 }
+#include <cassert>
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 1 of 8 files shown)


morrisonlevi:
Closing as the pprof proposal was closed.

PR #64, opened 2022-10-17
Allow setting cgroup path for integration testing purposes

pawelchcki:
# What does this PR do?

Currently Cgroup path is rightfully hardcoded in the source, however this makes it impossible to perform end-to-end tests on stubbed values of the file, as `/proc/groups/self` file is not overridable on linux systems

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/ddcommon/src/container_id.rs
+++ b/ddcommon/src/container_id.rs
@@ -37,7 +37,9 @@ Following environments are supported:
       `1:name=systemd:/ecs/8cd79a803caf4d2aa945152e934a5c00/8cd79a803caf4d2aa945152e934a5c00-1053176469`
 */
 
-const CGROUP_PATH: &str = "/proc/self/cgroup";
+const DEFAULT_CGROUP_PATH: &str = "/proc/self/cgroup";
+
+static mut CGROUP_PATH: Option<String> = None;
 
 const UUID_SOURCE: &str =
     r"[0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}";
@@ -85,10 +87,16 @@ fn extract_container_id(filepath: &Path) -> Result<String, Box<dyn std::error::E
     Err(ContainerIdNotFoundError.into())
 }
 
+pub unsafe fn set_cgroup_file(file: String) {
+    // Safety: Must not be called in multi-threaded contexts
+    CGROUP_PATH = Some(file)
+}
+
 pub fn get_container_id() -> Option<&'static str> {
     // cache container id in a static to avoid recomputing it at each call
     lazy_static! {
-        static ref CONTAINER_ID: Option<String> = extract_container_id(Path::new(CGROUP_PATH)).ok();
+        // Safety: we assume set_cgroup_file is not called when it shouldn't
+        static ref CONTAINER_ID: Option<String> = unsafe { extract_container_id(Path::new(CGROUP_PATH.as_ref().map(String::as_str).unwrap_or(&DEFAULT_CGROUP_PATH))).ok() };
     }
     CONTAINER_ID.as_deref()
 }
--- a/ddcommon/src/container_id.rs
+++ b/ddcommon/src/container_id.rs
@@ -8,6 +8,7 @@ use std::fmt;
 use std::fs::File;
 use std::io::{BufRead, BufReader};
 use std::path::Path;
+use std::path::PathBuf;
 
 /* Extract container id from /proc/self/group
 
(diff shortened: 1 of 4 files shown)


ddcommon/src/container_id.rs
```diff
@@ -85,10 +89,27 @@ fn extract_container_id(filepath: &Path) -> Result<String, Box<dyn std::error::E
...
 
+/// # Safety
+/// Must not be called in multi-threaded contexts
+pub unsafe fn set_cgroup_file(file: String) {
+    TESTING_CGROUP_PATH = Some(file)
+}
+
+fn get_cgroup_path() -> PathBuf {
+    // Safety: we assume set_cgroup_file is not called when it shouldn't
+    if let Some(path) = unsafe { TESTING_CGROUP_PATH.as_ref() } {
```
pawelchcki, line 100: @bwoebi would making this an env variable work as well ?

bwoebi: An env var would be okay as well I think.

ddtelemetry-ffi/Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2021"
 [lib]
 # LTO is ignored if "lib" is added as crate type
```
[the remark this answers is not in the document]
bwoebi: Yeah. The problem is: if this is _not_ lib, then you cannot use this as a dependency to another crate. (The use case being: coalesce all dependencies into a single static lib in php and link the tracer against that once.)

ddcommon/src/container_id.rs
ivoanjo: To be honest, I'm not a big fan of the env variable approach. The current approach adds a new API, which the application needs to opt in to for using (e.g. when it's running the tests), and I like that a bit more than using an environment variable where suddenly we've created a public API that any user of libdatadog will "inherit" and make available in product because the environment variable can be set by customers.

So... I dislike coming into reviews and giving advice that goes in the exact other direction, but @pawelchcki can I convince you that an env variable may not be the best approach here?

ddcommon/src/container_id.rs
```diff
@@ -85,10 +89,27 @@ fn extract_container_id(filepath: &Path) -> Result<String, Box<dyn std::error::E
...
+    }
+}
+
 pub fn get_container_id() -> Option<&'static str> {
     // cache container id in a static to avoid recomputing it at each call
+
     lazy_static! {
-        static ref CONTAINER_ID: Option<String> = extract_container_id(Path::new(CGROUP_PATH)).ok();
+        static ref CONTAINER_ID: Option<String> =
+            extract_container_id(get_cgroup_path().as_path()).ok();
```
ivoanjo, line 112: Suggestion: Why not make the path an argument to this function? Callers that don't care about it can pass in `get_container_id(DEFAULT_CGROUP_PATH)`; callers that do can pass another one.

This way we eliminate the extra state, but allow the behavior to be configurable. If Rust had default values for arguments, this would be a textbook great use for them -- `get_container_id()` for the default and `get_container_id(custom)` for other use-cases, but as we say in Portugal, "if you don't have a hunting dog, use a cat" ;)

ivoanjo: Fair enough. (Sigh... Rust is so low-level lol, it's definitely not for me)

ddcommon/src/container_id.rs
```diff
@@ -85,10 +89,27 @@ fn extract_container_id(filepath: &Path) -> Result<String, Box<dyn std::error::E
...
+        Path::new(path.as_str()).into()
+    } else {
+        Path::new(DEFAULT_CGROUP_PATH).into()
+    }
+}
+
 pub fn get_container_id() -> Option<&'static str> {
     // cache container id in a static to avoid recomputing it at each call
+
     lazy_static! {
```
r1viollet, line 110: noob question: lazy_static! guarantees thread safety ?

ddcommon/src/container_id.rs
pawelchcki: Those were some good points - @ivoanjo consider me convinced.

ddtelemetry-ffi/Cargo.toml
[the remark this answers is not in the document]
pawelchcki: Removed the comment :) - when Rust 1.64 lands in Alpine we'll be able to address this challenge differently as well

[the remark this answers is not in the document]
pawelchcki: https://github.com/rust-lang/cargo/pull/10838

ddcommon/src/container_id.rs
pawelchcki: Yes IIRC it uses CellOnce underneath to make this threadsafe.

bwoebi (approved):
Looks good to me.

r1viollet (approved):
LGTM
Very cool that we are enabling more sharing for PHP folks.

pawelchcki:
Thanks for the review ! :bow:

ddtelemetry-ffi/Cargo.toml
```diff
@@ -7,9 +7,7 @@ version = "0.9.1"
 edition = "2021"
 
 [lib]
-# LTO is ignored if "lib" is added as crate type
-# cf. https://github.com/rust-lang/rust/issues/51009
-crate-type = ["staticlib", "cdylib"]
+crate-type = ["lib", "staticlib", "cdylib"]
```
morrisonlevi, line 10: Reminder, this _will_ break your LTO and bloat binary size.

PR #65, opened 2022-10-17
Build ffi on alpine

pawelchcki:
# What does this PR do?

Add alpine and aarm64 builds via docker bake. 
For now in the CI I only added Alpine ffi build as the aarm64 one was ooming. 

Usage:
```
$ docker buildx bake debian-build-aarch64
$ docker buildx bake alpine-build

$ ls build/
aarch64-unknown-linux-gnu  x86_64-unknown-linux-gnu
```

It uses more or less the same Dockerfile as can be found in our release repository. 

Dockerfile was crafted to maximize caching - especially for local builds. In CI caching is handled by `buildkit` (directly stored build steps in github cache)

# Motivation

Alpine builds are not able to generate ffi images properly and this PR is a first step to fix that problem


ivoanjo:
Thanks for making our github setup more awesome :)

pawelchcki:
Thanks for quick review :bow: ! :)

PR #70, opened 2022-11-07
Add AAS metadata in event.json file

gleocadie:
# What does this PR do?

This PR adds AAS metadata as tags in the `event.json` file instead of adding them in the form body.

# Motivation

In a previous https://github.com/DataDog/libdatadog/pull/50, the metadata were added in the form body but with the 2.4 endpoint version, tags are added in the `event.json` file.

# Additional Notes

# How to test the change?

Same as the previous PR (https://github.com/DataDog/libdatadog/pull/50)

### DIFF
--- a/profiling/src/exporter/mod.rs
+++ b/profiling/src/exporter/mod.rs
@@ -187,15 +187,12 @@ impl ProfileExporter {
                     ("aas.site.type", aas_metadata.get_site_type()),
                     ("aas.subscription.id", aas_metadata.get_subscription_id()),
                 ];
-                aas_tags
-                    .into_iter()
-                    .for_each(|(name, value)| match Tag::new(name, value) {
-                        Ok(tag) => {
-                            tags_profiler.push_str(tag.as_ref());
-                            tags_profiler.push(',');
-                        }
-                        _ => (),
-                    });
+                aas_tags.into_iter().for_each(|(name, value)| {
+                    if let Ok(tag) = Tag::new(name, value) {
+                        tags_profiler.push_str(tag.as_ref());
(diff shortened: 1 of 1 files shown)


chrisnas (approved):
LGTM

profiling/src/exporter/mod.rs
```diff
@@ -209,13 +187,37 @@ impl ProfileExporter {
                     ("aas.site.type", aas_metadata.get_site_type()),
                     ("aas.subscription.id", aas_metadata.get_subscription_id()),
                 ];
-                aas_tags
-                    .into_iter()
-                    .for_each(|(name, value)| form.add_text(name, value));
+                aas_tags.into_iter().for_each(|(name, value)| {
+                    tags_profiler.push_str(Tag::new(name, value).unwrap().as_ref());
```
morrisonlevi, line 191: On this particular unwrap, the tag hasn't been validate yet, has it? This unwrap could panic if for some reason it contains something that isn't considered a valid tag, right?

gleocadie: Oh you are right.
Will fix it.

morrisonlevi (approved):
Looks good to me! I recommend asking someone on the backend to verify that the data is coming in cleanly like they want, since I don't know about that aspect ^_^

gleocadie:
> Looks good to me! I recommend asking someone on the backend to verify that the data is coming in cleanly like they want, since I don't know about that aspect ^_^

good idea, I'll ping @AlexJF

AlexJF (approved):
LGTM from a backend perspective!

PR #71, opened 2022-11-14
Add Profiled endpoints statistics

gleocadie:
# What does this PR do?

This PR allows to send the profiled endpoints count in the `event.json` file.

The counts should be sent in the `event.json` file:
```
{
 "start":"<START DATE>", 
 "end":"<END DATE>",
 "attachments":["<ATTACHMENT1>", "<ATTACHMENT2>"],
 "tags_profiler":"<PROFILER TAGS>",
 "endpoint_counts": {
   "/api/endpoint": 10,
   "/api/different-endpoint": 35
 },
 "family":"<FAMILY>",
 "version":"4"
}
```

# Motivation

The goal is to be able to see how much resources are used by an average individual call to an endpoint. From those statistics, the backend will do the computation.

# How to test the change?

For .NET, I compiled the code and generated the NuGet to be used by the profiler. I sniffed the HTTP request and checked that:
- Adding endpoints, we can see the `event.json` containing the endpoints stats.
- Not adding endpoints, we can see an empty list.


ddcommon/src/profiled_endpoints.rs
```diff
@@ -0,0 +1,72 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
```
ivoanjo, line 1: I'm curious about the decision to add this new feature to common, rather than under profiling. Did you consider having it all under profiling? Is there any constraint that makes it hard to do so?

ddcommon/src/profiled_endpoints.rs
```diff
@@ -0,0 +1,72 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+use std::ops::AddAssign;
+
+use indexmap::IndexMap;
```
ivoanjo, line 6: There's a few other places already where we use `IndexMap` as well. I'm curious, when is it worth using `IndexMap` over rust's `HashMap`?

ivoanjo (commented):
Left a few notes, and I think this is shaping up well!

ddcommon/src/profiled_endpoints.rs
gleocadie: To be honest, I did not know about `HashMap` and I just followed the usage in the repository

profiling/src/profile/mod.rs
```diff
@@ -463,7 +469,13 @@ impl Profile {
...
         let mut buffer: Vec<u8> = Vec::new();
         profile.encode(&mut buffer)?;
-        Ok(EncodedProfile { start, end, buffer })
+
+        Ok(EncodedProfile {
+            start,
+            end,
+            buffer,
+            endpoints_stats: self.endpoints.stats.clone().into(),
+        })
```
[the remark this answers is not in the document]
gleocadie: >Is there any limitation into exposing the actual ProfiledEndpointsStats opaque map-like object over the FFI?

ddcommon/src/profiled_endpoints.rs
gleocadie: Good question.
No constraint, to be honest I'm not used to the folder structure and I hope this PR would help in getting more insights :)

profiling/src/profile/mod.rs
```diff
@@ -431,6 +435,8 @@ impl Profile {
         self.endpoints
             .mappings
             .insert(interned_span_id, interned_endpoint);
+
+        self.endpoints.stats.add_endpoint(endpoint.to_string())
     }
```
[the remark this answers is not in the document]
gleocadie: Hum ok, I was under the impression that `profiling endpoint` and reporting the statistics of profiled endpoints were tightly coupled.

For .NET, if profiling endpoints is deactivated, we won't be able to provide the counts. But if other (like Ruby) can and will different both, I can split this into 2 functions.

> Am I getting it right? Because... this may be slightly problematic, at least for Ruby, I think. Which I don't necessarily see as a blocker, since I can opt out. Here's the issue I'm seeing: this design assumes that add_endpoint will be called at most once for a given <local_root_span_id, endpoint> pair.

You are right, my assumption (at least for .NET), is that `add_endpoint` is called once. Generally, it's called when the trace is closed (when we know the endpoint name)

But I agree, it's  be good to dissociate those 2 behaviors.

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
gleocadie: When I try:
```
#[repr(C)]
pub struct EncodedProfile {
    start: Timespec,
    end: Timespec,
    buffer: ddcommon_ffi::Vec<u8>,
    endpoints_stats: profiles::profiled_endpoints::ProfiledEndpointsStats,
}
```
I get this: 
image

ddcommon/src/profiled_endpoints.rs
ivoanjo: Hah, good question! Any insights @morrisonlevi? Rust noobs here ahaha

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
ivoanjo: My rust-foo sucks as usual but I think it's because the representation of the `IndexMap` is not ffi-able and we're returning it "by value". I think the trick is returning a pointer to it instead, something along the lines of (incomplete example):

```diff
diff --git a/profiling-ffi/src/profiles.rs b/profiling-ffi/src/profiles.rs
index b160731..64a002d 100644
--- a/profiling-ffi/src/profiles.rs
+++ b/profiling-ffi/src/profiles.rs
@@ -4,6 +4,7 @@
 use crate::Timespec;
 use datadog_profiling::profile as profiles;
 use ddcommon::profiled_endpoints::ProfiledEndpointStats;
+use ddcommon::profiled_endpoints::ProfiledEndpointsStats;
 use ddcommon_ffi::slice::{AsBytes, CharSlice, Slice};
 use std::convert::{TryFrom, TryInto};
 use std::str::Utf8Error;
@@ -375,6 +376,7 @@ pub struct EncodedProfile {
     end: Timespec,
     buffer: ddcommon_ffi::Vec<u8>,
     endpoints_stats: ddcommon_ffi::Vec<ProfiledEndpointStats>,
+    endpoints_stats_v2: *const ddcommon::profiled_endpoints::ProfiledEndpointsStats,
 }

 impl From<datadog_profiling::profile::EncodedProfile> for EncodedProfile {
@@ -383,12 +385,14 @@ impl From<datadog_profiling::profile::EncodedProfile> for EncodedProfile {
         let end = value.end.into();
         let buffer = value.buffer.into();
         let endpoints_stats = value.endpoints_stats.into();
+        let endpoints_stats_v2 = Box::into_raw(Box::new(value.endpoints_stats_v2));

         Self {
             start,
             end,
             buffer,
             endpoints_stats,
+            endpoints_stats_v2,
         }
     }
 }
diff --git a/profiling/src/profile/mod.rs b/profiling/src/profile/mod.rs
index b7ab9e9..f8d242f 100644
--- a/profiling/src/profile/mod.rs
+++ b/profiling/src/profile/mod.rs
@@ -211,6 +211,7 @@ pub struct EncodedProfile {
     pub end: SystemTime,
     pub buffer: Vec<u8>,
     pub endpoints_stats: Vec<ProfiledEndpointStats>,
+    pub endpoints_stats_v2: ProfiledEndpointsStats,
 }

 impl Endpoints {
@@ -475,6 +476,7 @@ impl Profile {
             end,
             buffer,
             endpoints_stats: self.endpoints.stats.clone().into(),
+            endpoints_stats_v2: self.endpoints.stats.clone(),
         })
     }
```

Ping me if you get stuck, we can get stuck together for more fun! ;)

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
ivoanjo: > Hum ok, I was under the impression that profiling endpoint and reporting the statistics of profiled endpoints were tightly coupled.

Not quite, as far as I understand it. Maybe @mar-kolya can chime in as well, but my understanding is that the statistics are supposed to represent what the tracer is seeing, not the profiling.

I believe the objective here is if in the space of 1 minute, the tracer sees 1000 traces on a given endpoint, and the profiler while sampling only sees 100 of those traces, we still want this counter to be 1000. (I'm not sure what the expected semantics are for traces that start before the 1-minute profile window, or those that extend beyond it).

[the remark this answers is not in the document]
mar-kolya: Hmm... Let me try to answer this to the best of my understanding.

@gleocadie Had asked me: should profiler provide these counts if endpoint profiler is disabled. And the answer to this - it largely doesn't matter. It is fine to have them but as things stand currently they will not be used.

But it looks like the question that is being discussed here may be different: should these counts include 'all' endpoint calls or only ones that are seen by the profiler - and the answer here is that these counts should include *all* andpoint calls. Even ones never sampled by the profiler. And even ones destined to be dropped by the tracer.

Does this make sense?

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
gleocadie: Nice I'll give it a try. Thx

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
gleocadie: Yeah that makes sense. Thx.

ddcommon/src/profiled_endpoints.rs
gleocadie: ✔️

profiling-ffi/src/profiles.rs
```diff
@@ -368,19 +369,47 @@ pub unsafe extern "C" fn ddog_Profile_set_endpoint<'a>(
...
+/// # Arguments
+/// * `profile` - a reference to the profile that will contain the samples.
+/// * `endpoint` - the value of the endpoint label to add for matching samples.
+///
+/// # Safety
+/// The `profile` ptr must point to a valid Profile object created by this
+/// module.
+/// This call is _NOT_ thread-safe.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_Profile_add_endpoint_count(
```
ivoanjo, line 383: Minor: Do you mind if we call this `ddog_prof_Profile_add_endpoint_count`? I'm working on a PR to namespace-prefix the profiling APIs and this would be one less that would need the rename :)

See also [libdatadog API naming review doc](https://docs.google.com/spreadsheets/d/1iuuSCJ0NcSOpo2RVvkQ8f8ugz2XlEbbpFZLVOg6WioE/edit#gid=0) (internal link only, sorry!)

ddcommon/src/profiled_endpoints.rs
morrisonlevi: Use `IndexMap` if you need to refer to the inserted thing by number, like in pprof. That's why you see a lot of it it. If you don't need this guarantee, you can use the standard `HashMap`.

profiling/src/profile/profiled_endpoints.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+impl From<IndexMap<String, i64>> for ProfiledEndpointsStats {
+    fn from(count: IndexMap<String, i64>) -> Self {
+        ProfiledEndpointsStats { count }
+    }
+}
+
+impl ProfiledEndpointsStats {
+    pub fn add_endpoint(&mut self, endpoint_name: String) {
```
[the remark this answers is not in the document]
gleocadie: I was thinking about that too, but was not sure if it would be of interest.

profiling-ffi/src/profiles.rs
gleocadie: no pb ;) will do

profiling/src/profile/profiled_endpoints.rs
```diff
@@ -0,0 +1,53 @@
...
+
+use std::ops::AddAssign;
+
+use indexmap::IndexMap;
+use serde::ser::SerializeMap;
+use serde::Serialize;
+
+#[derive(Default, PartialEq, Eq, Debug, Clone)]
+pub struct ProfiledEndpointsStats {
+    count: IndexMap<String, i64>,
```
morrisonlevi, line 12: This could be a regular HashMap, yeah? You don't use it by index anywhere, as far as I can tell?

profiling-ffi/src/profiles.rs
```diff
@@ -368,19 +369,47 @@ pub unsafe extern "C" fn ddog_Profile_set_endpoint<'a>(
     profile.add_endpoint(local_root_span_id, endpoint);
 }
 
+/// Count the number of times an endpoint has been seen.
+///
+/// # Arguments
+/// * `profile` - a reference to the profile that will contain the samples.
+/// * `endpoint` - the value of the endpoint label to add for matching samples.
```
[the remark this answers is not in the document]
gleocadie: ✔️

profiling-ffi/src/profiles.rs
gleocadie: ✔️

profiling/src/profile/profiled_endpoints.rs
gleocadie: ✔️

profiling/src/profile/profiled_endpoints.rs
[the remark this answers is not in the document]
gleocadie: ✔️

profiling-ffi/src/exporter.rs
```diff
@@ -180,6 +181,7 @@ pub unsafe extern "C" fn profile_exporter_build(
     end: Timespec,
     files: Slice<File>,
     additional_tags: Option<&ddcommon_ffi::Vec<Tag>>,
+    endpoints_stats: Option<&profiled_endpoints::ProfiledEndpointsStats>,
```
ivoanjo, line 184: An earlier version used a `Box` here, and this has now become a `&`. I'm actually not sure what the difference is -- @morrisonlevi is this fine in this situation? (Asking just in case)

gleocadie: We already do the same, for example profile or exporter object. 
- We create the pointer (with `Box`)
- Pass it to the library
- The library do calls passing the profiler/exporter, we use `&` instead
- In the `free` or `drop`, we box it again to release the memory

ivoanjo: Thanks for checking!

ddcommon/src/profiled_endpoints.rs
gleocadie: Use of `HashMap` done

profiling/src/profile/profiled_endpoints.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use serde::ser::SerializeMap;
+use serde::Serialize;
+
+#[derive(Default, PartialEq, Eq, Debug, Clone)]
+pub struct ProfiledEndpointsStats {
+    count: HashMap<String, i64>,
+}
+
+impl Serialize for ProfiledEndpointsStats {
```
paullegranddc, line 14: I think you could use `#[serde(transparent)]` https://serde.rs/container-attrs.html#transparent  on the struct instead of implementing serialization

profiling/src/profile/mod.rs
```diff
@@ -433,6 +438,12 @@ impl Profile {
             .insert(interned_span_id, interned_endpoint);
     }
 
+    pub fn add_endpoint_count(&mut self, endpoint: Cow<str>, value: i64) {
```
paullegranddc, line 441: If you're going to convert the endpoint to a `String` anyway, you should probably take a String as argument

profiling/src/profile/mod.rs
```diff
@@ -433,6 +438,12 @@ impl Profile {
...
+        self.endpoints
+            .stats
+            .add_endpoint_count(endpoint.to_string(), value);
```
paullegranddc, line 444: Here you want [into_owned](https://doc.rust-lang.org/std/borrow/enum.Cow.html#method.into_owned) I think.
To owned will no copy the data if the Cow is already of the owned variant, whereas to_string calls the fmt::Display traits which always allocates and clones the data.

profiling-ffi/src/profiles.rs
```diff
@@ -371,19 +372,47 @@ pub unsafe extern "C" fn ddog_prof_Profile_set_endpoint<'a>(
...
+    let endpoint = endpoint.to_utf8_lossy();
+    profile.add_endpoint_count(endpoint, value);
+}
+
 #[repr(C)]
 pub struct EncodedProfile {
     start: Timespec,
     end: Timespec,
     buffer: ddcommon_ffi::Vec<u8>,
+    endpoints_stats: Box<profiled_endpoints::ProfiledEndpointsStats>,
```
paullegranddc, line 400: Is the Box here needed to keep profiled_endpoints as an opaque pointer because it's not repr(C)-able?

profiling/src/profile/mod.rs
```diff
@@ -208,6 +211,7 @@ pub struct EncodedProfile {
     pub start: SystemTime,
     pub end: SystemTime,
     pub buffer: Vec<u8>,
+    pub endpoints_stats: Box<ProfiledEndpointsStats>,
```
paullegranddc, line 214: I don't see the need for Boxing here

paullegranddc: Actually I could see a use for it if you expect the `ProfiledEndpointsStats` to be empty most of the time.

An empty Hashmap is [48 bytes](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=89500f98178f6ca926bd25e3b887e63f) but if you use `Option<Box<ProfiledEndpointsStats>>`, in case the ProfiledEndpointsStats is empty you can put a None variant and have it take only 8 bytes instead.

profiling-ffi/src/profiles.rs
gleocadie: That and also release the profiled_endpoints when the encode profile is release

profiling/src/profile/profiled_endpoints.rs
gleocadie: I may miss something, I get this error
image
and cannot find a way to fix it. Do you have any idea?

profiling/src/profile/mod.rs
gleocadie: > I don't see the need for Boxing here since this is not a C struct

I'm not sure I understand this comment.

paullegranddc: > I'm not sure I understand this comment.

The profiling::EncodedProfile struct is the dual to the the ffi::EncodedProfile one, but that doesn't support repr(C). The `endpoints_stats` field in ffi::EncodedProfile needs to be Boxed, but this one doesn't so you could just have `pub endpoints_stats: ProfiledEndpointsStats` here.

profiling/src/profile/profiled_endpoints.rs
paullegranddc: The struct needs to be `#[derive(Serde)] and `#[serde(transparent)]` is a struct level attribute.

This works for me.

```
#[derive(Default, PartialEq, Eq, Debug, Clone, Serialize)]
#[serde(transparent)]
pub struct ProfiledEndpointsStats {
    count: HashMap<String, i64>,
}
```

gleocadie: ✔️

profiling/src/profile/mod.rs
gleocadie: ✔️

profiling/src/profile/profiled_endpoints.rs
```diff
@@ -0,0 +1,28 @@
...
+
+impl From<HashMap<String, i64>> for ProfiledEndpointsStats {
+    fn from(count: HashMap<String, i64>) -> Self {
+        ProfiledEndpointsStats { count }
+    }
+}
+
+impl ProfiledEndpointsStats {
+    pub fn add_endpoint_count(&mut self, endpoint_name: String, value: i64) {
+        *self.count.entry(endpoint_name).or_insert(0) += value;
```
paullegranddc, line 22: Btw, not a something to fix but more a general reflexion.

The HashMap::entry API needs to take ownership of the key even if it already exists in the HashMap, which requires allocating a String everytime this function is called, but it's possible to do this with copying only if the key doesn't exists using the [raw_entry api](https://doc.rust-lang.org/std/collections/hash_map/enum.RawEntryMut.html#) (which is currently unstable, but available on hashbrown::HashMap which the std uses underneath) allowing to propagate the Cow<str> to here.
So if allocation or heap fragmentation becomes an issue it might be a good idea to consider this change.

gleocadie: Thanks for this.

ddcommon/Cargo.toml
```diff
@@ -18,6 +18,7 @@ futures-util = {version = "0.3.0", default-features = false}
 hex = "0.4"
 hyper = {version = "0.14", features = ["http1", "client", "tcp", "stream"], default-features = false}
 hyper-rustls = {version = "0.23", default-features = false, features = ["native-tokio", "http1", "tls12"]}
+indexmap = "1.8"
```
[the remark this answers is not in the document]
gleocadie: ✔️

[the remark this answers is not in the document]
gleocadie: ✔️

paullegranddc (approved):

ddcommon/Cargo.toml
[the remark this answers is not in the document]
paullegranddc: Wait looking at the diff, I think you just removed it from the dev-deps but not from the general ones?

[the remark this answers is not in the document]
gleocadie: oh good one.

PR #72, opened 2022-11-17
Use rustc-hash's FxHasher for better performance

morrisonlevi:
# What does this PR do?

Changes the hasher used in the profiling crate to use rustc-hash's `FxHasher`.

# Motivation

The default hasher is resistant to HashDoS attacks but is slow. The rustc-hash's FxHasher is the hashing algorithm Firefox uses, and is generally faster while still being decent quality. See:

 - https://github.com/rust-lang/rustc-hash
 - https://crates.io/crates/rustc-hash

# Additional Notes

I have not yet actually tested the performance of this :)

Grégory tested it with .NET and said:

> I profiled an application for 30s with two versions of the .NET profiler:
> - With libdatadog  0.9
> - With  Levi's fix.

> We have function responsible for collecting samples, transforming them to FFI version and adding them to the profile:
> - For the 0.9.0, I spend 23.6% adding the samples to the profile.
> - With Levi fix, I spend 13% of adding the samples to the profile.

So, this seems to do what we expected!
# How to test the change?


gleocadie (commented):
Would it make sense/possible to use the hasher in the profiled_endpoint ?
https://github.com/DataDog/libdatadog/blob/f3a8992ff022ec416859fa6848f729d9aa563c1f/profiling/src/profile/profiled_endpoints.rs#L11

morrisonlevi:
> Would it make sense/possible to use the hasher in the profiled_endpoint ?

Endpoints could be susceptible to HashDos style attacks. The default algorithm is supposed to be a bit more resistant to those kinds of things, so without more data (just trusting what's been said in docs), I would say let's leave that one alone?

gleocadie:
> > Would it make sense/possible to use the hasher in the profiled_endpoint ?
> 
> Endpoints could be susceptible to HashDos style attacks. The default algorithm is supposed to be a bit more resistant to those kinds of things, so without more data (just trusting what's been said in docs), I would say let's leave that one alone?

fair point.

ivoanjo:
I have a simple Ruby profiler benchmark that creates 4 threads with very deep stacks and then just samples them as fast as possible (https://github.com/DataDog/dd-trace-rb/blob/master/benchmarks/profiler_sample_loop.rb). I had not yet updated the benchmark for the new libdatadog-based profiling code, but I did so (PR upcoming) and here's the results of running regular 0.9.0 vs 0.9.0 with fxhasher on my machine:

```
Calculating -------------------------------------
stack collector /Users/ivo.anjo/datadog/libdatadog/build-0.9.0-no-core-foundation-fxhasher/
                          1.715k (± 2.8%) i/s -     17.201k in  10.037444s

Comparison:
stack collector /Users/ivo.anjo/datadog/libdatadog/build-0.9.0-no-core-foundation-fxhasher/:     1715.1 i/s
stack collector /Users/ivo.anjo/datadog/libdatadog/build-0.9.0-no-core-foundation/:     1144.3 i/s - 1.50x  (± 0.00) slower
```

Aka this seems to be a nice win for Ruby!

ivoanjo (approved):
LGTM 👍

PR #74, opened 2022-11-28
Document assumption in Ruby profiler that serialize is safe to be interrupted by a fork

ivoanjo:
# What does this PR do?

This PR documents the assumption from the Ruby profiler that it's safe for the libdatadog serialize operation to be interrupted by a fork.

The intention is not to prevent modification of the API, but to create a stronger link so that if we need to change serialize, we remember that the Ruby profiler needs to be updated.

 # Motivation

I fear that without documenting this link, this piece of information can easily get lost and cause issues in the future.

 # Additional Notes

(Nothing)

 # How to test the change?

Use eyes to detect that documentation has been added ;)


ivoanjo:
We ended up discussing this and documenting the results in a separate (internal) document, so closing for now. Thanks Levi for the insights! :)

PR #75, opened 2022-11-30
Add .pprof to api::Profile helper

morrisonlevi:
# What does this PR do?

- Creates a new type `api::Profile`.
- Implements `TryFrom<&pprof::Profile> for api::Profile`.
- Tests a few samples in given WordPress profile to ensure it's correct.
- Drops the nightly clippy lints. These fail too often. Tracking the latest stable's clippy should be good enough.

# Motivation

This is a precursor PR for benchmarking hashing functions or other changes. This allows us to read in a pprof from disk and get an `api::Profile` from it, which we can use to add to the regular profile API. This allows us to benchmark adding samples, interning strings, etc.

# Additional Notes

I wrote the tests for those samples by hand -- they aren't auto-generated. It's a very slow process.

# How to test the change?

There are no changes to shipped code, so there shouldn't be anything to test.


profiling/src/profile/api.rs
```diff
@@ -120,3 +127,163 @@ pub struct Sample<'a> {
...
     pub labels: Vec<Label<'a>>,
 }
+
+pub struct Profile<'a> {
+    pub duration: Duration,
+    pub period: Option<(i64, ValueType<'a>)>,
+    pub sample_types: Vec<ValueType<'a>>,
+    pub samples: Vec<Sample<'a>>,
+    pub start_time: SystemTime,
+}
```
ivoanjo, line 137: Given the assumption

> There are no changes to shipped code, so there shouldn't be anything to test.

Should this be inside the `mod test`? And perhaps also document the assumption in the code, so that other/newer team members know that this isn't expected to be used outside of libdatadog testing.

profiling/tests/wordpress.pprof
```diff
@@ -0,0 +1,4441 @@
+
```
ivoanjo, line 1: Consider perhaps compressing this file as `pprof.gz` or something OR at least telling github not to render them https://thoughtbot.com/blog/github-diff-supression :)

ivoanjo (approved):
Other than making it clear that the new API is only for testing, I don't see any reason not to merge this in. Here sir, take my 👍

profiling/tests/wordpress.pprof
morrisonlevi: Good idea 👍🏻

profiling/src/profile/api.rs
morrisonlevi: Maybe. I need to see what configuration the `bench` configuration uses -- I know it uses public APIs only, but I'm not sure if it builds with test or not, so that's the only reason I didn't immediately slap it into `#[cfg(test)]`.

PR #77, opened 2022-12-02
build: upgrade to Rust 1.64 and Alpine 3.17

morrisonlevi:
# What does this PR do?

Upgrades Alpine to 3.17, and Rust to 1.64.0.

# Motivation

Alpine 3.17 was released recently. Alpine is the OS that holds us back from upgrading Rust versions, so it's time to try upgrading to Rust 1.64.0, which is the version of Rust in Alpine 3.17.

# Additional Notes

The PHP profiler has been building with Rust 1.64.0 since datadog/dd-trace-php#1922. No issues.

Additionally, #133 wants to update to at least 1.63.

# How to test the change?

Artifacts are [here](https://gitlab.ddbuild.io/DataDog/libddprof-build/-/jobs/224222900/artifacts/download). Build against these, and test as usual.


.gitlab-ci.yml
```diff
@@ -2,7 +2,7 @@ variables:
   # These are gitlab variables so that it's easier to do a manual deploy
   # If these are set witih value and description, then it gives you UI elements
   DOWNSTREAM_BRANCH:
-    value: "main"
+    value: "levi/rust-1.64"
```
thedavl, line 5: should this be changed back to main?

morrisonlevi:
Currently, there's a static analysis error on Windows:

```
error: constant `TRAMPOLINE_BIN` is never used
 --> spawn_worker\src\lib.rs:4:18
  |
4 | pub(crate) const TRAMPOLINE_BIN: &[u8] =
  |                  ^^^^^^^^^^^^^^
  |
  = note: `-D dead-code` implied by `-D warnings`
```

.gitlab-ci.yml
morrisonlevi: Yes, but not until that downstream PR is merged.

PR #78, opened 2022-12-09
Change `endpoints_count` into `endpoint_counts`

gleocadie:
# What does this PR do?

Rename the field `endpoints_count` into `endpoint_counts`.

# Motivation

This field is used in the backend, but the name expected by the backend is not the one I used.

# Additional Notes

# How to test the change?

Did the same test for the previous PR: run a .NET application with the Tracer and check what's inside the HTTP request.


ivoanjo (approved):
👍 LGTM

PR #80, opened 2022-12-12
Convert local root span id to u64, fix clippy lints

morrisonlevi:
# What does this PR do?

- Fix `clippy::uninlined_format_args` lints.
- Fix unused import `std::path::Path` in `cfg(test)`.
- Converts `local_root_span_id` in `add_endpoint` and related functions to be `u64` instead of a string.
- Removes unnecessary `Cow<str>` for `str` in label.
- Bumps version to v2.0.0.

# Motivation

The profiling backend can now accept numeric labels. This PR converts "local root span id" related parameters to use numbers in the endpoint API.

# Additional Notes

This should have been multiple PRs. I'm sorry.

# How to test the change?

- Change `add_endpoint` to use `u64` instead of a string.
- Change your `local root span id` labels to use the `.num` variant instead of `.str`. Transmute from `u64` to `i64` as necessary. If they aren't changed, then `profile.add` will fail.

And then test as normal. Double-check that Code Hotspots and Endpoint Profiling both work.

### DIFF
--- a/ddtelemetry/src/config.rs
+++ b/ddtelemetry/src/config.rs
@@ -50,7 +50,7 @@ impl FromEnv {
         let agent_host =
             env::var(DD_AGENT_HOST).unwrap_or_else(|_| String::from(DEFAULT_AGENT_HOST));
 
-        format!("http://{}:{}", agent_host, agent_port)
+        format!("http://{agent_host}:{agent_port}")
     }
 
     fn get_intake_base_url() -> String {
@@ -63,9 +63,9 @@ impl FromEnv {
 
         if let Ok(dd_site) = env::var(DD_SITE) {
             if dd_site.is_empty() {
-                format!("{}.{}", PROD_INTAKE_FORMAT_PREFIX, DEFAULT_DD_SITE)
+                format!("{PROD_INTAKE_FORMAT_PREFIX}.{DEFAULT_DD_SITE}")
             } else {
-                format!("{}.{}", PROD_INTAKE_FORMAT_PREFIX, dd_site)
+                format!("{PROD_INTAKE_FORMAT_PREFIX}.{dd_site}")
             }
         } else {
             String::from(STAGING_INTAKE)
@@ -77,9 +77,9 @@ impl FromEnv {
 
         let telemetry_url = if api_key.is_some() {
             let telemetry_intake_base_url = Self::get_intake_base_url();
-            format!("{}{}", telemetry_intake_base_url, DIRECT_TELEMETRY_URL_PATH)
+            format!("{telemetry_intake_base_url}{DIRECT_TELEMETRY_URL_PATH}")
         } else {
-            format!("{}{}", &agent_url, AGENT_TELEMETRY_URL_PATH)
+            format!("{}{AGENT_TELEMETRY_URL_PATH}", &agent_url)
         };
 
         let telemetry_uri = Uri::from_str(&telemetry_url).ok()?;
--- a/profiling/src/profile/mod.rs
+++ b/profiling/src/profile/mod.rs
@@ -1066,7 +1066,7 @@ mod api_test {
     }
 
     #[test]
-    fn root_span_id_label_as_str_or_i64() {
+    fn local_root_span_id_label_as_str_or_i64() {
         let sample_types = vec![
             api::ValueType {
                 r#type: "samples",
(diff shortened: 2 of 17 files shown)


profiling/src/profile/mod.rs
```diff
@@ -488,6 +489,35 @@ impl Profile {
...
+             */
+            let string = match self.strings.get_index(label.str as usize) {
+                None => anyhow::bail!(
+                    "failed to retrieve index {} from the string table when looking for endpoint data",
+                    label.str
+                ),
+                Some(string) => string,
+            };
+            string.as_str().parse()?
+        };
```
ivoanjo, line 511: As this PR changes `add_endpoint` to take a u64 and not a String, I'm not quite sure we need this.

E.g., what clients would use the new numberic `add_endpoint` but still send the labels as strings elsewhere?

morrisonlevi: This is coming from a _label_, and there's nothing that stops a client from attaching the label value as a string instead of a number.

profiling/src/profile/mod.rs
```diff
@@ -1034,4 +1064,82 @@ mod api_test {
...
+        let l2 = s2.labels.get(0).expect("label");
+
+        assert_eq!(
+            serialized_profile
+                .string_table
+                .get(l2.str as usize)
+                .unwrap(),
+            "11"
+        );
+    }
```
morrisonlevi, line 1144: This test is good, thanks Florian! On top of this, we should be able to check that both have the endpoint label too, I think?

profiling/src/profile/mod.rs
ivoanjo: Hmm... I think having the fallback path is a sharper edge, as it makes things work even when they're not set up the way they should.

E.g., if someone does it wrong, they will still get the right output, which is dangerous...

One suggestion is to consider failing serialization and complaining that this label should not be a string. The `serialize` operation already returns a `Result<EncodedProfile, EncodeError>` so clients already need to handle errors, and perhaps modifying `from` to potentially return an error would allow us to report back what's wrong.

morrisonlevi: Okay, I've  started working on fixing this. There will be two changes:

1. At serialization time, verify that any `local root span id` labels use `.num`.
2. When the sample is added, verify that any `local root span id` labels use `.num`.

morrisonlevi:
I confirmed that Code Hotspots and Endpoint profiling work for PHP with this branch.

ivoanjo:
:tada: I too was able to check that Code Hotspots and Endpoint profiling work fine for Ruby with this branch.

ivoanjo (commented):
Left a few notes! I ran out of time and didn't fully finish my review, will come back to it on monday :)

profiling/src/profile/mod.rs
```diff
@@ -312,28 +312,41 @@ impl Profile {
...
+                    "the label \"local root span id\" must not be 0"
+                )
+            }
+
+            labels.push(Label {
+                key,
+                str,
+                num: label.num,
+                num_unit,
+            });
```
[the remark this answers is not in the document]
morrisonlevi: It doesn't bother me 🤷🏻

profiling/src/profile/mod.rs
```diff
@@ -312,28 +312,41 @@ impl Profile {
...
-                }
-            })
-            .collect();
+        let mut labels = Vec::with_capacity(sample.labels.len());
+        for label in sample.labels.iter() {
+            let key = self.intern(label.key);
+            let str = label.str.map(|s| self.intern(s)).unwrap_or(0);
+            let num_unit = label.num_unit.map(|s| self.intern(s)).unwrap_or(0);
+
+            if key == self.endpoints.local_root_span_id_label {
```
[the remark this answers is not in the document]
morrisonlevi: Good point, good thinking!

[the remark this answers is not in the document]
morrisonlevi: Fixed in [https://github.com/DataDog/libdatadog/pull/80/commits/e1314557253d30112298f20ccd0900ed05f911f1](https://github.com/DataDog/libdatadog/pull/80/commits/e1314557253d30112298f20ccd0900ed05f911f1) by eagerly storing the label. Although this isn't "ideal" in the smallest profile sense if the feature is unused, we honestly expect it to be used by most profiles, and it's not wasting that much (and also reduces some tiny runtime cost to do it on startup).

profiling/src/profile/mod.rs
```diff
@@ -509,27 +558,28 @@ impl From<&Profile> for pprof::Profile {
...
-                        break;
+                // There _should_ only be one local root span id label, but it's not enforced.
+                let lrsi_labels: Vec<_> = sample
+                    .labels
+                    .iter()
+                    .filter(|label| label.key == profile.endpoints.local_root_span_id_label)
+                    .map(Clone::clone) // Need to clone to break borrows
+                    .collect();
+
+                for label in lrsi_labels {
```
ivoanjo, line 569: Minor: I somewhat wonder if adding a boolean to track that the label wasn't repeated in `Profile::add` is simpler than having to write extra code to handle the case it is here.

profiling/src/profile/mod.rs
```diff
@@ -488,10 +508,39 @@ impl Profile {
...
     }
+
+    /// Fetches the endpoint information for the label. There may be errors,
+    /// but there may also be no endpoint information for a given endpoint, or
+    /// it may not be an local root span id label. Hence, the return type of
+    /// Result<Option<_>, _>.
+    fn get_endpoint_for_label(&self, label: &Label) -> anyhow::Result<Option<i64>> {
+        if label.key != self.endpoints.local_root_span_id_label {
+            return Ok(None);
+        }
```
ivoanjo, line 519: Minor: Should this also be a an `anyhow::ensure!`?

```rust
anyhow::ensure!(
  label.key == self.endpoints.local_root_span_id_label,
  "cannot look up endpoint for a label that does not contain the local root span id, got string id {}",
  label.str
)
```

profiling/src/profile/mod.rs
```diff
@@ -1034,4 +1098,106 @@ mod api_test {
 
         assert_eq!(endpoints_stats, expected_endpoints_stats);
     }
+
+    #[test]
+    fn local_root_span_id_label_as_i64() {
```
ivoanjo, line 1103: Minor: Should we perhaps test using u64 and with one of the ids being UINT_MAX, just to doublecheck that it's handled correctly?

ivoanjo (approved):
:+1: LGTM, thanks for this tiny but really needed improvement :)

ddtelemetry/src/config.rs
```diff
@@ -77,9 +77,9 @@ impl FromEnv {
 
         let telemetry_url = if api_key.is_some() {
             let telemetry_intake_base_url = Self::get_intake_base_url();
-            format!("{}{}", telemetry_intake_base_url, DIRECT_TELEMETRY_URL_PATH)
+            format!("{telemetry_intake_base_url}{DIRECT_TELEMETRY_URL_PATH}")
         } else {
-            format!("{}{}", &agent_url, AGENT_TELEMETRY_URL_PATH)
+            format!("{}{AGENT_TELEMETRY_URL_PATH}", &agent_url)
```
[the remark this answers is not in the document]
paullegranddc: No, the inline format values can only contain simple identifiers https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html

For the sake of uniformity, maybe named arguments would be better
```suggestion
            format!("{agent_url}{AGENT_TELEMETRY_URL_PATH}", agent_url=&agent_url)
```

profiling/src/profile/mod.rs
morrisonlevi: At the moment it's called on all labels. We could change that but it doesn't really change too much, because it's just going to shift that code to the caller, and we can't short-circuit at the moment anyway because labels aren't unique.

PR #81, opened 2022-12-20
Add metrics.json support

gleocadie:
# What does this PR do?

This PR adds support for `metrics.json` file.

# Motivation

In the .NET profiler, we would like to send `metrics.json` file to the backend. The goal is to unlock this capability for other profilers too.

# How to test the change?

To be done with the .NET profiler.

### DIFF
--- a/ddcommon-ffi/src/metrics.rs
+++ b/ddcommon-ffi/src/metrics.rs
@@ -1,0 +1,76 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2022-Present Datadog, Inc.
+
+use crate::{slice::AsBytes, slice::CharSlice};
+use ddcommon::metric::Metric;
+
+#[must_use]
+#[no_mangle]
+pub extern "C" fn ddog_Vec_Metric_new() -> crate::Vec<Metric> {
+    crate::Vec::default()
+}
+
+#[no_mangle]
+pub extern "C" fn ddog_Vec_Metric_drop(_: crate::Vec<Metric>) {}
+
+#[repr(C)]
+pub enum PushMetricResult {
+    Ok,
+    Err(crate::Vec<u8>),
+}
+
+#[no_mangle]
+pub extern "C" fn ddog_Vec_Metric_PushResult_drop(_: PushMetricResult) {}
+
+/// Creates a new Tag from the provided `key` and `value` by doing a utf8
+/// lossy conversion, and pushes into the `vec`. The strings `key` and `value`
+/// are cloned to avoid FFI lifetime issues.
+///
+/// # Safety
+/// The `vec` must be a valid reference.
+/// The CharSlices `key` and `value` must point to at least many bytes as their
+/// `.len` properties claim.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_Vec_Metric_push(
+    vec: &mut crate::Vec<Metric>,
+    key: CharSlice,
+    value: f32,
+) -> PushMetricResult {
+    let name = key.to_utf8_lossy().into_owned();
+    match Metric::new(name, value) {
+        Ok(metric) => {
+            vec.push(metric);
+            PushMetricResult::Ok
+        }
+        Err(err) => PushMetricResult::Err(err.as_bytes().to_vec().into()),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::metrics::*;
+
+    #[test]
+    fn empty_metric_name() {
+        unsafe {
+            let mut metrics = ddog_Vec_Metric_new();
+            let result = ddog_Vec_Metric_push(&mut metrics, CharSlice::from(""), 42.);
+            assert_eq!(metrics.len(), 0);
+            assert!(!matches!(result, PushMetricResult::Ok));
+        }
+    }
+
+    #[test]
+    fn test_get() {
+        unsafe {
+            let mut metrics = ddog_Vec_Metric_new();
+            let metric_name = "my_metric_name";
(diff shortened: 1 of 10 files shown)


profiling/src/exporter/mod.rs
```diff
@@ -146,15 +146,16 @@ impl ProfileExporter {
             tags,
         })
     }
-
     /// Build a Request object representing the profile information provided.
+    #[allow(clippy::too_many_arguments)]
```
gleocadie, line 150: this is just for now (PR-time) to get your insight on this:
should I group `start` and `end` into a struct to it will reduce the number of parameters?
I could also group `additional_tags`, `endpoint_counts` and `metrics` into a struct `additional_info` ?

ddcommon-ffi/src/metrics.rs
```diff
@@ -0,0 +1,76 @@
...
+///
+/// # Safety
+/// The `vec` must be a valid reference.
+/// The CharSlices `key` and `value` must point to at least many bytes as their
+/// `.len` properties claim.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_Vec_Metric_push(
+    vec: &mut crate::Vec<Metric>,
+    key: CharSlice,
+    value: f32,
```
morrisonlevi, line 37: I don't know the answer: should this be `f64`? In PHP, tracer metrics are doubles, but that may because the language doesn't offer anything finer, I don't know.

ddcommon/src/metric.rs
```diff
@@ -0,0 +1,31 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Metric")
+            .field("name", &self.name)
+            .field("value", &self.value)
+            .finish()
+    }
+}
+
+impl Metric {
+    pub fn new(name: String, value: f32) -> Result<Self, Cow<'static, str>> {
```
morrisonlevi, line 24: This doesn't ever use `Cow::Owned`, so it could be `Result<Self,  &'static str>`.

profiling-ffi/src/exporter.rs
```diff
@@ -183,6 +184,7 @@ pub unsafe extern "C" fn ddog_prof_Exporter_Request_build(
     files: Slice<File>,
     additional_tags: Option<&ddcommon_ffi::Vec<Tag>>,
     endpoints_stats: Option<&profiled_endpoints::ProfiledEndpointsStats>,
+    metrics: Option<&ddcommon_ffi::Vec<Metric>>,
```
morrisonlevi, line 187: Why not send it as part of the `files` slice?

morrisonlevi (commented):
Thanks! I have some questions.

profiling-ffi/src/exporter.rs
gleocadie: can you elaborate on this please ?

From what I understand, it would mean that we will duplicate `code` in different profilers (creating a file with the metric, adding it to the `files` slice...) for `metrics.json` file which is not specific to a profiler (I mean, profilers may or may not send metrics, but this json file is a `standard` defined in the backend that profilers could take advantage on).

ddcommon-ffi/src/metrics.rs
gleocadie: you are right. Will fix

ddcommon/src/metric.rs
gleocadie: fixed

profiling-ffi/src/exporter.rs
gleocadie: Another thing using the `files` slice, it's not possible to give a `content-type`.

ivoanjo: I guess there's a trade-off here, that may be different for every language.

For instance, for Ruby, I think it's actually going to be roughly as much code between using Ruby to generate JSON, or using the proposed API (perhaps even a bit more C code for this API). So it's not _that_ useful for Ruby versus Ruby directly emitting the `json` and including it in the files -- I'd probably go with that approach instead of using this.

But if it's useful for .net, since the complexity is also low, I don't see why not -- it's not like we can't change our minds later and deprecate this if we really want to.

ddcommon-ffi/src/metrics.rs
```diff
@@ -0,0 +1,76 @@
...
+) -> PushMetricResult {
+    let name = key.to_utf8_lossy().into_owned();
+    match Metric::new(name, value) {
+        Ok(metric) => {
+            vec.push(metric);
+            PushMetricResult::Ok
+        }
+        Err(err) => PushMetricResult::Err(err.as_bytes().to_vec().into()),
+    }
+}
```
ivoanjo, line 47: I'm a bit hesitant to add this to ddcommon. It has the potential to become common, but it's also possible that (for instance) telemetry will want more out of a metric, and thus will need to create a different one anyway.

So I would suggest considering moving this into profiling for the time being :)

ivoanjo (approved):
Left a few notes, but overall it LGTM :+1:

ddcommon/src/metric.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+impl Debug for Metric {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Metric")
+            .field("name", &self.name)
+            .field("value", &self.value)
+            .finish()
+    }
+}
```
morrisonlevi, line 21: I think this can be replaced with a `#[derive(Debug)]` on the struct -- it appears to do exactly the same.

morrisonlevi (commented):
I have a specific suggestion below, but what is `metrics.json`? These aren't Datadog metrics as I'd generally think about them, as there would types associated to each metric type. I assume this is some specific thing for profiling?

PR #84, opened 2023-01-06
Remove unused optional `time` dependency

ivoanjo:
**What does this PR do?**:

This PR tweaks the optional features of the `chrono` crate to avoid pulling in the `time` crate.

**Motivation**:

We got warned by dependabot that the version of `time` pulled in by `chrono` is a deprecated version that should no longer be used.

Since we were not actually making use of any of the features provided by it, we can just remove it.

**Additional Notes**:

(N/A)

**How to test the change?**:

Validate that all tests still pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -123,7 +123,6 @@ dependencies = [
  "libc",
  "num-integer",
  "num-traits",
- "time",
  "winapi",
 ]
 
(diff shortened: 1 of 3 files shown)


ivoanjo:
See https://github.com/DataDog/libdatadog/security/dependabot/4 for details on the dependabot complaint.

gleocadie (approved):
LGTM

PR #88, opened 2023-01-17
fix(profiling)!: ensure a single label per sample has key "local root span id"

morrisonlevi:
# What does this PR do?

This enforces that only a single label on a sample may contain a key of "local root span id". This means a `Profile::add` can now return an error case where it wouldn't before if multiple labels with keys of "local root span id" are sent. This is inherited by the FFI equivalent function as well, which returns 0 for such cases.

# Motivation

I had a conversation with Ivo in #80 and continued over Slack in which he convinced me the code would be overall simpler if we did this. Aside from being non-sensical to have multiple local root span id labels, it's also an implementation optimization.

# Additional Notes

I like having Ivo on the team :)

# How to test the change?

Upgrade, and run tests as normal. Unless you are sending multiple labels with keys of "local root span id", then this will behave the same.

### DIFF
--- a/profiling/src/profile/mod.rs
+++ b/profiling/src/profile/mod.rs
@@ -535,13 +535,14 @@ impl Profile {
     }
 
     /// Fetches the endpoint information for the label. There may be errors,
-    /// but there may also be no endpoint information for a given endpoint, or
-    /// it may not be an local root span id label. Hence, the return type of
-    /// Result<Option<_>, _>.
+    /// but there may also be no endpoint information for a given endpoint.
+    /// Hence, the return type of Result<Option<_>, _>.
     fn get_endpoint_for_label(&self, label: &Label) -> anyhow::Result<Option<i64>> {
-        if label.key != self.endpoints.local_root_span_id_label {
-            return Ok(None);
-        }
+        anyhow::ensure!(
+            label.key == self.endpoints.local_root_span_id_label,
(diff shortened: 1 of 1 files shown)


profiling/src/profile/mod.rs
```diff
@@ -546,42 +572,39 @@ impl TryFrom<&Profile> for pprof::Profile {
...
-
-                for label in lrsi_labels {
-                    if let Some(endpoint_value_id) = profile.get_endpoint_for_label(&label)? {
-                        sample.labels.push(pprof::Label {
+            .map(|(sample, values)| {
+                // Clone the labels, but enrich them with endpoint profiling.
+                let mut labels = sample.labels.clone();
+                if let Some(offset) = sample.local_root_span_id_label_offset {
+                    // Safety: this offset was created internally and isn't be mutated.
+                    let lsri_label = unsafe { sample.labels.get_unchecked(offset) };
```
gleocadie, line 586: nit: `lrsi_label` instead of `lsri_label`

gleocadie (approved):
only a nit, otherwise LGTM

ivoanjo (approved):
:+1: LGTM!

When we initially discussed it, I was thinking that the `local_root_span_id_label_offset` could be a local variable in `::extract_sample_labels`, which would force serialization to still need to iterate the labels. Storing the label is probably a tiny improvement, although since we clone the labels they'd probably still be in the processor cache so it's probably a small one ;)

Either way, I like it, thanks!

PR #89, opened 2023-01-17
feat(profiling)!: send media type with files

morrisonlevi:
# What does this PR do?

Change the profiling `File` types to have a media type aka mime. This allows sending files like `metrics.json` with the right content type.

# Motivation

There was discussion on PR #81 and on Slack around the idea that it might be nicer for current users to send `metrics.json` in the `File` slice. However, there isn't support today for media types, which this fixes.

# Additional Notes

I like working with Grégory :)

# How to test the change?

All `File` objects now have a `mime` field which needs to be added. For encoded profiles, use type `mime::APPLICATION_OCTET_STREAM` or `"application/octet-stream"` as appropriate for which `File` type is being used.

### DIFF
--- a/examples/ffi/exporter.cpp
+++ b/examples/ffi/exporter.cpp
@@ -109,6 +109,7 @@ int main(int argc, char *argv[]) {
 
   ddog_prof_Exporter_File files_[] = {{
       .name = DDOG_CHARSLICE_C("auto.pprof"),
+      .mime = DDOG_CHARSLICE_C("application/octet-stream"),
       .file = ddog_Vec_U8_as_slice(&encoded_profile->buffer),
   }};
 
--- a/profiling-ffi/src/exporter.rs
+++ b/profiling-ffi/src/exporter.rs
@@ -9,6 +9,7 @@ use datadog_profiling::exporter;
 use datadog_profiling::profile::profiled_endpoints;
 use ddcommon::tag::Tag;
 use ddcommon_ffi::slice::{AsBytes, ByteSlice, CharSlice, Slice};
+use exporter::mime;
 use exporter::ProfileExporter;
 use std::borrow::Cow;
 use std::ptr::NonNull;
@@ -45,9 +46,13 @@ pub enum Endpoint<'a> {
     Agentless(CharSlice<'a>, CharSlice<'a>),
 }
 
+/// Represents a file. The `name` and `mime` must be valid UTF-8. The mime
+/// contents should be recognized by [mime::Mime::from_str]. Send encoded
+/// profiles as "application/octet-stream".
 #[repr(C)]
 pub struct File<'a> {
     name: CharSlice<'a>,
+    mime: CharSlice<'a>,
     file: ByteSlice<'a>,
 }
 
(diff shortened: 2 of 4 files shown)


morrisonlevi:
I tried this out just now in the PHP profiler. I only sent the encoded profile (no metrics.json or anything), but profiles came in and seemed to work. If someone can test with `metrics.json` or code provenance or something and check the HTTP payload to have the right mimes, that would be where the value actually comes :)

ivoanjo:
I'm a bit unsure about this being needed...

* Since libdatadog compresses all the files before sending, should we still send mimetypes that are not `application/octet-stream` even though the payload is compressed?
* Ruby has been sending `code-provenance.json` with libdatadog for quite a long time without this change. So I think it works without the mimetype... unless somehow the fallback support for Ruby that we had in the backend at some point is covering up an issue :thinking:

morrisonlevi:
I forgot momentarily about compression. Is that going to screw up `metrics.json`? If so, then if you send certain mime types we can choose not to compress those.

gleocadie:
> * Since libdatadog compresses all the files before sending, should we still send mimetypes that are not `application/octet-stream` even though the payload is compressed?

From what read here https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#:~:text=A%20multipart%20type%20is%20one,sent%20together%20in%20one%20transaction. for a `multipart`, each component can have its own mime type.

> I forgot momentarily about compression. Is that going to screw up metrics.json? If so, then if you send certain mime types we can choose not to compress those.

I do not think it will mess up the `metrics.json`. But I wonder if not adding the mime type won't screw up code that parse and read the HTTP request.

ivoanjo:
> From what read here https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#:~:text=A%20multipart%20type%20is%20one,sent%20together%20in%20one%20transaction. for a multipart, each component can have its own mime type.

> I do not think it will mess up the metrics.json. But I wonder if not adding the mime type won't screw up code that parse and read the HTTP request.

My doubt here is, afaik libdatadog compresses each file individually, so technically every payload is an application/octet-stream, none of them go as text (other than `event.json`, which is not currently compressed):

https://github.com/DataDog/libdatadog/blob/a9725b920bec6cfe13d4bf1d29dab5105c3a892a/profiling/src/exporter/mod.rs#L227-L237

gleocadie:
> > From what read here https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#:~:text=A%20multipart%20type%20is%20one,sent%20together%20in%20one%20transaction. for a multipart, each component can have its own mime type.
> 
> > I do not think it will mess up the metrics.json. But I wonder if not adding the mime type won't screw up code that parse and read the HTTP request.
> 
> My doubt here is, afaik libdatadog compresses each file individually, so technically every payload is an application/octet-stream, none of them go as text (other than `event.json`, which is not currently compressed):
> 
> https://github.com/DataDog/libdatadog/blob/a9725b920bec6cfe13d4bf1d29dab5105c3a892a/profiling/src/exporter/mod.rs#L227-L237

I see. Yeah 🤔 good point.

morrisonlevi:
Intake doesn't care about the content-type of anything other than `event.json`. Closing out as unnecessary.

PR #90, opened 2023-01-18
feat(profiling)!: pass errors through more FFI functions

morrisonlevi:
# What does this PR do?

This changes the return types for FFI functions:
 - `ddog_prof_Profile_add`
 - `ddog_prof_Exporter_Request_build`

Adds new structs:
 - `ddog_prof_Profile_AddResult`
 - `ddog_prof_Exporter_Request_BuildResult`
 
And adds functions to drop them:
 - `ddog_prof_Profile_AddResult_drop`
 - `ddog_prof_Exporter_Request_BuildResult_drop`

Removes a now-unnecessary newtype definition of `struct Request(exporter::Request)` as cbindgen handles the refactored code.

# Motivation

As we will be releasing v2.0 soon, I figured we could ship some other breaking changes that help FFI users know _what_ went wrong, rather than just that something did.

# Additional Notes

Please audit the names as per the recent naming scheme names.

# How to test the change?

Download the artifacts, adjust the code surrounding the two functions listed above, and ensure it all still works correctly.


profiling-ffi/src/exporter.rs
```diff
@@ -169,6 +170,28 @@ unsafe fn into_vec_files<'a>(slice: Slice<'a, File>) -> Vec<exporter::File<'a>>
...
+            ExporterRequestBuildResult::Err(err) => {
+                // Safety: not generally safe but this is for test code.
+                Err(unsafe { String::from_utf8_unchecked(err.into()) })
+            }
+        }
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn ddog_prof_ExporterRequestBuildResult_drop(_result: ExporterRequestBuildResult) {}
```
ivoanjo, line 193: Since we started from `ddog_prof_Exporter_Request_build` I think this should be `ddog_prof_Exporter_Request_BuildResult_drop`

(Similar to how `ddog_prof_Exporter_send` => `ddog_prof_Exporter_SendResult_drop`)

```suggestion
pub extern "C" fn ddog_prof_Exporter_Request_BuildResult_drop(_result: ExporterRequestBuildResult) {}
```

profiling-ffi/src/exporter.rs
```diff
@@ -169,6 +170,28 @@ unsafe fn into_vec_files<'a>(slice: Slice<'a, File>) -> Vec<exporter::File<'a>>
...
+}
+
+#[cfg(test)]
+impl From<ExporterRequestBuildResult> for Result<Box<Request>, String> {
+    fn from(result: ExporterRequestBuildResult) -> Self {
+        match result {
+            ExporterRequestBuildResult::Ok(ok) => Ok(ok),
+            ExporterRequestBuildResult::Err(err) => {
+                // Safety: not generally safe but this is for test code.
+                Err(unsafe { String::from_utf8_unchecked(err.into()) })
```
ivoanjo, line 186: Minor: I'm curious, why is this unsafe? Because the error may not be representable in utf8?

profiling-ffi/src/exporter.rs
```diff
@@ -169,6 +170,28 @@ unsafe fn into_vec_files<'a>(slice: Slice<'a, File>) -> Vec<exporter::File<'a>>
         .collect()
 }
 
+#[repr(C)]
+pub enum ExporterRequestBuildResult {
```
ivoanjo, line 174: When I did the FFI renames, I chose in some cases not to modify the Rust types, but instead to rename them using cbindgen configuration.

So I think we're missing one of the following:
1. Add `"ExporterRequestBuildResult" = "ddog_prof_Exporter_Request_BuildResult"` to `cbindgen.toml` section `[export.rename]`
2. Directly name this type `Exporter_Request_BuildResult`

profiling-ffi/src/profiles.rs
```diff
@@ -325,12 +325,36 @@ pub unsafe extern "C" fn ddog_prof_Profile_drop(
 ) {
 }
 
+#[repr(C)]
+pub enum ProfileAddResult {
```
ivoanjo, line 329: A similar note to `ExporterRequestBuildResult` above applies here, with the final name being `ddog_prof_Profile_AddResult`

profiling-ffi/src/profiles.rs
```diff
@@ -325,12 +325,36 @@ pub unsafe extern "C" fn ddog_prof_Profile_drop(
...
+            ProfileAddResult::Err(err) => {
+                // Safety: not generally safe but this is for test code.
+                Err(unsafe { String::from_utf8_unchecked(err.into()) })
+            }
+        }
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn ddog_prof_ProfileAddResult_drop(_result: ProfileAddResult) {}
```
ivoanjo, line 348: Naming: (Similar to other comments above)

```suggestion
pub extern "C" fn ddog_prof_Profile_AddResult_drop(_result: ProfileAddResult) {}
```

ivoanjo (approved):
:+1: 

Although I did not have the chance yet to try integrating this with Ruby, it LGTM!

profiling-ffi/src/exporter.rs
```diff
@@ -169,6 +170,28 @@ unsafe fn into_vec_files<'a>(slice: Slice<'a, File>) -> Vec<exporter::File<'a>>
...
+    Ok(Box<Request>),
```
ivoanjo, line 175: On a second pass through the PR, I noticed that this is exposing a `ddog_prof_Box_Request` to the FFI, instead of a `ddog_prof_Exporter_Request *`.

This will probably cause issues since the other APIs expect `ddog_prof_Exporter_Request *` as input.

ivoanjo (commented):
Sorry for approving and dismissing, but I noticed this extra note, and that the examples also need to be updated first :)

profiling-ffi/src/exporter.rs
morrisonlevi: Yes, and since it could come from C code, there's no guarantee.

ivoanjo: Aaah, I see. In our particular use, we know this is never going to happen (because we're not going to take in a `ExporterRequestBuildResult` and then look at the contents of its `Err`), but I guess the type system can't know that.

profiling-ffi/src/exporter.rs
```diff
@@ -169,6 +163,33 @@ unsafe fn into_vec_files<'a>(slice: Slice<'a, File>) -> Vec<exporter::File<'a>>
...
+
+/// Drops the result. Since `ddog_prof_Exporter_send` will take ownership of
+/// the Request object, do not call this method if you call
+/// `ddog_prof_Exporter_send` on the `ok` value!
+#[no_mangle]
+pub extern "C" fn ddog_prof_Exporter_Request_BuildResult_drop(_result: RequestBuildResult) {}
+
 /// Builds a Request object based on the profile data supplied.
 ///
 /// # Safety
```
ivoanjo, line 195: Minor: We were discussing it earlier over PM: I think it'd be cool to include here a note that you should either `send` or `drop` the result. I noticed both `drop` and `send` include similar notes, but I think since this is the supplier of the potentially-problematic object, it's cool to also drop a note here :)

profiling-ffi/src/exporter.rs
```diff
@@ -211,7 +232,7 @@ pub unsafe extern "C" fn ddog_prof_Exporter_Request_build(
 ///
 /// # Arguments
 /// * `exporter` - borrows the exporter for sending the request
-/// * `request` - takes ownership of the request
+/// * `request` - takes ownership of the request. Be careful not to drop it twice!
```
ivoanjo, line 235: Minor: I suggest being even more explicit here:

```suggestion
/// * `request` - takes ownership of the request. You do not need to (and must not) call `Request_drop` on a request after calling `send` with it.
```

examples/ffi/exporter.cpp
```diff
@@ -114,7 +114,7 @@ int main(int argc, char *argv[]) {
```
ivoanjo, line 114: :arrow_up: Would it perhaps make sense to update the `Profile_add` call above to check the result as well?

ivoanjo (approved):
:+1: LGTM! This is looking great! I think this is a really nice improvement, for two reasons a) Safer use of libdatadog and b) Libdatadog reporting back what's wrong, so that the libraries can log this info and it's available during investigations/support tickets.

examples/ffi/exporter.cpp
morrisonlevi: Yes, it does. The compiler didn't whine at me, so I added `#[must_use]`. Now it yells at me, and I fixed it in the C example too.

profiling-ffi/src/exporter.rs
morrisonlevi: The `Error` type I added would make this safer, at least if people follow the API and comments :)

PR #95, opened 2023-01-21
refactor(profiling)!: create FFI Error type and remove `*Result_drop` methods

morrisonlevi:
# What does this PR do?

This extracts an FFI `Error` type aka `ddog_Error` and it contains an FFI Vec internally.

```rust
/// Please treat this as opaque; do not reach into it, and especially don't
/// write into it!
#[derive(Debug)]
#[repr(C)]
pub struct Error {
    /// This is a String stuffed into the vec.
    message: Vec<u8>,
}
```

Then, the FFI error types use this in their `Err` field, instead of `ddcommon_ffi::Vec<u8>`:

```rust
#[repr(C)]
pub enum SendResult {
    HttpResponse(HttpStatus),
    Err(Error),
}
```

The `*Result_drop` methods have been removed:

- ddog_prof_Exporter_NewResult_drop
- ddog_prof_Exporter_Request_BuildResult_drop
- ddog_prof_Exporter_SendResult_drop
- ddog_prof_Profile_AddResult_drop
- ddog_prof_Profile_SerializeResult_drop
- ddog_Vec_Tag_PushResult_drop

And these were added instead:

- ddog_Error_drop (+ ddog_Error_message to get the message)
- ddog_prof_EncodedProfile_drop
- ddog_prof_Exporter_Request_drop

# Motivation

I thought it was a bit nicer this way to not have to reach into the guts, and instead have an API.

Additionally, we realized that with the `*Result_drop` APIs it was a bit easy to get into double-free scenarios, so we removed them.

# Additional Notes

I wish we could also extract a generic `Result` type, but the cbindgen output leaves a bad API. Maybe once some PRs land. The idea would be to hard-code the error type to always be the FFI Error:

```rust
#[repr(C)]
enum Result<T> {
    Ok(T),
    Err(Error), // the Error type being the one introduced here
};
```

# How to test the change?

Use the new `ddog_Error_message` and `ddog_Error_drop` APIs:

```c
/**
 * # Safety
 * Only pass null or a valid reference to an Error.
 */
void ddog_Error_drop(struct ddog_Error *error);

/**
 * Returns a CharSlice of the error's message that is valid until the error
 * is dropped.
 * # Safety
 * Only pass null or a valid reference to an Error.
 */
ddog_CharSlice ddog_Error_message(const struct ddog_Error *error);
```

As well as adjust your dropping as the `*Result_drop` APIs have been removed. You need to drop each case, if necessary (don't need to drop a uint64_t, obviously).

### DIFF
--- a/ddcommon-ffi/src/error.rs
+++ b/ddcommon-ffi/src/error.rs
@@ -62,7 +62,7 @@ impl From<Box<&dyn std::error::Error>> for Error {
 }
 
 /// # Safety
-/// Only pass null or a valid reference to an Error.
+/// Only pass null or a valid reference to a `ddog_Error`.
 #[no_mangle]
 pub unsafe extern "C" fn ddog_Error_drop(error: Option<&mut Error>) {
     if let Some(err) = error {
@@ -76,7 +76,7 @@ pub unsafe extern "C" fn ddog_Error_drop(error: Option<&mut Error>) {
 /// Returns a CharSlice of the error's message that is valid until the error
 /// is dropped.
 /// # Safety
-/// Only pass null or a valid reference to an Error.
+/// Only pass null or a valid reference to a `ddog_Error`.
 #[no_mangle]
 pub unsafe extern "C" fn ddog_Error_message(error: Option<&Error>) -> CharSlice {
     match error {
--- a/ddcommon-ffi/src/error.rs
+++ b/ddcommon-ffi/src/error.rs
@@ -1,0 +1,86 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+use crate::slice::CharSlice;
+use crate::vec::Vec;
+use std::fmt::{Display, Formatter};
+
+/// Please treat this as opaque; do not reach into it, and especially don't
+/// write into it!
+#[derive(Debug)]
+#[repr(C)]
+pub struct Error {
+    /// This is a String stuffed into the vec.
+    message: Vec<u8>,
+}
+
+impl AsRef<str> for Error {
+    fn as_ref(&self) -> &str {
+        // Safety: .message is a String (just FFI safe).
+        unsafe { std::str::from_utf8_unchecked(self.message.as_slice().as_slice()) }
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.as_ref())
+    }
+}
+
+impl std::error::Error for Error {}
+
+impl From<String> for Error {
+    fn from(value: String) -> Self {
+        let message = Vec::from(value.into_bytes());
+        Self { message }
+    }
+}
+
+impl From<Error> for String {
+    fn from(value: Error) -> String {
+        // Safety: .message is a String (just FFI safe).
+        unsafe { String::from_utf8_unchecked(value.message.into()) }
+    }
+}
+
+impl From<&str> for Error {
+    fn from(value: &str) -> Self {
+        Self::from(value.to_string())
+    }
+}
+
+impl From<anyhow::Error> for Error {
+    fn from(value: anyhow::Error) -> Self {
+        Self::from(value.to_string())
+    }
+}
+
+impl From<Box<&dyn std::error::Error>> for Error {
+    fn from(value: Box<&dyn std::error::Error>) -> Self {
+        Self::from(value.to_string())
+    }
+}
+
+/// # Safety
+/// Only pass null or a valid reference to an Error.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_Error_drop(error: Option<&mut Error>) {
+    if let Some(err) = error {
+        // Leave an empty String in place to help with double-free issues.
+        let mut tmp = Error::from(String::new());
+        std::mem::swap(&mut tmp, err);
+        drop(tmp)
+    }
+}
+
+/// Returns a CharSlice of the error's message that is valid until the error
+/// is dropped.
+/// # Safety
+/// Only pass null or a valid reference to an Error.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_Error_message(error: Option<&Error>) -> CharSlice {
+    match error {
+        None => CharSlice::default(),
+        Some(err) => CharSlice::from(err.as_ref()),
+    }
+}
--- a/examples/ffi/exporter.cpp
+++ b/examples/ffi/exporter.cpp
@@ -128,6 +128,7 @@ int main(int argc, char *argv[]) {
     nullptr,
     30000
   );
+  ddog_prof_EncodedProfile_drop(encoded_profile);
 
   if (build_result.tag == DDOG_PROF_EXPORTER_REQUEST_BUILD_RESULT_ERR) {
     print_error("Failed to build request: ", build_result.err);
--- a/examples/ffi/exporter.cpp
+++ b/examples/ffi/exporter.cpp
@@ -169,6 +169,8 @@ int main(int argc, char *argv[]) {
     printf("Response code: %d\n", send_result.http_response.code);
   }
 
+  // no ddog_prof_Exporter_Request_drop(request) since it was sent to Exporter_send
+
   ddog_prof_Exporter_drop(exporter);
   ddog_CancellationToken_drop(cancel);
   return exit_code;
--- a/examples/ffi/exporter.cpp
+++ b/examples/ffi/exporter.cpp
@@ -15,8 +15,9 @@ struct Deleter {
   void operator()(ddog_prof_Profile *object) { ddog_prof_Profile_drop(object); }
 };
 
-template <typename T> void print_error(const char *s, const T &err) {
-  printf("%s (%.*s)\n", s, static_cast<int>(err.len), err.ptr);
+void print_error(const char *s, const ddog_Error &err) {
+  auto charslice = ddog_Error_message(&err);
+  printf("%s (%.*s)\n", s, static_cast<int>(charslice.len), charslice.ptr);
 }
 
 int main(int argc, char *argv[]) {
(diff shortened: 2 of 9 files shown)


ivoanjo:
So... I'm a bit unsure about this one.

Things I like: 

* The idea of having a single "object" for containing the error, because that allows us to easily feed more information there (maybe some day the libdatadog stack trace? more details?) without having to change the API again.

* The FFI code seems cleaner and easier to follow, since it can focus on "this is an error", and the string handling stuff is only in a single place.

Things I'm not quite sure:
* This doesn't make it easier for the client to print these strings. E.g. consider that I turn every error into a Ruby string:

    ```c
    VALUE err_details = ruby_string_from_prof_vec_u8(exporter_result.err);
    ```

    so I'm imagining I'd update it to:

    ```c
    VALUE err_details = ruby_string_from_error(exporter_result.err);
    ```
    
    (where `ruby_string_from_error` calls `ddog_Error_message`.)

    So this doesn't particularly help out this part...

* One apparent benefit of this change would be to have a generic drop for every error. Which is true but... I don't think it alleviates the need for any of the other `_drop` methods, since there are still going to be cases where the clients need to drop the non-faulty result.

    Thus the API contract goes from "you need to either consume or drop most objects" to "you need to either consume or drop most objects, but if it's an error you can drop the error message instead", e.g. there's now two ways of doing the same thing.

    This is not necessarily a blocker; I'm thinking we can still keep the error, and not add the error-specific drop, which would prevent this issue.
    
* (Minor: The change to prefix arguments with `_`, while it makes sense from the Rust side, it's propagated to the FFI headers and its meaning there is kinda confusing. I realize that we were using a mix of `_` and no `_`, but I actually think the right thing in the FFI is to never use `_`)

morrisonlevi:
> Thus the API contract goes from "you need to either consume or drop most objects" to "you need to either consume or drop most objects, but if it's an error you can drop the error message instead", e.g. there's now two ways of doing the same thing.

What got me started on this path was actually a case where dropping the result was likely to result in double-free. So for that type, I specifically would _not_ create a drop for the result type, and force callers to drop the error case. I'm unsure if that ought to generalize or not. Maybe we should do a case-by-case review?

profiling-ffi/src/profiles.rs
```diff
@@ -444,12 +445,6 @@ impl From<datadog_profiling::profile::EncodedProfile> for EncodedProfile {
...
 }
 
-#[repr(C)]
-pub enum SerializeResult {
-    Ok(EncodedProfile),
-    Err(ddcommon_ffi::Vec<u8>),
-}
-
 /// Serialize the aggregated profile. Don't forget to clean up the result by
 /// calling ddog_prof_Profile_SerializeResult_drop.
```
ivoanjo, line 449: :arrow_up: This needs to be updated :)

ddcommon-ffi/src/error.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+/// # Safety
+/// Only pass null or a valid reference to a `ddog_Error`.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_Error_drop(error: Option<&mut Error>) {
+    if let Some(err) = error {
+        // Leave an empty String in place to help with double-free issues.
+        let mut tmp = Error::from(String::new());
```
ivoanjo, line 70: Is an empty string a `ddog_Vec_U8` with `ptr` set to `NULL`?

examples/ffi/exporter.cpp
```diff
@@ -15,8 +15,9 @@ struct Deleter {
   void operator()(ddog_prof_Profile *object) { ddog_prof_Profile_drop(object); }
 };
 
-template <typename T> void print_error(const char *s, const T &err) {
-  printf("%s (%.*s)\n", s, static_cast<int>(err.len), err.ptr);
+void print_error(const char *s, const ddog_Error &err) {
+  auto charslice = ddog_Error_message(&err);
+  printf("%s (%.*s)\n", s, static_cast<int>(charslice.len), charslice.ptr);
 }
```
ivoanjo, line 21: Minor: Since the error string is always going to be printed and accessed from the ffi client side, I wonder if it'd be nice to just make it a `char *` to save the effort of every API client having to turn it into a `char *` one way or another :)

ivoanjo (approved):
If I understood correctly, the guiding principle of the latest version  is: there is never any `_drop` for a `FooResult`, there's only `_drop` (if needed) for _the contents_ of `FooResult` -- there's one for the error, and another (if needed) for the non-error return.

I think I like it. There's only one correct way of doing things, and it de-emphasizes the `FooResult`s from being a thing you should care about vs just being effectively a shortcut for returning a (success flag, something) from our functions that would not exist if C allowed multiple returns natively.

One thing that occurred to me is -- should we add a note somewhere that when adding functions that return results via the FFI, they should always be marked as `#[must_use]`? That way clients get a compiler complaint if they just use a function that returns a result in a fire-and-forget-way (e.g. they're probably missing a drop).

I checked and every function _right now_ is correctly tagged, but it may be useful to document this somewhere so we don't forget this key fact :)

examples/ffi/exporter.cpp
morrisonlevi: It's a `ddog_CharSlice`, which does use `char*` for it's pointer. We can append a null byte for convenience if you want, but that doesn't guarantee that there aren't any interior null bytes. Meaning, you need the length anyway.

ddcommon-ffi/src/error.rs
ivoanjo: I never read this, so nothing to fix! But I was curious what ended up there :)

examples/ffi/exporter.cpp
ivoanjo: Right, what I was thinking is -- for error messages _specifically_ libdatadog could ensure that it output a real bona-fide `char *` with only a null byte at the end so that clients could print it directly.

morrisonlevi:
> I checked and every function _right now_ is correctly tagged, but it may be useful to document this somewhere so we don't forget this key fact :)

We can document it, yes, but will the programmer see/remember the documentation? :)

I added a README.md file to `profiling-ffi`.

morrisonlevi:
@ivoanjo, can you double-check before I commit? I did a leak check and it turns out on `main` we are leaking in the exporter.cpp example (I think it's a missing `ddog_prof_Profile_SerializeResult_drop` on the happy path). This branch doesn't leak anything, though.

profiling-ffi/src/exporter.rs
```diff
@@ -163,40 +160,23 @@ unsafe fn into_vec_files<'a>(slice: Slice<'a, File>) -> Vec<exporter::File<'a>>
...
 /// If unsuccessful, it returns an error message.
 ///
 /// The main use of the `ddog_prof_Exporter_Request` object is to be used in
 /// `ddog_prof_Exporter_send`, which will take ownership of the object. Do not pass it to
-/// `ddog_prof_Exporter_Request_drop` in such cases, nor call
-/// `ddog_prof_Exporter_Request_BuildResult_drop` on the result!
+/// `ddog_prof_Exporter_Request_drop` after that!
 ///
 /// # Safety
 /// The `exporter` and the files inside of the `files` slice need to have been
```
ivoanjo, line 182: Minor: Just noticed that this is actually incorrect -- the files can come from elsewhere e.g. for `metrics.json` or `code-provenance.json` :)

profiling-ffi/src/exporter.rs
```diff
@@ -163,40 +160,23 @@ unsafe fn into_vec_files<'a>(slice: Slice<'a, File>) -> Vec<exporter::File<'a>>
...
+            RequestBuildResult::Err(err) => Err(String::from(err)),
         }
     }
 }
 
-/// Drops the result. Since `ddog_prof_Exporter_send` will take ownership of
-/// the Request object, do not call this method if you call
-/// `ddog_prof_Exporter_send` on the `ok` value!
-#[no_mangle]
-pub extern "C" fn ddog_prof_Exporter_Request_BuildResult_drop(_result: RequestBuildResult) {}
```
ivoanjo, line 191: Just noticed that this one was correctly removed BUT we're missing `ddog_prof_Exporter_Request_drop`. We even reference it in some comments, but it's nowhere to be found :)

I actually use it in the Ruby profiler mostly because the files in the request are pointing to memory that is managed by Ruby so I do this dance of:

* Grab VM lock
* Build request (Ruby is not mutating the files because I'm holding the lock)
* Release lock
* Call send

Between "Build request" and "Call send" Ruby may want to abort (e.g. let's imagine the VM really wants to shut down) and in that situation it's when I used this API:

```c
  if (pending_exception) {
    // If we got here send did not run, so we need to explicitly dispose of the request
    ddog_prof_Exporter_Request_drop(request);

    // Let Ruby propagate the exception. This will not return.
    rb_jump_tag(pending_exception);
  }
```

profiling-ffi/src/exporter.rs
```diff
@@ -238,9 +218,8 @@ pub unsafe extern "C" fn ddog_prof_Exporter_Request_build(
 ///
 /// # Arguments
 /// * `exporter` - Borrows the exporter for sending the request.
-/// * `request` - Takes ownership of the request. Do not call `ddog_prof_Request_drop` nor
-///               `ddog_prof_Exporter_SendResult_drop` on `request` after calling
-///               `ddog_prof_Exporter_send` on it.
+/// * `request` - Takes ownership of the request. Do not call `ddog_prof_Request_drop` on
```
ivoanjo, line 221: Minor: Fixing namespace :)

```suggestion
/// * `request` - Takes ownership of the request. Do not call `ddog_prof_Exporter_Request_drop` on
```

ivoanjo (approved):
Last check with some last-minute suggested fixes!

Otherwise I've looked at the generated C headers and they also look good.

One last thing that may be interesting to mention in the PR description is that the following were dropped (pun not intended):

* `ddog_prof_Exporter_NewResult_drop`
* `ddog_prof_Exporter_Request_BuildResult_drop`
* `ddog_prof_Exporter_SendResult_drop`
* `ddog_prof_Profile_AddResult_drop`
* `ddog_prof_Profile_SerializeResult_drop`
* `ddog_Vec_Tag_PushResult_drop`

And these were added instead:

* `ddog_Error_drop` (+ `ddog_Error_message` to get the message)
* `ddog_prof_EncodedProfile_drop`
* `ddog_prof_Exporter_Request_drop`

PR #100, opened 2023-01-26
Create a span normalizer skeleton, fully implement span name normalization

thedavl:
# What does this PR do?

This PR adds a span normalizer skeleton (riddled with TODOs), but includes fully implemented span name normalization.

Future PRs (addressing TODOs) will complete:
- span tag normalization + unit tests
- span service normalization + unit tests.
- normalizer-wide unit tests
- trace normalization + unit tests

The span normalizer (as well as future functions completing TODOs) is modeled very closely after the existing agent trace normalizer, which can be found [here](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/normalizer.go). 

**Note: the unit tests are translated directly from the agent normalizer.** 

# Motivation

The trace normalizer will be used in our serverless agentless POC, which will use it to normalize traces before being sent directly to the datadog trace intake.

See: [the serverless branch](https://github.com/DataDog/libdatadog/tree/serverless)

# Additional Notes

- `agent/tracer/span` proto files were copied from https://github.com/DataDog/datadog-agent/tree/main/pkg/trace/pb

# How to test the change?

The changes have been manually tested using the agentless POC, normalizing span names and verifying the normalization works as expected. Unit tests have been copied/translated from the agent normalizer unit tests written in go.


trace/src/normalization/normalizer.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::pb;
+
+const MAX_TYPE_LEN: i64 = 100;
+
+// an arbitrary cutoff to spot weird-looking values
+// nanoseconds since epoch on Jan 1, 2000
+const YEAR_2000_NANOSEC_TS: i64 = 946684800000000000;
+
+#[allow(dead_code)]
+fn normalize(s: &mut pb::Span) -> Result<(), errors::NormalizerError> {
```
thedavl, line 18: A span normalizer which will be used by future normalize_trace function.

trace/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "trace"
```
morrisonlevi, line 2: I suspect you want a name like `datadog-trace-normalization` and to name the folder `trace-normalization`. At least with current conventions, which we can definitely change as we grow if we need to.

trace/src/normalization/normalize_utils.rs
```diff
@@ -0,0 +1,192 @@
...
+pub const DEFAULT_SPAN_NAME: &str = "unnamed_operation";
+// DEFAULT_SERVICE_NAME is the default name we assign a service if it's missing and we have no reasonable fallback
+pub const DEFAULT_SERVICE_NAME: &str = "unnamed-service";
+
+// MAX_NAME_LEN the maximum length a name can have
+pub const MAX_NAME_LEN: i64 = 100;
+// MAX_SERVICE_LEN the maximum length a service can have
+pub const MAX_SERVICE_LEN: i64 = 100;
+
+pub const MAX_TAG_LEN: i64 = 200;
```
morrisonlevi, line 18: A bit unusual to deal with `i64` for what appears to be a length, which is usually unsigned. Is there some particular aspect of design this stems from?

trace/src/normalization/normalizer.rs
```diff
@@ -0,0 +1,114 @@
...
+    match err {
+        Some(errors::NormalizeErrors::ErrorEmpty) => println!("Fixing malformed trace. Name is empty (reason:span_name_empty)"),
+        Some(errors::NormalizeErrors::ErrorTooLong) => println!("Fixing malformed trace. Name is too long (reason:span_name_truncate)"),
+        Some(errors::NormalizeErrors::ErrorInvalid) => println!("Fixing malformed trace. Name is invalid (reason:span_name_invalid)"),
+        None => ()
+    }
+    s.name = normalized_name;
+
+    if s.resource.is_empty() {
+        println!("Fixing malformed trace. Resource is empty (reason:resource_empty)");
```
morrisonlevi, line 49: I'm guessing this should be a log message instead, but we don't have a logging story in libdatadog yet. At least, this shouldn't print to stdout. The same goes for the other prints.

morrisonlevi (commented):
This wasn't a thorough review, just a drive-by to get started. Will check more later, probably tomorrow!

trace/Cargo.toml
thedavl: makes sense, I'll change that!

trace/src/normalization/normalize_utils.rs
thedavl: no good reason, I'll definitely change it over

trace/src/normalization/normalizer.rs
thedavl: Right, it should be. I'll remove all the print statements

trace-normalization/src/normalization/normalize_utils.rs
```diff
@@ -0,0 +1,191 @@
...
+    // if s == "" {
+    //     return (fallbackService(lang), errors::NormalizeErrors::ErrorInvalid)
+    // }
+    // return (s, err)
+    // (svc, None)
+// }
+
+// normalize_name normalizes a span name and returns an error describing the reason
+// (if any) why the name was modified.
+pub fn normalize_name(name: String) -> (String, Option<errors::NormalizeErrors>) {
```
algo-facile, line 58: I think you can use the built-in feature Result of rust instead of returning a tuple?

maxday, line 58: I think you can use the built-in feature Result of rust instead of returning a tuple?

trace-normalization/src/normalization/normalizer.rs
```diff
@@ -0,0 +1,96 @@
...
+        let status_code: String = s.meta.get("http.status_code").unwrap().to_string();
+        if !is_valid_status_code(status_code.clone()) {
+            s.meta.remove("http.status_code");
+        }
+    }
+
+    Ok(())
+}
+
+pub fn is_valid_status_code(sc: String) -> bool {
```
maxday, line 87: you could simplify this a bit with :
```
pub fn is_valid_status_code(sc: String) -> bool {
    if let Ok(code) = sc.parse::<i64>() {
        return (100..600).contains(&code);
    }
    false
}
```

trace-normalization/src/normalization/normalize_utils.rs
thedavl: I think the problem is this function has a case where the given name is over the max length, and this function needs to return the truncated name and an error indicating that the given name was too long

trace-normalization/src/normalization/normalizer.rs
thedavl: good idea, added!

trace-normalization/src/normalization/normalize_utils.rs
thedavl: I ended up refactoring these two functions to use Result<>

trace-normalization/src/normalization/normalize_utils.rs
```diff
@@ -0,0 +1,191 @@
...
+
+// pub fn is_valid_ascii_start_char(c: char) -> bool {
+//     ('a'..='z').contains(&c) || c == ':'
+// }
+
+// pub fn is_valid_ascii_tag_char(c: char) -> bool {
+//     is_valid_ascii_start_char(c) || ('0'..='9').contains(&c) || c == '.' || c == '/' || c == '-'
+// }
+
+pub fn normalize_metric_names(name: String) -> (String, bool) {
```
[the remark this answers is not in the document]
thedavl: refactored to return Result<>

trace-normalization/src/normalization/normalizer.rs
```diff
@@ -0,0 +1,100 @@
...
+pub const DEFAULT_SPAN_NAME: &str = "unnamed_operation";
+
+#[allow(dead_code)]
+pub fn normalize(s: &mut pb::Span) -> Result<(), errors::NormalizerError> {
+    if s.trace_id == 0 {
+        return Err(errors::NormalizerError::new("TraceID is zero (reason:trace_id_zero)"));
+    }
+    if s.span_id == 0 {
+        return Err(errors::NormalizerError::new("SpanID is zero (reason:span_id_zero)"));
+    }
```
morrisonlevi, line 27: I'm not sure if anyone on your team has opinions, but I would recommend pulling in the `anyhow` crate instead of rolling your own errors. It has some nice macros to simplify code around error paths. For instance, the `anyhow::ensure!` macro would reduce the code to something like this:

```rust
pub fn normalize(s: &mut pb::Span) -> anyhow::Result<()> {
    anyhow::ensure!(s.trace_id != 0, "TraceID is zero (reason:trace_id_zero)");
    anyhow::ensure!(s.span_id != 0, "SpanID is zero (reason:span_id_zero)");
```

trace-normalization/src/normalization/normalizer.rs
```diff
@@ -0,0 +1,96 @@
...
+	// with the Zipkin implementation. Furthermore, as described in the PR
+	// https://github.com/openzipkin/zipkin/pull/851 the constraint that the
+	// root span's ``trace id = span id`` has been removed
+	if s.parent_id == s.trace_id && s.parent_id == s.span_id {
+		s.parent_id = 0;
+	}
+
+    // Start & Duration as nanoseconds timestamps
+	// if s.Start is very little, less than year 2000 probably a unit issue so discard
+	// (or it is "le bug de l'an 2000")
```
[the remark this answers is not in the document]
thedavl: huh that's weird, there was a visual bug in my vscode that showed that code block with just one indent. Fixed it.

trace-normalization/src/normalization/normalize_utils.rs
```diff
@@ -0,0 +1,182 @@
...
+    //     return (fallbackService(lang), errors::NormalizeErrors::ErrorInvalid)
+    // }
+    // return (s, err)
+    // (svc, None)
+// }
+
+// normalize_name normalizes a span name and returns an error describing the reason
+// (if any) why the name was modified.
+// pub fn normalize_name(name: String) -> (String, Option<errors::NormalizeErrors>) {
+pub fn normalize_name(name: String) -> Result<String, errors::NormalizeErrors> {
```
morrisonlevi, line 54: This API means that you will pass a String, and `normalize_name` will take ownership of it, and on error, it will drop it. The name won't exist anymore. This is maybe fine, depends a bit on how you expect it to be used.

There are some other patterns out there, for instance `Cow<str>` might be nice here:

```
pub fn normalize_name(name: &str) -> Result<Cow<str>, errors::NormalizeErrors> {
```

If everything is fine, the string is simply borrowed, never copied. If a change needs to be made, then a new string is created. On an error, the caller still has the original `&str` that was being borrowed.

trace-normalization/src/pb.rs
```diff
@@ -0,0 +1,125 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present
+// Datadog, Inc.
+
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct Span {
```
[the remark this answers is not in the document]
thedavl: correct, this is autogenerated from the agent/tracer/span proto files.

trace-normalization/src/normalization/normalizer.rs
thedavl: this makes sense, I'll make these changes

trace-normalization/src/normalization/normalize_utils.rs
thedavl: I think` Cow<str>` definitely makes sense here, I'll make this change

trace-normalization/src/normalization/tests/normalize_utils_tests.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[cfg(test)]
+mod normalize_tests {
+
+    use crate::errors;
+    use crate::normalize_utils;
+
+    #[test]
+    fn test_normalize_name() {
+        let test_tuples: [(&str, &str, Option<errors::NormalizeErrors>); 5] = [
```
DarcyRaynerDD: Nit: One advanced way to write table tests like this is to use the duplicate crate
https://crates.io/crates/duplicate
Eg
```
#[duplicate_item(
  test_name input  expected expected_err;
  ["test_normalize_empty_string"] [ "" ]  [ "" ] [None];
)]
#[test]
fn test_name() {
          match normalize_utils::normalize_name(input.to_string()) {
              Ok(val) => {
                  assert_eq!(expected_err, None);
                  assert_eq!(val, expected);
              },
              Err(err) => {
                  assert_eq!(err, expected_err.unwrap());
              }
          }
}
```
It's good because it saves a lot of typing, and you will be individual unit test results.

trace-normalization/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "datadog-trace-normalization"
+version = "0.0.0"
+authors = ["David Lee <david.lee@datadoghq.com>"]
+edition = "2021"
+
+[workspace]
```
DarcyRaynerDD, line 7: The workspace field shouldn't be necessary, the module if already inside a workspace, (defined in the cargo.toml as the root of the repo).

trace-normalization/src/normalization/normalizer.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    // TODO: Implement tag normalization in future PR
+    // if s.meta.contains_key("env") {
+    //     let env_tag: String = s.meta.get("env").unwrap().to_string();
+    //     s.meta.insert("env".to_string(), normalize_utils::normalize_tag(env_tag));
+    // }
+
+    if s.meta.contains_key("http.status_code") {
+        let status_code: String = s.meta.get("http.status_code").unwrap().to_string();
```
DarcyRaynerDD, line 80: You can simplify this to remove the unwrap
```suggestion
    if let Some(status_code) = s.meta.get("http.status_code") {
```

DarcyRaynerDD (commented):
Left a few nits. Overall excellent work.

trace-normalization/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present
+// Datadog, Inc.
+
+#![deny(clippy::all)]
+
+pub mod pb {
+    include!("./pb.rs");
```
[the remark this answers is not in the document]
thedavl: I'll change the structure to use the standard approach

trace-normalization/src/normalization/tests/normalize_utils_tests.rs
thedavl: great idea, implemented

trace-normalization/src/normalization/normalizer.rs
morrisonlevi: @DarcyRaynerDD I hadn't heard of Snafu until you mention it, so that pretty much sums up why it wasn't considered :)

trace-normalization/src/normalization/normalize_utils.rs
```diff
@@ -0,0 +1,182 @@
...
+
+// TruncateUTF8 truncates the given string to make sure it uses less than limit bytes.
+// If the last character is an utf8 character that would be splitten, it removes it
+// entirely to make sure the resulting string is not broken.
+pub fn truncate_utf8(s: String, limit: usize) -> String {
+    if s.len() <= limit {
+        return s
+    }
+    let mut prev_index = 0;
+    for i in 0..s.len() {
```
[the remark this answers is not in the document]
thedavl: ah right, added.

trace-normalization/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "datadog-trace-normalization"
+version = "0.0.0"
```
morrisonlevi, line 3:
```suggestion
version = "2.0.0"
```

We are preparing to release libdatadog v2.0.0. Sound good to you?

trace-normalization/src/normalize_utils.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    // given a dummy value
+    let mut last_char: char = 'a';
+
+    let char_vec: Vec<char> = name.chars().collect();
+
+    let mut i = 0;
+
+    // skip non-alphabetic characters
+    while i < name.len() && !is_alpha(char_vec[0]) {
```
morrisonlevi, line 134: The `0` here seems suspect; is it supposed to be `i` in some form? There's probably a more Rust idiomatic way to express what you are doing, if so.

thedavl: oops, fixed.

morrisonlevi (approved):

PR #99, opened 2023-01-26
fix: clippy lints from Rust v1.67.0 release

morrisonlevi:
# What does this PR do?

Fixes some `clippy::uninlined_format_args` lints and also a `clippy::seek_to_start_instead_of_rewind` lint and that appeared on `stable` aka 1.67.0.

# Motivation

An unrelated PR started failing :)

# Additional Notes

Nope.

# How to test the change?

No change, regular tests apply.

PR #101, opened 2023-01-26
feat(build-profiling-ffi.sh): support CARGO_TARGET_DIR

morrisonlevi:
# What does this PR do?

Adds support for the environment variable `CARGO_TARGET_DIR`, which controls where cargo places its target directory. The script currently assumes `CARGO_TARGET_DIR` isn't set, and fails when trying to copy things from `./target`.

# Motivation

I was building in a docker container with a mounted volume for libdatadog source. To speed up the build, I set `CARGO_TARGET_DIR` to `/tmp/cargo`, but noticed this script failed.

# Additional Notes

I wish Mac ASAN supported leak detection already, then I wouldn't have to build in a container 🙃 .

Also, the clippy lint failures are already PR'd in #99.

# How to test the change?

`export CARGO_TARGET_DIR=/some/path` before running `build-profiling-ffi.sh`. Or, don't and note it should work the same as it does today.

### DIFF
--- a/build-profiling-ffi.sh
+++ b/build-profiling-ffi.sh
@@ -4,6 +4,11 @@
 # under the Apache License Version 2.0. This product includes software developed
 # at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
 
+# Location to place all artifacts
+if [ -z $CARGO_TARGET_DIR ] ; then
+    export CARGO_TARGET_DIR=$PWD/target
+fi
+
 set -eu
 
 destdir="$1"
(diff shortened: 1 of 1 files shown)


build-profiling-ffi.sh
```diff
@@ -4,6 +4,11 @@
 # under the Apache License Version 2.0. This product includes software developed
 # at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
 
+# Location to place all artifacts
+if [ -z $CARGO_TARGET_DIR ] ; then
+    export CARGO_TARGET_DIR=$PWD/target
```
realFlowControl, line 9: This changes where the `target` directory is, as you can run `cargo build` from a subdirectory and it would create the `target` dir from now on in this subdirectory. Before it would have been created relative to the `Cargo.toml`, now it would be relative to the `$PWD`. 
OTOH: The `build-profiling-ffi.sh` is only working when called from same directory where it is. If you go to the `profiling-ffi` directory and call it via `../build-profiling-ffi.sh` it fails anyway, so this might not be a problem after all.

realFlowControl (approved):
LGTM! 
Just a small thing I found, which might not be a problem after all.

build-profiling-ffi.sh
morrisonlevi: It's a good find, we should support it at some point. Since it doesn't work today and this doesn't change it, it doesn't block this PR.

PR #103, opened 2023-01-27
refactor(profiling)!: less chance of request double free

morrisonlevi:
# What does this PR do?

This refactors `ddog_prof_Exporter_send` to use a double-pointer for request `ddog_prof_Exporter_Request **request` instead of a single one. This allows it to replace the `*request` of the caller with a null-pointer, reducing the opportunity to double-free it by mistake.

Also does some lifetime refactoring around cancellation tokens. There weren't bugs before as far as I can tell, but now the lifetimes are more obviously bounded.

# Motivation

It bothered me that we kept calling out "don't double-free the request object." We should just make a better API, right?

# Additional Notes

Nope.

# How to test the change?

Handling of the request object is a bit different, it should now look a bit more like this C++ pseudo-code:

```cpp
auto build_result = ddog_prof_Exporter_Request_build(build, ...);
// handle err

auto &request = build_result.ok;
auto send_result = ddog_prof_Exporter_send(exporter, &request, nullptr);

// The build_result.ok now has null, so you can do this drop.
// Or, don't. Whatever you think is cleaner with your specific code.
ddog_prof_Exporter_Request_drop(request);

// handle send_result
```

### DIFF
--- a/profiling-ffi/src/exporter.rs
+++ b/profiling-ffi/src/exporter.rs
@@ -174,10 +174,6 @@ impl From<RequestBuildResult> for Result<Box<Request>, String> {
 /// If successful, builds a `ddog_prof_Exporter_Request` object based on the
 /// profile data supplied. If unsuccessful, it returns an error message.
 ///
-/// The main use of the `ddog_prof_Exporter_Request` object is to be used in
-/// `ddog_prof_Exporter_send`, which will take ownership of the object. Do not
-/// pass it to `ddog_prof_Exporter_Request_drop` after that!
-///
 /// # Safety
 /// The `exporter`, `additional_stats`, and `endpoint_stats` args should be
 /// valid objects created by this module, except NULL is allowed.
@@ -215,9 +211,8 @@ pub unsafe extern "C" fn ddog_prof_Exporter_Request_build(
 }
 
 /// # Safety
-/// The `request` must be valid, or null. Notably, it is invalid if it's been
-/// sent to `ddog_prof_Exporter_send` or `ddog_prof_Exporter_Request_drop`
-/// already.
+/// The `request` must be valid, or null. After this call, the `request` will
+/// be invalid.
 #[no_mangle]
 pub unsafe extern "C" fn ddog_prof_Exporter_Request_drop(request: *mut Request) {
     if !request.is_null() {
@@ -229,9 +224,9 @@ pub unsafe extern "C" fn ddog_prof_Exporter_Request_drop(request: *mut Request) 
 ///
 /// # Arguments
 /// * `exporter` - Borrows the exporter for sending the request.
-/// * `request` - Takes ownership of the request. Do not call
-///               `ddog_prof_Exporter_Request_drop` on `request` after calling
-///               `ddog_prof_Exporter_send` on it.
+/// * `request` - Takes ownership of the request, replacing it with a null
+///               pointer. This is why it takes a double-pointer, rather than
+///               a single one.
 /// * `cancel` - Borrows the cancel, if any.
 ///
 /// # Safety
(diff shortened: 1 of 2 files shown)


profiling-ffi/src/exporter.rs
```diff
@@ -12,19 +12,18 @@ use ddcommon::tag::Tag;
...
-use std::ptr::NonNull;
+use std::ptr::{drop_in_place, NonNull};
 use std::str::FromStr;
 
 #[repr(C)]
 pub enum ExporterNewResult {
-    Ok(*mut ProfileExporter),
+    Ok(NonNull<ProfileExporter>),
     Err(Error),
 }
```
ivoanjo, line 22: This changes the type exported in the FFI from `struct ddog_prof_Exporter_Request *ok` to `struct ddog_prof_NonNull_Request ok`.

I'm not sure I quite follow this change -- doesn't this `NonNull_Request` become null after being used/consumed?

(Also we probably need to update `cbindgen.toml` to call this... something. Perhaps this should still be named a `ddog_prof_Exporter_Request` as far as C clients care?)

profiling-ffi/src/exporter.rs
```diff
@@ -12,19 +12,18 @@ use ddcommon::tag::Tag;
...
 
 #[repr(C)]
 pub enum RequestBuildResult {
-    // This is a Box::into_raw pointer.
-    Ok(*mut Request),
+    Ok(NonNull<Request>),
```
ivoanjo, line 26: (Same note as for `ExporterNewResult`?)

profiling-ffi/src/exporter.rs
```diff
@@ -121,15 +121,15 @@ pub extern "C" fn ddog_prof_Exporter_new(
...
+    tags: *const ddcommon_ffi::Vec<Tag>,
     endpoint: Endpoint,
 ) -> ExporterNewResult {
     match || -> anyhow::Result<ProfileExporter> {
         let library_name = unsafe { profiling_library_name.to_utf8_lossy() }.into_owned();
         let library_version = unsafe { profiling_library_version.to_utf8_lossy() }.into_owned();
         let family = unsafe { family.to_utf8_lossy() }.into_owned();
         let converted_endpoint = unsafe { try_to_endpoint(endpoint)? };
-        let tags = tags.map(|tags| tags.iter().map(Tag::clone).collect());
+        let tags = unsafe { tags.as_ref() }.map(|tags| tags.iter().map(Tag::clone).collect());
```
ivoanjo, line 132: Just to doublecheck -- is it still ok to pass in a `NULL` for the tags?

profiling-ffi/src/exporter.rs
```diff
@@ -165,128 +169,125 @@ impl From<RequestBuildResult> for Result<Box<Request>, String> {
...
 pub unsafe extern "C" fn ddog_prof_Exporter_Request_build(
-    exporter: Option<NonNull<ProfileExporter>>,
+    exporter: *mut ProfileExporter,
     start: Timespec,
     end: Timespec,
     files: Slice<File>,
-    additional_tags: Option<&ddcommon_ffi::Vec<Tag>>,
-    endpoints_stats: Option<&profiled_endpoints::ProfiledEndpointsStats>,
+    additional_tags: *const ddcommon_ffi::Vec<Tag>,
+    endpoints_stats: *const profiled_endpoints::ProfiledEndpointsStats,
```
ivoanjo, line 192: Can these still be `NULL` now that they're not `Option`? I don't quite understand the trade-off in this change :sweat_smile:

profiling-ffi/src/exporter.rs
```diff
@@ -165,128 +169,125 @@ impl From<RequestBuildResult> for Result<Box<Request>, String> {
...
-/// already.
+/// The `request` must be valid, or null. After this call, the `request` will
+/// be invalid.
 #[no_mangle]
 pub unsafe extern "C" fn ddog_prof_Exporter_Request_drop(request: *mut Request) {
     if !request.is_null() {
-        drop(Box::from_raw(request));
+        drop_in_place(request)
     }
 }
```
ivoanjo, line 229: So one thing I don't understand in the change to `ddog_prof_Exporter_send` to receive a `request: *mut *mut Request` is -- why is `send` different than `drop`? E.g. why have `*mut *mut` in one but `*mut` in the other?

profiling-ffi/src/exporter.rs
```diff
@@ -165,128 +169,125 @@ impl From<RequestBuildResult> for Result<Box<Request>, String> {
...
-        None => {
-            return SendResult::Err(Error::from("failed to export: exporter was null"));
-        }
-        Some(e) => e,
+    let mut tmp = std::ptr::null_mut(); // leave a nullptr for the caller
+    std::mem::swap(&mut tmp, request_ptr);
+    let request = match tmp.as_mut() {
+        Some(r) => Box::from_raw(r as *mut _),
+        None => anyhow::bail!("request was null"),
     };
```
ivoanjo, line 270: This seems off -- do we really need to check if the request is null twice?

ivoanjo (commented):
Left a few comments! I feel that some of the subtleties of the changes are lost on a Rust n00b like me, so it's probably a good idea for someone else to give a pass :)

profiling-ffi/src/exporter.rs
morrisonlevi: Yes.

profiling-ffi/src/exporter.rs
morrisonlevi: Yes.

profiling-ffi/src/exporter.rs
morrisonlevi: I'll think more about this, thanks.

profiling-ffi/src/exporter.rs
morrisonlevi: It's `** Request`,  so there are two null checks, one for each pointer.

profiling-ffi/src/exporter.rs
morrisonlevi: > This changes the type exported in the FFI from struct ddog_prof_Exporter_Request *ok to struct ddog_prof_NonNull_Request ok.

This is not happening for me -- I double checked this. I shall triple-check!

morrisonlevi: As far as the `NonNull<T>` being accurate... yes, I think it is:

1. At the point these `*Result` types are created in Rust, they are not null. When using the Rust API to test these things, we get a nicer type to work with.
2. These `*Result` types are never _consumed_ by a Rust API except in tests, so there isn't an opportunity for `extern "C"` users to screw this up.

Does that sound right to you?

profiling-ffi/src/exporter.rs
morrisonlevi: Pushed this up.

profiling-ffi/src/exporter.rs
ivoanjo: I've just tried with the latest cbindgen (I was using the one from the ubuntu repos) and it looks good with that one :+1:

profiling-ffi/src/profiles.rs
```diff
@@ -434,10 +453,8 @@ pub struct EncodedProfile {
 /// call this twice on the same object);
 #[no_mangle]
 pub unsafe extern "C" fn ddog_prof_EncodedProfile_drop(profile: Option<&mut EncodedProfile>) {
```
ivoanjo, line 455: Unfortunately this does not seem to be correct. I was trying to run the Ruby profiler test suite and I get a crash whenever this function gets called. I was able to reproduce the same issue in the `exporter.cpp` example:

```
$ DD_API_KEY=abc ./exporter foo
free(): invalid pointer
Aborted
```

ivoanjo: (If I otherwise comment out the drop I can run the Ruby profiler test suite fine with this PR)

morrisonlevi: Yes, I messed up the `ddog_prof_EncodedProfile_drop` when refactoring. Leaving a comment to help me remember.

profiling-ffi/src/exporter.rs
```diff
@@ -165,128 +188,132 @@ impl From<RequestBuildResult> for Result<Box<Request>, String> {
...
-        None => RequestBuildResult::Err(anyhow::anyhow!("no exporter was provided").into()),
+    match exporter.as_ref() {
+        None => RequestBuildResult::Err(anyhow::anyhow!("exporter was null").into()),
         Some(exporter) => {
             let timeout = std::time::Duration::from_millis(timeout_ms);
             let converted_files = into_vec_files(files);
-            let tags = additional_tags.map(|tags| tags.iter().map(Tag::clone).collect());
+            let tags = additional_tags
+                .as_ref()
+                .map(|tags| tags.iter().map(Tag::clone).collect());
```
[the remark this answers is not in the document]
morrisonlevi: It works fine, but doesn't need `as_ref`, removing. 👍🏻

ivoanjo (approved):
:+1: LGTM

PR #106, opened 2023-02-02
Trace Normalizer: add service + env tag normalization

thedavl:
# What does this PR do?

Adds service tag and env tag normalization to the normalizer + associated unit tests.

This normalizer is modeled very closely after the existing agent trace normalizer, which can be found [here](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/normalizer.go).

The next PR will cover:
- `normalize_trace`
- `normalize_chunk`

These two last functions in the normalizer are much shorter than `normalize`

# Motivation

The trace normalizer will be used in our serverless agentless POC, which will use it to normalize traces before being sent directly to the datadog trace intake.

See: [the serverless branch](https://github.com/DataDog/libdatadog/tree/serverless)

# Additional Notes

- Tags allow non ascii alpha characters, while metrics names don't

# How to test the change?

The changes have been manually tested using the agentless POC, normalizing span names and verifying the normalization works as expected. Unit tests have been copied/translated from the agent normalizer unit tests written in go.

### DIFF
--- a/trace-normalization/src/normalize_utils.rs
+++ b/trace-normalization/src/normalize_utils.rs
@@ -127,8 +127,10 @@ pub(crate) fn is_normalized_ascii_tag(tag: &str) -> bool {
         None => return false,
     }
 
-    for mut i in 0..tag.len() {
-        let cur_char = match tag.chars().nth(i) {
+    let mut tag_iter = tag.chars().peekable();
+
+    while tag_iter.peek().is_some() {
+        let cur_char = match tag_iter.next() {
             Some(c) => c,
             None => return false,
         };
@@ -137,8 +139,7 @@ pub(crate) fn is_normalized_ascii_tag(tag: &str) -> bool {
         }
         if cur_char == '_' {
             // an underscore is only okay if followed by a valid non-underscore character
-            i += 1;
-            match tag.chars().nth(i) {
+            match tag_iter.next() {
                 Some(c) => {
                     if !is_valid_ascii_tag_char(c) {
                         return false;
--- a/trace-normalization/src/normalize_utils.rs
+++ b/trace-normalization/src/normalize_utils.rs
@@ -44,11 +44,7 @@ pub(crate) fn fallback_service() -> String {
 pub(crate) fn normalize_service(svc: &str) -> anyhow::Result<String> {
     anyhow::ensure!(!svc.is_empty(), "Normalizer Error: Empty service name.");
 
-    let truncated_service = if svc.len() > MAX_SERVICE_LEN {
-        truncate_utf8(svc, MAX_SERVICE_LEN)
-    } else {
-        svc
-    };
+    let truncated_service = truncate_utf8(svc, MAX_SERVICE_LEN);
 
     normalize_tag(truncated_service)
 }
--- a/trace-normalization/src/normalize_utils.rs
+++ b/trace-normalization/src/normalize_utils.rs
@@ -118,7 +118,10 @@ pub(crate) fn is_normalized_ascii_tag(tag: &str) -> bool {
     if tag.len() > MAX_TAG_LEN {
         return false;
     }
-    match tag.chars().next() {
+
+    let mut tag_iter = tag.chars();
+
+    match tag_iter.next() {
         Some(c) => {
             if !is_valid_ascii_start_char(c) {
                 return false;
@@ -126,8 +129,6 @@ pub(crate) fn is_normalized_ascii_tag(tag: &str) -> bool {
         }
         None => return false,
     }
-
-    let mut tag_iter = tag.chars();
 
     while let Some(cur_char) = tag_iter.next() {
         if is_valid_ascii_tag_char(cur_char) {
--- a/trace-normalization/src/normalize_utils.rs
+++ b/trace-normalization/src/normalize_utils.rs
@@ -34,17 +34,12 @@ pub(crate) fn truncate_utf8(s: &str, limit: usize) -> &str {
 
 // fallbackService returns the fallback service name for a service
 // belonging to language lang.
-pub(crate) fn fallback_service(lang: &str) -> String {
+pub(crate) fn fallback_service() -> String {
     if lang.is_empty() {
         return DEFAULT_SERVICE_NAME.to_string();
     }
-    let mut service_name = String::new();
-    service_name.push_str("unnamed-");
-    service_name.push_str(lang);
-    service_name.push_str("-service");
-    // TODO: the original golang implementation uses a map to cache previously created
-    // service names. Implement that here.
-    service_name
+    // In the go agent implementation, if a lang was specified in TagStats
+    // (extracted from the payload header) the fallback_service name would be "unnamed-{lang}-service". 
 }
 
 // NormalizeService normalizes a span service and returns an error describing the reason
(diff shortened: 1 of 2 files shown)


trace-normalization/src/normalize_utils.rs
```diff
@@ -25,23 +32,141 @@ pub(crate) fn truncate_utf8(s: &str, limit: usize) -> &str {
...
+    }
+
+    anyhow::ensure!(!tag.is_empty(), "Normalizer Error: Empty tag name.");
+
+    // given a dummy value
+    let mut last_char: char = 'a';
+
+    let mut result = String::with_capacity(tag.len());
+
+    let char_vec: Vec<char> = tag.chars().collect();
```
DarcyRaynerDD, line 72: I'm not sure you need to collect this into a vec, (which will allocate memory), since it's immutable you can just iterate over the original string:

```
for &char in tag.chars() {
```

trace-normalization/src/normalize_utils.rs
```diff
@@ -25,23 +32,141 @@ pub(crate) fn truncate_utf8(s: &str, limit: usize) -> &str {
...
+        Some(c) => {
+            if !is_valid_ascii_start_char(c) {
+                return false;
+            }
+        }
+        None => return false,
+    }
+
+    for mut i in 0..tag.len() {
+        let cur_char: char = match tag.chars().nth(i) {
```
DarcyRaynerDD, line 138: nit: Shortcut
```
let Some(cur_char) = match tag.chars().nth(i) else {return false; };
```

trace-normalization/src/normalize_utils.rs
```diff
@@ -190,4 +263,80 @@ mod tests {
...
+            Ok(val) => {
+                assert_eq!(expected_err, "");
+                assert_eq!(val, expected)
+            }
+            Err(err) => {
+                assert_eq!(format!("{err}"), expected_err);
+            }
+        }
+    }
+    #[duplicate_item(
```
DarcyRaynerDD, line 286: These are some nice thorough tests.

trace-normalization/src/normalizer.rs
```diff
@@ -198,4 +204,231 @@ mod tests {
         assert!(normalizer::normalize(&mut test_span).is_ok());
         assert_eq!(test_span.resource, test_span.name);
     }
+
+    #[test]
+    pub fn test_normalize_trace_id_passes() {
```
DarcyRaynerDD, line 209: Tests don't usually need to be pub

trace-normalization/src/normalize_utils.rs
```diff
@@ -25,23 +32,141 @@ pub(crate) fn truncate_utf8(s: &str, limit: usize) -> &str {
...
+        }
+        if cur_char.is_uppercase() {
+            let mut iter = cur_char.to_lowercase();
+            if let Some(c) = iter.next() {
+                result.push(c);
+                last_char = c;
+            }
+            continue;
+        }
+        if cur_char.is_alphabetic() {
```
DarcyRaynerDD, line 91: One thing to think about here, is that is_alphabetic uses the unicode definition of alphabetic characters, so includes characters from all alphabets, not just ascii. Whereas this helper function used elsewhere is only checking for ascii alpha characters.

https://github.com/DataDog/libdatadog/blob/main/trace-normalization/src/normalize_utils.rs#L162

Our documentation only mentions [alphanumerics](https://docs.datadoghq.com/getting_started/tagging/#define-tags):

What does the go trace agent do here? We should try to be consistent in which method we use

trace-normalization/src/normalize_utils.rs
```diff
@@ -25,23 +32,141 @@ pub(crate) fn truncate_utf8(s: &str, limit: usize) -> &str {
...
+            };
+        } else {
+            return false;
+        }
+    }
+    true
+}
+
+pub(crate) fn is_valid_ascii_start_char(c: char) -> bool {
+    ('a'..='z').contains(&c) || c == ':'
```
DarcyRaynerDD, line 164: Same deal here, also, are only lower case characters valid as a start character?

DarcyRaynerDD (commented):
Nice work, left some questions and nits

trace-normalization/src/normalize_utils.rs
thedavl: yes, thats correct ([corresponding function in the go agent normalizer](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/traceutil/normalize.go))

trace-normalization/src/normalize_utils.rs
thedavl: I tried to mirror exactly what the go tag normalizer does, which:

- uses `isNormalizedASCIITag` as a fast path ([here](https://github.com/DataDog/datadog-agent/blob/e5bb6a4bd4fb57625cac84f270afc943278e6a72/pkg/trace/traceutil/normalize.go#L282))
    - Only considers a lowercase ascii alpha as a valid starting char
    - single underscore is allowed as long as it's followed by a valid non-underscore char
    - `. - /` are valid
    - interestingly only sees ascii alphas as valid (so no other languages)

whereas the tag normalizer itself allows chars from all alphabets

[Here is the go tag normalizer](https://github.com/DataDog/datadog-agent/blob/e5bb6a4bd4fb57625cac84f270afc943278e6a72/pkg/trace/traceutil/normalize.go#L99)

trace-normalization/src/normalize_utils.rs
```diff
@@ -25,23 +32,138 @@ pub(crate) fn truncate_utf8(s: &str, limit: usize) -> &str {
...
+
+// normalize_service normalizes a span service
+pub(crate) fn normalize_service(svc: &str) -> anyhow::Result<String> {
+    anyhow::ensure!(!svc.is_empty(), "Normalizer Error: Empty service name.");
+
+    let truncated_service = if svc.len() > MAX_SERVICE_LEN {
+        truncate_utf8(svc, MAX_SERVICE_LEN)
+    } else {
+        svc
+    };
```
morrisonlevi, line 51: You can simplify this to just call `truncate_utf8`, right? The function looks designed to handle this just fine.

```suggestion
    let truncated_service = truncate_utf8(svc, MAX_SERVICE_LEN);
```

trace-normalization/src/normalize_utils.rs
```diff
@@ -25,23 +32,138 @@ pub(crate) fn truncate_utf8(s: &str, limit: usize) -> &str {
...
+            }
+        }
+        None => return false,
+    }
+
+    for mut i in 0..tag.len() {
+        let cur_char = match tag.chars().nth(i) {
+            Some(c) => c,
+            None => return false,
+        };
```
morrisonlevi, line 138: I think you can pull the iterator out, and then iterate manually using `.next()` instead this juggling. Then instead of calling `.nth(i)` to get the 'current' value, it will just be there. And 'next' can be obtained by calling `.next()` instead of `.nth(i + 1)`.

morrisonlevi (commented):
I have some stylistic suggestions. I'm happy to pair again if you want!

trace-normalization/src/normalize_utils.rs
thedavl: implemented!

morrisonlevi (approved):
Looks good from a generic Rust standpoint, as far as I can tell :)

PR #109, opened 2023-02-09
Trace Normalizer: add trace & trace chunk normalization

thedavl:
# What does this PR do?

This PR adds the public functions `normalize_trace` and `normalize_chunk` to the normalizer, as well as associated unit tests.

# Motivation

# Additional Notes

in the go agent trace normalizer, the `normalize_chunk` function takes in the chunk, and a pointer to the root span ([see here](https://github.com/DataDog/datadog-agent/blob/85f2ce92c3d71988a1cf4693a9b472b25dacdd49/pkg/trace/agent/normalizer.go#L148)). Because of the fact that in rust the root span will have been moved into the spans vec of the trace chunk itself, I've opted to pass the index of the root span within the trace chunk spans vec instead

# How to test the change?

unit tests cover these additions (unit tests are translated/copied from the go agent normalizer)

### DIFF
--- a/trace-normalization/src/normalizer.rs
+++ b/trace-normalization/src/normalizer.rs
@@ -115,7 +115,6 @@ pub(crate) fn is_valid_status_code(sc: &str) -> bool {
 
 // normalize_trace takes a trace and
 // * returns an error if there is a trace ID discrepancy between 2 spans
-// * returns an error if two spans have the same span_id
 // * returns an error if at least one span cannot be normalized
 pub fn normalize_trace(trace: &mut [pb::Span]) -> anyhow::Result<()> {
     if trace.is_empty() {
@@ -139,19 +138,17 @@ pub fn normalize_trace(trace: &mut [pb::Span]) -> anyhow::Result<()> {
     Ok(())
 }
 
-// normalizeChunk takes a trace chunk and
-// * populates Origin field if it wasn't populated
-// * populates Priority field if it wasn't populated
-// the root span is used to populate these fields, and it's index
-// (within the spans vec in TraceChunk) must be passed.
-pub fn normalize_chunk(chunk: &mut pb::TraceChunk, root_index: usize) -> anyhow::Result<()> {
+// normalize_chunk takes a trace chunk and
+// * populates origin field if it wasn't populated
+// * populates priority field if it wasn't populated
+// the root span is used to populate these fields, and it's index in TraceChunk spans vec must be passed.
+pub fn normalize_chunk(chunk: &mut pb::TraceChunk, root_span_index: usize) -> anyhow::Result<()> {
     anyhow::ensure!(
-        chunk.spans.len() > root_index,
-        "Normalize Chunk Error: root_index > length of trace chunk spans"
+        chunk.spans.len() > root_span_index,
+        "Normalize Chunk Error: root_span_index > length of trace chunk spans"
     );
     // check if priority is not populated
-    // a value of i8::MIN (-128) indicates no prior priority value set.
-    let root_span = &chunk.spans[root_index];
+    let root_span = &chunk.spans[root_span_index];
     if chunk.priority == SAMPLER_PRIORITY_NONE {
         // Older tracers set sampling priority in the root span.
         if let Some(root_span_priority) = root_span.metrics.get(TAG_SAMPLING_PRIORITY) {
@@ -165,6 +162,7 @@ pub fn normalize_chunk(chunk: &mut pb::TraceChunk, root_index: usize) -> anyhow:
             }
         }
     }
+    // check if origin is not populated
     if chunk.origin.is_empty() {
         if let Some(origin) = root_span.meta.get(TAG_ORIGIN) {
             // Older tracers set origin in the root span.
--- a/trace-normalization/src/normalizer.rs
+++ b/trace-normalization/src/normalizer.rs
@@ -134,7 +134,7 @@ pub fn normalize_trace(trace: &mut [pb::Span]) -> anyhow::Result<()> {
 // normalizeChunk takes a trace chunk and
 // * populates Origin field if it wasn't populated
 // * populates Priority field if it wasn't populated
-pub fn normalize_chunk(chunk: &mut pb::TraceChunk, root: &mut pb::Span) {
+pub fn normalize_chunk(chunk: &mut pb::TraceChunk, root: pb::Span) {
     // check if priority is not populated
     // a value of i8::MIN (-128) indicates no prior priority value set.
     if chunk.priority == i8::MIN as i32 {
@@ -163,6 +163,7 @@ mod tests {
 
     use crate::normalize_utils;
     use crate::normalizer;
+    use crate::normalizer::DEFAULT_SPAN_NAME;
     use crate::pb;
     use rand::Rng;
     use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


trace-normalization/src/normalizer.rs
```diff
@@ -16,8 +16,18 @@ const YEAR_2000_NANOSEC_TS: i64 = 946684800000000000;
 // DEFAULT_SPAN_NAME is the default name we assign a span if it's missing and we have no reasonable fallback
 const DEFAULT_SPAN_NAME: &str = "unnamed_operation";
 
+const TAG_SAMPLING_PRIORITY: &str = "_sampling_priority_v1";
+const TAG_ORIGIN: &str = "_dd.origin";
+
+#[allow(dead_code)]
```
maxday, line 22: maybe an enum would make sense? (and will avoid the allow dead code)

trace-normalization/src/normalizer.rs
```diff
@@ -103,11 +113,71 @@ pub(crate) fn is_valid_status_code(sc: &str) -> bool {
...
 
+// normalize_trace takes a trace and
+// * returns an error if there is a trace ID discrepancy between 2 spans
+// * returns an error if at least one span cannot be normalized
+pub fn normalize_trace(trace: &mut [pb::Span]) -> anyhow::Result<()> {
+    if trace.is_empty() {
+        anyhow::bail!("Normalize Trace Error: Trace is empty.");
+    }
+
+    let first_trace_id = trace[0].trace_id;
```
maxday, line 124: I think it might be safer to use get(0) which will return a result, then you can check the error.
having decorrelated `is_empty` and `[0]` split could be a risk later?

trace-normalization/src/normalizer.rs
```diff
@@ -103,11 +113,71 @@ pub(crate) fn is_valid_status_code(sc: &str) -> bool {
...
+// * populates origin field if it wasn't populated
+// * populates priority field if it wasn't populated
+// the root span is used to populate these fields, and it's index in TraceChunk spans vec must be passed.
+pub fn normalize_chunk(chunk: &mut pb::TraceChunk, root_span_index: usize) -> anyhow::Result<()> {
+    anyhow::ensure!(
+        chunk.spans.len() > root_span_index,
+        "Normalize Chunk Error: root_span_index > length of trace chunk spans"
+    );
+    // check if priority is not populated
+    let root_span = &chunk.spans[root_span_index];
```
maxday, line 151: same here

maxday (commented):
good work! left some small comments

trace-normalization/src/normalizer.rs
```diff
@@ -103,11 +113,71 @@ pub(crate) fn is_valid_status_code(sc: &str) -> bool {
     false
 }
 
+// normalize_trace takes a trace and
+// * returns an error if there is a trace ID discrepancy between 2 spans
+// * returns an error if at least one span cannot be normalized
```
DarcyRaynerDD, line 118: If you want comments to turn up in autocompletions, or rust documentation, each line should start with three [slashes](https://doc.rust-lang.org/rust-by-example/meta/doc.html#:~:text=Doc%20comments%20are%20very%20useful,a%20%2F%2F%2F%20%2C%20and%20support%20Markdown.&text=%2F%2F%2F%20%2F%2F%20If%20you%20pass,even%20test%20it%20for%20you!).

```suggestion
/// normalize_trace takes a trace and
/// * returns an error if there is a trace ID discrepancy between 2 spans
/// * returns an error if at least one span cannot be normalized
```

trace-normalization/src/normalizer.rs
```diff
@@ -103,11 +113,71 @@ pub(crate) fn is_valid_status_code(sc: &str) -> bool {
...
+        "Normalize Chunk Error: root_span_index > length of trace chunk spans"
+    );
+    // check if priority is not populated
+    let root_span = &chunk.spans[root_span_index];
+    if chunk.priority == SAMPLER_PRIORITY_NONE {
+        // Older tracers set sampling priority in the root span.
+        if let Some(root_span_priority) = root_span.metrics.get(TAG_SAMPLING_PRIORITY) {
+            chunk.priority = *root_span_priority as i32;
+        } else {
+            for span in &chunk.spans {
```
DarcyRaynerDD, line 157: I'm curious why normalize_chunk() doesn't call normalize() on this span list anywhere. Do we do that in another function somewhere?

trace-normalization/src/normalizer.rs
```diff
@@ -16,8 +16,18 @@ const YEAR_2000_NANOSEC_TS: i64 = 946684800000000000;
...
+#[allow(dead_code)]
+const SAMPLER_PRIORITY_AUTO_DROP: i32 = 0;
 #[allow(dead_code)]
-pub fn normalize(s: &mut pb::Span) -> anyhow::Result<()> {
+const SAMPLER_PRIORITY_AUTO_KEEP: i32 = 1;
+#[allow(dead_code)]
+const SAMPLER_PRIORITY_USER_KEEP: i32 = 2;
+const SAMPLER_PRIORITY_NONE: i32 = i8::MIN as i32;
+
+fn normalize(s: &mut pb::Span) -> anyhow::Result<()> {
```
DarcyRaynerDD, line 30: Perhaps this needs a more descriptive name given all the other normalize functions in this file

thedavl: good idea, I'll change it

trace-normalization/src/normalizer.rs
thedavl: totally, I'll change it

trace-normalization/src/normalizer.rs
DarcyRaynerDD: Sorry,  I think I meant normalize_chunk calling normalize_trace

thedavl: ah, I see. `normalize_trace` is called independently from `normalize_chunk`, who's sole purpose is just to populate the `origin` and `priority` fields. [See here in agent.to where the normalizer is called](https://github.com/DataDog/datadog-agent/blob/db20bdd7f29132a243c28feb178a5ff6514aaa90/pkg/trace/agent/agent.go#L252)

trace-normalization/src/normalizer.rs
thedavl: good idea, added

trace-normalization/src/normalizer.rs
```diff
@@ -103,11 +114,72 @@ pub(crate) fn is_valid_status_code(sc: &str) -> bool {
...
+/// normalize_trace takes a trace and
+/// * returns an error if there is a trace ID discrepancy between 2 spans
+/// * returns an error if at least one span cannot be normalized
+pub fn normalize_trace(trace: &mut [pb::Span]) -> anyhow::Result<()> {
+    let first_trace_id = match trace.get(0) {
+        Some(first_span) => first_span.trace_id,
+        None => anyhow::bail!("Normalize Trace Error: Trace is empty"),
+    };
+
+    for span in trace {
```
maxday, line 126: maybe you can simplify this a bit further using the `?` operator
something like
```
for span in trace {
  if span.trace_id != first_trace_id {
    return Err(format!("Normalize Trace Error: Trace has foreign span: {:?}", span));
  }
  normalize_span(span)?;
}
Ok(())
```

thedavl: changed!

maxday (approved):

PR #110, opened 2023-02-14
Initial implementation of spawn_worker crate

pawelchcki:
# What does this PR do?

To be able to safely spawn a worker - and avoid all the pitfalls of only using fork. We need to execute an entrypoint in a separate freshly exec'ed process.

This PR supports exectly that on `unix` systems.
It embeds a simple compiled binary which only purpose is to load a shared library and call one of its methods.

It requires an exported entrypoint to be available in the shared library and will not work from static binaries due to glibc limitations. So to test this code we need to have a test run that is linked dynamically.

# Motivation

# Additional Notes

# How to test the change?


spawn_worker/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+memfd = { version = "0.6" }
+nix = { version = "0.26" }
+tempfile = { version = "3" }
+io-lifetimes = { version = "1.0" }
+
+[build-dependencies]
+cc_utils = { path = "../tools/cc_utils" }
+
+[dev-dependencies]
+tempfile = { version = "3.3" }
```
paullegranddc, line 18: Is this right? I was under the impression you wanted to expose function using this dep not just for tests

paullegranddc (approved):
LGTM overall, although I think there might be simpler than the trampolines on linux because it requires a writable FS

spawn_worker/Cargo.toml
pawelchcki: Ah - sorry - its a leftover, from initial implementation, that only used memfd

spawn_worker/src/unix/spawn.rs
```diff
@@ -0,0 +1,266 @@
...
+            Target::Fork(_) => todo!(),
+            Target::Noop => return Ok(None),
+        };
+
+        let argv = argv.seal();
+
+        let mut envp = ExecVec::default();
+
+        if self.inherit_env {
+            for (k, v) in std::env::vars() {
```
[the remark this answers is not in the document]
paullegranddc: Maybe call `std::env::vars()` in a ctor and cache it?
It's more likely to not encounter race conditions when the library is getting loaded than during runtime. Although we wouldn't get env updates.

spawn_worker/src/unix/spawn.rs
```diff
@@ -0,0 +1,266 @@
...
+                let fd = linux::write_trampoline()?;
+                Box::new(move |argv, envp| {
+                    // not using nix crate here, as it would allocate args after fork, which will lead to crashes on systems
+                    // where allocator is not fork+thread safe
+                    unsafe { libc::fexecve(fd.as_raw_fd(), argv.as_ptr(), envp.as_ptr()) };
+                    // if we're here then exec has failed
+                    panic!("{}", std::io::Error::last_os_error());
+                })
+            }
+            SpawnMethod::Exec => {
```
r1viollet, line 163: what is the mechanism to know if we already did this write and that the daemon is already running ? I'm basically worried about short lived processes starting many instances.

spawn_worker/src/unix/spawn.rs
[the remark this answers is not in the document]
r1viollet: That is a good idea. Alternatively asking the user to launch in a non-threaded place is OK for now ?

spawn_worker/src/unix/spawn.rs
```diff
@@ -0,0 +1,266 @@
...
+                    // if we're here then exec has failed
+                    panic!("{}", std::io::Error::last_os_error());
+                })
+            }
+        };
+
+        match &self.target {
+            Target::Trampoline(f) => {
+                let (library_path, symbol_name) =
+                    match unsafe { crate::get_dl_path_raw(*f as *const libc::c_void) } {
```
r1viollet, line 185: I think this looks good, though we should be extra careful as this is a place a bad user would want to use to inject his library.

spawn_worker/build.rs
```diff
@@ -0,0 +1,11 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+fn main() {
```
r1viollet, line 4: I have trouble figuring out the life cycle of the trampoline. Can you help me out ? 
Compiled -> Stored somewhere in library? (this is the place I'm missing) -> written to memory --> Exec

r1viollet (approved):
LGTM. Nice mechanism !
A few questions that should not block merge on my side.

spawn_worker/src/unix/spawn.rs
```diff
@@ -0,0 +1,266 @@
...
+                )?;
+
+                Box::new(move |argv, envp| {
+                    // not using nix crate here, to avoid allocations post fork
+                    unsafe { libc::execve(path.as_ptr(), argv.as_ptr(), envp.as_ptr()) };
+                    // if we're here then exec has failed
+                    panic!("{}", std::io::Error::last_os_error());
+                })
+            }
+        };
```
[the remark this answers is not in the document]
pawelchcki: I'll add some clarifications. Mostly it was done to avoid doing any allocations between Fork and Exec

spawn_worker/src/unix/spawn.rs
```diff
@@ -0,0 +1,266 @@
...
+                    unsafe { libc::fexecve(fd.as_raw_fd(), argv.as_ptr(), envp.as_ptr()) };
+                    // if we're here then exec has failed
+                    panic!("{}", std::io::Error::last_os_error());
+                })
+            }
+            SpawnMethod::Exec => {
+                let path = CString::new(
+                    write_trampoline()?
+                        .into_temp_path()
+                        .keep()? // ensure the file is not auto cleaned in parent process
```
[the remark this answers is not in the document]
pawelchcki: ahh - yes. That is what I thought too :)

On OSX there is no fexecve - so to use it we need to keep the file - long enough.

By default the desctructor in the child process will remove the file however :)

spawn_worker/build.rs
pawelchcki: https://github.com/DataDog/libdatadog/pull/110/files#diff-29a0a855c8fb3650574054b856d1145f82bb1b8c8ccd5bc50720a2d8b9e76424R4-R5
Its included as bytes by rustc

spawn_worker/src/unix/spawn.rs
pawelchcki: nope, this PR - only has the raw, spawn a subprocess - type functionality.

The logic about when the subprocess should be spawn - lives in a different PR

pawelchcki:
Thanks @r1viollet and @paullegranddc for taking a look! :bow:

PR #113, opened 2023-02-21
Remove direct usage of fork(2) in most tests, use trampolined execution instead

pawelchcki:
# What does this PR do?
This addresses long standing issue with fork being inherently unsafe and unreliable way to run tests - and sidecar itself on production environments.

To enable running tests directly on linux - a new LD_PRELOAD based trampoline was added. It will automatically kick off when the spawn_worker crate is used from a linked executable.

However 1 caveat exists that enables this. The tests need to be compiled with "-rdynamic" flag which will export symbols so that they are available for LD_PRELOAD to look up and execute

# Motivation

Evils of fork

# Additional Notes

Fork is evil

# How to test the change?


ddtelemetry/tests/flock.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    c.write_all(&[0]).unwrap(); // signal readiness
+    let mut buf = [0; 10];
+    assert!(c.read(&mut buf).unwrap() > 0); // wait for signal to closepp
+
+    std::process::exit(0); // exit without explicitly freeing
+}
+
+#[test]
+fn test_file_locking_works_as_expected() {
```
paullegranddc, line 33: can the tests (and test entrypoints) be put in a `#[cfg(test)]` block to not get compiled on release?

paullegranddc (approved):

spawn_worker/src/unix/fork.rs
```diff
@@ -16,7 +18,9 @@ pub enum Fork {
 /// Existing state of the process must allow safe forking, e.g. no background threads should be running
 /// as any locks held by these threads will be locked forever
 ///
-pub unsafe fn fork() -> Result<Fork, std::io::Error> {
+/// When forking a multithreaded application, no code should allocate or access other potentially locked resources
+/// until call to exec is executed
+pub(crate) unsafe fn fork() -> Result<Fork, std::io::Error> {
     let res = libc::fork();
     match res {
         -1 => Err(std::io::Error::last_os_error()),
```
paullegranddc, line 26: Pretty sure this is strictly equivalent to [nix::fork](https://docs.rs/nix/latest/nix/unistd/fn.fork.html)

ddtelemetry/tests/flock.rs
pawelchcki: doesn't need to be. This file is compiled separately as an integration test. It will not get into release

ddtelemetry-ffi/build.rs
```diff
@@ -0,0 +1,9 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+fn main() {
+    // ensure symbols are properly exported for dlsym to be able to look them up
+    // https://github.com/rust-lang/cargo/issues/10937
+    // TODO: only apply this setting in tests
+    println!("cargo:rustc-link-arg=-rdynamic")
```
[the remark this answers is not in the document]
pawelchcki: Hmm. I need to refactor this stiff to integration tests - too. Minimal build.rs should compile long ?

spawn_worker/src/unix/fork.rs
pawelchcki: You're right :facepalm:  
I'll promise to fix this in other pr :pinky_swear:

pawelchcki:
Thanks @paullegranddc and @r1viollet for the reviews :bow:

spawn_worker/src/unix/mod.rs
```diff
@@ -45,53 +54,56 @@ pub unsafe fn get_dl_path_raw(addr: FnPointer) -> (Option<CString>, Option<CStri
...
+
+pub fn recv_passed_fd() -> Option<OwnedFd> {
+    let val = env::var(ENV_PASS_FD_KEY).ok()?;
+    let fd: RawFd = val.parse().ok()?;
+
+    // check if FD number is valid
+    nix::fcntl::fcntl(fd, nix::fcntl::FcntlArg::F_GETFD).ok()?;
+
+    Some(unsafe { OwnedFd::from_raw_fd(fd) })
+}
```
paullegranddc, line 86: This reminds me that we should probably close all non passed file descriptors, in order to not keep alive the applications files and socket open, since the sidecar could outlive the parent process

spawn_worker/src/unix/spawn.rs
```diff
@@ -48,61 +64,155 @@ impl ExecVec {
...
 
-pub enum Target {
-    ViaFnPtr(extern "C" fn()),
-    Manual(CString, CString),
-    Fork(fn()),
-    Noop,
+enum ChildStdio {
+    Inherit,
+    Owned(OwnedFd),
+    Ref(libc::pid_t),
```
paullegranddc, line 148: A raw fd should be of type `std::os::fd::RawFd`, `libc::pid_t` is confusing

spawn_worker/src/unix/spawn.rs
```diff
@@ -48,61 +64,155 @@ impl ExecVec {
...
+impl Stdio {
+    fn as_child_stdio(&self) -> std::io::Result<ChildStdio> {
+        match self {
+            Stdio::Inherit => Ok(ChildStdio::Inherit),
+            Stdio::Fd(fd) => {
+                if fd.as_raw_fd() >= 0 && fd.as_raw_fd() <= libc::STDERR_FILENO {
+                    Ok(ChildStdio::Owned(fd.try_clone()?))
+                } else {
+                    Ok(ChildStdio::Ref(fd.as_raw_fd()))
+                }
```
paullegranddc, line 176: I don't understand this one. Why couldn't we clone non stdin/stdout/stderr file descriptors?

spawn_worker/src/unix/spawn.rs
```diff
@@ -183,71 +319,138 @@ impl SpawnCfg {
             }
         }
 
-        type SpawnFn = dyn Fn(&ExecVec, &ExecVec);
+        // setup arbitrary fd passing
+        let _shorter_lived_fd = if let Some(src_fd) = &self.fd_to_pass {
+            // we're stripping the close on exec flag from the FD
+            // to ensure we will not modify original fd, whose expected lifetime is unknown
+            // we should clone the FD that needs passing to the subprocess, keeping its lifetime
+            // as short as possible
```
paullegranddc, line 327: I disagree with this, on fork all of the processes fd gets cloned, so I am not sure what cloning an fd to a "smaller" lifetime improves..
On the `CLOEXEC` argument I am split.
* Passing a fd with the flag set should be an error since the creator of the fd specified that they don't want them to get passed to forks
* On the other hand, by default `try_clone` on io-lifetimes fds adds this flag https://doc.rust-lang.org/src/std/os/fd/owned.rs.html#94 . But IMO the fix should be either stopping to clone OwnedFds all over the place and use Borrowed ones instead, or use our own clone function that skips this flag.

spawn_worker/src/unix/mod.rs
pawelchcki: That is the beauty of using exec! we don't have to! 
Assuming we can trust most cases where people might have some sockets, and if they don't like those sockets to be shared with other processes they will do set `CLOSEXEC`

....

Now I realized that its the APM world we're talking about here :| - and maybe we should close them

spawn_worker/src/unix/spawn.rs
pawelchcki: this code follows what Command::spawn is doing. Also since stdio fd's are not CLOSEXEC by default then we don't need to dupe them.

PR #114, opened 2023-02-27
ld preloadable libdatadog trace sidecar

pawelchcki:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


sidecar/src/connections.rs
```diff
@@ -0,0 +1,171 @@
...
+    ) -> Poll<std::io::Result<()>> {
+        self.project().inner.poll_read(cx, buf)
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct Tracker {
+    count: Arc<AtomicUsize>,
+    notifier: Arc<tokio::sync::Notify>,
+}
```
paullegranddc, line 105: One arc is better than two maybe?

```suggestion
struct TrackerInner {
    count: AtomicUsize,
    notifier: tokio::sync::Notify,
}

pub struct Tracker {
  inner: Arc<TrackerWatcherInner>
}
```

sidecar/src/connections.rs
```diff
@@ -0,0 +1,171 @@
...
+            if prev_count == count
+                && count == 0
+                && prev_time.elapsed() >= min_duration_without_instances
+            {
+                return;
+            }
+
+            prev_count = count;
+            prev_time = tokio::time::Instant::now();
+        }
```
paullegranddc, line 162: Note for myself to review this better

sidecar/src/connections.rs
pawelchcki: I need to be able to clone either the counter or notifier separately

sidecar/src/connections.rs
pawelchcki: :sweat_smile: 
better not ;)

PR #117, opened 2023-03-02
trace-obfuscation: Add is_card_number func to identify credit cards

ajgajg1134:
# What does this PR do?

# Motivation
As part of R&D week I've been playing with libdatadog and FFI in go, I realized some of the obfuscation the agent does was missing from libdatadog so I decided to try and port the datadog-agent's "IsCreditCard" function. For now this won't be used, but can serve as a starting point for additional obfuscation in libdatadog

# Additional Notes
My rust is a bit... rusty... 😎 . So please make sure what I'm doing is idiomatic and I'm not making any silly mistakes / missing any good refactors here. The original go code it pretty tough to read and I believe this new implementation is far more readable, however I avoided changing the design _too_ much to reduce the likelihood I introduce any bugs in the port. 

# How to test the change?
All the unit tests from datadog-agent were brought over (And a few new tests for my own benefit were added)


trace-obfuscation/src/credit_cards.rs
```diff
@@ -0,0 +1,422 @@
...
+        } else {
+            *val
+        };
+        acc += x;
+    }
+    10 - (acc % 10) % 10
+}
+
+#[derive(Debug, PartialEq)]
+enum FuzzyBool {
```
pawelchcki, line 79: :heart:

trace-obfuscation/src/credit_cards.rs
```diff
@@ -0,0 +1,422 @@
...
+/// If validateLuhn is true, the Luhn checksum is also applied to potential candidates.
+/// Note: This code is based on the code from datadog-agent/pkg/obfuscate/credit_cards.go
+pub fn is_card_number(s: &str, validate_luhn: bool) -> bool {
+    if s.len() < 12 {
+        // fast path
+        return false;
+    }
+
+    // Only valid characters are 0-9, space (" ") and dash("-")
+    let num_s = s.replace([' ', '-'], "");
```
pawelchcki, line 16: since its a hackathon, I'll dare to propose, doing one single loop. 
something like:
```rust 
let mut num_s = Vec::with_capacity(s.len());
for i in 0..s.len() {
let c = s[i];
match c {
  ' ' | '-' =>  continue
  _ => {}
}
if !c.is_numeric {
return false
}
num_s.push(c);
}

**Comment by pawelchcki** [comment]
Overall looks good to me. 
I'd suggest setting up a criterion benchmark for `is_card_number` too - just so we can track the :speedboat: in the future

File: trace-obfuscation/src/credit_cards.rs (hunk 2 above)
Comment (ajgajg1134) [reply trace-obfuscation/src/credit_cards.rs]: Doing this in a single loop was so much better, it also meant I could move the `to_digit` call from two places to a single place and then simplified my arguments so I didn't have to move and copy things around quite so much.

**Comment by ajgajg1134** [comment]
> Overall looks good to me. I'd suggest setting up a criterion benchmark for `is_card_number` too - just so we can track the 🚤 in the future

I added some criterion benchmarks which are hopefully about what you expected? I've never written criterion benchmarks before but these seemed to be "good enough". Let me know if I shouldn't be using a slice of data and grouping and just do a single test of 1 valid credit card

**Comment by thedavl** [comment]
you're going to want to add the path to your new bench file [here](https://github.com/DataDog/libdatadog/blob/d5e5aa96094628724e28767a3294d463bbf135cd/tools/docker/Dockerfile.build#L88) to fix that alpine docker build issue!

File: trace-obfuscation/src/credit_cards.rs [hunk 3]
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    let mut num_s = Vec::with_capacity(s.len());
+    for c in s.chars() {
+        // Only valid characters are 0-9, space (" ") and dash("-")
+        match c {
+            ' ' | '-' => continue,
+            _ => (),
+        };
+        if !c.is_numeric() {
```
morrisonlevi, line 22: Is this supposed to be ascii numeric? Rust's chars are not the same as C chars, FYI.

morrisonlevi: These would also be good test cases to add :)

ajgajg1134: Oooo that's good to know, yeah I think this should be `is_ascii_digit`  I'll fix this

trace-obfuscation/src/credit_cards.rs
```diff
@@ -0,0 +1,419 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present
+// Datadog, Inc.
+
+/// is_card_number checks if b could be a credit card number by checking the digit count and IIN prefix.
+/// If validateLuhn is true, the Luhn checksum is also applied to potential candidates.
+/// Note: This code is based on the code from datadog-agent/pkg/obfuscate/credit_cards.go
+pub fn is_card_number(s: &str, validate_luhn: bool) -> bool {
```
pawelchcki, line 9: A little usability improvement could be made if `AsRef` trait is used for s
https://doc.rust-lang.org/std/convert/trait.AsRef.html#examples

```suggestion
pub fn is_card_number<T: AsRef<str>>(s: T, validate_luhn: bool) -> bool {
     let s = s.as_ref();
```

pawelchcki (approved):
One nit from me and one important comment from Levi.

Otherwise it looks good to me.

Thanks @ajgajg1134 :bow:

r1viollet:
Should we merge this one ?

morrisonlevi:
@r1viollet, I think it's waiting on a review from the serverless team, since it's part of a folder they own.

thedavl (approved):
Aside from the clippy errors, everything looks good! Just fix those before merging

thedavl:
> > Aside from the clippy errors, everything looks good! Just fix those before merging
> 
> So it looks like those clippy failures are outside this PR? They seem to be in the trace-normalization code which isn't touched here? Do you still want me to fix those in this PR?

Ah, sorry. I missed that they weren't touched in this PR. Must have been some clippy change, as it was green before.

That's fine then, you're good to merge!

ajgajg1134:
> > > Aside from the clippy errors, everything looks good! Just fix those before merging
> > 
> > 
> > So it looks like those clippy failures are outside this PR? They seem to be in the trace-normalization code which isn't touched here? Do you still want me to fix those in this PR?
> 
> Ah, sorry. I missed what files it was failing on. Must have been some clippy change, as it was green before.
> 
> That's fine then, you're good to merge!

Sweet! I don't have permissions to merge so hopefully someone else can do so :)

ivoanjo:
Thanks for the great contribution y'all, pressing the magic button!

PR #118, opened 2023-03-03
Implement values upscaling

gleocadie:
# What does this PR do?

This PR add an API to allow upscaling values.

# Motivation

For performance reason, profilers do not collect every events they are notified (ex: every allocations, every exceptions...). The samples we collect statically represent  the behavior of the application but the numbers we show the sampled one and not the real one.

# How to test the change?

- Added tests 
- Tested the Proportional upscaling computation with .NET profiler: The application throws 4000 exceptions and the final result is 3998. So the error margin is really low.


profiling/src/profile/mod.rs
```diff
@@ -561,6 +633,49 @@ impl Profile {
...
+            anyhow::ensure!(
+                values_to_update.iter().all(|v| *v < 2),
+                "Multiple rules modifying the same offset for this sample"
+            );
+
+            group_of_rules.iter().for_each(|rules| {
+                rules.iter().for_each(|rule| {
+                    rule.values_offset
+                        .iter()
+                        .for_each(|offset| new_values[*offset] *= rule.ratio as i64)
```
nsavoire, line 672: This first converts `ratio` to a `i64` and then multiply the value by it, I would expect to first convert value to f64 then scale the value then convert it back to i64 (should the conversion do a round to nearest, floor or round towards zero ?).

profiling/src/profile/mod.rs
```diff
@@ -81,6 +81,13 @@ struct Sample {
     local_root_span_id_label_offset: Option<usize>,
 }
 
+struct UpscalingRules {
+    values_offset: Vec<usize>,
+    label_name: i64,
+    label_value: i64,
+    ratio: f64,
```
nsavoire, line 88: `scale` would be more explicit than `ratio`

profiling-ffi/src/profiles.rs
```diff
@@ -438,6 +466,42 @@ pub unsafe extern "C" fn ddog_prof_Profile_add_endpoint_count(
...
+) -> UpscalingRuleAddResult {
+    let label_name_n = label_name.to_utf8_lossy();
+    let label_value_n = label_value.to_utf8_lossy();
+    match profile.add_upscaling_rules(
+        offset_values.as_slice(),
+        label_name_n.as_ref(),
+        label_value_n.as_ref(),
+        upscaling_info.into(),
+    ) {
+        Ok(_) => UpscalingRuleAddResult::Ok(true),
```
morrisonlevi, line 500: What is the `bool` for, if it's always true?

gleocadie: Good point, I do not know how to make empty `void`
```
pub enum UpscalingRuleAddResult {
    Ok(()),
    Err(Error),
}
```
The C code generated from this is invalid.
What would you advise ?

gleocadie: thank :)
PR updated following our private discussion

profiling/src/profile/mod.rs
```diff
@@ -561,6 +662,50 @@ impl Profile {
             .get(&local_root_span_id)
             .map(i64::clone))
     }
+
+    fn upscale_values(&self, values: &[i64], labels: &[Label]) -> anyhow::Result<Vec<i64>> {
+        let mut new_values = values.to_vec();
+
+        if !self.upscaling_rules.is_empty() {
```
realFlowControl, line 669: nitpick (non-blocking): I personally prefer the use of guard clauses or some call it early-exit. Then the rest of the code would not be indented that far. It is just a personal preference, not hard feelings, feel free to ignore.

```suggestion
        if self.upscaling_rules.is_empty() {
            Ok(new_values)
        }
```

profiling/src/profile/mod.rs
```diff
@@ -1155,6 +1303,1179 @@ mod api_test {
         profile.add(sample).unwrap_err();
     }
 
+    #[test]
```
realFlowControl, line 1306: praise: Thanks for writing tests for all the added functionality 🎉

profiling/src/profile/mod.rs
```diff
@@ -81,6 +83,28 @@ struct Sample {
...
+    pub fn compute_scale(&self, values: &[i64]) -> f64 {
+        match self.upscaling_info {
+            UpscalingInfo::Poisson { x, y, threshold } => {
+                let avg = (values[x] / values[y]) as f64;
+                1_f64 / (1_f64 - (-avg / threshold as f64).exp())
+            }
+            UpscalingInfo::Proportional {
+                total_sampled,
+                total_real,
+            } => total_real as f64 / total_sampled as f64,
```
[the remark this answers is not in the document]
gleocadie: If `total_sampled` == `0` it means that `total_real` == `0`. In that case, why bothering adding a rule.
That said, better safe than sorry, I'll add a test and a test :)
Thx

profiling/src/profile/mod.rs
```diff
@@ -561,6 +662,50 @@ impl Profile {
...
+                    rule.values_offset
+                        .iter()
+                        .for_each(|offset| values_to_update[*offset] += 1)
+                })
+            });
+
+            anyhow::ensure!(
+                values_to_update.iter().all(|v| *v < 2),
+                "Multiple rules modifying the same offset for this sample"
+            );
```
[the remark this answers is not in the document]
gleocadie: In this case it's runtime, you cannot fail earlier.
There are static checks done when adding them but how can you check that two rules modifying the same values but with different label name / label value will apply to the same pprof row ?

This is a runtime problem and not sure you can find it earlier.

profiling/src/profile/api.rs
```diff
@@ -128,6 +128,11 @@ pub struct Sample<'a> {
     pub labels: Vec<Label<'a>>,
 }
 
+pub enum UpscalingInfo {
+    Poisson { x: usize, y: usize, threshold: i64 },
```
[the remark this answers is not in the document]
gleocadie: agree `lambda` will not be suitable in that case, mostly confusing.

profiling/src/profile/mod.rs
gleocadie: For me it depends of the cases, in this case I would have to return `Ok(new_values)` at two callsites (where you spot it and at the end).
Which means, potential error (if there is a change and on a callsite you have `Ok(new_values)` and the other you have `Ok(values)` )

realFlowControl (approved):
@ivoanjo I would appreciate a review from you

profiling/src/profile/mod.rs
```diff
@@ -1155,6 +1320,1224 @@ mod api_test {
...
+            total_real: 20,
+        };
+        profile
+            .add_upscaling_rules(
+                value_offsets.as_slice(),
+                id_label.key,
+                id_label.str.unwrap(),
+                upscaling_info2,
+            )
+            .expect("Rule added");
```
nsavoire, line 2535: `add_upscaling_rules` should fail at this point since rules always collide (if the second rule applies, then first one will also apply).

profiling/src/profile/mod.rs
```diff
@@ -81,6 +83,42 @@ struct Sample {
...
+            UpscalingInfo::Poisson {
+                x,
+                y,
+                sampling_distance,
+            } => {
+                if values[y] == 0 {
+                    return 0_f64;
+                }
+
+                let avg = (values[x] / values[y]) as f64;
```
nsavoire, line 105: **❔ question:** ‏ What happens if `avg==0` (eg. when `value[x]==0`) ?

profiling/src/profile/mod.rs
gleocadie: You cannot tell until it runs. Because, if you do not have a pprof row with this specific label, it's still valid

profiling/src/profile/mod.rs
```diff
@@ -1155,6 +1320,1224 @@ mod api_test {
         profile.add(sample).unwrap_err();
     }
 
+    #[test]
```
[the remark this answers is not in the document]
gleocadie: yeah maybe 😅

profiling/src/profile/mod.rs
nsavoire: When you check for rule overlap, you should also check for overlap for `by value` rules (I added a comment there too), then you would catch this.
You can argue that a pprof without this rule is valid, but the same argument could be used for not checking for overlap when adding rules (ie. if the overlapping rules do not apply, then pprof is valid).
My argument is that this rule should be rejected because if it applies, then it will overlap an existing by value rule.

profiling/src/profile/mod.rs
gleocadie: there will computation with this statement
`1_f64 / (1_f64 - (-0 / sampling_distance as f64).exp())` replacing avg with 0
Which might end up with to `1_f64`

nsavoire: More like `+Inf` ;)

profiling/src/profile/mod.rs
```diff
@@ -491,6 +531,84 @@ impl Profile {
...
+        upscaling_info: UpscalingInfo,
+    ) -> anyhow::Result<()> {
+        let label_name_id = self.intern(label_name);
+        let label_value_id = self.intern(label_value);
+
+        let mut new_values_offset = offset_values.to_vec();
+        new_values_offset.sort_unstable();
+
+        fn is_overlapping(v1: &[usize], v2: &[usize]) -> bool {
+            v1.iter().any(|x| v2.contains(x))
```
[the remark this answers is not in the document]
gleocadie: yep, I did not find an `intersection` function on vector, do you know of any ?
The other stuff I found was on hashset.... building 2 hashets for our case loop heavy and complex IMO

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
gleocadie: ✔️

gleocadie: oh I forgot the `exp()` :)

gleocadie: I missed the `exp()`.
In that case, what do you propose to return ?

profiling-ffi/src/profiles.rs
```diff
@@ -23,6 +23,48 @@ pub enum SerializeResult {
...
+pub enum UpscalingRuleAddResult {
+    // Do not use the value of Ok. This value only exists to overcome Rust -> C code generation.
+    Ok(bool),
+    Err(Error),
+}
+
+#[repr(C)]
+pub enum UpscalingInfo {
+    Poisson {
+        x: usize,
```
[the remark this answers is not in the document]
realFlowControl: `x` and `y` refer to offsets in a pprof row (which are `alloc-size` and `alloc-samples` in case of allocation profiling). I am also not happy with the naming, but I can't think of something better ATM

profiling/src/profile/mod.rs
gleocadie: That's what I had in mind, good point
When adding the rule, we can refuse the rule if the `distance_sampling` is 0. WDYT?

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
nsavoire: You could do something like:
```rust
fn is_overlapping(v1: &[usize], v2: &[usize]) -> bool {
    let mut i1 = 0;
    let mut i2 = 0;
    while i1 < v1.len() && i2 < v2.len() {
        match v1[i1].cmp(&v2[i2]) {
            Ordering::Equal => return true,
            Ordering::Less => i1 += 1,
            Ordering::Greater => i2 += 1,
        }
    }
    false
}
```

profiling/src/profile/mod.rs
nsavoire: While I agree that `value[x]` should not be 0 if `value[y]` is not 0, nothing prevents you from adding samples with 0 values. In the case `value[x]` or `value[y]` is 0, I would just return `1` (or `0`, but `1` is the neutral value for scaling).

profiling-ffi/src/profiles.rs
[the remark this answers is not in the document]
nsavoire: I agree it's not easy to name
`x` is the offset of the value that is a kind of sum => `sum_value_offset`
`y` is the offset of the value that is a kind of count => `count_value_offset`
WDYT ?

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
nsavoire: It's a nitpick, you are free to ignore this (and I won't be offended if do ;) )

profiling/src/profile/mod.rs
```diff
@@ -1155,6 +1320,1224 @@ mod api_test {
...
+                r#type: "wall-time",
+                unit: "nanoseconds",
+            },
+        ];
+
+        let mut profile: Profile = Profile::builder().sample_types(sample_types).build();
+
+        let id_label = api::Label {
+            key: "my label",
+            str: Some("coco"),
```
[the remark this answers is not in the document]
gleocadie: ✔️

profiling/src/profile/mod.rs
```diff
@@ -491,6 +531,84 @@ impl Profile {
...
+        anyhow::ensure!(
+            self.upscaling_rules
+                .iter()
+                .all(|(_, rules)| rules.iter().all(|rule| !is_overlapping(
+                    &rule.values_offset,
+                    &new_values_offset
+                ) || rule.label_name != label_name_id
+                    || rule.label_value != label_value_id)),
+            "Duplicated upscaling rule"
+        );
```
ivoanjo, line 560: I wonder: can this get a bit expensive? I'm assuming that for instance in the exception profiling case we may call this method hundreds or thousands of times (for every unique exception name in the app).

(Would it be preferable to perhaps only this validation inline with serialization? Or perhaps do it in a single pass after all rules were added?)

profiling/src/profile/mod.rs
```diff
@@ -81,6 +83,42 @@ struct Sample {
...
+            UpscalingInfo::Proportional {
+                total_sampled,
+                total_real,
+            } => {
+                if total_sampled == 0 {
+                    return 0_f64;
+                }
+
+                total_real as f64 / total_sampled as f64
+            }
```
ivoanjo, line 117: Should this perhaps be precalculated when creating the internal `UpscalingRule`? Unlike the poisson approach, this is always going to be a constant (perhaps we don't even need to store the sampled/read internally -- even though I agree that it's nicer if the API caller specifies them separately).

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
ivoanjo: This is definitely a lot of code -- this file grows by more than 50% with this change. I've left a few suggestions below of repeated things but... I think the overall tests do lose a bit by having so much repetition (e.g. it's hard to see what changes from test to test).

profiling/src/profile/mod.rs
```diff
@@ -1155,6 +1320,1224 @@ mod api_test {
...
+            },
+            api::ValueType {
+                r#type: "wall-time",
+                unit: "nanoseconds",
+            },
+            api::ValueType {
+                r#type: "cpu-time",
+                unit: "nanoseconds",
+            },
+        ];
```
ivoanjo, line 1376: Minor: I think factoring out the sample types would be a nice win in reducing a lot of the repetition.

profiling/src/profile/mod.rs
```diff
@@ -1155,6 +1320,1224 @@ mod api_test {
...
+            .add_upscaling_rules(values_offset.as_slice(), "", "", upscaling_info)
+            .expect("Rule added");
+
+        let serialized_profile = pprof::Profile::try_from(&profile).unwrap();
+
+        let first = serialized_profile.samples.get(0).expect("one sample");
+
+        assert_eq!(first.values[0], 0);
+        assert_eq!(first.values[1], 10000);
+        assert_eq!(first.values[2], 42);
```
ivoanjo, line 1403: Minor: Is it possible to assert on all values at once? E.g. `assert_eq!(first.values, vec![0, 10000, 42])` it would also save some of the repetition.

profiling/src/profile/mod.rs
```diff
@@ -1155,6 +1320,1224 @@ mod api_test {
...
+        };
+        profile
+            .add_upscaling_rules(
+                by_value_offsets.as_slice(),
+                "my label",
+                "coco",
+                upscaling_info,
+            )
+            .expect_err("Invalid offset");
+    }
```
ivoanjo, line 2476: Minor: Since we already test both x and y out of bounds, I'm not sure testing "both out of bounds" adds a lot, it exercises the same code as above. (Unless I'm missing something?)

ivoanjo (commented):
Sorry for taking so long! Here's my delayed review. There's a few improvements that I suggested, but at this point I think my feedback shouldn't be considered blocking.

profiling-ffi/src/profiles.rs
```diff
@@ -23,6 +23,48 @@ pub enum SerializeResult {
...
+        y: usize,
+        sampling_distance: i64,
```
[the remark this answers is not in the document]
gleocadie: From @realFlowControl , `sampling_distance` is the term he saw in math formula

profiling-ffi/src/profiles.rs
```diff
@@ -438,6 +480,43 @@ pub unsafe extern "C" fn ddog_prof_Profile_add_endpoint_count(
...
+/// * `ratio` - value to multiply the sample values at `offset_values` against
+///
+/// # Safety
+/// This function must be called before serialize and must not be called after.
+/// The `profile` ptr must point to a valid Profile object created by this
+/// module.
+/// This call is _NOT_ thread-safe.
+#[must_use]
+#[no_mangle]
+pub unsafe extern "C" fn ddog_prof_Profile_add_upscaling_rule(
```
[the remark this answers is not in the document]
gleocadie: To be honest, I did not see any "how to use API" documentation for the other, so that's why I did not think about adding.
In addition to that, there is the example where you could see how to use it.
Maybe we should reconsider this for all the FFI function.

profiling/src/profile/mod.rs
gleocadie: This can be a bit expensive, I changed a bit the implementation, so for `by-label` rules, this should be more efficient.
For the exception profiling, you will call this as much as unique exception name you have. For .NET application, we do not throw a lot of unique exception name, but maybe other can.

> Would it be preferable to perhaps only this validation inline with serialization? Or perhaps do it in a single pass after all rules were added?

Catching rule collision when adding can provide more information on what failed (again the new implementation tries to achieve that)

profiling/src/profile/mod.rs
gleocadie: You are right, I was wondering if I can cache it when access the first time or like you said, we know it will result in a constant, and compute it when adding it.

profiling/src/profile/mod.rs
gleocadie: will do

profiling/src/profile/mod.rs
gleocadie: good point, I did not know

profiling/src/profile/mod.rs
gleocadie: I agree that feels redundant, I wonder what happen if the code change and this test is removed and nothing fails ?
because in terms of tests cases, this is a valid one.

gleocadie:
> Sorry for taking so long! Here's my delayed review. There's a few improvements that I suggested, but at this point I think my feedback shouldn't be considered blocking.

no worries, your review is welcomed :) I'll take them into account.

profiling-ffi/src/profiles.rs
```diff
@@ -438,6 +480,43 @@ pub unsafe extern "C" fn ddog_prof_Profile_add_endpoint_count(
...
+    profile: &mut Profile,
+    offset_values: Slice<usize>,
+    label_name: CharSlice,
+    label_value: CharSlice,
+    upscaling_info: UpscalingInfo,
+) -> UpscalingRuleAddResult {
```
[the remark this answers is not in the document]
gleocadie: Good point, I went that way because I do not know how much upscaling function we would have.
I'll your way, will see later :)

profiling/src/profile/mod.rs
```diff
@@ -561,6 +679,50 @@ impl Profile {
...
+                    rule.values_offset
+                        .iter()
+                        .for_each(|offset| values_to_update[*offset] += 1)
+                })
+            });
+
+            anyhow::ensure!(
+                values_to_update.iter().all(|v| *v < 2),
+                "Multiple rules modifying the same offset for this sample"
+            );
```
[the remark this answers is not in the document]
gleocadie: It depends what we want: partial upscaling or no upscaling in case we have rules collision.
As you might read, I went for `do not touch my values if you intend to fail`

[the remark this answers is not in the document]
gleocadie: Side because I tried: you cannot return a value from a `for_each`. Type incompatibility (between `()` and `anyhow::Error`)

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
gleocadie: Sorry I tried to make it easy. I followed what was done for the other tests, naively thinking that others would understand (except me because not enough experience in Rust 😅 )

profiling/src/profile/mod.rs
```diff
@@ -491,6 +531,84 @@ impl Profile {
             .add_endpoint_count(endpoint.into_owned(), value);
     }
 
+    pub fn add_upscaling_rules(
```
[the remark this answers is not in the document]
gleocadie: ✔️

profiling/src/profile/mod.rs
gleocadie: In the end, when going from FFI to Rust `pure` (do not know the term), I compute the scale once and for all.

profiling-ffi/src/profiles.rs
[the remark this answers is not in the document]
ivoanjo: Ah! It wasn't completely obvious to me, but I realize naming is always hard :)

profiling-ffi/src/profiles.rs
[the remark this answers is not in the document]
ivoanjo: Yeah. The example in this case is a quite simple use of the feature (no-op scaling, no labels, only a single index "0") which is also why I couldn't get a lot from it.

profiling/src/profile/mod.rs
ivoanjo: (Not blocking for this PR!)

The newer version indeed is better but... one thing that I'm not a big fan of is that for instance the checks penalize adding by-value first, and then by-label later -- because in this case we check every with-label against all the by-value rules; it's much better to add all by-label first (check checks) and then by-value later.

But... the API doesn't really push clients to do that. So clients can get a more expensive behavior if they haven't noticed how the validation code works.

And that's.... a bit of a sharp edge. I guess we can document it, but it would be nice if either the checks either were independent of API calling order, of it the API forced clients to use it correctly -- by perhaps not allowing the addition of by-value rules after by-label rules have been added.

profiling/src/profile/mod.rs
ivoanjo: Maybe... my thinking was -- what weird change would need to happen for the "x out of bounds" and "y out of bounds" individual tests to still pass, and the "both x and y out of bounds" test to fail?

Nevertheless, indeed the reason I marked my feedback as minor is that it's a small thing, so if you're not convinced, I can definitely live with that ;)

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
ivoanjo: Right! My thinking is that if something is wrong, something is wrong. I'll probably not report a profile, will write out an error message, and stop the profiler.

But indeed if you were going for "let's try to keep going", this approach is slightly more annoying (although we could keep a copy of the values -- probably cheap enough with the amount we use).

profiling/src/profile/mod.rs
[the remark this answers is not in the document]
ivoanjo: Well I probably have less Rust-foo than you, so that's part of the trouble hahaha :)

profiling-ffi/cbindgen.toml
```diff
@@ -41,6 +41,8 @@ renaming_overrides_prefixing = true
 "SendResult" = "ddog_prof_Exporter_SendResult"
 "SerializeResult" = "ddog_prof_Profile_SerializeResult"
 "Slice_File" = "ddog_prof_Exporter_Slice_File"
+"UpscalingRuleAddResult" = "ddog_prof_Profile_UpscalingRuleAddResult"
+"UpscalingInfo" = "ddog_prof_Profile_UpscalingInfo"
```
ivoanjo, line 45: Minor: This is no longer needed (since now the FFI exposes specific functions to add each upscaling variant)

ivoanjo (approved):
:+1: LGTM

profiling-ffi/cbindgen.toml
gleocadie: thx :) will clean this up

PR #119, opened 2023-03-10
Add scripts to builds artifacts on windows

gleocadie:
# What does this PR do?

Add scripts and files to build windows artifacts: libraries and nuget package

# Motivation

The goal is to allow developers to build libdatadog locally on windows and also in CI.

# Additional Notes

None

# How to test the change?

Run the tests locally

### DIFF
--- a/windows/libdatadog.props
+++ b/windows/libdatadog.props
@@ -2,23 +2,23 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003"
         ToolsVersion="15.0">
   <PropertyGroup>
-    <LIBDDPROF-PLATFORM Condition="'$(Platform)'=='x64'">x64</LIBDDPROF-PLATFORM>
-    <LIBDDPROF-PLATFORM Condition="'$(Platform)'=='Win32' OR '$(Platform)'=='x86'">x86</LIBDDPROF-PLATFORM>
+    <LIBDATADOG-PLATFORM Condition="'$(Platform)'=='x64'">x64</LIBDATADOG-PLATFORM>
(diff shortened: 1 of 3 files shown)


windows/libdatadog.props
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003"
+        ToolsVersion="15.0">
+  <PropertyGroup>
+    <LIBDDPROF-PLATFORM Condition="'$(Platform)'=='x64'">x64</LIBDDPROF-PLATFORM>
+    <LIBDDPROF-PLATFORM Condition="'$(Platform)'=='Win32' OR '$(Platform)'=='x86'">x86</LIBDDPROF-PLATFORM>
```
ivoanjo, line 6: Minor: Should this be `libdatadog` instead of `libddprof`?

ivoanjo (approved):
I can't test this since I don't have a working windows C# build environment, but it looks reasonable and the level of risk of merging looks to me to be very low, so here, take my :+1: 

(If I'm wrong and there are risks here that I haven't thought about -- do chime in)

windows/libdatadog.props
gleocadie: oh good point.

Yeah the risk is low, since it used to build `nuget package`. 
I tested the generated package on the profiler and it looks like it's working.
This has been migrated from a `nuspec` file (old way to describe nuget package generation)

morrisonlevi (approved):
This is basically a foreign language to me. I don't see how it could hurt anything, though, so I've approved it. You should maybe double-check with Christophe, though.

gleocadie:
> This is basically a foreign language to me. I don't see how it could hurt anything, though, so I've approved it. You should maybe double-check with Christophe, though.

I'll check with him, but it's also foreign language to him 🤣

PR #120, opened 2023-03-10
Fix clippy and flaky test

pawelchcki:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/ipc/tests/blocking_client.rs
+++ b/ipc/tests/blocking_client.rs
@@ -25,6 +25,7 @@ fn test_blocking_client() {
             .unwrap();
         let _g = rt.enter();
         sock_a.set_nonblocking(true).unwrap();
+
         let socket = UnixStream::from_std(sock_a).unwrap();
         let server = ExampleServer::default();
 
@@ -37,7 +38,7 @@ fn test_blocking_client() {
     transport.set_nonblocking(false).unwrap(); // write should still be quick, but we'll have to block waiting for RPC worker to come up
 
     transport
-        .set_write_timeout(Some(Duration::from_nanos(1)))
+        .set_write_timeout(Some(Duration::from_millis(100)))
         .unwrap();
     match transport.call(ExampleInterfaceRequest::TimeNow {}).unwrap() {
         ExampleInterfaceResponse::TimeNow(time) => {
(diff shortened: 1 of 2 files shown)


ipc/tests/blocking_client.rs
```diff
@@ -70,7 +71,4 @@ fn test_blocking_client() {
 
     let mut f = f.into_instance().unwrap();
     writeln!(f, "test").unwrap(); // file should still be writeable
-
-    drop(transport);
-    worker.join().unwrap();
```
ivoanjo, line 75: I'm curious about this change, will this create flakiness? E.g. if this thread stays alive and is concurrent with other tests.

pawelchcki: You're right - I changed the test to drop the tokio runtime on exit. :bow:

spawn_worker/Cargo.toml
```diff
@@ -6,7 +6,7 @@ version = "0.0.1"
 [dependencies]
 anyhow = { version = "1.0" }
 memfd = { version = "0.6" }
-nix = { version = "0.26" }
+nix = { version = "0.24" }
```
ivoanjo, line 9: Minor: This revert from version 0.26 to 0.24 seems related to CentOS 7. Are there issues using 0.26 on it?

ipc/tests/blocking_client.rs
ivoanjo: Ah, I think I see -- it happens implicitly when `rt` is dropped at the end of the method, is that it?

ivoanjo (approved):
:+1:  Thanks for hunting down these tiny things and fixing them :)

spawn_worker/Cargo.toml
pawelchcki: Yes there are, its a weird one too. 
Since it links into libc calls/syscalls not available on Centos 7 by default (memfd_create wasn't available in the original kernel shipped with RHEL 7, and in turn its libc) . 

Curiously I wasn't using `memfd_create` from nix, instead I used a different crate, which calls the syscall directly and doesn't link to libc (via rustix). Just including `nix 0.26` was enough to trigger this problem

I need to investigate `memfd` further, to make sure most Centos7 installs in the wild will actually have the updated kernel.

AFAIR native profiler also uses the memfd /cc @r1viollet @sanchda - wonder how this looks on your side

pawelchcki: Anyway - I've added the Centos7 build to CI, to safeguard against mishaps like this.

sanchda: Well,
```
static inline int memfd_create(const char *name, unsigned int flags) {
  return syscall(SYS_memfd_create, name, flags);
}
```
:)

the `memfd_create()` syscall was added in kernel 3.17, which is basically as far back as the native profiler supports anyway, so this hasn't really been a problem in the profiler.

PR #122, opened 2023-03-23
Add cross tool for windows

pawelchcki:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


spawn_worker/build.rs
```diff
@@ -11,13 +13,13 @@ fn main() {
         .warnings_into_errors(true)
         .emit_rerun_if_env_changed(true);
 
-    if !cfg!(target_os = "windows") {
+    if !env::var("TARGET").unwrap().contains("windows") {
```
paullegranddc, line 16: Why didn't `target_os` work anymore?

Cross.toml
```diff
@@ -1,2 +1,5 @@
 [target.x86_64-unknown-linux-gnu]
-image = "ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main-centos"
\ No newline at end of file
+image = "ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main-centos"
+
+[target.x86_64-pc-windows-msvc]
+image = "ghcr.io/datadog/libdatadog/cross/x86_64-pc-windows-msvc-cross:testing-not-for-shippable-artifacts"
```
gleocadie, line 5: why not for shippable artifacts ? what is the issue ?
should you consider `i686-pc-windows-msvc`. If the .NET profiler uses it at some point, this will be required (yeah we have customer with 32bits applications :) )

spawn_worker/build.rs
pawelchcki: cfg is evaluated during compilation. When cross compiling - you're compiling on linux

Cross.toml
pawelchcki: Sorry for late response.
Its named "testing-not-for-shippable-artifacts" to just safeguard against folks actually using this to build artifacts for customers. The name was meant to be self descriptive. But it seems its confusing. I'll change it and add a note in Cross.toml

PR #124, opened 2023-03-28
Serverless Trace Mini Agent

thedavl:
# What does this PR do?

This PR introduces the skeleton for a modular trace mini agent, as a standalone executable binary.

Description of new directories added:
- `serverless/`
   - contains the main() function, runs when the binary is executed & starts up the mini agent with serverless trace processor + serverless trace flusher
- `trace-mini-agent/`
   - contains the mini agent code, including spinning up the mini agent http server + trace flusher
- `trace-utils/`
  - contains trace deserialization & conversion utils, as well as agent payload construction + function to send traces to the trace intake backend.

Overall flow of logic:

- serverless bin is compiled and executed by a child process spawned by the tracer
- serverless trace agent main creates a serverless `trace flusher` and `trace processor`, using them to create a new mini-agent
- `mini_agent.start_mini_agent()` is called
- the trace flusher is started up, listens over a tokio (mpsc) channel for processed traces
- http server is started up, listening on `127.0.0.1:8126`
- requests to the server on the endpoint `/v0.4/traces` are handled by the trace processor
- trace processor de-serializes traces from the request body, converts them to protobuf traces/spans. These are then sent over the channel to the trace flusher
- trace flusher appends traces to a buffer vec. when the vec is over a hard-coded size flush_traces is called which creates an agent payload + serialized it into protobuf, sent to the trace intake. The buffer vec is then emptied.

# Motivation

This PR sets the framework for further features required in the Serverless mini-agent MVP. The serverless mini-agent is executed by a child process spawned within the tracers, to enable trace collection for Google Cloud Functions, and Azure Functions. 

Further work includes:
- Replace debugging print statements with a logger
- Make agent payload header + trace chunk data construction more programmatic
- Integration of trace obfuscation + trace normalization crates
- Trace sampling
- Smart trace buffering/flushing (instead of the current hard-coded buffer size to flush)
- Reading configuration properties from a datadog.yaml file included in the gcp function / azure function

# Additional Notes

# How to test the change?

- This mini agent is tested by building the `serverless-trace-mini-agent` binary, including it within a node google cloud function, modifying dd-trace-js with the code below to spin up the mini agent in a child process, and running the function locally.

dd-trace-js code:
```
// within tracer.js, the following function runs when tracer.init() is called
function startMiniAgent () {
  try {
    const { spawn } = require('child_process');
    const { join } = require('path');

    const rust_binary_path = join(process.cwd(), 'serverless_mini_agent_macos');

    const mini_agent_process = spawn(rust_binary_path);
    ...
```

Spans:
image


trace-mini-agent/src/trace_flusher.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+dyn_clone::clone_trait_object!(TraceFlusher);
+
+#[derive(Clone)]
+pub struct ServerlessTraceFlusher {}
+
+#[async_trait]
+impl TraceFlusher for ServerlessTraceFlusher {
+    async fn start_trace_flusher(&self, mut rx: Receiver<Vec<Vec<pb::Span>>>) {
+        let buffer_handle: Buffer = Arc::new(Mutex::new(Vec::with_capacity(BUFFER_FLUSH_SIZE)));
```
paullegranddc, line 33: Not sure why this needs to be inside an `Arc<Mutex<>>`

trace-mini-agent/src/trace_flusher.rs
```diff
@@ -0,0 +1,62 @@
...
+        // receive traces from http endpoint handlers and add them to the buffer. flush if
+        // the buffer gets to BUFFER_FLUSH_SIZE size.
+        while let Some(mut traces) = rx.recv().await {
+            let buffer_handle = buffer_handle.clone();
+            let mut buffer = buffer_handle.lock().unwrap();
+
+            buffer.append(&mut traces);
+            if buffer.len() >= BUFFER_FLUSH_SIZE {
+                println!("Attempting to flush buffer");
+                self.flush_traces(buffer.to_vec());
```
paullegranddc, line 44: You're calling to a function is doing synchronous IO in an async context. This is usually a pretty bad idea and might grind the whole tokio runtime to a halt, even with the multi threaded scheduler.

ivoanjo:
Hey :wave:

I'm kinda curious about the choice of a separate binary -- is that a choice for prototyping, or is that the actual intended outcome?

My question derives from a more generic "if this is a binary that is going to be very independent of the client libraries -- does this belong in libdatadog"?

Note that I'm definitely definitely (believe me! lol) not trying to push anyone out if you think this is the best place to add the code :sweat_smile: :sweat_smile: :sweat_smile:!

Maybe the plan is for some libraries to deploy the mini-agent in-process at some point? On the other hand, if we don't think that's the case, I do wonder if it would be better to separate out both projects.

trace-mini-agent/src/trace_flusher.rs
thedavl: fixed

trace-mini-agent/src/trace_flusher.rs
thedavl: removed this instance, though I'm planning to replace all println!s later once I implement a logger. They're all here for debugging, as I add more code to the mini-agent!

thedavl:
> Hey 👋
> 
> I'm kinda curious about the choice of a separate binary -- is that a choice for prototyping, or is that the actual intended outcome?
> 
> My question derives from a more generic "if this is a binary that is going to be very independent of the client libraries -- does this belong in libdatadog"?
> 
> Note that I'm definitely definitely (believe me! lol) not trying to push anyone out if you think this is the best place to add the code 😅 😅 😅!
> 
> Maybe the plan is for some libraries to deploy the mini-agent in-process at some point? On the other hand, if we don't think that's the case, I do wonder if it would be better to separate out both projects.

The separate binary is indeed the intended outcome, as our goal is to start the mini-agent by spawning a child process from within the tracers, and executing the compiled binary from the child process. There are no current plans to start the mini-agent in-process.

I think that since the only standalone-binary specific code present is what lives in `serverless/`, keeping everything here would let other future applications reuse components used by the serverless mini-agent.

trace-utils/src/trace_utils.rs
```diff
@@ -0,0 +1,274 @@
...
+
+pub fn get_tracer_tags_from_request_header(headers: &HeaderMap<HeaderValue>) -> TracerTags {
+    let mut ts = TracerTags {
+        lang: "",
+        lang_version: "",
+        lang_interpreter: "",
+        lang_vendor: "",
+        tracer_version: "",
+    };
+    if let Some(lang) = headers.get("datadog-meta-lang") {
```
thedavl, line 92: let me know if theres a cleaner way to do this!

trace-utils/src/trace_utils.rs
```diff
@@ -0,0 +1,274 @@
...
+
+pub struct TracerTags<'a> {
+    lang: &'a str,
+    lang_version: &'a str,
+    lang_interpreter: &'a str,
+    lang_vendor: &'a str,
+    tracer_version: &'a str,
+}
+
+pub fn get_tracer_tags_from_request_header(headers: &HeaderMap<HeaderValue>) -> TracerTags {
```
thedavl, line 84: let me know if theres a cleaner way to do this!

ivoanjo:
Thanks for the answer @thedavl! To be honest, I'm not entirely convinced that given that the current plan is for this to be separate, that it's worth keeping around "just in case".

Since you think it is, and given that it doesn't impact the rest of libdatadog, I'll trust you on it ;)

trace-utils/src/trace_utils.rs
paullegranddc: Maybe a macro ?
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=322c8d3f161d2dee6ef73ce9b746fde2

serverless/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "datadog-serverless-trace-mini-agent"
```
paullegranddc, line 2: Not sure I see the necessity of having this as a separate crate, instead of a bin inside the `trace-mini-agent` crate.

You can condition the import of tokio to a feature, that would be required in the bin definition

https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries

trace-utils/src/trace_utils.rs
```diff
@@ -0,0 +1,274 @@
...
+        Ok(res) => res,
+        Err(err) => {
+            anyhow::bail!("Error deserializing trace from request body: {}", err)
+        }
+    };
+
+    let mut pb_traces = Vec::<Vec<pb::Span>>::new();
+    for trace in traces {
+        let mut pb_spans = Vec::<pb::Span>::new();
+        for span in trace.iter() {
```
paullegranddc, line 45: I don't think `trace` is used anywhere else, so you can probably just do
```suggestion
        for span in trace() {
```

and remove the clone in `span.<>.clone()`

paullegranddc (approved):

serverless/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    let trace_processor = Box::new(trace_processor::ServerlessTraceProcessor {});
+
+    let mini_agent = Box::new(mini_agent::MiniAgent {
+        trace_processor,
+        trace_flusher,
+    });
+
+    match mini_agent.start_mini_agent() {
+        Ok(_) => (),
```
maxday, line 17: if you're not doing anything with the Ok case, maybe `if let Err` makes more sense? Something like this?
```
if let Err(e) = mini_agent.start_mini_agent() {
    println!("error when starting serverless mini agent: {}", e);
}
```

trace-mini-agent/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "datadog-trace-mini-agent"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+hyper = { version = "0.14", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
```
maxday, line 8: do we need full features for hyper and tokio?

trace-mini-agent/src/mini_agent.rs
```diff
@@ -0,0 +1,85 @@
...
+                match trace_processor.process_traces(req, tx).await {
+                    Ok(res) => Ok(res),
+                    Err(err) => Ok(Response::new(Body::from(format!(
+                        "Error processing traces: {}",
+                        err
+                    )))),
+                }
+            }
+            _ => {
+                let mut not_found = Response::default();
```
maxday, line 79: Response has a builder method which you can use to build a 404

trace-utils/src/trace_utils.rs
```diff
@@ -0,0 +1,274 @@
...
+use hyper::{body::Buf, Body, Client, Method, Request};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::{env, str};
+
+use prost::Message;
+
+use datadog_trace_protobuf::pb;
+
+const TRACE_INTAKE_URL: &str = "http://trace.agent.datadoghq.com/api/v0.2/traces";
```
maxday, line 15: I think latest version is v0.7 ->https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/version.go#L120
Could we use latest endpoint?

trace-utils/src/trace_utils.rs
```diff
@@ -0,0 +1,274 @@
...
+pub struct TracerTags<'a> {
+    lang: &'a str,
+    lang_version: &'a str,
+    lang_interpreter: &'a str,
+    lang_vendor: &'a str,
+    tracer_version: &'a str,
+}
+
+pub fn get_tracer_tags_from_request_header(headers: &HeaderMap<HeaderValue>) -> TracerTags {
+    let mut ts = TracerTags {
```
maxday, line 85: maybe we could implement Default for TracerTags?

serverless/Cargo.toml
thedavl: I think I'll leave it separate for now, for the sake of clarity especially as we add more serverless-specific code.

trace-utils/src/trace_utils.rs
```diff
@@ -0,0 +1,274 @@
...
+
+pub fn construct_trace_chunk(trace: &[pb::Span]) -> pb::TraceChunk {
+    pb::TraceChunk {
+        priority: 1,
+        origin: "".to_string(),
+        spans: trace.to_vec(),
+        tags: HashMap::new(),
+        dropped_trace: false,
+    }
+}
```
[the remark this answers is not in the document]
thedavl: fixed

trace-utils/src/trace_utils.rs
thedavl: fixed!

trace-mini-agent/Cargo.toml
thedavl: fixed with just the features we need.

trace-utils/src/trace_utils.rs
thedavl: done

serverless/src/main.rs
thedavl: done!

trace-utils/src/trace_utils.rs
thedavl: that version.go document outlines agent endpoint versions (for the tracers to use), rather than backend trace endpoints. the current trace_intake_url should be correct!

maxday (approved):
great work David!

PR #125, opened 2023-03-30
Add metrics package for dogstatsd metric parsing

IvanTopolcic:
# What does this PR do?

This PR adds a `Metrics` crate that introduces a metrics agent. This agent has the ability to listen on a UDP socket, receive {statsd/dogstatsd} payloads, and flush them to Datadog.

# Motivation

This PR is the first of a larger effort to get a micro dogstatsd server implementation in Rust, to be used by Serverless workloads. We want customers in Google Cloud Functions, Azure Functions, etc. to be able to send statsd and dogstatsd-based metrics to Datadog.

# Additional Notes

Currently, we don't support sampling, and **only** support statsd/dogstatsd distribution metrics.

# How to test the change?

The PR includes tests, however, you can spin up a tiny Rust binary:

```rust
use std::env;

use metrics::agent::{ConfigBuilder, MetricsAgent};

#[tokio::main]
async fn main() {
    let config = ConfigBuilder::new()
        .api_key(env::var("DD_API_KEY").unwrap())
        .build()
        .expect("Error parsing config");

    MetricsAgent::with_config(config).run().await;
}
```

And send UDP packets to the endpoint yourself:

```bash
echo -n "ivan.test.metric:1|d|#tag1:val1" | nc -4u -w0 localhost 8125
```

On an unrelated note, this is also my first PR in Rust! :-)


metrics/src/agent.rs
IvanTopolcic, line 1: There are `println!()`s littered everywhere in this code. In a future PR we can refactor this and introduce a logging library. Right now we're looking to get something merged in so we have a base to work off of.

metrics/src/metrics.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub values: Vec<f64>,
+}
+
+impl Points {
+    pub fn new(values: Vec<f64>, timestamp: u64) -> Self {
+        Self { timestamp, values }
+    }
+}
+
+#[derive(PartialEq, Debug, Serialize, Clone)]
```
IvanTopolcic, line 19: There's a lot of different metrics types, but we only care about distribution metrics for now.

IvanTopolcic, line 19: There's a lot of different metrics types, but we only care about distribution metrics for now.

metrics/src/metrics.rs
```diff
@@ -0,0 +1,212 @@
...
+pub enum MetricType {
+    Distribution,
+}
```
IvanTopolcic, line 22: There's a lot of different metrics types, but we only care about distribution metrics for now.

metrics/src/agent.rs
```diff
@@ -0,0 +1,266 @@
...
+            buf: Arc::new(Mutex::new(Vec::new())),
+        }
+    }
+
+    pub async fn run(&self) {
+        // Create a UDP socket and bind to the dogstatsd port
+        let socket = UdpSocket::bind(
+            format!("0.0.0.0:{}", DOGSTATSD_PORT)
+                .parse::<SocketAddr>()
+                .unwrap(),
```
maxday, line 41: let's try not to use unwrap here?

metrics/src/agent.rs
```diff
@@ -0,0 +1,266 @@
...
+        )
+        .await
+        .expect("Error binding to socket");
+
+        println!("Listening for dogstatsd packets on port {}", DOGSTATSD_PORT);
```
maxday, line 46: we could use a logger for that? as this will be printed in customer stdout

metrics/src/agent.rs
```diff
@@ -0,0 +1,266 @@
...
+                        .collect(),
+                    Err(e) => {
+                        println!("Error converting metric to str: {}", e);
+                        continue;
+                    }
+                };
+
+                let mut metrics_to_process = buf_producer.lock().await;
+                for metric in metrics {
+                    metrics_to_process.push(metric.clone());
```
maxday, line 95: not sure we need to clone it?

metrics/src/agent.rs
```diff
@@ -0,0 +1,266 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+use crate::metrics::Metric;
+use crate::payload::construct_distribution_payload;
+
+use futures::lock::Mutex;
```
maxday, line 7: maybe you don't need an external library and go with: https://doc.rust-lang.org/std/sync/struct.Mutex.html

metrics/src/agent.rs
```diff
@@ -0,0 +1,266 @@
...
+        println!("Listening for dogstatsd packets on port {}", DOGSTATSD_PORT);
+
+        // We use a shared vec here instead of a channel because we need to buffer the data at
+        // some point anyways. If we had used a channel here, we'd still need to add it to
+        // some type of buffer, so we'd need to implement a vector buffer anyways.
+        // Instead, we can save a step and just add it to the buffer directly.
+        let buf_producer = self.buf.clone();
+        let buf_consumer = self.buf.clone();
+
+        let https = HttpsConnectorBuilder::new()
```
maxday, line 55: not sure you need this, you can just use the builder

metrics/src/agent.rs
IvanTopolcic: That doesn't work in an async context unfortunately, it's what we had originally.

metrics/src/agent.rs
maxday: (I didn't see your comment before, feel free to ignore this)

IvanTopolcic: Ah yeah we should definitely introduce a logging library but I'd like to get _something_ in this week before I leave for vacation. I think it's ok for now to leave it as println, but if you feel strongly I can change it

metrics/src/agent.rs
IvanTopolcic: Correct!

metrics/src/agent.rs
```diff
@@ -0,0 +1,266 @@
...
+                let mut metrics_to_process = buf_producer.lock().await;
+                for metric in metrics {
+                    metrics_to_process.push(metric.clone());
+                }
+            }
+        });
+
+        // Process metrics we've parsed and flush them to Datadog
+        loop {
+            tokio::time::sleep(time::Duration::from_millis(3000)).await;
```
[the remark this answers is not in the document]
IvanTopolcic: Done!

metrics/src/agent.rs
IvanTopolcic: My understanding is that if we use a synchronous mutex, it'll block the entire thread, which would defeat the purpose of having the function be async in the first place.

IvanTopolcic: I switched it to use `tokio::sync::Mutex` as we're already using the `tokio` dependency, and removed the `futures` dependency.

metrics/src/agent.rs
```diff
@@ -0,0 +1,267 @@
...
+                Err(e) => {
+                    println!("Error sending request to Datadog: {}", e);
+                }
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Config {
```
DarcyRaynerDD, line 158: nit:Might be nice to move this to it's own file

metrics/src/agent.rs
```diff
@@ -0,0 +1,267 @@
...
+    api_key: String,
+    site: String,
+}
+
+pub struct ConfigBuilder {
```
DarcyRaynerDD, line 163: https://docs.rs/derive_builder/latest/derive_builder/ Is another alternative that doesn't add any runtime overhead.

metrics/src/agent.rs
```diff
@@ -0,0 +1,267 @@
...
+                        .split('\n')
+                        .filter_map(|metric_str| Metric::from_string(metric_str, unix_timestamp))
+                        .collect(),
+                    Err(e) => {
+                        println!("Error converting metric to str: {}", e);
+                        continue;
+                    }
+                };
+
+                let mut metrics_to_process = buf_producer.lock().await;
```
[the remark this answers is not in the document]
IvanTopolcic: Yeah! So I realized that we'd still need a buffer to store the metrics until we flush them to Datadog. If we used a channel, we'd still need to construct this buffer, and write to it. We'd essentially have the exact same problem (shared memory), so instead I opted to just write to the buffer directly instead of using a channel. This is because we need to flush every n seconds, and we need a place to store the Metrics until we flush. For example:

```rust
let (tx, rx) = mspc::channel();
let buf Vec<Metric> = vec![];

loop {
  let metric = get_metric_from_udp_server();
  tx.send(metric);
}

loop {
  rx.recv(metric);
  buf.push(metric);
  
  // what happens here? if we want to flush every 3s (buffered, we don't just want to 
  // flush single metrics), we need a new thread,
  // which means we need to put buf behind an Arc<Mutex<>>
  // which means we have the same problem as what we have now
}
```

I'm not sure if there's a better way to solve this problem, so I opted for the current approach. Do you have any suggestions?

metrics/src/agent.rs
IvanTopolcic: The default client doesn't implement https, the docs recommend using an HttpsConnector for https requests. I think it's okay to leave it the way it is now.

metrics/src/agent.rs
IvanTopolcic: Oh that's awesome, I didn't know we could do it automagically! I'll see if I can get it working.

metrics/src/agent.rs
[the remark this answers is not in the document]
DarcyRaynerDD: You could do something like this with channels:
 ```rust
let mut buffer = vec![]; // Reuse the same vec every iteration to avoid allocations
loop {
    buffer.clear();
    if let Some(iter) = rx.try_iter() {
        buffer.extend(iter);
    }
}
```
I'm not sure though. Do you think that's any neater?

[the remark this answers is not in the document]
Comment (DarcyRaynerDD) [reply metrics/src/agent.rs:92]: I think the main difference is this won't lock either thread. It's possible the producer thread will be blocked from writing while the lock is held by the consumer, which might be as long as it takes to flush the metrics to the backend.

File: metrics/src/payload.rs [hunk 12]
```diff
@@ -0,0 +1,117 @@
...
+#[derive(Serialize)]
+pub struct DistributionPayloadBuf {
+    series: Vec<Metric>,
+}
+
+// construct_datadog_payload takes a vector of metrics and constructs the payload
+// for the submit distribution metrics payload.
+// docs: https://docs.datadoghq.com/api/latest/metrics/?code-lang=curl#submit-distribution-points
+pub fn construct_distribution_payload(metrics: Vec<Metric>) -> Result<String, serde_json::Error> {
+    #[derive(Serialize)]
```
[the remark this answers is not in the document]
IvanTopolcic: Done!

metrics/src/agent.rs
IvanTopolcic: Done, works like magic :-)

metrics/src/agent.rs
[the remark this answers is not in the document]
IvanTopolcic: I've gone ahead and implemented this.

Looking back, I'm not entirely sure why I thought we'd need another thread poll the buf every 3s. This approach works, and is a lot simpler, much appreciated!

metrics/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "metrics"
```
ivoanjo, line 2: Suggestion to align with how (most) other crates are being named:

```suggestion
name = "datadog-metrics"
```

Cargo.toml
```diff
@@ -17,7 +17,7 @@ members = [
   "trace-utils",
   "spawn_worker",
   "tests/spawn_from_lib",
-  "serverless"
+  "metrics",
```
ivoanjo, line 20: I'm curious -- is this really intended (e.g. remove the "serverless" folder?) Doesn't that mean it will stop getting built and tested in CI?

ivoanjo (commented):
Hey! To help move this PR forward, I've done a review focusing on things outside the `metrics/` folder (as I commented in the channel).

Cargo.toml
thedavl: I believe we do want to keep the serverless folder. I changed it back

thedavl:
Hi all 👋, I'll be taking this PR over the finish line since Ivan is out.

metrics/src/metrics.rs
paullegranddc: Since the Point format will be different for distributions and Count/Gauges `(timestamp, Vec<point>)` `(timestamp, point)`
 
 I think the points should be a member of the enum, instead of having a `MetricType` enum
 
```rust
 #[derive(PartialEq, Debug, Serialize, Clone)]
pub enum MetricPoint {
    Distribution(Vec<f64>),

    // When you want to add other types
    Count(f64),
    Gauge(f64),
}
```

metrics/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+#[macro_use]
+extern crate derive_builder;
```
paullegranddc, line 5: This syntax is a [bit legacy](https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#macros), and you should probably just import the macro you're using in modules 

```rust
use derive_builder::Builder;

#[derive(Debug, Builder)]
#[builder(build_fn(validate = "Self::validate"))]
pub struct Config {}
```

paullegranddc (approved):
I have mostly nitpicks, but lgtm

PR #126, opened 2023-04-03
[Telemetry worker] Upgrade to api v2 spec

paullegranddc:
# What does this PR do?

## Telemetry data schema changes

* Update the telemetry version tag to "v2" 🥳 
* Remove the type field in dependency since it's not used in the official schema
* Add skip serialisation fi fields are None in the host
* Remove lib_language and lib_version from generate metrics
* AppClientConfigurationChange, MessageBatch and AppExtdendedHearbeat type messages
* Fix configuration object schema

## Telemetry worker changes

* Instead of having a flush event per collection, have a single one for all
* Use the new MessageBatch type message to batch all data flush payloads (except logs and metrics)
* Keep the telemetry data already sent in storage, to be able to send it again for the extended hearbeat
* Because we keep already submitted data, keep data in a bounded collection to prevent memory leaks
* Resend data if the previous submission upstream failed

+ some refactoring here and there

# Motivation

Changes are specced, and explained in the following document https://docs.google.com/document/d/1Fai2gZlkvfa_WQs_yJsmi3nUwiOsMtNu2LFBnbTHJRA/edit

# Additional Notes

# How to test the change?

In progress


pawelchcki (approved):

LICENSE-3rdparty.yml
pawelchcki, line 1: Hmm interesting why this has so many changes :thinking: /cc @morrisonlevi - possible something changed in the script we use to generate this

paullegranddc: I think the changes are because the yaml file went from inline strings to multi line strings

```yaml
a: "quoted\nstring"
b: |
  quoted
  string
```

PR #128, opened 2023-04-04
Remove dependency on messagepack

paullegranddc:
# What does this PR do?

* Remove dependency on rpm-serde
* Groom dependencies and only enable necessary features

# Motivation

# Additional Notes

# How to test the change?


ivoanjo (approved):
:+1: Thanks for the nice cleanup, LGTM!

PR #130, opened 2023-04-04
Mini Agent: add trace normalization, root span calculation, logging

thedavl:
# What does this PR do?

- Introduces a logger (env_logger)
- Integrates the trace-normalization lib
- Use data from a tracer's request headers & trace root span tags to populate tracer payload fields
- Compute + tag top level spans
- Unit tests

# Motivation

Add basic features into the mini-agent for our serverless mvp

# Additional Notes

Future PRs will include:
- Change flushing logic to poll for traces to flush every 3 seconds, instead of based on buffer size
- Send JSON response to endpoint requests
- Trace sampling

# How to test the change?

Added unit tests.
Tested manually with a modified version of dd-trace-js. See prev PR: https://github.com/DataDog/libdatadog/pull/124


trace-mini-agent/src/mini_agent.rs
```diff
@@ -51,11 +51,24 @@ impl MiniAgent {
         });
 
         let addr = SocketAddr::from(([127, 0, 0, 1], MINI_AGENT_PORT as u16));
-        let server = Server::bind(&addr).serve(make_svc);
+        let server_builder = match Server::try_bind(&addr) {
```
maxday, line 54: you could use `map_err` to simplify this code block

trace-utils/src/trace_utils.rs
```diff
@@ -183,23 +206,107 @@ pub async fn send(data: Vec<u8>) -> anyhow::Result<()> {
...
 
+    for span in trace {
+        if parent_id_to_child_map.contains_key(&span.span_id) {
+            parent_id_to_child_map.remove(&span.span_id);
+        }
+    }
+
+    // if the trace is valid, parent_id_to_child_map should just have 1 entry at this point.
+    if parent_id_to_child_map.len() != 1 {
+        println!(
```
maxday, line 241: you could use your new logger here

trace-utils/src/trace_utils.rs
```diff
@@ -183,23 +206,107 @@ pub async fn send(data: Vec<u8>) -> anyhow::Result<()> {
...
+            &trace[0].trace_id,
+        );
+    }
+
+    // pick a span without a parent
+    let span_tuple = match parent_id_to_child_map.values().copied().next() {
+        Some(res) => res,
+        None => {
+            // just return the index of the last span in the trace.
+            println!("Returning index of last span in trace as root span index.");
```
maxday, line 252: same here

trace-utils/src/trace_utils.rs
thedavl: totally missed these. Thanks!

thedavl: fixed

trace-utils/src/trace_utils.rs
thedavl: fixed

trace-mini-agent/src/mini_agent.rs
thedavl: simplified this code block to propagate error back to whatever called start_mini_agent

pawelchcki (approved):
:heart: Looks good to me!

trace-utils/src/trace_utils.rs
```diff
@@ -183,23 +206,107 @@ pub async fn send(data: Vec<u8>) -> anyhow::Result<()> {
...
-        Err(e) => println!("Failed to send traces: {}", e),
+        Err(e) => error!("Failed to send traces: {}", e),
     }
     Ok(())
 }
 
-#[cfg(test)]
-mod tests {
-    use std::collections::HashMap;
+pub fn get_root_span_index(trace: &Vec<pb::Span>) -> anyhow::Result<usize> {
```
maxday, line 216: I think you can just use Result from std here

trace-utils/src/trace_utils.rs
```diff
@@ -183,23 +206,107 @@ pub async fn send(data: Vec<u8>) -> anyhow::Result<()> {
...
+            }
+            None => {
+                // span has no parent in chunk
+                set_top_level_span(span, true)
+            }
+        }
+    }
+}
+
+fn set_top_level_span(span: &mut pb::Span, is_top_level: bool) {
```
maxday, line 291: What do you think about moving this method to the Span implementation?

maxday (approved):

trace-utils/src/trace_utils.rs
morrisonlevi: Lots of code in libdatadog uses `anyhow::Result` specifically, either for its convenience helpers like `anyhow::bail!` (like this code does), or for its `.context` helpers for chaining errors together. I think it's fine to keep it this way for now, although last time this sort of thing came up there was another error library mentioned that may be better. Maybe we should investigate that a bit more, but for now it's fine, I think.

PR #133, opened 2023-04-10
Mini Agent: Modify flushing logic, Serverless root span tags

thedavl:
# What does this PR do?

- Changes trace flushing logic to flush traces every 3 seconds
- Add build script to trace-protobuf, re-generate protobuf files
   - Added `#[serde(default)]` annotations to generated protobuf struct fields to allow for direct deserialization of request body spans (without the need of an intermediate struct with Option<> fields). Annotations are programmatically added in the build script.
   - protobuf file regeneration is only done when the feature `generate-protobuf` is on
- Set the serverless mini agent logger to log to stdout instead of stderr (which is default for env_logger)
- Set serverless specific tags to a trace's root span
   - Adds `_dd.origin=gcp_function` and `functionname` tags into the root span's meta map
   - Sets the root span type to "serverless"
-  Send json response back to tracer
- Implement an `/info` endpoint, importantly adding `"client_drop_p0s": true` so that the tracer drops sampled out spans from being sent to the mini-agent

# Motivation

Serverless mini agent MVP

# Additional Notes

Future PRs will include:
- Trace flush retries
- Expose a metrics endpoint for client computed stats
- Process client computed stats payloads. Flush to backend.

# How to test the change?


trace-protobuf/Cargo.toml
```diff
@@ -6,3 +6,11 @@ edition = "2021"
 
 [dependencies]
 prost = "0.11.6"
+serde = { version = "1.0.145", features = ["derive"] }
+
+[build-dependencies]
+prost-build = { version = "=0.11.3", optional = true  }
```
thedavl, line 12: explicitly use an older version of prost-build, as versions > 0.11.3 introduce adds `#[allow(clippy::derive_partial_eq_without_eq)]` to generated protobuf structs. This lint doesn't exist for rust versions < 1.63.0 and throws an `unknown lint` error in our CI

trace-protobuf/build.rs
```diff
@@ -0,0 +1,76 @@
+use std::io::Result;
```
thedavl, line 1: this build script was added to re-compile `.proto` files we use, adding `#[serde(default)]` annotations to certain fields.

this allows us to directly de-serialize spans into these proto generated structs, as opposed to first de-serializing to a struct with Option<> fields, then converting to the proto generated Span struct (using unwrap_or_default on Option<> fields)

trace-utils/src/trace_utils.rs
```diff
@@ -31,63 +28,21 @@ macro_rules! parse_string_header {
     }
 }
 
-#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
```
thedavl, line 34: no longer needed, as the pb::Span struct now has serde default annotations on certain fields allowing for direct deserialization of requests into them

trace-utils/src/trace_utils.rs
```diff
@@ -31,63 +28,21 @@ macro_rules! parse_string_header {
...
-pub struct Span {
```
thedavl, line 35: no longer needed, as the pb::Span struct now has serde default annotations on certain fields allowing for direct deserialization of requests into them

trace-utils/src/trace_utils.rs
```diff
@@ -31,63 +28,21 @@ macro_rules! parse_string_header {
...
 
-    let traces: Vec<Vec<Span>> = match rmp_serde::from_read(buffer.reader()) {
+    let traces: Vec<Vec<pb::Span>> = match rmp_serde::from_read(buffer.reader()) {
         Ok(res) => res,
         Err(err) => {
             anyhow::bail!("Error deserializing trace from request body: {}", err)
         }
     };
 
-    let mut pb_traces = Vec::<Vec<pb::Span>>::new();
```
thedavl, line 59: we save all this conversion!

trace-mini-agent/src/trace_flusher.rs
```diff
@@ -25,29 +24,45 @@ pub struct ServerlessTraceFlusher {}
...
+
+        tokio::spawn(async move {
+            while let Some(tracer_payload) = rx.recv().await {
+                let mut buffer = buffer_producer.lock().await;
+                buffer.push(tracer_payload);
             }
+        });
+
+        loop {
+            tokio::time::sleep(time::Duration::from_millis(3000)).await;
```
thedavl, line 40: flush every 3 seconds

trace-mini-agent/src/trace_flusher.rs
```diff
@@ -25,29 +24,45 @@ pub struct ServerlessTraceFlusher {}
...
+
+            let mut buffer = buffer_consumer.lock().await;
+            if buffer.is_empty() {
+                continue;
```
maxday, line 44: you could remove the continue by reversing the condition and move flush + clear in the block?

trace-mini-agent/src/trace_flusher.rs
```diff
@@ -25,29 +24,45 @@ pub struct ServerlessTraceFlusher {}
...
+            self.flush_traces(buffer.to_vec()).await;
+            buffer.clear();
         }
     }
 
     async fn flush_traces(&self, traces: Vec<pb::TracerPayload>) {
         if traces.is_empty() {
             return;
         }
+        info!("Flushing traces");
```
maxday, line 55: it could be useful to add the number of traces?
+ add a debug log with the payload content?

trace-mini-agent/src/trace_processor.rs
```diff
@@ -111,13 +123,18 @@ impl TraceProcessor for ServerlessTraceProcessor {
...
-                .status(200)
-                .body(Body::from("Successfully buffered traces to be flushed.")),
-            Err(e) => Response::builder().status(500).body(Body::from(format!(
-                "Error sending traces to the trace flusher. {}",
-                e
-            ))),
+            Ok(_) => {
+                info!("Successfully buffered traces to be flushed.");
+                let body =
+                    json!({ "message": "Successfully buffered traces to be flushed." }).to_string();
```
maxday, line 129: do we need a body? who is reading it?

trace-mini-agent/src/trace_processor.rs
```diff
@@ -111,13 +123,18 @@ impl TraceProcessor for ServerlessTraceProcessor {
...
+                Response::builder().status(200).body(Body::from(body))
+            }
+            Err(e) => {
+                let error_message = format!("Error sending traces to the trace flusher: {}", e);
```
maxday, line 133: same here

trace-protobuf/Cargo.toml
maxday: it seems a bit dangerous to fix a older version for a new feature. are there any workarounds?

trace-protobuf/src/lib.rs
```diff
@@ -3,8 +3,5 @@
 // developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present
 // Datadog, Inc.
 
-#![deny(clippy::all)]
-
-pub mod pb {
-    include!("./pb/pb.rs");
-}
+#[rustfmt::skip]
```
maxday, line 6: why do we need this?

trace-mini-agent/src/trace_flusher.rs
thedavl: added

trace-protobuf/Cargo.toml
thedavl: A workaround would be to load and modify the autogenerated file in the build.rs rust script itself, removing the auto-added annotation that is causing problems.

I thought it would be cleaner not to do file processing in the code, but to just pin this earlier version. I couldn't figure out any other workarounds, what do you think the best approach would be?

trace-mini-agent/src/trace_flusher.rs
thedavl: done

trace-mini-agent/src/trace_processor.rs
thedavl: the JS tracer will log a debug log with a "unable to parse JSON" error if the body is not a JSON string. It doesn't effect any functionality to not include a body however. I thought this might be confusing to users with debug logs turned on, so I added bodies

trace-protobuf/Cargo.toml
thedavl: I've instead bumped the rust version we use in the CI tests `1.60.0` -> `1.63.0`

This lets us unpin the prost-build version, but if there are any objections I'll revert it.

.github/workflows/lint.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       matrix:
         # Ignore nightly for now, it fails too often
         #rust_version: ["nightly", "1.60.0", "stable"]
-        rust_version: ["1.60.0", "stable"]
+        rust_version: ["1.63.0", "stable"]
```
thedavl, line 24: the prost-build dependency will autogenerate spans annotated with `#[allow(clippy::derive_partial_eq_without_eq)]`

this lint rule doesn't exist in rust versions <1.63.0, so the workaround was to pin prost-build to an older version.

I think just bumping our rust_version we use is a better solution, but let me know if there are objections to this or implication's im missing

morrisonlevi: I opened a PR to bump it to 1.64.0 and all the tests at that time passed. Additionally, the PHP profiler has been building and shipping with Rust 1.64.0 for a while now. I recommend moving to 1.64.0.

trace-mini-agent/src/trace_processor.rs
```diff
@@ -111,13 +123,18 @@ impl TraceProcessor for ServerlessTraceProcessor {
...
+                error!("{}", error_message);
```
morrisonlevi, line 134:
```suggestion
                error!("Error sending traces to the trace flusher: {e}");
```

trace-utils/src/trace_utils.rs
```diff
@@ -31,63 +33,21 @@ macro_rules! parse_string_header {
...
-}
-
 pub async fn get_traces_from_request_body(body: Body) -> anyhow::Result<Vec<Vec<pb::Span>>> {
     let buffer = hyper::body::aggregate(body).await.unwrap();
 
-    let traces: Vec<Vec<Span>> = match rmp_serde::from_read(buffer.reader()) {
+    let traces: Vec<Vec<pb::Span>> = match rmp_serde::from_read(buffer.reader()) {
         Ok(res) => res,
         Err(err) => {
             anyhow::bail!("Error deserializing trace from request body: {}", err)
```
morrisonlevi, line 42:
```suggestion
            anyhow::bail!("Error deserializing trace from request body: {err}")
```

You will probably be required to fix these if you upgrade Rust because I think clippy started complaining about that in 1.64.0.

trace-utils/src/trace_utils.rs
```diff
@@ -149,7 +109,7 @@ pub fn construct_agent_payload(tracer_payloads: Vec<pb::TracerPayload>) -> pb::A
 
 pub fn construct_trace_chunk(trace: Vec<pb::Span>) -> pb::TraceChunk {
     pb::TraceChunk {
-        priority: 1,
+        priority: i8::MIN as i32, // SamplerPriority::None
```
morrisonlevi, line 112: I assume based on the comment that `SamplerPriority::None` is `i8::MIN` -- is there some reason we can't just use `SamplerPriority::None`?

trace-utils/src/trace_utils.rs
```diff
@@ -205,12 +164,16 @@ pub async fn send(data: Vec<u8>) -> anyhow::Result<()> {
...
+        Ok(response) => {
+            if response.status() != StatusCode::ACCEPTED {
+                let body_bytes = hyper::body::to_bytes(response.into_body()).await?;
+                let response_body = String::from_utf8(body_bytes.to_vec()).unwrap_or_default();
+                anyhow::bail!("Server did not accept traces: {}", response_body);
+            }
+            Ok(())
         }
-        Err(e) => error!("Failed to send traces: {}", e),
+        Err(e) => anyhow::bail!("Failed to send traces: {}", e),
```
morrisonlevi, line 175:
```suggestion
        Err(e) => anyhow::bail!("Failed to send traces: {e}"),
```

trace-protobuf/Cargo.toml
thedavl: [this PR](https://github.com/DataDog/libdatadog/pull/77) will bump the rust version, I'll revert the changes here and wait that PR to merge.

.github/workflows/lint.yml
thedavl: sounds good, I'll remove the changes here and wait for your PR to merge

trace-utils/src/trace_utils.rs
thedavl: switched over to using SamplerPriority enum from the normalizer crate

serverless/src/main.rs
```diff
@@ -24,6 +24,6 @@ pub fn main() {
     });
 
     if let Err(e) = mini_agent.start_mini_agent() {
-        error!("error when starting serverless mini agent: {}", e);
+        error!("Error when starting serverless trace mini agent: {}", e);
```
morrisonlevi, line 27: Another interpolation clean up:

```suggestion
        error!("Error when starting serverless trace mini agent: {e}");
```

trace-mini-agent/src/mini_agent.rs
```diff
@@ -55,11 +56,11 @@ impl MiniAgent {
...
-        info!("Mini Agent listening on port {}", MINI_AGENT_PORT);
+        info!("Mini Agent started: listening on port {MINI_AGENT_PORT}");
 
         // start hyper http server
         if let Err(e) = server.await {
-            error!("Server error: {}", e);
+            error!("Server error: {e}");
         }
 
         Ok(())
```
morrisonlevi, line 66: If the server fails, we still return `Ok(())`?  Seems wrong although I don't know why it would fail at this spot.

trace-mini-agent/src/trace_flusher.rs
morrisonlevi: Any reason you don't do `time::Duration::from_secs(3)` instead?

trace-mini-agent/src/trace_flusher.rs
```diff
@@ -42,12 +51,25 @@ impl TraceFlusher for ServerlessTraceFlusher {
         if traces.is_empty() {
             return;
         }
+        info!("Flushing {} traces", traces.len());
+
+        debug!("Traces to be flushed: {:?}", traces);
```
morrisonlevi, line 56: Another interpolation:

```suggestion
        debug!("Traces to be flushed: {traces:?}");
```

morrisonlevi (approved):
Just a few very small things, going to go ahead with the Approve button.

trace-mini-agent/src/mini_agent.rs
thedavl: you're right, I changed this to return the server error

PR #139, opened 2023-04-24
Mini Agent: Trace stats

thedavl:
# What does this PR do?

Rebase onto main after [Mini Agent: Modify flushing logic, Serverless root span tags](https://github.com/DataDog/libdatadog/pull/133) is merged.

Exposes a `/v0.6/stats` endpoint, receiving trace stats from the tracer for when `DD_TRACE_STATS_COMPUTATION_ENABLED=true`. Buffers the received trace stats, flushing every 3 seconds to the trace edge metrics intake.

# Motivation

We want google cloud function users to be able to sample traces at the tracer level, so a trace stats endpoint is needed to receive trace stats from the tracers.

# Additional Notes

# How to test the change?

Manually tested with modified versions of dd-trace-js & dd-trace-py, unit test.


trace-mini-agent/src/http_utils.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/// Logs a message (at the error level) and returns the same message in the body of response with status code 500.
+/// Response body:
+/// {
+///     "message": message
+/// }
+pub fn log_and_return_http_error_response(message: &str) -> http::Result<Response<Body>> {
+    error!("{message}");
+    let body = json!({ "message": message }).to_string();
```
maxday, line 28: nit: this looks a lot like the previous function, maybe create a new function with message and status code and reuse it in both functions?

trace-mini-agent/src/mini_agent.rs
```diff
@@ -69,29 +98,31 @@ impl MiniAgent {
...
+        trace_tx: Sender<pb::TracerPayload>,
+        stats_processor: Arc<dyn stats_processor::StatsProcessor + Send + Sync>,
+        stats_tx: Sender<pb::ClientStatsPayload>,
     ) -> http::Result<Response<Body>> {
         match (req.method(), req.uri().path()) {
             (&Method::PUT, TRACE_ENDPOINT_PATH) => {
-                match trace_processor.process_traces(req, tx).await {
+                match trace_processor.process_traces(req, trace_tx).await {
+                    Ok(res) => Ok(res),
+                    Err(err) => log_and_return_http_error_response(&format!(
```
maxday, line 109: maybe we can pass the str + err and format it in log_and_return_http_error_response?

trace-mini-agent/src/stats_flusher.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        tokio::spawn(async move {
+            while let Some(stats_payload) = rx.recv().await {
+                let mut buffer = buffer_producer.lock().await;
+                buffer.push(stats_payload);
+            }
+        });
+
+        loop {
+            tokio::time::sleep(time::Duration::from_millis(3000)).await;
```
maxday, line 40: should we move that to a constant on top of the file? that might be easier to change if we want

trace-utils/src/trace_utils.rs
```diff
@@ -147,7 +147,7 @@ pub fn serialize_agent_payload(payload: pb::AgentPayload) -> anyhow::Result<Vec<
 pub async fn send(data: Vec<u8>) -> anyhow::Result<()> {
     let api_key = match env::var("DD_API_KEY") {
         Ok(key) => key,
-        Err(_) => anyhow::bail!("oopsy, no DD_API_KEY was provided"),
+        Err(_) => anyhow::bail!("Sending traces failed. Missing DD_API_KEY"),
```
maxday, line 150: :)

trace-mini-agent/src/http_utils.rs
thedavl: combined the two functions

trace-mini-agent/src/stats_flusher.rs
thedavl: good idea, done

trace-protobuf/src/pb/stats.proto
```diff
@@ -0,0 +1,70 @@
+syntax = "proto3";
```
pawelchcki, line 1: Might be missing a copyright disclaimer - is protobof file - code (?)

trace-protobuf/src/pb/stats.proto
```diff
@@ -0,0 +1,70 @@
...
+
+import "gogo.proto";
+
+package pb;
+
+// protoc --gogofaster_out=. -I $GOPATH/src -I . stats.proto
+
+// StatsPayload is the payload used to send stats from the agent to the backend.
+message StatsPayload {
```
pawelchcki, line 10: One thing we've noticed in the past in [system-test](https://github.com/DataDog/system-tests) - is that these files are updated in the agent from time to time - would it make sense to have CI check that verifies if they are still in sync ?

trace-utils/src/stats_utils.rs
```diff
@@ -0,0 +1,82 @@
...
+use hyper_rustls::HttpsConnectorBuilder;
+use log::debug;
+use std::{env, io::Write};
+
+use datadog_trace_protobuf::pb;
+
+const STATS_INTAKE_URL: &str = "https://trace.agent.datadoghq.com/api/v0.2/stats";
+
+pub async fn get_stats_from_request_body(body: Body) -> anyhow::Result<pb::ClientStatsPayload> {
+    let buffer = hyper::body::aggregate(body).await?;
```
pawelchcki, line 15: We're not checking or constraining the size of the body here. 
A broken, oversized, or malicious request will cause an out of memory error.

trace-utils/src/stats_utils.rs
```diff
@@ -0,0 +1,82 @@
...
+    let mut encoder = GzEncoder::new(Vec::new(), Compression::fast());
+    encoder.write_all(&msgpack)?;
+    match encoder.finish() {
+        Ok(res) => Ok(res),
+        Err(e) => anyhow::bail!("Error serializing stats payload: {e}"),
+    }
+}
+
+pub async fn send_stats_payload(data: Vec<u8>) -> anyhow::Result<()> {
+    let api_key = match env::var("DD_API_KEY") {
```
pawelchcki, line 52: In telemetry crate - we already have some code to parse env variables - including the ability to set DD_SITE 
 - etc https://github.com/DataDog/libdatadog/blob/main/ddtelemetry/src/config.rs
 
Might be worth extracting this code into some common crate and attempting a reuse.

pawelchcki: Additionally I would recommend not using global variable like `env::var` inside the code here. Its generally better to create a `Config` object - that is constructed from env. To make the code more testable

trace-utils/src/trace_utils.rs
```diff
@@ -265,10 +265,16 @@ pub fn set_serverless_root_span_tags(span: &mut pb::Span) {
...
         .insert("_dd.origin".to_string(), "gcp_function".to_string());
-    span.meta.insert(
-        "functionname".to_string(),
-        env::var("K_SERVICE").unwrap_or_default(),
-    );
+
+    // Google cloud functions automatically sets either K_SERVICE or FUNCTION_NAME
+    // env vars to denote the cloud function name.
+    // K_SERVICE is set on newer runtimes, while FUNCTION_NAME is set on older deprecated runtimes.
+    if let Ok(function_name) = env::var("K_SERVICE") {
```
pawelchcki, line 272: Same as comment above, its better to rely on a config structure, and more generic way to handle these special cases. Than to add them in the middle of the implemetnation. e.g. If I wanted to use this code in another scenario (outside of GCP) - where someone decided to set the K_SERVICE env - I would get unexpeted span metadata.

pawelchcki (changes requested):
One important comment about constraining the received payload size - to avoid out of memory errors.
And few style/code composition suggestions.

Otherwise looks good to me !

trace-utils/src/stats_utils.rs
thedavl: Added a check to make sure the `content-length` header is less than a config set value (10MB). Hyper will stop reading request bodies once `content-length` is reached

trace-utils/src/stats_utils.rs
thedavl: moved api_key to a config struct

trace-utils/src/trace_utils.rs
thedavl: moved function_name to a config struct

trace-protobuf/src/pb/stats.proto
thedavl: i don't believe the proto files need copyright disclaimers (the proto files in the regular agent don't have them)

trace-protobuf/src/pb/stats.proto
thedavl: added a new CI workflow to diff our `span.proto`, `agent_payload.proto`, and `stats.proto` files against those present in datadog-agent. Fixed a couple missing fields that were added recently too.

trace-mini-agent/src/stats_flusher.rs
```diff
@@ -0,0 +1,87 @@
...
+use log::{debug, error, info};
+use std::{sync::Arc, time};
+use tokio::sync::{mpsc::Receiver, Mutex};
+
+use datadog_trace_protobuf::pb;
+use datadog_trace_utils::stats_utils;
+
+use crate::config::Config;
+
+const STATS_FLUSH_INTERVAL_SEC: u64 = 3;
```
pawelchcki, line 14: I missed it in previous review. Small nit and suggestion for the future:

this could also be configurable / at least stored in Config struct

trace-mini-agent/src/trace_flusher.rs
```diff
@@ -9,21 +9,25 @@ use tokio::sync::{mpsc::Receiver, Mutex};
 use datadog_trace_protobuf::pb;
 use datadog_trace_utils::trace_utils;
 
+use crate::config::Config;
+
+const TRACE_FLUSH_INTERVAL_SEC: u64 = 3;
```
pawelchcki, line 14: nit: same comment as above - about configurability

pawelchcki (approved):

pawelchcki:
LGTM :bow:

trace-mini-agent/src/stats_flusher.rs
thedavl: done

trace-mini-agent/src/trace_flusher.rs
thedavl: done

trace-mini-agent/src/config.rs
```diff
@@ -0,0 +1,38 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+use std::env;
+
+#[derive(Debug, Clone)]
+pub struct Config {
+    pub api_key: String,
+    pub gcp_function_name: Option<String>,
```
maxday, line 9: I don't think you need an optional here?

thedavl: I kept it Option<> for when the mini agent + config is used for non-serverless usecases where there is no gcp_function_name

trace-mini-agent/src/trace_processor.rs
```diff
@@ -95,7 +111,10 @@ impl TraceProcessor for ServerlessTraceProcessor {
                 trace_utils::compute_top_level_span(&mut chunk.spans);
             }
 
-            trace_utils::set_serverless_root_span_tags(&mut chunk.spans[root_span_index]);
+            trace_utils::set_serverless_root_span_tags(
+                &mut chunk.spans[root_span_index],
+                config.gcp_function_name.clone(),
```
maxday, line 116: what about passing the ref here to avoid a clone (as you're doing a toString in set_serverless_root_span_tags)?

thedavl: i think we should keep it as is, because `set_serverless_root_span_tags` doesn't re-clone `gcp_function_name`. 

Making this parameter a &str would require using.to_string() in `set_serverless_root_span_tags` anyways as the span's Meta map has String values, not &str.

maxday: oh yes, I got confused with other to_string() but they're actually not on the gcp_function_name param

maxday (approved):

PR #142, opened 2023-04-25
Mini Agent: Verify GCP Env

thedavl:
# What does this PR do?

Add an `env_verifier` field to the trace mini agent. The mini agent calls `env_verifier.verify_environment()` in a tokio async task during initialization. This function makes an HTTP call to `http://metadata.google.internal`, the Google metadata server.

This metadata server is only accessible within a Google Cloud VM, and will include the following header in the response, indicating a google cloud function environment:
`Server: Metadata Server for Serverless`

If the metadata server is not accessible or the correct header is not returned, exit the mini agent process.

# Motivation

In the node/python tracer, we only check for the presence of K_SERVICE or FUNCTION_NAME environment variables (which google cloud functions set) before spinning up the mini-agent.

To prevent the use of the mini agent in server-ful environments (thereby bypassing infrastructure/host billing), we want to shut down the mini agent if it's running in a non gcp function env.

# Additional Notes

# How to test the change?

### DIFF
--- a/trace-mini-agent/src/env_verifier.rs
+++ b/trace-mini-agent/src/env_verifier.rs
@@ -89,14 +89,14 @@ impl EnvVerifier for ServerlessEnvVerifier {
 /// This function extracts just the region (us-east1) from this GCP region string.
 /// If the string does not have 4 parts (separated by "/") or extraction fails, return "unknown"
 fn get_region_from_gcp_region_string(str: String) -> String {
-    let split_str = str.split("/").collect::<Vec<&str>>();
+    let split_str = str.split('/').collect::<Vec<&str>>();
     if split_str.len() != 4 {
         return "unknown".to_string();
     }
     match split_str.last() {
-        Some(res) => return res.to_string(),
-        None => return "unknown".to_string(),
-    };
+        Some(res) => res.to_string(),
+        None => "unknown".to_string(),
+    }
 }
 
 /// GoogleMetadataClient trait is used so we can mock a google metadata server response in unit tests
--- a/trace-mini-agent/src/env_verifier.rs
+++ b/trace-mini-agent/src/env_verifier.rs
@@ -30,7 +30,8 @@ pub struct GCPProject {
 
 #[async_trait]
 pub trait EnvVerifier {
-    /// verifies the mini agent is running in the intended environment. if not, exit the process.
+    /// Verifies the mini agent is running in the intended environment. if not, exit the process.
+    /// Returns MiniAgentMetadata, a struct of metadata collected from the environment.
     async fn verify_environment(&self) -> trace_utils::MiniAgentMetadata;
 }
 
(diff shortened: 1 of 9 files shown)


trace-mini-agent/src/env_verifier.rs
```diff
@@ -0,0 +1,213 @@
...
+use std::process;
+
+use datadog_trace_utils::trace_utils;
+
+const GCP_METADATA_URL: &str = "http://metadata.google.internal/computeMetadata/v1/?recursive=true";
+
+#[derive(Default, Debug, Deserialize, Serialize)]
+pub struct GCPMetadata {
+    pub instance: GCPInstance,
+    pub project: GCPProject,
```
maxday, line 16: I think there is also the location/region field in this endpoint (as least this is the case in CloudRun) and we might want this metadata as well for tagging purposes

trace-mini-agent/src/env_verifier.rs
```diff
@@ -0,0 +1,213 @@
...
+#[derive(Default, Debug, Deserialize, Serialize)]
+pub struct GCPInstance {
+    pub region: Option<String>,
+}
+
+#[derive(Clone, Default, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GCPProject {
+    pub numeric_project_id: Option<u64>,
+    pub project_id: Option<String>,
```
maxday, line 28: why do we need those to be optional?

trace-mini-agent/src/env_verifier.rs
```diff
@@ -0,0 +1,213 @@
...
+            res.unwrap_err().to_string(),
+            "In Google Cloud, but not in a function environment."
+        );
+    }
+
+    #[tokio::test]
+    async fn test_verify_env_true_if_cloud_function_env() {
+        struct MockGoogleMetadataClient {}
+        #[async_trait]
+        impl GoogleMetadataClient for MockGoogleMetadataClient {
```
maxday, line 201: 💯

trace-mini-agent/src/env_verifier.rs
thedavl: region field is within `GCPInstance`

trace-mini-agent/src/mini_agent.rs
```diff
@@ -27,11 +29,27 @@ pub struct MiniAgent {
     pub trace_flusher: Arc<dyn trace_flusher::TraceFlusher + Send + Sync>,
     pub stats_processor: Arc<dyn stats_processor::StatsProcessor + Send + Sync>,
     pub stats_flusher: Arc<dyn stats_flusher::StatsFlusher + Send + Sync>,
+    pub env_verifier: Arc<dyn env_verifier::EnvVerifier + Send + Sync>,
 }
 
 impl MiniAgent {
     #[tokio::main]
     pub async fn start_mini_agent(&self) -> Result<(), Box<dyn std::error::Error>> {
+        // verify we are in a google cloud funtion environment. if not, shut down the mini agent.
```
maxday, line 38: my understanding is that we're not blocking on this internal metadata api calls which could be an issue if we're receiving telemetry meanwhile, they won't be tagged properly. WDYT about making this a blocking call (with a short timeout value + default value if timeout exceeded or errored?

maxday (approved):
LGTM!

PR #148, opened 2023-05-02
Fix env_verifier + protobuf gitlab build issues

thedavl:
# What does this PR do?

Fixed env_verifier unit test + protobuf gitlab build issues (failing on windows/centos static analysis)

# Motivation

build failures

# Additional Notes

# How to test the change?


thedavl:
> LGTM, how does it impact the binary size?

No change in binary size!

PR #149, opened 2023-05-02
Mini Agent: Support DD_SITE

thedavl:
# What does this PR do?

Support `DD_SITE`, sending traces and trace stats to non-us1 intakes

# Motivation

Mini Agent mvp/closed beta with non-us1 customers

# Additional Notes

# How to test the change?

`config.rs` unit tests. Manually tested by sending traces and trace stats to each region (with the exception of `ddog-gov.com`), and verifying they show up in the UI

### DIFF
--- a/trace-mini-agent/src/config.rs
+++ b/trace-mini-agent/src/config.rs
@@ -24,12 +24,8 @@ pub struct Config {
 
 impl Config {
     pub fn new() -> Result<Config, Box<dyn std::error::Error>> {
-        let api_key = match env::var("DD_API_KEY") {
-            Ok(res) => res,
-            Err(_) => {
-                return Err(anyhow::anyhow!("DD_API_KEY environment variable is not set").into());
-            }
-        };
+        let api_key = env::var("DD_API_KEY")
+            .map_err(|_| anyhow::anyhow!("DD_API_KEY environment variable is not set"))?;
         let mut function_name = None;
 
         // Google cloud functions automatically sets either K_SERVICE or FUNCTION_NAME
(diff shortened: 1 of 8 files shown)


trace-mini-agent/src/config.rs
```diff
@@ -14,11 +17,19 @@ pub struct Config {
...
     pub verify_env_timeout: u64,
+    pub trace_intake_url: String,
+    pub trace_stats_intake_url: String,
+    pub dd_site: String,
 }
 
 impl Config {
     pub fn new() -> Result<Config, Box<dyn std::error::Error>> {
-        let api_key = env::var("DD_API_KEY")?;
+        let api_key = match env::var("DD_API_KEY") {
```
maxday, line 27: you could simplify this block I think as you're not doing anything special with res

maxday (approved):
LGTM! I left a small nit

trace-mini-agent/src/config.rs
thedavl: I didn't do the following because I wanted to return a more descriptive error in case the api key doesn't exist:
```
let api_key = env::var("DD_API_KEY")?
```

And the alternative I found was to do the following:
```
let api_key = env::var("DD_API_KEY").map_err(|_| -> Box<dyn std::error::Error> {
    anyhow::anyhow!("DD_API_KEY environment variable is not set").into()
})?;
```

which I thought was less clear than the current solution. was there an alternative you were thinking of?

maxday: what do you think about:
`let api_key = env::var("DD_API_KEY").map_err(|_| anyhow::anyhow!("DD_API_KEY environment variable is not set"))?;`

thedavl: ooh I like it, I'll do that. Thanks!

PR #152, opened 2023-05-04
Close unwanted fds before exec in sidecar

paullegranddc:
# What does this PR do?

A brief description of the change being made with this pull request.

# Motivation

What inspired you to submit this pull request?

# Additional Notes

Anything else we should know when reviewing?

# How to test the change?

Describe here in detail how the change can be validated.

PR #161, opened 2023-05-19
Fixing all the outstanding issues with the sidecar

bwoebi:
* This branch includes Pauls changes (#151, #152)
* Fix for Pauls PR about the sequential executor: Returning a future which instantly resolves instead of Future::Pending
* Fix for Pauls PR to exclude the memfd from being closed
* Close all the file descriptors in the intermediary forked process too (musl has an atexit handler which wreaks havoc *see below)
* Changed set_session_agent_url to not send a response (some tests were blocking in read(), waiting for the sidecar to reply)
* Temp files are controlled by the process spawning logic and can be removed by the spawned process.
* equeue_actions now can properly handle further updates after the initial flush, like a Lifecycle(Stop) event. Before that went mostly through thanks to the race condition that a stop immediately followed the flush event and then got enqueued before the flush dequeued it... (Also: why is there nothing like a completable future natively in std or tokio?)

With musl, when you read from a file via fread, it'll read up to the next kibibyte. I.e. musl reads 1024 bytes, and stores the current read position offset to the file descriptor position in its FILE struct.
Then, when you exit() a program, musl will update the file descriptions to seek to the actual position.
Now, in PHP CLI scripts, we open the script file and advance past a shebang line. That happens before ddtrace is loaded. When ddtrace is loaded and it forks, the new intermediary (i.e. fork 1 in the double fork) process will also see the FILE struct. It then exits. Helpful musl sees the file descriptor and ... does a seek() to the actual position. When the original process (race conditions ftw) then reads the file, it'll read from 1024 bytes too early (a negative position), leading to a segfault (outside of the mmap()'ped reagion).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -591,6 +591,7 @@ dependencies = [
  "io-lifetimes",
  "lazy_static",
  "libc",
+ "manual_future",
  "nix",
  "pin-project",
  "rand",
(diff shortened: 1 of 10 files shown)


pawelchcki (dismissed):
Looks like I already reviewed some of the code in previous unmerged PRs.
This looks good to me!

pawelchcki (approved):

PR #173, opened 2023-06-12
Mini Agent: Verify azure function env through filesystem

thedavl:
# What does this PR do?

Checks the azure function filesystem at specific paths for two files (`host.json` and `function.json`) to verify we're running in an azure function environment.

# Motivation

azure functions (consumption plan) support

# Additional Notes

# How to test the change?

Unit tested + manually tested with linux + windows azure functions.


trace-mini-agent/src/env_verifier.rs
```diff
@@ -178,20 +199,113 @@ async fn get_gcp_metadata_from_body(body: hyper::Body) -> anyhow::Result<GCPMeta
...
+        for entry in entries {
+            let entry = entry.map_err(|e| anyhow::anyhow!(e))?;
+            let entry_name = entry.file_name();
+            if entry_name == "node_modules" {
+                continue;
+            }
+
+            file_names.push(entry_name.to_string_lossy().to_string());
+
+            if entry.file_type()?.is_dir() {
```
thedavl, line 240: the `function.json` file is located in a sub directory of the main function root folder. This directory is named after the function's handler method which is not exposed via any environment variables. Hence the need to include all files of directories 1 layer deep from the function root dir.

PR #174, opened 2023-06-12
Mini Agent: Verify azure function env w/ filesystem

thedavl:
# What does this PR do?

Checks the azure function filesystem at specific paths for two files (host.json and function.json) to verify we're running in an azure function environment.

# Motivation

azure functions (consumption plan) support

# Additional Notes

# How to test the change?

Unit tested + manually tested with linux + windows azure functions.

### DIFF
--- a/trace-mini-agent/src/config.rs
+++ b/trace-mini-agent/src/config.rs
@@ -39,8 +39,17 @@ impl Config {
 
         let dd_site = env::var("DD_SITE").unwrap_or_else(|_| "datadoghq.com".to_string());
 
-        let trace_intake_url = construct_trace_intake_url(&dd_site, TRACE_INTAKE_ROUTE);
-        let trace_stats_intake_url = construct_trace_intake_url(&dd_site, TRACE_STATS_INTAKE_ROUTE);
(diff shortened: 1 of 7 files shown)


trace-mini-agent/src/config.rs
```diff
@@ -28,15 +36,29 @@ impl Config {
...
+            // Set by Google Cloud Functions for older runtimes
             function_name = Some(res);
+            maybe_env_type = Some(trace_utils::EnvironmentType::CloudFunction);
+        } else if let Ok(res) = env::var("WEBSITE_SITE_NAME") {
+            // Set by Azure Functions
+            function_name = Some(res);
+            maybe_env_type = Some(trace_utils::EnvironmentType::AzureFunction);
         }
 
+        let env_type = maybe_env_type.unwrap_or_else(|| {
```
maxday, line 54: Could we avoid using `#[cfg(not(test))]` and `#[cfg(test)]` here?

thedavl: fixed! process::exit is actually unnecessary here, we can just return

maxday (approved):

PR #176, opened 2023-06-13
Put Connector::default() into lazy_static! to avoid repeated overhead of root cert fetching

bwoebi:
With many different telemetry workers (due to different services in sidecar), overhead can be significant on startup.

### DIFF
--- a/ddcommon/src/connector/mod.rs
+++ b/ddcommon/src/connector/mod.rs
@@ -5,6 +5,7 @@ use futures::future::BoxFuture;
 use futures::{future, FutureExt};
 use hyper::client::HttpConnector;
 
+use lazy_static::lazy_static;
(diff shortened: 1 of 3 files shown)


paullegranddc (approved):

bwoebi:
@ivoanjo This is just the connector, which contains infos about parsed certs etc., it does not reference any file descriptions directly.

ivoanjo (approved):
Thanks for the clarification, I guess that should be fine.

I remember discussing at some point with @r1viollet that one other option to speed this up would be to not load the https certificates unless we need them (since often libdatadog will be talking to the Datadog agent), but I see no reason not to cache the configuration.

Maybe worth anyway dropping a note on slack in case anyone has any concerns, but here sir, have my :+1:

PR #178, opened 2023-06-14
Fix a data race condition when a service is initially flushed multiple times in parallel

bwoebi:
get_app() could be executed twice, while the AppInstance is still building. Prevent that by creating a future around it.

Also refactor register_service_and_flush_queued_actions to be a bit simpler and safe to repeated/concurrent calls of it.


paullegranddc (approved):
It looks ok too me.
Although main thing this achieve is the ability to lock a single key in the hashmap, and I feel like we should be able to find a cleaner pattern to write this because the code is pretty hard to reason about, but we can fix that latter.

bwoebi:
Yes, it's essentially a lock, but I'm not sure how to express this in simpler terms, a read/write sync-safe lock could do it too probably, but coming from an async world, this pattern is fairly standard to me.

PR #181, opened 2023-06-22
[PROF-7645] Add support for attaching internal metadata in profiling export

ivoanjo:
**What does this PR do?**:

This PR adds support for attaching internal metadata to profiles sent via the profiling exporter.

This is a (small) breaking API change, since the signatures of both `ProfileExporter::build` and `ddog_prof_Exporter_Request_build` now take an extra argument.
    
FYI if you're upgrading libdatadog, you'll probably want to supply `None` / `null` until in the future you figure out that you want to send internal metadata.

**Motivation**:

The intention of this internal metadata is to allow profilers to attach extra pieces of information to profiles that are not very interesting to show to customers by default (such as Ruby's "no_signal_workaround_enabled" or Go's "execution_trace_enabled").

For more details on this, check the Datadog internal "RFC: Attaching internal metadata to pprof profiles".

**Additional Notes**:

Design-wise, I could've gone in a few different directions for:

a. How to represent the internal metadata in `ProfileExporter::build`
b. How to represent the internal metadata across the FFI in `ddog_prof_Exporter_Request_build`

Starting with a: Since the information is going to be represented in JSON, I opted to "leak" this implementation detail by making the parameter a `serde_json::Value`. This means that callers can take full advantage of JSON to send whatever they want (e.g. nested objects, types other than string, etc), rather than being constrained to some smaller subset (e.g. if I imposed a list of pairs of strings).

This seemed a reasonable trade-off; I don't expect we'll go away from JSON for encoding this info anytime soon, and even if we do, we can always put a JSON string inside whatever we end up going with.

Concerning b: Getting complex types across the FFI boundary is really really really annoying, for both libdatadog (which needs to expose a bunch of types, and handle them), and the caller (which needs to think about how they're going to manage lifetimes and whatnot of the whole thing). To avoid this, I chose to instead represent the parameter as a raw JSON string across the FFI. This allows ffi users the same expressiveness as Rust users in terms of what they can send as internal metadata, with the trade-off that ffi callers need to serialize their stuff as JSON and libdatadog needs to deserialize it again.

Since internal metadata is something that gets passed along only once per minute AND it's not expected to have high complexity, I think the small overhead of throwing JSON strings across the ffi boundary is worth the simplification to code on both sides.

**How to test the change?**:

I have expanded the tests to test the `event.json` file, including the new parameter.

You should shortly see linked in this PR the Ruby PR to make use of this feature to send the `no_signal_workaround_enabled` parameter.


profiling/src/exporter/mod.rs
```diff
@@ -156,6 +161,8 @@ impl ProfileExporter {
         files: &[File],
         additional_tags: Option<&Vec<Tag>>,
         endpoint_counts: Option<&ProfiledEndpointsStats>,
+        // See Datadog-internal "RFC: Attaching internal metadata to pprof profiles" for details
+        internal_metadata: Option<serde_json::Value>,
```
realFlowControl, line 165: **not blocking**
In it's current state it would accept any valid JSON. Do we know more about what the backend accepts? Like is it always only `"key: "value"` and is value only allowed to be a primitive type (string, boolean, integers, ..) or would it also accept arrays or objects?

realFlowControl (approved):
I think we should somehow link to a resource where we documented the keys that the backend is actually handling. From my point of view it would be totally okay to link to some backend source code file that does the handling so folks can look up what is currently implemented as you would not get any feedback if part of your JSON is just ignored.

ivoanjo:
> I think we should somehow link to a resource where we documented the keys that the backend is actually handling. From my point of view it would be totally okay to link to some backend source code file that does the handling so folks can look up what is currently implemented as you would not get any feedback if part of your JSON is just ignored.

Sounds reasonable! I've added a section in the internal RFC to document current usage, as well as updated the docs to ask people to keep it up-to-date.

The way we've implemented it, the backend doesn't look for particular keys, which is part of why this approach was chosen over, e.g. using a prefix on the tag.

profiling/src/exporter/mod.rs
ivoanjo: It also accepts arrays or objects, see the RFC document for an example where we show it in use :)

realFlowControl (approved):

PR #186, opened 2023-06-29
[ddtelemetry] Flush logs and metrics

paullegranddc:
# What does this PR do?

Flush telemetry logs and metrics before exiting, when the app-closing event is emitted.

# Motivation

Telemetry logs and metrics generated in the last minute before the application exits were dropped, which is not ok.


morrisonlevi:
It looks like you forgot to replace the PR template placeholders ^_^

PR #196, opened 2023-07-31
Fix clippy lints

danielsn:
# What does this PR do?

Fixes two clippy lints

# Motivation

Clippy lints should be fixed when they make sense to

# Additional Notes

Can't think of any.

# How to test the change?

These are both syntax only changes.  Standard `cargo test` run


morrisonlevi (approved):
Thanks for fixing these!

PR #199, opened 2023-08-04
Intern stacktraces to reduce memory usage

danielsn:
# What does this PR do?

Interns the stack trace used by the profiler.  This still uses samples: FxIndexMap<Sample, Vec<i64>> for storage, but Samples now store a StackTraceId instead of Vec<PProfId> (where the ids are for locations). This makes it so when two samples have the same stack trace but different labels, they don't end up duplicating the vector of location ids, and instead share the same StackTraceId.

# Motivation

Currently, multiple samples with the same stack trace end up repeatedly paying the cost of storing that stack trace. Instead, 

# Additional Notes

# How to test the change?

I validated the memory savings using the modifier c-ffi test. 

We should run this on some client libraries to ensure that it doesn't break before merging.


morrisonlevi (approved):
Summary: this still uses `samples: FxIndexMap<Sample, Vec<i64>>` for storage, but `Sample`s now store a `StackTraceId` instead of `Vec<PProfId>` (where the ids are for locations). This makes it so when two samples have the same stack trace but different labels, they don't end up duplicating the vector of location ids.

I'd love to seem some benchmarks from the Ruby profiler's timeline work with before/after to see how it affects memory in practice. Looks good to me, though!

danielsn:
After:

```╰─➤  TIMELINE=true bundle exec ruby ruby_overhead_experiment.rb
Before serialization, process rss usage is 97040k
After serialization, process rss usage is 146000k
After shrinking, process rss usage is 146000k
```

Before:

```
╰─➤  TIMELINE=true bundle exec ruby ruby_overhead_experiment.rb
Before serialization, process rss usage is 139040k
After serialization, process rss usage is 154448k
After shrinking, process rss usage is 153024k
```

PR #200, opened 2023-08-07
Abstract OneWayShm out from AgentRemoteConfig

bwoebi:
Adds a trailing zero byte to keep C code happy. (Turns out PHPs json decoder is really unhappy if the last byte isn't a zero byte.)

And prepares work for being able to use this code for actual remote config (as in remote config from our backend).

### DIFF
--- a/sidecar/src/one_way_shared_memory.rs
+++ b/sidecar/src/one_way_shared_memory.rs
@@ -1,0 +1,207 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+use datadog_ipc::platform::{FileBackedHandle, MappedMem, NamedShmHandle, ShmHandle};
+use std::ffi::CString;
+use std::io;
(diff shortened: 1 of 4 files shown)


sidecar/src/one_way_shared_memory.rs
```diff
@@ -0,0 +1,207 @@
...
+        handle,
+    ))
+}
+
+impl<T: FileBackedHandle + From<MappedMem<T>>, D> OneWayShmReader<T, D> {
+    pub fn new(handle: Option<MappedMem<T>>, extra: D) -> OneWayShmReader<T, D> {
+        OneWayShmReader {
+            handle,
+            current_data: None,
+            extra,
```
paullegranddc, line 82: Is extra going to be anything else than an endpoint?

paullegranddc (approved):

sidecar/src/one_way_shared_memory.rs
bwoebi: No, effectively not (as it stands now), but I need two different "types" of endpoints at least to derive the correct path from the endpoint, for different usages of it (e.g. agent config must not write to the same file than actual remote config).

PR #202, opened 2023-08-07
Aggregate profiling sample observations by timestamp, using a HashMap

danielsn:
# What does this PR do?

This PR refactors the profiling `Sample` structure, moving the `timestamp` from a label to a first class value in the data-structure.  This leads to a significant memory use reduction when the timeline feature is enabled.

# Motivation

The internal format used to store profile samples deduplicates based on stack trace and labels.  In the current format, timestamps are stored as labels, which means that this deduplication cannot happen when timestamps are enabled, significantly increasing memory usage.  By considering the timestamp part of the observations on the sample, and not a label, we allow the deduplication to work and save memory.

# Additional Notes

This saves space on the internal representation, but we still pay the cost of generating the duplicate structures when expanding out the pprof.

# How to test the change?
Following the instructions here https://datadoghq.atlassian.net/wiki/spaces/PROF/pages/2660237526/Ruby+Profiling+Notes I ran the ruby timeline test both with and without the change enabled. 

Before
```
╰─➤  TIMELINE=true bundle exec ruby ruby_overhead_experiment.rb
Before serialization, process rss usage is 96608k
After serialization, process rss usage is 131440k
After shrinking, process rss usage is 131440k
```

After
```
╰─➤  TIMELINE=true bundle exec ruby ruby_overhead_experiment.rb
Before serialization, process rss usage is 75552k
After serialization, process rss usage is 137856k
After shrinking, process rss usage is 137856k
```

### DIFF
--- a/profiling/src/profile/mod.rs
+++ b/profiling/src/profile/mod.rs
@@ -908,7 +908,7 @@ impl TryFrom<&Profile> for pprof::Profile {
 
         Ok(pprof::Profile {
             sample_types: profile.sample_types.clone(),
-            samples: samples, // DSN
+            samples, // DSN
             mappings: profile
                 .mappings
                 .iter()
--- a/profiling/src/profile/mod.rs
+++ b/profiling/src/profile/mod.rs
@@ -558,8 +558,7 @@ impl Profile {
     }
 
     /// Validates labels and converts them to the internal representation.
-    /// Also tracks the index of the label with key "local root span id".
-    ///     /// Validates labels and converts them to the internal representation.
+    /// Extracts out the timestamp label, if it exists.
     /// Also tracks the index of the label with key "local root span id".
     fn extract_sample_labels(
         &mut self,
@@ -908,7 +907,7 @@ impl TryFrom<&Profile> for pprof::Profile {
 
         Ok(pprof::Profile {
             sample_types: profile.sample_types.clone(),
-            samples, // DSN
+            samples,
             mappings: profile
                 .mappings
                 .iter()
--- a/profiling/src/profile/mod.rs
+++ b/profiling/src/profile/mod.rs
@@ -125,7 +125,7 @@ impl UpscalingRule {
 
 pub struct Profile {
     sample_types: Vec<ValueType>,
-    samples: FxIndexMap<Sample, std::collections::HashMap<Option<Timestamp>, Vec<i64>>>,
+    samples: FxIndexMap<Sample, HashMap<Option<Timestamp>, Vec<i64>>>,
     mappings: FxIndexSet<Mapping>,
     locations: FxIndexSet<Location>,
     functions: FxIndexSet<Function>,
--- a/profiling/src/profile/mod.rs
+++ b/profiling/src/profile/mod.rs
@@ -834,33 +834,38 @@ impl Profile {
         Ok(new_values)
     }
 
+    /// Given a sample, and a set of observations at different timestamps,
+    /// expand to one Sample for each timestamp.
     fn expand_sample(
         &self,
         sample: &Sample,
         ts_val_map: &HashMap<Option<Timestamp>, Vec<i64>>,
     ) -> anyhow::Result<Vec<pprof::Sample>> {
-                        // Clone the labels, but enrich them with endpoint profiling.
-                        let mut labels = sample.labels.clone();
-                        if let Some(offset) = sample.local_root_span_id_label_offset {
-                            // Safety: this offset was created internally and isn't be mutated.
-                            let lrsi_label = unsafe { sample.labels.get_unchecked(offset) };
-                            if let Some(endpoint_value_id) = profile.get_endpoint_for_label(lrsi_label)? {
-                                labels.push(Label {
-                                    key: profile.endpoints.endpoint_label,
-                                    str: endpoint_value_id,
-                                    num: 0,
-                                    num_unit: 0,
-                                });
-                            }
-                        }
+        // Clone the labels, but enrich them with endpoint profiling.
+        let mut labels = sample.labels.clone();
+        if let Some(offset) = sample.local_root_span_id_label_offset {
+            // Safety: this offset was created internally and isn't be mutated.
+            let lrsi_label = unsafe { sample.labels.get_unchecked(offset) };
+            if let Some(endpoint_value_id) = self.get_endpoint_for_label(lrsi_label)? {
+                labels.push(Label {
+                    key: self.endpoints.endpoint_label,
+                    str: endpoint_value_id,
+                    num: 0,
+                    num_unit: 0,
+                });
+            }
+        }
         let stacktrace = self.get_stacktrace(sample.stacktrace);
 
+        // For now, pprof requires a separate Sample for each timestamp
+        // So expand it out.
         ts_val_map
             .iter()
             .map(|(timestamp, values)| {
                 let new_values = self.upscale_values(values.as_ref(), labels.as_ref())?;
                 let mut labels = labels.clone();
 
+                // pprof uses a label to store the timestamp so put it there
                 if let Some(ts) = timestamp {
                     labels.push(Label {
                         key: self.timestamp_key,
@@ -895,11 +900,7 @@ impl TryFrom<&Profile> for pprof::Profile {
         let samples: anyhow::Result<Vec<Vec<pprof::Sample>>> = profile
             .samples
             .iter()
-            .map(|(sample, ts_val_map)| {
-
-
-                profile.expand_sample(sample, ts_val_map, &labels)
-            })
+            .map(|(sample, ts_val_map)| profile.expand_sample(sample, ts_val_map))
             .collect();
 
         let samples = samples?;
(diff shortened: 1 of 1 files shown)


profiling/src/profile/mod.rs
```diff
@@ -545,17 +559,40 @@ impl Profile {
 
     /// Validates labels and converts them to the internal representation.
     /// Also tracks the index of the label with key "local root span id".
+    ///     /// Validates labels and converts them to the internal representation.
+    /// Also tracks the index of the label with key "local root span id".
```
morrisonlevi, line 563: "Duplicate" comments here.

profiling/src/profile/mod.rs
```diff
@@ -810,39 +897,18 @@ impl TryFrom<&Profile> for pprof::Profile {
...
+            .map(|(sample, ts_val_map)| profile.expand_sample(sample, ts_val_map))
             .collect();
 
+        let samples = samples?;
+        let samples: Vec<pprof::Sample> = samples.into_iter().flatten().collect();
+
         Ok(pprof::Profile {
             sample_types: profile.sample_types.clone(),
-            samples: samples?,
+            samples, // DSN
```
morrisonlevi, line 911: What's this comment for?

danielsn: Fixed

profiling/src/profile/mod.rs
danielsn: fixed

profiling/src/profile/mod.rs
```diff
@@ -121,7 +125,7 @@ impl UpscalingRule {
 
 pub struct Profile {
     sample_types: Vec<ValueType>,
-    samples: FxIndexMap<Sample, Vec<i64>>,
+    samples: FxIndexMap<Sample, HashMap<Option<Timestamp>, Vec<i64>>>,
```
ivoanjo, line 128: There's a bit of a memory trade-off here: by making this more efficient for timeline data, we're creating an extra `HashMap` (and `Option`?) for every sample, even though there may not ever be any timestamp data.

Note tham I'm not sure this is a blocker -- maybe this is good enough for the other cases that we don't care that much, but I wanted to raise it so we know we're making some cases worse.

profiling/src/profile/mod.rs
```diff
@@ -544,18 +558,40 @@ impl Profile {
...
+                    "the label \"{}\" must be sent as a number, not string {}",
+                    label.str.unwrap(),
+                    self.get_string(self.timestamp_key).unwrap()
+                );
+                anyhow::ensure!(
+                    label.num != 0,
+                    "the label \"{}\" must not be 0",
+                    self.get_string(self.timestamp_key).unwrap()
+                );
+                //TODO what should I ensure the units are?
```
ivoanjo, line 588: I actually wonder if we should get rid of the units and take the memory savings that come with that -- I'm not sure any of our profilers ever sets the units.

profiling/src/profile/mod.rs
```diff
@@ -796,6 +832,56 @@ impl Profile {
...
+                let new_values = self.upscale_values(values.as_ref(), labels.as_ref())?;
+                let mut labels = labels.clone();
+
+                // pprof uses a label to store the timestamp so put it there
+                if let Some(ts) = timestamp {
+                    labels.push(Label {
+                        key: self.timestamp_key,
+                        str: 0,
+                        num: ts.get(),
+                        num_unit: 0, // DSN CHECK THIS
```
ivoanjo, line 873: Minor: See my note above, I suspect we can get rid of this

ivoanjo (commented):
Seems reasonable! I'm not entirely sure about the trade-off in making non-timeline memory usage a bit higher, I wonder if there's a way to "have our cake and eat it too"?

profiling/src/profile/mod.rs
danielsn: We never set any units on any label? Or only on timestamp labels?

profiling/src/profile/mod.rs
danielsn: The `Option` is over a `NonZeroi64` so in memory, this translates to `HashMap<i64, Vec<i64>>`.  
My expectation is that in the case where we don't have a timeline, we will have relitively few samples, so paying the cost of a small `HashMap` should be affordable.  One option would be to move to `BTreeMap`, which is slower but appears to have somewhat less (and more predictable) memory usage.

profiling/src/profile/mod.rs
danielsn: https://github.com/DataDog/libdatadog/issues/204 to track.  That feels like an orthogonal change that is best handled in a separate PR.

profiling/src/profile/mod.rs
danielsn: On the ruby test, there is no significant difference with timeline off

With the `HashMap`
```
Before serialization, process rss usage is 64896k
After serialization, process rss usage is 64992k
After shrinking, process rss usage is 64992k
sample_map false
```

Without
```
Before serialization, process rss usage is 64080k
After serialization, process rss usage is 64176k
After shrinking, process rss usage is 64192k
main false
```

profiling/src/profile/mod.rs
ivoanjo: I mean on any label. I did a quick check:

* [None in Ruby profiler](https://github.com/search?q=repo%3ADataDog%2Fdd-trace-rb+ddog_prof_Label&type=code)
* [Quick check in Python profiler seems none are used](https://github.com/DataDog/dd-trace-py/blob/2eef3c3cad7e95b49dc92a180e9bf4914b2380d5/ddtrace/internal/datadog/profiling/src/exporter.cpp#L438-L456)
* [Don't see it in dotnet](https://github.com/DataDog/dd-trace-dotnet/blob/25f329a4d1af371d1fe42ab6bbce56c734d768a2/profiler/src/ProfilerEngine/Datadog.Profiler.Native/LibddprofExporter.cpp#L383-L396)
* [Also not in the native profiler](https://github.com/DataDog/ddprof/blob/9c3b87f0da7a616f10733ad2d06038640f6fe8a8/src/pprof/ddprof_pprof.cc#L218-L265)

...ahaha but last and not least, [php does seem to be using it](https://github.com/DataDog/dd-trace-php/blob/1114d91b278e1aa3f8de0d06e31e4646744706ed/profiling/src/profiling/mod.rs#L86-L109). So I guess something to raise with our estemeed colleagues?

profiling/src/profile/mod.rs
ivoanjo: The Ruby testing app is probably not a good proxy for this specific comparison, since the stacks never change, so you'll get a very low amount of samples.

I'm not sure if we have metrics for the average amount of stacks in a pprof in the backend, but if not, maybe choosing a few internal services and checking how many unique samples they have may provide a reasonable comparison point maybe?

gleocadie: 👋 I wondering why a `HashMap<Option<Timestamp>, Vec<i64>>>`  ? I was thinking about a `Vec<Pair<Timestamp, Vec<i64>>>` ? (sorry for my Rust, I do not know if `Pair` exists or is there an equivalent.

Because the probability that two identical samples have the same timestamp is really low, no?

profiling/src/profile/mod.rs
```diff
@@ -527,15 +535,21 @@ impl Profile {
...
                 let (_, existing_values) =
                     self.samples.get_index_mut(index).expect("index to exist");
-                for (a, b) in existing_values.iter_mut().zip(values) {
-                    a.add_assign(b)
+                if let Some(existing_values) = existing_values.get_mut(&timestamp) {
+                    for (a, b) in existing_values.iter_mut().zip(values) {
+                        a.add_assign(b)
+                    }
+                } else {
+                    existing_values.insert(timestamp, values);
```
gleocadie, line 550: I would think that most of the time we will go through l.550 but I might be wrong, which means that we would trade CPU against memory.

gleocadie (commented):
I suspect that we would trade CPU against memory. That's acceptable but it would be interesting to have CPU measurements too.

danielsn:
> I suspect that we would trade CPU against memory. That's acceptable but it would be interesting to have CPU measurements too.

I implemented the `vec<pair>` approach in #206.  I measured time and memory usage with timestamps on and off, three times each.  The TLDR is that the vec<pair>` approach might be slightly better, but its hard to tell given the measurement noise.  Which suggests that it comes down to style (or better experiments, if you have them)

profiling/src/profile/mod.rs
ivoanjo: > Because the probability that two identical samples have the same timestamp is really low, no?

I actually discussed this with @danielsn at some point. I was thinking that in a normal situation yes, but if we are run in an OS or VM where the system timestamp counter is not extremely accurate, we may get different samples with the same timestamp.

Which doesn't mean that we have to handle them by aggregation -- we could as well return an error back at the caller. I do still think it's worth dealing with explicitly, vs not handling it very gracefully.

danielsn:
Closed in favour of https://github.com/DataDog/libdatadog/pull/210

PR #205, opened 2023-08-08
Intern profiling labels

danielsn:
# What does this PR do?

Interns the `Label` datastructure used by the profiler.

# Motivation

Currently, the profiler deduplicates that the `Sample` level, where a sample consists of a `StackTrace` plus a set of `Label`s.  However, labels themselves (with some exceptions such as timestamps) are typically drawn a small set of values, and hence are also amenable to deduplication.

# Additional Notes

* For instances where the timeline is running, this will not have much effect until #202 is integrated.
* ~~I had an open question about whether we should remove the "span id" label from the label set, given that its already tracked on the `Sample`.~~. Decided to do this to save memory.
# How to test the change?


danielsn:
With timestamps on, this has a small memory savings
before
```
Before serialization, process rss usage is 95296k
After serialization, process rss usage is 129520k
After shrinking, process rss usage is 128496k
```

after
```
Before serialization, process rss usage is 88816k
After serialization, process rss usage is 123056k
After shrinking, process rss usage is 123056k
```

profiling/src/profile/mod.rs
```diff
@@ -961,12 +1170,17 @@ impl TryFrom<&Profile> for pprof::Profile {
...
                 // Clone the labels, but enrich them with endpoint profiling.
-                let mut labels = sample.labels.clone();
-                if let Some(offset) = sample.local_root_span_id_label_offset {
-                    // Safety: this offset was created internally and isn't be mutated.
-                    let lrsi_label = unsafe { sample.labels.get_unchecked(offset) };
+                let mut labels: Vec<pprof::Label> = profile
+                    .get_labelset(sample.labels)
+                    .iter()
+                    .map(|l| (*profile.get_label(*l)).into())
+                    .collect();
```
ivoanjo, line 1177: Maybe not for this PR, but now that we de-duplicate labelsets, could we reuse the labelset conversion for all samples that have the same labels?

(I guess this wouldn't make a difference for timeline so maybe it's not worth it, but I'm curious if it's simple, or if with the Rust lifetime rules we'd have a hard time doing it).

profiling/src/profile/mod.rs
```diff
@@ -1395,51 +1612,28 @@ mod api_test {
...
+        let mut labels: Vec<_> = s1
+            .labels
+            .iter()
+            .map(|l| {
+                let k = serialized_profile.string_table.get(l.key as usize).unwrap();
+                let s = serialized_profile.string_table.get(l.str as usize).unwrap();
+                (k, s, l.num)
+            })
+            .collect();
+        labels.sort();
```
ivoanjo, line 1624: I realize the existing test already did this but... relying on `sort` order here seems kinda brittle :sweat_smile:.

profiling/src/profile/mod.rs
```diff
@@ -221,16 +342,67 @@ struct StackTrace {
...
 #[derive(Eq, PartialEq, Hash)]
 struct Sample {
     pub stacktrace: StackTraceId,
 
     /// label includes additional context for this sample. It can include
     /// things like a thread id, allocation size, etc
-    pub labels: Vec<Label>,
+    pub labels: LabelSetId,
 
     /// Offset into `labels` for the label with key == "local root span id".
```
ivoanjo, line 404: (The old version of the comment is now outdated)

```suggestion
    /// This label is stored separately, to make it easy to add the endpoint label during serialization.
```

profiling/src/profile/mod.rs
```diff
@@ -127,6 +127,127 @@ pub struct StackTraceId(usize);
...
+
+    pub fn add(&mut self, l: LabelId) {
+        self.sorted_labels.push(l);
+        self.sorted_labels.sort_unstable();
+    }
+
+    pub fn extend(&mut self, ls: &[LabelId]) {
+        self.sorted_labels.extend(ls);
+        self.sorted_labels.sort_unstable();
+    }
```
ivoanjo, line 246: I wonder if we could get by without allowing the `LabelSet` to grow. E.g, only allow it to be created with a fixed-size vec and that's it?

ivoanjo (approved):
While reviewing this PR, I kept thinking back to -- is it worth deduplicating the labels individually, if we already did deduplication of samples?

In the end, I came down on: I think it's worth the extra complexity if it means that libdatadog uses less memory (and potentially less cpu).

I also ran this PR through the Ruby profiler benchmark where we sample as fast as possible, and here's what I saw:

```
stack collector build-pr-205 timeline_enabled=true:     4828.8 i/s
stack collector build-pr-205 timeline_enabled=false:     4574.2 i/s - 1.06x  slower
stack collector build-main-branch-a39a42f128 timeline_enabled=true:     4559.0 i/s - 1.06x  slower
stack collector build-main-branch-a39a42f128 timeline_enabled=false:     4538.6 i/s - 1.06x  slower
```

So this seems to also be a small cpu win, although I wouldn't claim the data used in this Ruby profiler benchmark is very realistic for these kinds of optimizations (since it repeats the same stack and labels over and over again).

I've left a pretty big comment on "how we're treating local_root_span_id" but I'm approving this PR already and I'll leave it up to you if you want to tackle that in this PR, or in a separate one.

As a final suggestion -- should we start extracting a few things from this file? It seems to be getting more and more complex >_>

profiling/src/profile/internal/labels/label_set.rs
```diff
@@ -0,0 +1,51 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+use super::super::*;
+
+/// A canonical representation for sets of labels.
+/// You should only use the impl functions to modify this.
+#[derive(Clone, Debug, Default, Eq, PartialEq, Hash)]
+pub struct LabelSet {
+    sorted_labels: Box<[LabelId]>,
+}
```
ivoanjo, line 11: Calling this a `LabelSet` seems somewhat weird to me since It's not actually a `Set` data structure, and we end up with weird things such as `FxIndexSet<LabelSet>` to actually use it as a `Set`.

Naming is always hard... Here's a few suggestions:

* `SampleLabels`
* `LabelGroup`

profiling/src/profile/internal/sample.rs
```diff
@@ -1,16 +1,17 @@
...
     /// label includes additional context for this sample. It can include
     /// things like a thread id, allocation size, etc
-    pub labels: Vec<Label>,
+    pub labels: LabelSetId,
 
-    /// Offset into `labels` for the label with key == "local root span id".
-    pub local_root_span_id_label_offset: Option<usize>,
+    /// This label is stored separately, to make it easy to add the endpoint
+    /// label during serialization.
+    pub local_root_span_id_label: Option<LabelId>,
```
ivoanjo, line 14: I had left this comment on another part of the code that was since changed but wanted to ask -- have you considered moving the `local_root_span_id_label` inside the `LabelSet`?

Since every sample that shares the same `LabelSet` will have the same `local_root_span_id_label` anyway, it seems the concept could live closer to the labels, rather than closer to the sample.

(Not blocking for this PR at all!)

profiling/src/profile/internal/sample.rs
```diff
@@ -1,16 +1,17 @@
...
 
-use super::{Label, StackTraceId};
+use super::{LabelId, LabelSetId, StackTraceId};
 use std::hash::Hash;
 #[derive(Eq, PartialEq, Hash)]
 pub struct Sample {
     /// label includes additional context for this sample. It can include
     /// things like a thread id, allocation size, etc
-    pub labels: Vec<Label>,
+    pub labels: LabelSetId,
```
ivoanjo, line 10: Minor: Should this perhaps be renamed `label_set`?

ivoanjo (dismissed):
LGTM :+1:

profiling/src/profile/mod.rs
```diff
@@ -678,28 +873,41 @@ impl Profile {
...
+                self.add_label(Label::str(key, str))
+            } else {
+                let num = label.num;
+                let num_unit = label.num_unit.map(|s| self.intern(s));
+                self.add_label(Label::num(key, num, num_unit))
+            };
+
+            // TODO: figure out if we need to push this anymore.
+            // Not pushing it would avoid duplication in the Sample,
+            // and increase the chance of deduplication.
```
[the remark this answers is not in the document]
danielsn: Given that labelsets are small, and we're already iterating over them, I just decide to lazily do the pass at the end when we're translating to pprof.

profiling/src/profile/mod.rs
danielsn: Yeah.  We should fix the broader set of these tests, which are indeed kinda brittle

profiling/src/profile/internal/labels/mod.rs
morrisonlevi, line 1: I wouldn't bother with separate files for label and label sets. I'd just put it all in `internal/labels.rs`.

profiling/src/profile/internal/labels/label_set.rs
morrisonlevi: I'm not sure if it matters in practice, but `[0, 1, 2]` and `[2, 1, 0]` wouldn't be the same "LabelSet". The number of items in these sets is expected to be small. I'm not sure if it's worth "normalizing" by either:

- Sorting the vec.
- Using a set structure.

But if we do nothing, then I agree with Ivo that it's not really a set, so we should probably avoid that name.

profiling/src/profile/internal/labels/label_set.rs
```diff
@@ -0,0 +1,51 @@
...
+impl LabelSet {
+    pub fn iter(&self) -> core::slice::Iter<'_, LabelId> {
+        self.sorted_labels.iter()
+    }
+
+    pub fn new(mut v: Vec<LabelId>) -> Self {
+        v.sort_unstable();
+        let sorted_labels = v.into_boxed_slice();
+        Self { sorted_labels }
+    }
```
morrisonlevi, line 22: I wonder how often `v` is the correct size already. It _may_ be prudent to take a `Box<[LabelId]>` in the constructor instead, so the caller is more aware of making a correctly sized slice in the first place, instead of having excess capacity which will then trigger re-allocation.

morrisonlevi: However, if we move to using an arena, it will may cause a re-allocation anyway. So not high priority.

profiling/src/profile/internal/labels/mod.rs
danielsn: Done

profiling/src/profile/internal/sample.rs
danielsn: Done

profiling/src/profile/internal/labels/label_set.rs
danielsn: I'll leave that for a later PR

profiling/src/profile/internal/sample.rs
danielsn: Personally, I find this name cleaner.

profiling/src/profile/internal/labels/label_set.rs
danielsn: Added a comment to make this more obvious

profiling/src/profile/mod.rs
```diff
@@ -127,6 +127,127 @@ pub struct StackTraceId(usize);
...
+impl From<LabelSetId> for usize {
+    fn from(s: LabelSetId) -> Self {
+        Self::from(&s)
+    }
+}
+
+impl From<&LabelSetId> for usize {
+    fn from(s: &LabelSetId) -> Self {
+        // Safety: we only run on platforms where usize >= u32
+        s.0 as usize
```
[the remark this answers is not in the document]
danielsn: It would silently succeed.

profiling/src/profile/mod.rs
danielsn: I'll look at this for a later PR.

profiling/src/profile/internal/observation/observations.rs
```diff
@@ -50,31 +53,31 @@ impl Observations {
...
+            observations.aggregated_data.insert(sample, trimmed);
         }
     }
 
     pub fn is_empty(&self) -> bool {
         self.inner.is_none()
     }
 
-    pub fn iter(&self) -> impl Iterator<Item = (SampleId, Option<Timestamp>, &[i64])> {
+    pub fn iter(&self) -> impl Iterator<Item = (&Sample, Option<Timestamp>, &[i64])> {
```
morrisonlevi, line 64: Since `Sample` is only 64-bit, I think it would be nicer to return a `Sample` instead of a `&Sample`, particularly because this is immutable.

profiling/src/profile/internal/observation/observations.rs
```diff
@@ -118,9 +131,9 @@ mod test {
         o.add(s2, None, vec![7, 8, 9]);
         o.iter().for_each(|(k, ts, v)| {
             assert!(ts.is_none());
-            if k == s1 {
+            if *k == s1 {
```
morrisonlevi, line 134: Notably, if you returned `Sample` intead of `&Sample`, you wouldn't have to do `*k` :)

profiling/src/profile/internal/sample.rs
```diff
@@ -1,25 +1,27 @@
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
 
-use super::{Id, Item, Label, StackTraceId};
+use super::{Id, Item, LabelSetId, StackTraceId};
 
 use std::hash::Hash;
-#[derive(Eq, PartialEq, Hash)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
```
morrisonlevi, line 7: I love it when things become `Copy` that weren't before 🚀

morrisonlevi (commented):
I have just a very small suggestion!

profiling/src/profile/internal/observation/observations.rs
danielsn: Done

profiling/src/profile/internal/observation/observations.rs
danielsn: done

morrisonlevi (approved):

profiling/src/profile/mod.rs
```diff
@@ -525,21 +543,41 @@ impl Profile {
...
-            }
-        }
+        let labels: Vec<_> = self
+            .get_label_set(sample.labels)
+            .iter()
+            .map(|l| self.get_label(*l).into())
+            .chain(
+                self.get_endpoint_for_labels(sample.labels)?
+                    .map(pprof::Label::from),
+            )
```
ivoanjo, line 577: Since the list of labels is quite small, maybe this is not a big difference, but it seems kinda weird that we're iterating through the labels twice: once to covert them into `pprof::Label`s, and then again to search for the `local root span id` label; in the old code we actually did both intertwined -- conversion and finding the label.

PR #207, opened 2023-08-09
Use a unique type for StringId

danielsn:
# What does this PR do?

Increases type safety by using a first class type for `StringId`.

# Motivation

Currently, strings are identified by an `i64` value, which represents an index into a string table.  This is error prone, since it is easy for a `i64` representing a value to be confused with a string index, potentially leading to a `panic`.  If we store the index in a transparent type, the type system both documents and enforces the correct use of string indicies.

As a side benefit, reduced the size of the index from `i64` to `u32`, saving 4 bytes per index.

# Additional Notes

There are a couple places where we have to unsafely take pprof indicies from a label and turn it back into a `StringId`.  This will be trivial to fix once #205 is merged.

# How to test the change?


profiling/src/profile/mod.rs
```diff
@@ -553,13 +582,19 @@ impl Profile {
...
+            let num_unit = label
+                .num_unit
+                .map(|s| self.intern(s))
+                .unwrap_or(StringId::zero());
 
             if key == self.endpoints.local_root_span_id_label {
                 // Panic: if the label.str isn't 0, then str must have been provided.
                 anyhow::ensure!(
-                    str == 0,
+                    str.is_zero(),
```
sanchda, line 597: Ah, I was a little confused why you added a special `is_zero` method, but this makes sense.  0 is special as far as these IDs go.

profiling/src/profile/mod.rs
```diff
@@ -907,7 +948,7 @@ mod api_test {
         }];
         let mut profiles = Profile::builder().sample_types(sample_types).build();
 
-        let expected_id: i64 = profiles.interned_strings_count().try_into().unwrap();
+        let expected_id = StringId::new(profiles.interned_strings_count());
```
sanchda, line 951: much cleaner

sanchda (approved):
Much cleaner.  Good idea--thank you!

Left a totally pointless comment, barely even a nit.

PR #208, opened 2023-08-09
Use unique types for PProfIds

danielsn:
# What does this PR do?

Increases type safety by using a first class type for various `PProfId` types, including `LocationId`, `MappingId` and `FunctionId`. Also does a bit of refactoring on the existing unique types as per PR comments.

# Motivation

Currently, Locations are identified by an `PProfId` value, which represents an index into a interning table.  This is error prone, since there are many PProfIds, generated by many different tables.  If we use the wrong one, we can get a panic, or nonsense values. If we store the index in a transparent type, the type system both documents and enforces the correct use of indicies.

As a side benefit, reduced the size of the index from `usize` to `u32`, saving 4 bytes per index.

# Additional Notes

# How to test the change?

### DIFF
--- a/profiling/src/profile/mod.rs
+++ b/profiling/src/profile/mod.rs
@@ -190,34 +190,6 @@ impl From<&StringId> for i64 {
     }
 }
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
-#[repr(transparent)]
-pub struct PProfId(usize);
-
-impl From<&PProfId> for u64 {
-    fn from(id: &PProfId) -> Self {
-        id.0 as u64
-    }
-}
-
-impl From<PProfId> for u64 {
-    fn from(id: PProfId) -> Self {
-        id.0.try_into().unwrap_or(0)
-    }
-}
-
-impl From<&PProfId> for i64 {
-    fn from(value: &PProfId) -> Self {
-        value.0.try_into().unwrap_or(0)
-    }
-}
-
-impl From<PProfId> for i64 {
-    fn from(value: PProfId) -> Self {
-        value.0.try_into().unwrap_or(0)
-    }
-}
-
 #[derive(Eq, PartialEq, Hash)]
 struct Mapping {
     /// Address at which the binary (or DLL) is loaded into memory.
--- a/profiling/src/profile/mod.rs
+++ b/profiling/src/profile/mod.rs
@@ -24,6 +24,38 @@ pub type FxIndexSet<K> = indexmap::IndexSet<K, BuildHasherDefault<rustc_hash::Fx
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 #[repr(transparent)]
+pub struct FunctionId(NonZeroU32);
+
+impl FunctionId {
+    pub fn new<T>(v: T) -> Self
+    where
+        T: TryInto<u32>,
+        T::Error: Debug,
+    {
+        let index: u32 = v.try_into().expect("FunctionId to fit into a u32");
+
+        // PProf reserves location 0.
+        // Both this, and the serialization of the table, add 1 to avoid the 0 element
+        let index = index.checked_add(1).expect("FunctionId to fit into a u32");
+        let index = NonZeroU32::new(index).unwrap();
+        Self(index)
+    }
+}
+
+impl From<FunctionId> for u64 {
+    fn from(s: FunctionId) -> Self {
+        Self::from(&s)
+    }
+}
+
+impl From<&FunctionId> for u64 {
+    fn from(s: &FunctionId) -> Self {
+        s.0.get().into()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


sanchda (approved):

profiling/src/profile/mod.rs
```diff
@@ -58,6 +57,38 @@ impl From<&SampleId> for u64 {
...
+
+impl LocationId {
+    pub fn new<T>(v: T) -> Self
+    where
+        T: TryInto<u32>,
+        T::Error: Debug,
+    {
+        let index = v
+            .try_into()
+            .expect("the machine to run out of memory far before this happens");
```
morrisonlevi, line 72: I think using an error message like this is probably more descriptive and helpful:

```suggestion
            .expect("location ids to fit in u32");
```

profiling/src/profile/mod.rs
```diff
@@ -58,6 +57,38 @@ impl From<&SampleId> for u64 {
...
+
+        // PProf reserves location 0.
+        // Both this, and the serialization of the table, add 1 to avoid the 0 element
+        Self(index + 1)
+    }
+}
+
+impl From<LocationId> for u64 {
+    fn from(s: LocationId) -> Self {
+        s.0.try_into().unwrap()
```
morrisonlevi, line 82: Delegate to the implementation.

```suggestion
        Self::from(&s)
```

profiling/src/profile/mod.rs
```diff
@@ -58,6 +57,38 @@ impl From<&SampleId> for u64 {
...
+        T: TryInto<u32>,
+        T::Error: Debug,
+    {
+        let index = v
+            .try_into()
+            .expect("the machine to run out of memory far before this happens");
+
+        // PProf reserves location 0.
+        // Both this, and the serialization of the table, add 1 to avoid the 0 element
+        Self(index + 1)
```
morrisonlevi, line 76: This can overflow, we should probably panic on that.

profiling/src/profile/mod.rs
```diff
@@ -58,6 +57,38 @@ impl From<&SampleId> for u64 {
 #[repr(transparent)]
 pub struct StackTraceId(usize);
 
+#[derive(Copy, Clone, Default, Debug, Eq, PartialEq, Hash)]
+#[repr(transparent)]
+pub struct LocationId(u32);
```
morrisonlevi, line 62: Thoughts on using `NonZeroU32` instead of `u32` since we literally are specifically avoiding 0 anyway?

danielsn: Good idea.  Will make `Option<LocationId>` cheaper

profiling/src/profile/mod.rs
danielsn: Done

profiling/src/profile/mod.rs
```diff
@@ -58,6 +57,38 @@ impl From<&SampleId> for u64 {
...
+
+impl From<LocationId> for u64 {
+    fn from(s: LocationId) -> Self {
+        s.0.try_into().unwrap()
+    }
+}
+
+impl From<&LocationId> for u64 {
+    fn from(s: &LocationId) -> Self {
+        s.0.try_into().unwrap()
```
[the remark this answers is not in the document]
danielsn: Done (well, I used `.into()`, but it now is an infallible conversion)

profiling/src/profile/mod.rs
danielsn: Done

profiling/src/profile/mod.rs
danielsn: Done

morrisonlevi (commented):
Florian and I are working on making benchmarks for the PHP profiler so we can before/after test this PR against it. At the moment, I don't think it will actually save memory because mostly we create the ids and then turn them back into the full size version to save them in the `pprof::*` structs. To get significant savings we'd have to create structs which mirror the pprof ones but have reduced indices. I'm happy to do this as a follow-up PR.

profiling/src/profile/mod.rs
```diff
@@ -22,98 +22,171 @@ use self::api::UpscalingInfo;
...
+    where
+        T: TryInto<u32>,
+        T::Error: Debug,
+    {
+        let index: u32 = v.try_into().expect("FunctionId to fit into a u32");
+
+        // PProf reserves location 0.
+        // Both this, and the serialization of the table, add 1 to avoid the 0 element
+        let index = index.checked_add(1).expect("FunctionId to fit into a u32");
+        let index = NonZeroU32::new(index).unwrap();
```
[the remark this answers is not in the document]
danielsn: I could go either way.  My personal aesthetic bias is against `unsafe` when it can be avoided. But this is indeed clearly safe, and would be a few instructions faster.

morrisonlevi (approved):

PR #206, opened 2023-08-08
Aggregate profiling sample observations by timestamp, using a vector

danielsn:
# What does this PR do?

This PR refactors the profiling `Sample` structure, moving the `timestamp` from a label to a first class value in the data-structure.  This leads to a significant memory use reduction when the timeline feature is enabled.  This is a competing PR with #202, using a different data-structure. 

# Motivation

The internal format used to store profile samples deduplicates based on stack trace and labels.  In the current format, timestamps are stored as labels, which means that this deduplication cannot happen when timestamps are enabled, significantly increasing memory usage.  By considering the timestamp part of the observations on the sample, and not a label, we allow the deduplication to work and save memory.

# Additional Notes

This saves space on the internal representation, but we still pay the cost of generating the duplicate structures when expanding out the pprof.

# How to test the change?
Following the instructions here https://datadoghq.atlassian.net/wiki/spaces/PROF/pages/2660237526/Ruby+Profiling+Notes I ran the ruby timeline test both with and without the change enabled. 

With this change, timestamps on
```
config: {:threads=>16, :depth=>50, :seconds=>60, :timeline_enabled=>true}
Before serialization, process rss usage is 71472k
After serialization, process rss usage is 113040k
Size of uncompressed serialized pprof: 9014k
After shrinking, process rss usage is 113056k
        2.94 real         2.47 user         0.19 sys
config: {:threads=>16, :depth=>50, :seconds=>60, :timeline_enabled=>true}
Before serialization, process rss usage is 71600k
After serialization, process rss usage is 113808k
Size of uncompressed serialized pprof: 9014k
After shrinking, process rss usage is 113824k
        2.75 real         2.47 user         0.19 sys
config: {:threads=>16, :depth=>50, :seconds=>60, :timeline_enabled=>true}
Before serialization, process rss usage is 71600k
After serialization, process rss usage is 113888k
Size of uncompressed serialized pprof: 9014k
After shrinking, process rss usage is 113904k
        2.75 real         2.48 user         0.19 sys
```

With the `HashMap`, timestamps on
```
config: {:threads=>16, :depth=>50, :seconds=>60, :timeline_enabled=>true}
Before serialization, process rss usage is 73488k
After serialization, process rss usage is 128656k
Size of uncompressed serialized pprof: 9014k
After shrinking, process rss usage is 127648k
        2.92 real         2.49 user         0.19 sys
config: {:threads=>16, :depth=>50, :seconds=>60, :timeline_enabled=>true}
[...]

PR #210, opened 2023-08-10
Reduce memory impact of timestamps by factoring out timestamped and aggregated data

danielsn:
# What does this PR do?

This is the third proposal in a set of options as to how to best reduce the memory usage of timestamping (#206 #202).

This approach has the lowest performance cost of the three approaches: 
* for non-timestamped profiles, the memory cost is of a single empty hash-map, and the time cost should be nil.
* for timestamped profiles, we keep a `Vec<(Timestamp, values)>` per sample, allowing samples to dedup on timestamps.  Other than the possible wasted space of a partially full vec, this should be close to optimal for storing timestamped values.

The downside is that there is a slight change to the API: Instead of having a single `SampleId`, we how have two (see additional Notes).  I didn't see anyone actually using the `SampleId` outside the profiler, but it is theoretically exposed in the API so this is a breaking change.

# Motivation

The internal format used to store profile samples deduplicates based on stack trace and labels. In the current format, timestamps are stored as labels, which means that this deduplication cannot happen when timestamps are enabled, significantly increasing memory usage. By considering the timestamp part of the observations on the sample, and not a label, we allow the deduplication to work and save memory.

# Additional Notes

A [quick code search](https://github.com/search?q=org%3ADataDog+ddog_prof_Profile_add&type=code) suggests no code that uses the sample_id return value except to check for error.

This change modified some of the hardcoded ids used in unit tests: I modified the tests to match the new results.

# How to test the change?

On a ruby benchmark: 

**Before, timeline on**
```
Before serialization, process rss usage is 97728k
After serialization, process rss usage is 145120k
After shrinking, process rss usage is 145120k
        3.05 real         2.52 user         0.25 sys
Before serialization, process rss usage is 95552k
After serialization, process rss usage is 130960k
After shrinking, process rss usage is 129936k
        2.83 real         2.50 user         0.24 sys
Before serialization, process rss usage is 95024k
After serialization, process rss usage is 126416k
After shrinking, process rss usage is 126416k
        2.84 real         2.51 user         0.25 sys
```

**After, timeline on**
```
Before serialization, process rss usage is 73968k
After serialization, process rss usage is 130208k
After shrinking, process rss usage is 129728k
        3.19 real         2.54 user         0.26 sys
Before serialization, process rss usage is 72256k
After serialization, process rss usage is 114672k
After shrinking, process rss usage is 114672k
        2.82 real         2.49 user         0.24 sys
Before serialization, process rss usage is 71584k
After serialization, process rss usage is 114864k
After shrinking, process rss usage is 113840k
        2.84 real         2.52 user         0.24 sys
```

**before, timeline off**
```
Before serialization, process rss usage is 66928k
After serialization, process rss usage is 67088k
After shrinking, process rss usage is 67088k
        2.92 real         2.46 user         0.23 sys
Before serialization, process rss usage is 66624k
After serialization, process rss usage is 66784k
After shrinking, process rss usage is 66784k
        2.81 real         2.50 user         0.23 sys
Before serialization, process rss usage is 65456k
After serialization, process rss usage is 65616k
After shrinking, process rss usage is 65616k
        2.81 real         2.49 user         0.23 sys
```

**after, timeline off**
```
Before serialization, process rss usage is 64288k
After serialization, process rss usage is 64432k
After shrinking, process rss usage is 64432k
        2.92 real         2.44 user         0.23 sys
Before serialization, process rss usage is 63888k
After serialization, process rss usage is 64048k
After shrinking, process rss usage is 64064k
        2.75 real         2.44 user         0.23 sys
Before serialization, process rss usage is 63824k
After serialization, process rss usage is 63968k
After shrinking, process rss usage is 64000k
        2.75 real         2.43 user         0.23 sys
```


profiling/src/profile/mod.rs
```diff
@@ -659,47 +666,77 @@ impl Profile {
...
+                    SampleId::timestamped(idx)
+                }
+                Some((idx, _, series)) => {
+                    let last_ts = series.last().unwrap().0;
+                    anyhow::ensure!(
+                        ts > last_ts,
+                        "Timestamps should be monotonically increasing, but instead saw {} {}",
+                        last_ts,
+                        ts
+                    );
```
morrisonlevi, line 683: How important is this guarantee? I'm not sure it would hold for runtimes like .NET where things are multi-threaded and queued. It seems probable that the order the queued things are processed will not result in monotonic timestamps at some point, even ignoring cases like time going backwards.

danielsn: This guarantee is not necessary to the implementation, (with one caveat).  This was more of a sanity check, that  

The caveat is the question is what to do in the case of duplicated timestamps for the same sample.  Without this check, we would simply record both observations to the sample.  With the check, we would reject it as impossible.  What is the desired behaviour in that case?

ivoanjo: Aggregating two observations with the same timestamp makes sense to me conceptually -- we're saying that multiple things happened with that same timestamp.

profiling/src/profile/mod.rs
```diff
@@ -942,6 +987,52 @@ impl Profile {
...
         Ok(new_values)
     }
+
+    /// Given a sample, and a set of observations at different timestamps,
+    /// expand to one Sample for each timestamp.
+    fn expand_timestamped_sample(
+        &self,
+        sample: &Sample,
+        observations: &[(Timestamp, Vec<i64>)],
+    ) -> anyhow::Result<Vec<pprof::Sample>> {
```
ivoanjo, line 997: I wonder if we could reuse more of the code for this step when comparing to the timestampless variant?

profiling/src/profile/mod.rs
danielsn: I removed the check

profiling/src/profile/mod.rs
```diff
@@ -260,9 +230,20 @@ impl UpscalingRule {
...
 }
 
+// For testing and debugging purposes
+impl Profile {
+    pub fn num_aggregated_samples(&self) -> usize {
+        self.aggregated_samples.len()
+    }
+    pub fn num_timestamped_samples(&self) -> usize {
+        self.timestamped_samples.len()
+    }
```
ivoanjo, line 240: Minor: Should we prefix these methods with `testing_` or even `only_for_testing_`? The objective being to make it kinda ugly and obvious even when doing an IDE autocomplete that you shouldn't be using these methods :)

profiling/src/profile/mod.rs
```diff
@@ -659,47 +645,60 @@ impl Profile {
...
-            None => {
-                self.samples.insert(s, values);
-                SampleId::new(self.samples.len() - 1)
+        if let Some(ts) = timestamp {
+            match self.timestamped_samples.get_mut(&s) {
+                None => {
+                    let series = vec![(ts, values)];
+                    self.timestamped_samples.insert(s, series);
+                }
+                Some(series) => series.push((ts, values)),
```
ivoanjo, line 654: This seems like a reasonable solution to the repeated timestamp issue -- do nothing!

The pprof spec doesn't say you need to aggregate stuff, but you usually want to do it to save on memory/space. But storing it as a separate sample even though it maybe could've been aggregated if a timestamp is repeated seems like a reasonable trade-off!

(I think it's worth leaving a comment stating this, in case people wonder about it in the future, since it's a quite subtle detail)

profiling/src/profile/mod.rs
```diff
@@ -659,47 +645,60 @@ impl Profile {
...
+
+            if key == self.timestamp_key {
+                anyhow::ensure!(
+                    label.str.is_none(),
+                    "the label \"{}\" must be sent as a number, not string {}",
+                    label.str.unwrap(),
+                    label.key
+                );
+                anyhow::ensure!(label.num != 0, "the label \"{}\" must not be 0", label.key);
+                //TODO what should I ensure the units are?
```
ivoanjo, line 691: Maybe barf an error back at the caller if they set the units up? In the spirit of trying to slowly push people away from using units? ;)

profiling/src/profile/mod.rs
```diff
@@ -659,47 +645,60 @@ impl Profile {
...
+
+                timestamp = Some(NonZeroI64::new(label.num).unwrap());
+                // Don't put the timestamp into the labels
```
ivoanjo, line 694: Minor: May be worth expanding the comment a bit to explain that this label gets treated specially and for that reason.

profiling/src/profile/mod.rs
```diff
@@ -986,9 +1036,15 @@ impl TryFrom<&Profile> for pprof::Profile {
             })
             .collect();
 
+        let samples = timestamped_samples?
+            .into_iter()
+            .flatten()
+            .chain(aggregated_samples?)
+            .collect();
```
ivoanjo, line 1043: I'm not quite sure about the `?` usage here -- what happens if there are only timestamped samples or only aggregated samples?

profiling/src/profile/mod.rs
danielsn: I made a helper function to reuse code

profiling-ffi/src/profiles.rs
```diff
@@ -13,7 +13,8 @@ use std::time::{Duration, SystemTime};
 
 #[repr(C)]
 pub enum ProfileAddResult {
-    Ok(u64),
+    // Do not use the value of Ok. This value only exists to overcome Rust -> C code generation.
```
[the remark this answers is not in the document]
danielsn: Done

profiling-ffi/src/profiles.rs
```diff
@@ -363,8 +364,8 @@ impl From<ProfileAddResult> for Result<u64, String> {
 /// module. All pointers inside the `sample` need to be valid for the duration
 /// of this call.
 ///
-/// If successful, it returns the internal id of the sample (> 0) in the Ok
-/// variant. On error, it holds an error message in the error variant.
+/// If successful, it returns `true` in the Ok variant.
+/// On error, it holds an error message in the error variant.
```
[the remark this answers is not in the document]
danielsn: fixed

profiling/src/profile/mod.rs
danielsn: done

profiling/src/profile/mod.rs
danielsn: Done

profiling/src/profile/mod.rs
danielsn: done

profiling/src/profile/mod.rs
danielsn: If either of the `timestamped_samples` or `aggregates_samples` had an error during conversion, the `?` will propagate the error upwards.  If neither had an error, then the iter will take the 0 or more elements from `timestamped_samples`, and then append the 0 or more elements from `aggregates_samples`, and then resolve to a list with both sets of samples in it.

ivoanjo: Ah, great, that makes a lot of sense! Thanks for putting up with my Rust noobness :)

(I was reading it with my Ruby eyes and suspecting the `?` may be similar to Ruby's `&.` which could have made this a bug, but now that you've explained it the `?` is nothing like the Ruby operator and I was just really confused :D )

profiling/src/profile/mod.rs
```diff
@@ -1020,6 +1112,45 @@ mod api_test {
         assert_eq!(label.num, 101);
         assert_eq!(str, "");
         assert_eq!(num_unit, "");
+
+        // For some reason, this test never covered index 1
```
ivoanjo, line 1116: Minor: This is a slightly weird comment to me... E.g. I'm not sure it's interesting going forward for this fact to be documented in the code, since you've fixed it now :)

ivoanjo (approved):
:+1: LGTM

profiling/src/profile/mod.rs
```diff
@@ -934,6 +992,16 @@ mod api_test {
...
 
+        let timestamp_lines = vec![api::Line {
+            function: api::Function {
+                name: "test",
+                system_name: "test",
+                filename: "index.php",
+                start_line: 3,
+            },
+            line: 0,
+        }];
```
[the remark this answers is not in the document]
danielsn: Merged, fixed conflicts

profiling/src/profile/mod.rs
```diff
@@ -770,44 +824,46 @@ impl TryFrom<&Profile> for pprof::Profile {
...
-                            num: 0,
-                            num_unit: 0,
-                        });
-                    }
-                }
-
-                let new_values = profile.upscale_values(values.as_ref(), &labels)?;
-                let stacktrace = profile.get_stacktrace(sample.stacktrace);
+            .map(|(sample, observations)| profile.expand_timestamped_sample(sample, observations))
+            .collect();
```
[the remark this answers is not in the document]
danielsn: It took a `flat_map` plus a `flatten()` but I got it to work

profiling/src/profile/mod.rs
danielsn: Fixed

ivoanjo (approved):
:+1: LGTM

PR #211, opened 2023-08-11
Refactor profiler to make a cleaner seperation with pprof

danielsn:
# What does this PR do?

Refactors the profiler to make a cleaner seperation between the profiler internal representation and pprof

# Motivation

The profiler has three layers: API, internal representation, and pprof.  The internal representation is currently entangled with the pprof types, which is undesirable:

* It ties the IR into (potentially memory inefficient) pprof structures, with larger indicies than needed.
* It limits memory safety: many indicies are `i64` rather than typesafe transparent types
* It makes the code less modular and harder to understand.

This PR addresses these issues by creating a cleaner IR without the references to pprof types.

# Additional Notes

* The new structures were factored out into their own file structure, away from the 2.5KLOC file.  More work on this in the future would be great, but is a topic for a future PR.
* 32 bits ought to be enough for everyone, right?  Are there any structures where `u32` is too small to be a reasonable index value?
* the `upscale` values function currently takes `pprof::Label`s.  This might be a good target for a future refactor.

# How to test the change?

### DIFF
--- a/profiling/src/profile/internal/function.rs
+++ b/profiling/src/profile/internal/function.rs
@@ -1,0 +1,72 @@
+pub use super::super::pprof;
+pub use super::super::StringId;
+use std::fmt::Debug;
+use std::num::NonZeroU32;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
+pub struct Function {
+    pub name: StringId,
+    pub system_name: StringId,
+    pub filename: StringId,
+    pub start_line: u32,
+}
+
+impl Function {
+    pub fn new<T>(name: StringId, system_name: StringId, filename: StringId, start_line: T) -> Self
+    where
+        T: TryInto<u32>,
+        T::Error: Debug,
+    {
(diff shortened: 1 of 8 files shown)


profiling/src/profile/internal/function.rs
```diff
@@ -0,0 +1,75 @@
...
+pub use super::super::StringId;
+use std::fmt::Debug;
+use std::num::NonZeroU32;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
+pub struct Function {
+    pub name: StringId,
+    pub system_name: StringId,
+    pub filename: StringId,
+    pub start_line: u32,
```
morrisonlevi, line 14: I think `start_line` needs to stay 64-bit. Presumably native could hit this.

morrisonlevi (commented):
By scanning, it looks good. I'm going to keep tugging at it because I think it can go a little more 👍🏻

danielsn:
Superseded by #214 which includes this work and further improvements

PR #212, opened 2023-08-12
feat: add datadog-profiling-replayer

morrisonlevi:
# What does this PR do?

This adds a new `datadog-profiling-replayer` crate which has a command line tool:

```
% datadog-profiling-replayer -h
Takes a pprof file and 'replays' it using libdatadog commands.

Usage: datadog-profiling-replayer [OPTIONS] -i <input>

Options:
  -i <input>       the pprof to replay
  -o <output>      the path to save the result to
  -h, --help       Print help
  -V, --version    Print version
  ```

It will print out how long it took to replay adding samples -- at the moment that's all it benchmarks. Before it adds samples, it it will remove `trace endpoint` labels if there is also a `local root span id` label, and then adds them back after benchmarking, and writes out the pprof to the specified location (if provided, that part is optional).

It also prints out some stats about the stack depths of the samples. Example output on a small PHP profile:

```
Number of samples: 9521.
Min stack depth is 1.
Q1 = 2, Q2 = 2, Q3 = 2.
Max stack depth is 51.
```

# Motivation

The goal is to be able to replay changes made to libdatadog to be able to determine their impact. For this research and development week, my goal is specifically to use it to benchmark changes to APIS.

# Additional Notes

It's not polished. For instance, we probably want to measure memory, and this doesn't do that. It also prints to stdout in an unstructured way -- we probably want to emit something that is easy to consume for programs.

# How to test the change?

Download a profile, and save it as `input.pprof`. Then, run `datadog-profile-replayer -i input.pprof -o output.pprof`. You can compare the profiles using tools like `go tool pprof -http=: input.pprof` and `go tool pprof -http=: output.pprof` or `protoc --decode`

### CODE REVIEW

File: profiling-replayer/Cargo.toml [hunk 1]
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "datadog-profiling-replayer"
+authors = ["Levi Morrison <levi.morrison@datadoghq.com>"]
+description = "Takes a pprof file and 'replays' it using libdatadog commands."
+version = "3.0.0"
```
danielsn, line 5: why 3.0?

danielsn:
https://docs.rs/sysinfo/latest/sysinfo/trait.ProcessExt.html#tymethod.memory might be a good way to get the memory usage

profiling-replayer/Cargo.toml
morrisonlevi: It matches the version field of `profiling/Cargo.toml`.

danielsn (dismissed):

danielsn (commented):
LGTM other than one math error

danielsn (approved):

PR #213, opened 2023-08-12
Add dd-trace-filter

Qard:
# What does this PR do?

Adds dd-trace-filter, which is a msgpack binary-to-binary transformer to apply truncation and (eventually) obfuscation logic in a way which can easily be shared across libraries.

# Motivation

This was a R&D Week project to see if it would be reasonable to build something in WASM to filter msgpack encoded trace chunks to truncate long fields and obfuscate anything that needs obfuscating.


ivoanjo:
So... My rust-foo is definitely not great to help out with the actual code side.

On the other hand, it seems harmless to me for us to merge this in -- worse thing we can move or remove it at some later point in the future. Do you mind updating the README with a bit of context on what this is?

I'm otherwise happy to approve. Or, you can try chasing people down in slack to get an actual review of the code :D

Qard:
I'm not in a huge rush to merge _immediately_. I just opened a PR mainly to have something for people to look at. It's still somewhat in-progress though. I can find a bit of time soon to do a bit more cleanup and land then. It should at least probably be passing the CI before it lands. Don't want to interfere with CI on other PRs. 😅

ivoanjo:
Yes, definitely fix CI first haha I was kinda assuming that as a given, I'm not that crazy ;) ;)

dbenamydd:
Guess this is stale :-) Please reopen if not!

PR #214, opened 2023-08-12
refactor(profiling): rework internal types

morrisonlevi:
# What does this PR do?

This continues pulling out things from `profiling/src/profiling/mod.rs` and puts them into `profiling/src/profile/internal/*`. Additionally, it introduces traits for a few concepts that were already there, just unnamed (and not tidy):

- `Id`, e.g. FunctionId, MappingId, LocationId, StackTraceId, etc.
- `Item`, e.g. Function, Mapping, Location, StackTrace, etc.
- `PprofItem`, e.g. Function, Mapping, Location, etc but not StackTrace.

The `DedupExt` trait was simplified, which is partly possible because I understand Rust better now, but also because of the new traits. It enforces the max-size and will panic if the containers get too big. Previously, only mappings checked for being full, which is a bit ironic because it's the least likely in practice to overflow.

Hopefully this explains why it's such a large diff: all these things are connected, although I probably snuck in a change or two that didn't strictly have to be in there (it's hard when it's this big to notice).

# Motivation

The goals are to increase cleanliness and separation between the different representations of profiles. We now have `api::*`, `internal::*`, and `pprof::*`, which are fairly contained and separated. There should probably be more cleaning up of the `internal::*` parts, but this PR is big enough already.

# Additional Notes

I would start the review by looking at `profiling/src/profile/internal/mod.rs`, and  reading the 3 trait definitions first.

# How to test the change?

Since this targets the internal representation, existing tests ought to be good enough.


profiling/src/profile/internal/mapping.rs
```diff
@@ -0,0 +1,69 @@
...
+use super::{Id, Item, PprofItem};
+use crate::profile::{pprof, StringId};
+use std::num::NonZeroU32;
+
+#[derive(Eq, PartialEq, Hash)]
+pub struct Mapping {
+    /// Address at which the binary (or DLL) is loaded into memory.
+    pub memory_start: u64,
+    /// The limit of the address range occupied by this mapping.
+    pub memory_limit: u64,
```
sanchda, line 13: As it stands now, we're stuck with this; but if we change from a `[start, end)` formulation to a `[start, start + size)` formulation, we could realistically squeeze the `size` parameter into a u32.  These mappings are actually ELF sections (otherwise including both start + file offset is kinda weird), and they probably need to be _executable_ unless we're providing location information for non-code resources (why lol), soooo I personally think u32 would be reasonable there.

Obviously this would be inappropriate for this PR, just making a comment.

profiling/src/profile/internal/mapping.rs
```diff
@@ -0,0 +1,69 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+use super::{Id, Item, PprofItem};
+use crate::profile::{pprof, StringId};
+use std::num::NonZeroU32;
+
+#[derive(Eq, PartialEq, Hash)]
+pub struct Mapping {
+    /// Address at which the binary (or DLL) is loaded into memory.
+    pub memory_start: u64,
```
sanchda, line 11: The PR description suggested something about squeezing this into a u32, but I think it needs to be u64.  Realistically you can bound userspace addresses to some range on the order of 2^47 or whatever, but I _think_ we'd really like to add kernel profiling for the native profiler, which would boost us back up to u64.

profiling/src/profile/internal/mapping.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// The limit of the address range occupied by this mapping.
+    pub memory_limit: u64,
+    /// Offset in the binary that corresponds to the first mapped address.
+    pub file_offset: u64,
```
sanchda, line 15: Realistically, I think this can be bound to a u32.  This combined with my other comment would shave a byte off of this, but it would require a refactor at serialization.

profiling/src/profile/internal/label.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
+pub struct Label {
+    key: StringId,
+    value: LabelValue,
+}
+
+impl Label {
+    pub fn has_num_value(&self) -> bool {
```
danielsn, line 22: This could also be `get_num_value(&self) -> Option<(i64, Option<StringId>>`, to allow easy destructuring

profiling/src/profile/internal/value_type.rs
```diff
@@ -0,0 +1,25 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+use super::super::{pprof, Id, StringId};
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct ValueType {
+    pub r#type: StringId,
```
danielsn, line 8: Since we're making our own struct, should we call this `typ` instead?

profiling/src/profile/mod.rs
```diff
@@ -438,62 +274,21 @@ impl<'a> ProfileBuilder<'a> {
...
-    }
-}
-
-#[derive(Debug)]
-pub struct FullError;
-
-impl fmt::Display for FullError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "Full")
+        <T as Item>::Id::from_offset(id)
```
danielsn, line 288: nice

profiling/src/profile/mod.rs
```diff
@@ -438,62 +274,21 @@ impl<'a> ProfileBuilder<'a> {
...
-    where
-        T: Eq + Hash + From<&'a Q> + Borrow<Q>,
-        Q: Eq + Hash + ?Sized;
-}
-
-impl<T: Sized + Hash + Eq> DedupExt<T> for FxIndexSet<T> {
-    fn dedup(&mut self, item: T) -> usize {
+impl<T: Item> Dedup<T> for FxIndexSet<T> {
+    fn dedup(&mut self, item: T) -> <T as Item>::Id {
         let (id, _) = self.insert_full(item);
```
danielsn, line 287: Do the same assertion as `intern` does

profiling/src/profile/mod.rs
```diff
@@ -555,65 +350,83 @@ impl Profile {
     }
 
     /// Interns the `str` as a string, returning the id in the string table.
-    fn intern(&mut self, str: &str) -> StringId {
-        // strings are special because the empty string is actually allowed at
-        // index 0; most other 0's are reserved and cannot exist
-        StringId::new(self.strings.dedup_ref(str))
+    fn intern(&mut self, item: &str) -> StringId {
```
danielsn, line 353: I was going to ask why we need this given that its a repeat of the `dedup` logic. But I assume this is because we don't want to have to convert `&str` to `String` for the `insert_full`?

profiling/src/profile/mod.rs
```diff
@@ -629,28 +442,8 @@ impl Profile {
 
         let mut locations = Vec::with_capacity(sample.locations.len());
```
danielsn, line 443:
```suggestion
        let locations : Vec<_> = sample.locations.iter().map(|l| self.add_location(l)).collect();
```

profiling/src/profile/internal/function.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    fn from_offset(v: usize) -> Self {
+        let index: u32 = v.try_into().expect("FunctionId to fit into a u32");
+
+        // PProf reserves function 0.
+        // Both this, and the serialization of the table, add 1 to avoid the 0 element
+        let index = index.checked_add(1).expect("FunctionId to fit into a u32");
+        // Safety: the `checked_add(1).expect(...)` guards this from ever being zero.
+        let index = unsafe { NonZeroU32::new_unchecked(index) };
+        Self(index)
```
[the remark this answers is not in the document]
morrisonlevi: I tried to make a proper abstraction for it that doesn't have tons of boiler-plate and I failed. So, I'll just make a helper function ^_^

profiling/src/profile/internal/line.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use crate::profile::{pprof, FunctionId, Id};
+use std::fmt::Debug;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
+pub struct Line {
+    /// The id of the corresponding Function for this line.
+    pub function_id: FunctionId,
+    /// Line number in source code.
+    pub line: i64,
```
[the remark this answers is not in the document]
morrisonlevi: Yes, think generated C code files that are basically the entire program in one file.

profiling/src/profile/internal/stack_trace.rs
```diff
@@ -0,0 +1,39 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+use super::{Id, Item, LocationId};
+use std::fmt::{Display, Formatter};
+
+#[derive(Eq, PartialEq, Hash)]
+pub struct StackTrace {
+    /// The ids recorded here correspond to a Profile.location.id.
+    /// The leaf is at location_id[0].
+    pub locations: Vec<LocationId>,
```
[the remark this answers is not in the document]
morrisonlevi: Let's save this kind of thing for another PR, this one is big enough! But good idea.

profiling/src/profile/internal/stack_trace.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl Item for StackTrace {
+    type Id = StackTraceId;
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+#[repr(transparent)]
+pub struct StackTraceId(u32);
+
+impl Display for StackTraceId {
```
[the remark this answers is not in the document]
morrisonlevi: We don't! Good catch. I expect at some point during my refactoring it was needed, but not anymore. Deleted.

profiling/src/profile/internal/label.rs
```diff
@@ -0,0 +1,77 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+use super::super::{pprof, Id, StringId};
```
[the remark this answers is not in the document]
morrisonlevi: Both are horrible, IMO. Maybe we should pull in those things into `internal/mod.rs` so they can be just be `super::{...}` and not `super::super::{...}`?

profiling/src/profile/mod.rs
morrisonlevi: I don't think it makes sense to in this case: if it already exists, that's fine. The string case is different because it can't do a simple `insert_full` because it has to deal with `&str` vs `String` differences.

profiling/src/profile/mod.rs
morrisonlevi: Yes, we don't want to convert `&str` into `String` unless we have to -- that's a good chunk of overhead that we can avoid for cases where it already exists in the set.

profiling/src/profile/mod.rs
morrisonlevi: Done!

profiling/src/profile/internal/line.rs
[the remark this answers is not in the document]
ivoanjo: Do we care enough about such use-cases? I'd probably take the memory savings win vs supporting accurate line-numbers for a program with more than 2^31/2^32 lines :P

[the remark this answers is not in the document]
morrisonlevi: We're getting rid of `Line` in another PR, and in that PR, shrinking the size of the `line` won't save any memory because of padding bytes anyway.

profiling/src/profile/mod.rs
```diff
@@ -686,20 +487,22 @@ impl Profile {
         let mut labels: Vec<Label> = Vec::with_capacity(sample.labels.len());
         let mut local_root_span_id_label_offset: Option<usize> = None;
         for label in sample.labels.iter() {
+            anyhow::ensure!(
+                label.str.is_none() || label.num == 0,
+                "Invalid label: {:?}",
```
danielsn, line 492: Do we want to explain in the message why the label is invalid?

profiling/src/profile/mod.rs
```diff
@@ -861,30 +665,39 @@ impl Profile {
...
+            "the local root span id label value must be sent as a number, not a string, given {:?}",
+            label
         );
 
-        /* Safety: the value is a u64, but pprof only has signed values, so we
-         * transmute it; the backend does the same.
-         */
-        let local_root_span_id: u64 = unsafe { std::intrinsics::transmute(label.num) };
+        let local_root_span_id: u64 = if let LabelValue::Num { num, .. } = label.get_value() {
+            // Manually specify the type here to be sure we're transmuting an
```
danielsn, line 691: 👍

danielsn (dismissed):
A couple minor ⛏️ and comments but LGTM

profiling/src/profile/internal/value_type.rs
ivoanjo: (I do prefer `type` personally :P )

profiling/src/profile/mod.rs
ivoanjo: (Should we add this as a comment on the code? Seems like a question that will come up when others look at this code :P )

profiling/src/profile/mod.rs
ivoanjo: +1 yes please! I left some suggestions for this on PR https://github.com/DataDog/libdatadog/pull/205

ivoanjo (approved):
Left a few notes, but  overall LGTM :+1:

profiling/src/profile/internal/label.rs
```diff
@@ -0,0 +1,77 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+use super::*;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
+pub enum LabelValue {
+    Str(StringId),
+    Num {
+        num: i64,
+        num_unit: Option<StringId>,
```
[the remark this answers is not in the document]
danielsn: If we need to, we could give an API allowing clients to specify a `key, num_unit` mapping, and then insert the `num_unit` to the pprof based on that mapping.  Keeps the feature, but saves memory

profiling/src/profile/mod.rs
morrisonlevi: I think we can combine both of the ensures into a single one:

```rust
            anyhow::ensure!(
                label.str.is_none() || (label.num == 0 && label.num_unit.is_none()),
                "Invalid label: used both str and num fields: {label:?}"
            );
```

If the `label.str` is none, then it doesn't matter what values are held in `label.num` and `label.num_unit`. If the `label.str` is something, on the other hand, then both `label.num` and `label.num_unit` should be zero-representations. Right?

morrisonlevi: I added some tests to double-check this ^_^ They seem equivalent.

profiling/src/profile/mod.rs
morrisonlevi: I added some commentary in the code 👍🏻

profiling/src/profile/internal/function.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use super::*;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
+pub struct Function {
+    pub name: StringId,
+    pub system_name: StringId,
+    pub filename: StringId,
+    pub start_line: i64,
+}
```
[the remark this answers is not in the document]
morrisonlevi: Done.

profiling/src/profile/internal/mapping.rs
morrisonlevi: What about needing to be _executable_ makes it reasonable to expect it to fit in a u32?

profiling/src/profile/internal/function.rs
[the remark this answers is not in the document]
morrisonlevi: The helper function is in `internal/mod.rs` and is called `small_non_zero_pprof_id`.

profiling/src/profile/internal/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+pub trait Item: Eq + Hash {
+    /// The Id associated with this Item, e.g. Function -> FunctionId.
+    type Id: Id;
+}
+
+/// Used to associate an Item with a pprof::* type. Not all Items can be
+/// converted to pprof::* types. For example, StackTrace doesn't have an
+/// associated pprof::* type.
+pub trait PprofItem: Item {
```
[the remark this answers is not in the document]
morrisonlevi: I added a comment above the `to_pprof` function.

danielsn (approved):
🚢

profiling/src/profile/mod.rs
ivoanjo: Seems good to me! The negations tripped me a bit so I had to pause a bit to mentally parse it, but the logic looks very correct!

PR #215, opened 2023-08-14
perf!: make profiling Location contain a function

morrisonlevi:
# What does this PR do?

Previously, the different representations contained some sort of vec or slice of `Line`. This PR removes this abstraction, and `Location` now directly contains a `Function` or `FunctionId` as appropriate for the representation, as well as the line number.

This also removes `is_folded`, as this is unused and just wasting bytes.

# Motivation

This shaves bytes off the `Location` struct (16 compared to main, 24 compared to the last release), and the extra allocation used for `Vec<Line>` is also removed. These are the primary motivations. I'll do some napkin math below. This also allows for `Location` to be `Copy`, which is a nice little UX perk as well.

I've copy/pasted the definitions for internal representations. Before:

```rust
#[derive(Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
pub struct Location {
    pub mapping_id: MappingId,
    pub address: u64,
    pub lines: Vec<Line>,
    pub is_folded: bool,
}

#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
pub struct Line {
    pub function_id: FunctionId,
    pub line: i64,
}
```

After:

```rust
#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord)]
pub struct Location {
    pub mapping_id: MappingId,
    pub function_id: FunctionId,
    pub address: u64,
    pub line: i64,
}
```

For a small profile with 9521 samples in it, and an average of 2 locations per sample:

9521 samples * 16 bytes saved per location * 2 average locations per sample = 457KiB saved in-memory.

Similarly, this saves 9521 * 1 allocations per Location * 2 average locations per sample = 19,042 allocations.

# Additional Notes

This is a breaking change. Profiling libraries will need to update their code to create new structs as appropriate.

This builds on a previous PR, and shouldn't be merged until that one is merged.

# How to test the change?

Regular tests apply, once you have fixed your struct initializations.


ivoanjo (commented):
+1 I like this. Ruby indeed even pre-allocates the locations like this:

```
  buffer->locations     = ruby_xcalloc(max_frames, sizeof(ddog_prof_Location));
  buffer->lines         = ruby_xcalloc(max_frames, sizeof(ddog_prof_Line));

  // Currently we have a 1-to-1 correspondence between lines and locations, so we just initialize the locations once
  // here and then only mutate the contents of the lines.
  for (unsigned int i = 0; i < max_frames; i++) {
    ddog_prof_Slice_Line lines = (ddog_prof_Slice_Line) {.ptr = &buffer->lines[i], .len = 1};
    buffer->locations[i] = (ddog_prof_Location) {.lines = lines};
  }
```

And then I only mutate the lines. We don't set `is_folded` to anything either :)

profiling/src/profile/pprof.rs
```diff
@@ -117,7 +117,7 @@ pub struct Mapping {
     #[prost(bool, tag = "8")]
     pub has_filenames: bool,
     #[prost(bool, tag = "9")]
-    pub has_line_numbers: bool,
+    pub has_lines: bool,
     #[prost(bool, tag = "10")]
```
ivoanjo, line 121: ‼ This change seems incorrect, since the pprof format shouldn't change itself...? (Maybe too much search&replace?)

profiling/src/profile/profile.proto
```diff
@@ -152,7 +152,7 @@ message Mapping {
   // The following fields indicate the resolution of symbolic info.
   bool has_functions = 7;
   bool has_filenames = 8;
-  bool has_line_numbers = 9;
+  bool has_lines = 9;
```
ivoanjo, line 155: ‼ This also seems incorrect (same note as `pprof.rs`)

ivoanjo (commented):
Found a couple of things that seem off! Still need to finish reviewing the other files :)

profiling/src/profile/api.rs
```diff
@@ -240,32 +226,29 @@ fn function_fetch(pprof: &pprof::Profile, id: u64) -> anyhow::Result<Function> {
...
-    }
-    Ok(output)
-}
-
 fn location_fetch(pprof: &pprof::Profile, id: u64) -> anyhow::Result<Location> {
     if id == 0 {
         return Ok(Location::default());
     }
 
     match pprof.locations.iter().find(|item| item.id == id) {
```
ivoanjo, line 234: I was a bit confused what this code was doing (and that iterating locations to find an id seems weird since supposedly the ids are sequential and whatnot), but then I realized that this is only used for `impl<'a> TryFrom<&'a pprof::Profile> for Profile<'a>`, which is definitely not something we use outside of tests (...right?)

Would it be worth adding a `#[cfg(test)]` around most of these methods, as a way of documenting "this is not production code"?

profiling/src/profile/api.rs
```diff
@@ -240,32 +226,29 @@ fn function_fetch(pprof: &pprof::Profile, id: u64) -> anyhow::Result<Function> {
...
-        Some(location) => Ok(Location {
-            mapping: mapping_fetch(pprof, location.mapping_id)?,
-            address: location.address,
-            lines: lines_fetch(pprof, &location.lines)?,
-            is_folded: location.is_folded,
-        }),
+        Some(location) => {
+            anyhow::ensure!(!location.is_folded, "expected Location to not be folded");
+            anyhow::ensure!(
+                location.lines.len() == 1,
```
ivoanjo, line 238: I noticed that this is actually incorrect, although it has been incorrect for a long time.

Looking at the pprof format ( https://github.com/DataDog/libdatadog/blob/a351844ed09611d01e853c9cf476291f519bc9e7/profiling/src/profile/profile.proto#L181 ) or even [Felix's great pic](https://github.com/DataDog/go-profiler-notes/blob/main/pprof.md): 

https://github.com/DataDog/go-profiler-notes/blob/main/pprof.md 

...there's no `lines` entry inside pprof (it's `line`). Our `pprof.rs` has incorrectly diverged from the `pprof` definition.

I'm guessing this hasn't caused issues because pprof doesn't encode the field names (?), so because we added at the same location the same field but with a different name, I think the resulting pprof is still equivalent.

So... we can probably leave it to a separate PR but it's worth maybe going back to the upstream `profile.proto` and make sure our copy is 1:1, and then pprof.rs is also 1:1 with it :)

ivoanjo (approved):
Left a couple more notes, but overall it LGTM :+1:

profiling/src/profile/api.rs
morrisonlevi: I think `Location.line` does exist, and I just named it `lines`. The name in our source code doesn't matter, as long as the type and order are the same.

profiling/src/profile/api.rs
morrisonlevi: Prior to this R&D week, I would have said that's fine, but the new profiling replayer added during R&D week uses this function ^_^

But we could a note of some kind.

profiling/src/profile/api.rs
ivoanjo: Yeah, definitely ok to not change it in this PR.

> But, if you want to have exactly the same names in our source code as in the pprof

Is it too nitpicky? It seems... weird to me that in the code that is implementing pprof we're not using the names that pprof uses.

It was definitely confusing to me -- I was referencing the pprof drawing above when understanding the code and the different name made me think I was misunderstanding the code and that I was looking at our own internal representation (which can use `lines`) vs the pprof one (which I was sure didn't use `lines`). I even wondered if perhaps you had made a typo and was the code compiling (maybe it wasn't?)? So I checked it out locally, verified it compiled, checked my assumptions, and that's when I found the name was changed.

So... Yeah, I was really led astray by the off name, hence why I suggested fixing it.

profiling/src/profile/api.rs
ivoanjo: Yeah, I think it'd helpful to mark this code as "not used/should not be used in production"; as that'd make it clear that improvements here are more of a "maybe some day" vs "woah maybe we're missing something".

ivoanjo (approved):
:+1: Re-approving to fix dismissed approval

PR #217, opened 2023-08-15
profiler: use Box<[i64]> instead of Vec<i64> to store observations, saving 8 bytes per sample

danielsn:
# What does this PR do?

Uses a new `Observation` type, instead of `Vec<i64>` to store observations on a sample.

# Motivation

Currently, we use `Vec<i64>` to store observations on a sample.  This is inefficient, as samples always have exactly the same number of values in the vec. Instead, using a boxed slice saves 8 bytes for each sample by not storing the unneeded `capacity` field.

Napkin math: 
For a small profile with 9521 samples in it, and an average of 2 locations per sample:

9521 samples * 8 bytes saved per location  = 76KiB saved in-memory.

# Additional Notes

I am working on a more involved change to also avoid the need to store the `len` field in the slice. But that requires `unsafe` and is risky, and this is a quick free win.

We might also want to store this data using a bump allocator, to get more efficient allocations.  Again, for a future PR

# How to test the change?


ivoanjo (approved):
:+1: LGTM

I'm not very confident on the rust-foo, but the changes seem reasonable, and I can confirm the memory improvement:

Using the `ruby_overhead_experiment.rb`:

Before:

```
$ TIMELINE=false bundle exec ruby ruby_overhead_experiment.rb 

Before serialization, process rss usage is 70244k
After serialization, process rss usage is 70404k
After shrinking, process rss usage is 70404k

$ TIMELINE=true bundle exec ruby ruby_overhead_experiment.rb 
Before serialization, process rss usage is 99112k
After serialization, process rss usage is 147140k
Size of uncompressed serialized pprof: 9031k
After shrinking, process rss usage is 98584k
``` 

After:

```
$ TIMELINE=false bundle exec ruby ruby_overhead_experiment.rb
Before serialization, process rss usage is 70716k
After serialization, process rss usage is 70876k
After shrinking, process rss usage is 70876k

$ TIMELINE=true bundle exec ruby ruby_overhead_experiment.rb
Before serialization, process rss usage is 97952k
After serialization, process rss usage is 146912k
Size of uncompressed serialized pprof: 9031k
After shrinking, process rss usage is 146912k
```

Without timeline, most of the samples get merged in this benchmark, so the small raise seems within the margin of error.

Using `examples/ffi/profiles`:

Before:

```
$ /bin/time -v ./profiles
Press any key to exit...
  Maximum resident set size (kbytes): 1712856
```

After:

```
$ /bin/time -v ./profiles
Press any key to exit...
  Maximum resident set size (kbytes): 1634600
```

ivoanjo (approved):
:+1: LGTM

danielsn:
Merged this change into #210 this seperate PR is no longer needed

PR #218, opened 2023-08-15
[draft] profiler: Don't store the unnecessary length field for Sample Observations

danielsn:
# What does this PR do?

Replaces the 24 byte `Vec<i64>` used by each sample with an 8 byte pointer.

# Motivation

Each sample has a vec of observations, which has both a capacity and a length member, both of which are 8 bytes.  This is wasteful, because all observations for a Profile are of the same length.  Instead, we only actually need a single pointer, and can store the length off to the side.

# Additional Notes

This has `unsafe` code and could use an extra careful review

The current implementation has a issue when multiple profiles are created, with different Sample Observation types. This happens for e.g. in our unit tests.  I'm trying to see if there is a clean solution.
# How to test the change?

### DIFF
--- a/profiling/src/profile/internal/mod.rs
+++ b/profiling/src/profile/internal/mod.rs
@@ -1,0 +1,5 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+
+mod observation;
+pub use observation::*;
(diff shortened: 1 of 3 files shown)


realFlowControl:
> The current implementation has a issue when multiple profiles are created, with different Sample Observation types.

Just to recap: you mean as in we create a profile with CPU and wall time and then another one with also has allocations and have both on memory at the same time?
Because this could happen in PHP

ivoanjo:
+1 I really like the improvement, but not having it per-profiler instance seems like a big potential footgun :sweat_smile:

PR #219, opened 2023-08-17
profiler: Don't store the unnecessary length field for Sample Observations

danielsn:
# What does this PR do?

Replaces the 24 byte `Vec<i64>` used by each sample with an 8 byte pointer.

# Motivation

Each sample has a vec of observations, which has both a capacity and a length member, both of which are 8 bytes.  This is wasteful, because all observations for a Profile are of the same length.  Instead, we only actually need a single pointer, and can store the length off to the side.

# Additional Notes

This has `unsafe` code and could use an extra careful review.

I realized that the correct abstraction layer for storing the length field is the `Map`.  All we do is put stuff into the map, and then iterate over it later. This means that each map carries its own `ObservationLength` along with it, allowing multiple profiles to coexist happily.

I still need to add some tests. But making this a draft PR so we can discuss the approach while I work on the tests.

# How to test the change?


profiling/src/profile/mod.rs
morrisonlevi, line 1: Is it possible to slap a `cfg(test)` on things like `only_for_testing_num_aggregated_samples`?

morrisonlevi (dismissed):

morrisonlevi (approved):

profiling/src/profile/mod.rs
morrisonlevi: Nope -- the `profiling-ffi` crate uses it in its tests, and when that crate is built, the dependency on `profiling` does not get built with `cfg(test)`.

PR #223, opened 2023-08-18
refactor: Move profiler internal types to the internal module

danielsn:
# What does this PR do?

Moves internal datastructures out of the `profile/mod.rs` file into `profile/internal/<structure_name>.rs`.

# Motivation

This work was partially completed in #214, but some types were missed.  Collecting them too.
# Additional Notes

# How to test the change?


morrisonlevi (approved):
I didn't bother to compare source to source, I assume the only changes are to make imports work?

danielsn:
> I didn't bother to compare source to source, I assume the only changes are to make imports work?

The only small change was to move `upscale_values` to be a member function on `UpscalingRules` rather than on the `Profile`, because it fits more naturally there (a bunch of calls `self.upscaling_rules.foo` became `self.foo`)

PR #225, opened 2023-08-22
Add cargo-miri github action

danielsn:
# What does this PR do?

Adds a `cargo miri` github action to our CI.

# Motivation

I added some `unsafe` code in #219, which I verified using `miri`.  But, this doesn't run in CI, so it could break without us knowing.  Adding it to CI increases safety. 

# Additional Notes

Miri is slower than regular test.  Currently, I'm only running this on the `unsafe` changes I made in #219.  We might want to consider expanding the filter to cover more code.

# How to test the change?


.github/workflows/miri.yml
```diff
@@ -0,0 +1,21 @@
+name: Miri test
```
ivoanjo, line 1: Since right now we're scoped to `profile::internal::observation`, I would suggest maybe including that here on the name (to avoid a misleading sense of safety -- oh we're running miri on stuff)?

ivoanjo (approved):
I'm curious -- how slow is miri? Although annoying, I think a CI run that takes a few minutes is probably fine (or we could add another job that runs full miri, but only every night and not on every PR).

One other thing we can make use of is a faster machine -- here's a PR where we did that for dd-trace-go to improve performance of CI stuff ( https://github.com/DataDog/dd-trace-go/pull/2068 ) .

None of this is blocking for this PR, but I'm excited about the possibilities :)

.github/workflows/miri.yml
danielsn: Done

ivoanjo (approved):
:+1: LGTM

PR #227, opened 2023-08-22
perf(profiling): store string data in an arena allocator

morrisonlevi:
# What does this PR do?

This uses `bumpalo::Bump` as an arena allocator to store the string data contiguously with fewer calls to the system allocator.

Since the `StringTable` owns the arena and stores references to that data inside other members of the `StringTable`, it is a self-referencing data structure, which is generally unsafe. This uses `ouroboros::self_referencing` to have a safe abstraction for this pattern.

# Motivation

This produces a better memory layout with fewer system allocations. It likely reduces the total memory used, and it _may_ improve the amount of memory returned to the OS when arena chunks are dropped (see below).

# Additional Notes

The initial size of the arena is `16384` bytes aka 16KiB. It reserves a few bytes for itself. When it's full, it creates another chunk that seems at least as big as the previous chunk. This relieves pressure on the system allocator with string-by-string allocation, and reduces fragmentation.

This `StringTable` is copied and simplified from the PHP profiler. I intend for the PHP profiler to use this version of the `StringTable` going forward.

# How to test the change?

This is an internal change only. Nothing special needs to be done for testing.

### DIFF
--- a/profiling/src/collections/identifiable/string_table.rs
+++ b/profiling/src/collections/identifiable/string_table.rs
@@ -1,0 +1,266 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+use super::*;
+use bumpalo::Bump;
+use ouroboros::self_referencing;
+use std::collections::HashMap;
+use std::hash::BuildHasherDefault;
+
+#[cfg(test)]
+use std::ops::Range;
+
+struct BorrowedStringTable<'b> {
+    /// The arena to store the characters in.
+    arena: &'b Bump,
+
+    /// Used to have efficient lookup by [StringId], and to provide an
+    /// [Iterator] over the strings.
+    vec: Vec<&'b str>,
+
+    /// Used to have efficient lookup by [&str].
+    map: HashMap<&'b str, StringId, BuildHasherDefault<rustc_hash::FxHasher>>,
+}
+
+#[self_referencing]
+struct StringTableCell {
+    // This arena holds the characters of the strings. The memory used to hold
+    // the vec and map to implement the lookups required are not in the arena,
+    // because they would be re-allocated over time but the previous data
+    // wouldn't generally get reclaimed. This makes them a poor fit for an
+    // arena allocator.
+    owner: Bump,
+
+    // This says that the BorrowedStringTable will hold a reference to the
+    // field `owner`. Rust does not allow this as there are many ways to make
+    // this code unsafe. The ouroboros crate is used to provide a safe
+    // abstraction for a subset of self-referential behavior.
+    #[borrows(owner)]
+    #[covariant]
+    dependent: BorrowedStringTable<'this>,
+}
+
+/// The [StringTable] stores strings and associates them with [StringId]s,
+/// which correspond to the order in which strings were inserted. The empty
+/// string is always associated with [StringId::ZERO].
+pub struct StringTable {
+    // ouroboros will add a lot of functions to this struct, which we don't
+    // want to expose publicly, so the internals are wrapped and private.
+    inner: StringTableCell,
+}
+
+impl StringTable {
+    // Not guaranteed for a given system, but a very common size.
+    const PAGE_SIZE: usize = 4096;
+
+    // not guaranteed, has a test to double-check.
+    const BUMP_OVERHEAD: usize = 64;
+
+    /// A good initial capacity for the [StringTable]. Used by
+    /// `[StringTable::new]` and its default impl.
+    ///
+    /// Ideally, we'd want the system allocator to allocate X pages. Pages are
+    /// the granularity that operating systems typically give out memory,
+    /// although some are definitely more efficient with larger number of
+    /// pages, and some have a concept of huge pages. The point is, it's hard
+    /// to totally generalize.
+    ///
+    /// However, we're not using a system allocator directly, we're using
+    /// [Bump], which has some overhead. So we need to subtract off some
+    /// overhead of whatever number we want, to avoid going into the next
+    /// size. The good news is that it rounds up to page sizes of 4096, so we
+    /// only need to get close.
+    ///
+    /// So, what remains is choosing how many pages to reserve up front. From
+    /// one perspective, it would be nice to use a size that goes directly to
+    /// `mmap` on Linux by a given malloc implementation, but if a certain
+    /// number of profiles don't reach that number, than that's wasteful. We
+    /// don't currently have metrics on this.
+    ///
+    /// So... for now, the selected number of pages is arbitrarily chosen.
+    pub const GOOD_INITIAL_CAPACITY: usize = 8 * Self::PAGE_SIZE - Self::BUMP_OVERHEAD;
+
+    #[inline]
+    pub fn new() -> Self {
+        Self::with_arena_capacity(Self::GOOD_INITIAL_CAPACITY)
+    }
+
+    /// Creates a new [StringTable] with an arena capacity of at least
+    /// `capacity` in bytes. Keep in mind that the other structures will also
+    /// use memory that is not included in this capacity.
+    #[inline]
+    fn with_arena_capacity(capacity: usize) -> Self {
+        let arena = Bump::with_capacity(capacity);
+        let inner = StringTableCell::new(arena, |arena| BorrowedStringTable {
+            arena,
+            vec: Default::default(),
+            map: Default::default(),
+        });
+
+        let mut s = Self { inner };
+        // string tables always have the empty string at 0.
+        let (_id, _inserted) = s.insert_full("");
+        debug_assert!(_id == StringId::ZERO);
+        debug_assert!(_inserted);
+        s
+    }
+
+    #[allow(unused)]
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.inner.with_dependent(|table| table.vec.len())
+    }
+
+    #[allow(unused)]
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Inserts the string into the table, if it did not already exist. The id
+    /// of the string is returned.
+    ///
+    /// # Panics
+    /// Panics if a new string needs to be inserted but the offset of the new
+    /// string doesn't fit into a [StringId].
+    #[inline]
+    pub fn insert(&mut self, str: &str) -> StringId {
+        self.insert_full(str).0
+    }
+
+    /// Inserts the string into the table, if it did not already exist. The id
+    /// of the string is returned, along with whether the string was inserted.
+    ///
+    /// # Panics
+    /// Panics if a new string needs to be inserted but the offset of the new
+    /// string doesn't fit into a [StringId].
+    #[inline]
+    pub fn insert_full(&mut self, str: &str) -> (StringId, bool) {
+        // For performance, delay converting the &str to a String until after
+        // it has been determined to not exist in the set. This avoids
+        // temporary allocations.
+        self.inner
+            .with_dependent_mut(|table| match table.map.get(str) {
+                None => {
+                    let id = StringId::from_offset(table.vec.len());
+                    let bumped_str = table.arena.alloc_str(str);
+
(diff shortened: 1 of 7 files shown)


profiling/src/profile/internal/string_table.rs
```diff
@@ -0,0 +1,227 @@
...
+pub struct StringTable {
+    // ouroboros will add a lot of functions to this struct, which we don't
+    // want to expose publicly, so the internals are wrapped and private.
+    inner: StringTableCell,
+}
+
+impl StringTable {
+    #[inline]
+    pub fn new() -> Self {
+        Self::with_capacity(4000)
```
danielsn, line 55: why 4000?

profiling/src/profile/internal/string_table.rs
```diff
@@ -0,0 +1,227 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// If this fails, bumpalo may have changed its allocation patterns, and
+    /// [StringTable::new] may need adjusted.
+    #[test]
+    fn test_bump() {
+        let arena = Bump::with_capacity(4000);
+        assert_eq!(4096 - 64, arena.chunk_capacity());
```
danielsn, line 168: Where does the 64 come from? Is this guaranteed not to change when Bumpalo is upgraded?

profiling/src/profile/internal/string_table.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+impl StringTable {
+    #[inline]
+    pub fn new() -> Self {
+        Self::with_capacity(4000)
+    }
+
+    #[inline]
+    pub fn with_capacity(capacity: usize) -> Self {
```
danielsn, line 59: Is this bytes? Document this

danielsn (approved):
LGTM.

profiling/src/profile/internal/string_table.rs
morrisonlevi: There's no guarantee, which is why there's a test -- if we upgrade bumpalo and it's not what we expect, we'll know.

As for the magic numbers, they come from internal details. I guess I could link this:
https://docs.rs/bumpalo/latest/src/bumpalo/lib.rs.html#435-444

```rs
// Maximum typical overhead per allocation imposed by allocators.
const MALLOC_OVERHEAD: usize = 16;

// This is the overhead from malloc, footer and alignment. For instance, if
// we want to request a chunk of memory that has at least X bytes usable for
// allocations (where X is aligned to CHUNK_ALIGN), then we expect that the
// after adding a footer, malloc overhead and alignment, the chunk of memory
// the allocator actually sets aside for us is X+OVERHEAD rounded up to the
// nearest suitable size boundary.
const OVERHEAD: usize = (MALLOC_OVERHEAD + FOOTER_SIZE + (CHUNK_ALIGN - 1)) & !(CHUNK_ALIGN - 1);
```

profiling/src/profile/internal/string_table.rs
morrisonlevi: The goal is to get the system allocator to use large, good allocations. I could see this being `4096`, but due to the overhead, it will actually allocate something larger like 8192 - 64. Not sure if that's better, could be.

morrisonlevi: At least on some glibc versions (search `man mallopt`), the point malloc starts using `mmap` is 128 * 1024 bytes (128KiB). So if we knew some statistics about how large string tables are in real-world profiles, grouped by language, we could possibly bump this to be near this limit and see improved performance, but also the memory will definitely get returned to the OS.

I guess I could start by analyzing a symfony-demo profile, this gives us a floor (not special, on the small side for PHP profiles almost assuredly).

morrisonlevi: A symfony-demo profile I analyzed came out with 1611 strings, which used at least 101,342 bytes of memory (for the string data, not any other kind of overhead in the vecs themselves or anything).

morrisonlevi: I documented many things and bumped this number to 8 pages, or 32,768 bytes. Feel free to argue for something different, but without metrics from intake or something, it's hard to say what it ought to be.

profiling/src/profile/internal/string_table.rs
morrisonlevi: Renamed and documented.

ivoanjo:
I've re-ran some of the experiments I mentioned in the memory usage RFC doc:

#### `examples/ffi/profiles.c`:

| libdatadog version                 | RSS Before drop (task manager) | RSS After drop (task manager) | Maximum RSS |
|------------------------------------|--------------------------------|-------------------------------|-------------|
| v3.0.0                             | 2.2 GB                         | 1.1 GB                        | 2181076 KB  |
| main@62e329fd4ddb7a34 (2023-08-23) | 1.7 GB                         | 640.2 MB                      | 1747392 KB  |
| levi/string-table                  | 1.8 GB                         | 640.3 MB                      | 1748420 KB  |

#### `ruby_overhead_experiment.rb`:

(`config: {:threads=>16, :depth=>50, :seconds=>60, :timeline_enabled=>true}`)

| libdatadog version                 | Before serialization rss | After serialization rss | After GC rss ('shrinking', in the script) |
|------------------------------------|--------------------------|-------------------------|--------------|
| main@62e329fd4ddb7a34 (2023-08-23) | 78484k                   | 134068k                 | 134068k      |
| levi/string-table                  | 78732k                   | 134352k                 | 85504k       |

---

I think overall the `profiles.c` benchmark doesn't gain a lot since it doesn't contain that many strings, so most of the overhead is elsewhere.

The Ruby benchmark does get a reduced rss after GC, which does seem to match our expectation that more memory is able to be returned to the OS (although there's still a lot hanging around -- we're still using up more rss than even before serialization; if we were releasing all libdatadog memory we would get below that number).

ivoanjo:
After yesterday's discussion of rss impact, I got curious on what we'd see for `examples/ffi/profiles.c` with tcmalloc and jemalloc, so here's the results:

| libdatadog version | RSS Before drop (task manager) | RSS After drop (task manager) | Maximum RSS |
|----------------------------|--------|----------|------------|
| levi/string-table glibc    | 1.8 GB | 640.3 MB | 1748420 KB |
| levi/string-table tcmalloc | 1.7 GB | 1.7 GB   | 1656732 KB |
| levi/string-table jemalloc | 1.3 GB | 165.9 MB | 1299504 KB |

profiling/src/internal/profile.rs
```diff
@@ -417,9 +404,10 @@ impl Profile {
             encoder.encode(ProfileFunctionsEntry::from(item))?;
         }
 
-        for item in self.strings.into_iter() {
+        for item in self.strings.iter() {
             encoder.encode(ProfileStringTableEntry::from(item))?;
         }
+        drop(self.strings);
```
danielsn, line 410: Does this drop the memory that is backing the `'a &str` in the `ProfileStringTableEntry`?

profiling/src/pprof/sliced_proto.rs
```diff
@@ -85,10 +88,52 @@ pub struct ProfileFunctionsEntry {
...
+
+impl<'a> ::prost::Message for ProfileStringTableEntry<'a> {
+    fn encode_raw<B>(&self, buf: &mut B)
+    where
+        B: BufMut,
+        Self: Sized,
+    {
+        let value = self.string_table_entry;
+        // See prost::encoding::string::encode
+        prost::encoding::encode_key(6, WireType::LengthDelimited, buf);
```
danielsn, line 105: `6` appears as a magic number a few times in this code

danielsn (commented):
Mostly looks good, one worry about possible use-after-free

profiling/src/internal/profile.rs
morrisonlevi: Miri didn't complain and I don't think so, but I haven't stepped through the code to be sure.

profiling/src/collections/identifiable/string_table.rs
```diff
@@ -0,0 +1,266 @@
...
+/// string is always associated with [StringId::ZERO].
+pub struct StringTable {
+    // ouroboros will add a lot of functions to this struct, which we don't
+    // want to expose publicly, so the internals are wrapped and private.
+    inner: StringTableCell,
+}
+
+impl StringTable {
+    // Not guaranteed for a given system, but a very common size.
+    const PAGE_SIZE: usize = 4096;
```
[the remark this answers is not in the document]
morrisonlevi: We could but it doesn't really matter because we're too far away in abstraction levels. Just really just documenting some of the choices of magic numbers.

profiling/src/internal/profile.rs
morrisonlevi: `encode` calls `encode_raw` which calls `put_slice` which does the copy, so I'm pretty confident this is fine.

profiling/src/collections/identifiable/string_table.rs
```diff
@@ -0,0 +1,266 @@
...
+#[cfg(test)]
+use std::ops::Range;
+
+struct BorrowedStringTable<'b> {
+    /// The arena to store the characters in.
+    arena: &'b Bump,
+
+    /// Used to have efficient lookup by [StringId], and to provide an
+    /// [Iterator] over the strings.
+    vec: Vec<&'b str>,
```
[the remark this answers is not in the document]
morrisonlevi: No, because the vector will resize. If we have a 32 element vec and it needs to resize to 64, the 32 elements are going to stay where they are and a new chunk of 64 elements will get allocated, and now there are 96 elements in the arena to store 64 items.

profiling/src/collections/identifiable/string_table.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    /// Used to have efficient lookup by [&str].
+    map: HashMap<&'b str, StringId, BuildHasherDefault<rustc_hash::FxHasher>>,
```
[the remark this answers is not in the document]
morrisonlevi: It looks like IndexMap may be slightly lighter in bytes, I'd have to check with some code. But in the next version (if we get there), there isn't a vec + map, only a map, because _the allocator storage is itself the vec_. So unless you care to, I'd just leave it for simplicity.

morrisonlevi:
On 50 runs of the replayer using the PHP timeline symfony-demo pprof, the average memory delta was:
 - `main`: 800.72
 - `levi/string-table`: 792.14

This seems within the noise tolerance of each other.

profiling/src/collections/identifiable/string_table.rs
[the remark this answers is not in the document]
danielsn: So this would make sense in a `mmap` world, but not in the current arena allocator

profiling/src/internal/profile.rs
```diff
@@ -301,9 +299,10 @@ impl Profile {
             encoder.encode(ProfileFunctionsEntry::from(item))?;
         }
 
-        for item in self.strings.into_iter() {
+        for item in self.strings.iter() {
```
danielsn, line 302: This is the one case where we don't `into_iter`.  Might be worth a comment explaining why

danielsn (approved):
One ⛏️ about adding a comment explaining why we iterate over strings differently than everything else.

profiling/src/collections/identifiable/string_table.rs
```diff
@@ -0,0 +1,266 @@
...
+    #[cfg(test)]
+    #[allow(unused)]
+    #[inline]
+    pub fn get_range(&self, range: Range<usize>) -> &[&str] {
+        self.inner.with_dependent(|table| &table.vec[range])
+    }
+
+    /// Returns an iterator over the strings in the table. The items are
+    /// returned in the order they were inserted, matching the [StringId]s.
+    pub fn iter(&self) -> impl Iterator<Item = &str> {
```
danielsn, line 181: Not necessary, and might be tricky with lifetimes but this got me wondering if we can have an `into_iter` implementation that drops the underlying storage when the iteration is done.

morrisonlevi: It's not worth it, IMO, because we'd have to make another self-referencing struct, and it doesn't actually buy us anything (end result is the same).

ivoanjo:
This is superceded by https://github.com/DataDog/libdatadog/pull/404, closing! :)

PR #230, opened 2023-08-24
cicd rfc

sanchda:
Brings libdatadog into adherence with our policy for public repos.  This is especially important here because libdatadog artifacts are consumed as published from this repo.


ivoanjo (approved):
LGTM :+1:

PR #231, opened 2023-08-25
perf(profiling): tune initial buffer sizes

morrisonlevi:
# What does this PR do?

This tunes the initial buffer sizes for encoding the profiles into the serialized pprof format, and also when compressing  files for upload.

# Motivation

The goal is to reduce allocator fragmentation, pressure, and generally return more memory to the system. In a separate PR, Ivo recorded some RSS numbers before/after serialization, and noticed in that PR how it freed up dozens of KiB of memory. The thing is, that change only allocated ~1KiB of string data into a separate arena allocator. The entire difference is therefore about the memory allocation patterns, and not their sizes.

That PR is a bit more complex, and Ivo wasn't sure about shipping it. This PR is very simple, and has a similar effect by simplifying allocations in a different area.

# Additional Notes

None.

# How to test the change?

Just test as normal, but notice that your system RSS when compared to before should return more memory after serialization than it did before.

## For Reviewers


ivoanjo:
This looks great, thank you! :)

PR #234, opened 2023-08-29
refactor: extract profiling::collections module

morrisonlevi:
# What does this PR do?

This moves the traits like Id, Item, and Dedup to a new module in `profiling::collections`.

# Motivation

I think there's likely to be more collections work in the future, but even now `profiling/src/profile/mod.rs` is still over 2k lines so it's worth splitting some of this out where it makes sense.

# Additional Notes

I will be rebasing my string table PR on the top of this one.

# How to test the change?

Some things moved around, but given that PHP is the only profiler not using the `profiling-ffi`, which is unaffected, you can just test as normal.

## For Reviewers


profiling/src/collections/identifiable/mod.rs
```diff
@@ -0,0 +1,102 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+mod string_id;
```
danielsn, line 4: Why is this the only ID that is under here?

danielsn (approved):
LGTM.  One small question

profiling/src/collections/identifiable/mod.rs
morrisonlevi: I picked it out of a +3,000 line change branch and didn't notice ^_^

The reason it's there in that branch is because `StringTable` is a data-structure which uses `StringId`, so it made sense to move it. But for the generic `Table` version, it works abstractly on `crate::collections::{Id, Item}`, so it doesn't need to know anything about `Function` or `FunctionId`, etc.

PR #235, opened 2023-08-29
Run Miri on all of profiling and fix the issues it found

danielsn:
# What does this PR do?

This PR is a followup to #225.  
1. It adds runs Miri on all feasible tests in `profiling` and `profiling-ffi`.
2. It fixes UB in Miri found in some profiling-ffi tests
3. It updates the CI action to run these tests on all future PRs.

# Motivation

Miri is a great way to find UB.  We shouldn't have UB we can prevent.

# Additional Notes

There was a subtle stacked-borrows violation cause by the ffi destructor taking an `Option<&mut>` instead of a `*mut`.  

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b3fb4e106c1d19183c831f38389f2658
shows how this plays out on a small example in the playground (use the tools/miri option to compare `fails` vs `doesnt_fail`.

# How to test the change?

See in CI that its running.

## For Reviewers


profiling-ffi/src/profiles.rs
```diff
@@ -320,11 +320,9 @@ pub unsafe extern "C" fn ddog_prof_Profile_new(
...
 /// created by the Rust Global allocator.
 #[no_mangle]
-pub unsafe extern "C" fn ddog_prof_Profile_drop(profile: Option<&mut Profile>) {
-    if let Some(reference) = profile {
-        // Safety: Profile is not repr(C), and is therefore boxed.
-        drop(Box::from_raw(reference as *mut _))
-    }
+pub unsafe extern "C" fn ddog_prof_Profile_drop(profile: *mut Profile) {
+    // Safety: Profile is not repr(C), and is therefore boxed.
+    drop(Box::from_raw(profile))
```
morrisonlevi, line 325: Help me understand this one. I don't get why this is better, partly because it will not handle a nullptr.

danielsn: Good catch on the `nullptr`. I had actually done this in the playground when I was testing out the code, but forgot to copy that back to the main fix.

The issue is that its not legal to `drop` a `&mut`: to quote *Rust for Rustacians* p11: "the difference between owning a value and having a mutable reference is that the owner is responsible for dropping the value when it is no longer necessary ... if you move the value behind a mutable reference, you must leave another value in its place".  Miri identified that several tests were doing UB because of this: changing to a pointer made miri happy.  Playground link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b3fb4e106c1d19183c831f38389f2658

If you have a better proposal, I'm happy to do that instead.  But I'd rather not leave UB hanging around.

profiling-ffi/src/profiles.rs
```diff
@@ -2,31 +2,64 @@
...
+        Profile {
+            inner: Box::into_raw(Box::new(profile)),
+        }
+    }
+
+    fn take(&mut self) -> Option<Box<profile::Profile>> {
+        // Leaving a null will help with double-free issues that can
+        // arise in C. Of course, it's best to never get there in the
+        // first place!
+        let raw = std::mem::replace(&mut self.inner, std::ptr::null_mut());
```
danielsn, line 32: since `null` is `default`, I think you can just `std::mem::take` here

profiling-ffi/src/profiles.rs
```diff
@@ -305,34 +338,35 @@ pub unsafe extern "C" fn ddog_prof_Profile_new(
...
 
-    NonNull::new_unchecked(Box::into_raw(Box::new(builder.build())))
+    Profile::new(builder.build())
 }
 
 /// # Safety
-/// The `profile` can be null, but if non-null it must point to a valid object
-/// created by the Rust Global allocator.
+/// The `profile` can be null, but if non-null it must point to a Profile
+/// made by this module, which has not previously been dropped.
```
danielsn, line 354: I like this comment better :)

examples/ffi/exporter.cpp
```diff
@@ -40,8 +40,9 @@ int main(int argc, char *argv[]) {
 
   const ddog_prof_Slice_ValueType sample_types = {&wall_time, 1};
   const ddog_prof_Period period = {wall_time, 60};
-  std::unique_ptr<ddog_prof_Profile, Deleter> profile{
+  ddog_prof_Profile profile_data{
       ddog_prof_Profile_new(sample_types, &period, nullptr)};
+  std::unique_ptr<ddog_prof_Profile, Deleter> profile{&profile_data};
```
ivoanjo, line 45: Minor: I wonder if such a simple example needs this kind of sophistication, e.g. maybe we could simplify this a bit?

profiling/src/profile/mod.rs
```diff
@@ -364,20 +364,21 @@ impl Profile {
...
-        }
-        Some(sample_types)
+            .collect();
+        Ok(sample_types)
     }
 
     /// Resets all data except the sample types and period. Returns the
     /// previous Profile on success.
-    pub fn reset(&mut self, start_time: Option<SystemTime>) -> Option<Profile> {
+    pub fn reset(&mut self, start_time: Option<SystemTime>) -> anyhow::Result<Profile> {
```
ivoanjo, line 381: Minor: Is it still worth returning the old profile in this situation? As an API think, it's kinda weird -- when would we want to keep around the old profile after a reset?

ivoanjo (approved):
:+1: LGTM, thanks for making our ffi safer and more strictly checked :)

profiling/src/profile/mod.rs
morrisonlevi: I need to get this release moving so let's talk about that for v5.0. I think originally I was hoping to do  something like:

```rs
let old = profile.reset();
pprof::Profile::from(old); // owned, not borrowed
```

But that didn't work for whatever reason and forgot to revisit. Or may it's that since the structures don't match anymore (did at one time), there's not really value to take ownership of the `profile::Profile`, since it has to copy things anyway.

examples/ffi/exporter.cpp
morrisonlevi: I just fixed it up for the changes I made ^_^

I get that in C++, they want to have destructors run, rather than doing it manually. But obviously the C bindings don't have destructors.

I think we can maybe simplify this code, or acknowledge that we care more about C++ and try to _also_ generate C++ headers and bindings, or something.

PR #237, opened 2023-09-01
Profiler: Take timestamp as an explicit argument, rather than hiding in the label

danielsn:
# What does this PR do?

Changes the API for the profiler to take the timestamp explicitly

# Motivation

Currently, the client does the work of building a label with the timestamp, and adding it to the Labels on the sample.
libdatadog then finds that label, extracts the information, and removes it from the vector.  Making it a function argument is cleaner and clearer.

# Additional Notes

# How to test the change?

## For Reviewers

### DIFF
--- a/profiling-ffi/src/profiles.rs
+++ b/profiling-ffi/src/profiles.rs
@@ -6,6 +6,7 @@ use datadog_profiling::profile::{self, api, profiled_endpoints};
 use ddcommon_ffi::slice::{AsBytes, CharSlice, Slice};
 use ddcommon_ffi::Error;
 use std::convert::{TryFrom, TryInto};
+use std::num::NonZeroI64;
 use std::str::Utf8Error;
 use std::time::{Duration, SystemTime};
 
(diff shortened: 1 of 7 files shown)


profiling-ffi/src/profiles.rs
```diff
@@ -388,8 +389,9 @@ impl From<ProfileResult> for Result<(), String> {
 pub unsafe extern "C" fn ddog_prof_Profile_add(
     profile: *mut Profile,
     sample: Sample,
+    timestamp: Option<NonZeroI64>,
```
ivoanjo, line 392: Just to doublecheck -- this gets translated to a `int64_t` in the C headers; does that mean `None` gets represented by `0`?

ivoanjo (commented):
Seems reasonable -- but there's still a few more places that need updatin' (see broken CI)

profiling-ffi/src/profiles.rs
paullegranddc: It's specifically documented in the std https://doc.rust-lang.org/stable/std/num/struct.NonZeroI64.html

ivoanjo: > NonZeroI64 is guaranteed to have the same layout and bit validity as i64 with the exception that 0 is not a valid instance. Option<NonZeroI64> is guaranteed to be compatible with i64, including in FFI.

Is that what this means? Because... it's rather cryptic lol.

paullegranddc: yes I think it's quite explicit

danielsn:
> Seems reasonable -- but there's still a few more places that need updatin' (see broken CI)

Fixed the issues that showed in CI

ivoanjo (approved):
:+1: LGTM

PR #241, opened 2023-09-06
Support obfuscating all redis arguments + integrate redis obfuscator (Redis obfuscation 3/3)

thedavl:
# What does this PR do?

Redis obfuscation PR 3/3

Implementation is closely modeled after the go agent's trace obfuscator. Unit tests are copied over near 1:1

# Motivation

# Additional Notes

# How to test the change?

## For Reviewers


trace-obfuscation/src/redis.rs
```diff
@@ -169,11 +169,55 @@ fn obfuscate_redis_args_step(mut args: Vec<String>, start: usize, step: usize) -
...
+    if redis_cmd_iter.peek().is_none() {
+        return obfuscated_cmd;
+    }
+
+    let cmd = redis_cmd_iter.next().unwrap_or_default();
+    obfuscated_cmd.push_str(cmd);
+
+    if redis_cmd_iter.peek().is_none() {
+        return obfuscated_cmd;
+    }
```
maxday, line 184: Is there a way to simplify this code block? It seems that we are doing some peek/next/peek which makes it a bit hard to understand

thedavl: I was trying to avoid collecting the iterator, hence the `peek().is_none()` to check that the iterator actually has a next item

The first peek ensures the string isn't empty. If it isn't then we can consume the first token as the "command" of the redis string

The second peek is to determine if there are any tokens after the command token. If not, we return just he command. Otherwise we proceed with some obfuscation

maxday (approved):

PR #242, opened 2023-09-07
Reduce memory consumption by consuming the profile as we serialize it

danielsn:
# What does this PR do?

Takes the `Profile` by ownership instead of borrow while serializing, allowing a more memory-efficient translation

# Motivation

Currently, we clone all values as we build the serializer.  Instead, we can take them by owned value, and save a lot of clones (and hence customer memory use)

# Additional Notes

# How to test the change?

## For Reviewers

### DIFF
--- a/profiling-ffi/src/profiles.rs
+++ b/profiling-ffi/src/profiles.rs
@@ -658,6 +658,7 @@ impl From<profile::EncodedProfile> for EncodedProfile {
 }
 
 /// Serialize the aggregated profile.
+/// Drains the data, and then resets the profile for future use.
 ///
 /// Don't forget to clean up the ok with `ddog_prof_EncodedProfile_drop` or
 /// the error variant with `ddog_Error_drop` when you are done with them.
@@ -671,6 +672,7 @@ impl From<profile::EncodedProfile> for EncodedProfile {
 ///                      under anomalous conditions this may fail as system clocks can be adjusted,
 ///                      or the programmer accidentally passed an earlier time. The duration of
 ///                      the serialized profile will be set to zero for these cases.
+/// * `start_time` - Optional start time for the next profile.
 ///
 /// # Safety
 /// The `profile` must point to a valid profile object.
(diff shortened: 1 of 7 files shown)


profiling-ffi/src/profiles.rs
```diff
@@ -691,13 +694,22 @@ pub unsafe extern "C" fn ddog_prof_Profile_serialize(
             ))
         }
     };
+    let old_profile = match profile.reset(start_time.map(SystemTime::from)) {
```
morrisonlevi, line 697: Probably should add something in the `ddog_prof_Profile_serialize` docs saying that the profile will be reset.

profiling/src/collections/identifiable/mod.rs
```diff
@@ -75,9 +75,9 @@ impl<T: Item> Dedup<T> for FxIndexSet<T> {
     }
 }
 
-pub fn to_pprof_vec<T: PprofItem>(collection: &FxIndexSet<T>) -> Vec<T::PprofMessage> {
+pub fn to_pprof_vec<T: PprofItem>(collection: FxIndexSet<T>) -> Vec<T::PprofMessage> {
```
morrisonlevi, line 78: Since it consumes, rename it `into_pprof_vec`?

profiling/src/profile/mod.rs
```diff
@@ -400,12 +400,14 @@ impl Profile {
...
     ///                these cases.
     pub fn serialize(
-        &self,
+        self,
         end_time: Option<SystemTime>,
         duration: Option<Duration>,
     ) -> anyhow::Result<EncodedProfile> {
         let end = end_time.unwrap_or_else(SystemTime::now);
         let start = self.start_time;
+        let endpoints_stats = self.endpoints.stats.clone();
```
morrisonlevi, line 409: Why do these need cloned since we have it by `self`?

morrisonlevi (commented):
Did a quick drive-by review before I take a little break, will look at it more later.

profiling-ffi/src/profiles.rs
danielsn: https://github.com/DataDog/libdatadog/pull/242/files#diff-e1c96667d7e8245b5d270aa32f45b199929b795e3e26e7100a98d9e8125dbdc6R661

profiling/src/collections/identifiable/mod.rs
danielsn: done

profiling/src/profile/mod.rs
danielsn: We fully consume `self` on the next line, so we can't partially consume it here.  But `std::mem::take` would work, and save the clone.

danielsn:
Closed in favour of #243

PR #243, opened 2023-09-07
Profiler: Rewrite serializer to stream compressed output, reducing memory usage

danielsn:
# What does this PR do?

Reduces memory needed to serialize the profiler into a pprof. In particular, it enables serialization to occur in a streaming manner which consumes the profile, allowing memory usage to remain flat.

# Motivation

Currently, serialization takes significant memory (see below, ~60MB on a Ruby benchmark with timeline enabled).  The standard `prost` model of serializing to protobuf works on complete messages, and requires multiple copies of the data.  Instead, tweak the serialization to:

1. Consume the in-memory representation of the Profile as it is serialized
2. Stream the creation of the serialized pprof
3. stream lz4 encoding of the pprof (timeline pprofs compress extremely well, so keeping only the compressed version in memory saves memory).

Running @ivoanjo Ruby benchmark gives

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-4f1ff639-7fff-dbc6-336e-b6d4bce873bf"><div dir="ltr" style="margin-left:0pt;" align="center">

Config | Timeline | Process Memory Rss before serialization (bytes) | Process Memory Rss after serialization (bytes) | Process Memory Rss after GC (bytes) | Maximum rss (from /bin/time -v, bytes)
-- | -- | -- | -- | -- | --
16 threads with depth 50 (libdatadog 4.0) | false | 70980k | 70980k | 70980k | 70980k
16 threads with depth 50(libdatadog 4.0) | true | 78636k | 134004k | 85196k | 153676k
16 threads with depth 50 (libdatadog 5.0) | false | 70900k | 71060k | 71060k | 71060k
16 threads with depth 50 (libdatadog 5.0) | true | 78808k | 76540k | 76540k | 79128k
4 threads with depth 400(libdatadog 5.0) | true | 76512k | 76136k | 76136k | 77472k

</div><br /></b>

# Additional Notes

* Builds on @ivoanjo work in https://github.com/DataDog/libdatadog/commit/3bb98dc5074e034ca8b26f50360e2291274c8652
* The key insight is that the **top level** protobuf message doesn't carry any frame information. So the emitted bytes are identical if we either (1) serialize the `Profile` top level message or (2) serialize a series of single field messages, each with the tag of their corresponding `Profile` field.  This only works for the top-level message, but luckily that's all we need.*

# How to test the change?
* Rerun the ruby benchmark, and see the memory savings
* I modified all tests to round-trip through the serializer, which ensures that both serialization and deserialization work. 

## For Reviewers


profiling/src/exporter/mod.rs
```diff
@@ -238,22 +236,25 @@ impl ProfileExporter {
...
-            // well, but these are just initial Vec sizes, not a hard-bound.
-            // Using 1/10 gives us a better start than starting at zero, while
-            // not reserving too much for things that compress really well, and
-            // power-of-two capacities are almost always the best performing.
-            let capacity = (file.bytes.len() / 10).next_power_of_two();
-            let buffer = Vec::with_capacity(capacity);
-            let mut encoder = FrameEncoder::new(buffer);
-            encoder.write_all(file.bytes)?;
-            let encoded = encoder.finish()?;
+            //TODO : confirm that this is not needed since we now pre-encode
```
morrisonlevi, line 239: This will stop compressing provenance files and other files which are uploaded.

danielsn: Added an option for files to control whether they should be compressed before export

profiling/src/profile/mod.rs
```diff
@@ -430,14 +434,68 @@ impl Profile {
...
+
+            encoder.encode(ProfileSamplesEntry::from(item))?;
+        }
+
+        // We need to do this out of order because we use the sample_types while
+        // upscaling
+        for sample_type in self.sample_types.into_iter() {
+            let item: pprof::ValueType = sample_type.into();
+            encoder.encode(ProfileSampleTypesEntry::from(item))?;
+        }
```
morrisonlevi, line 464: Sorry, what is out of order here? Aren't the sample types the first field in a profile?

profiling/src/profile/pprof/sliced_proto.rs
```diff
@@ -0,0 +1,115 @@
...
+//!
+//! The `tag` number and type for each of these sliced messages matches the
+//! corresponding field in the `Profile` message in `profile.proto`.
+
+use super::*;
+
+#[derive(Eq, Hash, PartialEq, ::prost::Message)]
+pub struct ProfileSampleTypesEntry {
+    #[prost(message, tag = "1")]
+    pub sample_types_entry: Option<ValueType>,
```
morrisonlevi, line 19: I am confused here. Why are so many of these sliced proto fields using `Option` aka 0 or 1 instead of of a type which is 0 or more? We definitely have more than one sample type, for instance. What am I missing?

profiling/src/profile/mod.rs
danielsn: We first emitted the `Sample`, then the `SampleTypes`, which is the opposite order of the profile.  I'll try to clean up the comment to clarify.

profiling-ffi/src/profiles.rs
```diff
@@ -691,13 +694,22 @@ pub unsafe extern "C" fn ddog_prof_Profile_serialize(
...
     let end_time = end_time.map(SystemTime::from);
     let duration = match duration_nanos {
         None => None,
         Some(x) if *x < 0 => None,
         Some(x) => Some(Duration::from_nanos((*x) as u64)),
     };
-    match profile.serialize(end_time, duration) {
+    match old_profile.serialize_into_compressed_pprof(end_time, duration) {
         Ok(ok) => SerializeResult::Ok(ok.into()),
         Err(err) => SerializeResult::Err(err.into()),
```
morrisonlevi, line 714: Since `serialize_into_compressed_pprof` returns an `anyhow::Result`, I'd convert the error branch to use `err.context("ddog_prof_Profile_serialize failed")` as well.

profiling/src/serializer/mod.rs
```diff
@@ -0,0 +1,6 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+pub mod compressed_streaming_encoder;
+
+pub use compressed_streaming_encoder::*;
```
morrisonlevi, line 6: Personally, I like either `pub mod` without using stuff, or private `mod`, with `pub use mod::*`. Any particular reason for doing both here?

morrisonlevi (approved):

ivoanjo:
Before reviewing the PR, I've spent some time updating the [Ruby timeline memory testing branch](https://github.com/datadog/dd-trace-rb/tree/ivoanjo/timeline-memory-experiments-v2) with the latest changes.

The test suite is 100% green with the new serializer, which I think is great news -- there's a lot of tests in the Ruby profiler test suite where we turn on the profiler, then serialize the result, then deserialize it and assert on it, and no issues there with the serializer.

I've also tested it in staging with our prof-fibonacci minimal test app, and both timeline, as well as regular profiles are in good shape. I also did some really quick memory measurements for prof-fibonacci:

* dd-trace-rb v1.14.0 (libdatadog 3) + timeline: maximum rss 151548 kb
* timeline testing branch (libdatadog 5) + timeline: maximum rss 99620 kb
* timeline testing branch (libdatadog 5) + no timeline: maximum rss 92812 kb

profiling/src/profile/pprof/sliced_proto.rs
```diff
@@ -0,0 +1,133 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+//! This file is a companion to the standard pprof protobuf definition.
+//! `Profile` fields are "sliced" into separate messages, allowing them to be
+//! serialized in a streamed manner.  This takes advantage of the fact that
+//! the top level message in a protobuf doesn't have a length header, so the
+//! bits on the wire are indistinguishable between serializing these sliced
+//! messages, and serializing the top-level message.
```
ivoanjo, line 9: May be worth adding here, for future context, why streamed emission is useful ;)

profiling/src/profile/pprof/sliced_proto.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+#[derive(Eq, Hash, PartialEq, ::prost::Message)]
+pub struct ProfileStringTableEntry {
+    // profile.proto requires that 'string_table[0] must always be "".'
+    // Writing "" to a protobuf is a no-op, unless the field is "repeated".
+    // Making this field repeated (and hence take Vec<String>) ensures that the
+    // initial "" will be correctly serialized.
+    #[prost(string, required, tag = "6")]
+    pub string_table_entry: String,
```
ivoanjo, line 71: The comment is outdated :P

profiling/src/profile/pprof/sliced_proto.rs
```diff
@@ -0,0 +1,133 @@
...
+pub struct ProfileLocationsEntry {
+    #[prost(message, required, tag = "4")]
+    pub locations_entry: Location,
+}
+
+#[derive(Eq, Hash, PartialEq, ::prost::Message)]
+pub struct ProfileFunctionsEntry {
+    #[prost(message, required, tag = "5")]
+    pub function_entry: Function,
+}
```
ivoanjo, line 62: Minor: For consistency with the above, I think this should be `functions_entry`

```suggestion
#[derive(Eq, Hash, PartialEq, ::prost::Message)]
pub struct ProfileFunctionsEntry {
    #[prost(message, required, tag = "5")]
    pub functions_entry: Function,
}
```

profiling/src/profile/mod.rs
```diff
@@ -430,14 +434,70 @@ impl Profile {
...
+        let mut encoder = CompressedProtobufSerializer::with_capacity(INITIAL_PPROF_BUFFER_SIZE);
+
+        for (sample, timestamp, mut values) in std::mem::take(&mut self.observations).into_iter() {
+            let labels = self.translate_and_enrich_sample_labels(sample, timestamp)?;
+            let location_ids: Vec<_> = self
+                .get_stacktrace(sample.stacktrace)
+                .locations
+                .iter()
+                .map(Id::to_raw_id)
+                .collect();
```
ivoanjo, line 446: Minor: Probably not this PR, but it occurs to me that we may be repeating some work here, because, e.g. we're repeating the `location_ids` conversion for all samples that have the example same stacktrace, but different timestamp.

profiling/src/profile/mod.rs
```diff
@@ -430,14 +434,70 @@ impl Profile {
...
+                labels,
+            };
+
+            encoder.encode(ProfileSamplesEntry::from(item))?;
+        }
+
+        // `Sample`s must be emitted before `SampleTypes` since we use
+        // the `sample_types` during upscaling.
+        // It is valid to emit protobuf fields out of order. See example in:
+        // https://protobuf.dev/programming-guides/encoding/#optional
```
ivoanjo, line 462: Minor: This comment makes sense, buttt I think there's an underlying assumption that's not clearly explained, which is that the iterators are consuming/releasing stuff as they are working, and thus if we were to reorder this to be the first step of the serialization process, we'd consume stuff right away.

ivoanjo (commented):
This is looking great! I've left some notes regarding forcing an API break on the ffi for the compression stuff, but other than that I think it's good to go!

(I'll admit the stuff inside `internal/observation` is a bit of dark magic, I'll leave those bits for Levi to check :P )

ivoanjo:
P.s: Also probably not a bad idea to update the PR title to reflect that we also kinda rebuilt the serializer ;)

profiling/src/profile/pprof/sliced_proto.rs
danielsn: Updated with comments in the file

profiling/src/serializer/mod.rs
danielsn: fixed

profiling-ffi/src/profiles.rs
danielsn: Done

examples/ffi/exporter.cpp
```diff
@@ -86,7 +86,7 @@ int main(int argc, char *argv[]) {
   }
 
   ddog_prof_Profile_SerializeResult serialize_result =
-      ddog_prof_Profile_serialize(profile.get(), nullptr, nullptr);
+      ddog_prof_Profile_serialize(profile.get(), nullptr, nullptr, nullptr);
```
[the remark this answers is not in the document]
danielsn: There's an RFC for that. https://internals.rust-lang.org/t/pre-rfc-named-arguments/16413

profiling-ffi/src/profiles.rs
```diff
@@ -691,13 +694,22 @@ pub unsafe extern "C" fn ddog_prof_Profile_serialize(
             ))
         }
     };
+    let old_profile = match profile.reset(start_time.map(SystemTime::from)) {
+        Ok(ok) => ok,
+        Err(err) => {
+            return SerializeResult::Err(Error::from(
+                err.context("ddog_prof_Profile_serialize failed"),
+            ))
```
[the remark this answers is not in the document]
danielsn: done

profiling/src/profile/pprof/sliced_proto.rs
danielsn: done

profiling/src/profile/pprof/sliced_proto.rs
danielsn: done

profiling/src/profile/pprof/sliced_proto.rs
danielsn: done

examples/ffi/exporter.cpp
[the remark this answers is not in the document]
ivoanjo: If that RFC gets approved, maybe I'll finally start enjoying Rust haha :D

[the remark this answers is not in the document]
morrisonlevi: That's all well and good but the snippet you commented on is in C++ so... expect it never ^_^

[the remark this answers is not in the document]
danielsn: Turns out you can hack it into C++

profiling-ffi/src/exporter.rs
```diff
@@ -43,6 +43,7 @@ pub enum Endpoint<'a> {
 pub struct File<'a> {
     name: CharSlice<'a>,
     file: ByteSlice<'a>,
+    compress_before_exporting: bool,
 }
```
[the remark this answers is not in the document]
danielsn: Done

examples/ffi/exporter.cpp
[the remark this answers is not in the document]
ivoanjo: Yeah... it's something I guess.

But then it's boilerplate-y, and it suffers from the exact same "changing signature" issue we were discussing for the compressed files -- the compiler won't bat an eyelash if you omit some of the arguments, so it's easy to do an oops :(

examples/ffi/exporter.cpp
```diff
@@ -120,18 +120,24 @@ int main(int argc, char *argv[]) {
...
   }};
-
-  ddog_prof_Exporter_Slice_File files = {.ptr = files_, .len = sizeof files_ / sizeof *files_};
+  ddog_prof_Exporter_Slice_File files_to_compress_and_export = {.ptr = files_to_compress_and_export_, .len = sizeof files_to_compress_and_export_ / sizeof *files_to_compress_and_export_};
+  
+  // Ensure we get a non-null pointer: it is required that:
+  // "data must be non-null and aligned even for zero-length slices."
+  // https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html
+  ddog_prof_Exporter_File files_to_export_unmodified_[1];
+  ddog_prof_Exporter_Slice_File files_to_export_unmodified = {.ptr = files_to_export_unmodified_, .len = 0};
```
ivoanjo, line 133: Hmmmmmmmmmmmmm.... can we do better for ffi consumers? Having to do this if you don't have any extra files is... painful.

(Alternatively, maybe we could `#define` a macro `NO_FILES` that ffi consumers can use for this case in the headers generated by cbindgen?)

To be clear -- my concern is not on about this example, it's more that I believe some of our profilers may need to do something similar.

profiling/src/profile/mod.rs
```diff
@@ -430,14 +437,76 @@ impl Profile {
...
+                location_ids,
+                values,
+                labels,
+            };
+
+            encoder.encode(ProfileSamplesEntry::from(item))?;
+        }
+
+        // `Sample`s must be emitted before `SampleTypes` since we use consume
+        // fields as we convert (using `into_iter`).  This allows Rust to
```
ivoanjo, line 463: Minor:

```suggestion
        // `Sample`s must be emitted before `SampleTypes` since we consume
        // fields as we convert (using `into_iter`).  This allows Rust to
```

profiling/src/profile/pprof/sliced_proto.rs
```diff
@@ -0,0 +1,153 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+//! This file is a companion to the standard pprof protobuf definition.
+//!
+//! When `prost` serializes the standard `pprof` protobuf, it does a batch
+//! operation:
+//! 1. the entire `internal::Profile` is converted to `pprof::Profile`,
+//! 2. which is then serialized into a byte-vector holding the protobuf,
+//! 3. which is then compressed.
```
ivoanjo, line 10: Minor: Technically, 1. is on our side, not prost that does it, and 3. prost does do at all

profiling/src/profile/pprof/sliced_proto.rs
```diff
@@ -0,0 +1,153 @@
...
+//! If we stream the creation of the protobuf, we can avoid this memory blowup.
+//! In particular, if each `pprof` Message (e.g. Sample, Location, etc) is
+//! generated in a streaming fashion, and then immediately serialized and
+//! compressed, we go directly from the efficient `internal` format to an
+//! a compact compressed serialized format, without needing to ever create large
+//! intermediate data-structures.
+//!
+//! We do this by taking advantage of the fact that the top level message in a
+//! protobuf doesn't have a length header.  This means that the bits on the wire
+//! are indistinguishable between serializing a single the top-level message,
```
ivoanjo, line 30: Minor:

```suggestion
//! are indistinguishable between serializing a single top-level message,
```

Also the paragraph mixes "top-level" and "top level" ;)

ivoanjo (dismissed):
:+1: LGTM

The hack to support a 0-length list of files to export for ffi consumers is kinda ugly, but I don't think we need to block the PR on it.

examples/ffi/exporter.cpp
morrisonlevi: I'm going to ask @sanchda to back me up here, but this should be fine:

```cpp
ddog_prof_Exporter_Slice_File files_to_export_unmodified = {
    .ptr = &(ddog_prof_Exporter_File){},
    .len = 0,
};
```

Even if that address `&(ddog_prof_Exporter_File){}` becomes a dangling pointer, it doesn't matter: it will never be dereferenced, but it will still be properly aligned for a `ddog_prof_Exporter_File`. Right, David?

sanchda: Right; the struct you've defined is going to have a pointer to a default-constructed `ddog_prof_Exporter_File` on the stack.  It'll be a valid pointer for the duration of the current scope, after which it'll become invalid.  The requirement that `.ptr` be "valid" is a little painful, but that's imposed on us.

I like @ivoanjo's note; create a macro (equivalently, just a global pointer to a static-initialized element) that references a static singleton element with 0 .len and a payload that contains some kind of clear error message or whatever, then pass that.

For discussion, I have a vague sense that the standard might be a little overbearing here; this is a fairly simple struct, so maybe just passing `{}` will be acceptable (I didn't test it or think too hard about this).  I'd argue that _if_ this is viable, then wrapping it in a macro anyway is better, and thus just going through the trouble of having a macro pass a sentinel zero-file is better.

I don't really have a strong opinion.  I'd wrap this in a helper function on my side anyway if I had to do this in more than one place.  That said, I sense @ivoanjo's comment was less about the unbearable burden of doing this once and more about having to follow _yet another thing_ across API changes as the library evolves.  :shrug:

morrisonlevi: I'm also fine with making `ddog_prof_Exporter_Slice_File_empty()` which returns an empty `ddog_prof_Exporter_Slice_File` for you.

ivoanjo: `ddog_prof_Exporter_Slice_File_empty()` is quite reasonable, happy to go with that :)

ivoanjo (approved):
:+1: LGTM

PR #244, opened 2023-09-08
Redis query tokenizer (Redis obfuscation 1/3)

thedavl:
# What does this PR do?

Redis obfuscation PR 1/3

Creates a redis tokenizer to be used in redis query obfuscation.

Implementation is closely modeled after the go agent's trace obfuscator. Unit tests are copied over near 1:1

# Motivation

Mini Agent trace obfuscation

# Additional Notes

# How to test the change?

Unit tested

## For Reviewers


trace-obfuscation/src/redis_tokenizer.rs
```diff
@@ -0,0 +1,443 @@
...
+#[derive(Debug)]
+pub enum RedisTokenType {
+    RedisTokenCommand,
+    RedisTokenArgument,
+}
+
+pub struct RedisTokenizer {
+    data: Vec<char>,
+    cur_char: char,
+    offset: Option<usize>,
```
maxday, line 13: What do you think about removing this option, setting 0 instead of None. I feel that the code would be simpler but I may have missed something?

thedavl: The main purpose was so that `next()` knows whether or not it's being called for the very first time. Alternatively, I could add a `has_started` field, and remove the option?

maxday (approved):

PR #250, opened 2023-09-13
Profiling: refactor upscaler

danielsn:
# What does this PR do?

Refactors the upscaling code to make it more self-contained, and removes some duplicated checks.

# Motivation

1. Make the upscaling code more self-contained, simplifying the code in the Profile
2. Removes duplicated checks
3. General code cleanup

# Additional Notes

I removed the duplicate check when we do the upscaling, since the only way to add a rule already does a duplicate check.  

# How to test the change?

## For Reviewers


gleocadie (approved):
LGTM
at least for the upscaling process :)

PR #252, opened 2023-09-18
refactor!(profiling-ffi): ddog_prof_Profile_new can fail

morrisonlevi:
# What does this PR do?

`ddog_prof_Profile_new` now returns a `ddog_prof_Profile_NewResult` instead of a `ddog_prof_Profile`.

# Motivation

Currently, this routine doesn't fail. In the future, it's expected to be possible to fail, so we want to get clients to be prepared for it since we're also shipping breaking changes in v5.0.0 already.

# Additional Notes

None.

# How to test the change?

Adapt to the above change and test as usual otherwise. Note that currently it cannot fail, but do not rely on that. Add proper error handling.

## For Reviewers

### DIFF
--- a/ddcommon-ffi/src/vec.rs
+++ b/ddcommon-ffi/src/vec.rs
@@ -21,6 +21,10 @@ pub struct Vec<T: Sized> {
     _marker: PhantomData<T>,
 }
 
+unsafe impl<T: Send> Send for Vec<T> {}
+
+unsafe impl<T: Sync> Sync for Vec<T> {}
+
 impl<T: PartialEq> PartialEq for Vec<T> {
     fn eq(&self, other: &Self) -> bool {
         self.len() == other.len() && self.iter().zip(other.iter()).all(|(s, o)| s == o)
--- a/examples/ffi/exporter.cpp
+++ b/examples/ffi/exporter.cpp
@@ -40,9 +40,12 @@ int main(int argc, char *argv[]) {
 
   const ddog_prof_Slice_ValueType sample_types = {&wall_time, 1};
   const ddog_prof_Period period = {wall_time, 60};
-  ddog_prof_Profile profile_data{
-      ddog_prof_Profile_new(sample_types, &period, nullptr)};
-  std::unique_ptr<ddog_prof_Profile, Deleter> profile{&profile_data};
+  ddog_prof_Profile_NewResult profile_new_result =
+      ddog_prof_Profile_new(sample_types, &period, nullptr);
+  if (profile_new_result.tag != DDOG_PROF_PROFILE_NEW_RESULT_OK) {
+    print_error("Failed to make new profile: ", profile_new_result.err);
+  }
+  std::unique_ptr<ddog_prof_Profile, Deleter> profile{&profile_new_result.ok};
 
   ddog_prof_Location root_location = {
       // yes, a zero-initialized mapping is valid
@@ -130,13 +133,13 @@ int main(int argc, char *argv[]) {
   };
 
   ddog_prof_Exporter_Slice_File files_to_export_unmodified = ddog_prof_Exporter_Slice_File_empty();
-  
+
   ddog_CharSlice internal_metadata_example = DDOG_CHARSLICE_C(
       "{\"no_signals_workaround_enabled\": \"true\", \"execution_trace_enabled\": \"false\"}");
 
-  ddog_prof_Exporter_Request_BuildResult build_result =
-      ddog_prof_Exporter_Request_build(exporter, encoded_profile->start, encoded_profile->end,
-                                       files_to_compress_and_export, files_to_export_unmodified, nullptr, nullptr, &internal_metadata_example, 30000);
+  ddog_prof_Exporter_Request_BuildResult build_result = ddog_prof_Exporter_Request_build(
+      exporter, encoded_profile->start, encoded_profile->end, files_to_compress_and_export,
+      files_to_export_unmodified, nullptr, nullptr, &internal_metadata_example, 30000);
   ddog_prof_EncodedProfile_drop(encoded_profile);
 
   if (build_result.tag == DDOG_PROF_EXPORTER_REQUEST_BUILD_RESULT_ERR) {
--- a/examples/ffi/profiles.c
+++ b/examples/ffi/profiles.c
@@ -6,6 +6,7 @@
 #include <datadog/common.h>
 #include <datadog/profiling.h>
 #include <stdio.h>
+#include <stdlib.h>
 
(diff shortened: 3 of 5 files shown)


ddcommon-ffi/src/vec.rs
```diff
@@ -21,6 +21,10 @@ pub struct Vec<T: Sized> {
     _marker: PhantomData<T>,
 }
 
+unsafe impl<T: Send> Send for Vec<T> {}
+
+unsafe impl<T: Sync> Sync for Vec<T> {}
```
morrisonlevi, line 26: These were required for `datadog_profiling_ffi::Error` to be able to be converted to certain other Error types.

examples/ffi/exporter.cpp
```diff
@@ -40,9 +40,12 @@ int main(int argc, char *argv[]) {
...
   const ddog_prof_Slice_ValueType sample_types = {&wall_time, 1};
   const ddog_prof_Period period = {wall_time, 60};
-  ddog_prof_Profile profile_data{
-      ddog_prof_Profile_new(sample_types, &period, nullptr)};
-  std::unique_ptr<ddog_prof_Profile, Deleter> profile{&profile_data};
+  ddog_prof_Profile_NewResult profile_new_result =
+      ddog_prof_Profile_new(sample_types, &period, nullptr);
+  if (profile_new_result.tag != DDOG_PROF_PROFILE_NEW_RESULT_OK) {
+    print_error("Failed to make new profile: ", profile_new_result.err);
+  }
```
ivoanjo, line 47: Are we missing the `ddog_Error_drop` here? I'm thinking it's relevant to have it in the example, as people may be inspired to copy/paste from here, and this way they won't forget the cleanup.

Also, like some of the other error checks, I think we should probably terminate the example if there's an error, since right now we're continuing with a weird broken state.

examples/ffi/profiles.c
```diff
@@ -15,7 +16,14 @@ int main(void) {
...
 
-  ddog_prof_Profile profile = ddog_prof_Profile_new(sample_types, &period, NULL);
+  ddog_prof_Profile_NewResult new_result = ddog_prof_Profile_new(sample_types, &period, NULL);
+  if (new_result.tag != DDOG_PROF_PROFILE_NEW_RESULT_OK) {
+    ddog_CharSlice message = ddog_Error_message(&new_result.err);
+    fprintf(stderr, "%*s", (int)message.len, message.ptr);
+    exit(EXIT_FAILURE);
+  }
+
+  ddog_prof_Profile *profile = &new_result.ok;
```
ivoanjo, line 26: This change is kinda... weird/interesting, since in a way, we're going back from what we did in https://github.com/DataDog/libdatadog/pull/235 . That is, we're again mostly using `ddog_prof_Profile *` in client code, with `ddog_prof_Profile_NewResult` now taking the place of how `ddog_prof_Profile` was supposed to be used after #235.

My question is -- do we still need the "new" (post-#235) `ddog_prof_Profile` if `NewResult` can take its place?

ivoanjo (commented):
Left a few notes! Overall seems reasonable, if a bit... inconsistent with some of the changes in #235 .

examples/ffi/exporter.cpp
r1viollet: Is there any way we could make this clearer ? 
Especially as we have made this mistake many times.  @danielsn would you have ideas ? Maybe even naming ?

examples/ffi/profiles.c
morrisonlevi: I don't think it works against that change at all. The `struct ddog_prof_Profile` was made so things like double-drop and use-after-free can be detected and protected against (shouldn't be relied on, just a defensive practice). All these still stand, even when it's tucked into `NewResult`. Is there some aspect I'm not thinking about?

ivoanjo: It looks like we have a `NewResult` that tucks inside a `Profile` that then has a reference to the actual profile inside itself.

My question is -- why do we need both `NewResult` and `Profile`? Are they not effectively equivalent to the ffi user? E.g. in practice a `NewResult` seems to already be a `Profile` that is not an error ;)

For instance, notice that in this case, the example doesn't need to keep a `profile` -- it's just keeping the whole `NewResult`.

morrisonlevi: Consider the next PR which adds an arena allocator to the profile. Creating a profile can now. Boom. I mean that's literally my plan though I don't know if it will actually be the _next_ PR. I know from my experiments that the PR will be large, so I was trying to 1) minimize disruption in API and 2) trying to make that PR a bit smaller.

examples/ffi/exporter.cpp
morrisonlevi: Yes, we are missing some error stuff here. I wrote this part kind of hastily. Sorry about that. I'll fix it later.

examples/ffi/profiles.c
morrisonlevi: As for the `Profile`, the _real_ profile type is not `repr(C)`. The FFI Profile type is there to be a more safe FFI wrapper which is also `repr(C)`. `NewResult` has no bearing on this; that is merely a Success/Failure indicator.

ivoanjo: Ah, wait, note that my comment is not "NewResult shouldn't exist" -- it's more "Does Profile need to exist as a carrier of a reference to the internal profile if NewResult can also be a carrier to the internal profile?".

But either way, my comment is not a blocker! More of a observation about how our API looks like.

examples/ffi/exporter.cpp
danielsn: For C++, could we wrap some of this stuff in smart pointers, guaranteeing that the drops happen.  I wonder if it would be worth having a seperate C and C++ API to give more protection to the C++ side

morrisonlevi: I will leave that up to the C++ users (maybe native + .NET?)

ivoanjo (approved):
I'm OK with the PR as-is! Thinking back on it, my previous comment

> Left a few notes! Overall seems reasonable, if a bit... inconsistent with some of the changes in https://github.com/DataDog/libdatadog/pull/235 .

Made a bigger thing of out if than what I was intended! I spent some time thinking about it, and I think the addition of `NewResult` makes sense for consistency, even if technically we could maybe merge it with the `Profile` ffi type (or vice-versa).

So, let's get this show on the road! :)

ddcommon-ffi/src/vec.rs
paullegranddc: This looks correct to me

PR #253, opened 2023-09-19
Sql Obfuscation (sql obfuscation 2/3)

thedavl:
# What does this PR do?

Adds the sql obfuscator, using the prev sql tokenizer. PR 2/3 for sql trace obfuscation

# Motivation

# Additional Notes

# How to test the change?
 Unit tests have been copied over from the go agent sql obfuscator.

## For Reviewers


trace-obfuscation/src/sql.rs
```diff
@@ -0,0 +1,1385 @@
...
+
+pub fn obfuscate_sql_string(s: &str, config: &ObfuscationConfig) -> AttemptSqlObfuscationResult {
+    let use_literal_escapes = false;
+    let mut tokenizer = SqlTokenizer::new(s, config.sql_literal_escapes);
+    let result = attempt_sql_obfuscation(tokenizer, config.sql_replace_digits);
+    if result.error.is_none() || !result.seen_escape {
+        return result;
+    }
+
+    tokenizer = SqlTokenizer::new(s, !use_literal_escapes);
```
thedavl, line 22: if obfuscation failed and we saw an escape character, try treating backslashes as character rather than escapes.

thedavl: We're going to want to store this in a global mutex eventually, so that future obfuscation attempts start with the correct literal escape config

PR #247, opened 2023-09-12
Use Cargo workspace properties and document MSRV

morrisonlevi:
# What does this PR do?

This adds `workspace.package` to the root `Cargo.toml`:

```rs
rust-version = "1.69"
edition = "2021"
version = "4.0.0"
license = "Apache-2.0"
```

# Motivation

This allows changing things in fewer places. I hope to make changing the minimum supported rust version (MSRV), edition, and libdatadog version easier.

# Additional Notes

Not every crate inherits these new properties. The serverless and sidecar crates, and some of their dependencies, have remained under their own versioning and control.

# How to test the change?

Regular tests should apply. This shouldn't change any behavior.

## For Reviewers
- [ ] If this PR touches code that signs or publishes builds or packages, or handles credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.
- [x] This PR doesn't touch any of that.

PR #236, opened 2023-08-30
Bump ddcommon, ddtelemetry, and profiling crates to v4.0.0

morrisonlevi:
# What does this PR do?

Bumps these crates to v4.0.0:

- `ddcommon` and `ddcommon-ffi`
- `ddtelemetry` and `ddtelemetry-ffi`
- `profiling`, `profiling-ffi`, `profiling-replayer`

It does not bump the sidecar*, serverless, nor trace* crates as their version numbers are out-of-sync and I wasn't sure if I should touch them, as I was not involved in any conversations about this.

# Motivation

I want to release a new PHP profiler that is based on all the lovely work we did during R&D week.

# Additional Notes

This is waiting on #235 to merge, as there is a potential undefined behavior issue there related to `_drop` routines.

# How to test the change?

There are some breaking changes for profiling* crates in here. Upgrade, then test as usual.

## For Reviewers
- [ ] If this PR touches code that signs or publishes builds or packages, or handles credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.
- [x] This PR doesn't touch any of that.

PR #266, opened 2023-10-03
Bump libdatadog version to 5.0.0 in preparation for release

ivoanjo:
**What does this PR do?**

This PR bumps the libdatadog version to 5.0.0. This PR looks a bit different from previous PRs to bump libdatadog version [(here's the 4.0.0 PR as an example)](https://github.com/DataDog/libdatadog/pull/236) because in [this PR](https://github.com/DataDog/libdatadog/pull/247) we've centralized the version on the single `Cargo.toml` file.

Furthermore, we're going from 4.0.0 to 5.0.0 because there were a number of backwards-incompatible changes to the profiling APIs.

**Motivation:**

Release libdatadog 5.0.0.

**Additional Notes:**

If I haven't missed anything, the backwards incompatible API changes were the following:

* The value of the `end_timestamp_ns` label is now provided as a regular argument to `ddog_prof_Profile_add`
* The libdatadog 5 serializer outputs compressed pprof files
* The exporter has a new API that takes two lists, a list of files to compress and a list of files to assume are compressed when exporting
* The libdatadog 5 serializer now resets profiles as part of serializing them
* The `ddog_prof_Profile_new` now returns a result structure

**How to test the change?**

I've tested the libdatadog 5 releases using the Ruby profiler, see https://github.com/DataDog/dd-trace-rb/pull/3169 for my draft PR.

**For Reviewers**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -595,7 +595,7 @@ dependencies = [
 
 [[package]]
 name = "datadog-profiling"
-version = "4.0.0"
+version = "5.0.0"
 dependencies = [
  "anyhow",
  "bitmaps",
@@ -626,7 +626,7 @@ dependencies = [
 
 [[package]]
 name = "datadog-profiling-ffi"
-version = "4.0.0"
+version = "5.0.0"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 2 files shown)


ivoanjo:
> There's a failing test because the serverless's protos are out of sync again. I don't consider this a blocker, personally.

:+1: Yeah, I'll drop a note to the serverless folks to look into it, but indeed that part of the code is not part of the binary releases for profiling so I agree it doesn't make sense to block on it.

PR #268, opened 2023-10-04
[ddtelemetry] Support metric distributions with ddsketch

paullegranddc:
# What does this PR do?
- Add minimal ddsketch implementation
- Add support for ddsketch in ddtelemetry

# Motivation
DDSketches are smaller than the current distribution implementation.

# Additional Notes
The [PR](https://github.com/DataDog/dd-go/pull/105925) to add support for ddsketches in the telemetry-intake needs to merged first.

# How to test the change?
The example `ddtelemetry::examples::tm-send-sketch` can be use to send sketches to staging with
`DD_API_KEY=*** cargo run --package ddtelemetry --example tm-send-sketch`
Results can be seen here https://ddstaging.datadoghq.com/notebook/3695892/test-instrumentation-telemetry-distribution?range=3600000&view=view-mode&start=1718366241675&live=true


Cargo.toml
```diff
@@ -28,7 +28,8 @@ members = [
   "serverless",
   "bin_tests",
   "data-pipeline",
-  "data-pipeline-ffi"
+  "data-pipeline-ffi",
+  "ddsketch",
```
pierotibou, line 32: What's the impact on the size of the binary? (I should add an automatic comment for any modif in that file :p)

ddsketch/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "datadog-ddsketch"
+authors = ["Paul Le Grand des Cloizeaux <paul.legranddescloizeaux@datadoghq.com>"]
```
pierotibou, line 3: You should add your name as well @VianneyRuhlmann

ddsketch/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
```
pierotibou, line 1: I believe we said we wanted to improve the state of the documentation, starting by improving the `description` field of our crates

ddsketch/build.rs
```diff
@@ -0,0 +1,59 @@
...
+// Datadog, Inc.
+
+// This file has been automatically generated from build.rs
+
+";
+
+fn main() -> Result<(), Box<dyn Error>> {
+    #[cfg(feature = "generate-protobuf")]
+    {
+        // protoc is required to compile proto files. This uses protobuf_src to compile protoc
```
pierotibou, line 30: is this automatically handle by cargo? Do we need to improve internal or public (e.g in repo) doc to ease building locally?

ddtelemetry/examples/tm-metrics-worker-test.rs
```diff
@@ -52,7 +52,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let dist_metric = handle.register_metric_context(
         "test_telemetry.dist".into(),
         Vec::new(),
-        data::metrics::MetricType::Distribution,
+        data::metrics::MetricType::Sketch,
```
pierotibou, line 55: nitpicking - Isn't it still a distribution?

ddtelemetry/src/worker/mod.rs
```diff
@@ -601,8 +607,8 @@ impl TelemetryWorker {
                 }
             }
             AppHeartbeat(()) | AppClosing(()) => {}
-            // TODO Paul lgdc keep metrics until we know if the flush was a success
-            GenerateMetrics(_) | Distributions(_) => {}
+            // TODO: Paul lgdc keep metrics until we know if the flush was a success
```
pierotibou, line 610: Should we add a ticket for this in the backlog? Is this a real issue? Aren't we ok to drop one payload (as long as we log, at least in debug, that we drop them)

pierotibou (commented):
Thanks a lot for doing this. 
I just have a few comments/questions. Haven't looked at the Rust implem though as usual.

ddsketch/build.rs
paullegranddc: If the proto files were to be modified, you would need to `cargo build --feature generate-protobuf` to generate the corresponding rust code, but it adds significant compile time, so it's hidden behind a feature.

It's the same mechanism as the tracer-payloads crate in this repo btw

ddsketch/Cargo.toml
ekump: I don't think we should be using the author attribute here at all, actually.

ddsketch/Cargo.toml
ekump, line 1: Since this is a new crate, the [CODEOWNERS](https://github.com/DataDog/libdatadog/blob/main/.github/CODEOWNERS) file should probably be updated as well. I assume @DataDog/libdatadog-telemetry should own it.

paullegranddc: I think both @Datadog/libdatadog-apm and @DataDog/libdatadog-telemetry  since it's going to be used for stats computation. This should require a review from at least one of the two groups, not both looking at [this](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file).

ddsketch/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+        bins
+    }
+
+    /// Add a point with value `point` to the sketch
+    pub fn add(&mut self, point: f64) -> Option<()> {
+        self.add_with_count(point, 1.0)
+    }
+
+    /// Add `count` point with value `point` to the sketch
+    pub fn add_with_count(&mut self, point: f64, count: f64) -> Option<()> {
```
ekump, line 55: Is there a particular reason that an empty Option was used as the return type? It feels non-idiomatic to me. If I were a caller of this function and the add didn't succeed because I passed in invalid inputs I'd expect a `Result<(), Error>` and for the error to explain why (even if it's just a generic "invalid input" error).

ddsketch/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+            return None;
+        }
+        Some(Self {
+            bins: VecDeque::new(),
+            offset: 0,
+            max_size,
+        })
+    }
+
+    /// Return an iterator over the bins
```
ekump, line 129: nitpick: I prefer to use regular comments over rustdoc comments for private functions. You wouldn't want this exposed if you generate docs.

ddsketch/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+                self.max_size as usize - self.bins.len(),
+            );
+
+            self.bins.reserve(additional_low_bins);
+            for _ in 0..additional_low_bins {
+                self.bins.push_front(0.0);
+            }
+
+            self.offset -= additional_low_bins as i32;
+            #[allow(clippy::needless_return)]
```
ekump, line 171: Is this clippy allow needed? Can't you just replace `return 0;` with `0` ?

ddsketch/src/lib.rs
paullegranddc: Private functions aren't printed in rustdocs I think

ddsketch/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+        if gamma <= 1.0 {
+            return None;
+        }
+        let multiplier = Self::multiplier_from_gamma(gamma);
+        Some(Self {
+            gamma,
+            multiplier,
+            min_indexable_value: max(
+                // So that the value representing the lowest bucket is >= std::f64::MIN_POSITIVE
+                std::f64::MIN_POSITIVE * gamma,
```
ekump, line 236: Just an FYI, it looks like `std::f64::MIN_POSITIVE` has been deprecated in favor of `f64::MIN_POSITIVE`.

ddsketch/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+    }
+
+    /// Returns the index of the bucket containing `value`
+    fn index(&self, value: f64) -> i32 {
+        (value.ln() * self.multiplier + self.index_offset).floor() as i32
+    }
+
+    /// Returns the value representing the bucket at `index`
+    fn value(&self, index: i32) -> f64 {
+        ((index as f64 - self.index_offset) / self.multiplier).exp()
```
ekump, line 261: Just curious, is there no need to be concerned with floating point math accuracy for ddsketch? Are we safe to not use an arbitrary-precision floating point library like BigDecimal?

ddsketch/src/lib.rs
ekump: Ah, I just tried it and you're correct. It still comes through in IDEs, which is nice.

VianneyRuhlmann: You can also use `--document-private-items` to generate docs with it.

ddtelemetry/examples/tm-metrics-worker-test.rs
VianneyRuhlmann: You're right, I renamed it so it's called Distribution for public interfaces, and sketche is just an implementation detail

ddtelemetry/src/worker/mod.rs
VianneyRuhlmann: @paullegranddc ?

paullegranddc: i think's it's fine to lose metrics payloads from time to time

Cargo.toml
VianneyRuhlmann: The release artifacts are 293,2MB and the last build on main is 290,8MB

ddtelemetry/src/metrics.rs
```diff
@@ -148,7 +146,16 @@ impl MetricBuckets {
             series: self.series.len() as u32,
             series_points: self.series.values().map(|v| v.len() as u32).sum(),
             distributions: self.distributions.len() as u32,
-            distributions_points: self.distributions.values().map(|v| v.len() as u32).sum(),
+            distributions_points: self
```
ekump, line 149: It doesn't look like this public function is covered with unit tests.

rustfmt.toml
```diff
@@ -6,7 +6,7 @@ ignore = [
 ]
 
 format_macro_matchers = true
-
+format_generated_files = false
```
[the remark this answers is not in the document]
VianneyRuhlmann: This only applies to files with `@generated` in the header. It is useful for files generated with prost_build which generates files with a different line length for wrapping. This means that someone regenerating the file by mistake (e.g. by running --all-features) would modify this file. I think it is better to have it be idempotent and it seemed simpler to have a slightly different formatting in generated files than running rustfmt during the build.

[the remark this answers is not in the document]
VianneyRuhlmann: I just noticed that `trace-protobuf` is using #[rustfmt::skip] on the module declaration. I guess we should disable formatting in the same way everywhere.

PR #269, opened 2023-10-05
Run Miri everywhere

danielsn:
# What does this PR do?

1. Runs Miri in CI on all of the library, not just profiling
2. Explicitly "ignores" the tests where Miri cannot currently run

# Motivation

Miri is great for catching bugs, and as much of the library as possible should take advantage of that.

# Additional Notes

1. Miri does not run on Tokio tests for now: https://github.com/tokio-rs/tokio/issues/3360
2. Miri cannot be directly disabled on doc-tests, so I had to use a slightly ugly patch there https://github.com/rust-lang/miri/issues/584#issuecomment-825798646
3. None of the issues I saw appeared to be bugs: they generally related to syscalls that Miri doesn't currently support.  We should investigate enabling these tests as Miri advances

# How to test the change?
Should just run in the CI

## For Reviewers

### DIFF
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -3,7 +3,7 @@ on:
   push:
 
 jobs:
-  profiler-test:
+  run-miri:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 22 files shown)


morrisonlevi (approved):
Looks good to me, unless some specific team doesn't want to add `#[cfg_attr(miri, ignore)]` to every new test to some subdirectory.

ivoanjo (approved):
Looks really good! :+1:

PR #270, opened 2023-10-12
Sidecar: propagate environment to telemetry

Anilm3:
# What does this PR do?

Telemetry data sent through the sidecar currently appears in the dashboard as a service without an environment, this PR fixes it by propagating the environment all the way from the PHP extension to the telemetry worker builder.

# Motivation

Fixing OSS vulnerability detection on ASM services.

# How to test the change?

This will be tested through https://github.com/DataDog/dd-trace-php/pull/2313

## For Reviewers

### DIFF
--- a/sidecar/src/interface.rs
+++ b/sidecar/src/interface.rs
@@ -65,6 +65,7 @@ pub trait SidecarInterface {
         queue_id: QueueId,
         meta: RuntimeMeta,
         service_name: String,
+        env_name: String,
     );
     async fn set_session_config(session_id: String, config: SessionConfig);
     async fn shutdown_runtime(instance_id: InstanceId);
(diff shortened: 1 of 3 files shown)


sidecar/src/interface.rs
```diff
@@ -276,17 +277,19 @@ impl RuntimeInfo {
...
         let mut apps = self.apps.lock().unwrap();
-        if let Some(found) = apps.get(service_name) {
+        let key = (service_name.clone(), env_name.clone());
+        if let Some(found) = apps.get(&key) {
             (found.clone(), None)
         } else {
             let (future, completer) = ManualFuture::new();
             let shared = future.shared();
-            apps.insert(service_name.clone(), shared.clone());
+            apps.insert((service_name.clone(), env_name.clone()), shared.clone());
```
bwoebi, line 292:
```suggestion
            apps.insert(key, shared.clone());
```

bwoebi (approved):
Thanks Anil - LGTM :-)

PR #272, opened 2023-10-23
[Profiling] Put StackTraceId into the sample labels to simplify deduplication on the backend

danielsn:
# What does this PR do?

Adds a new sample label "dd_stack_id" whose value is the `StackTraceId` of the sample.

# Motivation

Timeline profiles create a huge number of samples with identical stack-traces, which require resources to process and deduplicate on the backend.  We've already done the work of deduplicating the stack traces, and have unique ids per trace.  Dumping these into the label allows the backend to avoid having to even look at stacktraces its seen before.

# Additional Notes

This change should not affect any other pprof based tools: they will simply ignore the unknown label.

This will slightly increase the size of the pprof file, since we're still sending the stacktrace for each sample.  We could imagine a larger optimization where we only attach the actual stacktrace to the first sample, but this would break any other pprof based tools.

# How to test the change?

* Correctness: added a test
* Performance: lets enable this for a test service, and see whether the backend can take advantage

## For Reviewers

### DIFF
--- a/profiling/src/internal/profile.rs
+++ b/profiling/src/internal/profile.rs
@@ -22,6 +22,7 @@ pub struct Profile {
     observations: Observations,
     period: Option<(i64, ValueType)>,
     sample_types: Vec<ValueType>,
+    stack_id_key: StringId,
     stack_traces: FxIndexSet<StackTrace>,
     start_time: SystemTime,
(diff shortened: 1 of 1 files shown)


ivoanjo:
Duplicate of #278

PR #275, opened 2023-10-25
RFC: Allow verify trace-protobuf to continue on error

ivoanjo:
## What does this PR do?

This PR changes the "Verify trace-protobuf" github action to include "continue-on-error", so that unrelated PRs don't get marked as having failed CI when this action triggers.

Additionally, it makes use of the
`mainmatter/continue-on-error-comment@v1` action to turn these errors into a comment that gets added into the PR, so that we don't miss that this action failed -- we just make failures less noisy.

## Motivation:

I like having a green CI, and it's annoying to have a red CI because this action failed, since for many libdatadog PRs we don't care (e.g. this only impacts the serverless agent).

Thus, I decided to experiment with this alternative approach. As the PR title says -- comments welcome! Do you think this is reasonable?

## Additional Notes:

N/A

## How to test the change?

I'm hoping the action will start triggering for this PR itself, so we can see it running.

## For Reviewers


paullegranddc (approved):

paullegranddc:
It looks like the `Verify trace-protobuf` tests didn't run for this PR. Is this expected?

paullegranddc (changes requested):
https://github.com/DataDog/libdatadog/actions/runs/6638247018

It looks like the action is invalid

paullegranddc:
I added the action with the specific hash in this PR here https://github.com/DataDog/libdatadog/settings/actions

ivoanjo:
Ahh, thanks for spotting that -- I had added with the `v1` tag and forgot that to make it match with the hash.

ivoanjo:
It seems it has run now -- but doesn't output anything because it only runs on new PRs, not on branch updates ( https://github.com/DataDog/libdatadog/actions/runs/6642718147/job/18048114414?pr=275 ).

My suggestion is: Let's merge this in, and then right after I can open a tiny PR with a dummy fix to verify it's working fine. If it is, it stays, if not, I'll open a PR to revert it (or tweak it, if needed).

paullegranddc (approved):

ivoanjo:
Update: Almost got it! https://github.com/DataDog/libdatadog/pull/280 has a tiny fix and otherwise this seems to work fine.

PR #276, opened 2023-10-25
Fix clippy lint

paullegranddc:
# What does this PR do?

Fix a small clippy lint

# Motivation

The tarpc code has been vendored because we forked it with minimal changes, and the fact that package managers didn't want us to depend on other github repositories to distribute libdatadog.

I wish we had a way to disable the clippy lint just for tarpc, without having to diverge further from upstream but [it looks like it's not possible](https://github.com/EmbarkStudios/rust-ecosystem/issues/22).


ivoanjo (approved):
:+1: LGTM

PR #277, opened 2023-10-25
[Profiling] Put LabelSetId into the sample labels to simplify deduplication on the backend

danielsn:
# What does this PR do?

Adds a new sample label "dd_following_labels_belong_to_set" whose value is the `LabelSetId` of the sample.

# Motivation

Timeline profiles create a huge number of samples with identical label sets, which require resources to process and deduplicate on the backend.  We've already done the work of deduplicating the labelsets, and have unique ids.  Dumping these into the label allows the backend to avoid having to even look at labels its seen before.

# Additional Notes

This change should not affect any other pprof based tools: they will simply ignore the unknown label.

This will slightly increase the size of the pprof file, since we're still sending the labelset for each sample. 

This change makes labels positional: all labels following the "dd_following_labels_belong_to_set" label must indeed belong to the set.  This restriction does not seem onerous, but YMMV.

# How to test the change?

* Updated existing tests to ensure the correct presence of this label.
* Performance: lets enable this for a test service, and see whether the backend can take advantage

## For Reviewers

### DIFF
--- a/profiling/src/internal/profile.rs
+++ b/profiling/src/internal/profile.rs
@@ -586,7 +586,7 @@ mod api_test {
         check_label(profile, label, key, num, "", "");
     }
 
-    fn check_str_label(profile: &pprof::Profile, label: &pprof::Label, key: &str, str: &str) {
+    fn _check_str_label(profile: &pprof::Profile, label: &pprof::Label, key: &str, str: &str) {
         check_label(profile, label, key, 0, "", str);
     }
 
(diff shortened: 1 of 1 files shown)


profiling/src/internal/profile.rs
```diff
@@ -162,6 +164,7 @@ impl Profile {
         profile.endpoints.local_root_span_id_label = profile.intern("local root span id");
         profile.endpoints.endpoint_label = profile.intern("trace endpoint");
         profile.timestamp_key = profile.intern("end_timestamp_ns");
+        profile.label_set_start_key = profile.intern("dd_following_labels_belong_to_set");
```
ivoanjo, line 167: Minor: Up until now, most of our labels were separated by spaces, not underscores. The timeline timestamp label broke with this pattern, but I would suggest we don't keep making it more inconsistent from now on lol.

PR #281, opened 2023-11-16
Initial implementation of new Container Id

pawelchcki:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

## For Reviewers

### DIFF
--- a/ddcommon/src/container_id.rs
+++ b/ddcommon/src/container_id.rs
@@ -6,7 +6,9 @@ use regex::Regex;
 use std::error;
 use std::fmt;
 use std::fs::File;
+use std::io;
 use std::io::{BufRead, BufReader};
+use std::os::unix::fs::MetadataExt as _;
 use std::path::Path;
 use std::path::PathBuf;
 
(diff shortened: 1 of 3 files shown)


ddcommon/src/container_id.rs
```diff
@@ -107,12 +120,47 @@ pub fn get_container_id() -> Option<&'static str> {
...
+        let device = meta.dev();
+
+        Ok(CGroupSurrogateId {
+            device: device,
+            inode: inode,
+        })
+    }
+
+    pub fn read() -> Result<Self, ContainerIdError> {
+        Self::read_from_path(DEFAULT_SYS_CGROUP_PATH)
```
AliDatadog, line 155: I think we need to verify if `sys/fs/cgroup` is always the cgroup node of the container. In our tests it was always the case but IIRC the spec says that`/proc/<pid>/cgroup` contains the exact path: https://man7.org/linux/man-pages/man7/cgroups.7.html

ddcommon/src/container_id.rs
```diff
@@ -107,12 +120,47 @@ pub fn get_container_id() -> Option<&'static str> {
...
+    }
+
+    pub fn as_header_value(&self) -> String {
+        let Self { device, inode } = self;
+        format!("id_v2://{device:x}h/{device}d {inode}")
```
AliDatadog, line 160:
```suggestion
        format!("dev://{device:x},inode://{inode}")
```
If I understand correctly, it will be sent in the header `Datadog-Entity-Id`. Perhaps we don't need the device in both hexa and decimal. Also for parsing perhaps it's better to avoid spaces. Finally we're not sure if `device` can be optional so maybe we can split them ? I would assume that a linux host will use the same device for all its cgroups.

PR #282, opened 2023-11-28
Implement crash-tracking handler

danielsn:
# What does this PR do?

This module implements a crash tracker that enables libdatadog using profilers to send crash-reports to the datadog backend.
 
It has three related parts:
1. A `CrashInfo` struct which stores information about a crash in a structured format, and has built-in functions to upload itself to the datadog backend.
   This struct can be generated by the crash-handlers described below (e.g. for python or Ruby), or by language runtime tools from a crash dump (e.g. .Net).

2. A UNIX crash-handler, which registers signal handlers to detect crashes.
   The legal operations within a crash-handler are limited, so the handler does as much out of process as possible.
   When a crash occurs, this handler collects relevant information, and puts it on a pipe to the receiver, which processes it out of process.

3. A binary application `libdatadog-crashtracking-receiver`, which receives the crash-report over a pipe and formats it for transmission to the backend.

# Motivation

Currently, if crashes occur in profiling code, there is no notification until we get a support ticket.  Learning early will enable us to close the loop to faster customer resolution.

# Additional Notes

There are a number of open TODOs: see the README.md

# How to test the change?

Run the test in profiling/crashtracker/api

## For Reviewers
- [x] If this PR touches code that signs or publishes builds or packages, or handles credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.


sanchda (approved):
Mostly discussed offline, but LGTM!  The only question I had was about whether stdio might get directed from the shell into the crash handler if the underlying application as e.g., on a REPL.  Typically the way this would be avoided is by disconnecting stdin in the child (close()), creating a socket pair, then using something like dup2 to reposition the read end of that pair into FD 0 (you may need to create the pair, _then_ close stdin, _then_ dup2).  It's possible that this is already being done and I just missed it.


EDIT:  keeping my comment to document my ignorance, but it was pointed out to me that the fork is called with `.stdin(Stdio::piped())`, which addresses my thought.

sanchda (approved):
Looking pretty good!  Thank you!

I'm not 100% sure how the CMake will work out, but I'm happy to accept things as-is.  I think there might be value in revealing the source file through a variable; that way I can include this cmake in order to find the sourcefile, then add my own linker stuff (for instance, setting a runpath and giving it a wrapper library with suitably exposed symbols).

But again, I don't know and I'll have to play around with it to see how everything fits together.  I can easily infer the source location based on other hints, so my comment here is probably non-actionable.

The rest still looks good.

PR #285, opened 2023-12-12
Export crashes through telemetry

paullegranddc:
# What does this PR do?

Crashtracker additions:
* Add a telemetry uploader for crashes
* Send part of the crash info though tags, and part of it as the message.
* Extract necessary telemetry tags from profiler tags

Logs payload addition:
* Add the "tags" field to telemetry logs to be able to search on fields in crashes
* Add a "sensitive" flag to telemetry logs to mark crash logs as possibly containing customer data

# Additional Notes

Currently, I implemented the telemetry uploader in the same crate as the crashtracker binary, because the uploader has dependencies on the crash info code in in the profiling crate, and the ddtelemetry crate.

If we implement crashtracking in situations where the profiler lib is not running, we might want to move the crash_info code to ddcommon.

## For Reviewers


ddtelemetry/src/data/payloads.rs
```diff
@@ -72,6 +72,11 @@ pub struct Log {
     pub level: LogLevel,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub stack_trace: Option<String>,
+    #[serde(skip_serializing_if = "String::is_empty")]
```
danielsn, line 75: Nice, I should use this too :)

ddtelemetry/src/worker/mod.rs
```diff
@@ -675,6 +675,8 @@ impl TelemetryWorkerHandle {
                 message,
                 level,
                 stack_trace,
+                tags: String::new(),
+                is_sensitive: false,
```
danielsn, line 679: should this default to false? Where would we have `is_sensitive: true`?

profiling/src/crashtracker/crash_info.rs
```diff
@@ -30,20 +30,20 @@ pub struct StackFrame {
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct SigInfo {
-    signum: u64,
-    signame: Option<String>,
+    pub signum: u64,
```
danielsn, line 33: why do these need to be `pub`

profiling-crashtracking-receiver/src/main.rs
```diff
@@ -36,6 +39,9 @@ pub fn main() -> anyhow::Result<()> {
                 // TODO Experiment to see if 30 is the right number.
                 crash_info.upload_to_dd(endpoint, Duration::from_secs(30))?;
             }
+            if let Some(uploader) = telemetry_uploader {
+                uploader.upload_to_telemetry(&crash_info, Duration::from_secs(30))?;
```
danielsn, line 43: Should we make 30 seconds configurable? Or is that a standard value?

profiling-crashtracking-receiver/src/main.rs
```diff
@@ -25,6 +26,8 @@ pub fn main() -> anyhow::Result<()> {
     std::io::stdin().lock().read_line(&mut metadata)?;
     let metadata: Metadata = serde_json::from_str(&metadata)?;
 
+    let telemetry_uploader = telemetry::TelemetryCrashUploader::new(&metadata, &config).ok();
```
danielsn, line 29: Do we want to do something if there is an error? Or just elide the upload?

ddtelemetry/src/data/payloads.rs
paullegranddc: Actually I think I missed a matching `#[serde(default)]` annotation, because otherwise the field will be considered as missing on deserialization

profiling/src/crashtracker/crash_info.rs
paullegranddc: I'm reading them from the profiling-crashtracking-receiver crate to create the telemetry message

profiling-crashtracking-receiver/src/main.rs
paullegranddc: I'm not sure, there needs to be a timeout, but 30s seems too much for me since we're blocking the parent process until we return, but I think the evp intake the data goes too after being proxied by the agent out can have pretty long outliers from time to time (requests blocking for up to 15s in rare cases).

But I don't know if it's a good idea to let users modify it 🤔 Maybe the crashtracker should daemonize itself and do crash submission asynchronously after we receive it, and then we can keep the long timeout.

profiling-crashtracking-receiver/src/main.rs
paullegranddc: I'd say elide the upload. Crashtracking is best effort anyway

ddtelemetry/src/worker/mod.rs
paullegranddc: I set it to false because the existing behavior is that telemetry logs are not assumed to contain sensitive data.

profiling-crashtracking-receiver/src/main.rs
danielsn: The issue which I'm worried about is the pod being killed when the parent process dies, and the crash-tracker info never getting out.

profiling-crashtracking-receiver/src/main.rs
danielsn: We could have a metric for failed uploads. 🐢 all the way down!

profiling-crashtracking-receiver/src/main.rs
paullegranddc: Hmm true, and if the crash tracker daemonize itself if the crashing process is not PID 1 in the container we might end up with zombie processes

profiling-crashtracking-receiver/src/main.rs
paullegranddc: I'm a bit afraid that if the crashtracking fails, metrics wouldn't work although I guess sending a single statsd point wouldn't hurt

profiling-crashtracking-receiver/src/main.rs
paullegranddc: Also, if we block the process from exiting for a long time we might prevent any orchestrator from restarting it and worsen the impact a crash might have on availability

bin_tests/src/bin/test_the_tests.rs
```diff
@@ -0,0 +1,8 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+use std::{thread, time};
+
+fn main() {
```
danielsn, line 6: Comment explaining what this is testing, and why we need the sleep

bin_tests/src/bin/crashtracker_bin_test.rs
```diff
@@ -0,0 +1,39 @@
...
+fn main() -> anyhow::Result<()> {
+    let args = env::args().collect::<Vec<_>>();
+    let output_filename = &args[1];
+    let receiver_binary = &args[2];
+    crashtracker::init(
+        Configuration {
+            create_alt_stack: true,
+            endpoint: None,
+            output_filename: Some(output_filename.clone()),
+            path_to_receiver_binary: receiver_binary.clone(),
```
danielsn, line 20: Why are these clones needed here?

profiling-crashtracking-receiver/src/telemetry.rs
```diff
@@ -0,0 +1,284 @@
...
+            "http://localhost:8126/telemetry/proxy/api/v2/apmtelemetry"
+        );
+    }
+
+    #[test]
+    fn test_crash_request_content() {
+        let tmp = tempfile::tempdir().unwrap();
+        let output_filename = {
+            let mut p = tmp.into_path();
+            p.push("crahs_info");
```
danielsn, line 244: typo

profiling-crashtracking-receiver/src/telemetry.rs
```diff
@@ -0,0 +1,284 @@
...
+        let payload: serde_json::value::Value =
+            serde_json::de::from_str(&fs::read_to_string(&output_filename).unwrap()).unwrap();
+        assert_eq!(payload["request_type"], "logs");
+        assert_eq!(payload["application"]["service_name"], "foo");
+        assert_eq!(payload["application"]["language_name"], "native");
+
+        assert_eq!(payload["payload"].as_array().unwrap().len(), 1);
+        assert_eq!(
+            payload["payload"][0]["tags"],
+            "uuid:1d6b97cb-968c-40c9-af6e-e4b4d71e8781,signum:11,signame:SIGSEGV"
```
danielsn, line 280: Is the ordering of tags deterministic?

bin_tests/src/lib.rs
```diff
@@ -0,0 +1,175 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
```
danielsn, line 1: Why is this machinery necessary?

bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -0,0 +1,53 @@
...
+    assert!(!exit_status.success());
+
+    // Check the crash data
+    let crash_profile = fs::read(crash_profile_path).unwrap();
+    let prof = serde_json::from_slice::<serde_json::Value>(&crash_profile).unwrap();
+    prof.as_object().unwrap();
+
+    let crash_telemetry = fs::read(crash_telemetry_path).unwrap();
+    let telemetry = serde_json::from_slice::<serde_json::Value>(&crash_telemetry).unwrap();
+    telemetry.as_object().unwrap();
```
danielsn, line 52: Check that we have expected values here?

bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -0,0 +1,53 @@
...
+fn test_crash_tracking_bin() {
+    let crash_tracking_receiver = ArtifactsBuild {
+        name: "profiling-crashtracking-receiver".to_owned(),
+        profile: Profile::Debug,
+        artifact_type: ArtifactType::ExecutablePackage,
+        triple_target: None,
+    };
+    let crashtracker_bin = ArtifactsBuild {
+        name: "crashtracker_bin".to_owned(),
+        profile: Profile::Debug,
```
danielsn, line 22: Might be interesting to do a prop-test covering both debug and release builds?

bin_tests/tests/test_the_tests.rs
```diff
@@ -0,0 +1,42 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2023-Present Datadog, Inc.
+
+use std::{fs, process};
+
+use bin_tests::{build_artifacts, ArtifactType, ArtifactsBuild, Profile};
+
+#[test]
+#[cfg_attr(miri, ignore)]
```
danielsn, line 9: no kidding!

bin_tests/src/bin/crashtracker_bin_test.rs
paullegranddc: They're not, I have refactored a bit

profiling-crashtracking-receiver/src/telemetry.rs
paullegranddc: Yes

bin_tests/src/lib.rs
paullegranddc: Added a comment at the top of the module describing what it does

>/// This module implements an abstraction over compilation with cargo with the purpose
/// of testing full binaries or dynamic libraries, instead if just rust static libraries.
///
/// The main entrypoint is `fn build_artifacts` which takes a list of artifacts to build,
/// either executable crates, cdylib, or extra binaries, invokes cargo and return the path
/// of the built artifact.
///
/// Builds are cached between invocations so that multiple tests can use the same artifact
/// without doing expensive work twice.
///
/// It is assumed that functions in this module are invoked in the context of a cargo #[test]
/// item, or a `cargo run` command to be able to locate artifacts built by cargo from the position
/// of the current binary.

bin_tests/tests/crashtracker_bin_test.rs
paullegranddc: Added.
Interestingly, the null pointer dereference gets optimized out in release mod, and we receive a SIGTRAP instead.  I guess it's because it's an undefined behavior

profiling/Cargo.toml
```diff
@@ -18,6 +18,7 @@ bitmaps = "3.2.0"
 bytes = "1.1"
 chrono = {version = "0.4", default-features = false, features = ["std", "clock", "serde"]}
 ddcommon = {path = "../ddcommon"}
+ddtelemetry = {path = "../ddtelemetry"}
```
danielsn, line 21: The crashtracker was moved to its own package, this might no longer be necessary

bin_tests/src/bin/crashtracker_bin_test.rs
```diff
@@ -0,0 +1,65 @@
...
+    #[inline(never)]
+    unsafe fn deref_ptr(p: *mut u8) {
+        *std::hint::black_box(p) = std::hint::black_box(1);
+    }
+
+    pub fn main() -> anyhow::Result<()> {
+        let mut args = env::args().skip(1);
+        let output_filename = args
+            .next()
+            .ok_or(anyhow::anyhow!("Unexpected number of arguments"))?;
```
danielsn, line 30:
```suggestion
            .context("Unexpected number of arguments")?;
```

bin_tests/src/bin/crashtracker_bin_test.rs
```diff
@@ -0,0 +1,65 @@
...
+        crashtracker::init(
+            CrashtrackerConfiguration {
+                create_alt_stack: true,
+                endpoint: Some(datadog_profiling::exporter::Endpoint {
+                    url: ddcommon::parse_uri(&format!("file://{}", output_filename))?,
+                    api_key: None,
+                }),
+                path_to_receiver_binary: receiver_binary,
+                resolve_frames: crashtracker::CrashtrackerResolveFrames::ExperimentalInProcess,
```
danielsn, line 42: This may crash.  Might be better to use `Never` for now

bin_tests/tests/test_the_tests.rs
```diff
@@ -0,0 +1,48 @@
...
+    test_the_tests_inner(Profile::Debug);
+}
+
+#[test]
+#[cfg_attr(miri, ignore)]
+fn test_the_tests_release() {
+    test_the_tests_inner(Profile::Release);
+}
+
+fn test_the_tests_inner(profile: Profile) {
```
danielsn, line 22: It took me a ⏱️ to realize that `Profile` is not the same as `Profile` in profiling.  Maybe use a different name?

danielsn (commented):
Looks pretty good.  Should we add a CI step?

bin_tests/tests/test_the_tests.rs
paullegranddc: Changed to `BuildProfile`

bin_tests/src/bin/crashtracker_bin_test.rs
paullegranddc: It in fact, does not pass the tests on the CI with `ExperimentalInProcess`, crashes the process in debug mode and produces an invalid json when receiving the crash report in release mode.

crashtracker/src/telemetry.rs
```diff
@@ -0,0 +1,313 @@
...
+            }
+        }
+
+    };
+}
+
+#[derive(Debug, serde::Serialize)]
+/// This struct represents the part of the crash_info that we are sending in the
+/// log `message` field as a json
+struct TelemetryCrashInfoMessage<'a> {
```
danielsn, line 44: Should the `additional_stacktraces` go in here as well?

paullegranddc: I'd prefer adding it in a followup PR

danielsn (approved):

PR #287, opened 2023-12-14
Fix codewoner ordering

paullegranddc:
# What does this PR do?

In codewoners file, the last entry takes precedence so any entry before appearing before the wildcard is ignored, which is probably not the expected behaviour.

# Motivation

# Additional Notes

# How to test the change?

## For Reviewers


morrisonlevi (approved):

ivoanjo (approved):
:+1: LGTM

Thanks for spotting this. Indeed in https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file this exact situation is mentioned as well.

PR #289, opened 2024-01-02
[PROF-8037] System info support for pprof profiles

AlexJF:
# What does this PR do?

Enables the upload of system info data for pprof-based profiles following the logic described in Datadog's internal RFC 'Pprof System Info Support`.

# Motivation

Java profiles have had system info support for a long time. This PR is one step in enabling non-java profiles to have it as well:

image

The system info mechanism is also being used as a source for application start time, unlocking the showing of app uptime for pprof profiles as well (notice the header in the screenshot).

# Additional Notes

* This adds a new param so it's a breaking change that will need updating on profilers that make use of libdatadog.
* For consistency and simplicity I followed a similar design to https://github.com/DataDog/libdatadog/pull/181 but open to suggestions given the slightly more structured nature of system info (i.e. there's an expectation of usage of at least one inner map, with the outer map keys representing info categories). Do we feel adhering to the RFC is good enough or would we want to solidify the structure a bit more in code?

# How to test the change?

Testing in combination with https://github.com/DataDog/dd-trace-rb/pull/3357

## For Reviewers

### DIFF
--- a/profiling/src/crashtracker/crash_info.rs
+++ b/profiling/src/crashtracker/crash_info.rs
@@ -179,7 +179,17 @@ impl CrashInfo {
             tags,
             endpoint,
         )?;
-        let request = exporter.build(time, time, &[crash_file], &[], None, None, None, None, timeout)?;
+        let request = exporter.build(
+            time,
+            time,
+            &[crash_file],
+            &[],
+            None,
+            None,
+            None,
+            None,
+            timeout,
+        )?;
         let response = exporter.send(request, None)?;
         //TODO, do we need to wait a bit for the agent to finish upload?
         Ok(response)


profiling-ffi/src/exporter.rs
```diff
@@ -283,6 +293,25 @@ unsafe fn parse_internal_metadata_json(
...
+                Ok(parsed) => Ok(Some(parsed)),
+                Err(error) => Err(anyhow::anyhow!(
+                    "Failed to parse contents of system_info json string (`{}`): {}.",
+                    json,
+                    error
+                )),
+            }
+        }
+    }
+}
```
ivoanjo, line 313: Minor: It's maybe worth having a `parse_json` that is used for both system info and internal metadata -- the only practical difference on them seems to be the name of the field ("internal_metadata"/"system_info") on the error string.

examples/ffi/exporter.cpp
```diff
@@ -139,9 +139,13 @@ int main(int argc, char *argv[]) {
   ddog_CharSlice internal_metadata_example = DDOG_CHARSLICE_C(
       "{\"no_signals_workaround_enabled\": \"true\", \"execution_trace_enabled\": \"false\"}");
 
+  ddog_CharSlice system_info_example = DDOG_CHARSLICE_C(
+      "{\"application\": {\"start-time\": \"2023-09-08 13:45:29.415742 UTC\"}, \"platform\": {\"kernel\": \"Darwin Kernel 22.5.0\"}}");
```
ivoanjo, line 143: I noticed that in the RFC, we're using `start_time` and not `start-time`. It may be worth dropping some guidance in the comments for these things?

(Also nitpick: should we use ISO8601 strings for dates?)

ivoanjo (approved):
Looks reasonable to me!

profiling/tests/form.rs
```diff
@@ -16,7 +16,11 @@ fn open<P: AsRef<Path>>(path: P) -> Result<Vec<u8>, Box<dyn Error>> {
     Ok(buffer)
 }
 
-fn multipart(exporter: &ProfileExporter, internal_metadata: Option<serde_json::Value>) -> Request {
+fn multipart(
+    exporter: &ProfileExporter,
+    internal_metadata: Option<serde_json::Value>,
+    system_info: Option<serde_json::Value>,
```
paullegranddc, line 22: Maybe passing the `internal_metadata` and `system_info` as [`Box<serde_json::RawValue>`](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) would be more efficient, because you can keep the jsons in String format.

AlexJF: Probably. But it also feels like it may hide issues with json definitions in test a bit further down in the stacktrace rather than at the json definition part. Since this only applies for tests where efficiency is not such a priority imo, I decided to leave as is.

ivoanjo (approved):
:+1: LGTM

AlexJF:
Validated alongside https://github.com/DataDog/dd-trace-rb/pull/3357: 
image

PR #292, opened 2024-01-11
[profiling] Forward Blazesym's symbolization APIs

r1viollet:
# What does this PR do?

Forward C-APIs from blazesym.
There is an internal RFC doc around the choice of blazesym

# Motivation

Within `ddprof` we will switch to using blazesym.

# Additional Notes

In a similar way we can choose to forward the normalize APIs.

# How to test the change?

There is a C test within the repository.
Blazesym also has some tests within the repository

## For Reviewers


examples/ffi/symbolizer.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "datadog/symbolizer.h"
+
+#define _THIS_IP_                                                              \
+  ({                                                                           \
+    __label__ __here;                                                          \
+  __here:                                                                      \
+    (unsigned long)&&__here;                                                   \
+  })
+
+// Virtual only works as privileged user
```
r1viollet, line 16: @danielsn I am puzzled as to why this would not work without specific privileges, even within the same process. Have you investigated this API by any chance?

symbolizer-ffi/README.md
```diff
@@ -0,0 +1,4 @@
+# Symbolizer-ffi
+
+This forwards parts of the blazesym C APIs.
+Please ensure changes to this folder can be upstreamed to [blazesym](https://github.com/libbpf/blazesym).
```
danielsn, line 4: Is there a reason we are doing this here rather than upstreaming directly?

danielsn (commented):
Mostly looks good.  A few high-level comments:
1. How is this tested?
2. Is there a plan to upstream this?

symbolizer-ffi/README.md
r1viollet: So we should upstream all changes.

The capis are here: 
https://github.com/libbpf/blazesym/blob/main/capi/Cargo.toml

This is just taking the bits that we need. Progressively, we can remove these in favour of higher level APIs with better integration with our current C APIs. However I think this is a good first step to see how good the library is.

We can not compose the C APIs (due to the current Cargo definition of how cbindgen works).

.github/workflows/test.yml
```diff
@@ -106,10 +106,19 @@ jobs:
       - name: "Test building C bindings"
         shell: bash
         run: |
+          set -e
           mkdir examples/ffi/build
           cd examples/ffi/build
-          cmake -S .. -DDatadog_ROOT=$OUTPUT_FOLDER/profiling
-          cmake --build .
+          # Add BUILD_SYMBOLIZER variable only for Linux platforms
+          if [[ "${{ matrix.platform }}" == "ubuntu-latest" ]]; then
```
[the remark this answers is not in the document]
r1viollet: The matrix does not cover these for now.
I would need to fix the privileged behaviour before I can push tests to gitlab where we run these platforms. Gitlab will not allow sudo.

symbolizer-ffi/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+#[cfg(not(target_os = "windows"))]
+#[allow(non_camel_case_types)]
```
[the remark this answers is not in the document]
r1viollet: Adding a comment

.github/workflows/lint.yml
```diff
@@ -41,7 +41,8 @@ jobs:
       - name: Install licensecheck
         run: sudo apt-get install -y licensecheck
       - name: Check licenses
-        run: '! find . \( -name "*.rs" -o -name "*.c" -o -name "*.sh" \) -not -path "*/tarpc/*" | xargs licensecheck -c ".*" | grep -v "Apache License 2.0"'
+        # Exclude symbolizer-ffi from the checks (mostly imported code)
```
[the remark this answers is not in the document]
r1viollet: This falls under bsd-3 license.

The bsd-3 license is generally more permissive than our Apache v2 license. We should be careful about the usual :
- Preserving copyright and licensing. Our 3rd party file ensures this.
- We are using the package at our own risk

symbolizer-ffi/src/symbolize.rs
```diff
@@ -0,0 +1,675 @@
...
+use blazesym::symbolize::Symbolizer;
+use blazesym::Addr;
+use std::ptr::NonNull;
+use std::slice;
+
+/// # Safety
+/// This function is unsafe because it creates a slice from a raw pointer.
+/// The caller must ensure that `items` points to a valid memory location
+/// and `num_items` does not exceed the number of elements `items` points to.
+pub unsafe fn slice_from_user_array<'t, T>(items: *const T, num_items: usize) -> &'t [T] {
```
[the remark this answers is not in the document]
r1viollet: miri runs on the upstream version of this

symbolizer-ffi/src/symbolize.rs
```diff
@@ -0,0 +1,675 @@
...
+        .as_ref()
+        .and_then(|info| info.column)
+        .unwrap_or(0);
+}
+
+/// Convert [`Sym`] objects to [`blaze_result`] ones.
+///
+/// The returned pointer should be released using [`blaze_result_free`] once
+/// usage concluded.
+fn convert_symbolizedresults_to_c(results: Vec<Symbolized>) -> *const blaze_result {
```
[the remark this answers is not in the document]
r1viollet: The upstream implementation has some amount of tests. I could not easily port these to our codebase.

Let's talk about how we can compose the APIs in a better way before I make some changes.

r1viollet:
Ruby pulls in the shared library. This would have an impact on ruby binaries.

examples/ffi/symbolizer.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "datadog/blazesym.h"
+
+#define _THIS_IP_                                                              \
+  ({                                                                           \
+    __label__ __here;                                                          \
+  __here:                                                                      \
+    (unsigned long)&&__here;                                                   \
+  })
+
+// Virtual only works as privileged user
```
r1viollet, line 16: this is not true anymore

profiling-ffi/Cargo.toml
```diff
@@ -14,8 +14,9 @@ license.workspace = true
 crate-type = ["staticlib", "cdylib"]
 
 [features]
-default = ["cbindgen"]
+default = []
```
r1viollet, line 17: I think that having a default that does not generate can make sense.
@paullegranddc wdyt ?

build-common/src/cbindgen.rs
```diff
@@ -67,8 +59,26 @@ pub fn generate_and_configure_header(header_name: &str) {
...
+/// Expects CARGO_MANIFEST_DIR to be set.
+/// If DESTDIR is set, it will be used as the base directory for the header file.
+/// DESTDIR can be either relative or absolute.
+/// Either CARGO_TARGET_DIR is set, or `cargo locate-project --workspace` is used to find the base
+/// of the target directory.
+///
+/// # Arguments
+///
+/// * `header_name` - The name of the header file to generate.
+pub fn generate_and_configure_header(header_name: &str) {
```
danielsn, line 77: Do we want this to panic on failure, or return an `Error`

build-profiling-ffi.sh
```diff
@@ -100,9 +103,14 @@ cp -v LICENSE LICENSE-3rdparty.yml NOTICE "$destdir/"
 export RUSTFLAGS="${RUSTFLAGS:- -C relocation-model=pic}"
 
 datadog_profiling_ffi="datadog-profiling-ffi"
-echo "Building the ${datadog_profiling_ffi} crate (may take some time)..."
+FEATURES="--features cbindgen"
+if [[ "$symbolizer" -eq 1 ]]; then
+    FEATURES="--features cbindgen,symbolizer"
+fi
+cargo build --features "cbindgen,symbolizer"
```
danielsn, line 110: Shouldn't this use `$FEATURES`

symbolizer-ffi/src/blazesym.h
```diff
@@ -0,0 +1,895 @@
+// BSD 3-Clause License
```
danielsn, line 1: SPDX header

symbolizer-ffi/src/blazesym.h
```diff
@@ -0,0 +1,895 @@
...
+// Synchronized from blazesym repository
```
danielsn, line 2: How will we make sure this stays in sync?

crashtracker/Cargo.toml
```diff
@@ -12,7 +12,7 @@ crate-type = ["lib"]
 
 [target.'cfg(unix)'.dependencies]
 # TODO: move to a named release once https://github.com/libbpf/blazesym/pull/520 makes it to release
```
danielsn, line 14: This comment is no longer needed

crashtracker/Cargo.toml
```diff
@@ -12,7 +12,7 @@ crate-type = ["lib"]
...
-blazesym = {git = "https://github.com/libbpf/blazesym.git", rev = "4fa15cb"}
+blazesym = {git = "https://github.com/libbpf/blazesym.git", rev = "v0.2.0-alpha.11"}
```
danielsn, line 15: No need to specify the git repo, since this is published to crates.io

.github/workflows/test.yml
```diff
@@ -118,10 +118,18 @@ jobs:
...
+            cmake -S .. -DDatadog_ROOT=$OUTPUT_FOLDER/profiling -DBUILD_SYMBOLIZER=true
```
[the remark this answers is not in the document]
r1viollet: I think this could be true. 
Having C generation and C tests within a build script sounds good.

If that is OK, I would prefer pushing this to the next PR.

build-common/src/cbindgen.rs
r1viollet: In a build script, I feel the panic option is reasonable, though I will defer to you on such questions.

build-profiling-ffi.sh
r1viollet: this is a forgotten test. Thank you for spotting it :facepalm: 
build line is bellow

crashtracker/Cargo.toml
r1viollet: Although this is true there might by value in keeping a consistent setting (in terms of version), and the C ffi parts are not systematically published.

danielsn (approved):
Thanks for doing this!

symbolizer-ffi/src/blazesym.h
r1viollet: I could do a clone / copy in a build script, but I don't like adding git steps to the build process.
Maybe if this is a conditional build step :thinking:

PR #294, opened 2024-01-25
refactor(profiling) Move crashtracker to its own package

danielsn:
# What does this PR do?

Moves the crash-tracker to its own top-level module

# Motivation

The crashtracker started as a profiling feature, but is of interest to tracing as well.  Moving to its own module to make it easier to import.

# Additional Notes

The crashtracker still uses the profiling exporter, so it has a dependency on profiling.  This will go away when we move to instrumentation telemetry as the export format.

# How to test the change?

## For Reviewers


r1viollet (approved):
LGTM
This is goes in the right direction :+1:

PR #295, opened 2024-01-29
Improve logging of the sidecar

bwoebi:
This allows dynamically updating the log level of the sidecar as well as tracking logs from telemetry and the sidecar operation itself in trace mode.

Further, given --cfg tokio-taskdump, the sidecar is able to dump a task snapshot of itself.

Notes: telemetry required the tracing-subscriber, but it's only for the example. Moved it to dev-dependencies.


sidecar/src/setup/unix.rs
```diff
@@ -58,15 +63,15 @@ impl Liaison for SharedDirLiaison {
             // failing to acquire lock
             // means that another process is creating the socket
             Err(err) => {
-                tracing::debug!("failed_locking");
+                debug!("failed_locking");
```
morrisonlevi, line 66: Maybe shift this to `warn!`? Seems more important than some of the other debugs like "already_listening" which should be fine.

sidecar/src/interface.rs
```diff
@@ -682,7 +726,9 @@ impl SidecarServer {
             None => return,
         };
 
-        session.shutdown().await
+        info!("Shutting down session: {}", session_id);
+        session.shutdown().await;
+        debug!("Successfully shut down session: {}", session_id);
```
morrisonlevi, line 731: It seems like maybe the first item should be `debug!` and the second should be `info!`?

morrisonlevi (commented):
Overall it looks good. I have some suggestions for possible changes to log levels, and this is minor of course.

I have a more serious question, though: why are only some things guarded behind `cfg(tracing)`? Seems like it might break if you build without the feature, but maybe they are protected somehow by code that's not in the diff.

bwoebi:
It's only the tracing_subscriber which must be guarded, everything else is always present.

sidecar/src/interface.rs
bwoebi: The intention was noting the moment in time where no new data is accepted for the session. The second is really just "successfully cleaned up".

sidecar/src/log.rs
```diff
@@ -0,0 +1,324 @@
...
+    let log_file = std::fs::File::options()
+        .create(true)
+        .truncate(false)
+        .write(true)
+        .append(true)
+        .open(path)?;
+    Ok(log_file)
+}
+
+pub struct TemporarilyRetainedMap<K, V>
```
bantonsson, line 38: Maybe the code and the logic in this file is self explanatory to everyone else, but I would welcome some comments outlining how things fit together. Especially the interaction between this code and the code in `tracing` that calls back into this code to compute the log levels.

Also, are there any tests for this functionality?

bwoebi: I agree, let me add a couple comments.

The sidecar code is mostly tested through tests from the dd-trace-php repository. I haven't written self-contained tests for the sidecar yet.

bantonsson: I would really like something as complex as a reference counting data structure (with the potential to leak memory) to have some unit tests.

bwoebi:
@bantonsson Added a test :-) Looks good now?

bantonsson (approved):

PR #296, opened 2024-01-30
Metrics sidecar shutdown

pawelchcki:
# What does this PR do?

- Add server.memory_usage telemetry metric
- Add a watchdog to terminate the sidecar upon reaching 1GB of memory usage
- Fix windows clippy complaints

## For Reviewers


sidecar/src/watchdog.rs
```diff
@@ -0,0 +1,198 @@
...
+            ),
+            active_sessions: worker.register_metric_context(
+                "server.active_sessions".to_string(),
+                vec![],
+                MetricType::Gauge,
+                true,
+                MetricNamespace::Sidecar,
+            ),
+            memory_usage: worker.register_metric_context(
+                "server.memory_usage".to_string(),
```
pawelchcki, line 171: @paullegranddc Hey Paul, do we have to add this new metric somewhere ?

sidecar/src/self_telemetry.rs
```diff
@@ -60,62 +64,93 @@ pub fn self_telemetry(
...
+                true,
+                MetricNamespace::Sidecar,
+            ),
+            memory_usage: worker.register_metric_context(
+                "server.memory_usage".to_string(),
+                vec![],
+                MetricType::Gauge,
+                true, // TODO: make sure its a common metric
+                MetricNamespace::Sidecar,
+            ),
```
paullegranddc, line 134: This is not going to do what you expect. We don't tag telemetry metrics by runtime-id, only by the cardinality of a service.
So your gauges are not going to be aggregated correctly. If we receive multiple points, only one of them is going to stay.

What you want is a distribution

sidecar/src/watchdog.rs
paullegranddc: Yes, you would have to add it here https://github.com/DataDog/dd-go/blob/b52550e0ca6462443f66286bc717dc635947a365/trace/apps/tracer-telemetry-intake/telemetry-metrics/static/common_metrics.json

sidecar/src/self_telemetry.rs
bwoebi: What is a distribution / how do I use that? Under MetricType I just see Gauge and Count as options.
Or are you talking about something that doesn't exist yet?

bwoebi (approved):

PR #297, opened 2024-01-30
[crashtracker] Enable dynamic updating of metadata and config at runtime

danielsn:
# What does this PR do?

Makes the `CrashtrackerMetadata` and `CrashtrackerConfiguration` dynamically updatable, rather than fixed at crashtracker startup.

# Motivation

Ruby allows the user to update tags on the fly, so we need to support that.

# Additional Notes

This will also be useful for a sidecar setup, where we would want to keep the metadata local until its time to upload.

# How to test the change?

Run the `test_crash` test in crashtracker/api, and notice that the extra tag is output in the crash report.

## For Reviewers


ivoanjo:
The way that the "endpoint" gets configured in Ruby is also dynamic, and can also change when configuration via code is used. Would it be possible to also allow that to be reconfigured?

crashtracker/src/crash_handler.rs
```diff
@@ -73,14 +74,29 @@ fn make_receiver(
...
+/// receiver when a crash occurs.
+///
+/// PRECONDITIONS:
+///     None
+/// SAFETY:
+///     Crash-tracking functions are not reentrant.
+///     No other crash-handler functions should be called concurrently.
+/// ATOMICITY:
+///     This function is not atomic. A crash during its execution may lead to
+///     unexpected crash-handling behaviour.
```
ivoanjo, line 93: Should the crash tracker maybe "suspend" itself or otherwise take care not to avoid issues here? E.g. in runtimes where multiple threads may be running, we can't guarantee that a crash won't happen in the middle of this update.

crashtracker/src/crash_handler.rs
```diff
@@ -142,6 +158,17 @@ extern "C" fn handle_posix_signal(signum: i32) {
...
+    writeln!(w, "{}", metadata)?;
+
+    writeln!(w, "{DD_CRASHTRACK_END_METADATA}")?;
+
+    Ok(())
+}
+
 fn handle_posix_signal_impl(signum: i32) -> anyhow::Result<()> {
     let mut receiver = match std::mem::replace(unsafe { &mut RECEIVER }, GlobalVarState::Taken) {
         GlobalVarState::Some(r) => r,
```
ivoanjo, line 174: One thing that occurred to me while reviewing the PR is -- should we have something protecting `handle_posix_signal_impl` so that it can't get called multiple times in the same process?

Afaik if a thread is handling a signal (e.g. `SIGSEGV`), no other SIGSEGVs can be delivered. But I... don't think that applies to other threads in the same process.

May be worth giving it a try?

crashtracker/src/crash_info.rs
```diff
@@ -156,7 +163,7 @@ impl CrashInfo {
         //let site = "datad0g.com";
         //let api_key = std::env::var("DD_API_KEY")?;
```
ivoanjo, line 164: Minor: :arrow_up: Maybe remove this test code?

crashtracker/src/crash_info.rs
```diff
@@ -105,6 +107,11 @@ impl CrashInfo {
         Ok(())
     }
 
+    pub fn set_metadata(&mut self, metadata: CrashtrackerMetadata) -> anyhow::Result<()> {
+        anyhow::ensure!(self.metadata.is_none());
+        self.metadata = Some(metadata);
+        Ok(())
+    }
     pub fn set_siginfo(&mut self, siginfo: SigInfo) -> anyhow::Result<()> {
```
ivoanjo, line 115: Minor: Kinda surprised rustfmt doesn't add a newline here?

```suggestion
    pub fn set_metadata(&mut self, metadata: CrashtrackerMetadata) -> anyhow::Result<()> {
        anyhow::ensure!(self.metadata.is_none());
        self.metadata = Some(metadata);
        Ok(())
    }
    
    pub fn set_siginfo(&mut self, siginfo: SigInfo) -> anyhow::Result<()> {
```

ivoanjo (commented):
Seems reasonable! Left a few notes :)

crashtracker/src/crash_handler.rs
danielsn: https://github.com/DataDog/libdatadog/issues/318 to track

crashtracker/src/crash_handler.rs
danielsn: Replaced with an atomic variable

crashtracker/src/crash_handler.rs
```diff
@@ -66,89 +62,133 @@ fn make_receiver(
...
-    config: &CrashtrackerConfiguration,
-    metadata: &CrashtrackerMetadata,
+    config: CrashtrackerConfiguration,
+    metadata: CrashtrackerMetadata,
 ) -> anyhow::Result<()> {
-    let new_receiver = make_receiver(config, metadata)?;
-    let old_receiver =
-        unsafe { std::mem::replace(&mut RECEIVER, GlobalVarState::Some(new_receiver)) };
+    let new_receiver = Box::into_raw(Box::new(make_receiver(config, metadata)?));
+    let old_receiver = RECEIVER.swap(new_receiver, SeqCst);
```
ivoanjo, line 124: Really minor: Since outside callers only see `setup_receiver` this note is more of a code arrangement thing.

These two lines

```
    let new_receiver = Box::into_raw(Box::new(make_receiver(config, metadata)?));
    let old_receiver = RECEIVER.swap(new_receiver, SeqCst);
```

suggest to me that the receiver is being set up atomically. Which in fact isn't the case -- the metadata and config are "global" even though they don't look like it.

So I'd suggest maybe not calling `update_metadata`/`update_config` directly in `make_receiver`, and instead doing something like:

```rust
update_config(config);
update_metadata(metadata);

let new_receiver = Box::into_raw(Box::new(make_receiver()?));
let old_receiver = RECEIVER.swap(new_receiver, SeqCst);
```

here directly?

crashtracker/src/crash_handler.rs
```diff
@@ -66,89 +62,133 @@ fn make_receiver(
...
     anyhow::ensure!(
-        matches!(old_receiver, GlobalVarState::Unassigned),
-        "Error registering crash handler receiver: receiver already existed {old_receiver:?}"
+        old_receiver.is_null(),
+        "Error registering crash handler receiver: receiver already existed"
     );
```
ivoanjo, line 128: If the semantic here is to say an existing receiver already existing is an error, should we do a compare and swap on `RECEIVER` instead, expecting it to be `null`?

Right now it's kinda weird that we successfully set up the new receiver, and only then we return an error.

crashtracker/src/crash_handler.rs
```diff
@@ -157,11 +197,34 @@ fn handle_posix_signal_impl(signum: i32) -> anyhow::Result<()> {
...
 
+fn handle_posix_signal_impl(signum: i32) -> anyhow::Result<()> {
+    // Leak receiver, config, and metadata to avoid calling 'drop' during a crash
+    let receiver = RECEIVER.swap(ptr::null_mut(), SeqCst);
+    anyhow::ensure!(!receiver.is_null(), "No crashtracking receiver");
+    let receiver = unsafe { receiver.as_mut().context("")? };
+
+    let config = CONFIG.swap(ptr::null_mut(), SeqCst);
+    anyhow::ensure!(!config.is_null(), "No crashtracking config");
+    let (config, config_str) = unsafe { config.as_mut().context("")? };
```
ivoanjo, line 214: Minor: Should we provide labels to `.context` for these too?

crashtracker/src/crash_handler.rs
```diff
@@ -157,11 +197,34 @@ fn handle_posix_signal_impl(signum: i32) -> anyhow::Result<()> {
...
+
+    let metadata_ptr = METADATA.swap(ptr::null_mut(), SeqCst);
+    anyhow::ensure!(!metadata_ptr.is_null());
```
ivoanjo, line 217: Minor: For consistency with the above approaches:

```suggestion
    anyhow::ensure!(!metadata_ptr.is_null(), "No metadata config");
```

crashtracker/src/crash_handler.rs
```diff
@@ -173,46 +236,43 @@ fn handle_posix_signal_impl(signum: i32) -> anyhow::Result<()> {
...
     // https://doc.rust-lang.org/std/process/struct.Child.html#method.wait
     // The stdin handle to the child process, if any, will be closed before waiting.
     // This helps avoid deadlock: it ensures that the child does not block waiting
     // for input from the parent, while the parent waits for the child to exit.
-    //TODO, use a polling mechanism that could recover from a crashing child
+    // TODO, use a polling mechanism that could recover from a crashing child
     receiver.wait()?;
+    // Calling "free" in a signal handler is dangerous, so don't do that.
     Ok(())
 }
```
ivoanjo, line 256: Minor: I'm not sure I understand what this comment is getting to?

crashtracker/src/crash_handler.rs
```diff
@@ -173,46 +236,43 @@ fn handle_posix_signal_impl(signum: i32) -> anyhow::Result<()> {
...
 
-pub fn register_crash_handlers(config: &CrashtrackerConfiguration) -> anyhow::Result<()> {
+// TODO, there is a small race condition here, but we can keep it small
```
ivoanjo, line 258: Minor: Consider explaining the race condition? (It's that the signal handlers aren't setup atomically, right?)

ivoanjo (commented):
Looks reasonable! I think the biggest note I have is around the design of the APIs e.g. exposing explicit `update` arguments vs allowing further calls to `init`.

(But if you're not convinced with that option, happy to approve as-is)

crashtracker/src/crash_handler.rs
danielsn: done

crashtracker/src/crash_handler.rs
```diff
@@ -230,34 +290,21 @@ unsafe fn register_signal_handler(signal_type: signal::Signal) -> anyhow::Result
...
+    let prev = OLD_HANDLERS.swap(ptr::null_mut(), SeqCst);
+    anyhow::ensure!(!prev.is_null());
+    // Safety: The only nonnull pointer stored here comes from Box::into_raw()
+    let prev = unsafe { Box::from_raw(prev) };
+    // Safety: The value restored here was returned from a previous sigaction call
+    unsafe { signal::sigaction(signal::SIGBUS, &prev.sigbus)? };
+    unsafe { signal::sigaction(signal::SIGSEGV, &prev.sigsegv)? };
+    // We want to avoid freeing memory inside the handler, so just leak it
+    // This is fine since we're crashing anyway at this point
+    if leak {
```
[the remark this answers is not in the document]
danielsn: done

profiling-ffi/src/crashtracker.rs
```diff
@@ -118,6 +118,46 @@ pub unsafe extern "C" fn ddog_prof_crashtracker_shutdown() -> ProfileResult {
...
+        )),
+    }
+}
+
+unsafe fn ddog_prof_crashtracker_update_metadata_impl(
+    metadata: CrashtrackerMetadata,
+) -> anyhow::Result<()> {
+    let metadata = metadata.try_into()?;
+    datadog_crashtracker::update_metadata(metadata)
+}
```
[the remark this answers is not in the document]
danielsn: done

crashtracker/src/crash_handler.rs
danielsn: I modified the semantics of registering a reciever

crashtracker/src/crash_handler.rs
danielsn: yes

crashtracker/src/crash_handler.rs
danielsn: done

crashtracker/src/crash_handler.rs
danielsn: Updated

crashtracker/src/crash_handler.rs
```diff
@@ -116,31 +117,57 @@ pub fn update_config(config: CrashtrackerConfiguration) -> anyhow::Result<()> {
...
+/// PRECONDITIONS:
+///     None
+/// SAFETY:
+///     Crash-tracking functions are not guaranteed to be reentrant.
+///     No other crash-handler functions should be called concurrently.
+/// ATOMICITY:
+///     This function uses a compare_and_exchange on an atomic pointer.
+///     If two simultaneous calls to this function occur, the first will win,
+///     and the second will cleanup the redundant receiver.
+pub fn update_reciever_after_fork(config: CrashtrackerConfiguration) -> anyhow::Result<()> {
```
ivoanjo, line 162: Minor: Typo `reciever` => `receiver`

crashtracker/src/crash_handler.rs
```diff
@@ -255,9 +295,30 @@ fn handle_posix_signal_impl(signum: i32) -> anyhow::Result<()> {
...
+///     If a crash occurs during execution of this function, it is possible that
+///     the crash handler will have been registered, but the old signal handler
+///     will not yet be stored.  This would lead to unexpected behaviour for the
+///     user.  This should only matter if something crashes concurrently with
+///     this function executing.
 pub fn register_crash_handlers(create_alt_stack: bool) -> anyhow::Result<()> {
-    anyhow::ensure!(OLD_HANDLERS.load(SeqCst).is_null());
+    if !OLD_HANDLERS.load(SeqCst).is_null() {
+        return Ok(());
+    }
```
ivoanjo: Minor: Not sure this is something we want to care about, but one thing we could do when the function is called is to check that our signal handlers are still active, e.g. because something else on the system may have overwritten them.

(I'm thinking of this, because there may be a situation where someone starts the crash tracker, and then discovers that the VM overwrites their signal handlers after X, and then calls `register_crash_handlers` again expecting libdatadog to be restored but isn't. Maybe too farfetched?)

ivoanjo (approved):
:+1: LGTM

crashtracker/src/crash_handler.rs
danielsn: done

profiling-ffi/src/crashtracker.rs
```diff
@@ -127,7 +167,7 @@ pub unsafe extern "C" fn ddog_prof_crashtracker_update_on_fork(
     match ddog_prof_crashtracker_update_on_fork_impl(config, metadata) {
         Ok(_) => ProfileResult::Ok(true),
         Err(err) => ProfileResult::Err(Error::from(
-            err.context("ddog_prof_crashtracker_init failed"),
+            err.context("ddog_prof_crashtracker_update_on_fork failed"),
         )),
```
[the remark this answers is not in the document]
danielsn: done

profiling-ffi/src/crashtracker.rs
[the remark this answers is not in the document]
danielsn: removed

crashtracker/src/crash_handler.rs
danielsn: https://github.com/DataDog/libdatadog/issues/325

PR #300, opened 2024-01-31
refactor: Slice and Vec

morrisonlevi:
# What does this PR do?

There are two main aspects to this:

 1. Implement `Deref<Target=[T]>` for `Slice` and `Vec`. This effectively allows callers to use all slice methods on both types. This culls some code while definitely being more expressive.
 2. Removes unsafe from most functions for these types. If you don't trust type coming from FFI, then check it at the boundary anyway. Once the types are correctly constructed, they aren't unsafe, and we don't need to infect everything.

# Motivation

I recently learned about the `Deref<Target=[T]>` trick and finally got around to applying it here.

# Additional Notes

I renamed `Slice::new` to `Slice::from_raw_parts` and added a `Slice::new` which is a `const fn`. I fixed all the code in the repository, but if you use it from outside, you may need to change.

# How to test the change?

Everything should work as before 👍🏻.

## For Reviewers

### DIFF
--- a/ddcommon-ffi/src/slice.rs
+++ b/ddcommon-ffi/src/slice.rs
@@ -1,46 +1,52 @@
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
 
+use core::slice;
 use std::borrow::Cow;
 use std::fmt::{Debug, Formatter};
 use std::marker::PhantomData;
 use std::os::raw::c_char;
 use std::str::Utf8Error;
 
-/// Remember, the data inside of each member is potentially coming from FFI,
-/// so every operation on it is unsafe!
 #[repr(C)]
 #[derive(Copy, Clone)]
 pub struct Slice<'a, T: 'a> {
     /// Must be non-null and suitably aligned for the underlying type.
     ptr: *const T,
 
-    /// The number of elements (not bytes) that `.ptr` points to.
+    /// The number of elements (not bytes) that `.ptr` points to. Must be less
+    /// than or equal to [isize::MAX].
     len: usize,
-    marker: PhantomData<&'a [T]>,
+    _marker: PhantomData<&'a [T]>,
+}
+
+impl<'a, T: 'a> core::ops::Deref for Slice<'a, T> {
+    type Target = [T];
+
+    fn deref(&self) -> &Self::Target {
+        self.as_slice()
+    }
 }
 
 impl<'a, T: Debug> Debug for Slice<'a, T> {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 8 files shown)


ddcommon-ffi/src/slice.rs
```diff
@@ -1,158 +1,119 @@
...
-/// Remember, the data inside of each member is potentially coming from FFI,
-/// so every operation on it is unsafe!
 #[repr(C)]
 #[derive(Copy, Clone)]
 pub struct Slice<'a, T: 'a> {
     /// Must be non-null and suitably aligned for the underlying type.
     ptr: *const T,
 
-    /// The number of elements (not bytes) that `.ptr` points to.
+    /// The number of elements (not bytes) that `.ptr` points to. Must be less
```
danielsn, line 17: 💯

ddcommon-ffi/src/slice.rs
```diff
@@ -1,158 +1,119 @@
...
+    /// than or equal to [isize::MAX].
     len: usize,
-    marker: PhantomData<&'a [T]>,
+    _marker: PhantomData<&'a [T]>,
+}
+
+impl<'a, T: 'a> core::ops::Deref for Slice<'a, T> {
```
danielsn, line 23: 💯

ddcommon-ffi/src/slice.rs
```diff
@@ -1,158 +1,119 @@
...
+        self.as_slice().fmt(f)
     }
 }
 
 impl<'a, T: Eq> PartialEq<Self> for Slice<'a, T> {
     fn eq(&self, other: &Self) -> bool {
-        unsafe {
-            return self.as_slice() == other.as_slice();
-        }
+        **self == **other
```
danielsn, line 39: 🤣

ddcommon-ffi/src/vec.rs
```diff
@@ -73,7 +74,7 @@ impl<'a, T> IntoIterator for &'a Vec<T> {
     type IntoIter = core::slice::Iter<'a, T>;
 
     fn into_iter(self) -> Self::IntoIter {
-        unsafe { self.as_slice().into_slice() }.iter()
+        self.deref().iter()
```
danielsn, line 77: Would `*self` work here instead?

danielsn (approved):
Couple questions, but LGTM

ddcommon-ffi/src/vec.rs
morrisonlevi: Yes, it would be `(*self).iter()` but that felt uncomfortable so I just used `.deref()` instead. However, the compiler sometimes auto-derefs so this could possibly just be `self.iter()`, but that feels a bit more magical and I'm not sure code style wise what would be preferred.

realFlowControl: I have seen clippy (in other situations) suggesting to remove `.deref()` or the `*` because it would auto-deref anyway, but I personally prefer the explicit deref over the implicit one.

PR #302, opened 2024-02-01
RFC proposal around building and delivering different features

r1viollet:
# What does this PR do?

It proposes an RFC

# Motivation

Making an evolution to the build system

# Additional Notes

N.A.

# How to test the change?

N.A.

## For Reviewers


docs/RFCs/delivering-different-features.md
```diff
@@ -0,0 +1,78 @@
...
+- crash tracking
+- symbolization (ongoing).
+
+Not every profiler / tracer needs every feature, so we need it here way of Building and delivering exactly what we need.
+
+## Not a solution
+
+### Delivering different libraries
+
+By delivering features in several libraries, we expose ourselves to duplicated Rust runtimes (with possibly different versions and behaviours).
```
ivoanjo, line 35: While it's clear to me what you mean by "delivering different libraries" here, I'm not sure it'll be clear for all readers that you mean different libraries per-feature (telemetry, profiling, ...) or even finer-grained VS different libraries among other axis (e.g. for each language).

docs/RFCs/delivering-different-features.md
```diff
@@ -0,0 +1,78 @@
...
+cargo build --features symbolizer telemetry crash-tracking
+```
+
+#### Delivery
+
+CI pipelines are created with the required features.
+
+## Recommended
+
+I think we can do both proposed solutions. For the languages that require shared libraries, we can have the feature solution.
```
ivoanjo, line 77: Arguably, the "build everything together" option doesn't really preclude the shared library approach. We could still produce the one static library, and leave turning that into a shared library into a downstream per-library concern.

For instance, considering Ruby:
* The main libdatadog release artifact could be a static library
* This static library would be the new input to the Ruby packaging scripts, which would build a shared library from it, prior to publishing it to rubygems.org
* Alternatively -- we could re-examine our assumptions regarding using a shared library for Ruby being the best solution, and see if we can make a static library work and what would be the trade-offs in doing so

ivoanjo (commented):
Thanks for opening this up! Left some feedback :)

docs/RFCs/delivering-different-features.md
r1viollet: I feel this is a slightly different solution that I could list.

r1viollet: Also, this forces ruby to bundle a rust pipeline. Does this not mean solving the feature issue downstream ?

r1viollet: I added this to the list of options

ivoanjo: > Also, this forces ruby to bundle a rust pipeline. Does this not mean solving the feature issue downstream ?

Not in the way I was thinking... Let me try to clarify:

Current state:
1. Libdatadog release builds a static and a dynamic library for profiling, which gets uploaded as a tarball to GitHub
2. The scripts in the `ruby/` folder in the libdatadog repository take the GitHub release, keeps only the dynamic library, adds a few Ruby helpers, packages and uploads it as the "libdatadog" gem to rubygems.org
3. Downstream dd-trace-rb consumes the "libdatadog" gem, using the shared library inside

I'm thinking this may be an option:
1. Libdatadog release builds only a static library with everything, which gets uploaded as a tarball to GitHub
2. We update the scripts in the `ruby/` folder to take as an input the static library and to repackage it as a dynamic library containing only the APIs that Ruby makes use of. That dynamic library gets uploaded as the "libdatadog" gem to rubygems.org
3. Same as above -- probably no changes needed

ivoanjo: To clarify, I'm not necessarily arguing for this approach, mostly stating that it may be an option if we want to focus on producing a single static library and then try to base everything on it, rather than having multiple paths to building from source.

r1viollet: That is exactly what I intended to describe with my 3rd solution. Mission failed ;-) 
I'll add your description to illustrate

pawelchcki:
Thanks @r1viollet - looks perfect from my perspective (and to my existing knowledge) 


> I would recommend a mix of solution 1 and 2. For the languages that require shared libraries, we can use the feature solution (solution 2).
We can deliver a single static library (with all features) for languages that can link against a static library.

LGTM ! - worth double checking with @Anilm3 to ensure it aligns with ASM experience

pawelchcki (approved):

Anilm3:
> Thanks @r1viollet - looks perfect from my perspective (and to my existing knowledge)
> 
> > I would recommend a mix of solution 1 and 2. For the languages that require shared libraries, we can use the feature solution (solution 2).
> > We can deliver a single static library (with all features) for languages that can link against a static library.
> 
> LGTM ! - worth double checking with @Anilm3 to ensure it aligns with ASM experience

We don't have the issue of multiple distinct feature-sets within the single native library so perhaps my feedback isn't particularly helpful.

In terms of the options suggested:

1. Seems perfectly reasonable, I think the static library size is not particularly important as long as it's within reason. The only concern is the size of the final deployed binary and primarily only a concern for serverless, although it can still be an annoyance in other environments due to the need to deploy a binary per supported architecture and libc.
2. I'm not sure if this will scale well as you add more distinct features into libdatadog, but it's a start and if the goal is for everyone to use a single implementation I suppose languages will start to converge towards the complete libdatadog binary. 

On the other hand, this is an age-old problem which is traditionally solved by having multiple interdependent dynamic libraries. The most simple scenario is the one in which you have a core dynamic library and then multiple others dependent on it. No doubt there's still some redundancy and this typically doesn't involve statically linked dynamic libraries only depending on each other and libc so I'm not sure if it's feasible.

r1viollet:
@Anilm3 thank you for the feedback. To sum up, you are OK, though feel like solution 2 could get out of hand. 
To answer this, I think only Ruby will continue linking dynamically (and we are talking about removing this requirement)

I can list the solution around having many different dynamic libraries as an alternative.

Though I am not in favour of this. This will increase the likelihood of duplication and adds complexity. The aim will be to integrate modules together (example: symbolization can be done within the same step as pprof aggregation), and reduce the need for FFIs. Happy to reconsider if someone can make a strong case for this.

Anilm3:
> @Anilm3 thank you for the feedback. To sum up, you are OK, though feel like solution 2 could get out of hand. To answer this, I think only Ruby will continue linking dynamically (and we are talking about removing this requirement)

That's interesting as with libddwaf, libraries have generally headed in the other direction (using the dynamic library more than the static one).

> I can list the solution around having many different dynamic libraries as an alternative.
> 
> Though I am not in favour of this. This will increase the likelihood of duplication and adds complexity. The aim will be to integrate modules together (example: symbolization can be done within the same step as pprof aggregation), and reduce the need for FFIs. Happy to reconsider if someone can make a strong case for this.

I have no strong opinion about including this as an alternative option and I completely agree with the complexity argument. In most cases the approach I mentioned is typically used for general purpose libraries, while what we aim to do here is a little bit different.

r1viollet:
@Anilm3 Thanks, I added the extra option. The fact it does not apply here does not mean we should not document it as a possibility

Anilm3 (approved):

bantonsson (approved):
I agree that we need to be mindful to avoid a features explosion, but option 1 and 2 does make the most sense.

r1viollet:
> I agree that we need to be mindful to avoid a features explosion, but option 1 and 2 does make the most sense.
Yes, that is a risk. If we turn around and we have 10 different CI pipelines, we will know that we failed somewhere along the way. 

Thanks for the feedback!

PR #303, opened 2024-02-05
perf: less mem per entry in the string table

morrisonlevi:
# What does this PR do?

This uses a `Box<str>` instead of `String`, saving 1 `usize` per entry. When it's serialized to pprof proto, it gets encoded directly without allocating a temporary `String`. 

# Motivation

I'm working on reducing memory and CPU overhead of profiles and noticed this pretty cheap reduction in memory.

# Additional Notes

I've opened a PR upstream to put a generic str encoder helper there, as I think others could benefit from it as well. This part enables future optimizations like using a memory arena for string table data.

[PROF-9125](https://datadoghq.atlassian.net/browse/PROF-9125)

# How to test the change?

Test the same as usual, public API should be the same.

## For Reviewers

[PROF-9125]: https://datadoghq.atlassian.net/browse/PROF-9125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ivoanjo:
Nice! :D

profiling/src/pprof/sliced_proto.rs
```diff
@@ -85,12 +85,6 @@ pub struct ProfileFunctionsEntry {
     pub functions_entry: Function,
 }
 
-#[derive(Eq, Hash, PartialEq, ::prost::Message)]
-pub struct ProfileStringTableEntry {
-    #[prost(string, required, tag = "6")]
-    pub string_table_entry: String,
-}
-
```
danielsn, line 93: Why the change to do this differently from all the other field serializers?

profiling/src/serializer/compressed_streaming_encoder.rs
```diff
@@ -17,6 +27,26 @@ impl CompressedProtobufSerializer {
         Ok(())
     }
 
+    /// Only meant for string table strings. This is essentially an
+    /// implementation of [prost::Message::encode] but for any `AsRef<str>`,
+    /// and specialized for handling the unlikely OOM conditions of writing
+    /// into a `Vec<u8>`.
+    pub(crate) fn encode_str(&mut self, item: impl AsRef<str>) -> anyhow::Result<()> {
+        let tag = 6u32;
```
danielsn, line 35: Instead of hardcoding the tag, could take it as a function argument and be general.

profiling/src/pprof/sliced_proto.rs
morrisonlevi: Because `prost::Message` requires a `String` to encode strings. You can't encode any other `String`-like representations such as `&str`, `Box<str>`, etc with a `prost::Message`.

profiling/src/serializer/compressed_streaming_encoder.rs
morrisonlevi: The tag still has to be hard-coded somewhere. Previously it was here:

```rs
pub struct ProfileStringTableEntry {
    #[prost(string, required, tag = "6")]
    pub string_table_entry: String,
}
```

Would you rather I rename the method to something like `encode_string_table_entry` instead?

morrisonlevi: I have renamed this method to `encode_string_table_entry`.

bantonsson (approved):

PR #306, opened 2024-02-07
feat(profiling)!: add arena-based string table

morrisonlevi:
# What does this PR do?

Creates a string table that stores the bytes of the strings into an arena, instead of using individual allocations. The arena has a maximum capacity, and uses virtual memory to acquire pages lazily.

This is a breaking change:

 1. Many routines in Rust which weren't fallible now are to account for the string table hitting its capacity, or if initial allocations fail.
 2. The minimum string arena capacity in bytes must be passed to `ddog_prof_Profile_new`.

# Motivation

I'm working on reducing overhead of the PHP profiler, and reducing memory of the string table is a nice step. It also puts less pressure onto the system allocator, which should help others like the Ruby profiler as well.

# Additional Notes

The PR is quite large. Maybe I should split it into two pieces, but it's kind of hard to know where to split it.

[PROF-9125](https://datadoghq.atlassian.net/browse/PROF-9125)

# How to test the change?

The function `ddog_prof_Profile_new` has a new parameter for the minimum size of the arena, which will round up to the nearest page size for you. The memory is virtually allocated, so it will be paged in by the OS as needed. You don't need to worry _too_ much about its size, as long as you are being reasonable (e.g. 2 GiB may be fine, 2 TiB is not).

## For Reviewers

[PROF-9125]: https://datadoghq.atlassian.net/browse/PROF-9125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


morrisonlevi:
In PHP, I got a 3-4% memory savings on the container I used to test. I'm happy with this number for PHP and look forward to hearing numbers from any other runtime which can test it.

profiling-ffi/src/profiles.rs
```diff
@@ -333,22 +332,24 @@ impl<'a> TryFrom<Sample<'a>> for api::Sample<'a> {
...
-/// All slices must be have pointers that are suitably aligned for their type
+/// All slices must have pointers that are suitably aligned for their type
 /// and must have the correct number of elements for the slice.
 #[no_mangle]
 #[must_use]
 pub unsafe extern "C" fn ddog_prof_Profile_new(
     sample_types: Slice<ValueType>,
     period: Option<&Period>,
     start_time: Option<&Timespec>,
+    string_arena_min_capacity: libc::size_t,
```
gleocadie, line 343: Do we really need to expose this to the client libraries ? I feel like that should be internal.
To be honest, when I create a profile object I do not know the minimal.
I feel like it's worth hiding it (since it's a technical detail)

profiling-ffi/src/profiles.rs
```diff
@@ -333,22 +332,24 @@ impl<'a> TryFrom<Sample<'a>> for api::Sample<'a> {
...
     let types: Vec<api::ValueType> = sample_types.into_slice().iter().map(Into::into).collect();
     let start_time = start_time.map_or_else(SystemTime::now, SystemTime::from);
     let period = period.map(Into::into);
 
-    let internal_profile = internal::Profile::new(start_time, &types, period);
-    let ffi_profile = Profile::new(internal_profile);
-    ProfileNewResult::Ok(ffi_profile)
+    match internal::Profile::new(start_time, &types, period, string_arena_min_capacity) {
+        Ok(internal_profile) => ProfileNewResult::Ok(Profile::new(internal_profile)),
+        Err(err) => ProfileNewResult::Err(Error::from(err.context("ddog_prof_Profile_new failed"))),
```
gleocadie, line 351: maybe add a test that actually fails no?

profiling-ffi/src/profiles.rs
```diff
@@ -468,8 +469,13 @@ pub unsafe extern "C" fn ddog_prof_Profile_set_endpoint(
         }
     };
     let endpoint = endpoint.to_utf8_lossy();
-    profile.add_endpoint(local_root_span_id, endpoint);
-    ProfileResult::Ok(true)
+    if let Err(err) = profile.add_endpoint(local_root_span_id, endpoint) {
```
gleocadie, line 472: same comment about testing

profiling-ffi/src/profiles.rs
morrisonlevi: It's exposed because every runtime has different ideas about how much memory usage is okay. You making this comment made me realize that "min_capacity" is not the best name: it rounds up to a page size, and if you exceed that it will fail. So this number should be close to the maximum you want. It's lazily allocated, so you don't need to worry too much about the memory unless you actually use it, as long as your request is sane (you know, like 1 GiB, not 1 TiB).

gleocadie: > It's exposed because every runtime has different ideas about how much memory usage is okay.
To be honest, I'm not sure about that: I do not know that information for .NET...

Why not making :
- Make it optional with a default value set to 4096
- If the value is provided, aligned it to a page size
- And prevent for programming error by adding a max value
?

profiling/src/internal/profile.rs
```diff
@@ -451,21 +474,12 @@ impl Profile {
...
-                // we try to insert it, but suddenly it exists now?
-                debug_assert!(_inserted);
-                index
-            }
-        };
-        StringId::from_offset(index)
+    #[inline]
+    fn intern<S>(&mut self, item: &S) -> anyhow::Result<StringId>
+    where
+        S: ?Sized + Borrow<str>,
```
danielsn, line 480: Do we need this?

profiling/src/internal/profile.rs
```diff
@@ -840,21 +854,23 @@ mod api_test {
...
          */
-        let mut profile = provide_distinct_locations();
+        let mut profile = provide_distinct_locations()?;
 
         let period = (
             10_000_000,
             ValueType {
-                r#type: profile.intern("wall-time"),
-                unit: profile.intern("nanoseconds"),
+                r#type: profile.strings.insert_full("wall-time")?,
```
danielsn, line 872: This is because of the weird type we have for valuetypes

danielsn:
What happens if HUGE_PAGES are enabled?

profiling-replayer/src/main.rs
```diff
@@ -151,7 +151,8 @@ fn main() -> anyhow::Result<()> {
         replayer.start_time,
         &replayer.sample_types,
         replayer.period,
-    );
+        u32::MAX as usize,
```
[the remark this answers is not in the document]
morrisonlevi: I had to provide _some_ value. A max of 4 GiB seemed okay for request replayer.

profiling/src/alloc/arena.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /// Allocates the given layout. It will be zero-initialized. Allows for
+    /// zero-sized allocations, which are not guaranteed to have unique
+    /// addresses.
+    pub fn allocate_zeroed(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
+        let layout = layout.pad_to_align();
+        let size = layout.size();
+        if size == 0 {
+            return Ok(NonNull::from(&mut []));
```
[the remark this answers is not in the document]
morrisonlevi: Because it's a wide pointer `NonNull<[u8]>`, so we also need size info.

```
error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
  --> profiling/src/alloc/arena.rs:70:23
   |
70 |             return Ok(NonNull::dangling());
   |                       ^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
   |
   = help: the trait `Sized` is not implemented for `[u8]`
note: required by a bound in `NonNull::<T>::dangling`
  --> /Users/levi.morrison/.rustup/toolchains/1.69.0-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/non_null.rs:67:6
   |
67 | impl<T: Sized> NonNull<T> {
   |      ^ required by this bound in `NonNull::<T>::dangling`
```

profiling/src/alloc/arena.rs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        let mapping = match self.mapping.as_ref() {
```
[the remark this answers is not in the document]
morrisonlevi: Close, `.ok_or_else(|| AllocError)?`

profiling/src/alloc/arena.rs
```diff
@@ -0,0 +1,215 @@
...
+    use super::*;
+    use crate::alloc::page_size;
+    use std::mem;
+
+    #[test]
+    fn test_capacity_0() -> anyhow::Result<()> {
+        let arena = ArenaAllocator::with_capacity(0)?;
+
+        // This should fail to allocate, arena no capacity cannot allocate.
+        arena.allocate_zeroed(Layout::new::<u8>()).unwrap_err();
```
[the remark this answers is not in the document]
morrisonlevi: I asked a few folks what they'd want at the API boundary--let's see what they say.

profiling/src/alloc/arena.rs
```diff
@@ -0,0 +1,215 @@
...
+    #[test]
+    fn test_arena_basic_exhaustion() -> anyhow::Result<()> {
+        let arena = ArenaAllocator::with_capacity(1)?;
+
+        let expected_size = page_size();
+        let actual_size = arena.remaining_capacity();
+        assert_eq!(expected_size, actual_size);
+
+        // This should consume the whole arena.
+        arena.allocate_zeroed(Layout::from_size_align(expected_size, 1)?)?;
```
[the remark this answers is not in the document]
morrisonlevi: Hmm. I wonder if miri considers these bytes initialized... if not, this is UB to even check.

profiling/src/alloc/arena.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Creates a new arena allocator which has a capacity of zero. It will
+    /// not request a virtual mapping from the OS.
+    pub const fn new() -> Self {
+        Self {
+            mapping: None,
+            first_free_offset: Cell::new(0),
+        }
+    }
+
+    unsafe fn from_mapping(mapping: Mapping<A>) -> ArenaAllocator<A> {
```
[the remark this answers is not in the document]
morrisonlevi: Fixed by inlining the only remaining use ^_^

profiling-ffi/src/profiles.rs
```diff
@@ -884,6 +893,7 @@ mod test {
             Slice::from_raw_parts(sample_type, 1),
             None,
             None,
+            4096,
```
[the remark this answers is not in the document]
morrisonlevi: Not specifically in FFI, no. There are tests on arena and string table for small values like `64`, and there is a test on string table which does `4 * u16::MAX` or in other words, a number just under 256 KiB.

But not currently testing other values in FFI.

profiling/src/internal/profile.rs
```diff
@@ -133,7 +143,8 @@ impl Profile {
         start_time: SystemTime,
         sample_types: &[api::ValueType],
         period: Option<api::Period>,
-    ) -> Self {
+        string_arena_capacity_hint: usize,
```
ivoanjo, line 146: Here and in the ffi, it may be worth specifying the unit here is in bytes maybe?

morrisonlevi: It gets rounded to a page size. There's mostly no consequence for setting this high, because it's virtual memory so you pay for what you actually use. There are some edges though. like on Linux, OOM kill will be more attracted to your process if you have high virtual memory use. And .NET supports 32-bit processes which only has 2 GiB of virtual memory available per process.

I met with Christophe and Gregory and they are concerned about hitting a max, any max, because that will screw up the accuracy of the upscaling, and they would consider this inaccuracy intolerable in .NET. I knew that the data would incomplete, and that may have been okay. But it will also be inaccurate, and that's not.

Making a more complicated and advanced allocator is a possibility, but then the risk surface goes up too. I'm currently exploring an arena which has fixed-sized chunks, and it allocates new chunks as needed. It uses the system allocator to hold these in a `Vec`, which isn't ideal, but that's far less memory than is represented by the chunks, so it would still greatly reduce the effects of commingling with the system allocator. I could avoid this by using an intrusive linked-list, but this slows performance and makes it harder to provide a global offset for it. Anyway, this is the direction I'm looking at right now.

morrisonlevi:
Closing in favor of #404.

PR #313, opened 2024-02-13
Build refactoring - cbindgen

r1viollet:
# What does this PR do?

Ensure cbindgen is included in the package's build steps. By removing logics from the shell scripts, we can make build steps more modular.
This is step 1 of moving towards a common "feature" driven build system.

# Motivation

You can refer to the [RFC](https://github.com/DataDog/libdatadog/blob/main/docs/RFCs/delivering-different-features.md) around features

# Additional Notes

# How to test the change?

## For Reviewers


build-telemetry-ffi.sh
```diff
@@ -68,7 +68,7 @@ export RUSTFLAGS="${RUSTFLAGS:- -C relocation-model=pic}"
 
 datadog_telemetry_ffi="ddtelemetry-ffi"
 echo "Building the ${datadog_telemetry_ffi} crate (may take some time)..."
-cargo build --package="${datadog_telemetry_ffi}" --release --target "${target}"
+DESTDIR="$destdir" cargo build --package="${datadog_telemetry_ffi}" --release --target "${target}"
```
r1viollet, line 71: I am trying to debate if this is the right way of doing things.
For now we some `cp` underneath and this different mechanism.

I don't really see the benefit from the `cp` operations bellow.

r1viollet: The `OUT_DIR` set by cargo is sadly not the right place.
I think this is still slightly better, but not perfect.

ddcommon-ffi/build.rs
```diff
@@ -0,0 +1,16 @@
...
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    let crate_dir = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").unwrap());
+    let header_name = "common.h";
+    let output_base_dir = env::var("DESTDIR").ok(); // Use `ok()` to convert Result to Option
+
+    generate_header(crate_dir, header_name, output_base_dir.as_deref());
+    println!("cargo:rerun-if-env-changed=DESTDIR");
```
danielsn, line 15: We have the exact same code for all three crates, with the only difference being the header name.  Maybe worth factoring into a function?

build-common/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+    let cargo_target_dir = output_base_dir.unwrap_or("target");
+
+    // Determine if `cargo_target_dir` is absolute or relative
+    let cargo_target_path = Path::new(cargo_target_dir);
+    let output_path = if cargo_target_path.is_absolute() {
+        // If absolute, use it directly
+        cargo_target_path.join("include/datadog/").join(header_name)
+    } else {
+        // If relative, adjust the path accordingly. we are in a crate, so get back to top level
+        let adjusted_path = Path::new("..").join(cargo_target_path);
```
danielsn, line 25: is this guaranteed to work? What if the command is run deeper in the crate?

ddcommon-ffi/build.rs
r1viollet: I have mixed feelings about hiding some of this behaviour :-)
I added the magical function all the same.

build-common/src/lib.rs
r1viollet: No.
However I can not think of a way to magically infer where the headers should be placed.
I looked into using the OUT_DIR variable, but that also has different depths depending on build parameters.
So I think this being not perfect still guarantees that it works in the use cases we have today.
Happy to reconsider if you find a different logic that works well.

danielsn: If we want the root of the workspace, can we try https://doc.rust-lang.org/cargo/commands/cargo-locate-project.html#options

morrisonlevi (approved):

build-common/src/lib.rs
r1viollet: Here is a new iteration using your proposal.

build-common/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2021-Present Datadog, Inc.
+use cbindgen::Config;
+use std::path::PathBuf;
+use std::process::Command;
+use std::str;
+use std::{env, fs};
+
+/// Configure the header generation using environment variables.
+/// call into `generate_header` with the appropriate arguments.
+///
```
r1viollet, line 11: I can still find ways to break this, but I feel we already improve the state of the build.

bantonsson (approved):

danielsn (approved):

PR #326, opened 2024-02-21
Allow the sidecar to spawn, even if the log file could not be opened

bwoebi:
But still report failure, so that it's logged; and subsequent connections will be able to successfully connect.

This allows for a bit more graceful handling than having every sidecar attempt completely fail.


bantonsson (approved):
I think this is a reasonable approach.

sidecar/src/entry.rs
```diff
@@ -163,7 +173,7 @@ pub fn daemonize(listener: IpcServer, cfg: Config) -> anyhow::Result<()> {
         .map_err(|err| io::Error::new(io::ErrorKind::Other, err))
         .context("Could not spawn the sidecar daemon")?;
 
-    Ok(())
+    log_error.context("Failed to open the log file")
```
paullegranddc, line 176: But if you return an error here, you'll fail to connect to the sidecar in the function that called, even though subsequent calls will succeed right?

I think that's a bad idea to return the error.
* Either, it's a hard error to not be able to log to file and we shouldn't spawn the sidecar
* We admit it as acceptable an the initial call should connect

paullegranddc (approved):

sidecar/src/entry.rs
bantonsson: Yes, I think you are right. It's better to push the _logging_ of what happened down here (or maybe treat certain errors as recoverable) than right out fail the first attempt.

bantonsson (changes requested):

bantonsson (approved):

PR #327, opened 2024-02-21
[crashtracker] Use blazesym to resolve stack trace names in the receiver 

danielsn:
# What does this PR do?

Uses blazesym to resolve symbol addresses into names.  

# Motivation

Crash reports with names are better than crashreports with pointers.

# Additional Notes

This depends on an unreleased change in Blazesym.  We pin to a commit that has the change.  Once they release it, we can move to use that version. 

# How to test the change?

run `crashtracker::api::test_crash` and see that the the emitted crash report now includes 

after:
```
  "stacktrace": [
    {
      "ip": "0xaaaab1b32350",
      "module_base_address": null,
      "names": [
        {
          "colno": 5,
          "filename": "mod.rs",
          "lineno": 66,
          "name": "backtrace::backtrace::trace_unsynchronized"
        }
      ],
      "sp": "0xffffa5f758c0",
      "symbol_address": "0xaaaab1b3231c"
    },
    {
      "ip": "0xaaaab1b32438",
      "module_base_address": null,
      "names": [
        {
          "colno": 5,
          "filename": "collectors.rs",
          "lineno": 30,
          "name": "datadog_crashtracker::collectors::emit_backtrace_by_frames"
        }
      ],
      "sp": "0xffffa5f75900",
      "symbol_address": "0xaaaab1b32380"
    },
```

Before:
```
  "stacktrace": [
    {
      "ip": "0xaaaadd44234c",
      "module_base_address": null,
      "names": null,
      "sp": "0xffffa4b508c0",
      "symbol_address": "0xaaaadd442318"
    },
    {
      "ip": "0xaaaadd442434",
      "module_base_address": null,
      "names": null,
      "sp": "0xffffa4b50900",
      "symbol_address": "0xaaaadd44237c"
    },
```

## For Reviewers

### DIFF
--- a/crashtracker/Cargo.toml
+++ b/crashtracker/Cargo.toml
@@ -13,7 +13,8 @@ crate-type = ["lib"]
 [dependencies]
 anyhow = "1.0"
 backtrace = "0.3.69"
-blazesym = {git = "https://github.com/libbpf/blazesym.git"}
+# TODO: move to a named release once https://github.com/libbpf/blazesym/pull/520 makes it to release
+blazesym = {git = "https://github.com/libbpf/blazesym.git", rev = "4fa15cb"}
 chrono = {version = "0.4", default-features = false, features = ["std", "clock", "serde"]}
 datadog-profiling = { path = "../profiling"}
 ddcommon = {path = "../ddcommon"}
--- a/crashtracker/Cargo.toml
+++ b/crashtracker/Cargo.toml
@@ -13,6 +13,7 @@ crate-type = ["lib"]
 [dependencies]
 anyhow = "1.0"
 backtrace = "0.3.69"
+blazesym = {git = "https://github.com/libbpf/blazesym.git"}
 chrono = {version = "0.4", default-features = false, features = ["std", "clock", "serde"]}
 datadog-profiling = { path = "../profiling"}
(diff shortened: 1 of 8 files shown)


crashtracker/Cargo.toml
```diff
@@ -10,6 +10,10 @@ license.workspace = true
 [lib]
 crate-type = ["lib"]
 
+[target.'cfg(unix)'.dependencies]
+# TODO: move to a named release once https://github.com/libbpf/blazesym/pull/520 makes it to release
+blazesym = {git = "https://github.com/libbpf/blazesym.git", rev = "4fa15cb"}
```
r1viollet, line 15: I need to make a mental note to take the same revision

r1viollet (approved):
LGTM
I was also thinking about the implications.
This means that Ruby is pulling blazesym through the crashtracking functionality.
So we can merge my work on having the symbolizer available.

r1viollet:
This is great! LGTM

danielocfb:
FWIW, we just released `blazesym` `0.2.0-alpha.11`, which contains the changes you need, so you can use that. (for the C API you'll still need to pull from Git directly, but you could at least import a tag, if that makes it any better)

PR #330, opened 2024-02-23
Updated rustfmt rules

bantonsson:
# What does this PR do?

Updates the `rustfmt` rules to format comments and code in doc comments.

# Motivation

The more ppl/projects working in the repository, the more important it becomes to enforce common formatting for comments and doc comments as well, to avoid very long comment lines and other differences in comment style.

# Additional Notes

The biggest bulk of the changes are reformatting of the copyright headers (which rustfmt doesn't do in a nice way automatically, but will maintain).

The PR is split into 3 different commits for easier review, with the copyright header reformatting in a separate commit.

There is a helper script in `./scripts/reformat_copyright.sh` that can be used to update the copyright headers in new files in PRs.
```
fd --type f '\.(rs|toml|cpp|c|sh|in)$' | while read fname; do; echo $fname; ./scripts/reformat_copyright.sh $fname; done
```

# How to test the change?

## For Reviewers

### DIFF
--- a/scripts/reformat_copyright.sh
+++ b/scripts/reformat_copyright.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+# Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+# SPDX-License-Identifier: Apache-2.0
+
+set -eu
+
+fname=$1
+ftmp="${fname}.tmp"
+# preserve file attributes
+cp -p "${fname}" "${ftmp}"
+# process the file
+cat "${fname}" | sed 's/^\([^a-zA-Z]*\).*Datadog.*Copyright \([0-9]*\)-Present.*$/\1Copyright \2-Present Datadog, Inc\. https\:\/\/www\.datadoghq\.com\/\n\1SPDX-License-Identifier: Apache-2.0/' | sed '/^\([^a-zA-Z]*\)Unless explicitly stated/d' | sed '/^\([^a-zA-Z]*\)under the Apache/d' | sed '/^\([^a-zA-Z]*\)Datadog, Inc.$/d' >| "${ftmp}" && mv "${ftmp}" "${fname}"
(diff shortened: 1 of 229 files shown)


bwoebi:
I find it hard to read through many new lines with comments.
I encourage writing shorter sentences than breaking sentences randomly through enforced code formatting. Please don't do that.
It also makes for uglier diffs when changing something small in a comment.

bantonsson:
If ppl would just write shorter comments, then there would be no problems, but we don't. This is also an attempt at enforcing formatting of doc comments and code in doc comments as well, and the comment formatting comes with the package.

I personally don't think that the status quo is acceptable, so what is your proposal @bwoebi ?

bwoebi:
I personally think the status quo is absolutely fine and I never was annoyed by any comment formatting in this codebase ever.

scripts/reformat_copyright.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
paullegranddc, line 1: When should this be called?
It looks like it's not used in the CI

paullegranddc (approved):

scripts/reformat_copyright.sh
bantonsson: It's only a helper script for end users (that is us) to use manually when fixing up a PR.

> There is a helper script in `./scripts/reformat_copyright.sh` that can be used to update the copyright headers in new files in PRs.

danielsn:
Do we actually need the copyright headers?  Can we move to just SPDX headers? https://spdx.dev/learn/handling-license-info/

ekump (approved):

danielsn (approved):
I agree with @bwoebi that the changes make the existing comments look worse.  But, in the long run having a formatter seems like a worthwhile improvement.

bantonsson:
I switched the license headers to a Copyright line and SPDX.

PR #331, opened 2024-02-26
[ddtelemetry] Implement ffi code for telemetry metrics

paullegranddc:
# What does this PR do?

* Add an entrypoint to the telemetry worker that doesn't send app-events, just metrics and logs for use in languages that already have a telemetry client sending app-events, but want to send telemetry metrics from the native side

* Expose telemetry metrics API to the ffi

* Refactor the telemetry API to use `NonNull<Box<Type>>` instead of `&mut *mut Type` to represent out parameters.
This doesn't have any impact on the generated C headers, or the ffi API since this types have the exact same ABI, but it expresses intent better IMO.

# Motivation

This is needed because Christophe Nasarre wants to use libdatadog to send telemetry metrics from the dotnet profiler, from the native library side.

# Additional Notes

# How to test the change?

## For Reviewers


gleocadie:
Is it possible to have an example in C/C++ in `https://github.com/DataDog/libdatadog/tree/main/examples/ffi` to show how to use the API please ?
It could be also used to check that the functions are correctly exported and can be used.

ddtelemetry-ffi/src/worker_handle.rs
```diff
@@ -82,7 +88,39 @@ pub extern "C" fn ddog_handle_stop(handle: &TelemetryWorkerHandle) -> MaybeError
...
 }
 
+#[allow(clippy::missing_safety_doc)]
+#[no_mangle]
+pub unsafe extern "C" fn ddog_handle_register_metric_context(
+    handle: &TelemetryWorkerHandle,
+    name: ffi::CharSlice,
+    metric_type: MetricType,
+    tags: ffi::Vec<Tag>,
+    common: bool,
```
gleocadie, line 98: What is `common` ? what is it meaning ?
Could you add a docstring please ?

paullegranddc: Added 👍

gleocadie (approved):
LGTM

PR #332, opened 2024-02-26
Also move #defines when deduplicating headers

bwoebi:
Small change in dedup_headers, so that a pub const (translated as a #define) used in the definition of an enum will be in the same file than the enum.

```
pub const LOG_ONCE: isize = 1 << 3;

pub enum Log {
    Deprecated = 3 | LOG_ONCE,
}
```


bantonsson (approved):
Is this for future changes, or are there any such cases currently?

bwoebi:
@bantonsson It's for my work-in-progress branch in dd-trace-php.

PR #336, opened 2024-03-01
Use bincode for much better performance

bwoebi:
This also requires dropping skip_serializing_if statements which are incompatible with non-self-describing formats like msgpack or bincode. (See also https://github.com/serde-rs/serde/issues/1732)

Changing the format gave a 3.5 times speedup according to benchmarking profiles in serialization for dd-trace-php.


paullegranddc (approved):

danielsn:
Curious what the bandwidth effect of this change is

bwoebi:
About halved for telemetry, just because of the avoided names.

PR #340, opened 2024-03-04
Add composer installed.json parsing in sidecar

bwoebi:
Up to now this happened on the tracer side and could be quite expensive. We're moving it to the sidecar so that we can a) apply caching and b) not execute it in the request path of the tracer.

This brings considerable performance improvements in our benchmarks.

### DIFF
--- a/sidecar/src/interface.rs
+++ b/sidecar/src/interface.rs
@@ -461,6 +461,7 @@ impl EnqueuedTelemetryData {
             lazy_static! {
                 static ref COMPOSER_CACHE: tokio::sync::Mutex<ComposerCache> =
                     tokio::sync::Mutex::new(Default::default());
+                static ref LAST_CACHE_CLEAN: AtomicU64 = AtomicU64::new(0);
             }
 
             let mut cache = COMPOSER_CACHE.lock().await;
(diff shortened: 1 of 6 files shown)


sidecar/src/interface.rs
```diff
@@ -404,6 +436,65 @@ impl EnqueuedTelemetryData {
...
+                        let parsed: ComposerPackages = simd_json::from_slice(json.as_mut_slice())?;
+                        Ok(parsed.packages)
+                    }
+                    let packages = Arc::new(parse(&path).await.unwrap_or_else(|e| {
+                        warn!("Failed to report dependencies from {path:?}: {e:?}");
+                        vec![]
+                    }));
+                    cache.insert(path, (now, packages.clone()));
+                    // cheap way to avoid unbounded caching
+                    cache.retain(|_, (inserted, _)| *inserted > now.sub(Duration::from_secs(2000)));
```
bantonsson, line 490: So retain is `O(capacity)` since it visits all buckets (even empty ones), but I guess that there aren't that many entries (i.e. capacity is small) and/or not that many updates to the files.

bantonsson (commented):
This seems ok, but I can't find any tests at all of either the `json` parsing or the caching.

bantonsson (approved):

PR #346, opened 2024-03-06
Ignore some PHP specific code on 32 bit, fixing CI

bwoebi:
Fixes i686 windows CI

### DIFF
--- a/sidecar/src/interface.rs
+++ b/sidecar/src/interface.rs
@@ -480,7 +480,10 @@ impl EnqueuedTelemetryData {
                     }
                     async fn parse(path: &PathBuf) -> Result<Vec<data::Dependency>> {
                         let mut json = tokio::fs::read(&path).await?;
+                        #[cfg(not(target_arch = "x86"))]
                         let parsed: ComposerPackages = simd_json::from_slice(json.as_mut_slice())?;
+                        #[cfg(target_arch = "x86")]
+                        let parsed = ComposerPackages { packages: vec![] }; // not interested in 32 bit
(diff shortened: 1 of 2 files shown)


sidecar/src/interface.rs
```diff
@@ -480,7 +480,10 @@ impl EnqueuedTelemetryData {
                     }
                     async fn parse(path: &PathBuf) -> Result<Vec<data::Dependency>> {
                         let mut json = tokio::fs::read(&path).await?;
+                        #[cfg(not(target_arch = "x86"))]
                         let parsed: ComposerPackages = simd_json::from_slice(json.as_mut_slice())?;
+                        #[cfg(target_arch = "x86")]
```
bantonsson, line 485: Which project is interested in x86?

bantonsson (approved):

sidecar/src/interface.rs
bwoebi: I don't know, the sidecar isn't. Just making it pass here.

r1viollet:
Should we not prevent the full sidecar build on non x86 ?

bwoebi:
@r1viollet I don't care - the sidecar generally should still work, this a PHP specific routine where the x86 problem occurs.

r1viollet:
> @r1viollet I don't care - the sidecar generally should still work, this a PHP specific routine where the x86 problem occurs. And if yes, please provide a PR, not sure how to do that.

I see your point that this is not the focus. I am worried that we are creating new branches that should be tested in a different way.
Fixing CI is more important right now.

PR #305, opened 2024-02-06
Windows support for the sidecar

bwoebi:
Adds NamedPipes and spawning of the sidecar process for windows support.

PR #347, opened 2024-03-11
Introduce `DDOG_CHARSLICE_C_BARE` and undo change to `DDOG_CHARSLICE_C` helper

ivoanjo:
# What does this PR do?

This PR introduces a new `DDOG_CHARSLICE_C_BARE` helper, as well as undoes the change to `DDOG_CHARSLICE_C` from
https://github.com/DataDog/libdatadog/pull/305/files#diff-5c00b9c49c81bf0d0a754e0bcc6e0add4fe849bf0ea4a4d1ebfac8f698150f0b

There are now two helpers: `DDOG_CHARSLICE_C` that includes the `(ddog_CharSlice)` cast and `DDOG_CHARSLICE_C_BARE` which doesn't.

# Motivation

In a nutshell @bwoebi explained to me that on Windows there's times where you need to use `(ddog_CharSlice) {.ptr = ..., .len = ...}` and others where you need to use `{.ptr = ..., .len = ...}` and cannot include the cast.

Thus in #305 we tweaked the helper to never have `(ddog_CharSlice)`, with the intention of using `(ddog_CharSlice) DDOG_CHARSLICE_C(...)` some times, and `DDOG_CHARSLICE_C(...)` the other times.

But... this had the downside of making some of our Ruby Profiler code really ugly, as we'd need to move from `DDOG_CHARSLICE_C` to `(CharSlice) DDOG_CHARSLICE_C` in a lot of cases.

Thus, this PR introduces one helper for each variant, so that we always have a nice user-friendly option for each case, and avoid needing to sprinkle casts.

# Additional Notes

Arghhh C/C++ you always manage to be weird and annoying.

# How to test the change?

I've tested this change by compiling the Ruby profiler to use this branch of libdatadog, and validating that it compiles successfully on Linux.

## For Reviewers

### DIFF
--- a/ddcommon-ffi/cbindgen.toml
+++ b/ddcommon-ffi/cbindgen.toml
@@ -17,7 +17,10 @@ sys_includes = ["stdbool.h", "stddef.h", "stdint.h"]
 after_includes = """
 
 #define DDOG_CHARSLICE_C(string) \\
-/* NOTE: Compilation fails if you pass in a char* instead of a literal */ {.ptr = "" string, .len = sizeof(string) - 1}
+/* NOTE: Compilation fails if you pass in a char* instead of a literal */ ((ddog_CharSlice){ .ptr = "" string, .len = sizeof(string) - 1 })
+
+#define DDOG_CHARSLICE_C_BARE(string) \\
+/* NOTE: Compilation fails if you pass in a char* instead of a literal */ { .ptr = "" string, .len = sizeof(string) - 1 }
 
 #if defined __GNUC__
 #  define DDOG_GNUC_VERSION(major) __GNUC__ >= major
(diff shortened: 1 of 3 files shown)


bwoebi (approved):

bwoebi:
You could define `DDOG_CHARSLICE_C(str)` as `((ddog_CharSlice)DDOG_CHARSLICE_C_BARE(str))` :-D

ivoanjo:
> You could define DDOG_CHARSLICE_C(str) as ((ddog_CharSlice)DDOG_CHARSLICE_C_BARE(str)) :-D

I decided to leave as-is, to hopefully simplify a bit error messages (e.g. when passing in a non-literal), but don't have a strong opinion so if you're really not convinced I can live with your suggestion as well ;)

PR #348, opened 2024-03-11
[POC] Use hyper and allow deserialization/serialization

bantonsson:
# What does this PR do?

APM R&D week. Switch to using Hyper, and the existing deserialization/serialization.


data-pipeline/src/trace_exporter.rs
```diff
@@ -1,19 +1,124 @@
...
+        if self.no_proxy {
+            self.send_deser_ser(data)
+        } else {
+            self.send_proxy(data, trace_count)
+        }
+    }
+
+    fn send_proxy(self, data: Bytes, trace_count: usize) -> Result<String, String> {
+        let uri = self.endpoint.url.clone();
+        block_on(async {
```
paullegranddc, line 34: I'm not sure that guaranteed works? `hyper` futures need to be polled on a thread that has a tokio runtime activated, but if it's a same_thread runtime, that will deadlock, because you're blocking the runtime from advancing.
If you look at what the profiler does, they hold a `tokio::Runtime` in the profiling::Exporter struct and use with spawn_blocking to run async functions

bantonsson: Thanks. I'll fix that.

data-pipeline/src/trace_exporter.rs
```diff
@@ -25,60 +132,92 @@ pub struct TraceExporterBuilder {
...
+        self
     }
-}
 
+    pub fn build(&mut self) -> anyhow::Result<TraceExporter> {
+        let mut tags = TracerHeaderTags::default();
+        self.language
+            .as_ref()
+            .into_iter()
+            .for_each(|lang| tags.lang = lang);
```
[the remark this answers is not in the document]
bantonsson: 🤷🏼‍♂️ Probably some Scala residue that tipped me in that direction. It was a 1-liner before `rustfmt` had its way with it. I'll change them.

hoolioh (approved):

PR #352, opened 2024-03-12
data-pipeline-ffi

hoolioh:
# What does this PR do?

- Add FFI bindings to data-pipeline crate.
- Fix ownership issues where TraceExporterBuilder's string references were stored in TraceExporter object impeding the exporter outlive the builder.


data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn dd_trace_exporter_send(
+    ctx: *mut TraceExporter,
+    trace: ByteSlice,
+    trace_count: usize,
+) -> *const c_char {
+    let handle = Box::from_raw(ctx);
+    let response = handle
+        .send(Bytes::copy_from_slice(trace.as_bytes()), trace_count)
```
bantonsson, line 52: So can't this use the bytes directly instead of doing a copy? Isn't there a way to create bytes from a `Box<[u8]>`, or should the `send` in `TraceExporter` really operate on `&[u8]` instead of `Bytes`?

bantonsson (approved):
Only one comment.

To get to green, you need to add the `data-pipelie-ffi` project to the docker image here `tools/docker/Dockerfile.build` and update the 3rd party license file:

```
cargo bundle-licenses --format yaml --output ./LICENSE-3rdparty.yml
```

data-pipeline-ffi/src/trace_exporter.rs
bwoebi: This should operate on &[u8], I mean if you use Bytes, it will also free it, which it must not.

hoolioh: Good point. I didn't want to take ownership over the buffer just in case we needed to manipulate it. Maybe we can use COW semantics to use it and try to avoid the copy when it's not necessary. If you agree I will carry on with that change in another PR.

data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -0,0 +1,56 @@
...
+use data_pipeline::trace_exporter::TraceExporter;
+use data_pipeline::trace_exporter::TraceExporterBuilder;
+use ddcommon_ffi::{
+    slice::{AsBytes, ByteSlice},
+    CharSlice,
+};
+use std::ffi::{c_char, CString};
+
+#[no_mangle]
+pub unsafe extern "C" fn dd_trace_exporter_new(
```
morrisonlevi, line 14: Was there a discussion anywhere about `dd_trace`, `ddtrace`, etc? We've generally been using `ddog_` as a prefix to avoid possible conflicts in client libraries (which have been around a lot longer) and may be using those `dd_trace`, `ddtrace` style prefixes.

It's not a blocker, just curious.

hoolioh: No discussion at all :), just pick that name up to start coding. I'm gonna change it to be consistent with the rest of the crates. Thanks for the suggestion.

PR #359, opened 2024-03-15
Rename profiling-ffi Endpoint to ProfilingEndpoint/prof_Endpoint to remove confusion with ddcommon::Endpoint

paullegranddc:
# What does this PR do?

We have currently have an issue with different types sharing the same name in the C headers, but no being the same.
Cbindgen doesn't understand rust modules, so two types sharing the same name in different modules will be considered as the same for cbindgen.
The ffi currently manipulates two Endpoint types which are roughly equivalent
* ddcommon::Endpoint [code](https://github.com/DataDog/libdatadog/blob/9f4f3cf58f57763c9ef22c6930597c69d5e44ad2/ddcommon/src/lib.rs#L34), used as an opaque pointer in ffi for telemetry and the sidecar
 * datadog_profiling_ffi::Endpoint [code](https://github.com/DataDog/libdatadog/blob/9f4f3cf58f57763c9ef22c6930597c69d5e44ad2/profiling-ffi/src/exporter.rs#L37) an repr(C) enum, which is passed through the ffi and eventually transformed into ddcommon::Endpoint

If you look at the common.h header, you'll see a definition for the ddog_Endpoint  struct and, void ddog_endpoint_drop(struct ddog_Endpoint*); but these types are not the same.

This PR renames the datadog_profiling_ffi::Endpoint type to datadog_profiling_ffi::ProfilingEndpoint and generates C definition with type name ddog_prof_Endpoint

# Motivation

# Additional Notes

# How to test the change?

## For Reviewers

### DIFF
--- a/bin_tests/src/bin/crashtracker_bin_test.rs
+++ b/bin_tests/src/bin/crashtracker_bin_test.rs
@@ -33,7 +33,7 @@ mod unix {
         crashtracker::init(
             CrashtrackerConfiguration {
                 create_alt_stack: true,
-                endpoint: Some(datadog_profiling::exporter::Endpoint {
+                endpoint: Some(ddcommon::Endpoint {
                     url: ddcommon::parse_uri(&format!("file://{}", output_filename))?,
                     api_key: None,
                 }),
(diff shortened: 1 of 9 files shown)


r1viollet (approved):
LGTM
This seems like a dangerous issue for libdatadog users. Thanks for fixing.
Minor:why do we need 2 different endpoints?

PR #362, opened 2024-03-17
[trace-obfuscation] Add SQL obfuscator and make other obfuscators run faster

paullegranddc:
# What does this PR do?

This PR:

* Adds a simple SQL obfuscator based on the java and dotnet implementations
* Refactor the existing redis obfuscator to use less allocations, bringing a nice (60%) gain in performance
* Refactor the existing trace tags obfuscator  to use less allocations and pre-compute some values, bringing a nice (40%) gain in performance
* Refactor the trace-obfuscation benches and adding some new ones.

The changes to the existing redis and tags obfuscators should not affect their semantics as they are still passing the existing tests.
 
# Motivation

An SQL obfuscator is needed to make the stats computation in data-pipeline implementation spec compliant. 

# Additional Notes

# How to test the change?

## For Reviewers
- [ x ] This PR doesn't touch any of that.


trace-obfuscation/benches/credit_cards_bench.rs
```diff
@@ -44,4 +44,3 @@ fn is_card_number_no_luhn_bench(c: &mut Criterion) {
 }
 
 criterion_group!(benches, is_card_number_bench, is_card_number_no_luhn_bench);
-criterion_main!(benches);
```
bantonsson, line 47: I'm a bit confused to how these things fit together. In `main` it's possible to run `cargo bench --bench credit_cards_bench` and it will run the these benchmarks only, but in this branch, they are skipped, and nothing is run.

Is the `benches_main` thing really necessary, and shouldn't the different bench files be defined in the `Cargo.toml` file?

trace-obfuscation/src/redis_tokenizer.rs
```diff
@@ -1,179 +1,141 @@
...
-            escape = false
+            escape = false;
+            self.offset += 1;
         }
+        (start, self.offset)
     }
 
-    fn next(&mut self) {
-        if self.has_started {
+    fn skip_whitespace(&mut self) {
```
bantonsson, line 122: Very nice, keeping the code DRY.

trace-obfuscation/src/sql.rs
```diff
@@ -0,0 +1,271 @@
...
+
+fn is_hex_litteral_prefix(bytes: &[u8], start: usize, end: usize) -> bool {
+    (bytes[start] | b' ') == b'x' && start + 1 < end && bytes[start + 1] == b'\''
+}
+
+fn is_quoted(bytes: &[u8], start: usize, end: usize) -> bool {
+    bytes[start] == b'\'' && bytes[end - 1] == b'\''
+}
+
+/// Normalizes an sql string by replacing litterals with '?' chars.
```
bantonsson, line 28: Confusing to have SQL normailization in the obfuscation project.

bantonsson (commented):
Nice work. My only real comment is about the wiring up and running of the benchmarks, standard [documented rust way](https://doc.rust-lang.org/cargo/commands/cargo-bench.html#target-selection) is not working anymore.

trace-obfuscation/src/sql.rs
```diff
@@ -0,0 +1,271 @@
...
+///
+/// The algorithm works by finding the places where a litteral could start (so called splitters)
+/// and then identifies them by looking at their first few characters.
+///
+/// It does not attempt at rigorous parsing of the SQL syntax, and does not take any context
+/// sensitive decision, contrary to the more exhaustive datadog-agent implementation.
+///
+/// based off
+/// https://github.com/DataDog/dd-trace-java/blob/36e924eaa/internal-api/src/main/java/datadog/trace/api/normalize/SQLNormalizer.java
+pub fn obfsucate_sql_string(s: &str) -> String {
```
bantonsson, line 38: Spelling `obfuscate`

trace-obfuscation/benches/credit_cards_bench.rs
paullegranddc: Hmm true, I added this because I found it annoying to edit the Cargo.toml file and was under the impression that it ran the linking part for each individual benchmark binary, which raised compile times when running all of them at once.

I was inspired by how the criterion-rs repo organizes their own benchmarks. https://github.com/bheisler/criterion.rs/tree/master/benches

You can still use `cargo bench -p datadog-trace-obfuscation card_number` to run the card number benches.

trace-obfuscation/src/sql.rs
paullegranddc: The confusion spawns from the fact that I was following the java implementation which is named `SQLNormalizer`, when in practice, it is part of the obfuscation for trace stats in library spec

trace-obfuscation/benches/credit_cards_bench.rs
paullegranddc: Having a single bench binary also makes editing the Dockerfile for testing somewhat simpler since you don't have to add each of them individually https://github.com/DataDog/libdatadog/blob/main/tools/docker/Dockerfile.build#L106

trace-obfuscation/src/sql.rs
bantonsson: Yeah, the line between `normalization` and `obfuscation` isn't that clear cut. I prefer to think about this as _normalizing_ the SQL statements when you remove _irrelevant_ (and maybe sensitive) data for the stats grouping, but let's not dwell on it.

trace-obfuscation/benches/credit_cards_bench.rs
bantonsson: I don't see any changes to the docker file in this PR, so how would it change?

paullegranddc: I added the changes (the alpine test were failing without them)

bantonsson: @paullegranddc I couldn't let got of the weird way that `cargo bench --bench` was displaying benches that doesn't exist (leading to confusion), and the solution is to move the benchmarks into a sub module in the same way as is done in the `criterion` repo https://github.com/bheisler/criterion.rs/tree/master/benches/benchmarks

Could we also rename the `benches_main` to `trace_obfuscation` to make it more clear where the `bench` lives? Then it would be possible to do things like `cargo bench --bench trace_obfuscation -- --list` to see all benchmarks available there.

paullegranddc: done

bantonsson (approved):

Issue #353, opened 2024-03-12
[FEATURE] Build system handles different feature configurations


r1viollet:
## Description

When building libraries in libdatadog, we want to ensure that it is easy to add in or remove modules.
Any logics associated to a module should be within the associated folder.

This means that shell scripts with embedded build logics should be removed.

Example of logics to migrate
https://github.com/DataDog/libdatadog/blob/main/build-telemetry-ffi.sh
https://github.com/DataDog/libdatadog/blob/main/build-profiling-ffi.sh

Package config generation:
https://github.com/DataDog/libdatadog/blob/main/profiling-ffi/datadog_profiling.pc.in

Associated FFI examples:
https://github.com/DataDog/libdatadog/tree/main/examples/ffi

## **Describe the goal of the feature**

The goal of the feature is to be able to do:
```
cargo build --feature crash-tracking,telemetry,profiling
cargo build --feature trace-pipeline,profiling
```
Which would produce the expected deliverable.
The associated test commands should work.

## **Describe alternatives you've considered**
We can review the [RFC](https://github.com/DataDog/libdatadog/blob/main/docs/RFCs/delivering-different-features.md).
The aim is to have a single build for Rust features.

## **Additional context**
As we are looking to add tracing features in the build, we want a clean build system.

brettlangdon:
The approach I have been taking in Python is build the Rust FFI/wrappers in the dd-trace-py repo. The goal being I want to be able to pull in specific crates only in our own Cargo.toml instead of worrying about downloading binaries/artifacts.

I've seen the benefit a lot as being able to pick and choose only what I need, as well as building the rust extensions on the same systems that we are building the library on this way we don't have to try and duplicate build environments (os/arch/etc) across multiple repos.

I _think_ this approach should work for most languages. Do you think this is a viable path to achieve the same goals you have? Or are we always going to have to build the FFI/binaries for libdatadog anyways?

https://github.com/DataDog/dd-trace-py/tree/f60060dbed3841244abead9008ff04eb804461af/src/core

r1viollet:
> I _think_ this approach should work for most languages. Do you think this is a viable path to achieve the same goals you have? Or are we always going to have to build the FFI/binaries for libdatadog anyways?
> 
> https://github.com/DataDog/dd-trace-py/tree/f60060dbed3841244abead9008ff04eb804461af/src/core

Rebuilding from source has some real benefits. I know PHP have chosen this path.

[...]

PR #368, opened 2024-03-25
Initial code and CI for trace data pipeline

bantonsson:
# What does this PR do?

The work done on a shared component data pipeline done during the R&D week.

# Motivation

Add a base build and CI that we can continue to iterate on to refine the API and functionality of the shared data pipeline component.

# Additional Notes

Please don't be to hard on the public APIs, and the implementation. They will evolve over time, but in the interest of not having a ginormous PR from a feature branch, we would like to get the machinery into main.

# How to test the change?

## For Reviewers


paullegranddc:
Are we going to ship different shared libraries for data-pipeline and profiling?

bantonsson:
We are not there yet, and in some languages we definitely will since they don't use the profiler in libdatadog.

ivoanjo:
Not a blocker at this point, but more of an observation, note that shipping profiling AND data-pipeline separately for the same language will probably increase the binary size a lot more than if a unified libdatadog was produced ;)

bantonsson:
Yes we are aware of the added binary size and are going to work for reducing this and combining the features into a single binary where it makes sense.

paullegranddc:
> Not a blocker at this point, but more of an observation, note that shipping profiling AND data-pipeline separately for the same language will probably increase the binary size a lot more than if a unified libdatadog was produced ;)

It's also going to be hard to load both of them in the same process because of symbols exposed in ddcomon being in both shared library

ivoanjo:
> It's also going to be hard to load both of them in the same process because of symbols exposed in ddcomon being in both shared library

Good point! [For Ruby we explicitly load the profiling native extension with `RTLD_LOCAL`](https://github.com/DataDog/dd-trace-rb/blob/master/ext/datadog_profiling_loader/datadog_profiling_loader.c#L68) and I believe that's a potential solution to this issue.

I'm saying this more out of curiosity / sharing notes :)

build-data-pipeline-ffi.sh
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env bash
+
+# Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+# SPDX-License-Identifier: Apache-2.0
+
+set -eu
+
+destdir="$1"
+
```
r1viollet, line 9: do we want to fix the build logics before adding the tracing logics or after ?

bantonsson: Could you please elaborate a bit on what this would mean.

r1viollet: This is relative to the RFC + issue https://github.com/DataDog/libdatadog/issues/353
The idea would be to have a single build where we can select what is required to build.
A good first step could be to unify the build scripts.
Ideally all logics would be embedded in cargo in the long term.
cargo build --feature trace-pipeline,profiling,telemetry

gleocadie (commented):
LGTM
can you remove this change please windows/libdatadog.csproj
It will generate a +300MB nuget package. We need to rethink how we package this on windows. CI (dd-trace-dotnet for example, this can be a problem on the runners - not enough disk space)

build-data-pipeline-ffi.sh
hoolioh: Fixed in the latest commit. As discussed internally profiling library will be the main artifact and data-pipeline symbols will be exported there.

profiling-ffi/src/lib.rs
```diff
@@ -18,6 +18,10 @@ pub use crashtracker::*;
 #[cfg(feature = "ddtelemetry-ffi")]
 pub use ddtelemetry_ffi::*;
 
+#[cfg(feature = "data-pipeline-ffi")]
+#[allow(unused_imports)]
+pub use data_pipeline_ffi::*;
```
r1viollet, line 23: :+1:

build-data-pipeline-ffi.sh
r1viollet: Why do we export the symbols here ?

hoolioh: Forgot the last commit (pushed it but the prompt asked for the credentials while I was AFK) now it's complete.

examples/ffi/trace_exporter.c
```diff
@@ -0,0 +1,41 @@
...
+int main(int argc, char** argv)
+{
+    ddog_CharSlice host = DDOG_CHARSLICE_C("localhost");
+    uint16_t port = 8126;
+    ddog_CharSlice tracer_version = DDOG_CHARSLICE_C("v0.1");
+    ddog_CharSlice language = DDOG_CHARSLICE_C("dotnet");
+    ddog_CharSlice language_version = DDOG_CHARSLICE_C("10.0");
+    ddog_CharSlice language_interpreter = DDOG_CHARSLICE_C("X");
+    ddog_TraceExporter* trace_exporter = ddog_trace_exporter_new(
+        host,
```
r1viollet, line 18: we have an endpoint object that could be useful to share

.github/workflows/test.yml
```diff
@@ -97,7 +97,7 @@ jobs:
       - name: "Generate profiling FFI"
```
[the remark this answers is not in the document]
r1viollet: The important part is having at least one build per delivered combination.

r1viollet (approved):
LGTM
Minor suggestions. Exciting to see this progress :clap:

PR #369, opened 2024-03-26
style(profiling): remove redundant imports

morrisonlevi:
# What does this PR do?

This removes redundant imports in the `profiling/` subdir.

# Motivation

Miri was complaining about these for me on nightly. 

# Additional Notes

It is probable other crates will have similar warnings. I only worked on `profiling/`'s.

# How to test the change?

Everything should compile and work as normal, this should be a net-zero change.

## For Reviewers


ivoanjo (approved):
:+1: LGTM, thanks for the cleanup!

bantonsson (approved):

PR #372, opened 2024-03-27
build: clean up profiling deps

morrisonlevi:
# What does this PR do?

Trims down the dependencies of the `profiling/` subdirectory, and updates patch versions of some dependencies as well.

# Motivation

The removed deps were added when crashtracking was in this subdir, but it's been moved out and these were left.

# Additional Notes

None.

# How to test the change?

Everything should work as before without any changes.

## For Reviewers


morrisonlevi:
The test failure is:

> LLVM ERROR: IO failure on output stream: No space left on device

I've seen someone already ask about this on Slack, so I'll ignore for now.

ivoanjo (approved):
:+1: Thanks for the great little cleanup :)

PR #373, opened 2024-03-29
Fix linking issue on Windows

gleocadie:
# What does this PR do?

This PR fixes a linking with for Windows native libraries that link against the VC runtime statically that was introduced by [the PR adding telemetry.
](https://github.com/DataDog/libdatadog/pull/339)
# Motivation

The .NET profiler links against the VC runtime statically so it can be deployed easily everywhere (on windows) without the customers having to install additional stuffs (VC runtime for example).

The PR that introduced the telemetry in the datadog_profiling lib, brought some new dependencies. Now the .NET profiler is unable to successfully build on windows.
The build produces this linking error:
```
datadog_profiling.lib(windows.o) : error LNK2019: unresolved external symbol __imp_strncpy referenced in function get_os_release [D:\a\libdatadog\libdatadog\examples\ffi\build_static\telemetry_metrics.vcxproj]
```

# Implementation

To fix this, I added a `.cargo/config.toml` file in the `profiling-ffi` crate and build the crate directly inside (not from the workspace, otherwise, we will have to move the `config.toml` in the workspace and it will apply to all crates)

# Additional Notes

Nop

# How to test the change?

I added in the CI 2 configurations when building and testing FFI:
- One where the projects are linked dynamically against the VC runtime
- One where the projects are linked statically against the VC runtime.

## For Reviewers

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,13 +97,11 @@ jobs:
       - name: "Generate profiling FFI"
         shell: bash
         run: |
-           chmod +x build-profiling-ffi.sh
            ./build-profiling-ffi.sh ${OUTPUT_FOLDER}/profiling
 
       - name: "Generate Telemetry FFI"
         shell: bash
         run: |
-           chmod +x build-telemetry-ffi.sh
            ./build-telemetry-ffi.sh ${OUTPUT_FOLDER}/telemetry
 
       - name: 'Publish libdatadog'
@@ -115,8 +113,29 @@ jobs:
           path: ${{ github.workspace }}/artifacts
           retention-days: 1
 
+      - name: "(Windows) Test building C bindings - dynamic link vc runtime"
+        if: ${{ matrix.platform == 'windows-latest' }}
+        shell: bash
+        run: |
+          set -e
+          mkdir examples/ffi/build_dll
+          cd examples/ffi/build_dll
+          cmake -S .. -DDatadog_ROOT=$OUTPUT_FOLDER/profiling -DVCRUNTIME_LINK_TYPE=DLL
+          cmake --build .
+
+      - name: "(Windows) Test building C bindings - static link vc runtime"
+        if: ${{ matrix.platform == 'windows-latest' }}
+        shell: bash
+        run: |
+          set -e
+          mkdir examples/ffi/build_static
+          cd examples/ffi/build_static
+          cmake -S .. -DDatadog_ROOT=$OUTPUT_FOLDER/profiling -DVCRUNTIME_LINK_TYPE=STATIC
+          cmake --build .
+
       - name: "Test building C bindings"
         shell: bash
+        if: ${{ matrix.platform != 'windows-latest' }}
         run: |
           set -e
           mkdir examples/ffi/build
--- a/build-profiling-ffi.sh
+++ b/build-profiling-ffi.sh
@@ -100,7 +100,6 @@ sed < cmake/DatadogConfig.cmake.in \
 
 cp -v LICENSE LICENSE-3rdparty.yml NOTICE "$destdir/"
 
-export RUSTFLAGS="${RUSTFLAGS:- -C relocation-model=pic}"
 
 datadog_profiling_ffi="datadog-profiling-ffi"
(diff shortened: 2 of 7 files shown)


build-profiling-ffi.sh
```diff
@@ -162,7 +162,6 @@ unexpected_native_libs=$(comm -13 <(echo "$unique_expected_libs") <(echo "$uniqu
 if [ -n "$unexpected_native_libs" ]; then
     echo "Error - More native static libraries are required for linking than expected:" 1>&2
     echo "$unexpected_native_libs" 1>&2
-    exit 1
```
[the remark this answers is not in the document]
gleocadie: no you are right. It was during my debugging time (that's why it's still a draft :P)

r1viollet (approved):
LGTM

r1viollet (approved):
LGTM

.github/workflows/test.yml
```diff
@@ -115,8 +113,29 @@ jobs:
           path: ${{ github.workspace }}/artifacts
           retention-days: 1
 
+      - name: "(Windows) Test building C bindings - dynamic link vc runtime"
+        if: ${{ matrix.platform == 'windows-latest' }}
+        shell: bash
```
danielsn, line 118: It never ceases to amaze me that there is now supported bash on windows

.github/workflows/test.yml
```diff
@@ -115,8 +113,29 @@ jobs:
...
+        run: |
+          set -e
```
danielsn, line 120: for debugging, might be useful to `set -eux`

.github/workflows/test.yml
```diff
@@ -115,8 +113,29 @@ jobs:
...
+          cmake --build .
+
+      - name: "(Windows) Test building C bindings - static link vc runtime"
+        if: ${{ matrix.platform == 'windows-latest' }}
+        shell: bash
+        run: |
+          set -e
+          mkdir examples/ffi/build_static
+          cd examples/ffi/build_static
+          cmake -S .. -DDatadog_ROOT=$OUTPUT_FOLDER/profiling -DVCRUNTIME_LINK_TYPE=STATIC
```
danielsn, line 133: Do we also need to test FFI in both configurations?

build-profiling-ffi.sh
```diff
@@ -100,15 +100,17 @@ sed < cmake/DatadogConfig.cmake.in \
...
 FEATURES="--features cbindgen,datadog-profiling-ffi/ddtelemetry-ffi"
 if [[ "$symbolizer" -eq 1 ]]; then
     FEATURES="--features cbindgen,datadog-profiling-ffi/ddtelemetry-ffi,symbolizer"
 fi
 
-DESTDIR="$destdir" cargo build --package="${datadog_profiling_ffi}" ${FEATURES} --release --target "${target}"
+# build inside the crate to use the config.toml file
+pushd profiling-ffi
+DESTDIR="$destdir" cargo build ${FEATURES} --release --target "${target}"
+popd
```
danielsn, line 113: Would it be cleaner to 
```suggestion
( cd profiling-ffi && DESTDIR="$destdir" cargo build ${FEATURES} --release --target "${target}")
```

profiling-ffi/.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[build]
+rustflags = ["-C", "relocation-model=pic"]
```
danielsn, line 2: Does this have any performance implications?

.github/workflows/test.yml
danielsn: https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425

profiling-ffi/.cargo/config.toml
gleocadie: TBH, I do not know. I just moved this from the bash script

.github/workflows/test.yml
gleocadie: Yeah Git bash is installed :) Not really conventional but it works 🤷

danielsn (approved):

PR #374, opened 2024-03-29
[crashtracker] Chain signal handlers 

danielsn:
# What does this PR do?

Reworks the signal handler chaining code to properly handle cases where the signal was manually raised.

# Motivation

The example in `crashtracker.c` would fail to call the underlying handler in the case of a `raise(SEGV)` call, and then would continue past what ought to be a core-dump.  Now we don't do that anymore.

# Additional Notes

The POSIX signal infra is 🤯 

# How to test the change?

`crashtracking.c` now works the way its expected to.

## For Reviewers

### DIFF
--- a/crashtracker/src/crash_handler.rs
+++ b/crashtracker/src/crash_handler.rs
@@ -212,6 +212,8 @@ extern "C" fn handle_posix_sigaction(signum: i32, sig_info: *mut siginfo_t, ucon
     // Once we've handled the signal, chain to any previous handlers.
     // SAFETY: This was created by [register_crash_handlers].  There is a tiny
     // instant of time between when the handlers are registered, and the
+    // `OLD_HANDLERS` are set.  This should be very short, but is hard to fully
+    // eliminate given the existing POSIX APIs.
     let old_handlers = unsafe { &*OLD_HANDLERS.load(SeqCst) };
     let old_sigaction = if signum == libc::SIGSEGV {
         old_handlers.sigsegv
--- a/crashtracker/src/crash_handler.rs
+++ b/crashtracker/src/crash_handler.rs
@@ -11,8 +11,8 @@ use super::constants::*;
 use super::counters::emit_counters;
 use anyhow::Context;
 use libc::{
-    mmap, sigaltstack, MAP_ANON, MAP_FAILED, MAP_PRIVATE, PROT_NONE, PROT_READ, PROT_WRITE,
-    SIGSTKSZ,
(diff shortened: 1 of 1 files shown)


paullegranddc (approved):

paullegranddc:
Could you edit the CODEOWNERS file in a separate PR to put the crashtracker crate under the responsability of Datadog/libdatadog-profiling?

crashtracker/src/crash_handler.rs
```diff
@@ -204,14 +204,54 @@ pub fn shutdown_receiver() -> anyhow::Result<()> {
...
+extern "C" fn handle_posix_sigaction(signum: i32, sig_info: *mut siginfo_t, ucontext: *mut c_void) {
+    // Handle the signal.  Note this has a guard to ensure that we only generate
+    // one crash report per process.
     let _ = handle_posix_signal_impl(signum);
 
-    // return to old handler (chain).  See comments on `restore_old_handler`.
+    // Once we've handled the signal, chain to any previous handlers.
+    // SAFETY: This was created by [register_crash_handlers].  There is a tiny
+    // instant of time between when the handlers are registered, and the
+    let old_handlers = unsafe { &*OLD_HANDLERS.load(SeqCst) };
```
ivoanjo, line 215: Minor: Comment cut off abruptly?

crashtracker/src/crash_handler.rs
```diff
@@ -244,6 +284,13 @@ fn emit_siginfo(w: &mut impl Write, signum: i32) -> anyhow::Result<()> {
...
+    static NUM_TIMES_CALLED: AtomicU64 = AtomicU64::new(0);
+    if NUM_TIMES_CALLED.fetch_add(1, SeqCst) > 0 {
+        // In the case where some lower-level signal handler recovered the error
+        // we don't want to spam the system with calls.  Make this one shot.
+        return Ok(());
+    }
+
     // Leak receiver, config, and metadata to avoid calling 'drop' during a crash
     let receiver = RECEIVER.swap(ptr::null_mut(), SeqCst);
     anyhow::ensure!(!receiver.is_null(), "No crashtracking receiver");
```
ivoanjo, line 296: I guess this is somewhat harmless, but wasn't the existing code already handling this correctly? E.g. the first crash would grab the receiver and any other crash would return with the "No crashtracking receiver" option.

ivoanjo (approved):
This looks reasonable. Two notes:

1. May be worth getting a review from the native profiler wizards (e.g. @sanchda / @nsavoire / @r1viollet )
2. I know testing this is a pain but... have you considered perhaps modifying the crash tracker example to have a few more situations?

sanchda (approved):
LGTM!

PR #379, opened 2024-04-03
[profiling] Refactor FFI implementation to simplify error handling

danielsn:
# What does this PR do?

Adds a `.into` method to the FFI error types to auto convert from Rust Error types, and then uses that to make cleaner error handling possible in the FFI functions.

# Motivation

The old way had cruft and was hard to read.  This is nicer.

# Additional Notes

# How to test the change?

## For Reviewers


r1viollet (approved):
LGTM
Thanks for figuring out how to use anyhow results!

PR #385, opened 2024-04-10
Fix sidecar_mockgen for mac

bwoebi:
On mac symbol sizes are not known and underscore prefixed.

This allows the sidecar to not crash when built as a shared php extension.


bantonsson (commented):
Hard to understand if it has been fixe since there is nothing explaining what it is supposed to do.
What is the `sidecar_mockgen` really?

bantonsson (approved):

PR #388, opened 2024-04-11
Minor adjustments to the data-pipeline/TraceExporter

ekump:
# What does this PR do?

Just some minor changes to the TraceExporter.

- Added some tests around the `From` trait implementations for `TracerTags`. These were already tested indirectly. 
- Tweaked the unit tests to test the end result of calling the builder rather than the values of the builder struct fields.
- Changed `no_proxy` to `use_proxy` to be (possibly) clearer and added necessary default trait impl for the builder.
- Made the TracerTag fields on `TraceExporterBuilder` non-optional since they weren't actually optional. 

# Motivation

I incorrectly started to work on TraceExporter for APMSP-1020 before I was informed we should be implementing that ticket in the sidecar's trace export logic for now. (We'll address implementing in TraceExporter later). 

# Additional Notes

# How to test the change?

Unit tests are in place

## For Reviewers

### DIFF
--- a/data-pipeline/src/trace_exporter.rs
+++ b/data-pipeline/src/trace_exporter.rs
@@ -157,10 +157,10 @@ impl TraceExporter {
 pub struct TraceExporterBuilder {
     host: Option<String>,
     port: Option<u16>,
-    tracer_version: Option<String>,
-    language: Option<String>,
-    language_version: Option<String>,
-    interpreter: Option<String>,
+    tracer_version: String,
+    language: String,
+    language_version: String,
+    language_interpreter: String,
     no_proxy: bool,
 }
 
@@ -176,17 +176,17 @@ impl TraceExporterBuilder {
     }
 
     pub fn set_tracer_version(&mut self, tracer_version: &str) -> &mut TraceExporterBuilder {
-        self.tracer_version = Some(String::from(tracer_version));
+        self.tracer_version = tracer_version.to_owned();
         self
     }
 
     pub fn set_language(&mut self, lang: &str) -> &mut TraceExporterBuilder {
-        self.language = Some(String::from(lang));
+        self.language = lang.to_owned();
         self
     }
 
     pub fn set_language_version(&mut self, lang_version: &str) -> &mut TraceExporterBuilder {
-        self.language_version = Some(String::from(lang_version));
+        self.language_version = lang_version.to_owned();
         self
     }
 
@@ -194,7 +194,7 @@ impl TraceExporterBuilder {
         &mut self,
         lang_interpreter: &str,
     ) -> &mut TraceExporterBuilder {
-        self.interpreter = Some(String::from(lang_interpreter));
+        self.language_interpreter = lang_interpreter.to_owned();
         self
     }
 
@@ -217,16 +217,18 @@ impl TraceExporterBuilder {
             )?,
             api_key: None,
         };
+
         let runtime = tokio::runtime::Builder::new_current_thread()
             .enable_all()
             .build()?;
+
         Ok(TraceExporter {
             endpoint,
             tags: TracerTags {
-                tracer_version: self.tracer_version.clone().unwrap(),
-                language_version: self.language_version.clone().unwrap(),
-                language_interpreter: self.interpreter.clone().unwrap(),
-                language: self.language.clone().unwrap(),
+                tracer_version: std::mem::take(&mut self.tracer_version),
+                language_version: std::mem::take(&mut self.language_version),
+                language_interpreter: std::mem::take(&mut self.language_interpreter),
+                language: std::mem::take(&mut self.language),
             },
             no_proxy: self.no_proxy,
             runtime,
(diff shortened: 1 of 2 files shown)


data-pipeline/src/trace_exporter.rs
```diff
@@ -256,12 +258,11 @@ mod tests {
...
             "http://192.168.1.1:8127/v0.7/traces"
         );
-        assert_eq!(builder.host.unwrap(), "192.168.1.1");
-        assert_eq!(builder.port.unwrap(), 8127);
-        assert_eq!(builder.tracer_version.unwrap(), "v0.1");
-        assert_eq!(builder.language.unwrap(), "nodejs");
-        assert_eq!(builder.language_version.unwrap(), "1.0");
-        assert_eq!(builder.interpreter.unwrap(), "v8");
+
+        assert_eq!(exporter.tags.tracer_version, "v0.1");
```
ekump, line 262: I switched to testing the exporter tags because this test is calling `build()` which sets the exporter tags and we should be testing outputs in unit tests, unless we have a special reason not to.

data-pipeline/src/trace_exporter.rs
```diff
@@ -157,10 +157,10 @@ impl TraceExporter {
 pub struct TraceExporterBuilder {
     host: Option<String>,
     port: Option<u16>,
-    tracer_version: Option<String>,
```
ekump, line 160: The unwrapping of the tag values in the build function and the fact that the corresponding `TracerTag` values are not optional implies that they aren't actually optional. I switched them to `String` since that better signals intent and also leads to a very minor perf improvement in the builder (not a primary concern). 

If there are future plans that rely on these values truly being options I can change them back.

data-pipeline/src/trace_exporter.rs
```diff
@@ -176,25 +176,25 @@ impl TraceExporterBuilder {
     }
 
     pub fn set_tracer_version(&mut self, tracer_version: &str) -> &mut TraceExporterBuilder {
-        self.tracer_version = Some(String::from(tracer_version));
+        self.tracer_version = tracer_version.to_owned();
```
ekump, line 179: I've always known `to_owned()` to be idiomatic for going from &str to String. Happy to revert this if it was done intentionally.

data-pipeline/src/trace_exporter.rs
```diff
@@ -217,16 +217,18 @@ impl TraceExporterBuilder {
...
             .build()?;
+
         Ok(TraceExporter {
             endpoint,
             tags: TracerTags {
-                tracer_version: self.tracer_version.clone().unwrap(),
-                language_version: self.language_version.clone().unwrap(),
-                language_interpreter: self.interpreter.clone().unwrap(),
-                language: self.language.clone().unwrap(),
+                tracer_version: std::mem::take(&mut self.tracer_version),
```
ekump, line 228: I haven't run benchmarks, but using `mem::take` is slightly more performant versus `to_owned()` because `to_owned()` does another memory allocation. `mem::take` just swaps the existing value with the default. The downside to this is that the builder doesn't retain values after `build()` is called and thus can cause unexpected results if it is reused. I don't think I've ever reused a builder after I finished building, so I don't think it should be a problem? But, I'm happy to change this to `to_owned()` if people disagree.

data-pipeline/src/trace_exporter.rs
```diff
@@ -279,14 +280,20 @@ mod tests {
...
-        assert_eq!(builder.interpreter.unwrap(), "v8");
+        assert_eq!(exporter.tags.tracer_version, "v0.1");
+        assert_eq!(exporter.tags.language, "nodejs");
+        assert_eq!(exporter.tags.language_version, "1.0");
+        assert_eq!(exporter.tags.language_interpreter, "v8");
     }
 
     #[test]
-    fn configure() {}
+    #[ignore]
```
ekump, line 290: This may seem insignificant, but I think this better conveys intention for these empty tests. Otherwise, they always silently pass (reducing the likelihood of never being implemented). Now, cargo will at least print that the tests were ignored. And if you run ignored tests via `cargo test -- --ignored` you'll get a failure saying these tests have not been implemented.

data-pipeline/src/trace_exporter.rs
```diff
@@ -153,15 +153,28 @@ impl TraceExporter {
...
-    interpreter: Option<String>,
-    no_proxy: bool,
+    tracer_version: String,
+    language: String,
+    language_version: String,
+    language_interpreter: String,
+    use_proxy: bool,
+}
+
+impl Default for TraceExporterBuilder {
```
ekump, line 166: I'm split on whether or not this is a good idea. On the one hand, we should avoid negative bool variables like `no_proxy` when possible as it can easily lead to confusion versus bools that are affirmatively named. On the other hand, I see why `no_proxy` was probably chosen in the first place since we want the default behavior to use the proxy and bools default to false naturally. Instead of using derive Default macro we had to implement it manually which can seem like overkill for just one struct field.

I can see an argument that this change makes the code more complicated in its attempt to make it simpler so I can revert back to `no_proxy` if anyone feels strongly.

data-pipeline/src/trace_exporter.rs
hoolioh: It was done intentionally. The reasoning behind it is, at least in this case, the builder won't cross multiple function boundaries. It's called from an init function and it's never reused so it doesn't need to own all the resources. Therefore cloning all the values and then dropping them at the end of the function didn't seem a good choice in terms of memory management. If you foresee any use case in which the builder has to own all the resources go for it if not it's probably better to not allocate all that strings but that's my opinion.

ekump: Nowadays, `String::From` and `to_owned()` are functionally equivalent, in both cases you're creating a String that the Builder owns. From a style perspective I personally prefer `to_owned()` as it looks cleaner (to me) and conveys intention of creating something the builder will own. 

If the intention is that the builder should only hold the references, then I can change it to do that. But, is that something we really want to do? Because at the end of the day we probably want `TracerTags` to not hold references as they will likely get used in another thread and the function that calls the builder may drop the references before we're done with them?

We could keep `TracerTags` members as `String`, but have the Builder use `&str` and do the clone in the `build()` function to defer the memory allocation as long as possible if we think it's worth it.

bantonsson: I'm in favor of `to_owned` as well.

data-pipeline/src/trace_exporter.rs
bantonsson: If the `build` function is _destructive_ in that way, then it should take ownership of the builder with `fn build(self) ...`, to avoid misuse.

bantonsson (approved):

hoolioh (approved):

PR #389, opened 2024-04-11
[crashtracker] Add `args` and `env` parameters to the receiver

danielsn:
# What does this PR do?

Adds `args` and `env` parameters to allow clients to control that facet of the receiver

# Motivation

As crashtracking is used in more places, we see use cases for setting args and env.  I also cleaned up the structs.
# Additional Notes

# How to test the change?

## For Reviewers


crashtracker/src/configuration.rs
```diff
@@ -31,27 +55,21 @@ impl CrashtrackerConfiguration {
...
-        );
-        anyhow::ensure!(stderr_filename.is_none() && stdout_filename.is_none() || stderr_filename != stdout_filename,
-        "Can't give the same filename for stderr and stdout, they will conflict with each other"
-    );
+        //     anyhow::ensure!(
+        //         !path_to_receiver_binary.is_empty(),
+        //         "Expected a receiver binary"
+        //     );
+        //     anyhow::ensure!(stderr_filename.is_none() && stdout_filename.is_none() ||
+        // stderr_filename != stdout_filename,     "Can't give the same filename for stderr
```
sanchda, line 66: Whimsical comment follows.

In many cases, stderr and stdout are already interleaved by the time they get to the user (either their terminal, or an aggregation platform that just rips everything _similarly_ to a terminal).  This system uses `dup2()` in order to duplicate the file descriptors, so the interleaving would actually happen at the level of the `write()` syscall.  In userland, even libc doesn't know better and will provide two separate buffers to the singular file descriptor, but that doesn't decrease the safety of the operation.

I sense that this situation may actually be fine (in the sense that it is precisely what the user asked for and should have no _unwanted_ side-effects).  Of course, if someone complains that this interleaves the outputs in a confusing way, then they deserve the stick-bicycle meme.  Possibly twice.

crashtracker/src/configuration.rs
```diff
@@ -18,11 +18,35 @@ pub struct CrashtrackerConfiguration {
     pub collect_stacktrace: bool,
     pub create_alt_stack: bool,
     pub endpoint: Option<Endpoint>,
-    pub path_to_receiver_binary: String,
     pub resolve_frames: CrashtrackerResolveFrames,
+    pub timeout: Duration,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct CrashtrackerReceiverConfig {
```
sanchda, line 26: As mostly an end-user, this makes sense.

examples/ffi/crashtracking.c
```diff
@@ -41,7 +49,6 @@ int main(int argc, char **argv) {
   }
 
   // raise(SIGSEGV);
```
sanchda, line 51: Does this still need to be here?  If it's useful, might be interesting to toggle with a preprocessor macro.

examples/ffi/crashtracking.c
```diff
@@ -18,21 +18,29 @@ int main(int argc, char **argv) {
     return -1;
   }
 
+  ddog_prof_CrashtrackerReceiverConfig receiver_config = {
+      .args = {},
+      .env = {},
+      .path_to_receiver_binary = DDOG_CHARSLICE_C("FIXME - point me to receiver binary path"),
```
sanchda, line 24: Will this be ironed out?  If not, suggest changing the string literal to explicitly tell the user that they are expected to adapt this to their build process.

profiling-ffi/src/crashtracker/datatypes.rs
```diff
@@ -9,18 +9,30 @@ use ddcommon_ffi::{Error, Slice, StringWrapper};
...
+    key: CharSlice<'a>,
+    val: CharSlice<'a>,
+}
+
+#[repr(C)]
+pub struct CrashtrackerReceiverConfig<'a> {
+    pub args: Slice<'a, CharSlice<'a>>,
+    pub env: Slice<'a, EnvVar<'a>>,
+    pub path_to_receiver_binary: CharSlice<'a>,
+    /// Optional filename to forward stderr to (useful for logging/debugging)
```
sanchda, line 23: If we keep the requirement that stderr != stdout, might be useful to specify right in the header that they are assumed to be unique

profiling-ffi/src/crashtracker/datatypes.rs
```diff
@@ -9,18 +9,30 @@ use ddcommon_ffi::{Error, Slice, StringWrapper};
...
+    pub optional_stdout_filename: CharSlice<'a>,
+}
+
 #[repr(C)]
 pub struct CrashtrackerConfiguration<'a> {
     /// Should the crashtracker attempt to collect a stacktrace for the crash
     pub collect_stacktrace: bool,
     pub create_alt_stack: bool,
-    /// The endpoint to send the crash repor to (can be a file://)
+    /// The endpoint to send the crash report to (can be a file://)
```
sanchda, line 34: Whoa!  I didn't know this!  Way cool!

sanchda (approved):
LGTM.  Left some minor notes/comments, but no dealbreakers.

profiling-ffi/src/crashtracker/datatypes.rs
danielsn: For now, we're opening them using `File::create`, so conflicting names would be an issue.

crashtracker/src/configuration.rs
danielsn: As noted below, we're creating them with `File::create`, so the same name twice would be an issue.  If we want to allow interleaving, I can change the code to allow that in a subsequent PR

crashtracker/src/configuration.rs
danielsn: 👍

examples/ffi/crashtracking.c
danielsn: done

examples/ffi/crashtracking.c
danielsn: done

profiling-ffi/src/crashtracker/datatypes.rs
sanchda: 👍

PR #390, opened 2024-04-12
Collect the 'logs_created' metric for the sidecar

iamluc:
# What does this PR do?

Count the number of logs created and send the it as a self-telemetry metric.

Part of https://datadoghq.atlassian.net/browse/AIT-4125
Similar to what is done in dd-trace-php (https://github.com/DataDog/dd-trace-php/pull/2604)

# Motivation

# Additional Notes

# How to test the change?

## For Reviewers


sidecar/src/log.rs
```diff
@@ -44,6 +44,7 @@ where
     live_counter: Mutex<HashMap<K, i32>>,
     pending_removal: Mutex<PriorityQueue<K, Instant>>,
     pub expire_after: Duration,
+    logs_created: Mutex<HashMap<Level, u32>>,
```
bwoebi, line 47: Wrong place, the datastructure is e.g. used for both filter and logs.
Instead of stuffing it into TemporarilyRetainedMap, please try to keep the logic contained to MultiEnvFilter.

e.g. replace `pub type MultiEnvFilter = TemporarilyRetainedMap<String, EnvFilter>;` by `struct MultiEnvFilter { pub map: TemporarilyRetainedMap<String, EnvFilter>;  logs_created: Mutex<HashMap<Level, u32>>, }`.

sidecar/src/log.rs
```diff
@@ -136,7 +136,32 @@ where
...
+    pub fn add<'a, 's: 'a>(&'s self, key: String) -> TemporarilyRetainedMapGuard<'a, String, EnvFilter> {
+        self.map.add(key)
+    }
+
+    pub fn collect_logs_created_count<'a, 's: 'a>(&'s self) -> HashMap<Level, u32> {
+        let mut map = self.logs_created.lock().unwrap();
+        let clone = map.clone();
+        map.clear();
+
+        clone
```
bwoebi, line 161: I think `std::mem::take()` would work too, without copying.

sidecar/src/log.rs
```diff
@@ -136,7 +136,32 @@ where
...
+            map: TemporarilyRetainedMap::default(),
+            logs_created: Mutex::new(HashMap::new()),
+        }
+    }
+
+    pub fn add<'a, 's: 'a>(&'s self, key: String) -> TemporarilyRetainedMapGuard<'a, String, EnvFilter> {
+        self.map.add(key)
+    }
+
+    pub fn collect_logs_created_count<'a, 's: 'a>(&'s self) -> HashMap<Level, u32> {
```
bwoebi, line 156: I think the lifetime annotations are unnecessary here? Don't know.

iamluc: Correct

sidecar/src/log.rs
iamluc: Looks OK 👌 

But I had to use `std::mem::take(map.deref_mut())` as `map` is a `MutexGuard`. I hope this is just fine

bwoebi: Yes, that's correct. I just could not tell you the full codeline off my head without autocompletion :-)

bwoebi (approved):
Looks good :-)

PR #391, opened 2024-04-12
Fix leaking metrics in the sidecar

bwoebi:
AppInstance is apparently cloned each time when extracted from the Shared<ManualFuture<AppInstance>>.

Not having the guard telemetry_metrics HashTable in an Arc<> will cause the "already inserted" checks to always fail. And thus leak memory.

Also allow the telemetry worker to have a mode where it's continuing execution after a start-stop cycle, otherwise it won't send any more metrics afterwards.

### DIFF
--- a/ddtelemetry/src/worker/mod.rs
+++ b/ddtelemetry/src/worker/mod.rs
@@ -154,7 +154,10 @@ impl TelemetryWorker {
         };
 
         // if no action is received, then it means the channel is stopped
-        action.unwrap_or(TelemetryActions::Lifecycle(LifecycleAction::Stop))
+        action.unwrap_or_else(|| {
+            self.config.restartable = false;
+            TelemetryActions::Lifecycle(LifecycleAction::Stop)
+        })
     }
 
     async fn run_metrics_logs(mut self) {
@@ -167,7 +170,7 @@ impl TelemetryWorker {
 
             match self.dispatch_metrics_logs_action(action).await {
                 ControlFlow::Continue(()) => {}
-                ControlFlow::Break(()) => break,
+                ControlFlow::Break(()) => if !self.config.restartable { break; },
             };
         }
     }
@@ -234,6 +237,8 @@ impl TelemetryWorker {
                 {
                     self.log_err(&e);
                 }
+                self.data.started = false;
+                self.deadlines.clear_pending();
                 return BREAK;
             }
         };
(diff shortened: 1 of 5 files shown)


sidecar/src/interface.rs
```diff
@@ -368,13 +368,14 @@ impl RuntimeInfo {
 struct AppInstance {
```
bantonsson, line 368: Yeah, `#[derive(Clone)]` can have scary side effects.

ddtelemetry/src/worker/mod.rs
```diff
@@ -167,7 +170,11 @@ impl TelemetryWorker {
 
             match self.dispatch_metrics_logs_action(action).await {
                 ControlFlow::Continue(()) => {}
-                ControlFlow::Break(()) => break,
+                ControlFlow::Break(()) => {
+                    if !self.config.restartable {
```
bantonsson, line 174: It would be nice to have some test that shows that the restartable feature works as expected.

ddtelemetry/src/worker/mod.rs
```diff
@@ -154,7 +154,10 @@ impl TelemetryWorker {
         };
 
         // if no action is received, then it means the channel is stopped
-        action.unwrap_or(TelemetryActions::Lifecycle(LifecycleAction::Stop))
+        action.unwrap_or_else(|| {
+            self.config.restartable = false;
+            TelemetryActions::Lifecycle(LifecycleAction::Stop)
```
bantonsson, line 159: Some documentation about how this `restartable` feature interacts with the rest of the `Lifecycle` of the `TelemetryWorker` would be good. This is not the only place where `LifecycleAction::Stop` is sent, and how do we know that we won't have `TelemetryWorker` instances staying around in some unexpected way.

bantonsson (commented):
Nice find. Some questions about the `restartable` feature.

ddtelemetry/src/worker/mod.rs
bwoebi: It's the only place where it's sent as a _definite_ termination event (i.e. where the handle has been dropped, causing the channel to be closed).

ddtelemetry/src/worker/mod.rs
bwoebi: It would, however there are currently no tests for the overall functionality of the worker lifecycle, which I could extend.
This PR is not the place to add a full-blown testsuite.

ddtelemetry/src/worker/mod.rs
bwoebi: Added comments :-)

ddtelemetry/src/worker/mod.rs
bantonsson: The joys of _shared_ ownership...

bantonsson (approved):

PR #392, opened 2024-04-12
Add stats reporting for sidecar and telemetry

bwoebi:
Allows to collect stats about the fundamental live datastructures of the sidecar. (like amount of apps, metrics stored, etc.)

This should be the goto tool to diagnose high (or runaway) memory usage of the sidecar.


sidecar/src/interface.rs
```diff
@@ -1000,6 +1094,128 @@ impl SidecarServer {
...
+                .values()
+                .map(|s| s.runtimes.lock().unwrap().len() as u32)
+                .sum(),
+            apps: sessions
+                .values()
+                .map(|s| {
+                    s.runtimes
+                        .lock()
+                        .unwrap()
+                        .values()
```
bwoebi, line 1135: can we tell the formatter somehow that lock().unwrap().values() is perfectly readable on one line :-(

ddtelemetry/src/metrics.rs
```diff
@@ -151,6 +170,14 @@ impl<'a> MetricContextGuard<'a> {
     pub fn read(&self, key: ContextKey) -> Option<&MetricContext> {
         self.guard.store.get(key.0 as usize)
     }
+
+    pub fn is_empty(&self) -> bool {
+        self.guard.store.is_empty()
+    }
```
bwoebi, line 176: Unused, but clippy wanted it.

sidecar/src/interface.rs
bantonsson: As far as I know, it doesn't can't distinguish between different methods, so I guess no.

bantonsson: I think that the repeated use of `lock().unwrap` should be hidden behind helper functions on this form to lessen the noise.
```
    fn runtimes(&self) -> MutexGuard<HashMap<String, RuntimeInfo>> {
        self.runtimes.lock().unwrap()
    }
```

Can be done for `sessions`, `runtimes`, `apps,` `app_or_actions` to make things at least a little bit nicer to read.

bantonsson (commented):
The `interface.rs` file is just way too big. It needs to be split up (and commented).

sidecar/src/interface.rs
bwoebi: Good idea, though I'd call it lock_runtimes() to be explicit about the locking.

bwoebi:
Agree, interface.rs has grown too big. I'll make sure to split it into multiple modules in a future PR.

bantonsson (approved):
Looking forward to the refactor 😉

PR #393, opened 2024-04-16
Make verify timeout test in datadog-trace-mini-agent trigger timeout

bantonsson:
# What does this PR do?

Fixes a test in the `datadog-trace-mini-agent` that failed spuriously.

# Motivation

Flaky tests are bad™

# Additional Notes

Only setting the timeout to 0 ms is not enough, and there was a race in the test between the dns lookup and the timeout that intermittently failed the test (most likely on Windows).

# How to test the change?

## For Reviewers


ekump (approved):
Not that familiar with the mini-agent, but LGTM FWIW

PR #395, opened 2024-04-17
refactor sidecar/interface.rs into smaller files

ekump:
# What does this PR do?

- Break up the various types and implementations in [sidecar/interface.rs](https://github.com/DataDog/libdatadog/blob/main/sidecar/src/interface.rs) into separate files within a sidecar::service module. (**I'm very open to a better name than "service"**).
- Introduce _some_ tests _and_ some documentation. I only added some tests that didn't require much mocking or setup. APMSP-1079 has been created to track a more comprehensive unit testing solution.
- Fixes some minor bugs uncovered through addition of tests.
- Miscellaneous small fixes for typos, cleanup, and Rust idiomaticness.  
- As I am fairly new to the sidecar I made my best guesses at rustdoc comments, tests, and naming. Please call out any errors.

# Motivation

Prep work for introducing retry logic sending traces via the sidecar

# Additional Notes

This is a a fairly disruptive refactor. If there are non-critical things we can address in follow up PRs I would prefer to do that versus keeping this open for a long time. Obviously, any critical issues should be addressed ASAP. 
# How to test the change?

Unit tests are being added, but some sort of integration test with PHP is probably a good idea.

## For Reviewers


sidecar/src/service/session_info.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// ```
+    /// use datadog_sidecar::service::SessionInfo;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let session_info = SessionInfo::default();
+    ///     session_info.shutdown_running_instances().await;
+    /// }
+    /// ```
+    pub async fn shutdown_running_instances(&self) {
```
ekump, line 116: Calling this function out for review as it has not only moved, but changed. The previous implementation was cloning the runtimes and never actually shutting them down. This implementation drains the runtimes hashmap and then shuts them down. 

I doubt this is the optimal implementation, but it at least works as advertised now.

sidecar/src/service/session_info.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// ```
+    /// use datadog_sidecar::service::SessionInfo;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let session_info = SessionInfo::default();
+    ///     session_info.shutdown_runtime(&"runtime1".to_string()).await;
+    /// }
+    /// ```
+    pub async fn shutdown_runtime(&self, runtime_id: &String) {
```
ekump, line 148: Calling this function out as there is a minor change (in addition to being moved). This was the only public function that took `self` instead of `&self`. This meant that after shutting down a single runtime the `session_info` object was moved and you couldn't do anything else with it. I'm not that familiar with how `SessionInfo` should be used, but since it supports multiple runtimes it should be usable after a single one is shutdown.

sidecar/src/service/sidecar_server.rs
```diff
@@ -0,0 +1,723 @@
+// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+// TODO: APM-1076 - This file contains a fair amount of expects. While in most cases it is unlikely
+// we will ever hit these, we should consider adding more robust error handling.
```
bwoebi, line 5: I think in these cases the error handling we have is quite fine. In fact, the error handling is tokio catching panics in async tasks and transforming the tasks into an error Result.
The important part is that these panics on expect() don't bring the application into a deadlocked state where e.g. no further traces are processed at all.

On top of that a crash of the whole sidecar, while a minor data loss, is not catastrophic either as it will be restarted by the PHP processes quickly.

sidecar/src/service/session_info.rs
```diff
@@ -0,0 +1,254 @@
...
+
+use futures::future;
+
+use tracing::{enabled, info, Level};
+
+use crate::log::{MultiEnvFilterGuard, MultiWriterGuard};
+use crate::tracer;
+
+use crate::service::InstanceId;
+/// `SessionInfo` holds information about a session.
```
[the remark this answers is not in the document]
bwoebi: The very dumb answer is: a session is identified by a session id, and a runtime by a runtime id :-P
Now, what _exactly_ these are depends a bit on the language, but session ids are supposed to be shared for the main process and all it's forks while runtimes are mostly per process.
In context of PHP it's: one session+runtime id tuple = one sidecar connection.

sidecar/src/service/sidecar_server.rs
```diff
@@ -47,6 +47,18 @@ fn no_response() -> NoResponse {
     future::ready(())
 }
 
+/// The `SidecarServer` struct represents a server that handles sidecar operations.
+///
+/// It maintains a list of active sessions and a counter for each session.
+/// It also holds a reference to a `TraceFlusher` for sending trace data,
+/// and a `Mutex` guarding an optional `ManualFutureCompleter` for telemetry configuration.
+///
+/// # Fields
```
bwoebi, line 56: I think it's more common (and I prefer it too) to put docs about fields inline in the struct, i.e.
```
    /// Used for sending trace data
   pub trace_flusher: Arc<TraceFlusher>
```

sidecar/src/service/tracing/trace_flusher.rs
```diff
@@ -0,0 +1,240 @@
...
+use std::time::{Duration, Instant};
+use tokio::select;
+use tokio::task::{JoinError, JoinHandle};
+use tracing::{debug, error, info};
+
+/// `TraceFlusherStats` holds stats of the trace flusher like the count of allocated shared memory
+/// for agent config, agent config writers, last used entries in agent configs, and the size of send
+/// data.
+#[derive(Serialize, Deserialize)]
+pub struct TraceFlusherStats {
```
ekump, line 28: Not sure if "stats" belong in their own namespace, or under telemetry, or close to the functionality they are "stat'ing" so put them near the functionality for now.

bwoebi: I think they should belong near the functionality as they are explicitly meant to describe the state of the functionality, primarily as a debugging mechanism.

ekump: @bwoebi That's a fair point. Near the functionality works for me.

sidecar/src/service/runtime_info.rs
```diff
@@ -0,0 +1,139 @@
...
+        #[cfg(feature = "tracing")]
+        debug!(
+            "Successfully shut down runtime_id {} for session {}",
+            self.instance_id.runtime_id, self.instance_id.session_id
+        );
+    }
+
+    // TODO-EK: Reduce the type complexity before merging
+    #[allow(clippy::type_complexity)]
+    // TODO-EK - Investigate why this needs to be exposed this way before merging
```
bwoebi, line 115: The goal is simply simplifying the code a bit whether every access is multiple lines, see https://github.com/DataDog/libdatadog/pull/392#discussion_r1565757801 for context.

sidecar/src/service/runtime_info.rs
```diff
@@ -0,0 +1,139 @@
+// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+// TODO-EK: Re-evaluate access levels for the structs and methods in this file after refactor,
+// before merging
+// TODO-EK: Better understand why apps and actions are conflated in this struct before merging. And,
+// if they should be separated.
```
bwoebi, line 7: Because service and environment names are only known late. Hence we first queue actions - once the service and env names are finalized, we create the app and flush everything from queue to app.

sidecar/src/service/runtime_info.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(Clone, Default)]
+pub(crate) struct RuntimeInfo {
+    pub(crate) apps: AppMap,
+    app_or_actions: Arc<Mutex<HashMap<QueueId, AppOrQueue>>>,
+    #[cfg(feature = "tracing")]
+    pub(crate) instance_id: InstanceId,
+}
+
+// TODO-EK: Can this be refactored more before merging? We may be able to encapsulate some of the
+// functionality in session_info to RuntimeInfo.
```
bwoebi, line 43: Like for example?

sidecar/src/service/runtime_info.rs
bwoebi: Reducing the type complexity here sort of defeats the point of just being a simple helper function.

sidecar/src/service/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+    RegisterTelemetryMetric(MetricContext),
+    AddTelemetryMetricPoint((String, f64, Vec<Tag>)),
+    PhpComposerTelemetryFile(PathBuf),
+}
+
+#[allow(clippy::large_enum_variant)]
+pub(crate) enum AppOrQueue {
+    App(Shared<ManualFuture<(String, String)>>),
+    Queue(EnqueuedTelemetryData),
+}
```
bwoebi, line 66: Fits better into telemetry/mod.rs

sidecar/src/service/telemetry/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub(crate) apps: u32,
+    pub(crate) active_apps: u32,
+    pub(crate) enqueued_apps: u32,
+    pub(crate) enqueued_telemetry_data: EnqueuedTelemetryStats,
+    pub(crate) telemetry_metrics_contexts: u32,
+    pub(crate) telemetry_worker: TelemetryWorkerStats,
+    pub(crate) telemetry_worker_errors: u32,
+    pub(crate) log_writer: TemporarilyRetainedMapStats,
+    pub(crate) log_filter: TemporarilyRetainedMapStats,
+}
```
bwoebi, line 30: Should be located in sidecar_server.rs, like other Stats are in their related module.

sidecar/src/service/tracing/mod.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+pub(crate) use trace_flusher::TraceFlusher;
+
+use trace_send_data::TraceSendData;
```
bwoebi, line 6: What's the point of this use?

sidecar/src/service/serialized_tracer_header_tags.rs
```diff
@@ -0,0 +1,163 @@
...
+///     client_computed_stats: false,
+/// };
+///
+/// let serialized: Result<SerializedTracerHeaderTags, _> = tracer_header_tags.try_into();
+/// assert!(serialized.is_ok());
+/// ```
+// DEV: This implementation is not tested for the case where bincode raises an error because there
+// is no reasonable way to force bincode to fail during serialization without modifying the
+// `TracerHeaderTags` struct.
+impl<'a> TryFrom<TracerHeaderTags<'a>> for SerializedTracerHeaderTags {
```
bwoebi, line 85: bincode is infallible for Strings. That's why the code did not use TryFrom, but just From. This was a deliberate choice, not unsafety.

Just as you note, there's no reasonable way for it to fail.

bwoebi:
I’m a bit annoyed at big docblocks explaining obvious things. Like yes, _obviously_ the fn add() of the Add trait impl on EnqueuedTelemetryStats will add two EnqueuedTelemetryStats together and return a new EnququedTelemetryStats instance which is the result of rhs and the instance. And this interrupts my reading flow with 10 lines of fat docblock.
There are a couple instances of such obvious docblocks, would you mind removing these? Thanks!

bwoebi:
Is there a reason why you’ve aggressively put `pub(crate)` on everything instead of simply `pub` within the individual rust files?
Looks easier to read if it's just pub.

And I assume it’s enough to just make the modules then pub(crate) in the `mod.rs` files? What you’re anyway already doing.
Feels cleaner, and gives the ability to just toggle the visibility of a whole module from pub(crate) to pub when needed instead of touching every single codeline in that mod.

sidecar/src/service/serialized_tracer_header_tags.rs
bwoebi: For deserialization it's probably okay to check.

sidecar/src/service/telemetry/enqueued_telemetry_stats.rs
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `rhs` - An `EnqueuedTelemetryStats` instance that needs to be added to the current
+    ///   instance.
+    ///
+    /// # Returns
+    ///
+    /// * A new `EnqueuedTelemetryStats` instance that is the result of the addition.
+    fn add(self, rhs: Self) -> Self::Output {
```
bantonsson, line 38: Implementations of a _trivial_ trait doesn't need the doc.

bantonsson (approved):
Looks good. Only some minor things.

* Inline doc of members in struct/enum instead of before.
* Remove doc from _"trivial"_ trait implementations like `Add` and `Sum`.

sidecar/src/service/sidecar_server.rs
ekump: @bwoebi 
> I'm not a big proponent of having a proper error handling code for things which are not supposed to be possible

I tend to agree as long as 1) we are confident that these errors should never happen and 2) If 1) is wrong, we don't put the application in a bad state. Your comment indicates that we satisfy both these points so I'm ok with leaving these expects as is. I can remove the TODO.

sidecar/src/service/sidecar_server.rs
ekump: No objection from me, I'll update them to be inline!

sidecar/src/service/runtime_info.rs
ekump: @bwoebi your comments are from an outdated version of this file, so I'm not sure if still applies. 

The complexity isn't due to the `MutexGuard`, it's due to the app HashMap type of `HashMap<(String, String), Shared<ManualFuture<Option<AppInstance>>>>`. I created the type `AppMap` and use that throughout `RuntimeInfo` and was able to remove the clippy ignore. 

I think we should strive to ignore linter errors only when it is absolutely necessary. I can't speak for @bantonsson on https://github.com/DataDog/libdatadog/pull/392#discussion_r1565757801, but I would argue that a helper function isn't all that helpful if you have to suppress valid linter errors. The current implementation resolves the clippy error and maintains the helpfulness Bjorn was suggesting...at least in my opinion. 

I added a TODO (with a jira ticket) to just investigate if it makes sense / is possible to not mutably expose the hashmap at a later time. In the past, I've found that this pattern can lead to problems but I understand there may be no reasonable way around it.

sidecar/src/service/runtime_info.rs
ekump: Thanks, that's good to know. I'll update the comment to reflect this.

sidecar/src/service/runtime_info.rs
ekump: This TODO was just a reminder for myself while the PR was in draft. At the moment, I don't think this is possible in the short-term and isn't really blocking progress. I created a jira ticket and updated the TODO comment to reflect it.

sidecar/src/service/runtime_info.rs
bwoebi: To be honest, if it were me, I would just disable the type complexity clippy error for the whole repository.
Sometimes types are simply complex and using a `type` is basically just a workaround and requiring you to search for the `type` definition to just see the actual type.

sidecar/src/service/session_info.rs
bwoebi: It didn't matter, because it implements Clone and all the relevant data within is Arc. But yes, using `&self` is correct.

sidecar/src/service/tracing/mod.rs
ekump: it elevates `TraceSendData` to the top level of the module so you can import it as `tracing::TraceSendData` instead of `tracing::trace_send_data::TraceSendData`. 

There are situations where you want the full namespace path, but I think `TraceSendData` is an example where there is no benefit.

bwoebi: I was mostly referring to the fact that TraceSendData has only a single use, namely in the neighboring trace_flusher module; I'd agree with you if it had more uses. Anyway, this one is not important.

ekump:
> Is there a reason why you’ve aggressively put pub(crate) on everything instead of simply pub within the individual rust files?
Looks easier to read if it's just pub.

> And I assume it’s enough to just make the modules then pub(crate) in the mod.rs files? What you’re anyway already doing.
Feels cleaner, and gives the ability to just toggle the visibility of a whole module from pub(crate) to pub when needed instead of touching every single codeline in that mod.

@bwoebi I did this for two reasons:

1) Consistency and safety (not rust "safety"). There are situations where we don't want to mark an entire file as `pub(crate)` but we do mix exposing certain functions, types, or struct members as either `pub(crate)` or just `pub`. In those cases we have no choice but to be verbose in the file. It seems awkward to me that sometimes we set `pub(crate)` on just the `mod.rs` level and sometimes we do it at the implementation. I also don't find it hard to read and I like knowing the access level with the implementation details. But if the consensus is that this is superfluous I can revert this in another PR. 

2) It's related to reason 1, but I generally believe that we need to lock down the public surface area of the library as much as possible. We should only expose things publicly that we actually want public, and we're willing to support long-term and properly document. At the moment, I don't think we have any tools in place to help us not "leak" things we don't want public. Until we have something to do this for us (via static analysis or testing or something else) I'd prefer to err on the side of caution.

Your assumption is correct. Functionally, we don't need to be this verbose. If there is a general consensus against this, I can revert in a follow-up PR.

PR #396, opened 2024-04-17
Move MaybeError from ddtelemetry-ffi to ddcommon-ffi

bantonsson:
# What does this PR do?

Moves the `MaybeError` type from the `ddtelemetry-ffi` crate to the `ddcommon-ffi` crate, and changes it to use the `ddcommon-ffi::Error` type.

# Motivation

It's a nice convenience type that we want to usein the `data-pipeline-ffi`, and I don't understand why it didn't use the `ddcommon-ffi::Error` type.

# Additional Notes

# How to test the change?

## For Reviewers

### DIFF
--- a/ddcommon-ffi/src/error.rs
+++ b/ddcommon-ffi/src/error.rs
@@ -3,13 +3,13 @@
 
 use crate::slice::CharSlice;
 use crate::vec::Vec;
-use std::fmt::{Display, Formatter};
+use std::fmt::{Debug, Display, Formatter};
 
 /// Please treat this as opaque; do not reach into it, and especially don't
 /// write into it! The most relevant APIs are:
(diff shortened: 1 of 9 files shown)


ddcommon-ffi/src/error.rs
```diff
@@ -9,7 +9,7 @@ use std::fmt::{Display, Formatter};
 /// write into it! The most relevant APIs are:
 /// * `ddog_Error_message`, to get the message as a slice.
 /// * `ddog_Error_drop`.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
```
danielsn, line 12: When do we want to compare errors using eq?

bwoebi (approved):
Yup, no reason to have two separate impls here.

PR #397, opened 2024-04-19
Fix use-after-free in non-fdexec spawn_worker config

bwoebi:
Assigning `argv.items[]` directly does not update `argv.ptrs[]`, causing an use-after-free.

### DIFF
--- a/spawn_worker/src/unix/spawn.rs
+++ b/spawn_worker/src/unix/spawn.rs
@@ -71,6 +71,11 @@ impl ExecVec {
         self.ptrs.push(ptr::null());
         self.items.push(item);
     }
+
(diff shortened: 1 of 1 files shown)


paullegranddc:
Could you put `ExecVec` in it's own module, and set it's fields to private to prevent any further modifications of it's fields without respecting it's invariants?

paullegranddc (approved):

bantonsson (approved):

bwoebi:
@paullegranddc good idea, sadly rust has no `pub(impl)` restricting the visibility to implementations of the specific struct.

bwoebi:
I mean, what you proposed is fine. I just meant, there's no means in rust to restrict access to fields within a same module, to e.g. only direct impls of the struct.

PR #399, opened 2024-04-22
Provide a DogStatsD API in the sidecar

iamluc:
# What does this PR do?

Add a DogStatsD client in the sidecar, and expose an API to send metrics through it.

# Motivation

Is part of the PHP tasks for Q2. It adds new possibilities for the integration team.

# Additional Notes

The underlying StatsD client used by this PR is [Cadence](https://docs.rs/cadence/1.0.0/cadence/index.html).
It is already used by some teams at Datadog, even if does not complies exactly to other DogStatsD implementations like the one in [PHP](https://github.com/DataDog/php-datadogstatsd) or [Go](https://github.com/DataDog/datadog-go)

For example, it lacks support of the "sample rate", or requires an integer for the `Set` metric, while Go and PHP can accept a string.

But I think there is nothing critical that blocks us.

# How to test the change?

## For Reviewers


sidecar/src/dogstatsd.rs
```diff
@@ -0,0 +1,152 @@
...
+where
+    T: Metric + From<String>,
+{
+    for tag in tags {
+        builder = builder.with_tag_value(tag.as_ref());
+    }
+    match builder.try_send() {
+        Ok(_) => Ok(()),
+        Err(err) => Err(anyhow!(err)),
+    }
```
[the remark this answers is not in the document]
bwoebi: I prefer the latter, but you're right.

ddcommon-ffi/src/endpoint.rs
```diff
@@ -49,5 +49,11 @@ pub extern "C" fn ddog_endpoint_from_api_key_and_site(
     None
 }
 
+#[no_mangle]
+#[must_use]
+pub extern "C" fn ddog_endpoint_empty() -> Box<Endpoint> {
+    Box::new(Endpoint::default())
+}
+
```
bwoebi, line 57: I would not provide that, but just do nothing if an API key is provided in the Endpoint instead of handling an empty endpoint.

bwoebi (approved):
Looks good to me :-)

sidecar/src/service/session_info.rs
```diff
@@ -143,6 +144,17 @@ impl SessionInfo {
     {
         f(&mut self.get_trace_config());
     }
+
+    pub(crate) fn get_dogstatsd(&self) -> MutexGuard<dogstatsd::Flusher> {
+        self.dogstatsd.lock().unwrap()
```
ekump, line 149: I think we should use `expect()` instead of `unwrap()`, however unlikely it is that we'll panic (this applies to the entire PR, not just this one line).

sidecar/src/dogstatsd.rs
```diff
@@ -0,0 +1,268 @@
...
+                QUEUE_SIZE,
+            );
+
+            Ok(StatsdClient::from_sink("", sink))
+        }
+    };
+}
+
+#[cfg(test)]
+mod test {
```
ekump, line 162: 👍 love to see tests being added

ekump (approved):
Warning - I know nothing about dogstatsd

Some non-blocking questions and style comments, but LGTM

sidecar/src/dogstatsd.rs
```diff
@@ -0,0 +1,268 @@
...
+        match endpoint.api_key {
+            Some(_) => {
+                info!("DogStatsD is not available in agentless mode");
+                self.endpoint = None;
+            }
+            None => {
+                debug!("Updating DogStatsD endpoint to {}", endpoint.url.clone());
+                self.endpoint = Some(endpoint);
+            }
+        }
```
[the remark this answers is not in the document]
bwoebi: Is the clone() call is the debug!() macro not redundant?

sidecar-ffi/src/lib.rs
```diff
@@ -523,3 +527,118 @@ pub unsafe extern "C" fn ddog_sidecar_stats(
...
 }
+
+#[no_mangle]
+#[allow(clippy::missing_safety_doc)]
+pub unsafe extern "C" fn ddog_sidecar_dogstatsd_count(
+    transport: &mut Box<SidecarTransport>,
+    instance_id: &InstanceId,
+    metric: ffi::CharSlice,
+    value: u64,
+    tags: Option<&ddcommon_ffi::Vec<Tag>>,
```
pierotibou, line 538: So we passa all the tags all the time? Some of them (e.g lang, lang_interpreter, lang_version, tracer_version, runtimeId won't change from one session to another. 
Also, not saying we should do it, but we do process at least some tags in .NET (cf [this](https://github.com/DataDog/dd-trace-dotnet/blob/bcc8488d70163ab8ec0bfb95163a58545874915b/tracer/src/Datadog.Trace/TracerManagerFactory.cs#L369)), do you do it on php side?

pierotibou (commented):
I had started the review yesterday so pushing those 2 questions anyway

sidecar-ffi/src/lib.rs
bwoebi: Yes, we do that on the PHP side. The language/version/runtime id tags could be probably directly provided within the sidecar. PR #398 anyway transfers that language/version info as part of session info. We could just use that part and merge it here.
Would make sense to me, great point, Pierre!

pierotibou: If processing happens on the client side we should document it Somewhere in the API.

sidecar/src/dogstatsd.rs
```diff
@@ -0,0 +1,268 @@
...
+// Queue with a maximum capacity of 32K elements
+const QUEUE_SIZE: usize = 32 * 1024;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum DogStatsDAction {
+    Count(String, u64, Vec<Tag>),
+    Distribution(String, f64, Vec<Tag>),
+    Gauge(String, f64, Vec<Tag>),
+    Histogram(String, f64, Vec<Tag>),
+    // Cadence only support i64 type as value
```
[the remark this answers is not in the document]
iamluc: Good catch, It' a mistake!
 I also fixed the `Count` metric type, both should accept `i64`

sidecar/src/dogstatsd.rs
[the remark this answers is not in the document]
iamluc: Addressed both comments

sidecar/src/service/session_info.rs
iamluc: If you don't mind, I'd prefer to do it in a separate PR.
Because for example, `get_dogstatsd` is similar to many other methods of the `SessionInfo` struct, which all use `unwrap`

Cargo.lock
```diff
@@ -334,6 +334,15 @@ dependencies = [
  "serde",
 ]
 
+[[package]]
+name = "cadence"
```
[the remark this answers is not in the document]
iamluc: You mean on `libdatadog_profiling.{a,so}`? The size is exactly the same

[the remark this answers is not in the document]
pierotibou: It won't change the profiling lib. It will change the final package tracer size I assume as the sidecar is used by tracing only for now. Basically, our release artifacts are 500MB already, the profiling team starts looking into reducing package size, we should globally be very careful of the size impact of adding any new dependency.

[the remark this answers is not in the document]
pierotibou: We support 12 versions IIRC. So I assume we have an extra 400KB per version here. Not saying it comes from cadence only though. Anyway, this is something that we will need to be more and more careful about I believe even for the sidecar (not a blocker here I suppose).

sidecar/src/service/session_info.rs
ekump: @iamluc - A separate PR is ok.

PR #401, opened 2024-04-25
[crashtracker] Simplify the stack collection options

danielsn:
# What does this PR do?

Combines the "should we collect stack traces" and "should we symbolize them" into one enum

# Motivation

The current setup confused @ivoanjo. Hopefully this is clearer.
# Additional Notes

Naming things is hard. Feel free to suggest better names.

# How to test the change?

## For Reviewers

### DIFF
--- a/crashtracker/src/configuration.rs
+++ b/crashtracker/src/configuration.rs
@@ -4,19 +4,25 @@ use ddcommon::Endpoint;
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 
+/// Stacktrace collection occurs in the context of a crashing process.
+/// If the stack is sufficiently corruputed, it is possible (but unlikely),
+/// for stack trace collection itself to crash.
+/// We recommend fully enabling stacktrace collection, but having an environment
+/// variable to allow downgrading the collector.
 #[repr(C)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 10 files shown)


crashtracker/src/configuration.rs
```diff
@@ -6,19 +6,17 @@ use std::time::Duration;
...
-pub enum CrashtrackerResolveFrames {
-    Never,
-    /// Resolving frames in process is experimental, and can fail/crash
-    ExperimentalInProcess,
-    InReceiver,
+pub enum CrashtrackerStacktraceCollectionOptions {
+    DontCollectStacktrace,
+    CollectStacktraceButDoNotResolveSymbols,
+    CollectStacktraceAndResolveSymbolsInReceiver,
 }
```
ivoanjo, line 13: I think these options / new API looks good!

Small suggestion in terms of naming (feel free to ignore, naming is hard :P)

```suggestion
pub enum StacktraceCollection {
    Disabled,
    WithoutSymbols,
    Enabled,
}
```

Also, for API users, it may be relevant to document when to use Disabled vs WithoutSymbols vs Enabled. E.g. what are the reasons to not always go with Enabled and forget about the others?

ivoanjo (approved):
This looks great, thanks for looking into it! :)

crashtracker/src/configuration.rs
danielsn: Good idea.  Done.

PR #403, opened 2024-04-25
refactor(profiling): backup sample types/period 

morrisonlevi:
# What does this PR do?

This refactors the Profile to backup the sample types and period outside of the string table.

Adds `api::ValueType::new`, and uses that instead of direct instantiation. In most cases, this shortens from 4 lines to 1 (but not always).

Simplifies some `vec![...]` usage to `[...]` in tests, since the vector isn't needed at all.

# Motivation

Restoring the sample types and period was the only place that used `fn get_string(&self, id: StringId) -> &str` outside of tests. Requiring the data structure to map in both directions to provide this functionality makes it hard for innovation. The upcoming `StringTable` PR will not provide the `StringId` to `&str` functionality.

# Additional Notes

This is stacked on top of another PR, and there will be another PR stacked on this. They should all be merged at roughly the same time.

# How to test the change?

This should be an internal change only, so test as usual.


examples/ffi/exporter.cpp
```diff
@@ -89,7 +89,7 @@ int main(int argc, char *argv[]) {
   }
 
   ddog_prof_Profile_SerializeResult serialize_result =
-      ddog_prof_Profile_serialize(profile.get(), nullptr, nullptr, nullptr);
+      ddog_prof_Profile_serialize(profile.get(), nullptr, nullptr, sample_types, &period, nullptr);
   if (serialize_result.tag == DDOG_PROF_PROFILE_SERIALIZE_RESULT_ERR) {
```
ivoanjo, line 93: I guess we can live with this, although it makes the API a bit... weird? Since the sample types usually never change so forcing API clients to keep providing them is somewhat awkward.

Since the strings provided by the profile types are up to a dozen small strings, one way of avoiding the additional complexity on the string table + on the API would be something like storing the types both in the string table as well as regular Rust strings somewhere in the profile.

Then, on reset, we would just re-feed the same strings into the string table. Something like this maybe?

morrisonlevi: Yes, but it does not fit into the long-term goal of avoiding allocations on the system allocator. These strings are not a big deal, but the user almost certainly has them in static memory, right?

Maybe it's worth trusting the user here instead. Meaning, we _trust_ they are static strings and document it as such?

ivoanjo: > Yes, but it does not fit into the long-term goal of avoiding allocations on the system allocator.

True but... I think we have a bunch of other things to fix, and perhaps the fix for those things would also work for the strings?

> These strings are not a big deal, but the user almost certainly has them in static memory, right?
>
> Maybe it's worth trusting the user here instead. Meaning, we trust they are static strings and document it as such?

It's a good point. On the other hand, it seems like a bit of a potential footgun for a small-ish gain, so between footgun and awkward API (e.g having to specify the types on a bunch of API calls), I'd say let's go with the awkward API and no footgun.

None of this is set in stone anyway, since we've been successfully iterating on these APIs all the time, so I guess we can live with the awkward API if needed :P

danielsn: I don't have a strong preference here, but keeping the (small) set of strings in a vec seems like a reasonable trade-off

morrisonlevi: I rewrote the code to backup the sample types and period using `String`s and such. Let me know what you think of the revised PR. This version does not require a BC break.

ivoanjo: Nice, thanks! I'm falling behind on a few PRs, will take a look at it soon-ish if nobody beats me to it :sweat_smile:

profiling/src/api.rs
```diff
@@ -10,6 +10,11 @@ pub struct ValueType<'a> {
     pub r#type: &'a str,
     pub unit: &'a str,
 }
+impl<'a> ValueType<'a> {
+    pub fn new(r#type: &'a str, unit: &'a str) -> Self {
+        Self { r#type, unit }
+    }
+}
```
ivoanjo, line 17: Really minor: Since this is a trivial constructor, is it useful to keep? We don't seem to have constructors for most of the other structs below, and even in this PR there's a mix of constructing the struct via `new` and directly.

profiling/src/internal/profile.rs
```diff
@@ -12,6 +12,7 @@ use std::collections::HashMap;
 use std::time::{Duration, SystemTime};
 
 pub struct Profile {
+    backup_sample_types: BackupSampleTypesAndPeriod,
```
ivoanjo, line 15: Minor: This seems "weird enough" to deserve a bit of a hint ;)

```suggestion
    // used when resetting the profile, see struct definition for details
    backup_sample_types: BackupSampleTypesAndPeriod,
```

profiling/src/internal/profile.rs
```diff
@@ -846,36 +819,37 @@ mod api_test {
...
         // table offsets may not match).
-        let (value, period_type) = profile.period.expect("profile to have a period");
-        assert_eq!(value, period.0);
-        assert_eq!(profile.get_string(period_type.r#type), "wall-time");
-        assert_eq!(profile.get_string(period_type.unit), "nanoseconds");
+        for (value, period_type) in [profile.period.unwrap(), prev.period.unwrap()] {
+            assert_eq!(value, period.value);
+            let strings = profile.strings.iter().collect::<Vec<_>>();
+            let r#type: &str = strings[period_type.r#type.to_offset()];
+            let unit: &str = strings[period_type.unit.to_offset()];
```
ivoanjo, line 849: Minor: It may be useful to introduce a test helper for doing lookups in the strings, as it looks like something we may need to use for more tests.

profiling/src/internal/backup_sample_types.rs
```diff
@@ -0,0 +1,115 @@
...
+/// we can just re-use them as needed. It does a single allocation for
+/// everything from the global allocator.
+pub struct BackupSampleTypesAndPeriod {
+    /// The static bounds are lies, they are tied to the `storage`. Only pass
+    /// out references with the same lifetime as &self.
+    sample_types: ManuallyDrop<Box<[ValueType<'static>]>>,
+    /// Despite our manual drop order, it is still recommended to order the
+    /// fields in the order things should be dropped.
+    storage: ManuallyDrop<Vec<(String, String)>>,
+    period: Option<((String, String), i64)>,
```
ivoanjo, line 19: Minor: Out of curiosity, it seems that often we either omit or pass in a dummy period. I  wonder if we should just get rid of this parameter for libdatadog API users (and libdatadog would always just insert a default one).

profiling/src/internal/backup_sample_types.rs
```diff
@@ -0,0 +1,115 @@
...
+            // SAFETY: transforming the lifetime here is _NOT_ safe! We are
+            // appeasing the borrow-checker. Only expose it to the outside
+            // with a lifetime bound to &self!
+            let value_type = unsafe {
+                ValueType {
+                    r#type: core::mem::transmute(sample_type.0.as_str()),
+                    unit: core::mem::transmute(sample_type.1.as_str()),
+                }
+            };
+            new_sample_types.push(value_type);
```
ivoanjo, line 46: So, I'm not quite sure what's the advantage here between having the separate storage and sample_types. Specifically, the Rust docs say that a `String` always allocates its data on the heap, so assuming that's correct, I don't get what we're gaining from not having the sample types own the strings directly?

(This is probably me missing some context on some Rust limitation, as usual)

ivoanjo (commented):
This looks quite reasonable! The only part I'm not very confident about is the magic around internal memory management in `BackupSampleTypesAndPeriod`, I left a few questions there.

profiling/src/api.rs
morrisonlevi: I think so, because it fits on 1 line in many cases, but directly constructing it will take 4 lines. That's honestly why I made it. Thoughts?

profiling/src/internal/backup_sample_types.rs
morrisonlevi: No, that makes sense. Originally I had a smarter implementation which reduced allocations but miri didn't like it. I usually can figure out miri errors or at least understand them, and on that one I didn't. So I did a simpler thing.

But taking a step back, no, you are right. If I can't reduce this to like 1 allocation for the whole thing, may as well just make simple structs that own `String`s instead.

profiling/src/api.rs
ivoanjo: Sounds reasonable. In that case, maybe worth a quick pass on the PR to make sure we're using it everywhere we can perhaps?

profiling/src/internal/profile.rs
morrisonlevi: Removed :D

danielsn (approved):

profiling/src/internal/backup_sample_types.rs
ivoanjo: Yeap, for sure, more of a passing observation.

PR #404, opened 2024-04-25
perf(profiling): datadog-alloc based StringTable

morrisonlevi:
# What does this PR do?

Creates a `StringTable` where the individual strings are allocated in
an arena allocator. This uses the Chain, Linear, and Virtual Allocators
provided by datadog-alloc to create a chain of 4 MiB chunks, each of
which is virtually allocated.

Also changes the chain allocator to handle large allocations, in case
we encounter a large string. Previously it would error if the
allocation size was larger than the node size.

# Motivation

This separates the StringTable from the system allocator. Since strings
are variable length, separating them from the system allocator has more
benefits than many other items.

This also is a bit faster. This is probably due to the simpler
allocation strategy and better data locality. See this for details:
https://github.com/DataDog/libdatadog/pull/404#discussion_r1600836784

# Additional Notes

This has been a long-time coming. I'm excited for it to finally merge!

# How to test the change?

As long as you weren't using string table implementation details,
nothing should change. The apis for FFI users is unchanged, for
instance.


profiling/benches/interning_strings.rs
```diff
@@ -0,0 +1,28 @@
...
+            assert_eq!(n_strings, table.len());
+
+            // re-insert, should nothing should be inserted.
+            for string in WORDPRESS_STRINGS {
+                black_box(table.intern(string));
+            }
+            assert_eq!(n_strings, table.len())
+        })
+    });
+}
```
ivoanjo, line 26: I'm curious: did you get a chance to compare this with the previous implementation?

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl StringTable {
+    /// Creates a new string table, which initially holds the empty string and
+    /// no others.
+    pub fn new() -> Self {
+        // Christophe and Grégory think this is a fine size for 32-bit .NET.
```
ivoanjo, line 38: Minor: This is the kind of comment that tends to age poorly, e.g. maybe in some time we won't remember what this meant; may be worth expanding a bit ;)

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+        const SIZE_HINT: usize = 4 * 1024 * 1024;
+        let bytes = ChainAllocator::new_in(SIZE_HINT, VirtualAllocator {});
+
+        let mut strings = HashSet::with_hasher(Hasher::default());
+        // It various by implementation, but frequently I've noticed that the
```
ivoanjo, line 43: Minor:

```suggestion
        // It varies by implementation, but frequently I've noticed that the
```

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+        // - At least one sample type
+        // - At least one sample unit--already at 3 without any samples.
+        // - "local root span id"
+        // - "span id"
+        // - "trace endpoint"
+        // - A file and/or function name per frame.
+        // So with a capacity like 3, we end up reallocating a bunch on or
+        // before the very first sample. The number here is not fine-tuned,
+        // just skipping some obviously bad, tiny sizes.
+        strings.reserve(32);
```
ivoanjo, line 57: To be honest, I'd go even bigger -- 512 or maybe more? E.g. for one minute of data of a non-idle realistic application this seems like we'll hit it easily?

(I guess we could even try measuring this in staging or something)

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+                };
+
+                // Copy the bytes of the string into the allocated memory.
+                // SAFETY: this is guaranteed to not be overlapping because an
+                // allocator must not return aliasing bytes in its allocations.
+                unsafe {
+                    let src = str.as_ptr();
+                    let dst = uninit_ptr.as_ptr() as *mut u8;
+                    let count = str.len();
+                    core::ptr::copy_nonoverlapping(src, dst, count)
```
ivoanjo, line 109: Minor: More of a structure kinda of thing, I'd kinda expect these two things (especially the unsafe bits) to be internal to `ChainAllocator`, rather than having the users need to get their hands dirty.

ivoanjo (commented):
Oh wow, I have to say, after the PR with the really scary looking things for the allocator, this one is surprisingly clean and straightforward. (I always kinda fear a Rust PR will turn gnarly eventually lol)

I've left a few notes. I'd be curious to see the results of benchmarking this new implementation vs the previous one.

profiling/src/collections/string_table/mod.rs
morrisonlevi: If only we had statistics on string data from libdatadog profiles...

profiling/src/collections/string_table/mod.rs
morrisonlevi: In Rust, allocators don't initialize things. They deal with the shapes of the things (`Layout`) and give you bytes. The caller is responsible to initialize them.

I could create a custom `String` type that has a method like `String::new_in(str: &str, allocator: &impl Allocator)`. I don't feel like it's worth it for _this_ purpose only, but if there's some other reason you can think of to want it, then we could do that. We could bind the lifetime of the custom String struct, but then we'd have to deal with self-referential structs, which are also unsafe. The [ouroboros](https://crates.io/crates/ouroboros) crate can help with that, but it's a bit complex. We can do it if you or others think it's worth it.

profiling/benches/interning_strings.rs
morrisonlevi: I haven't yet, but intend to!

profiling/src/collections/string_table/mod.rs
morrisonlevi: Thoughts on the reworded comment?

profiling/src/collections/string_table/mod.rs
morrisonlevi: I did a little refactoring to pull the alloc + initialization into a single operation so the interning code reads in a better way. Let me know what you think. We can make these `'self -> 'static -> 'self` operations safe, if we really want to. It's just a lot of boilerplate to use ouroboros for it.

profiling/src/collections/string_table/mod.rs
ivoanjo: I guess we could gather this if we wanted to -- we know in the backend which runtime a profile comes from, and so could even look at the data per-runtime.

profiling/src/collections/string_table/mod.rs
ivoanjo: I think the new version is clearer, but I'm still kinda curious (and I don't think any of these are blockers, it's fine as-is):

* Currently the allocation exposes its guts. Then the caller uses them for what it wants. I was kinda wondering if it would make sense to do it the other way -- by telling the allocator: here's a pointer, and here's length, return me a pointer that's a result of copying the pointer/length inside yourself. Maybe even use generics to be able to return a pointer to whatever was passed in, rather than a generic pointer?

* The transmute in `intern` is it just a really fancy way of saying "never try to deallocate this string directly, as I'm managing the memory in another way"?

ivoanjo (commented):
Just to clarify, is the idea not to merge this PR until we merge #402?

I think it's in pretty good state, but since a draft I'm not sure what plans you have and if I should approve as-is or wait for next steps, let me know! :pray:

profiling/src/collections/string_table/mod.rs
morrisonlevi: - The allocation API is a mirror of what is in the work-in-progress Allocator API for standard Rust. It's not something I think we should change.
- The transmute of the lifetime doesn't deal with the deallocation in this case, because it's just a reference either way. The transmute is to tell the compiler that the lifetime of these references outlives the data it refers to. As mentioned, we can safely do this with self-referential lifetimes but Rust doesn't have those.

morrisonlevi:
> Just to clarify, is the idea not to merge this PR until we merge [...]

Yes, although that is merged now and doesn't matter.

> I think it's in pretty good state, but since a draft I'm not sure what plans you have and if I should approve as-is or wait for next steps, let me know! 🙏

I think the code is overall in a good state. I've updated the PR description to match the state of current. There are still a few things like benchmarking the implementation vs the previous, as well as profiling the code to see if there is low hanging fruit. In particular, I wonder about `IndexMap` vs using a `Vec` + `HashMap`.

alloc/src/chain.rs
```diff
@@ -298,6 +298,32 @@ mod tests {
...
+        let allocator = ChainAllocator::new_in(4096, Global);
+
+        // Force an allocation, so it makes a chunk of the minimum size.
+        {
+            let ptr = allocator.allocate(Layout::new::<u8>()).unwrap();
+            unsafe { allocator.deallocate(ptr.cast(), Layout::new::<u8>()) };
+        }
+        // Should be a bit less than 4096, but use this over hard-coding a
+        // number, to make it more resilient to implementation changes.
+        let remaining_capacity = allocator.remaining_capacity();
```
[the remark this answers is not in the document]
ivoanjo: Then we learn about it through our tests and update them accordingly? ;)

profiling/benches/interning_strings.rs
morrisonlevi: Alright, some numbers! Changing from the old implementation to the new one resulted in:

```
benching string interning on wordpress profile
                        time:   [53.695 µs 53.977 µs 54.279 µs]
                        change: [-28.080% -27.464% -26.850%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  2 (2.00%) high severe
```

And then changing it back resulted in:

```
benching string interning on wordpress profile
                        time:   [75.224 µs 75.501 µs 75.820 µs]
                        change: [+40.269% +41.386% +42.503%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

This is unsurprising, because we spend less time in allocators, but I'm glad to have some data!

morrisonlevi:
> In particular, I wonder about IndexMap vs using a Vec + HashMap.

It's worse than using `IndexMap`, which is good (you would hope a structure meant to solve this, does it well).

I added the old implementation for other people to A/B test if they want. Import whichever `StringTable` you want to use, and then do:

```
cargo bench -p datadog-profiling
```

Performance is improved. I used `perf` to look for low hanging fruit, there was nothing obvious to me there. We're ready for final review.

profiling/src/internal/profile.rs
```diff
@@ -847,32 +838,37 @@ mod api_tests {
...
+        let mut profile = Profile::new(SystemTime::now(), &sample_types, Some(period));
 
         let prev = profile
             .reset_and_return_previous(None)
             .expect("reset to succeed");
 
         // Resolve the string values to check that they match (their string
         // table offsets may not match).
+        let mut strings = Vec::with_capacity(profile.strings.len());
+        let mut strings_iter = profile.strings.into_lending_iter();
```
danielsn, line 863: Could this become a `.map().collect()`?

profiling/src/collections/string_table/wordpress_test_data.rs
```diff
@@ -0,0 +1,1064 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+pub const WORDPRESS_STRINGS: [&str; 1059] = [
```
danielsn, line 4: A comment explaining where this came from, and how to modify it as needed, would be helpful to future maintainers.

profiling/benches/interning_strings.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        #[inline]
+        #[allow(clippy::len_without_is_empty)]
+        pub fn len(&self) -> usize {
+            self.strings.len()
+        }
+    }
+}
+
+// To benchmark a different implementation, import a different one.
```
danielsn, line 45: Make this a `cfg` rather than commenting out a line

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /// Test inserting strings from a real WordPress profile, although it's
+    /// unknown at this point which profiler version generated it. It's a small
+    /// sample. Here we're checking that we don't panic or otherwise fail, and
+    /// that the total number of strings and the bytes of those strings match
+    /// the profile.
+    #[test]
+    fn test_wordpress() {
+        test_from_src(&wordpress_test_data::WORDPRESS_STRINGS);
```
danielsn, line 277: We're only adding each string once, in order.  Would it make sense to randomly add strings multiple times too?

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+use crate::iter::{IntoLendingIterator, LendingIterator};
+use datadog_alloc::{AllocError, Allocator, ChainAllocator, VirtualAllocator};
+use std::alloc::Layout;
+
+/// A trait that indicates an allocator is arena allocator, meaning it doesn't
+/// deallocate individual items, but deallocates their memory as a  group when
+/// the arena is dropped.
+pub trait ArenaAllocator: Allocator {
+    /// Copies the str into the arena, and returns a slice to the new str.
+    fn allocate(&self, str: &str) -> Result<&str, AllocError> {
```
danielsn, line 17: `allocate` is a funny name for this

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    bytes: ChainAllocator<VirtualAllocator>,
+
+    /// The ordered hash set of unique strings. The order becomes the StringId.
+    /// The static lifetime is a lie, it is tied to the `bytes`, which is only
+    /// moved if the string table is moved e.g.
+    /// [StringTable::into_lending_iterator].
+    /// References to the underlying strings should generally not be handed,
+    /// but if they are, they should be bound to the string table's lifetime
+    /// or the lending iterator's lifetime.
+    strings: HashSet<&'static str>,
```
danielsn, line 62: Should we take a lifetime paramater / is there way to tie this to the lifetime of the underlying object?

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+impl<A: Allocator + Clone> ArenaAllocator for ChainAllocator<A> {}
+
+type Hasher = core::hash::BuildHasherDefault<rustc_hash::FxHasher>;
+type HashSet<K> = indexmap::IndexSet<K, Hasher>;
+
+/// Holds unique strings and provides [StringId]s that correspond to the order
+/// that the strings were inserted.
+pub struct StringTable {
```
danielsn, line 51: Where do we enforce that the first element is always `""`

profiling/src/collections/string_table/mod.rs
danielsn: The cost of reserving a bit too much shouldn't be that bad

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+        // - "trace endpoint"
+        // - A file and/or function name per frame.
+        // So with a capacity like 3, we end up reallocating a bunch on or
+        // before the very first sample. The number here is not fine-tuned,
+        // just skipping some obviously bad, tiny sizes.
+        strings.reserve(32);
+
+        // Always hold the empty string as item 0. Do not insert it via intern
+        // because that will try to allocate zero-bytes from the storage,
+        // which is sketchy.
```
danielsn, line 102: Seems like a worthwhile special case to handle in the allocator.

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+        strings.insert("");
+
+        Self { bytes, strings }
+    }
+
+    /// Returns the number of strings currently held in the string table.
+    #[inline]
+    #[allow(clippy::len_without_is_empty)]
```
danielsn, line 110: Since `is_empty` would be `false`?

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn len(&self) -> usize {
+        self.strings.len()
+    }
+
+    /// Adds the string to the string table if it isn't present already, and
+    /// returns a [StringId] that corresponds to the order that this string
+    /// was originally inserted.
+    ///
+    /// # Panics
+    /// This panics if the allocator fails to allocate a new chunk/node.
```
danielsn, line 120: I know errors are annoying to propagate, but panic will bring down the customer process.

profiling/src/collections/string_table/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+                    // underlying allocator fails when asking for a new chunk.
+                    // This is expected to be rare.
+                    let s = ArenaAllocator::allocate(&self.bytes, str)
+                        .expect("allocator for StringTable::intern to succeed");
+
+                    // SAFETY: all references to this value get re-narrowed to
+                    // the lifetime of the string table or iterator when
+                    // exposed to the user. The string table and iterator will
+                    // keep the arena alive, making the access safe.
+                    unsafe { core::mem::transmute::<&str, &'static str>(s) }
```
danielsn, line 145: is it possible to give it the same lifetime as `self`?

profiling/src/collections/string_table/mod.rs
morrisonlevi: Yes, I am aware, but this is just like the `GlobalAlloc` panic'ing if it runs out of memory. Same fringe error condition.

profiling/src/collections/string_table/mod.rs
morrisonlevi: There is no such thing as a `'self` lifetime, unfortunately 🙁

profiling/src/collections/string_table/mod.rs
morrisonlevi: Yes, it doesn't make sense to ask the question "is the string table empty?" because it's never empty.

profiling/src/collections/string_table/mod.rs
morrisonlevi: The trouble is, what allocators are supposed to do for zero-sized allocations is not pinned down. In C, you get implementation defined behavior for `malloc(0)` and it's generally recommended to avoid it because of this. The debate is still open for Rust.

profiling/src/collections/string_table/mod.rs
morrisonlevi: Have any ideas for a better one? If I had a better one, I'd have used it ^_^

profiling/src/collections/string_table/mod.rs
morrisonlevi: Yes, there is a way to do it using `ouroboros`. The code to do it is quite ugly, because they have to restrict all kinds of things to make it safe. Under the hood, it's basically doing the same trick.

profiling/src/collections/string_table/mod.rs
morrisonlevi: It's in `StringTable::new`, a few lines down.

profiling/src/collections/string_table/mod.rs
morrisonlevi: The `test_from_src` function adds every string twice.

profiling/src/internal/profile.rs
morrisonlevi: The lending iterator doesn't have it implemented. I could do that if you want, but I'd prefer to do it in another PR. Let me know.

profiling/src/collections/string_table/wordpress_test_data.rs
morrisonlevi: 👍🏻 I'll move the comment from `test_wordpress` to the top of this file.

profiling/benches/interning_strings.rs
morrisonlevi: When you are benchmarking you generally make tweaks to things and run it again. It's not just a static A/B thing we're going to run in CI so you want to automate swapping it out.. So it seems weird to have a cfg for this reason.

I think to use the cfg without making code changes, that would mean making a feature for it. Having a feature just for a benchmarking thing seems odd. Plus it breaks the idea of feature unification (features should be additive).

Do these things make sense? Thoughts?

r1viollet (approved):
LGTM
We should be careful about performance measures while deploying this.
The CI failure is not relevant to this PR.

danielsn (approved):
LGTM

profiling/src/internal/profile.rs
morrisonlevi: Doing `map` is possible it has usability issues. `collect` seems even harder, because you need to have a `Map` or something which breaks the borrow of the item (e.g. `String::from`) because otherwise the borrow prevents the lending iterator from moving to the next item (because it's `mut`).

PR #405, opened 2024-04-26
[crashtracker] fix fprintf formal string in receiver binary

danielsn:
# What does this PR do?

Fixes the format to work with non null terminated strings

# Motivation

It was wrong before

# Additional Notes

# How to test the change?

## For Reviewers


sanchda (approved):
LGTM, thank you!

sanchda (approved):

PR #406, opened 2024-04-27
[crashtracker] Remove profiling endpoint since we're using IT now

danielsn:
# What does this PR do?

Removes unneeded code to post crashreports as profile objects.

# Motivation

We were not using it, and this lets us remove any dependence of crashtracker on profiler.

# Additional Notes

# How to test the change?

## For Reviewers


r1viollet (approved):
LGTM, thanks for keeping things simple

PR #408, opened 2024-05-02
[profiling] Ensure shared library name has no soname set.

AlexJF:
# What does this PR do?

Ensure the libdatadog profiling shared library has no SONAME set in `.so` shared libraries.

# Motivation

In some environments, the libdatadog profiling shared library will include a SONAME and it will point to the library name previous to the rename in:

https://github.com/DataDog/libdatadog/blob/f5c7e888ee3aa6646986c90ed6917fe51c06a8c0/build-profiling-ffi.sh#L138

The existence of this SONAME makes it impossible to load libdatadog at runtime because the loading will be done based on the SONAME, not the actual library name:

```
2024-04-29_16:17:11.44159 W, [2024-04-29T16:17:11.441513 #43868]  WARN -- datadog: [datadog] Profiling was requested but is not supported, profiling disabled: There was an error loading the profiling native extension due to 'RuntimeError Failure to load datadog_profiling_native_extension.2.7.2_x86_64-linux due to libdatadog_profiling_ffi.so: cannot open shared object file: No such file or directory' at '/opt/dd-trace-rb/lib/datadog/profiling/load_native_extension.rb:41:in `<top (required)>''
```

# Additional Notes

We haven't been able to fully pinpoint what leads to different SONAME handling between different environments (e.g. my Ubuntu 24 LTS VM always includes them but not Ivo's Ubuntu 22 laptop) but we found the following places that show how "finicky" this can be:
* https://github.com/rust-lang/cargo/blob/d34d0a1e15b1005ed6ebddbe34aca2058e96293b/src/cargo/core/compiler/build_runner/compilation_files.rs#L691-L733
* https://github.com/lu-zero/cargo-c/blob/9c5c603f5b2b0d7c2228d13fe88ab631e0b195d4/src/target.rs#L76-L87 (I don't think we use this one but some similar code may be used somewhere in our build pipeline? 🤷‍♂️ )

# How to test the change?

## For Reviewers

### DIFF
--- a/build-profiling-ffi.sh
+++ b/build-profiling-ffi.sh
@@ -66,6 +66,7 @@ library_prefix="lib"
 remove_rpath=0
 fix_macos_rpath=0
 symbolizer=0
+strip_soname=0
 
 # Rust provides this note about the link libraries:
 # note: Link against the following native artifacts when linking against this
@@ -83,6 +84,7 @@ case "$target" in
         # on alpine musl, Rust adds some weird runpath to cdylibs
         remove_rpath=1
         symbolizer=1
+        strip_soname=1
         ;;
 
     "x86_64-apple-darwin"|"aarch64-apple-darwin")
@@ -98,6 +100,7 @@ case "$target" in
         expected_native_static_libs=" -ldl -lrt -lpthread -lgcc_s -lc -lm -lrt -lpthread -lutil -ldl -lutil"
         native_static_libs=" -ldl -lrt -lpthread -lc -lm -lrt -lpthread -lutil -ldl -lutil"
         symbolizer=1
+        strip_soname=1
         ;;
 
     "x86_64-pc-windows-msvc")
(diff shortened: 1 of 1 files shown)


r1viollet:
We talked about this. We are OK with removing the so name as we are not using it today. Would it be possible to just drop this within a cargo build script as part of the effort described in https://github.com/DataDog/libdatadog/issues/353

r1viollet:
can it be part of the rust flags (like linker flags)? I think @gleocadie did work that is similar, we should put settings in a similar place.

ivoanjo:
> can it be part of the rust flags (like linker flags)? I think @gleocadie did work that is similar, we should put settings in a similar place.

From what I chatted with @AlexJF, he did not find such an option, hence why he went with the "fix-after-the-fact" approach.

ivoanjo:
I see some failures in our build pipeline

> ./build-profiling-ffi.sh: line 151: patchelf: command not found

We'll need a PR on that side to make sure our images have patchelf.

ivoanjo:
PR to add `patchelf` to our centos build images: https://github.com/DataDog/libddprof-build/pull/40

ivoanjo:
CI is now green for this one! @r1viollet thoughts on merging? Or do you want to hold on until we can see if there's a cleaner way to tell the rust compiler to not emit the soname at all?

ivoanjo:
Actually CI isn't green lol, either I was looking at the wrong tab, or some of the jobs hadn't finished when I looked at it. Pardon the temporary insanity. Will look into the missing issue -- I think I didn't bump the CI image version thingy correctly.

ivoanjo:
I'm moving this back to draft until we can pick it up again

ivoanjo:
Closing as superceded by #628

PR #412, opened 2024-05-02
[crashtracker] Take a user defined list of additional files

danielsn:
# What does this PR do?

Allows the client to specify which files to record, rather than hard coding it.

# Motivation

Previously, we had hardcoded to use `/proc/meminfo` and `/proc/cpuinfo` as the additional files.  The client may not want these, or may want others.  Give them the choice.

# Additional Notes

# How to test the change?

## For Reviewers


sanchda (approved):
LGTM--thank you for this!

PR #414, opened 2024-05-06
[PROF-9476] Managed string storage for interning over several profiles

AlexJF:
# What does this PR do?

PoC for allowing string interning to survive across several profiles. This is used in https://github.com/DataDog/dd-trace-rb/pull/3628 to reduce the overhead of heap profiling.

THIS CODE IS NOT PRODUCTION READY AND IS SIMPLY A HACKISH PoC!

# Motivation

Heap profiling is a stateful mode of profiling where samples associated with live objects may be emitted across several profiles (those where the live object stays alive). Because of this, information such as the allocation class and allocation stacktrace needs to be preserved alongside the tracked object so heap samples can be re-inserted in subsequent profiles.

libdatadog already does a good job of handling and deduplicating strings in the span of a single profile. Exposing an API to allow doing this work across several profiles would prevent users from having to re-implement this outside of libdatadog and duplicating a lot of work.

Initial testing with dd-trace-rb shows great promise, allowing us to increase heap sampling rate by 10x with negligible overhead compared to the currently released implementation:

image

image

# Additional Notes

# How to test the change?

## For Reviewers


ivoanjo:
I've added keep-open for this one, as I plan to pick it up soon :tm:

ivoanjo:
Closing in favor of https://github.com/DataDog/libdatadog/pull/607

PR #415, opened 2024-05-06
Run tests with nextest and publish results

bantonsson:
# What does this PR do?

This PR does the following:

* Adds the `nextest` test runner and configuration to run _single threaded_ tests separately.
* Adds test result reporting as [checks to the PR](https://github.com/DataDog/libdatadog/actions/runs/9001508636#summary-24727897494)
* Renames the modules for single threaded tests to `single_threaded_tests`, so they can be easily run separately with `nextest`.
* Removes the use of `ignore` to separate out single threaded tests.
* Adds a `skip_root_tests` module name for tests (that were previously ignored 🤷🏼‍♂️) that can't run as `root` on GitLab CI.
* Changes the module name for most other tests to `tests` for consistency and easy skipping when building with `cross`, which doesn't support `nextest`.

# Motivation

* Tests were using `skip` as a way to separate out different test configuration needs (single threaded tests).
* Failures were hard to find (scrolling through test output).

# Additional Notes

Tests should now be run with `cargo nextest run`

To install `nextest` run:

```
cargo install --locked 'cargo-nextest@0.9.67'
```

or if you are on a rust version newer than `1.71.1` run:

```
cargo install --locked 'cargo-nextest@0.9.*'
```

# Dependencies

This PR depends on new build images for the GitLab build.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -14,13 +14,16 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust
-        run: rustup update stable
+        run: rustup install nightly && rustup default nightly
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@v2
         with:
-          tool: cargo-llvm-cov@0.6.9
-      - name: Generate code coverage
-        run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
+          tool: cargo-llvm-cov@0.6.9,nextest@0.9.68
+      - name: Generate code coverage (including doc tests)
+        run: |
+          cargo llvm-cov --all-features --workspace --no-report nextest
+          cargo llvm-cov --all-features --workspace --no-report --doc
+          cargo llvm-cov report --doctests --lcov --output-path lcov.info
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
--- a/README.md
+++ b/README.md
@@ -31,3 +31,23 @@ bash build-profiling-ffi.sh /opt/libdatadog
 
 - Rust 1.71 or newer with cargo
 - `cmake` and `protoc`
+
+### Running tests
+
+This project uses [cargo-nextest][nt] to run tests.
+
+```bash
+cargo nextest run
+```
+
+#### Installing cargo-nextest
+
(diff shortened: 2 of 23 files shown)


.github/workflows/coverage.yml
```diff
@@ -14,13 +14,16 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust
-        run: rustup update stable
+        run: rustup install nightly && rustup default nightly
```
ekump, line 17: Just curious - why does coverage require nightly now?

danielsn:
What happens if someone runs `cargo test`?

.gitlab-ci.yml
```diff
@@ -2,7 +2,7 @@ variables:
   # These are gitlab variables so that it's easier to do a manual deploy
   # If these are set witih value and description, then it gives you UI elements
   DOWNSTREAM_BRANCH:
-    value: "main"
+    value: "ban/switch-to-nextest" # TODO switch back
```
ekump, line 5: Just calling this out in case you meant switch it back before posting for review.

README.md
```diff
@@ -30,3 +30,23 @@ bash build-profiling-ffi.sh /opt/libdatadog
 #### Build Dependencies
 
 Rust 1.69 or newer with cargo. Some platforms may need protoc; others have it shipped in prost-build.
+
```
ekump, line 33: Beyond the scope of this PR, but just a thought...should we offer a dev docker image for working on datadog with all the necessary tools pre-installed? Some client libraries do this to make setting up a dev env easier.

ekump (approved):
looks good! Glad we can get rid of the `ignore` workflow for single-threaded tests.

.github/workflows/coverage.yml
bantonsson: It's needed to merge the doctest coverage with the normal test coverage (don't know why it's not in stable). This little dance is an unfortunate side effect of `nextest` not running the doctest (yet).

bantonsson:
@danielsn if you do `cargo test`, all tests will be run as normal, and some tests that need to be run single threaded will fail.

morrisonlevi (approved):
Looks fine to me, assuming CI passes and whatnot.

PR #416, opened 2024-05-06
Reduce log verbosity for duplicate labels

sanchda:
# What does this PR do?

When duplicate labels are detected, we print both labels.  Is there any value to that?

# Motivation

Make logs a little bit easier to read.  It's not a big deal, if anyone has a strong opinion to the contrary.

# Additional Notes

No.

# How to test the change?

It's a change to the output of logs, so I'm not sure a test is necessary.  We already test for the condition which causes the logs, AFAICT.

## For Reviewers


ivoanjo (commented):
I think the intention here was to allow debugging when the same key (label name) is duplicated, but not its value :thinking:

PR #418, opened 2024-05-07
Collect some of the `trace_api.*` metrics in the trace flusher

iamluc:
# What does this PR do?

This PR splits the `tracer_utils` into smaller parts and collects metrics for:
* trace_api.requests.
* trace_api.errors.
* trace_api.responses.
These metrics are then sent by the "self telemetry" worker of the sidecar.

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/sidecar/src/self_telemetry.rs
+++ b/sidecar/src/self_telemetry.rs
@@ -24,6 +24,9 @@ struct MetricData<'a> {
     active_sessions: ContextKey,
     memory_usage: ContextKey,
     logs_created: ContextKey,
+    trace_api_requests: ContextKey,
+    trace_api_responses: ContextKey,
+    trace_api_errors: ContextKey,
 }
 impl<'a> MetricData<'a> {
     async fn send(&self, key: ContextKey, value: f64, tags: Vec<Tag>) {
@@ -34,6 +37,8 @@ impl<'a> MetricData<'a> {
     }
 
     async fn collect_and_send(&self) {
+        let trace_metrics = self.server.trace_flusher.collect_metrics();
+
         let mut futures = vec![
             self.send(
                 self.submitted_payloads,
@@ -61,6 +66,42 @@ impl<'a> MetricData<'a> {
                 self.logs_created,
                 count as f64,
                 vec![Tag::new("level", level.as_str().to_lowercase()).unwrap()],
+            ));
+        }
+
+        if trace_metrics.api_requests > 0 {
+            futures.push(self.send(
+                self.trace_api_requests,
+                trace_metrics.api_requests as f64,
+                vec![],
+            ));
+        }
+        if trace_metrics.api_errors_network > 0 {
+            futures.push(self.send(
+                self.trace_api_errors,
+                trace_metrics.api_errors_network as f64,
+                vec![Tag::new("type", "network").unwrap()],
+            ));
+        }
+        if trace_metrics.api_errors_timeout > 0 {
+            futures.push(self.send(
+                self.trace_api_errors,
+                trace_metrics.api_errors_timeout as f64,
+                vec![Tag::new("type", "timeout").unwrap()],
+            ));
+        }
+        if trace_metrics.api_errors_status_code > 0 {
+            futures.push(self.send(
+                self.trace_api_errors,
+                trace_metrics.api_errors_status_code as f64,
+                vec![Tag::new("type", "status_code").unwrap()],
+            ));
+        }
+        for (status_code, count) in &trace_metrics.api_responses_count_per_code {
+            futures.push(self.send(
+                self.trace_api_responses,
+                *count as f64,
+                vec![Tag::new("status_code", status_code.to_string().as_str()).unwrap()],
             ));
         }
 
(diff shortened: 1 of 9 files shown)


trace-utils/src/send_data.rs
```diff
@@ -0,0 +1,229 @@
...
+    T: prost::Message,
+{
+    let mut buf = Vec::with_capacity(payload.encoded_len());
+    payload.encode(&mut buf)?;
+    Ok(buf)
+}
+
+#[derive(Debug, Clone)]
+pub struct SendData {
+    pub tracer_payloads: Vec<pb::TracerPayload>,
```
ekump, line 112: Do we really need to expose `tracer_payloads` publicly? From what I can tell, it's only for the tests in the `trace-mini-agent`, which is not a good pattern. We shouldn't be exposing internal implementation details just for unit tests in this way. 

If we are only exposing `tracer_payloads` for tests, could we refactor the tests to use httpmock and properly test the outputs? If that is something that we want to explore, I think we could do that in a follow-up PR since we want to get this merged in soon. Maybe we can just add a TODO comment then?

trace-utils/src/send_data.rs
```diff
@@ -0,0 +1,229 @@
...
+            last_result: Err(anyhow!("No requests sent")),
+            requests_count: 0,
+            responses_count_per_code: Default::default(),
+            errors_timeout: 0,
+            errors_network: 0,
+            errors_status_code: 0,
+        }
+    }
+
+    async fn update(
```
ekump, line 41: How do you envision this working with retry logic? Would we call `update` for each attempt of a request payload? Or should update only be called for a payload once, regardless of how many retries it took?

trace-utils/src/send_data.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub last_result: anyhow::Result<Response<Body>>,
+    pub requests_count: u64,
+    pub responses_count_per_code: HashMap<u16, u64>,
+    pub errors_timeout: u64,
+    pub errors_network: u64,
+    pub errors_status_code: u64,
+}
+
+impl SendDataResult {
+    fn new() -> SendDataResult {
```
ekump, line 30: Since you're not taking in any parameters, should this just be a `Default` impl?

trace-utils/src/trace_utils.rs
```diff
@@ -1,40 +1,22 @@
...
-use futures::stream::FuturesUnordered;
-use futures::StreamExt;
-use hyper::http::HeaderValue;
-use hyper::{body::Buf, Body, Client, HeaderMap, Method, Response, StatusCode};
+use hyper::{body::Buf, Body};
 use log::{error, info};
-use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
+pub use crate::send_data::SendData;
```
ekump, line 8: nitpick: Should we move `trace_utils` to use a `mod.rs` file like we do in the sidecar to handle exporting like this? Feels a bit awkward / indirect like this. (If we want to do this, we can do in a separate PR).

sidecar/src/self_telemetry.rs
```diff
@@ -64,6 +69,42 @@ impl<'a> MetricData<'a> {
...
+                self.trace_api_errors,
+                trace_metrics.api_errors_status_code as f64,
+                vec![Tag::new("type", "status_code").unwrap()],
+            ));
+        }
+        for (status_code, count) in &trace_metrics.api_responses_count_per_code {
+            futures.push(self.send(
+                self.trace_api_responses,
+                *count as f64,
+                vec![Tag::new("status_code", status_code.to_string().as_str()).unwrap()],
```
ekump, line 104: The most minor of nitpicks, feel free to ignore: `&status_code.to_string()` also works and is a bit less verbose.

sidecar/src/service/tracing/trace_flusher.rs
```diff
@@ -131,6 +158,12 @@ impl TraceFlusher {
             send_data_size: self.inner.lock().unwrap().traces.send_data_size as u32,
         }
     }
+    
+    pub fn collect_metrics(&self) -> TraceFlusherMetrics {
```
ekump, line 163: I know it's technically not mutable, but I'm not sure the name `collect_metrics` conveys the actual mutability of this function on `TraceFlusherMetrics`? Also, I don't think it's actually collecting metrics at this point (the caller is doing the collecting)?

What about calling the function `remove()`, similar to the [remove function in HashMap](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.remove)? Or something like `fetch_and_clear`?

trace-utils/src/send_data.rs
bantonsson: This is a good point. We shouldn't expose internals unnecessarily in this way, but I think that going all the way to sending to the `httpmock` isn't necessary. Just adding a test only method on the `SendData` implementation that hands out payload number X, is fine by me.

trace-utils/src/send_data.rs
bantonsson: Agreed

trace-utils/src/send_data.rs
hoolioh: I struggled with this because [coalesce_send_data](https://github.com/DataDog/libdatadog/blob/luc/sidecar-collect-trace_api-metrics/sidecar/src/service/tracing/trace_flusher.rs#L232) needs to mutate the `tracer_payloads` fields so I thought about implementing a getter to return a mutable reference but in the end is kind of the same as making the the field public. Also I thought about restructuring the SendData and TraceFlusher to improve the encapsulation but I realized it'd overcomplicate this PR. Anyway if you guys think the getter is a better solution we can go that way or if you can think of a better solution please let me know.

bantonsson: That can be public to the crate, right?

bantonsson (approved):
As mentioned internally, it’s fine to merge this as is and make changes in the follow up PR.

trace-utils/src/send_data.rs
hoolioh: Since the retry logic tries to minimize errors I'd choose the latter because it'd be confusing to send errors or dropped statistics for traces that will end up being sent.

ekump: @hoolioh Ah, that's a great point. Maybe retry metrics should be it's own thing at some point? It may be useful information.

bantonsson: I think that the retries can be deduced from the number of API requests, the number of failures of different types, as well as the number of _dropped_ traces.

PR #421, opened 2024-05-08
[WIP] Update GitHub workflows

bantonsson:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


.github/workflows/lint.yml
```diff
@@ -102,8 +102,9 @@ jobs:
 
 
       - name: export the generated license file on failure
-        if: ${{ failure() }}
-        uses: actions/upload-artifact@v3
+        if: failure()
```
paullegranddc, line 105: Does that work? I was under the impression that expressions had to be kept between double curly brackets? https://docs.github.com/en/actions/learn-github-actions/expressions#failure

pawelchcki (approved):

.github/workflows/lint.yml
bantonsson: I thought so as well, but then I saw that the files where mixing this and it turns out that GH evaluates the things after `if` as if it is an expression. I can switch all of them to the `${{ ... }}` syntax if you prefer.

ekump (approved):

PR #422, opened 2024-05-09
[PROF-9742] Package libdatadog v9.0.0 for Ruby

ivoanjo:
# What does this PR do?

This PR includes the changes documented in the "Releasing a new version to rubygems.org" part of the README: https://github.com/datadog/libdatadog/tree/main/ruby#releasing-a-new-version-to-rubygemsorg

(It's also very similar to the v7.0.0 release PR).

It also includes a tiny paper cut fix: it disables the Rake `build` task which is the default way to package a Ruby gem, since we have our own custom packaging steps in the `Rakefile`.

# Motivation

Enable Ruby to use libdatadog v9.0.0.

# Additional Notes

We ended up not packaging the previous release (v8) for Ruby, so we're directly jumping from v7 to v9.

# How to test the change?

I've tested this release locally using the changes in https://github.com/DataDog/dd-trace-rb/pull/3627 .

As a reminder, new libdatadog releases don't get automatically picked up by dd-trace-rb, so the PR that bumps the Ruby profiler will also test this release against all supported Ruby versions.

### DIFF
--- a/ruby/Rakefile
+++ b/ruby/Rakefile
@@ -11,7 +11,7 @@ require "rubygems/package"
 
 RSpec::Core::RakeTask.new(:spec)
 
-LIB_VERSION_TO_PACKAGE = "7.0.0"
+LIB_VERSION_TO_PACKAGE = "9.0.0"
 unless LIB_VERSION_TO_PACKAGE.start_with?(Libdatadog::LIB_VERSION)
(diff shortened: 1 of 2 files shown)


danielsn (approved):
This PR appears totally mechanical.  Would it be worth adding creating this PR to the release CI action?

ivoanjo:
> This PR appears totally mechanical. Would it be worth adding creating this PR to the release CI action?

Yes, but I haven't had the chance to do it so far :)

PR #423, opened 2024-05-09
[PROF-8917] Add Ruby helper to get `ld_library_path` for libdatadog

ivoanjo:
# What does this PR do?

This PR adds a `Libdatadog.ld_library_path` helper to the libdatadog Ruby support code.

This helper returns the full path to the `lib/` folder where the libdatadog `.so` files are contained.

# Motivation

I'm adding this to support the crashtracker for Ruby: we need this information when starting the crashtracker, and rather than having the knowledge of folder layouts inside dd-trace-rb, I think it makes sense for the libdatadog Ruby code to keep this knowledge.

# Additional Notes

I'm planning to include this change in the libdatadog 9 release for Ruby.

# How to test the change?

Test coverage is included (and CI runs these tests).

### DIFF
--- a/ruby/lib/libdatadog.rb
+++ b/ruby/lib/libdatadog.rb
@@ -64,4 +64,12 @@ module Libdatadog
 
     File.absolute_path("#{pkgconfig_folder}/../../bin/libdatadog-crashtracking-receiver")
   end
+
+  def self.ld_library_path
+    pkgconfig_folder = self.pkgconfig_folder
+
+    return unless pkgconfig_folder
+
+    File.absolute_path("#{pkgconfig_folder}/../")
+  end
 end
--- a/ruby/spec/libdatadog_spec.rb
+++ b/ruby/spec/libdatadog_spec.rb
@@ -30,11 +30,7 @@ RSpec.describe Libdatadog do
       end
     end
 
-    context "when no binaries are available in the vendor directory" do
-      describe ".available_binaries" do
-        it { expect(Libdatadog.available_binaries).to be_empty }
-      end
-
+    shared_examples_for "libdatadog not in usable state" do
       describe ".pkgconfig_folder" do
         it { expect(Libdatadog.pkgconfig_folder).to be nil }
       end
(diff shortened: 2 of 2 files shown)


ruby/lib/libdatadog.rb
```diff
@@ -64,4 +64,12 @@ def self.path_to_crashtracking_receiver_binary
 
     File.absolute_path("#{pkgconfig_folder}/../../bin/libdatadog-crashtracking-receiver")
   end
+
+  def self.ld_library_path
+    pkgconfig_folder = self.pkgconfig_folder
+
+    return unless pkgconfig_folder
+
+    File.absolute_path("#{pkgconfig_folder}/../")
```
bantonsson, line 73: I trust you know what you are doing... 😉

bantonsson (approved):

ruby/lib/libdatadog.rb
ivoanjo: Ahaha thanks! To be clear, what this does is take the structure of our binary builds:

```
% tree
.
├── bin
│   └── libdatadog-crashtracking-receiver
├── cmake
│   └── DatadogConfig.cmake
├── include
│   └── datadog
│       ├── blazesym.h
│       ├── common.h
│       ├── profiling.h
│       └── telemetry.h
├── lib
│   ├── libdatadog_profiling.a
│   ├── libdatadog_profiling.so
│   ├── libdatadog_profiling.so.debug
│   └── pkgconfig
│       ├── datadog_profiling.pc
│       ├── datadog_profiling-static.pc
│       └── datadog_profiling_with_rpath.pc
├── LICENSE
├── LICENSE-3rdparty.yml
└── NOTICE
```

and encodes the knowledge that the .so file is in the parent folder to the pkgconfig file.

I could've implemented this another way (e.g. search for the so file directly or something), but since this is internal to the libdatadog helper, and we already have another helper that does this relative find thing, I just went with it.

(Since on the Ruby library side we only call these methods and don't encode any assumptions on the folder structure, we can always update the helpers and move stuff around whenever we want to).

PR #425, opened 2024-05-12
Fix comments parsing in dedup_headers

bwoebi:
After d3f85442971ec6074e9912f4b9df578cb4817811 it broke cbindgen generation in dd-trace-php.

The previous /\*\*.*?\*/ parsing would match from a comment to the next typedef fulfilling the criteria, instead of matching just the comment itself.

This is fixed by matching against /\*\*([^*]|\*[^/])*\*/, where the */ sequence is disallowed within a comment. (Also saves us from needing the ungreedy match.)

### DIFF
--- a/tools/src/bin/dedup_headers.rs
+++ b/tools/src/bin/dedup_headers.rs
@@ -13,7 +13,7 @@ use std::io::{self, BufReader, BufWriter, Read, Seek, Write};
 
 fn collect_definitions(header: &str) -> Vec<regex::Match<'_>> {
     lazy_static::lazy_static! {
-        static ref HEADER_TYPE_DECL_RE: Regex = RegexBuilder::new(r"^(/\*\*.*?\*/\n)?(#define [a-zA-Z_0-9]+ [^\n]+|typedef (struct|enum) [a-zA-Z_0-9]+ +(\{.*?\} )?[a-zA-Z_0-9]+;)\n+")
+        static ref HEADER_TYPE_DECL_RE: Regex = RegexBuilder::new(r"^(/\*\*([^*]|\*+[^*/])*\*+/\n)?(#define [a-zA-Z_0-9]+ [^\n]+|typedef (struct|enum) [a-zA-Z_0-9]+ +(\{.*?\} )?[a-zA-Z_0-9]+;)\n+")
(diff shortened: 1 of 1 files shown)


tools/src/bin/dedup_headers.rs
```diff
@@ -13,7 +13,7 @@ use std::io::{self, BufReader, BufWriter, Read, Seek, Write};
 
 fn collect_definitions(header: &str) -> Vec<regex::Match<'_>> {
     lazy_static::lazy_static! {
-        static ref HEADER_TYPE_DECL_RE: Regex = RegexBuilder::new(r"^(/\*\*.*?\*/\n)?(#define [a-zA-Z_0-9]+ [^\n]+|typedef (struct|enum) [a-zA-Z_0-9]+ +(\{.*?\} )?[a-zA-Z_0-9]+;)\n+")
+        static ref HEADER_TYPE_DECL_RE: Regex = RegexBuilder::new(r"^(/\*\*([^*]|\*[^/])*\*/\n)?(#define [a-zA-Z_0-9]+ [^\n]+|typedef (struct|enum) [a-zA-Z_0-9]+ +(\{.*?\} )?[a-zA-Z_0-9]+;)\n+")
```
bantonsson, line 16: Could we add  some tests for these things? To me it's not clear what assumptions can/should be made about the comments we are matching against. This new pattern has problems with comments ending in `**/` and will not stop there. Is that an issue?

bwoebi: No, it's not an issue, but trivial to fix, added a `+` there.

tools/src/bin/dedup_headers.rs
```diff
@@ -93,3 +93,72 @@ fn main() {
...
+typedef struct ddog_Vec_U8 {
+  const uint8_t *ptr;
+} ddog_Vec_U8;
+";
+    let matches = collect_definitions(header);
+
+    assert_eq!(matches.len(), 1);
+    assert_eq!(
+        matches[0].as_str(),
+        r"typedef struct ddog_Vec_U8 {
```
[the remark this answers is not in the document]
bwoebi: Yes, it's just a test that it doesn't malfunction completely.

PR #426, opened 2024-05-13
Don't validate that endpoint is not empty because it doesn't need to be present

paullegranddc:
# What does this PR do?

Remove the validation that endpoint is not empty when casting the crashtrackers parameters.
The endpoint can be empty, because it will be read from env variables here https://github.com/DataDog/libdatadog/blob/paullgdc/crashtracker/fix_enpoint_env_read/crashtracker/src/telemetry.rs#L63

# Motivation

# Additional Notes

# How to test the change?


profiling-ffi/src/crashtracker/datatypes.rs
```diff
@@ -89,7 +89,7 @@ impl<'a> TryFrom<CrashtrackerConfiguration<'a>>
             vec
         };
         let create_alt_stack = value.create_alt_stack;
-        let endpoint = unsafe { Some(exporter::try_to_endpoint(value.endpoint)?) };
+        let endpoint = unsafe { exporter::try_to_endpoint(value.endpoint).ok() };
```
bantonsson, line 92: Tests?

r1viollet, line 92: What does it mean for the crashtracker to have an empty endpoint ? We can add some comments to ensure the ffi API is clear.

paullegranddc:
For context, this is a request from @kevingosse . He want to use the crashtracker in situations were the it doesn't have access to the library variables, so the goal of this PR is to use agent host detection if no endpoint is provided by the crashtracker

ddcommon/src/lib.rs
```diff
@@ -79,48 +79,45 @@ where
...
+///     * For windows, interprets everything after windows: as path
+///     * For unix, interprets everything after unix:// as path
+/// * For file scheme implementation will simply backfill missing authority section
 pub fn parse_uri(uri: &str) -> anyhow::Result<hyper::Uri> {
-    let scheme_pos = if let Some(scheme_pos) = uri.find("://") {
-        scheme_pos
+    if let Some(path) = uri.strip_prefix("unix://") {
+        encode_uri_path_in_authority("unix", path)
+    } else if let Some(path) = uri.strip_prefix("windows:") {
+        encode_uri_path_in_authority("windows", path)
```
paullegranddc, line 94: After checking with @gleocadie , this is actually the format of windows pipe urls, no "//".
This fucntion is only used in the telemetry lib, which is not yet used on windows so this should not be a breaking change.

profiling-ffi/src/crashtracker/datatypes.rs
paullegranddc: Added 

>   /// If ProfilingEndpoint is left to a zero value (enum value for Agent + empty charslice),
    /// the crashtracker will infer the agent host from env variables.

ddcommon/src/lib.rs
```diff
@@ -79,48 +79,45 @@ where
...
+    } else if let Some(path) = uri.strip_prefix("file://") {
+        let mut parts = uri::Parts::default();
+        parts.scheme = uri::Scheme::from_str("file").ok();
+        parts.authority = Some(uri::Authority::from_static("localhost"));
+
+        // TODO: handle edge cases like improperly escaped url strings
+        //
+        // this is eventually user configurable field
+        // anything we can do to ensure invalid input becomes valid - will improve usability
```
danielsn, line 103: Should we give an error message if the URL is invalid, so that the user can be encouraged to fix it?

ddtelemetry/src/config.rs
```diff
@@ -51,40 +52,50 @@ fn endpoint_with_telemetry_path(
...
 /// (either through env variable, or that could be set from the )
 pub struct Settings {
-    pub agent_host: String,
-    pub trace_agent_port: u16,
-    pub trace_agent_url: Option<Uri>,
+    // Env parameter
+    pub agent_host: Option<String>,
+    pub trace_agent_port: Option<u16>,
+    pub trace_agent_url: Option<String>,
+    pub trace_pipe_name: Option<String>,
```
danielsn, line 59: I assume only some of these can be set at once? Would it be possible to use an `Enum` to encode that?

ddtelemetry/src/config.rs
```diff
@@ -203,91 +243,202 @@ impl Config {
         &CFG
     }
 
-    pub fn set_url(&mut self, url: &str) -> anyhow::Result<()> {
+    /// set_host sets the host telemetry should connect to.
+    ///
+    /// It handles http(s?), and unix and windows pipes of the format (unix)|(windows)://<path to
+    /// object> If the host_url is http/https, any path will be ignored and replaced by the
+    /// appropriate telemetry endpoint path
+    pub fn set_host(&mut self, host_url: &str) -> anyhow::Result<()> {
```
danielsn, line 251: Does it also handle `file://`

ddtelemetry/src/config.rs
paullegranddc: This struct is a wrapper around env variables for easier testing (you can look at this file's tests for instance).
All of them can be set at once, but some of them take precedence over each other ([see here](https://github.com/DataDog/architecture/blob/ee430517f3435cd686a7236e4b1b0a9715ef8b7c/rfcs/apm/integrations/trace-autodetect-agent-config/rfc.md#explicit-configuration)), and the logic on how is in Configuration::from_setting.

ddtelemetry/src/config.rs
paullegranddc: It does, adding it to the comment

crashtracker/src/telemetry.rs
```diff
@@ -70,7 +70,7 @@ impl TelemetryCrashUploader {
 
             // ignore result because what are we going to do?
             let _ = if endpoint.url.scheme_str() == Some("file") {
-                cfg.set_url(&format!("file://{}.telemetry", endpoint.url.path()))
+                cfg.set_host(&format!("file://{}.telemetry", endpoint.url.path()))
```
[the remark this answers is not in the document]
paullegranddc: I think `set_host` is clearer than`set_url` (I wrote the method and I ended up confused on what it did exactly, and naming wasn't a great help), because this method just sets the host and discards any path on the string

ddcommon/src/lib.rs
```diff
@@ -79,48 +79,45 @@ where
...
-    let scheme_pos = if let Some(scheme_pos) = uri.find("://") {
-        scheme_pos
+    if let Some(path) = uri.strip_prefix("unix://") {
+        encode_uri_path_in_authority("unix", path)
+    } else if let Some(path) = uri.strip_prefix("windows:") {
+        encode_uri_path_in_authority("windows", path)
+    } else if let Some(path) = uri.strip_prefix("file://") {
+        let mut parts = uri::Parts::default();
+        parts.scheme = uri::Scheme::from_str("file").ok();
+        parts.authority = Some(uri::Authority::from_static("localhost"));
```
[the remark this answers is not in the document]
paullegranddc: the authority value doesn't matter in this case, but it shouldn't be empty, otherwise Uri::from_parts returns an error.

crashtracker/src/telemetry.rs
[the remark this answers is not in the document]
paullegranddc: `set_host_from_url` ?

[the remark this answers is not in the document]
bantonsson: I still feel that `set_host...` implies that there should be a `set_port...`, but `set_host_from_url` is fine.

bantonsson (approved):

PR #432, opened 2024-05-15
Fix telemetry shutdown and log max filter

bwoebi:
Fixing max() in max_level_hint().

And properly dropping the telemetry worker:
When the runtime is shutdown with pending apps, it will send Stop, but never actually shutdown the telemetry instances. This regressed with d1fb3bc934e4b51b2ba1cb45818717351ad43b0. Now we simply drop the telemetry handle, which implicitly stops the worker as well causes the worker to be joined properly.

At least this led to the occasional deadlocked sidecar.

### DIFF
--- a/sidecar/src/log.rs
+++ b/sidecar/src/log.rs
@@ -243,7 +243,7 @@ impl<S: Subscriber> Filter<S> for &MultiEnvFilter {
             .unwrap()
             .values()
             .map(|f| f.max_level_hint())
-            .min()
+            .max()
             .flatten()
     }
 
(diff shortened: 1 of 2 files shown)


sidecar/src/log.rs
```diff
@@ -243,7 +243,7 @@ impl<S: Subscriber> Filter<S> for &MultiEnvFilter {
             .unwrap()
             .values()
             .map(|f| f.max_level_hint())
-            .min()
+            .max()
```
bantonsson, line 246: The whole _ordering_ implementations for `Level` and `LevelFilter` makes my head spin since the actual underlying `enum` for both is ordered the other way around, i.e. `Debug` is `1` and `Warn` is `3`, but apparently `Debug` is greater than `Warn`.

bantonsson (approved):

sidecar/src/log.rs
bwoebi: Yes, that's exactly why I made the mistake in the first place. Really a what the fuck moment :-D

PR #433, opened 2024-05-17
Add retry logic to trace_utils::SendData

ekump:
# What does this PR do?

Introduces retry logic to send traces in trace_utils::SendData. The retry strategy is configurable via the `RetryStrategy` struct. `SendData` will use a Default strategy, but callers have the option to configure the strategy via the SendData.set_retry_strategy() function. **TraceExporter, Mini-agent, and SIdecar will have the default retry strategy behavior.** 

# Motivation

In order to turn the sidecar on by default for PHP there needs to be support for retries when flushing traces fail. Adding the retry logic to trace_utils made the most sense (and can be used outside of the sidecar)

# Additional Notes

The `RetryStrategy` struct should live in another file. And some of the test helper functions should move to a common place. But, there are other PRs in flight that will be modifying `send_data.rs`. So we can do that work in a follow-up PR to minimize disruption. TODO comments with Jira IDs have been added to the code in these spots.

# How to test the change?

Unit tests have been added. Longer-term we may want to consider using the test-agent.


trace-utils/src/send_data.rs
```diff
@@ -151,79 +302,508 @@ impl SendData {
...
+    async fn send_with_msgpack(&self) -> SendDataResult {
+        let mut result = SendDataResult::new();
+
+        let mut req = self.create_request_builder();
+        req = req.header("Content-type", "application/msgpack");
+
+        let (template, _) = req.body(()).unwrap().into_parts();
+
+        let mut futures = FuturesUnordered::new();
+        for tracer_payload in self.tracer_payloads.clone().into_iter() {
```
bantonsson, line 355: This can be `(&self.tracer.payloads).into_iter()` to avoid the clone.

trace-utils/src/send_data.rs
```diff
@@ -151,79 +302,508 @@ impl SendData {
...
-                .await
-                .map_err(SendRequestError::Hyper)
-        }
+        req
+    }
+
+    async fn send_request(
+        &self,
+        req: HttpRequestBuilder,
+        payload: Vec<u8>,
```
bantonsson, line 311: If this is a `Bytes` instead, then there can be a shallow arc copy in the calling code, since the `Bytes` will be constructed from a `Vec<u8>`

trace-utils/src/send_data.rs
```diff
@@ -133,14 +222,76 @@ impl SendData {
...
+    ) -> Result<Response<Body>, SendRequestError> {
+        let mut request_attempt = 0;
+
+        loop {
+            request_attempt += 1;
+            let mut req = self.create_request_builder();
+            req = req.header("Content-type", content_type);
+
+            let result = self.send_request(req, payload.clone()).await;
```
bantonsson, line 261: This should be a `Bytes` that you clone instead to share the underlying data, and the `Bytes` can be created outside the `loop` with `Bytes::from(payload)`

trace-utils/src/send_data.rs
```diff
@@ -151,79 +302,508 @@ impl SendData {
...
+            .await
+            .map_err(SendRequestError::Hyper)
+    }
+
+    async fn send_with_protobuf(&self) -> SendDataResult {
         let mut result = SendDataResult::new();
 
-        if target.api_key.is_some() {
-            req = req.header("Content-type", "application/x-protobuf");
+        let agent_payload = construct_agent_payload(self.tracer_payloads.clone());
```
bantonsson, line 327: I can't figure out a way to avoid this `clone` 😢 The _raw_ mapping of `proto` definitions to `rust` structs make this messy.

bantonsson (commented):
Some general perf comments. I'll continue looking at it.

trace-utils/src/send_data.rs
```diff
@@ -151,79 +302,508 @@ impl SendData {
...
+                    <= retry_strategy.delay_ms
+                        + Duration::from_millis(RETRY_STRATEGY_TIME_TOLERANCE_MS),
+            "Elapsed time was not within expected range"
+        );
+
+        let start = Instant::now();
+        retry_strategy.delay(2).await;
+        let elapsed = start.elapsed();
+
+        // For the Exponential strategy, the delay for the second attempt should be double the
```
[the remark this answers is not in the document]
ekump: Good call!

trace-utils/src/send_data.rs
```diff
@@ -107,12 +110,99 @@ where
...
+    /// value is added to the delay.
+    ///
+    /// # Arguments
+    ///
+    /// * `attempt`: The number of the current attempt (1-indexed).
+    pub(crate) async fn delay(&self, attempt: u32) {
+        let delay = match self.backoff_type {
+            RetryBackoffType::Double => self.delay_ms * 2u32.pow(attempt - 1),
+            RetryBackoffType::Constant => self.delay_ms,
+            RetryBackoffType::Exponential => self.delay_ms * attempt,
```
bantonsson, line 187: `Exponential` seems linear to me and `Double` seems exponential 🤷🏼‍♂️

bantonsson (commented):
Thanks for the great tests. This looks good, but I'm confused by the `RetryBackoffType` naming.

bantonsson (approved):

PR #437, opened 2024-05-20
[profiling] Add fuzz test for StringTable

danielsn:
# What does this PR do?

Adds a fuzz test for the `StringTable` implementation in the profiler

# Motivation

Its "Reliability week"

# Additional Notes

It appears that `cargo test` runs a small number of tests.  We should add a `cargo bolero` test run to test this more effectively in CI.

# How to test the change?

This IS a test

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,3 +53,9 @@ opt-level = "s" # optimize for size
 
 [profile.release.package.datadog-serverless-trace-mini-agent]
 strip = true
+
+[profile.fuzz]
+inherits = "dev"
+opt-level = 3
+incremental = false
+codegen-units = 1
--- a/profiling/Cargo.toml
+++ b/profiling/Cargo.toml
@@ -44,6 +44,9 @@ serde_json = {version = "1.0"}
 tokio = {version = "1.23", features = ["rt", "macros"]}
 tokio-util = "0.7.1"
 byteorder = { version = "1.5", features = ["std"] }
+bolero-generator = "0.10.2"
 
 [dev-dependencies]
+bolero = "0.10.1"
 criterion = "0.5.1"
+
--- a/profiling/src/collections/string_table/mod.rs
+++ b/profiling/src/collections/string_table/mod.rs
@@ -202,6 +202,42 @@ impl IntoLendingIterator for StringTable {
 mod tests {
     use super::*;
 
+    /// This is a fuzz test for the allocation optimized `StringTable`.
+    /// It checks both safety (lack of crashes / sanitizer failures),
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -53,3 +53,9 @@ opt-level = "s" # optimize for size
 
 [profile.release.package.datadog-serverless-trace-mini-agent]
 strip = true
+
+[profile.fuzz]
+inherits = "dev"
+opt-level = 3
+incremental = false
+codegen-units = 1
```
morrisonlevi, line 61: Did you craft these yourself or follow guidelines from somewhere? Basically, can I read up somewhere on these choices?

danielsn: https://camshaft.github.io/bolero/library-installation.html  I'll add a comment

profiling/Cargo.toml
```diff
@@ -44,6 +44,9 @@ serde_json = {version = "1.0"}
 tokio = {version = "1.23", features = ["rt", "macros"]}
 tokio-util = "0.7.1"
 byteorder = { version = "1.5", features = ["std"] }
+bolero-generator = "0.10.2"
```
taegyunkim, line 47: Should this be in dev-dependencies?

profiling/src/collections/string_table/mod.rs
```diff
@@ -202,6 +207,78 @@ impl IntoLendingIterator for StringTable {
...
+            });
+    }
+
+    /// This is a fuzz test for the allocation optimized `StringTable`.
+    /// It checks both safety (lack of crashes / sanitizer failures),
+    /// as well as functional correctness (the table should behave like an
+    /// ordered set).  
+    /// Limitations:
+    ///   - The length of strings appears to be limited: `assert!(string.len() < 50);` failed, but
+    ///     `100` passed (in 10s)
```
morrisonlevi, line 242: I didn't quite understand this comment. Could you elaborate?

taegyunkim: Bolero, the crate used here to generate `String`s internally has a default range for the length of a string, (0..=64)
https://github.com/camshaft/bolero/blob/f401669697ffcbe7f34cbfd09fd57b93d5df734c/lib/bolero-generator/src/alloc/mod.rs#L17

If a string generated had length >= 50, then `assert!(string.len() < 50);` would have failed but not `100`.

morrisonlevi: Yes, I think that is a better way to explain it, thanks!

taegyunkim (approved):

PR #438, opened 2024-05-20
[profiling] Add fuzz test for internal::observation

taegyunkim:
# What does this PR do?

Fuzz test TrimmedObservation

# Motivation

Hack week

# Additional Notes

# How to test the change?

`cargo +nightly bolero test internal::observation::trimmed_observation::tests::fuzz_trimmed_observation -T 1min`


profiling/src/internal/observation/observations.rs
```diff
@@ -96,6 +88,62 @@ impl Observations {
...
+
+    fn add(&mut self, sample: Sample, values: Vec<i64>) -> anyhow::Result<()> {
+        if let Some(v) = self.data.get_mut(&sample) {
+            // SAFETY: This method is only way to build one of these, and at
+            // the top we already checked the length matches.
+            unsafe { v.as_mut_slice(self.obs_len) }
+                .iter_mut()
+                .zip(values)
+                // HACK: fix this
+                .for_each(|(a, b)| match a.checked_add(b) {
```
danielsn, line 113: Would saturating add make more sense here?

profiling/src/internal/observation/observations.rs
```diff
@@ -96,6 +88,62 @@ impl Observations {
...
+                });
+        } else {
+            let trimmed = TrimmedObservation::new(values, self.obs_len);
+            self.data.insert(sample, trimmed);
+        }
+
+        Ok(())
+    }
+
+    fn len(&self) -> usize {
```
danielsn, line 131: I think its normally best practice to have a `is_empty` if you have `len`

profiling/src/internal/observation/observations.rs
```diff
@@ -340,4 +378,78 @@ mod tests {
...
+            // TODO: Figure out sane limits for these numbers. We don't simply want to go up to
+            // usize::MAX as that would result in crashes with too large Vec allocations.
+            .with_generator(((1..=1024usize), (1..1024usize), (1..1024usize)))
+            .and_then(|(observations_len, num_ts_samples, num_samples)| {
+                let ts_samples = Vec::<(Sample, Timestamp, Vec<i64>)>::gen()
+                    .with()
+                    .values((
+                        Sample::gen(),
+                        Timestamp::gen(),
+                        Vec::<i64>::gen().with().len(observations_len),
```
danielsn, line 398: This is only going to fuzz with correct length samples, correct?

profiling/src/internal/stack_trace.rs
```diff
@@ -16,6 +16,7 @@ impl Item for StackTrace {
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 #[repr(transparent)]
+#[cfg_attr(test, derive(bolero_generator::TypeGenerator))]
```
danielsn, line 19: Wondering about this.  This is an opaque type that represents an interned index in the profile.  If we ever want to use it as such, we would need a constraint connecting the two.  But maybe for fuzzing for now this is ok?

profiling/src/internal/observation/observations.rs
danielsn: I think it would be worth having two tests: one that leaves the lengths arbitrary, one that only looks at correct lengths

profiling/src/internal/observation/observations.rs
```diff
@@ -340,4 +378,78 @@ mod tests {
...
+                for (sample, ts, values) in iter {
+                    assert!(ts.is_none());
+                    assert!(aggregated_observations.data.contains_key(&sample));
+                    let expected_values = aggregated_observations.data.remove(&sample).unwrap();
+                    unsafe {
+                        let b = expected_values.into_vec(obs_len);
+                        assert_eq!(*b, values);
+                    }
+                }
+                assert!(aggregated_observations.data.is_empty());
```
danielsn, line 452: this should be exposed at the `aggregated_observations` level

profiling/src/internal/observation/trimmed_observation.rs
```diff
@@ -183,4 +183,33 @@ mod tests {
             assert_eq!(*b, vec![1, 2]);
         }
     }
+
+    // cargo +nightly bolero test
+    // internal::observation::trimmed_observation::tests::fuzz_trimmed_observation -T 1min
```
danielsn, line 188: Not sure if we need this as a comment.  But a block comment explaining what you do and don't fuzz over might be nice

profiling/src/internal/observation/trimmed_observation.rs
```diff
@@ -183,4 +183,33 @@ mod tests {
...
+                }
+            }
+            {
+                let mut t = TrimmedObservation::new(v.clone(), o);
+                unsafe {
+                    assert_eq!(&t.as_mut_slice(o), &v);
+                    assert_eq!(t.into_boxed_slice(o).as_ref(), v.as_slice());
+                }
+            }
+            {
```
danielsn, line 206: Not immediatly obvious how this differs from the above example

profiling/src/internal/observation/trimmed_observation.rs
```diff
@@ -183,4 +183,33 @@ mod tests {
...
+                let t = TrimmedObservation::new(v.clone(), o);
+                unsafe {
+                    t.consume(o);
+                }
+            }
+            {
+                let mut t = TrimmedObservation::new(v.clone(), o);
+                unsafe {
+                    assert_eq!(&t.as_mut_slice(o), &v);
```
danielsn, line 202: would be good to test that mutation works as expected

profiling/src/internal/observation/trimmed_observation.rs
```diff
@@ -183,4 +183,33 @@ mod tests {
...
     }
+
+    // cargo +nightly bolero test
+    // internal::observation::trimmed_observation::tests::fuzz_trimmed_observation -T 1min
+    #[test]
+    fn fuzz_trimmed_observation() {
+        bolero::check!().with_type::<Vec<i64>>().for_each(|v| {
+            let o = ObservationLength::new(v.len());
+            {
+                let t = TrimmedObservation::new(v.clone(), o);
```
danielsn, line 194: One other check we can do is that the length of the trimmed observation is the same as o

profiling/src/internal/observation/observations.rs
taegyunkim: Updated this one so that sample lengths are arbitrary, and check for the counts appropriately. Also, this now checks samples with incorrect lengths results in error when called `add()`

profiling/src/internal/stack_trace.rs
danielsn: Fair

profiling/src/internal/observation/observations.rs
```diff
@@ -340,4 +383,82 @@ mod tests {
...
+        // The body of this test simply adds these samples to the Observations and then uses the
+        // iterator to check that the samples are the same as added.
+        bolero::check!()
+            // TODO: Figure out sane limits for these numbers. We don't simply want to go up to
+            // usize::MAX as that would result in crashes with too large Vec allocations.
+            .with_generator(((1..=1024usize), (1..1024usize), (1..1024usize)))
+            .and_then(|(observations_len, num_ts_samples, num_samples)| {
+                let ts_samples = Vec::<(Sample, Timestamp, Vec<i64>)>::gen()
+                    .with()
+                    .values((Sample::gen(), Timestamp::gen(), Vec::<i64>::gen()))
```
danielsn, line 400: Depending on how smart the fuzzer is, it might be worth having one test with malformed sampled (e.g. different lengths) and a separate one where we constrain the samples to be of the right length, to get more coverage

profiling/src/internal/observation/observations.rs
```diff
@@ -96,6 +88,67 @@ impl Observations {
...
+        self.data.remove(sample)
+    }
+}
+
+impl Drop for AggregatedObservations {
+    fn drop(&mut self) {
+        let o = self.obs_len;
+        self.data.drain().for_each(|(_, v)| {
+            // SAFETY: The only way to build one of these is through
+            // [Self::add], which already checked that the length was correct.
```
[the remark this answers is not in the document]
taegyunkim: Added a check in `Self::add`

danielsn (approved):

PR #440, opened 2024-05-21
Allow the SidecarTransport to recreate its underlying transport

iamluc:
# What does this PR do?

Replace the `SidecarTransport` type alias by a real struct that handles reconnection.

This is handled in the Rust code so that sidecar's clients (like the PHP tracer) can use the SidecarTransport in a thread safe way.

# Motivation

In the PHP tracer, the reconnection was previously done in the C code, but was not thread safe.

# Additional Notes

# How to test the change?

### DIFF
--- a/sidecar-ffi/src/lib.rs
+++ b/sidecar-ffi/src/lib.rs
@@ -258,6 +258,7 @@ pub unsafe extern "C" fn ddog_sidecar_runtimeMeta_drop(meta: Box<RuntimeMetadata
     drop(meta)
 }
 
+/// Reports the runtime configuration to the telemetry.
 #[no_mangle]
 #[allow(clippy::missing_safety_doc)]
 pub unsafe extern "C" fn ddog_sidecar_telemetry_enqueueConfig(
@@ -282,6 +283,7 @@ pub unsafe extern "C" fn ddog_sidecar_telemetry_enqueueConfig(
     MaybeError::None
 }
 
+/// Reports a dependency to the telemetry.
 #[no_mangle]
 #[allow(clippy::missing_safety_doc)]
 pub unsafe extern "C" fn ddog_sidecar_telemetry_addDependency(
@@ -309,6 +311,7 @@ pub unsafe extern "C" fn ddog_sidecar_telemetry_addDependency(
     MaybeError::None
 }
 
+/// Reports an integration to the telemetry.
 #[no_mangle]
 #[allow(clippy::missing_safety_doc)]
 pub unsafe extern "C" fn ddog_sidecar_telemetry_addIntegration(
@@ -340,6 +343,7 @@ pub unsafe extern "C" fn ddog_sidecar_telemetry_addIntegration(
     MaybeError::None
 }
 
+/// Registers a service and flushes any queued actions.
 #[no_mangle]
 #[allow(clippy::missing_safety_doc)]
 pub unsafe extern "C" fn ddog_sidecar_telemetry_flushServiceData(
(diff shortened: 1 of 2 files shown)


sidecar-ffi/src/lib.rs
```diff
@@ -642,3 +642,12 @@ pub unsafe extern "C" fn ddog_sidecar_dogstatsd_set(
 
     MaybeError::None
 }
+
+#[no_mangle]
+#[allow(clippy::missing_safety_doc)]
```
pierotibou, line 647: I believe we should document the propose and arguments of all public ffi APIs.

sidecar/src/service/blocking.rs
```diff
@@ -6,22 +6,92 @@ use super::{
...
 /// This transport is used for communication between different parts of the sidecar service.
 /// It is a blocking transport, meaning that it will block the current thread until the operation is
 /// complete.
-pub type SidecarTransport = BlockingTransport<SidecarInterfaceResponse, SidecarInterfaceRequest>;
+pub struct SidecarTransport {
+    pub inner: Mutex<BlockingTransport<SidecarInterfaceResponse, SidecarInterfaceRequest>>,
+}
+
+impl SidecarTransport {
+    pub fn reconnect<F>(&mut self, factory: F)
```
pierotibou, line 31: Could you add utests for all those fonctions?

pierotibou (commented):
Too complicated for me ;) but a few comments around testing and comments

sidecar/src/service/blocking.rs
```diff
@@ -6,22 +6,92 @@ use super::{
...
+        match self.inner.lock() {
+            Ok(mut t) => t.set_write_timeout(timeout),
+            Err(e) => Err(io::Error::new(io::ErrorKind::Other, e.to_string())),
+        }
+    }
+
+    pub fn is_closed(&self) -> bool {
+        match self.inner.lock() {
+            Ok(t) => t.is_closed(),
+            Err(_) => true, // Well... what can we do?
```
ekump, line 69: I like that we're not just unwrapping and potentially panicking on the lock acquisitions, but I'm not sure what here is safe? I believe the majority of the time we'd get an error on a lock is if the lock is poisoned, which means we might be in a bad state? is it ok to assume that if self.inner is poisoned that the connection is closed and we should move forward with trying to reconnect?

bwoebi: The scenario where this will happen is when the reconnect callback executes functions on the sidecar instance (only location which calls back to caller from within the lock). And ... during the reconnect the original sidecar instance is always closed (otherwise the reconnect wouldn't happen in the first place), so actually `true` is the right value to report there.

sidecar/src/service/blocking.rs
iamluc: Added some tests, but I'm not really happy with them as they are really coupled to the BlockingTransport.

Any advice is welcome!

sidecar-ffi/src/lib.rs
iamluc: 👍 
I added a few comments. I suggest to continue in separate PRs.

sidecar/src/service/blocking.rs
iamluc: Comment updated 😛

sidecar-ffi/src/lib.rs
pierotibou: Very nice. Thanks for adding comments to things you didn't change :)

PR #441, opened 2024-05-21
Refactor sidecar trace_flusher and add basic tests before adding retry logic

ekump:
# What does this PR do?

This is a minor refactor the sidecar's trace_flusher. It
1. Adds unit tests and leverages httpmock to mock calling the trace agent. It would be more appropriate for the call to trace_utils::send_data.send to be mocked in this unit test, but introducing such mocking would be a bigger lift than we'd like to take on right now. We may revisit this in the future. 
2. Introduces sane defaults for the TraceFlusher. The previous values all defaulted to 0. The new values were taken from [here](https://github.com/DataDog/libdatadog/blob/main/sidecar-ffi/tests/sidecar.rs#L128).
3. Breaks up the large `start_trace_flusher` function in to a couple of smaller functions. 

No functional changes.

# Motivation

This was originally part of APMSP-1020 to add retry logic for the sidecar. It was broken out into a separate PR because there is concurrent work being done to introduce trace_api.* metrics that will touch this file and it will be easier to handle the merge conflicts now rather than later.

# Additional Notes

# How to test the change?

New unit tests have been added


sidecar/src/service/tracing/trace_flusher.rs
```diff
@@ -78,16 +82,26 @@ impl TraceFlusherMetrics {
...
-    pub(crate) min_force_flush_size: AtomicU32,
-    pub(crate) min_force_drop_size: AtomicU32, // put a limit on memory usage
+    pub(crate) interval_ms: AtomicU64,
+    pub(crate) min_force_flush_size_bytes: AtomicU32,
+    pub(crate) min_force_drop_size_bytes: AtomicU32, // put a limit on memory usage
     remote_config: Mutex<AgentRemoteConfigs>,
     pub metrics: Mutex<TraceFlusherMetrics>,
 }
-
+impl Default for TraceFlusher {
```
ekump, line 93: Previously, everything defaulted to 0 which made it unusable. There is no point in having defaults if you can't use them. These values are are a best guess from existing tests. In reality, the sidecar-ffi code is setting the values based on the config, so this shouldn't impact existing usage.

sidecar/src/service/tracing/trace_flusher.rs
```diff
@@ -200,12 +214,63 @@ impl TraceFlusher {
         }
     }
 
+    fn replace_trace_send_data(&self, completer: ManualFutureCompleter<()>) -> Vec<SendData> {
```
ekump, line 217: No logic changes here, just breaking up the `start_trace_flusher` function into smaller functions.

sidecar/src/service/tracing/trace_flusher.rs
```diff
@@ -268,3 +296,169 @@ impl TraceFlusher {
...
+    use datadog_trace_protobuf::pb;
+    use datadog_trace_utils::trace_utils::TracerHeaderTags;
+    use httpmock::{Mock, MockServer};
+    use std::sync::Arc;
+
+    // This function will poll the mock server for "hits" until the expected number of hits is
+    // observed. In its current form it may not correctly report if more than the asserted number of
+    // hits occurred. More attempts at lower sleep intervals is preferred to reduce flakiness and
+    // test runtime.
+    async fn poll_for_mock_hit(
```
ekump, line 312: Sleeping during tests are not ideal, but somewhat necessary with the given implementation of the flusher. The public `enqueue` method doesn't ever block on flushing. This function attempts to minimize the badness of sleeping by sleeping for shorter durations, but polling frequently to see if our mocks have been hit.

hoolioh: Tried a few ideas but none of them worked without tweaking the flusher. Given the fact that the future inside the flusher is completed in another thread I guess we don't have other choice to just calling await while in the loop.

hoolioh (approved):

bwoebi (approved):
Thanks for the tests :-)

PR #443, opened 2024-05-21
[profiling] Fuzz chain allocator

danielsn:
# What does this PR do?

Adds fuzz tests for the chain allocator. Fixes an issue found by the fuzz tester https://github.com/DataDog/libdatadog/pull/443/files#r1609080275

# Motivation

R&D week

# Additional Notes

# How to test the change?

This IS A TEST


alloc/src/chain.rs
```diff
@@ -238,10 +238,7 @@ unsafe impl<A: Allocator + Clone> Allocator for ChainAllocator<A> {
 
         debug_assert!(chain_node.remaining_capacity() >= layout.size());
 
-        let result = chain_node.linear.allocate(layout);
-        // If this fails, there's a bug in the allocator.
-        debug_assert!(result.is_ok());
-        result
```
morrisonlevi, line 244: Was there a specific reason to get rid of this?

taegyunkim (changes requested):

alloc/src/chain.rs
```diff
@@ -290,6 +287,62 @@ mod tests {
...
+    /// https://doc.rust-lang.org/beta/std/primitive.pointer.html#method.is_aligned_to
+    /// Convenience function until the std lib standardizes this.
+    fn is_aligned_to<T: ?Sized>(p: *const T, align: usize) -> bool {
+        (p as *const u8 as usize) % align == 0
+    }
+
+    #[test]
+    fn fuzz() {
+        bolero::check!()
+            .with_type::<(usize, Vec<(usize, u32, usize, u8)>)>()
```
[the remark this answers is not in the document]
danielsn: Done

alloc/src/chain.rs
danielsn: It was causing the fuzzer to spuriously fail.  Do you know of a way to control whether this assert appears using `cfg`?

danielsn: found it

alloc/src/chain.rs
```diff
@@ -239,6 +239,7 @@ unsafe impl<A: Allocator + Clone> Allocator for ChainAllocator<A> {
         debug_assert!(chain_node.remaining_capacity() >= layout.size());
 
         let result = chain_node.linear.allocate(layout);
+        #[cfg(not(fuzzing))]
         // If this fails, there's a bug in the allocator.
         debug_assert!(result.is_ok());
```
morrisonlevi, line 244: I've been thinking about this more, and I think the comment is accurate. The issue is the capacity checks that the chain allocator does do not account for alignment, but the linear allocator does. So the chain thinks there is capacity, and the linear says there isn't. There's the problem.
Of course, for our string table, this is always align=1 so there's no need for alarm. But maybe we should make a method for the linear allocator that is like `has_capacity_for(layout: Layout)` that the chain allocator can call?

alloc/src/utils.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+/// https://doc.rust-lang.org/beta/std/primitive.pointer.html#method.is_aligned_to
```
taegyunkim, line 4: the std lib impl panics when align is not a power of two. And we don't want that right?

alloc/src/chain.rs
```diff
@@ -239,6 +239,7 @@ unsafe impl<A: Allocator + Clone> Allocator for ChainAllocator<A> {
         debug_assert!(chain_node.remaining_capacity() >= layout.size());
 
         let result = chain_node.linear.allocate(layout);
+        #[cfg(not(fuzzing))]
```
taegyunkim, line 242: Does `cargo-bolero` or sets `fuzzing` when it's invoking the tests?

alloc/src/utils.rs
morrisonlevi: I think we do. It derives from alignments being a power of two in `Layout`, and since you need a `Layout` to make an allocation in the first place, how did we get to a not-power-of-2?

alloc/src/chain.rs
danielsn: appears to

taegyunkim (approved):

PR #446, opened 2024-05-22
[profiling] Fuzz virtual allocator

danielsn:
# What does this PR do?

Fuzzes the virtual allocator.

# Motivation

R&D week

# Additional Notes

`virtual` is a reserved name, internally renamed to `r#virtual` and bolero was not happy running a test with `#` in the name.  Moved the file to fix that.

# How to test the change?


alloc/src/virtual_alloc.rs
```diff
@@ -183,8 +183,37 @@ pub mod os {
...
     use super::*;
+    use crate::utils::*;
     use allocator_api2::alloc::Allocator;
 
+    #[test]
+    fn fuzz() {
+        use bolero::TypeGenerator;
+
+        #[cfg(miri)]
+        const MAX_SIZE: usize = 1_000_000;
```
morrisonlevi, line 194: Why is miri different here? Too slow on bigger sizes?

danielsn: Miri had a memout if we used too large a size

morrisonlevi (approved):

PR #448, opened 2024-05-23
Improve container id computation

VianneyRuhlmann:
- **Implement entity-id and cgroup inode computation**
- **Add entity-id headers to Endpoint**
- **Add custom error type**

# What does this PR do?

Implements [this RFC](https://datadoghq.atlassian.net/wiki/spaces/CONT/pages/3230337019/Support+origin+detection+on+CgroupV2+without+UDS) by adding:
- `get_cgroup_inode` to compute the inode of a cgroup node
- `get_entity_id` to compute the entity id using either the container id or the inode as a fallback
- `datadog-entity-id` as a header in `Endpoint`
- `CgroupFileParsingError` as a new error type for `/proc/self/cgroup` parsing

It tries to be compliant with the implementation done in [dd-trace-go](https://github.com/DataDog/dd-trace-go/blob/main/internal/container_linux.go#L150).

# Motivation

This PR allows tagging of the container id when using cgroupv2 by using the cgroup node inode as a fallback.

# Additional Notes

Please feel free to give feedback especially about the `CgroupFileParsingError` error type and the `get_cgroup_node_path` function which parses `/proc/self/cgroup.

I am also not curious if you think that the entity id computation should stay in `container-id.rs` or be moved to a new module.

# How to test the change?

TODO


ddcommon/src/container_id.rs
```diff
@@ -113,6 +178,33 @@ pub fn get_container_id() -> Option<&'static str> {
     CONTAINER_ID.as_deref()
 }
 
+/// Returns the `cgroup_inode` if available, otherwise `None`
+pub fn get_cgroup_inode() -> Option<&'static str> {
```
pierotibou, line 182: This one doesn't need to be public. 
Also, I don't know if in Rust we tend to put public methods at the top and private below. But if you have followed libdd format, that's fine by me.

ddcommon/src/container_id.rs
pierotibou, line 1: Not saying this isn't good but we seem to have done a bit more things in .NET.
Didn't have the time to go through the whole diff but here's [Zach PR](https://github.com/DataDog/dd-trace-dotnet/commit/b314b9c52f709f922fd60a1feb46652eeeff7c10) (you could look at the current code instead it has changed). In this PR, there's at least a check around `/proc/self/ns/cgroup` that we don't have here.

pierotibou: When ready, you could ask the container team for a review just in case.

pierotibou (commented):
Congrats on the first PR, even if only a draft for now!
Added a few comments.

ddcommon/src/container_id.rs
VianneyRuhlmann: You're right I fixed that. I don't know about any convention in Rust so I put them next to the existing ones.

ddcommon/src/container_id.rs
VianneyRuhlmann: I was indeed missing a check to avoid getting an entity-id when running on host and not in a container.

ddcommon/src/entity_id/mod.rs
```diff
@@ -0,0 +1,245 @@
+// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Extract the entity id and container id
+//!
+//! The container id can be extracted from `/proc/self/group`
+//!
+//! The entity id is either:
+//! - `cid:<container id>` if available
```
AliDatadog, line 9: Isn't in `cid-<container_id>` or `in-<container-id>` ?

bantonsson (approved):
Great work. Thanks for the extensive documentation/testing.

ddcommon/tests/cgroup.multiple_controllers
```diff
@@ -0,0 +1,10 @@
+10:blkio:/user.slice/user-0.slice/session-14.scope
```
bantonsson, line 1: _Nitpick:_ I would prefer that these files were in `resource/test` instead of the standard integration test directory.

ddcommon/src/entity_id/unix/cgroup_inode.rs
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+    }
+    node_path.ok_or(CgroupFileParsingError::CgroupNotFound)
+}
+
+#[cfg(not(test))]
+/// Checks if the agent is running in the host cgroup namespace.
+/// This check is disabled when testing
+fn is_host_cgroup_namespace() -> Result<(), ()> {
```
AliDatadog, line 71: Perhaps worth renaming `is_not_host_cgroup_namespace` for readability.

ddcommon/src/entity_id/unix/cgroup_inode.rs
```diff
@@ -0,0 +1,143 @@
...
+) -> Result<PathBuf, CgroupFileParsingError> {
+    let file = File::open(cgroup_path).map_err(|_| CgroupFileParsingError::CannotOpenFile)?;
+    let reader = BufReader::new(file);
+
+    let mut node_path: Option<PathBuf> = None;
+
+    for (index, line) in reader.lines().enumerate() {
+        let line_content = &line.map_err(|_| CgroupFileParsingError::InvalidFormat)?;
+        let cgroup_entry: Vec<&str> = line_content.split(':').collect();
+        if cgroup_entry.len() != 3 {
```
AliDatadog, line 43: IMO here we can just skip the line

ddcommon/src/entity_id/unix/cgroup_inode.rs
```diff
@@ -0,0 +1,143 @@
...
+/// Returns the inode of file at `path`
+fn get_inode(path: &Path) -> io::Result<u64> {
+    let meta = fs::metadata(path)?;
+    Ok(meta.ino())
+}
+
+/// Returns the cgroup mount path associated with `base_controller` or the default one for
+/// cgroupV2
+fn get_cgroup_node_path(
+    base_controller: &str,
```
AliDatadog, line 31: IMO it's worth renaming (and propagating) the variable as `cgroup_v1_base_controller`

ddcommon/src/entity_id/unix/cgroup_inode.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    for (index, line) in reader.lines().enumerate() {
+        let line_content = &line.map_err(|_| CgroupFileParsingError::InvalidFormat)?;
+        let cgroup_entry: Vec<&str> = line_content.split(':').collect();
+        if cgroup_entry.len() != 3 {
+            return Err(CgroupFileParsingError::InvalidFormat);
+        }
+        let controllers: Vec<&str> = cgroup_entry[1].split(',').collect();
+        // Only keep empty controller if it is the first line as cgroupV2 uses only one line
+        if controllers.contains(&base_controller) || (controllers.contains(&"") && index == 0) {
```
AliDatadog, line 48: It's adding the assumption that the cgroup line will always have index 0 which should hold in practice but the Agent and other implementations don't have this assumption. IMO it's better to harmonize just in case the behavior ever changes.

ddtelemetry/src/lib.rs
```diff
@@ -15,7 +15,7 @@ pub mod worker;
 pub fn build_host() -> data::Host {
     data::Host {
         hostname: info::os::real_hostname().unwrap_or_else(|_| String::from("unknown_hostname")),
-        container_id: container_id::get_container_id().map(|f| f.to_string()),
+        container_id: entity_id::get_container_id().map(|f| f.to_string()),
```
AliDatadog, line 18: Not sure how this field is used

ddcommon/src/entity_id/unix/cgroup_inode.rs
bantonsson: I think it's perfectly readable this way, and how it's being used in the code.

ddcommon/tests/cgroup.multiple_controllers
bantonsson: There is no rust/cargo convention for resources, but there is a convention for integration tests being in the `tests` directory. I’m not sure how we want to organize this, so this is fine for now.

paullegranddc (approved):

PR #449, opened 2024-05-23
[profiling] Fuzz profiler API

danielsn:
# What does this PR do?

Adds fuzzable versions of the data structures, and uses them to fuzz the profiler API.
 
# Motivation

R&D week

# Additional Notes

# How to test the change?


profiling/src/internal/profile.rs
```diff
@@ -544,6 +544,188 @@ impl Profile {
...
+        bolero::check!()
+            .with_generator((
+                Vec::<owned_types::ValueType>::gen(),
+                Vec::<(Option<Timestamp>, owned_types::Sample)>::gen(),
+            ))
+            .for_each(|(val, samples)| {
+                let expected_sample_types: Vec<_> = val.iter().map(api::ValueType::from).collect();
+                let mut expected_profile =
+                    Profile::new(SystemTime::now(), &expected_sample_types, None);
+                let mut expected_samples: Vec<&owned_types::Sample> = Vec::new();
```
danielsn, line 567: This will have an issue with aggregated samples

taegyunkim: Yes, updated the code to keep track of timestamped and non-timestamped samples.

profiling/src/internal/profile/mod.rs
```diff
@@ -544,6 +547,7 @@ impl Profile {
 #[cfg(test)]
 mod api_tests {
     use super::*;
+    use std::collections::HashMap;
```
danielsn, line 550: not needed anymore

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,597 @@
...
+use super::*;
+use crate::collections::identifiable::FxIndexMap;
+use crate::internal::{owned_types, Profile, Timestamp};
+use crate::pprof;
+use bolero::TypeGenerator;
+use bolero_generator::{TypeGeneratorWithParams, ValueGenerator};
+use std::collections::{HashMap, HashSet};
+use std::time::SystemTime;
+
+#[derive(Clone, Debug, Default, Eq, PartialEq, Hash, bolero_generator::TypeGenerator)]
```
danielsn, line 13: `TypeGenerator` type is imported, can be shortened

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,597 @@
...
+            let key = (&owned_locations, &owned_labels);
+            let expected_values = samples_without_timestamps
+                .get(&key)
+                .expect("Value not found for an aggregated sample");
+            assert_eq!(&sample.values, expected_values);
+        }
+    }
+}
+
+fn add_sample<'a>(
```
danielsn, line 365: rename to something like `fuzz_add_sample_inner`

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+    fn from(value: &'a Sample) -> Self {
+        Self {
+            locations: value.locations.iter().map(api::Location::from).collect(),
+            values: value.values.clone(),
+            labels: value.labels.iter().map(api::Label::from).collect(),
+        }
+    }
+}
+
+/// PartialEq and Hash work on just the key to avoid getting duplicate
```
danielsn, line 214: Are we using `Eq` to compare labels in the tests? Will this miss issues where the values don't agree?

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+        self.key.hash(state)
+    }
+}
+impl TypeGenerator for Sample {
+    fn generate<D: bolero_generator::Driver>(driver: &mut D) -> Option<Self> {
+        let locations = Vec::<Location>::gen_with().generate(driver)?;
+        let values = Vec::<i64>::gen_with().generate(driver)?;
+        let mut labels = std::collections::HashSet::<Label>::gen_with().generate(driver)?;
+
+        // Ensure that the label has "local root span id" key
```
danielsn, line 237: Why do we need to ensure this? Isn't it OK for a sample to not have a "local root span id"

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+        "Samples length mismatch"
+    );
+
+    let mut expected_timestamped_samples = samples_with_timestamps.iter();
+
+    for sample in profile.samples.iter() {
+        // Recreate owned_locations from vector of pprof::Location
+        let mut owned_locations = Vec::new();
+        for loc_id in sample.location_ids.iter() {
+            let location = &profile.locations[*loc_id as usize - 1];
```
danielsn, line 297: comment explaining the `-1`

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+            let mapping = &profile.mappings[location.mapping_id as usize - 1];
+            assert!(location.lines.len() == 1);
```
danielsn, line 299: why?

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+            let line = location.lines[0];
+            let function = profile.functions[line.function_id as usize - 1];
+            assert!(!location.is_folded);
```
danielsn, line 302: why?

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+
+            let owned_mapping = Mapping::new(
```
danielsn, line 304: Its probably fine given that this is a test, but you could probably use the `&str` from the string table and make an `api::` mapping, saving allocations.

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+
+            if *key == *"end_timestamp_ns" {
+                // TODO: Check end timestamp label
+                continue;
+            } else if *key == *"trace endpoint" {
+                let actual_str = profile.string_table_fetch(label.str);
+
+                let prev_label: &Label = owned_labels
+                    .last()
+                    .expect("Previous label to exist for endpoint label");
```
danielsn, line 354: Is the order of labels mandated?

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+                    num,
+                    num_unit,
+                });
+            }
+        }
+
+        if let Some(expected_sample) = expected_timestamped_samples.next() {
+            assert_eq!(owned_locations, expected_sample.locations);
+            assert_eq!(sample.values, expected_sample.values);
+            assert_eq!(owned_labels, expected_sample.labels);
```
danielsn, line 396: These should potentially be sorted: there is no requirement that labels remain in the same order

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+    {
+        assert_eq!(*profile.string_table_fetch(typ.r#type), *expected_typ.r#typ);
+        assert_eq!(*profile.string_table_fetch(typ.unit), *expected_typ.unit);
+    }
+}
+
+fn assert_samples_eq(
+    profile: &pprof::Profile,
+    samples_with_timestamps: &[&Sample],
+    samples_without_timestamps: &HashMap<(&Vec<Location>, &Vec<Label>), Vec<i64>>,
```
danielsn, line 282: `&[]` instead of `&Vec`

profiling/src/internal/profile/fuzz_tests.rs
```diff
@@ -0,0 +1,631 @@
...
+                profile
+                    .string_table_fetch(mapping.build_id)
+                    .clone()
+                    .into_boxed_str(),
+            );
+            let owned_function = Function::new(
+                profile
+                    .string_table_fetch(function.name)
+                    .clone()
+                    .into_boxed_str(),
```
danielsn, line 321: This is repeated often enough to justify a `string_table_fetch_owned` that returns a `Box<str>`

profiling/src/internal/profile/fuzz_tests.rs
morrisonlevi: Locally I'm sorting the labels and got a new failure:

```
======================== Test Failure ========================

Input: 
(
    [],
    [
        (
            None,
            Sample {
                locations: [],
                values: [],
                labels: [
                    Label {
                        key: "local root span id",
                        str: None,
                        num: 281474976710656,
                        num_unit: None,
                    },
                    Label {
                        key: "",
                        str: Some(
                            "",
                        ),
                        num: 0,
                        num_unit: Some(
                            "",
                        ),
                    },
                ],
            },
        ),
    ],
)

Error: 
panicked at profiling/src/internal/profile/fuzz_tests.rs:402:18:
Value not found for an aggregated sample
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

==============================================================
```

Line 402 is this expect:

```rs
            let key = (&owned_locations, &owned_labels);
            let expected_values = samples_without_timestamps
                .get(&key)
                .expect("Value not found for an aggregated sample");
            assert_eq!(&sample.values, expected_values);
```

I think this happens because `Vec<Label>` and such are used as keys in the hash, and then they don't work out because of order?

morrisonlevi: I pushed this up as `fuzz_failure_001`.

profiling/src/internal/profile/fuzz_tests.rs
taegyunkim: You're right, it's OK for a sample to not have it. However, I wasn't sure whether codes that deal with this logic would be covered?

profiling/src/internal/profile/fuzz_tests.rs
taegyunkim: `internal::Location` to `pprof::Location` impl here
https://github.com/DataDog/libdatadog/blob/d43e6bccffecb10fb58163038d67e7235f5a26ac/profiling/src/internal/location.rs#L35

profiling/src/internal/profile/fuzz_tests.rs
taegyunkim: https://github.com/DataDog/libdatadog/blob/dsn/fuzz-profile-api/profiling/src/internal/location.rs#L31-L34

profiling/src/internal/profile/fuzz_tests.rs
danielsn: Take a boolean and add it iff the boolean is true?

profiling/src/internal/profile/fuzz_tests.rs
taegyunkim: `[internal::Location]` is created via its `from_offset()` function and the function uses `small_non_zero_pprof_id()`.

https://github.com/DataDog/libdatadog/blob/f3d0382d1c7c43dde7e22d3b27aad270d146bcee/profiling/src/internal/location.rs#L48

The function's documentation says 
```
/// Creates a non-zero, 32-bit unsigned id from the offset. It's guaranteed to
/// be the offset + 1, with guards to not overflow the size of u32.
///
/// This is useful because many pprof collections do not allow an item with an
/// id of zero, even if it's the first item in the collection.
```

r1viollet (approved):
LGTM
I don't know much about what a bolero_generator is, but the changes look fine :+1:

r1viollet:
I am guessing we have some amount of work to fix the tests before we can merge this ?
```
test internal::profile::fuzz_tests::fuzz_failure_001 ... FAILED
```

profiling/src/internal/profile/fuzz_tests.rs
taegyunkim: That's not possible as this is an implementation of a trait function.
https://docs.rs/bolero-generator/latest/bolero_generator/trait.TypeGenerator.html#tymethod.generate

I can try modifying what's generated by bolero using [mutate](https://docs.rs/bolero-generator/latest/bolero_generator/trait.TypeGenerator.html#method.mutate)?

taegyunkim:
> I am guessing we have some amount of work to fix the tests before we can merge this ?
> 
> ```
> test internal::profile::fuzz_tests::fuzz_failure_001 ... FAILED
> ```

Yes, looking into this...

profiling/src/internal/profile/fuzz_tests.rs
taegyunkim: Also, `validate_sample_labels` doesn't allow us to add a label with `local root span id` key. Just deleting, this whole custom impl of `TypeGenerator`
 https://github.com/DataDog/libdatadog/blob/6f37a289241f85b7d379d5b8767925761dce09e4/profiling/src/internal/profile/mod.rs#L88

profiling/src/internal/profile/fuzz_tests.rs
taegyunkim: Left a comment

profiling/src/internal/profile/fuzz_tests.rs
taegyunkim: `std::hash::Hash` doc says this property must hold when implementing `Eq` and `Hash`, `k1 == k2 -> hash(k1) == hash(k2)`. In our case, values can disagree even if the hashes are the same. 

This was introduced mainly to use Labels as keys for HashMap in https://github.com/DataDog/libdatadog/blob/dsn/fuzz-profile-api/profiling/src/internal/profile/fuzz_tests.rs#L369

PR #450, opened 2024-05-24
Add metrics

hoolioh:
# What does this PR do?

Adds support for sending the following metrics:

- trace_api.bytes
- trace_chunks_sent
- trace_chunks_dropped

Aside from the introduction of those new metrics a small refactor has been done on send_data.rs in order to address some technical debt from previous PRs. This changes can be summarized in:

- Limit visibility of `SendData` fields.
- Add default trait implementation for `SendDataResult`.
- Add tests for both `TraceHeaderTags` and `SendData`.

# Motivation

In order to turn the sidecar on by default for PHP there is the need to support those metrics.

### DIFF
--- a/sidecar/src/self_telemetry.rs
+++ b/sidecar/src/self_telemetry.rs
@@ -28,6 +28,9 @@ struct MetricData<'a> {
     trace_api_requests: ContextKey,
     trace_api_responses: ContextKey,
     trace_api_errors: ContextKey,
+    trace_api_bytes: ContextKey,
+    trace_chunk_sent: ContextKey,
+    trace_chunk_dropped: ContextKey,
 }
 impl<'a> MetricData<'a> {
     async fn send(&self, key: ContextKey, value: f64, tags: Vec<Tag>) {
@@ -69,7 +72,6 @@ impl<'a> MetricData<'a> {
                 vec![Tag::new("level", level.as_str().to_lowercase()).unwrap()],
             ));
         }
-
         if trace_metrics.api_requests > 0 {
             futures.push(self.send(
                 self.trace_api_requests,
@@ -96,6 +98,27 @@ impl<'a> MetricData<'a> {
                 self.trace_api_errors,
                 trace_metrics.api_errors_status_code as f64,
                 vec![tag!("type", "status_code")],
+            ));
+        }
+        if trace_metrics.bytes_sent > 0 {
+            futures.push(self.send(
+                self.trace_api_bytes,
+                trace_metrics.bytes_sent as f64,
+                vec![],
+            ));
+        }
+        if trace_metrics.chunks_sent > 0 {
+            futures.push(self.send(
+                self.trace_chunk_sent,
+                trace_metrics.chunks_sent as f64,
+                vec![],
+            ));
+        }
+        if trace_metrics.chunks_dropped > 0 {
+            futures.push(self.send(
+                self.trace_chunk_dropped,
+                trace_metrics.chunks_dropped as f64,
+                vec![],
             ));
         }
         for (status_code, count) in &trace_metrics.api_responses_count_per_code {
@@ -214,6 +237,27 @@ impl SelfTelemetry {
             ),
             trace_api_errors: worker.register_metric_context(
                 "trace_api.errors".to_string(),
+                vec![],
+                MetricType::Count,
+                true,
+                MetricNamespace::Tracers,
+            ),
+            trace_api_bytes: worker.register_metric_context(
+                "trace_api_bytes".to_string(),
+                vec![],
+                MetricType::Count,
+                true,
+                MetricNamespace::Tracers,
+            ),
+            trace_chunk_sent: worker.register_metric_context(
+                "trace_chunk_sent".to_string(),
+                vec![],
+                MetricType::Count,
+                true,
+                MetricNamespace::Tracers,
+            ),
+            trace_chunk_dropped: worker.register_metric_context(
+                "trace_chunk_dropped".to_string(),
                 vec![],
                 MetricType::Count,
                 true,
--- a/sidecar/src/self_telemetry.rs
+++ b/sidecar/src/self_telemetry.rs
@@ -28,9 +28,10 @@ struct MetricData<'a> {
     trace_api_requests: ContextKey,
     trace_api_responses: ContextKey,
     trace_api_errors: ContextKey,
-    trace_api_bytes: ContextKey,
-    trace_chunk_sent: ContextKey,
-    trace_chunk_dropped: ContextKey,
+    // TODO: APMSP-1157 - Enable this metric when support is enabled.
+    // trace_api_bytes: ContextKey,
+    trace_chunks_sent: ContextKey,
+    trace_chunks_dropped: ContextKey,
 }
 impl<'a> MetricData<'a> {
     async fn send(&self, key: ContextKey, value: f64, tags: Vec<Tag>) {
@@ -112,23 +113,24 @@ impl<'a> MetricData<'a> {
                 ],
             ));
         }
-        if trace_metrics.bytes_sent > 0 {
-            futures.push(self.send(
-                self.trace_api_bytes,
-                trace_metrics.bytes_sent as f64,
-                vec![tag!("src_library", "libdatadog")],
-            ));
-        }
+        // TODO: APMSP-1157 - Enable this metric when support is enabled.
+        // if trace_metrics.bytes_sent > 0 {
+        //     futures.push(self.send(
+        //         self.trace_api_bytes,
+        //         trace_metrics.bytes_sent as f64,
+        //         vec![tag!("src_library", "libdatadog")],
+        //     ));
+        // }
         if trace_metrics.chunks_sent > 0 {
             futures.push(self.send(
-                self.trace_chunk_sent,
+                self.trace_chunks_sent,
                 trace_metrics.chunks_sent as f64,
                 vec![tag!("src_library", "libdatadog")],
             ));
         }
         if trace_metrics.chunks_dropped > 0 {
             futures.push(self.send(
-                self.trace_chunk_dropped,
+                self.trace_chunks_dropped,
                 trace_metrics.chunks_dropped as f64,
                 vec![tag!("src_library", "libdatadog")],
             ));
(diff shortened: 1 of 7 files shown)


trace-utils/src/send_data.rs
```diff
@@ -153,24 +218,48 @@ impl SendData {
...
+                        RequestResult::TimeoutError(chunks)
+                    } else {
+                        RequestResult::NetworkError(chunks)
+                    }
+                }
+            }
         }
 
-        let mut result = SendDataResult::new();
+        //let mut result = SendDataResult::new();
```
pierotibou, line 252: You should remove it I assume.

trace-utils/src/send_data.rs
```diff
@@ -193,14 +278,12 @@ impl SendData {
...
+                let chunks = u64::try_from(tracer_payload.chunks.len()).unwrap();
                 let mut builder = HttpRequestBuilder::new()
                     .method(template.method.clone())
                     .uri(template.uri.clone())
                     .version(template.version)
-                    .header(
-                        "X-Datadog-Trace-Count",
-                        tracer_payload.chunks.len().to_string(),
-                    );
+                    .header("X-Datadog-Trace-Count", chunks.to_string());
```
pierotibou, line 286: nitpicking - Ignore for this PR, but I'm not a big fan of inlined string. I believe that it's better to define constants somewhere. It's also easier to document what they are meant for.

pierotibou (commented):
Thanks a lot for doing this. TLDR, I assume we should comment the sending of the `bytes` metrics (we can keep the rest) as it's not yet supported. And add a few missing `s` to metric names. 
The rest is mainly questions for my knowledge and some nitpicks that you should ignore :p

trace-utils/src/send_data.rs
```diff
@@ -4,81 +4,130 @@
...
+type BytesSent = u64;
+type ChunksSent = u64;
+type ChunksDropped = u64;
+
+pub enum RequestResult {
+    Success((Response<Body>, BytesSent, ChunksSent)),
+    Error((Response<Body>, ChunksDropped)),
+    TimeoutError(ChunksDropped),
+    NetworkError(ChunksDropped),
+    BuildError(ChunksDropped),
```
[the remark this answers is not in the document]
hoolioh: As Bob said that `enum` is to track the errors coming from the request builder. Probably I haven't picked the best name, maybe `BuildRequestError` or `RequestBuilderError` could work better?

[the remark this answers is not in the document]
pierotibou: Maybe a comment on the same line will do the work. Don't worry about naming with me, I'm very bad at naming

trace-utils/src/send_data.rs
hoolioh: Nice catch.

trace-utils/src/send_data.rs
```diff
@@ -4,81 +4,130 @@
...
+                *self
+                    .responses_count_per_code
+                    .entry(response.status().as_u16())
+                    .or_default() += 1;
+                self.chunks_dropped += chunks;
+
+                let body_bytes = hyper::body::to_bytes(response.into_body()).await;
+                let response_body =
+                    String::from_utf8(body_bytes.unwrap_or_default().to_vec()).unwrap_or_default();
+                self.last_result = Err(anyhow::format_err!(
```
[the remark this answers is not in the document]
pierotibou: I don't know if there's a preferred format. Either what you suggest or something more verbose.
`Status code: <status_code> - Message: <body>`? 
I don't have a real preference so as you wish.

trace-utils/src/send_data.rs
hoolioh: Since I needed to address some comments I've done this too. Besides it would've taken me more effort to just create the ticket than fixing it.

sidecar/src/self_telemetry.rs
```diff
@@ -218,6 +256,27 @@ impl SelfTelemetry {
                 true,
                 MetricNamespace::Tracers,
             ),
+            trace_api_bytes: worker.register_metric_context(
+                "trace_api_bytes".to_string(),
+                vec![],
+                MetricType::Count,
```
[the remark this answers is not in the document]
hoolioh: I've commented the code and created a new ticket to track the issue.

sidecar/src/self_telemetry.rs
```diff
@@ -27,6 +27,10 @@ struct MetricData<'a> {
     trace_api_requests: ContextKey,
     trace_api_responses: ContextKey,
     trace_api_errors: ContextKey,
+    // TODO: APMSP-1157 - Enable this metric when support is enabled.
+    // trace_api_bytes: ContextKey,
+    trace_chunk_sent: ContextKey,
+    trace_chunk_dropped: ContextKey,
```
bantonsson, line 33: They are both named `trace_chunks_*` in the docs.

sidecar/src/self_telemetry.rs
```diff
@@ -65,43 +68,78 @@ impl<'a> MetricData<'a> {
...
+        if trace_metrics.bytes_sent > 0 {
+            futures.push(self.send(
+                self.trace_api_bytes,
+                trace_metrics.bytes_sent as f64,
+                vec![Tag::new("src_library", "libdatadog").unwrap()],
+            ));
+        }
+        if trace_metrics.chunks_sent > 0 {
+            futures.push(self.send(
+                self.trace_chunk_sent,
```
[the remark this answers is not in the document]
bantonsson: I see that I'm late to the game...

sidecar/src/self_telemetry.rs
```diff
@@ -65,43 +68,78 @@ impl<'a> MetricData<'a> {
...
+                ],
             ));
         }
-
         if trace_metrics.api_requests > 0 {
             futures.push(self.send(
                 self.trace_api_requests,
                 trace_metrics.api_requests as f64,
-                vec![],
+                vec![Tag::new("src_library", "libdatadog").unwrap()],
```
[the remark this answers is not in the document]
bantonsson: I'm not too concerned about this part since it's only during reporting.

pierotibou (approved):
LGTM functionally (so please still wait for another review please). 
Thanks a lot for the implementation and the follow up on all comments.

sidecar/src/self_telemetry.rs
```diff
@@ -65,43 +69,79 @@ impl<'a> MetricData<'a> {
...
                 self.trace_api_errors,
                 trace_metrics.api_errors_status_code as f64,
-                vec![Tag::new("type", "status_code").unwrap()],
+                vec![
+                    Tag::new("type", "status_code").unwrap(),
+                    Tag::new("src_library", "libdatadog").unwrap(),
+                ],
+            ));
+        }
+        // TODO: APMSP-1157 - Enable this metric when support is enabled.
```
ekump, line 115: Would it make sense to feature flag this support instead of commenting it out?

trace-utils/src/trace_utils.rs
```diff
@@ -42,14 +42,14 @@ pub async fn get_traces_from_request_body(
...
-struct RootSpanTags<'a> {
+pub struct RootSpanTags<'a> {
     pub env: &'a str,
     pub app_version: &'a str,
     pub hostname: &'a str,
     pub runtime_id: &'a str,
 }
 
-fn construct_trace_chunk(trace: Vec<pb::Span>) -> pb::TraceChunk {
+pub fn construct_trace_chunk(trace: Vec<pb::Span>) -> pb::TraceChunk {
```
ekump, line 52: If we're making things public we should probably include doc comments.

sidecar/src/self_telemetry.rs
```diff
@@ -65,43 +69,79 @@ impl<'a> MetricData<'a> {
             futures.push(self.send(
                 self.logs_created,
                 count as f64,
-                vec![Tag::new("level", level.as_str().to_lowercase()).unwrap()],
+                vec![
+                    Tag::new("level", level.as_str().to_lowercase()).unwrap(),
```
ekump, line 73: I would be remiss if I didn't call out all of these unwraps on `Tag::new()`. If the `Result` is not Ok, should we log an error and not attempt to push the future? I assume we could wrap that logic in a helper function.

ekump: Actually, [this](https://github.com/DataDog/libdatadog/pull/450#issuecomment-2130221322) would solve the problem. So feel free to leave the unwraps for now.

trace-utils/src/trace_utils.rs
hoolioh: I got it wrong, it's used only inside the crate so I'm changing to `pub(crate)`.

trace-utils/src/send_data.rs
```diff
@@ -5,83 +5,171 @@ use anyhow::{anyhow, Context};
 use bytes::Bytes;
 use futures::stream::FuturesUnordered;
 use futures::StreamExt;
-use hyper::{Body, Client, Method, Response, StatusCode};
+use hyper::{Body, Client, Method, Response};
 use std::collections::HashMap;
+// use std::fmt::Result;
```
ekump, line 10: probably want to remove this

trace-utils/src/send_data.rs
```diff
@@ -383,17 +517,410 @@ impl SendData {
 }
 
 #[cfg(test)]
-// For RetryStrategy tests the observed delay should be approximate.
```
ekump, line 386: DId you mean to remove these comments?

trace-utils/src/send_data.rs
```diff
@@ -383,17 +517,410 @@ impl SendData {
...
+
+        assert_eq!(data.target.api_key.unwrap(), "TEST-KEY");
+        assert_eq!(data.target.url.path(), "/foo/bar");
+
+        assert_eq!(data.headers.get("DD-API-KEY").unwrap(), "TEST-KEY");
+    }
+
+    #[test]
+    fn send_data_new_no_api_key() {
+        let header_tags = TracerHeaderTags {
```
ekump, line 621: you could probably just create a `get_tracer_header_tags()` function or implement a default trait and add a `#[cfg(test)]` to it to avoid having to create one in every test.

trace-utils/src/send_data.rs
```diff
@@ -5,83 +5,171 @@ use anyhow::{anyhow, Context};
...
+            RequestError::Network => write!(f, "Network error"),
+            RequestError::Build => write!(f, "Request failed due to invalid property"),
+        }
+    }
 }
 
+impl std::error::Error for RequestError {}
+
+pub enum RequestResult {
+    // Holds information from a succesful request.
```
ekump, line 49: These should probably be `///` instead of `//` so they're treated as doc comments?

trace-utils/src/send_data.rs
```diff
@@ -5,83 +5,171 @@ use anyhow::{anyhow, Context};
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use datadog_trace_utils::send_data::RequestResult;
+    /// use datadog_trace_utils::trace_utils::SendDataResult;
+    ///
+    /// let result = RequestResult::NetworkError((1, 0));
+    /// let mut data_result = SendDataResult::default();
+    /// data_result.update(result);
```
ekump, line 115: This example is a bit misleading since the `.await` is missing from the call to update. You could put the example code within an async function and add the miri ignore so it compiles and doc tests passs

/// #[cfg_attr(miri, ignore)]
/// async fn test_with_poll() {
///    ...
///}

trace-utils/src/send_data.rs
hoolioh: Nope, a casualty from the rebase. Putting it back.

trace-utils/src/send_data.rs
hoolioh: Good catch, I didn't realize I was cluttering the tests when I could use the default impl.

ekump (approved):

PR #451, opened 2024-05-24
Fix coalescing send data

bwoebi:
There's an interesting tidbit of information in the documentation of dedup_by:

> The elements are passed in opposite order from their order in the slice, so if same_bucket(a, b) returns true, a is removed.

So, we were adding the data to a.tracer_payloads ... and then a gets dropped. Awesome, we just successfully dropped everything but one value.

On top of that, have a size restriction for the possible payload size, to ensure we'll never exceed it.

### DIFF
--- a/trace-utils/src/trace_utils.rs
+++ b/trace-utils/src/trace_utils.rs
@@ -17,6 +17,8 @@ const TOP_LEVEL_KEY: &str = "_top_level";
 /// Span metric the tracer sets to denote a top level span
 const TRACER_TOP_LEVEL_KEY: &str = "_dd.top_level";
 
+const MAX_PAYLOAD_SIZE: usize = 52428800;
+
 /// First value of returned tuple is the payload size
 pub async fn get_traces_from_request_body(
     body: Body,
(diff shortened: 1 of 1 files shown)


trace-utils/src/trace_utils.rs
```diff
@@ -17,6 +17,8 @@ const TOP_LEVEL_KEY: &str = "_top_level";
 /// Span metric the tracer sets to denote a top level span
 const TRACER_TOP_LEVEL_KEY: &str = "_dd.top_level";
 
+const MAX_PAYLOAD_SIZE: usize = 52428800;
```
bantonsson, line 20: Could we use `_` or multiplication in long numbers, like `52_428_800` or `50 * 1024 * 1024`?

trace-utils/src/trace_utils.rs
```diff
@@ -82,9 +84,15 @@ pub fn coalesce_send_data(mut data: Vec<SendData>) -> Vec<SendData> {
...
-            a.tracer_payloads.append(&mut b.tracer_payloads);
-            a.size += b.size;
-            return true;
+            // Size is only an approximation. In practice it won't vary much, but be safe here.
+            // We also don't care about the exact maximum size, like two 25 MB or one 50 MB request
+            // has similar results. The primary goal here is avoiding many small requests.
+            // TODO: maybe make the MAX_PAYLOAD_SIZE configurable?
+            if a.size + b.size < MAX_PAYLOAD_SIZE / 2 {
+                b.tracer_payloads.append(&mut a.tracer_payloads);
+                b.size += a.size;
```
bantonsson, line 93: Maybe call out the order of `dedup_by` in the comment so we don't break this again?

bantonsson (commented):
Nice find.

trace-utils/src/trace_utils.rs
bwoebi: Yes, absolutely.

trace-utils/src/trace_utils.rs
bwoebi: Done :-)

bantonsson (approved):

PR #453, opened 2024-05-24
feat(common): add tag! macro

morrisonlevi:
# What does this PR do?

This does a few things:

 1. Creates a `tag!` macro which does compile-time checking for some invariants.
 2. Changes a `Tag` to use `Cow<'static, str>` instead of a `String`.
 3. Makes `Tag::from_value` private. Use `tag!` if the key and value are known at compile-time. If not, use `Tag::new`.
 4. `Tag::new` returns an `anyhow::Result<Tag>` now instead of a `Result<Tag, Cow<_>>`.

# Motivation

As libdatadog usage has grown, we have more Rust users of the lib and therefore more static Rust strs we can trust at compile-time and avoid runtime allocations.

# Additional Notes

This was previously stacked on another PR, but not anymore.

# How to test the change?

Everything should run as normal. It's recommended to use `tag!` over `Tag::new` if the key and value are both static strs.


profiling-ffi/src/exporter.rs
```diff
@@ -511,7 +512,7 @@ mod tests {
     #[cfg_attr(miri, ignore)]
     fn profile_exporter_new_and_delete() {
         let mut tags = ddcommon_ffi::Vec::default();
-        let host = Tag::new("host", "localhost").expect("static tags to be valid");
+        let host = tag!("host", "localhost");
         tags.push(host);
```
danielsn, line 516: would this work?
```
let mut tags = vec![tag!("host", "localhost")].into();
```

sidecar/src/self_telemetry.rs
```diff
@@ -138,7 +133,7 @@ impl<'a> MetricData<'a> {
                 *count as f64,
                 vec![
                     Tag::new("status_code", status_code.to_string().as_str()).unwrap(),
```
danielsn, line 135: Does the `tag!` macro only work on constant strings?

danielsn (approved):
Nice

ddcommon/src/tag.rs
```diff
@@ -1,15 +1,68 @@
...
+            value: Cow::Borrowed(value),
+        }
+    }
+}
+
+// todo: what's a good way to keep these in-sync with Tag::from_value?
+// This can be a little more strict because it's compile-time evaluated.
+// https://docs.datadoghq.com/getting_started/tagging/#define-tags
+#[macro_export]
+macro_rules! tag {
```
danielsn, line 38: Put a doc comment explaining that this only works for compile time constant strings

profiling-ffi/src/exporter.rs
morrisonlevi: It turns out that yes, it works!

ekump:
This is cool. While reviewing #450 I was thinking to myself "man, it sucks we can't know these tags are valid at compile time".

ddcommon/src/tag.rs
```diff
@@ -32,9 +89,8 @@ impl Display for Tag {
 }
 
 impl Tag {
-    /// It's recommended to use Tag::new when possible, as tags that are in
-    /// the <KEY>:<VALUE> format are preferred.
-    pub fn from_value<'a, IntoCow>(chunk: IntoCow) -> Result<Self, Cow<'static, str>>
+    /// Validates a tag.
+    fn from_value<'a, IntoCow>(chunk: IntoCow) -> Result<Self, Cow<'static, str>>
```
danielsn, line 93: Should this be `try_from`?

ddcommon/src/tag.rs
```diff
@@ -32,9 +89,8 @@ impl Display for Tag {
 }
 
 impl Tag {
-    /// It's recommended to use Tag::new when possible, as tags that are in
-    /// the <KEY>:<VALUE> format are preferred.
-    pub fn from_value<'a, IntoCow>(chunk: IntoCow) -> Result<Self, Cow<'static, str>>
+    /// Validates a tag.
```
danielsn, line 92: Cleanup: this function could use `anyhow::ensure`

bantonsson (approved):
Really nice!

I wasn't completely sure what this would end up as in the generated binary, so I dug into it with `readelf` and `objdump` and it's indeed reusing constant tag data defined in multiple places.

ddcommon/src/tag.rs
morrisonlevi: That would make it a public since it's a trait impl. It's an implementation detail used by `parse_tags` and `new`.

ekump (approved):

PR #454, opened 2024-05-24
[profiling] use internal Label instead of pprof::Label

taegyunkim:
# What does this PR do?

Resolves a TODO. Use `internal::Label` instead of `pprof::Label` to make it easier to fuzz test. Even though `pprof::Label` can also derive `bolero_generator::TypeGenerator` guarded by `cfg(test)`, it would be best to keep internal things internal.
 Also, this lets us use `pprof::Label` only when returning pprof.

# Motivation

TODO

# Additional Notes

# How to test the change?

`cargo nextest run`

### DIFF
--- a/profiling/src/internal/profile.rs
+++ b/profiling/src/internal/profile.rs
@@ -214,7 +214,7 @@ impl Profile {
         let mut encoder = CompressedProtobufSerializer::with_capacity(INITIAL_PPROF_BUFFER_SIZE);
 
         for (sample, timestamp, mut values) in std::mem::take(&mut self.observations).into_iter() {
-            let labels = self.translate_and_enrich_sample_labels(sample, timestamp)?;
+            let labels = self.enrich_sample_labels(sample, timestamp)?;
             let location_ids: Vec<_> = self
                 .get_stacktrace(sample.stacktrace)?
                 .locations
@@ -223,6 +223,7 @@ impl Profile {
                 .collect();
             self.upscaling_rules.upscale_values(&mut values, &labels)?;
 
+            let labels = labels.into_iter().map(pprof::Label::from).collect();
             let item = pprof::Sample {
                 location_ids,
                 values,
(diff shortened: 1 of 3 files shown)


morrisonlevi:
I believe this means we can also remove `StringId::new` now, yes?

danielsn (approved):
LGTM

profiling/src/internal/profile.rs
```diff
@@ -481,20 +482,16 @@ impl Profile {
...
         self.get_label_set(sample.labels)?
             .iter()
-            .map(|l| self.get_label(*l).map(pprof::Label::from))
-            .chain(
-                self.get_endpoint_for_labels(sample.labels)
-                    .transpose()
-                    .map(|res| res.map(pprof::Label::from)),
-            )
-            .chain(timestamp.map(|ts| Ok(Label::num(self.timestamp_key, ts.get(), None).into())))
+            .map(|l| self.get_label(*l).copied())
```
danielsn, line 492: Thought: since `Label` is `Copy`, should `get_label` just return an owned object rather than a reference?

danielsn: (not needed for this PR, but might be interesting to not bother taking `Copy` types by reference

profiling/src/internal/upscaling.rs
```diff
@@ -151,17 +151,20 @@ impl UpscalingRules {
...
             // get bylabel rules first (if any)
             let mut group_of_rules = labels
                 .iter()
-                .filter_map(|label| self.get(&(StringId::new(label.key), StringId::new(label.str))))
+                .filter_map(|label| {
+                    self.get(&(
+                        label.get_key(),
+                        match label.get_value() {
+                            LabelValue::Str(str) => *str,
+                            LabelValue::Num { .. } => StringId::ZERO,
```
danielsn, line 164: Is this an error if this case is reached?

profiling/src/internal/profile.rs
morrisonlevi: I know for small types, yes, it's definitely preferred to work by-value instead of reference. I'm not sure what "small" means for Rust calling conventions, though. Is `Label` small?

profiling/src/internal/upscaling.rs
taegyunkim: I simply looked at `impl From<&Label> for pprof::Label` here.  https://github.com/DataDog/libdatadog/blob/0593cd71b018dcb53dcc37b82d7ab1306cbbf277/profiling/src/internal/label.rs#L59-L73

I don't think this was treated an error before this PR.

profiling/src/internal/profile.rs
taegyunkim: `Label` has `StringId`, which is `u32` and `LabelValue`. 
https://github.com/DataDog/libdatadog/blob/633b0b02ce317ad3145ecd2859b3d9b61d5933c2/profiling/src/internal/label.rs#L16-L19

`LabelVale` is an enum which can be 
- `StringId`, again `u32`
- `i64` and optional `StringId`

https://github.com/DataDog/libdatadog/blob/633b0b02ce317ad3145ecd2859b3d9b61d5933c2/profiling/src/internal/label.rs#L7

PR #457, opened 2024-05-28
Use a SIDECAR_VERSION by default for the reported version

bwoebi:
Apparently there's no env_or_default, option_env!() exists, but it's not fine with concat!(), ... :-)


bantonsson (approved):
Too bad that you can't do option manipulation in a `const fn`, or that could maybe have been a more readable solution.

PR #459, opened 2024-05-29
Bump rustls version to 0.23

bwoebi:
We were running into symbol issues with older rustls versions, see https://github.com/briansmith/ring/issues/1808.

Note that the new rustls version requires nasm to be installed on windows, so we install it in CI.


bantonsson (approved):
🎉

PR #463, opened 2024-05-31
[Serverless Mini Agent] Add Azure App Service Span Tags

duncanpharvey:
# What does this PR do?

Adds `aas.*` span tags to all spans submitted from Azure Functions.

# Motivation

https://datadoghq.atlassian.net/browse/SVLS-4920

Allows for proper billing of Azure Functions and for traces to be correlated for the app on the Azure tab of the Serverless page.

# Additional Notes

- Reused `azure_app_services` in `ddcommon` to build these span tags
  * Changes are backwards compatible with the profiling package that currently uses `azure_app_services` by creating `get_function_metadata` which has Azure Function specific logic. `get_metadata` expects the `DD_AZURE_APP_SERVICES` environment variable to be set to true, which only occurs when the [Datadog Extension for Windows](https://github.com/DataDog/datadog-aas-extension) is installed
    - https://github.com/DataDog/libdatadog/blob/ebe9a6a502a03b32d06b108b19cb3048c339b85a/profiling/src/exporter/mod.rs#L18
- Operating system now falls back to `std::env::consts::OS.to_string()` if `WEBSITE_OS` environment variable is not set
- Resource group now falls back to being parsed from the `WEBSITE_ONWER_NAME` environment variable if `WEBSITE_RESOURCE_GROUP` is not set (Azure Functions on Consumption plans only)
- Added `mini_agent_version` property to mini agent config to surface in a span tag
  * Value from `CARGO_PKG_VERSION` environment variable

Screenshot 2024-05-31 at 9 51 52 AM

# How to test the change?

* Build mini agent locally
* Deploy Azure Function with mini agent binary in root path
* Set `DD_MINI_AGENT_PATH` to `/home/site/wwwroot/datadog-serverless-trace-mini-agent`


ddcommon/src/azure_app_services.rs
```diff
@@ -93,6 +94,15 @@ impl AzureMetadata {
             .map(|v| v.to_string())
     }
 
+    fn extract_resource_group(s: Option<String>) -> Option<String> {
+        let re: Regex = Regex::new(r"(.+)\+(.+)-(.+)-(.+)").unwrap();
```
astuyve, line 98: We typically avoid using `unwrap()` as if the call fails, it'll crash our function.
Here it's probably fine as the regex is static (so it [shouldn't error](https://docs.rs/regex/latest/regex/struct.Regex.html#method.new)).

I'm not sure what the best answer here would be, maybe matching and returning a safe default?

ddcommon/src/azure_app_services.rs
```diff
@@ -155,6 +160,32 @@ impl AzureMetadata {
...
+
+        if !is_relevant {
+            return None;
+        }
+
+        AzureMetadata::build_metadata(query)
+    }
+
+    pub fn new_function<T: QueryEnv>(query: T) -> Option<Self> {
+        let is_relevant = matches!(
```
astuyve, line 177: I suspect this could be re-written as a match statement, although the `matches!` macro makes that a little weird.

Something like:
```rust
match matches!(...) {
  true => AzureMetadata::build_metadata(query),
  false => None
}
```

astuyve (approved):
minor nits, otherwise looks good. Make sure to run `cargo clippy` and `cargo fmt` if CI hasn't (guessing libdatadog does this, but I didn't check)

trace-utils/src/trace_utils.rs
```diff
@@ -238,6 +239,35 @@ pub fn enrich_span_with_mini_agent_metadata(
     }
 }
 
+pub fn enrich_span_with_azure_metadata(span: &mut pb::Span, mini_agent_version: &str) {
+    match azure_app_services::get_function_metadata() {
```
duncanista, line 243: This method can be rewritten as an 
```rust
if let Some(aas_metadata) = match ... {
   ...
}```

duncanpharvey: Done!

ddcommon/src/azure_app_services.rs
duncanpharvey: Done!

ddcommon/src/azure_app_services.rs
duncanpharvey: I added some unit tests to validate the behavior of the `extract_resource_group` regex function. If the pattern matches we return the resource group as designed. If the pattern doesn't match then `None` is returned and the resource group resolves to an unknown value gracefully.

Let me know if you think that's sufficiently safe! All of the examples in the regex docs use `unwrap()` so I wasn't sure what the best alternative was.
https://docs.rs/regex/latest/regex/struct.Regex.html#method.captures

https://github.com/DataDog/libdatadog/blob/88c6c960320842194f342d214e685762385fca65/ddcommon/src/azure_app_services.rs#L401-L448

gleocadie (approved):
LGTM

PR #465, opened 2024-05-31
Fix bug introduced in APMSP-1020 that was not setting trace count header for msgpack requests to tracer

ekump:
# What does this PR do?

Introduced a bug in APMSP-1020 where we were setting `X-Datadog-Trace-Count` for msgpack requests and then throwing the header away and rebuilding the request. This PR now sets the header appropriately. 

# Motivation

# Additional Notes

# How to test the change?

Unit tests for msgpack requests now validate headers are correctly sent.

### DIFF
--- a/trace-utils/src/send_data.rs
+++ b/trace-utils/src/send_data.rs
@@ -5,7 +5,10 @@ use anyhow::{anyhow, Context};
 use bytes::Bytes;
 use futures::stream::FuturesUnordered;
 use futures::StreamExt;
-use hyper::{Body, Client, Method, Response};
+use hyper::{
+    header::{HeaderMap, HeaderValue},
+    Body, Client, Method, Response,
+};
 use std::collections::HashMap;
 use std::time::Duration;
 use tokio::time::sleep;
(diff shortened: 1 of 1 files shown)


trace-utils/src/send_data.rs
```diff
@@ -725,16 +737,27 @@ mod tests {
...
                     .header("Content-type", "application/msgpack")
+                    .header("datadog-meta-lang", header_tags.lang)
+                    .header(
+                        "datadog-meta-lang-interpreter",
+                        header_tags.lang_interpreter,
+                    )
+                    .header("datadog-meta-lang-version", header_tags.lang_version)
+                    .header("datadog-meta-lang-vendor", header_tags.lang_vendor)
+                    .header("datadog-meta-tracer-version", header_tags.tracer_version)
+                    .header("datadog-container-id", header_tags.container_id)
```
ekump, line 754: @hoolioh The to HashMap [implementation](https://github.com/DataDog/libdatadog/blob/main/trace-utils/src/tracer_header_tags.rs#L40) for `TracerHeadertags` always ignores `client_computed_stats` and `client_computed_top_level` tags. I assume that is intentional?

pierotibou: I assume that's because PHP don't support it. cf [this](https://github.com/DataDog/dd-trace-php/blob/92560faa353e7d5412a0b0517d13f4e9a0506d0c/ext/auto_flush.c#L66). @bwoebi can confirm.

That said we should open a ticket for it not to forget. It's gonna be important when we will compute stats on the client side.

trace-utils/src/send_data.rs
```diff
@@ -725,16 +737,27 @@ mod tests {
     async fn request_msgpack() {
         let server = MockServer::start_async().await;
 
+        let header_tags = HEADER_TAGS;
         let mock = server
             .mock_async(|when, then| {
                 when.method(POST)
+                    .header(HEADER_DD_TRACE_COUNT, "1")
```
pierotibou, line 744: `when` asserts the header is present, right? Not just setting up the mock, right?

pierotibou (approved):
LGTM. Feel free to wait for a 2nd pair of eyes.

trace-utils/src/send_data.rs
ekump: correct, if the headers aren't set correctly the mock won't return a 200 and the test will fail.

trace-utils/src/send_data.rs
pierotibou: APMSP-1167 for the follow up

bwoebi: Yeah, no need for it yet.

bwoebi (approved):
Looks right to me now, thanks for tackling it quickly!

PR #469, opened 2024-06-04
Spawn one sidecar per user

bwoebi:
Restricting permissions to shm data to individual users. This solves any sort of permission problems:
- agent shm config could not be read despite being readable
- we probably don't want other users to write configs for us
- it will be required with remote config, which uses signals as notification mechanism, which need the same euid on both ends

Also includes fixes for shared memory on macos:
- shared memory can only be ever sized once, resizes, even though required by ftruncate() POSIX, are not supported on macos.
- Adds some tests for shared memory


sidecar/src/agent_remote_config.rs
```diff
@@ -22,7 +23,12 @@ fn path_for_endpoint(endpoint: &Endpoint) -> CString {
     // We need a stable hash so that the outcome is independent of the process
     let mut hasher = ZwoHasher::default();
     endpoint.url.authority().unwrap().hash(&mut hasher);
-    CString::new(format!("/libdatadog-agent-config-{}", hasher.finish())).unwrap()
+    CString::new(format!(
+        "/ddcfg-{}-{}",
```
bwoebi, line 27: macos has a length limit of 30 chars, this fits it.

bantonsson: Maybe add a comment in the code explaining that?

bantonsson (approved):

sidecar/src/agent_remote_config.rs
bwoebi: Yes :-D

ekump:
Overriding branch protection. dd-gitlab/report_gitlab_CI_status is green, but still blocking the PR.

PR #471, opened 2024-06-05
ekump/APMSP-1076 break up trace utils send data file

ekump:
# What does this PR do?

Moves `RetryStrategy` and `SendDataResult` to separate files and do some minor code cleanup. No functional changes.

# Motivation

A lot has been added to `SendData` recently and the file has become unwieldy. 

# How to test the change?

Existing unit tests remain unchanged and should still pass.

### DIFF
--- a/trace-utils/src/send_data/mod.rs
+++ b/trace-utils/src/send_data/mod.rs
@@ -1,21 +1,21 @@
 // Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
+pub mod retry_strategy;
+pub mod send_data_result;
+
+pub use crate::send_data::retry_strategy::{RetryBackoffType, RetryStrategy};
+
+use crate::trace_utils::{SendDataResult, TracerHeaderTags};
 use anyhow::{anyhow, Context};
 use bytes::Bytes;
+use datadog_trace_protobuf::pb::{AgentPayload, TracerPayload};
+use ddcommon::{connector, Endpoint, HttpRequestBuilder};
 use futures::stream::FuturesUnordered;
 use futures::StreamExt;
-use hyper::{
-    header::{HeaderMap, HeaderValue},
-    Body, Client, Method, Response,
-};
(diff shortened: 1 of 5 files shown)


trace-utils/src/send_data/retry_strategy.rs
```diff
@@ -0,0 +1,228 @@
...
+#[derive(Debug, Clone)]
+pub struct RetryStrategy {
+    /// The maximum number of retries to attempt.
+    pub max_retries: u32,
+    /// The minimum delay between retries.
+    pub delay_ms: Duration,
+    /// The type of backoff to use for the delay between retries.
+    pub backoff_type: RetryBackoffType,
+    /// An optional jitter to add randomness to the delay.
+    pub jitter: Option<Duration>,
```
hoolioh, line 48: Should we limit the visibility of the fields to the crate?

ekump: I'd have to implement  `new()` for RetryStrategy...which is probably a good idea.

hoolioh (approved):

trace-utils/src/send_data/mod.rs
```diff
@@ -60,243 +60,65 @@ pub enum RequestResult {
...
-        } else {
-            sleep(delay).await;
-        }
-    }
-}
-
-#[derive(Debug, Clone)]
 pub struct SendData {
-    pub(crate) tracer_payloads: Vec<pb::TracerPayload>,
+    pub(crate) tracer_payloads: Vec<TracerPayload>,
```
bantonsson, line 99: A bit late to the game here, but I think that the `pb::*` structs should always be fully qualified since some of the names are _generic_ and will be used by other types in these files. I mean there will likely be a `Span` that is not the `pb::Span` to represent some of these things. Anyway, nothing to change now, just something to keep in mind going forward.

ekump: Ah, that's a good point. I slightly prefer aliasing the import in that situation, but this is fine too. Added back in #495

PR #473, opened 2024-06-06
[crashtracker] Take relative address and build id for remote symbolication

danielsn:
# What does this PR do?

Adds two new fields optional to each stack trace frame: `relative_address` and `build_id`

# Motivation

This is the information required by a backend symbolizer

# Additional Notes

Getting relative addresses from absolute addresses is the responsibility of the client.  In the future, we may add functionality to libdatadog to support this.

# How to test the change?


profiling-ffi/src/crashtracker/datatypes.rs
```diff
@@ -325,12 +327,15 @@ impl<'a> TryFrom<&StackFrame<'a>> for datadog_crashtracker::StackFrame {
             }
             Some(vec)
         };
+        let relative_address = to_hex(value.relative_address);
```
danielsn, line 330: Is this valid to assume a relative offset of `0` is `None`? Or do we need to be more clever here?

r1viollet: the term relative is ambiguous, as it can have different meanings.
we use `elf`, blazesym uses `normalized`. I can imagine you are aiming for cross compatibility ?
0 for an elf address is not possible (there are headers)

peterg17 (dismissed):
We should need "just" these 2 pieces of information, but I suspect most of the complexity is in deriving the elf address, which could be helpful to exist in libdatadog. Nicolas and Erwan would be the experts here and their code in ddprof probably would be useful to reference

crashtracker/src/stacktrace.rs
```diff
@@ -58,6 +97,19 @@ mod unix {
     }
 
     impl StackFrame {
+        pub fn normalize_ip(&mut self, normalizer: &Normalizer, pid: Pid) -> anyhow::Result<()> {
+            if let Some(ip) = &self.ip {
+                let ip = ip.trim_start_matches("0x");
+                let ip = u64::from_str_radix(ip, 16)?;
+                let normed = normalizer.normalize_user_addrs(pid, &[ip])?;
```
peterg17, line 104: Makes sense to me based on my reading of the blazesym docs - https://docs.rs/blazesym/latest/i686-unknown-linux-gnu/blazesym/normalize/struct.Normalizer.html

crashtracker/src/stacktrace.rs
```diff
@@ -33,16 +34,54 @@ pub struct StackFrame {
...
+        Pid,
+    };
+
+    impl From<&UserMeta> for NormalizedAddressMeta {
+        fn from(value: &UserMeta) -> Self {
+            match value {
+                UserMeta::Apk(a) => Self::Apk(a.path.clone()),
+                UserMeta::Elf(e) => Self::Elf {
+                    path: e.path.clone(),
+                    build_id: e.build_id.clone(),
```
peterg17, line 78: Does this handle go build id, which can be in a different part of the metadata than the GNU build id? Or is this irrelevant because we don't currently use libdatadog in dd-trace-go?

peterg17 (approved):
Looks good!

crashtracker/src/stacktrace.rs
danielsn: I'd guess the second, but Felix/Nick may have thoughts.  If they want it, I'll add it in a followup PR

PR #475, opened 2024-06-07
Skip normalization & obfuscation and coalesce instead

bwoebi:
Turns out the agent ignores the root_span_tags for v0.7 anyway. This is useful to coalesce chunks with common data, allowing us to actually save us from doing redundant requests. As it was currently, the agentful trace sender always sent one trace per request, instead of merging them.

We also skip normalization and rely on the agent to do it, for sake of consistency with other tracers and reduce the potentially duplicated work across tracer and agent. We may change that back in future, but for now we've determined it to be the easiest way to work, also with respect to normalization-unaware testing.

### DIFF
--- a/trace-mini-agent/src/trace_processor.rs
+++ b/trace-mini-agent/src/trace_processor.rs
@@ -85,7 +85,7 @@ impl TraceProcessor for ServerlessTraceProcessor {
                     obfuscate_span(span, &config.obfuscation_config);
                 }
             },
-            true,
+            true, // In mini agent, we always send agentless
         );
 
         let send_data = SendData::new(body_size, payload, tracer_header_tags, &config.trace_intake);
--- a/trace-mini-agent/src/trace_processor.rs
+++ b/trace-mini-agent/src/trace_processor.rs
@@ -85,6 +85,7 @@ impl TraceProcessor for ServerlessTraceProcessor {
                     obfuscate_span(span, &config.obfuscation_config);
                 }
             },
+            true,
         );
 
         let send_data = SendData::new(body_size, payload, tracer_header_tags, &config.trace_intake);
--- a/trace-utils/Cargo.toml
+++ b/trace-utils/Cargo.toml
@@ -23,12 +23,12 @@ datadog-trace-normalization = { path = "../trace-normalization" }
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 rand = "0.8.5"
 bytes = "1.6.0"
-# This should only be used for testing. It isn't under dev-dependencies because test-utils can't be under #[cfg(test)].
-httpmock = {  version = "0.7.0", optional = true}
+httpmock = { version = "0.7.0", optional = true}
 
 [dev-dependencies]
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 serde_json = "1.0"
+httpmock = { version = "0.7.0"}
 
 [features]
 test-utils = ["httpmock"]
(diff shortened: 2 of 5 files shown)


trace-utils/src/trace_utils.rs
```diff
@@ -103,6 +116,20 @@ pub fn coalesce_send_data(mut data: Vec<SendData>) -> Vec<SendData> {
...
     });
+    // Merge chunks with common properties. Reduces requests for agentful mode.
+    // And reduces a little bit of data for agentless.
+    for send_data in data.iter_mut() {
+        send_data
+            .tracer_payloads
+            .sort_unstable_by(cmp_send_data_payloads);
+        send_data.tracer_payloads.dedup_by(|a, b| {
+            if cmp_send_data_payloads(a, b) == Ordering::Equal {
+                b.chunks.append(&mut a.chunks);
```
pierotibou, line 127:
```suggestion
        send_data.tracer_payloads.dedup_by(|a, b| {
            if cmp_send_data_payloads(a, b) == Ordering::Equal {
                // Note: dedup_by drops a, and retains b.
                b.chunks.append(&mut a.chunks);
```

trace-utils/src/trace_utils.rs
```diff
@@ -103,6 +116,20 @@ pub fn coalesce_send_data(mut data: Vec<SendData>) -> Vec<SendData> {
         }
         false
     });
+    // Merge chunks with common properties. Reduces requests for agentful mode.
```
pierotibou, line 119: Is it an optimization that we should keep for another PR? To stay on the safe side?

trace-utils/src/trace_utils.rs
```diff
@@ -288,10 +315,12 @@ pub fn collect_trace_chunks(
     mut traces: Vec<Vec<pb::Span>>,
     tracer_header_tags: &TracerHeaderTags,
     process_chunk: impl Fn(&mut TraceChunk, usize),
+    is_agentless: bool,
 ) -> pb::TracerPayload {
     let mut trace_chunks: Vec<pb::TraceChunk> = Vec::new();
 
-    let mut gathered_root_span_tags = false;
+    // We'll skip setting the global metadata and reply on the agent to unpack these
```
pierotibou, line 322:
```suggestion
    // We'll skip setting the global metadata and rely on the agent to unpack these
```

trace-mini-agent/src/trace_processor.rs
```diff
@@ -85,6 +85,7 @@ impl TraceProcessor for ServerlessTraceProcessor {
                     obfuscate_span(span, &config.obfuscation_config);
                 }
             },
+            true,
```
pierotibou, line 88:
```suggestion
            true, // In mini agent, we always send agentless
```

pierotibou (commented):
nitpicks on comments

trace-utils/src/trace_utils.rs
bwoebi: Sort of mandatory fix, yes.

pierotibou (approved):

hoolioh (approved):

trace-utils/Cargo.toml
```diff
@@ -23,12 +23,12 @@ datadog-trace-normalization = { path = "../trace-normalization" }
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
 rand = "0.8.5"
 bytes = "1.6.0"
-# This should only be used for testing. It isn't under dev-dependencies because test-utils can't be under #[cfg(test)].
```
ekump, line 26: This needs to be put back.

ekump:
bypassing branch protection and merging at @pierotibou's request. dd-gitlab/report_gitlab_CI_status job is not working as expected. Gitlab manually verified to pass.

PR #477, opened 2024-06-07
[crashtracker] Add a 'wait_for_receiver' param, and use it in both the stdin and unix socket cases

danielsn:
# What does this PR do?

Adds a `wait_for_receiver` flag to the collector config.

# Motivation

In some cases, the client may wish the collector to wait until the receiver has uploaded before finishing crashing, e.g. to keep the pod alive, or to allow out-of-process symbolization.  In other cases, they may wish the process to die immediately.  This lets the client decide. 
# Additional Notes

# How to test the change?


sanchda (approved):
Simple improvement for fixing coordination between the two actors.  I think this is a great, multiplatform-safe, robust strategy.  LGTM

ivoanjo:
Note to self: I was gonna add this to Ruby but then realized it's not yet on libdatadog 10.0 . I'll come back later ;)

PR #480, opened 2024-06-10
Update protobuf files for stats for new trilean type

bantonsson:
# What does this PR do?

Updates protobuf files to match Datadog Agent.

### DIFF
--- a/.github/workflows/verify-proto-files.yml
+++ b/.github/workflows/verify-proto-files.yml
@@ -4,7 +4,7 @@ on:
     types: [ opened, synchronize, reopened ]
 
 env:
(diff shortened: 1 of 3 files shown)


hoolioh (approved):

paullegranddc (approved):
LGTM, but I don't quite get the change, because if the struct tags and enums are the same, this is basically a no-op in term of format.

bantonsson:
@paullegranddc it is indeed a no-op on the wire, but the names have been changed in the Datadog Agent protobuf files, so our diff to make sure we don't miss new changes to the files will still fail.

PR #481, opened 2024-06-10
Drop fd acking in sidecar transport

bwoebi:
This has effectively been never working - it probably got added in long times past, before unidirectional sending was added. There's no ack, and thus it would simply leak and stay unclosed forever.

And the functionality is simply not needed. It's not like linux would close file descriptions as long as they're referenced in sendmsg buffers.

### DIFF
--- a/ipc/src/platform/unix/channel.rs
+++ b/ipc/src/platform/unix/channel.rs
@@ -101,10 +101,7 @@ impl Write for Channel {
                         "failed to write whole buffer",
                     ));
                 }
-                Ok(n) => {
-                    self.metadata.defer_close_handles(handles);
-                    buf = &buf[n..]
-                }
+                Ok(n) => buf = &buf[n..],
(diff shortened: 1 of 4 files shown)


iamluc (approved):

bantonsson (commented):
I can only repeat that it would have been nice with some comments trying to explain the whole machinery in here.

As for removing this ack thing, what does that mean for the rest of the convoluted code in `PlatformHandle`, and how do we test that properly?

bwoebi:
I haven't written that code, but the primary issue is that, if a write couldn't be sent through, we'll have to retry the write. The rust async writes machinery will retry the write with a given buf then. But it doesn't have any idea of the adjacent file descriptions sent. So we'll have to store these separately and try sending them again when possible.
As Linux doesn't allow sending just file descriptors, it _requires_ a non-empty accompanying message, we have to mix in the fd sending with the writes.

PR #482, opened 2024-06-11
Refactor to reduce signature errors risks

VianneyRuhlmann:
# What does this PR do?

Refactor the fallback for unix specific functions in `entity_id` to reduce risks of signature inconsistency between the unix and the fallback implementation.

# Motivation

Different signatures between fallback and unix versions fixed by https://github.com/DataDog/libdatadog/pull/474 

# Additional Notes

Only refactor, no functional changes


ddcommon/src/entity_id/mod.rs
```diff
@@ -49,16 +49,50 @@
...
+    #[cfg(not(unix))]
+    {
+        None
+    }
+}
+
+/// Set the path to cgroup file to mock it during tests
+/// # Safety
+/// Must not be called in multi-threaded contexts
+pub unsafe fn set_cgroup_file(file: String) {
```
bantonsson, line 83: `#[allow(unused_variables)]` 😉

bantonsson (approved):
Nice cleanup.

ddcommon/src/entity_id/mod.rs
VianneyRuhlmann: I used `_file` clippy doesn't complain and it seems a bit more specific (even though in such a small function it doesn't  make any difference)

PR #483, opened 2024-06-11
[crashtracker] Send pid to the receiver

danielsn:
# What does this PR do?

Sends the PID to the receiver from the crashing process

# Motivation

Currently, we use `getppid`, which only works because the receiver is the child of the crashing process.  If we want to have a sidecar receiver, we'll need to know the actual pid.

# Additional Notes

# How to test the change?


r1viollet (approved):
LGTM
This seems like something interesting to add tests to.

PR #484, opened 2024-06-11
[crashtracker] Add tests for unix socket receiver, and mark slow tests ignore

danielsn:
# What does this PR do?

* Adds an end-to-end test for the unix socket receiver for the crashtracker.
* Marks a slow test `ignore`

# Motivation

* We didn't actually have testing for this yet, now we do.
* Testing the "release" version was more of a test-the-tests thing, don't actually need to do that regularly to get confidence in the underlying code.
# Additional Notes

These tests pass a bunch of arguments positionally.  This is not ideal, and made them harder to maintain, but since they're tests I felt leaving it that way was OK for now.

# How to test the change?

This is a test!


bin_tests/src/bin/crashtracker_bin_test.rs
```diff
@@ -43,34 +49,59 @@ mod unix {
...
+                CrashtrackerReceiverConfig::new(
+                    vec![],
+                    env::vars().collect(),
+                    receiver_binary,
+                    Some(stderr_filename),
+                    Some(stdout_filename),
+                )?,
+                metadata,
+            )?;
+        } else if mode == format!("{:?}", ReceiverType::UnixSocket) {
```
danielsn, line 85: This is a little ugly.  Is there a better way?

bin_tests/tests/test_the_tests.rs
```diff
@@ -14,6 +14,7 @@ fn test_the_tests_debug() {
 }
 
 #[test]
+#[ignore] // This test is slow, only run it if explicitly opted in
 #[cfg_attr(miri, ignore)]
```
taegyunkim, line 18: If this test is ignore, can `#[cfg_attr(miri, ignore)]` this be removed?

bin_tests/src/bin/crashtracker_bin_test.rs
taegyunkim: How about converting the `mode` string into `ReceiverType` enum and use match? https://docs.rs/strum/latest/strum/derive.EnumString.html

taegyunkim (changes requested):

bin_tests/src/bin/crashtracker_bin_test.rs
danielsn: Done.

bin_tests/tests/test_the_tests.rs
danielsn: done

taegyunkim (approved):

PR #486, opened 2024-06-13
style: fix clippy lints from Rust 1.79 release

morrisonlevi:
# What does this PR do?

Fixes lints from Rust 1.79 release.

# Motivation

A new stable release was made, so CI starts failing.

# Additional Notes

Nope.

# How to test the change?

Everything should be the same.


profiling-ffi/src/exporter.rs
```diff
@@ -15,31 +15,37 @@ use std::borrow::Cow;
 use std::ptr::NonNull;
 use std::str::FromStr;
 
+#[allow(dead_code)]
```
danielsn, line 18: do we need both of these?

profiling/src/internal/profile/mod.rs
```diff
@@ -359,12 +359,9 @@ impl Profile {
         );
 
         let local_root_span_id: u64 = if let LabelValue::Num { num, .. } = label.get_value() {
-            // Manually specify the type here to be sure we're transmuting an
-            // i64 and not a &i64.
-            let id: i64 = *num;
             // Safety: the value is a u64, but pprof only has signed values, so we
             // transmute it; the backend does the same.
-            unsafe { std::intrinsics::transmute(id) }
+            unsafe { std::intrinsics::transmute::<i64, u64>(*num) }
```
danielsn, line 364: 💯

danielsn (approved):
One small question, but LGTM

profiling-ffi/src/exporter.rs
morrisonlevi: Oops, forgot to remove the first one I tried. Thanks.

PR #490, opened 2024-06-14
[chore] Require all checks to pass on a PR

danielsn:
# What does this PR do?

Adds a CI job to ensure that all checks pass on a PR.

# Motivation

Currently, marking checks "required" is manual, and hard to do for lots of tests.  This automates that.

# Additional Notes

https://github.com/marketplace/actions/allcheckspassed

# How to test the change?

### DIFF
--- a/.github/workflows/all-checks.yml
+++ b/.github/workflows/all-checks.yml
@@ -14,5 +14,5 @@ jobs:
           - uses: wechuli/allcheckspassed@v1
             with:
                 delay: '3'
-                retries: '20'
+                retries: '30'
                 polling_interval: '1'
--- a/.github/workflows/all-checks.yml
+++ b/.github/workflows/all-checks.yml
@@ -1,0 +1,14 @@
+name: All checks pass
+on:
+    pull_request:
+    push:
+      branches:
+        - main
+jobs:
+    allchecks:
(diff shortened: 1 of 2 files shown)


.github/workflows/all-checks.yml
```diff
@@ -0,0 +1,18 @@
...
+    allchecks:
+        runs-on: ubuntu-latest
+        permissions:
+            checks: read
+            contents: read
+        steps:
+          - uses: wechuli/allcheckspassed@v1
+            with:
+                delay: '3'
+                retries: '30'
```
morrisonlevi, line 17: 30 retries?

danielsn: It polls once a minute to see if the PR has succeeded, up to `retries` times.  Our seems to take a bit less than 20 minutes: 30 seemed like a reasonable buffer

.github/workflows/all-checks.yml
```diff
@@ -0,0 +1,18 @@
+name: All checks pass
```
bantonsson, line 1: Could this be named `Required`, since we might have checks that we would like to exclude at some point.

bantonsson (approved):
Very nice. Only minor comment about workflow name.

PR #491, opened 2024-06-15
Add support for sending v0.4 traces.

hoolioh:
# What does this PR do?

Add support for sending v0.4 encoded traces to the agent.

### DIFF
--- a/trace-utils/src/send_data/mod.rs
+++ b/trace-utils/src/send_data/mod.rs
@@ -7,6 +7,7 @@ pub mod send_data_result;
 pub use crate::send_data::retry_strategy::{RetryBackoffType, RetryStrategy};
 
 use crate::trace_utils::{SendDataResult, TracerHeaderTags};
+use crate::tracer_payload::TracerPayloadCollection;
 use anyhow::{anyhow, Context};
 use bytes::Bytes;
 use datadog_trace_protobuf::pb::{AgentPayload, TracerPayload};
@@ -18,7 +19,9 @@ use hyper::{Body, Client, HeaderMap, Method, Response};
 use std::collections::HashMap;
 
 const DD_API_KEY: &str = "DD-API-KEY";
+
 const HEADER_DD_TRACE_COUNT: &str = "X-Datadog-Trace-Count";
+
 const HEADER_HTTP_CTYPE: &str = "Content-Type";
 const HEADER_CTYPE_MSGPACK: &str = "application/msgpack";
 const HEADER_CTYPE_PROTOBUF: &str = "application/x-protobuf";
@@ -74,13 +77,15 @@ pub(crate) enum RequestResult {
 ///     SendData,
 /// };
 /// use datadog_trace_utils::trace_utils::TracerHeaderTags;
+/// use datadog_trace_utils::tracer_payload::TracerPayloadCollection;
 /// use ddcommon::Endpoint;
 /// use std::time::Duration;
 ///
 /// #[cfg_attr(miri, ignore)]
 /// async fn update_send_results_example() {
 ///     let size = 100;
-///     let tracer_payload = TracerPayload::default(); // Replace with actual payload
+///     let tracer_payload = TracerPayloadCollection::V07(
+///         vec![TracerPayload::default()]); // Replace with actual payload
 ///     let tracer_header_tags = TracerHeaderTags::default(); // Replace with actual header tags
 ///     let target = Endpoint::default(); // Replace with actual endpoint
 ///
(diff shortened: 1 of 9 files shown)


trace-utils/src/tracer_payload.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// v0.4 endoding (TracerPayloadV04).
+    V04,
+    /// v0.7 endoding (TracerPayload).
+    V07,
+}
+
+#[derive(Debug, Clone)]
+/// Enum representing a general abstraction for a collection of tracer payloads.
+/// ```
+pub enum TracerPayloadCollection {
```
hoolioh, line 24: I've used an enum to encapsulate the abstraction for using both encoding types in SendData. I thought about using generics and traits but this way seemed less convoluted. Let me know what you think.

trace-utils/src/send_data/mod.rs
```diff
@@ -187,7 +192,7 @@ impl SendData {
     /// # Returns
     ///
     /// A `SendDataResult` instance containing the result of the operation.
-    pub async fn send(self) -> SendDataResult {
+    pub async fn send(&self) -> SendDataResult {
         if self.use_protobuf() {
```
hoolioh, line 196: This part seemed a little bit odd because we're using the api key in the endpoint to distinguish the method to send the payload. Would it be better to match the collection type? WDYT?

trace-utils/src/send_data/mod.rs
```diff
@@ -337,53 +342,72 @@ impl SendData {
...
-                Err(e) => return result.error(anyhow!(e)),
-            };
-            futures.push(self.send_payload(
-                HEADER_CTYPE_MSGPACK,
-                payload,
-                chunks,
-                additional_payload_headers,
-            ));
+
+        match &self.tracer_payloads {
```
hoolioh, line 379: Here I've kept the compatibility to send v07 payloads serialized as a msgpack. Is something we want to do? sending v07 to the agent? WDYT?

trace-utils/src/send_data/mod.rs
pierotibou: On each review I did, using the api key was stressing me. I needed to check twice that we cannot be in a case where we set the api key inadvertedly and use agentless. Overall, I'd rather have a clear parameter `isAgentless`. That said I don't think it's in the scope of that PR.

trace-utils/src/send_data/mod.rs
pierotibou: > Is something we want to do?

We don't plan ever doing this honestly

trace-utils/src/send_data/mod.rs
```diff
@@ -473,18 +497,30 @@ mod tests {
...
+            }
+            _ => 0,
+        }
     }
 
-    fn rmp_compute_payload_len(payload: &Vec<TracerPayload>) -> usize {
-        let mut total: usize = 0;
-        for payload in payload {
-            total += rmp_serde::to_vec_named(payload).unwrap().len();
+    fn rmp_compute_payload_len(collection: &TracerPayloadCollection) -> usize {
```
pierotibou, line 511: what `rmp` means?

pierotibou (approved):
LGTM, but you should get another review honestly

trace-utils/src/send_data/mod.rs
hoolioh: rmp stands for rust messagepack (https://docs.rs/rmp-serde/latest/rmp_serde/index.html)

trace-utils/src/tracer_payload.rs
bantonsson: I think this is a straight forward way to represent it. We can always revisit it if we feel the need for generics.

bantonsson (approved):

trace-utils/src/send_data/mod.rs
ekump: +100

trace-utils/src/tracer_payload.rs
```diff
@@ -0,0 +1,186 @@
...
+pub enum TraceEncoding {
+    /// v0.4 endoding (TracerPayloadV04).
+    V04,
+    /// v0.7 endoding (TracerPayload).
+    V07,
+}
+
+#[derive(Debug, Clone)]
+/// Enum representing a general abstraction for a collection of tracer payloads.
+/// ```
```
ekump, line 23: Some of your doc comments have ///``` at the bottom even though there are no code examples.

trace-utils/src/tracer_payload.rs
```diff
@@ -0,0 +1,186 @@
...
+
+type TracerPayloadV04 = Vec<Span>;
+
+#[derive(Debug, Clone)]
+/// Enumerates the different encoding types.
+/// ```
+pub enum TraceEncoding {
+    /// v0.4 endoding (TracerPayloadV04).
+    V04,
+    /// v0.7 endoding (TracerPayload).
```
ekump, line 17:
```suggestion
    /// v0.7 encoding (TracerPayload).
```

ekump (approved):

PR #499, opened 2024-06-21
Encode the file path in url authority so we don't crash on whitespaces

paullegranddc:
# What does this PR do?

Encode the file path in url authority so we don't crash on whitespaces

# Motivation

The telemetry lib would fail because of a panic on URI violation, because we were trying to encode the file path in the url path, which doesn't accept some characters.

This PR encodes the path as hex, like we do for unix socket and windows pipe.

# Additional Notes

# How to test the change?

### DIFF
--- a/bin_tests/tests/crashtracker_bin_test.rs
+++ b/bin_tests/tests/crashtracker_bin_test.rs
@@ -226,12 +226,13 @@ fn setup_test_fixtures<'a>(crates: &[&'a ArtifactsBuild]) -> TestFixtures<'a> {
     let artifacts = build_artifacts(crates).unwrap();
 
     let tmpdir = tempfile::TempDir::new().unwrap();
+    let dirpath = tmpdir.path();
     TestFixtures {
-        crash_profile_path: extend_path(tmpdir.path(), "crash"),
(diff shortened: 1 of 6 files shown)


bantonsson (commented):
So why don't we URL encode the path with `%20` for space etc?

chrisnas:
> So why don't we URL encode the path with `%20` for space etc?
Would %20 work for a file path on Windows?

Since I found this issue by calling ddog_endpoint_from_url(), would it be possible to add a test to validate that it does not return a null pointer?

paullegranddc:
> So why don't we URL encode the path with `%20` for space etc?

I reused the existing method to encode paths in urls. In the end it doesn't really matter because it's just an internal representation.
As to why we use the authority instead of the path of the uri to smuggle the path for unix sockets and windows, it was because the authority of hyper::Uri needs to be non empty anyway, and I was afraid the authority was used in hyper client to implement connection reuse, so a connection to a socket would be reused to connect to another path.

bantonsson (approved):
Thanks for the explanation.

yahya-mouman (approved):

PR #500, opened 2024-06-24
Skip serialization of field Span.error if equals to 0

iamluc:
# What does this PR do?

It skips the serialization of the field `Span.error` if the value is 0 as it's done by tracers (at least [.net](https://github.com/DataDog/dd-trace-dotnet/blob/f4ba8212dc62f8a61e947086202ba21f2fdd0bb2/tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs#L199) and [PHP](https://github.com/DataDog/dd-trace-php/blob/master/ext/serializer.c#L1360-L1362))

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/trace-protobuf/Cargo.toml
+++ b/trace-protobuf/Cargo.toml
@@ -10,6 +10,8 @@ license.workspace = true
 prost = "0.11.6"
 serde = { version = "1.0.145", features = ["derive"] }
 serde_bytes = "0.11.9"
+
+[dev-dependencies]
 serde_json = "1.0.117"
 
 [build-dependencies]
--- a/trace-protobuf/Cargo.toml
+++ b/trace-protobuf/Cargo.toml
@@ -10,6 +10,7 @@ license.workspace = true
 prost = "0.11.6"
 serde = { version = "1.0.145", features = ["derive"] }
 serde_bytes = "0.11.9"
+serde_json = "1.0.117"
 
 [build-dependencies]
 prost-build = { version = "0.11.9", optional = true  }
--- a/trace-protobuf/build.rs
+++ b/trace-protobuf/build.rs
@@ -63,6 +63,10 @@ fn generate_protobuf() {
         ".pb.Span.spanLinks",
         "#[serde(skip_serializing_if = \"::prost::alloc::vec::Vec::is_empty\")]",
     );
+    config.field_attribute(
+        ".pb.Span.error",
+        "#[serde(skip_serializing_if = \"is_default\")]",
+    );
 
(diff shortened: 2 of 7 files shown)


iamluc:
> Can we easily add a utest?

Done!

bantonsson (approved):

trace-protobuf/Cargo.toml
```diff
@@ -10,6 +10,7 @@ license.workspace = true
 prost = "0.11.6"
 serde = { version = "1.0.145", features = ["derive"] }
 serde_bytes = "0.11.9"
+serde_json = "1.0.117"
```
hoolioh, line 13:
```suggestion

[dev-dependencies]
serde_json = "1.0.117"
```

hoolioh (changes requested):

trace-protobuf/build.rs
```diff
@@ -63,6 +63,10 @@ fn generate_protobuf() {
         ".pb.Span.spanLinks",
         "#[serde(skip_serializing_if = \"::prost::alloc::vec::Vec::is_empty\")]",
     );
+    config.field_attribute(
+        ".pb.Span.error",
+        "#[serde(skip_serializing_if = \"is_default\")]",
```
bwoebi, line 68: You can use ::i32::is_zero instead.

iamluc: It fails with 
```
error[E0425]: cannot find function `is_zero` in module `std::i32`
   --> libdatadog/trace-protobuf/src/pb.rs:115:35
    |
115 |     #[serde(skip_serializing_if = "::std::i32::is_zero")]
    |      
```

bwoebi: Hm. Well, try `i32::is_zero` without the std stuff.

iamluc: Same.

At first try, I used `is_negative` (https://moshg.github.io/rust-std-ja/std/primitive.i32.html#method.is_negative), but it gave me an other error:
```
error[E0308]: mismatched types
   --> trace-protobuf/src/pb.rs:57:23
    |
57  | #[derive(Deserialize, Serialize)]
    |                       ^^^^^^^^^ expected `i32`, found `&i32`
...
114 |     #[serde(skip_serializing_if = "i32::is_negative")]
    |                                   ------------------ arguments to this function are incorrect
```

hoolioh (approved):

hoolioh (approved):

PR #505, opened 2024-06-25
Support v0.5 trace endpoint

astuyve:
# What does this PR do?
- Adds a few tiny lambda-related enums
- Adds support for `/v0.5/traces` which are string-compressed trace chunks.

# Motivation
- I needed it

# Additional Notes
- I've tested this with python v0.5

# How to test the change?
- automated tests should be sufficient


trace-utils/src/config_utils.rs
```diff
@@ -10,6 +10,10 @@ const TRACE_INTAKE_ROUTE: &str = "/api/v0.2/traces";
 const TRACE_STATS_INTAKE_ROUTE: &str = "/api/v0.2/stats";
 
 pub fn read_cloud_env() -> Option<(String, trace_utils::EnvironmentType)> {
+    if let Ok(res) = env::var("AWS_LAMBDA_FUNCTION_NAME") {
+        // Set by Google Cloud Functions for older runtimes
```
astuyve, line 14: lol copilot

trace-utils/src/trace_utils.rs
```diff
@@ -40,6 +45,210 @@ pub async fn get_traces_from_request_body(body: Body) -> anyhow::Result<(usize,
...
+        read_array_len(reader).map_err(|err| anyhow!("Error reading dict size: {err}"))?;
+    if dict_size as i32 > 25 * 1e6 as i32 {
+        anyhow::bail!(
+            "Error deserializing strings dictionary. Dict size is too large: {dict_size}"
+        );
+    }
+    let mut dict: Vec<String> = Vec::with_capacity(dict_size.try_into()?);
+    for _ in 0..dict_size {
+        let val: Value = read_value(reader)?;
+        match val {
```
duncanista, line 60:
```suggestion
        match read_value(reader)? {
```
Would this compile?

trace-utils/src/trace_utils.rs
```diff
@@ -40,6 +45,210 @@ pub async fn get_traces_from_request_body(body: Body) -> anyhow::Result<(usize,
...
+    }
+    Ok(dict)
+}
+
+#[inline]
+fn get_v5_span(reader: &mut Reader<impl Buf>, dict: &[String]) -> anyhow::Result<Span> {
+    let mut span: Span = Default::default();
+    let span_size = rmp::decode::read_array_len(reader)
+        .map_err(|err| anyhow!("Error reading span size: {err}"))? as usize;
+    if span_size != 12 {
```
duncanista, line 76: 12 into constant too?

trace-utils/src/trace_utils.rs
```diff
@@ -40,6 +45,210 @@ pub async fn get_traces_from_request_body(body: Body) -> anyhow::Result<(usize,
...
+        anyhow::bail!("Expected an array of exactly 12 elements in a span, got {span_size}");
+    }
+    //0 - service
```
duncanista, line 79: So values come in order and I imagine that's why we are doing 12 reads here?

trace-utils/src/trace_utils.rs
```diff
@@ -40,6 +45,210 @@ pub async fn get_traces_from_request_body(body: Body) -> anyhow::Result<(usize,
...
+        }
+        val => anyhow::bail!("Error reading span duration, value is not an integer: {val}"),
+    };
+    //8 - error
+    match read_value(reader)? {
+        Value::Integer(i) => {
+            span.error = i
+                .as_i64()
+                .ok_or_else(|| anyhow!("Error reading span error, value is not an integer: {i}"))?
+                as i32;
```
duncanista, line 151: Is the `as i32` expected here in error? If so, should we add a comment on why we do this here and not for the other ones?

duncanista (approved):
LGTM – saw the e2e test, should we add individual tests?

trace-utils/src/trace_utils.rs
```diff
@@ -450,6 +661,88 @@ mod tests {
...
+                    u64,
+                    u64,
+                    u64,
+                    i64,
+                    i64,
+                    i32,
+                    HashMap<u8, u8>,
+                    HashMap<u8, f64>,
+                    u8,
+                )>,
```
morrisonlevi, line 683: I'm not familiar with the domain here--is there a struct/proto or something I can look at to see what each slot of the tuple means? At a first glance, I am inclined to agree with `clippy::type_complexity`, although I realize this is just in a test.

trace-utils/src/trace_utils.rs
astuyve: Yup, msgpack is self describing but as far as I know all fields are required here. Maybe Bjorn can confirm?

trace-utils/src/trace_utils.rs
astuyve: Weird to read on github, but yeah an i32 is required. I don't think a comment is necessary because the compiler complains with out it.

Here's the i32: https://github.com/DataDog/libdatadog/blob/main/trace-protobuf/src/pb/span.proto#L49

trace-utils/src/trace_utils.rs
astuyve: According to the docs, all are required: https://github.com/DataDog/datadog-agent/blob/579d60cbfc8635559338b19fa855bf59b6d10f00/pkg/trace/api/version.go#L110

trace-utils/src/trace_utils.rs
astuyve: great question, I wish I had one!
What I did find was this [documentation](https://github.com/DataDog/datadog-agent/blob/579d60cbfc8635559338b19fa855bf59b6d10f00/pkg/trace/api/version.go#L110) and finally this [test](https://github.com/DataDog/datadog-agent/blob/main/pkg/proto/pbgo/trace/decoder_v05_test.go).

I used the same fields to create the test to confirm this all worked (as well as implementing it using the python tracer).

trace-utils/src/trace_utils.rs
```diff
@@ -40,6 +45,210 @@ pub async fn get_traces_from_request_body(body: Body) -> anyhow::Result<(usize,
...
+    if span_size != 12 {
+        anyhow::bail!("Expected an array of exactly 12 elements in a span, got {span_size}");
+    }
+    //0 - service
+    match read_value(reader)? {
+        Value::Integer(s) => {
+            span.service = str_from_dict(dict, s)?;
+        },
+        val => anyhow::bail!("Error reading span service, value is not an integer and can't be looked up in dict: {val}")
+    };
```
bantonsson, line 85: This repeated code could be moved out to
```rust
fn get_v05_string(reader: &mut Reader<impl Buf>, dict: &[String], field_name: &str) -> anyhow::Result<String> {
    match read_value(reader)? {
        Value::Integer(s) => {
            str_from_dict(dict, s)
        },
        val => anyhow::bail!("Error reading {field_name}, value is not an integer and can't be looked up in dict: {val}")
    }
}
```

trace-utils/src/trace_utils.rs
bantonsson: That doc comment is unfortunately the reference documentation.

trace-utils/src/trace_utils.rs
```diff
@@ -40,6 +45,210 @@ pub async fn get_traces_from_request_body(body: Body) -> anyhow::Result<(usize,
...
+        anyhow::bail!(
+            "Error deserializing strings dictionary. Dict size is too large: {dict_size}"
+        );
+    }
+    let mut dict: Vec<String> = Vec::with_capacity(dict_size.try_into()?);
+    for _ in 0..dict_size {
+        let val: Value = read_value(reader)?;
+        match val {
+            Value::String(s) => {
+                let s = s.to_string().replace(&['\"'][..], "");
```
bantonsson, line 62: This feels so weird. Are the strings escaped this way by the Python tracer, or is it the `rmpv` library that does this (which would be really strange). I'm pretty sure that the Java Tracer does not do this, but things like AppSec and CI Visibility would send escaped JSON in fields. It should at least have a comment.

bantonsson (changes requested):
Looks good. Only minor comments (and some nitpicks 😉)

trace-utils/src/trace_utils.rs
```diff
@@ -40,6 +45,210 @@ pub async fn get_traces_from_request_body(body: Body) -> anyhow::Result<(usize,
     Ok((size, traces))
 }
 
+#[inline]
+fn get_v5_strings_dict(reader: &mut Reader<impl Buf>) -> anyhow::Result<Vec<String>> {
+    let dict_size =
+        read_array_len(reader).map_err(|err| anyhow!("Error reading dict size: {err}"))?;
+    if dict_size as i32 > 25 * 1e6 as i32 {
```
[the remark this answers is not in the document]
astuyve: I'm not sure, this is from the go implementation so I assume it's canonical: https://github.com/DataDog/datadog-agent/blob/main/pkg/proto/pbgo/trace/decoder_v05.go#L223

trace-utils/src/trace_utils.rs
astuyve: Yup fixed, just shouldn't use `to_string()` on the `Value::String`

trace-utils/src/trace_utils.rs
astuyve: Done

trace-utils/src/trace_utils.rs
astuyve: Yup, good call.

trace-utils/src/trace_utils.rs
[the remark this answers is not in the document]
bantonsson: I agree. Let's stick with the Agent limits, even if I think they are confusing number of entries with payload size (I mean 25 million array entries is _huge_ 🤷🏼‍♂️).

trace-utils/src/trace_utils.rs
```diff
@@ -40,6 +47,209 @@ pub async fn get_traces_from_request_body(body: Body) -> anyhow::Result<(usize,
...
+) -> anyhow::Result<String> {
+    match read_value(reader)? {
+        Value::Integer(s) => {
+            str_from_dict(dict, s)
+        },
+        val => anyhow::bail!("Error reading {field_name}, value is not an integer and can't be looked up in dict: {val}")
+    }
+}
+
+pub async fn v5_get_traces_from_request_body(
```
bantonsson, line 224: Almost all `v5` to `v05` 😉

bantonsson (approved):

trace-utils/src/trace_utils.rs
astuyve: oops, fixed!

PR #458, opened 2024-05-28
Paullgdc/data pipelines/stats computation

paullegranddc:
# What does this PR do?

A brief description of the change being made with this pull request.

# Motivation

What inspired you to submit this pull request?

# Additional Notes

Anything else we should know when reviewing?

# How to test the change?

Describe here in detail how the change can be validated.

PR #506, opened 2024-06-26
Use inplace normalization and add benchmark

VianneyRuhlmann:
# What does this PR do?

Work from @paullegranddc in #458
- Use inplace normalization to improve performance
- Handle chars as uint instead of utf8 chars to improve performance
- Add benchmark for trace-normalization  

# Motivation

Improve trace normalization performance

# Additional Notes

# How to test the change?


VianneyRuhlmann:
Benchmark summary for performance compared to `main`:
- **normalization/normalize_service/normalize_service/test_ASCII**
                        time:   [24.606 ns 24.855 ns 25.198 ns]
                        change: [-54.072% **-53.559%** -53.003%] (p = 0.00 < 0.05)
                        Performance has improved.
- **normalization/normalize_service/normalize_service/Test Conversion 0f Weird !@#$%^&**() Characters**
                        time:   [141.12 ns 141.49 ns 141.91 ns]
                        change: [+10.668% **+11.138%** +11.597%] (p = 0.00 < 0.05)
                        Performance has regressed.
- **normalization/normalize_service/normalize_service/Data🐨dog🐶 繋がっ⛰てて**
                        time:   [338.55 ns 339.26 ns 340.03 ns]
                        change: [+67.316% **+68.103%** +68.841%] (p = 0.00 < 0.05)
                        Performance has regressed.
- normalization/normalize_service/normalize_service/A0000000000000000000000000000000000000000000000000...
                        time:   [191.64 ns 192.19 ns 192.84 ns]
                        change: [-41.650% **-41.457%** -41.253%] (p = 0.00 < 0.05)
                        Performance has improved.
- normalization/normalize_name/normalize_name/good
                        time:   [10.918 ns 10.950 ns 10.984 ns]
                        change: [-63.638% **-63.478%** -63.328%] (p = 0.00 < 0.05)
                        Performance has improved.
- normalization/normalize_name/normalize_name/bad-name
                        time:   [15.268 ns 15.300 ns 15.336 ns]
                        change: [-77.047% **-76.969%** -76.897%] (p = 0.00 < 0.05)
                        Performance has improved.
- normalization/normalize_name/normalize_name/Too-Long-.Too-Long-.Too-Long-.Too-Long-.Too-Long-.Too-Lo...
                        time:   [123.12 ns 123.63 ns 124.15 ns]
                        change: [-76.671% **-76.584%** -76.499%] (p = 0.00 < 0.05)
                        Performance has improved.

The regression for the empty service name case is probably due to the fact that in `main` the default value is written by `normalize_span` instead of `normalize_service`.
There is also a regression for string using non-ASCII characters, it seems to be caused by the handling of strings as bytes instead of using standard strings.

trace-normalization/src/normalize_utils.rs
```diff
@@ -1,332 +1,367 @@
...
+use std::time::SystemTime;
+
+// MAX_TYPE_LEN the maximum size for a span type
+pub(crate) const MAX_TYPE_LEN: usize = 100;
+/// an arbitrary cutoff to spot weird-looking values
+/// nanoseconds since epoch on Jan 1, 2000
+const YEAR_2000_NANOSEC_TS: i64 = 946684800000000000;
+/// DEFAULT_SPAN_NAME is the default name we assign a span if it's missing and we have no reasonable
+/// fallback
+pub(crate) const DEFAULT_SPAN_NAME: &str = "unnamed_operation";
```
bantonsson, line 13: Where does this come from? I found this ancient [RFC](https://github.com/Datadog/architecture/blob/master/rfcs/apm/agent/trace-agent-normalization/rfc.md) saying `unnamed-operation`.

trace-normalization/src/normalize_utils.rs
```diff
@@ -1,332 +1,367 @@
...
 }
 
-pub(crate) fn is_alpha_num(c: char) -> bool {
-    is_alpha(c) || c.is_ascii_digit()
+// truncate_utf8 truncates the given string to make sure it uses less than limit bytes.
+// If the last character is a utf8 character that would be split, it removes it
+// entirely to make sure the resulting string is not broken.
+pub(crate) fn truncate_utf8(s: &mut String, limit: usize) {
+    let boundary = crate::utf8_helpers::floor_char_boundary(s, limit);
+    s.truncate(boundary);
```
bantonsson, line 278: Is `s.truncate(...)` a no-op if the boundary is the length of `s`?

bantonsson (approved):
Looks good. Only some small questions.

trace-normalization/src/normalize_utils.rs
VianneyRuhlmann: The [agent](https://github.com/DataDog/datadog-agent/blob/c746874a5c0e7c1745a5d10d87d0847721fd54b1/pkg/trace/traceutil/normalize.go#L18) and most [tracers](https://github.com/search?q=org%3ADataDog+unnamed+operation&type=code) seem to be using `unnamed_operation`. Although I agree that `unnamed-operation` makes more sense.

trace-normalization/src/normalize_utils.rs
VianneyRuhlmann: [Almost](https://doc.rust-lang.org/src/alloc/string.rs.html#1406) it is checking if the boundary is higher or equal than the length. And since all the functions used are inline, the compiler might be reusing the call to `len` from `floor_char_boundary`.

trace-normalization/src/normalize_utils.rs
bantonsson: Let's keep doing what we have been doing, and use `unnamed_operation`.

PR #507, opened 2024-06-26
[Data-Pipeline] Add stats computation

VianneyRuhlmann:
# What does this PR do?

Work from @paullegranddc in #458
- Add `stats_exporter` to receive, aggregate and send span stats
- Move try_c to `ddcommon_ffi`

This is a first implementation which still have some caveats:
- We only store one time bucket and all SpanStats received are aggregated in this bucket, this requires the client to call `send` regularly to avoid a time bucket too wide.
- We don't aggregate on `peer.tags`
- We currently have two different exporter one for sampled spans and another for stats. The sampled spans need to be sent to both. To merge both exporter we need 
   1) a way to send partial spans for the unsampled spans, to avoid serialization of fields not used for stats computation
   2) replace the synchrone send of the spans exporter, to be able to aggregate the stats on time interval chosen by 
        libdatadog instead of the client span submission rate.

# Motivation

Moving stats computation to libdatadog would allow all tracers to compute stats locally and only send sampled spans to the agent. This will reduce traffic to the agent and avoid serialization of data which are going to be dropped in the agent

## Drawbacks
The current implementation doesn't reduce the amount of serialization since we need to serialize spans to send them from the tracer to libdatadog.

# Additional Notes

# How to test the change?

You can modify and run the example to send stats to a test agent


ddcommon-ffi/src/error.rs
```diff
@@ -111,3 +111,15 @@ pub type MaybeError = crate::Option<Error>;
 
 #[no_mangle]
 pub extern "C" fn ddog_MaybeError_drop(_: MaybeError) {}
+
+#[macro_export]
+macro_rules! try_c {
```
ekump, line 116: Does this have to be a macro? Would something like this work instead?
```rust
pub fn try_c<T, E: std::fmt::Debug>(failable: Result<T, E>) -> Result<T, MaybeError> {
    match failable {
        Ok(o) => Ok(o),
        Err(e) => Err(MaybeError::Some(Error::from(format!("{:?}", e)))),
    }
}
```

bwoebi: The point of try_c() is to return within a function returning MaybeError. The conversion itself, to MaybeError::Some isn't the important part here.

ddcommon/src/tag.rs
```diff
@@ -130,6 +130,10 @@ impl Tag {
 
         Tag::from_value(format!("{key}:{value}"))
     }
+
+    pub fn into_string(self) -> String {
```
ekump, line 134: Any reason this can't be:

```rust
impl ToString for Foo {
    fn to_string(&self) -> String {
        self.value.to_string()
    }
}
```

ddcommon-ffi/src/error.rs
ekump: @bwoebi I'm not sure I understand. My comment is asking if we _need_ to use a macro here versus a generic function. Personally, I'm not a fan of macros unless they are truly necessary / really the best option.

data-pipeline/Cargo.toml
```diff
@@ -16,11 +16,15 @@ log = "0.4"
...
 tokio = {version = "1.23", features = ["rt"], default-features = false}
+http = "0.2"
 
 ddcommon = { path = "../ddcommon" }
 datadog-trace-protobuf = { path = "../trace-protobuf" }
 datadog-trace-utils = { path = "../trace-utils" }
 datadog-trace-normalization = { path = "../trace-normalization" }
+datadog-trace-obfuscation = { path = "../trace-obfuscation" }
+datadog-ddsketch = { path = "../ddsketch" }
+httpmock = "0.7.0"
```
ekump, line 27: this should either be in dev.dependencies or under a feature like in [trace-utils](https://github.com/DataDog/libdatadog/blob/main/trace-utils/Cargo.toml#L35).

data-pipeline-ffi/src/stats_exporter.rs
```diff
@@ -0,0 +1,121 @@
...
+/// * `hostname` - The tracer hostname
+/// * `env` - env tag, used for aggregation
+/// * `version` - version tag, used for aggregation
+/// * `lang` - The language of the client library.
+/// * `tracer_version` - The version of the client library.
+/// * `runtime_id` - Id used by the agent to identify uniquely a source
+/// * `service` - service name
+/// * `git_commit_sha` - A git commit sha set through source code integration
+/// * `stats_computation_interval_seconds` - the size of stats buckets in seconds
+/// * `request_timeout_ms` - request timeout in ms, no temout if zero
```
ajgajg1134, line 28:
```suggestion
/// * `request_timeout_ms` - request timeout in ms, no timeout if zero
```

data-pipeline-ffi/src/stats_exporter.rs
```diff
@@ -0,0 +1,121 @@
...
+};
+use ddcommon::{parse_uri, Endpoint};
+use ddcommon_ffi as ffi;
+use ffi::slice::AsBytes;
+
+/// Create a new StatExporter instance.
+///
+/// # Arguments
+///
+/// * `out_handle` - The handle to write the TraceExporter instance in.
```
ajgajg1134, line 17:
```suggestion
/// * `out_handle` - The handle to write the StatsExporter instance in.
```

data-pipeline-ffi/src/stats_exporter.rs
```diff
@@ -0,0 +1,121 @@
...
+use std::{ptr::NonNull, time};
+
+use data_pipeline::stats_exporter::{
+    blocking::StatsExporter, stats_url_from_agent_url, Configuration, LibraryMetadata, SpanStats,
+};
+use ddcommon::{parse_uri, Endpoint};
+use ddcommon_ffi as ffi;
+use ffi::slice::AsBytes;
+
+/// Create a new StatExporter instance.
```
ajgajg1134, line 13:
```suggestion
/// Create a new StatsExporter instance.
```

data-pipeline-ffi/src/stats_exporter.rs
```diff
@@ -0,0 +1,121 @@
...
+    exporter: &StatsExporter,
+    resource_name: ffi::CharSlice,
+    service_name: ffi::CharSlice,
+    operation_name: ffi::CharSlice,
+    span_type: ffi::CharSlice,
+    http_status_code: u32,
+    is_synthetics_request: bool,
+    is_top_level: bool,
+    is_error: bool,
+    duration: u64,
```
ajgajg1134, line 94: Do we need to document the units here for `duration`?

data-pipeline-ffi/src/stats_exporter.rs
```diff
@@ -0,0 +1,121 @@
...
+        is_synthetics_request,
+        is_top_level,
+        is_error,
+        duration,
+    })
+}
+
+#[no_mangle]
+#[allow(clippy::missing_safety_doc)]
+/// Flush stats form a stats exporter and send them to the agent
```
ajgajg1134, line 111:
```suggestion
/// Flush stats from a stats exporter and send them to the agent
```

data-pipeline/src/stats_exporter.rs
```diff
@@ -0,0 +1,600 @@
...
+    /// optional timeout for sending stats
+    pub request_timeout: Option<time::Duration>,
+    /// endpoint used to send stats to the agent
+    pub endpoint: Endpoint,
+}
+
+/// An exporter aggregating stats from traces and sending them to the agent
+///
+/// Currently we only keep one time bucket starting at the time of the exporter creation and
+/// resetting to current time on flush. All `SpanStats` sent between flushesare added to the same
```
ajgajg1134, line 120:
```suggestion
/// resetting to current time on flush. All `SpanStats` sent between flushes are added to the same
```

data-pipeline/src/stats_exporter.rs
```diff
@@ -0,0 +1,600 @@
...
+/// bucket.
+/// This raises two issues:
+/// - We expect SpanStats to be submitted right after the span ended (since the aggregation is done
+///   on endTime)
+/// - We expect the tracer to call send when we reach start_time + bucket_duration to make sure the
```
ajgajg1134, line 125: Is this a reasonable / safe assumption to make for users of this library? Is there any reason we avoided doing this flushing on the library side? complexity with FFI / wanting to avoid buffering too many stats buckets?

ajgajg1134 (commented):
Mostly just some small typos and questions from me (since I'm very not familiar with this code base :P ) I haven't finished reviewing all of `stats_exporter.rs` but wanted to post this initial review before the weekend :)

data-pipeline/src/stats_exporter.rs
```diff
@@ -0,0 +1,600 @@
...
+        } else {
+            let _ = self.ok_summary.add(span_stats.duration as f64);
+        }
+        if span_stats.is_top_level {
+            self.top_level_hits += 1;
+        }
+    }
+}
+
+/// Metadata required in a ClientStatsPayload
```
ekump, line 60: I think the rustdoc comments for public functions and structs should be more robust and follow the patterns we've been using recently in trace-utils and the sidecar. Let me know if you need specific examples, or a hand writing them.

PR #512, opened 2024-06-27
[Serverless Mini Agent] Add Azure App Service Runtime Tags

duncanpharvey:
# What does this PR do?

Adds the following span tags to traces sent by the Serverless Mini Agent in Azure Functions.

* `aas.environment.runtime`
* `aas.environment.runtime_version`
* `aas.environment.function_runtime`

# Motivation

https://datadoghq.atlassian.net/browse/SVLS-4935

* Usage tracking for mini agent private beta
* Span tag parity with Datadog Agent

# Additional Notes

# How to test the change?

- Build mini agent locally
- Deploy Azure Function with mini agent binary in root path
- Set `DD_MINI_AGENT_PATH`
  - Azure: `/home/site/wwwroot/datadog-serverless-trace-mini-agent`


alexgallotta (approved):
Looks good! I would add a test with the failing case to make sure it works

ddcommon/src/azure_app_services.rs
```diff
@@ -222,6 +237,18 @@ impl AzureMetadata {
     pub fn get_site_kind(&self) -> &str {
         self.site_kind.as_str()
     }
+
+    pub fn get_runtime(&self) -> &str {
+        get_value_or_unknown!(self.runtime)
```
bantonsson, line 242: Couldn't this filtering be moved to when the strings are read out of the query?

duncanpharvey: It could but the idea was to map empty strings to `unknown` for all of the fields that are read with `get_value_or_unknown`, not just the runtime related fields.

bantonsson: Since the fields of the `AzureMetadata` struct are `private` and the only way to get them out is via the `get_*` methods, that are the only users of `get_value_or_unknown!`, you could as well do the filtering for all the fields during construction.

duncanpharvey: I see, so the idea is to handle the empty string case on construction so we aren't doing the same operation every time we call `get_*`. I moved the empty string handling to `get_trimmed_env_var`. Let me know if that looks good to you!

https://github.com/DataDog/libdatadog/blob/62704d036e0463ce5e6e2dbb5c30750ba81183b7/ddcommon/src/azure_app_services.rs#L27-L34

bantonsson: Wait, I just realized that `get_trimmed_env_var` is not being used in most of the code, which use `query.get_var`, so that is probably not the right place to put it.

duncanpharvey: `get_trimmed_env_var` is used in the implementation of `query.get_var`, right? And `query.get_var` is used in `build_metadata` to get the environment variables. So I think everywhere we would want to filter out empty strings is handled with the filter in `get_trimmed_env_var`.

https://github.com/DataDog/libdatadog/blob/62704d036e0463ce5e6e2dbb5c30750ba81183b7/ddcommon/src/azure_app_services.rs#L61-L65

bantonsson: It is indeed. Completely missed it.

PR #513, opened 2024-06-27
Add option to skip checks when building profiling

p-datadog:
# What does this PR do?

Adds -T option to build-profiling-ffi.sh script to skip post-build checks, which is frequently done during development.

Also adds a usage blurb which can be printed with -h option, and prints the usage blurb if the required positional parameter of destination directory is not provided instead of failing with a non-descriptive error.

# Motivation

Current diagnostics are not indicating what is wrong when the positional argument is not supplied.

# Additional Notes

None

# How to test the change?

Run `./build-profiling-ffi.sh -h` or `./build-profiling-ffi.sh -T`.


ivoanjo (approved):
LGTM :+1: 

I know @r1viollet has been working to get rid of this script, but this seems a useful + harmless UX improvement to have until we can kill it off :)

bantonsson (approved):

PR #515, opened 2024-07-01
telemetry worker: flush data after stops, and two other fixes

cataphract:
Telemetry workers are functionally dead after a Stop lifecycle action, provided there's no intervening Start. While AddPoint actions are still processed, their data is never flushed, since the Stop action handler unschedules FlushMetrics and FlushData actions.

PHP sends a Stop action at the end of every request via ddog_sidecar_telemetry_end(), but a Start action is only generated just after a telemetry worker is spawned. With no more Start actions generated, no metrics can effectively be sent after the first Stop.

It is not clear to me whether the intention is to have a Start/Stop pair on every PHP request (where Stop flushes the metrics) or if the intention is to to have only such a pair in the first request, with the Stop event generated by ddog_sidecar_telemetry_end() effectively a noop. It would appear, judging by [this
comment](https://github.com/DataDog/libdatadog/pull/391):

> Also allow the telemetry worker to have a mode where it's continuing
execution after a start-stop cycle, otherwise it won't send any more metrics afterwards.

that the intention is to keep sending metrics after a Start/Stop pair. It also makes more sense, insofar as data is flushed only on the interval, rather than after every request via Stop. In that case:

* The Stop action handler should not unschedule FlushData and FlushMetrics events and
* FlushData, if called outside a Start-Stop pair, should not be a noop.

Finally: swap the order in which FlushData and FlushMetrics are scheduled so that FlushMetrics runs first and therefore its generated data can be sent by the next FlushData.


ddtelemetry/src/worker/mod.rs
```diff
@@ -296,7 +293,9 @@ impl TelemetryWorker {
                     self.log_err(&e);
                 }
                 self.data.started = false;
-                self.deadlines.clear_pending();
+                if !self.config.restartable {
```
bantonsson, line 296: Wouldn't it be enough to only include `self.data.started = false;` inside the if statement as well, and leave the exit early checks?

ddtelemetry/src/worker/mod.rs
```diff
@@ -458,7 +454,9 @@ impl TelemetryWorker {
                 .await;
 
                 self.data.started = false;
-                self.deadlines.clear_pending();
+                if !self.config.restartable {
```
bantonsson, line 457: Same here.

bantonsson (commented):
These booleans are woefully undocumented and I'm not completely sure about the expected life cycle, but from reading the code it looks like your interpretation is correct.

I think that it might be better to leave the worker as `started` if it's `restartable` and leave the checks in place.

ddtelemetry/src/worker/mod.rs
cataphract: I guess it depends. Do we want Start, Stop, Stop to generate two stops? Because that's what PHP ends up generating. The second stop is a noop, but if I moved the assignment `self.data.started = false` under the condition `if !self.config.restartable`, then the stops would be effective

bantonsson: If the exit early checks are still in the code, then how would the second `Stop` be effective?

cataphract: From what I understand, what you're proposing is that I reset `started = false` only if `!restartable`. In that case `started` would stay true forever once there is a start. So the early check

```
 if !self.data.started {
   return BREAK;
 }
```

would never be hit.

bantonsson: Yes, you're right. I was confusing the early checks.

There is something in the logic that feels a bit broken. The `FlushData` is also protected by this `!self.data.started` check. Should it work after a _restartable_ stop?

The things that `Stop` does, should they happen when the first request ends, or when the worker stops?

cataphract:
@bwoebi Can you look at this? I need it for https://github.com/DataDog/dd-trace-php/pull/2735

ddtelemetry/src/worker/mod.rs
```diff
@@ -341,10 +343,10 @@ impl TelemetryWorker {
                         Err(err) => self.log_err(&err),
                     }
                     self.deadlines
-                        .schedule_event(LifecycleAction::FlushData)
+                        .schedule_event(LifecycleAction::FlushMetricAggr)
```
bantonsson, line 346: This is a nice optimization. Maybe add a comment about why the order matters.

ddtelemetry/src/worker/mod.rs
```diff
@@ -474,27 +474,32 @@ impl TelemetryWorker {
...
     // Builds telemetry payloads containing lifecycle events
-    fn build_app_events_batch(&self) -> Vec<Payload> {
+    fn build_app_events_batch(&mut self) -> Vec<Payload> {
         let mut payloads = Vec::new();
 
         if self.data.dependencies.flush_not_empty() {
             payloads.push(data::Payload::AppDependenciesLoaded(
                 data::AppDependenciesLoaded {
-                    dependencies: self.data.dependencies.unflushed().cloned().collect(),
+                    dependencies: self.data.dependencies.drain_unflushed().cloned().collect(),
```
bantonsson, line 483: Shouldn't these already be removed on a successful send by the call to `removed_flushed`? Yeah the method name has a `d` too much.

iamluc (approved):

ddtelemetry/src/worker/mod.rs
cataphract: I've added a new commit that uses the existing method.

ddtelemetry/src/worker/mod.rs
```diff
@@ -429,40 +429,36 @@ impl TelemetryWorker {
...
+                    payloads.push(data::Payload::MessageBatch(obsevability_events));
+                }
+
+                let self_arc = Arc::new(tokio::sync::RwLock::new(self));
+                let futures = payloads.into_iter().map(|payload| {
+                    let self_arc = self_arc.clone();
+                    async move {
+                        // This is different from the non-functional:
+                        // match self_arc.read().await.send_payload(&payload).await { ... }
+                        // presumably because the temp read guard would live till end of match
```
bantonsson, line 444: Interesting find

bantonsson (approved):

PR #518, opened 2024-07-03
Implement timeouts for both sending traces and telemetry.

hoolioh:
# What does this PR do?

This PR adds a new timeout field to the `Endpoint` struct in order to set a timeout for outgoing connections used for sending telemetry and traces.

# Motivation

Prevent situations where holding too many connections could lead to memory pressure or resource exhaustion in the system by dropping them after a period of inactivity.

### DIFF
--- a/bin_tests/src/bin/crashtracker_bin_test.rs
+++ b/bin_tests/src/bin/crashtracker_bin_test.rs
@@ -45,7 +45,7 @@ mod unix {
         } else {
             Some(ddcommon::Endpoint {
                 url: ddcommon::parse_uri(&output_url)?,
-                api_key: None,
+                ..Default::default()
             })
         };
 
--- a/bin_tests/src/bin/crashtracker_bin_test.rs
+++ b/bin_tests/src/bin/crashtracker_bin_test.rs
@@ -13,13 +13,13 @@ fn main() -> anyhow::Result<()> {
 mod unix {
     use anyhow::Context;
     use bin_tests::ReceiverType;
-    use std::{env, fs::File, str::FromStr, time::Duration};
+    use std::{env, fs::File, str::FromStr};
 
     use datadog_crashtracker::{
         self as crashtracker, CrashtrackerConfiguration, CrashtrackerMetadata,
         CrashtrackerReceiverConfig,
     };
-    use ddcommon::tag;
+    use ddcommon::{parse_uri, tag, Endpoint};
 
     #[inline(never)]
     unsafe fn deref_ptr(p: *mut u8) {
@@ -37,13 +37,16 @@ mod unix {
         let socket_path = args.next().context("Unexpected number of arguments")?;
         anyhow::ensure!(args.next().is_none(), "unexpected extra arguments");
 
-        let timeout = Duration::from_secs(30);
         let wait_for_receiver = true;
 
         let endpoint = if output_url.is_empty() {
             None
         } else {
-            Some(ddcommon::Endpoint::from_slice(&output_url))
+            Some(Endpoint {
+                url: parse_uri(&output_url).unwrap(),
+                timeout_ms: 30_000,
+                ..Default::default()
+            })
         };
 
         let config = CrashtrackerConfiguration {
@@ -51,7 +54,6 @@ mod unix {
             create_alt_stack: true,
             resolve_frames: crashtracker::StacktraceCollection::WithoutSymbols,
             endpoint,
-            timeout,
             wait_for_receiver,
         };
 
--- a/bin_tests/src/bin/crashtracker_bin_test.rs
+++ b/bin_tests/src/bin/crashtracker_bin_test.rs
@@ -19,7 +19,7 @@ mod unix {
         self as crashtracker, CrashtrackerConfiguration, CrashtrackerMetadata,
         CrashtrackerReceiverConfig,
     };
-    use ddcommon::{parse_uri, tag, Endpoint};
+    use ddcommon::{tag, Endpoint};
 
     #[inline(never)]
     unsafe fn deref_ptr(p: *mut u8) {
@@ -42,11 +42,7 @@ mod unix {
         let endpoint = if output_url.is_empty() {
             None
         } else {
-            Some(Endpoint {
-                url: parse_uri(&output_url).unwrap(),
-                timeout_ms: 30_000,
-                ..Default::default()
-            })
+            Some(Endpoint::from_slice(&output_url))
         };
 
         let config = CrashtrackerConfiguration {
--- a/bin_tests/src/bin/crashtracker_bin_test.rs
+++ b/bin_tests/src/bin/crashtracker_bin_test.rs
@@ -43,10 +43,7 @@ mod unix {
         let endpoint = if output_url.is_empty() {
             None
         } else {
-            Some(ddcommon::Endpoint {
-                url: ddcommon::parse_uri(&output_url)?,
-                ..Default::default()
-            })
+            Some(ddcommon::Endpoint::from_slice(&output_url))
         };
 
         let config = CrashtrackerConfiguration {
--- a/crashtracker/src/api.rs
+++ b/crashtracker/src/api.rs
@@ -143,7 +143,7 @@ fn test_crash() -> anyhow::Result<()> {
 
     let endpoint = Some(Endpoint {
         url: parse_uri(&output_url).unwrap(),
-        api_key: None,
+        ..Default::default()
     });
 
     let path_to_receiver_binary =
--- a/crashtracker/src/api.rs
+++ b/crashtracker/src/api.rs
@@ -140,11 +140,7 @@ fn test_crash() -> anyhow::Result<()> {
     let dir = "/tmp/crashreports/";
     let output_url = format!("file://{dir}{time}.txt");
 
-    let endpoint = Some(Endpoint {
-        url: ddcommon::parse_uri(&output_url).unwrap(),
-        timeout_ms: 30_000,
-        ..Default::default()
-    });
+    let endpoint = Some(Endpoint::from_slice(&output_url));
 
     let path_to_receiver_binary =
         "/tmp/libdatadog/bin/libdatadog-crashtracking-receiver".to_string();
--- a/crashtracker/src/api.rs
+++ b/crashtracker/src/api.rs
@@ -140,7 +140,11 @@ fn test_crash() -> anyhow::Result<()> {
     let dir = "/tmp/crashreports/";
     let output_url = format!("file://{dir}{time}.txt");
 
-    let endpoint = Some(Endpoint::from_slice(&output_url));
+    let endpoint = Some(Endpoint {
+        url: ddcommon::parse_uri(&output_url).unwrap(),
+        timeout_ms: 30_000,
+        ..Default::default()
+    });
 
     let path_to_receiver_binary =
         "/tmp/libdatadog/bin/libdatadog-crashtracking-receiver".to_string();
@@ -148,7 +152,6 @@ fn test_crash() -> anyhow::Result<()> {
     let resolve_frames = StacktraceCollection::EnabledWithInprocessSymbols;
     let stderr_filename = Some(format!("{dir}/stderr_{time}.txt"));
     let stdout_filename = Some(format!("{dir}/stdout_{time}.txt"));
-    let timeout = Duration::from_secs(30);
     let wait_for_receiver = true;
     let receiver_config = CrashtrackerReceiverConfig::new(
         vec![],
@@ -162,7 +165,6 @@ fn test_crash() -> anyhow::Result<()> {
         create_alt_stack,
         endpoint,
         resolve_frames,
-        timeout,
         wait_for_receiver,
     )?;
     let metadata = CrashtrackerMetadata::new(
--- a/crashtracker/src/api.rs
+++ b/crashtracker/src/api.rs
@@ -132,7 +132,6 @@ pub fn init_with_unix_socket(
 fn test_crash() -> anyhow::Result<()> {
     use crate::{begin_profiling_op, StacktraceCollection};
     use chrono::Utc;
-    use ddcommon::parse_uri;
     use ddcommon::tag;
     use ddcommon::Endpoint;
     use std::time::Duration;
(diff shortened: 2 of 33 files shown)


bwoebi:
Slightly surprised that we opt modifying the Endpoint struct, but it probably makes sense :-)
 
Can you please also add a FFI API to set the timeout on the Endpoint?

ddcommon/src/lib.rs
```diff
@@ -32,11 +32,22 @@ pub type HttpClient = hyper::Client<connector::Connector, hyper::Body>;
 pub type HttpResponse = hyper::Response<hyper::Body>;
 pub type HttpRequestBuilder = hyper::http::request::Builder;
 
-#[derive(Default, Clone, PartialEq, Eq, Hash, Debug, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Hash, Debug, Serialize, Deserialize)]
 pub struct Endpoint {
     #[serde(serialize_with = "serialize_uri", deserialize_with = "deserialize_uri")]
     pub url: hyper::Uri,
     pub api_key: Option<Cow<'static, str>>,
+    pub timeout: u64,
```
bantonsson, line 40: Is there a particular reason to not have this as a `Duration`?

hoolioh: The aim was to reduce a little bit of memory. Since `Endpoint` is embedded in each `SendData` structurure and, therefore, the `TraceFlusher` holds a collection of `SendData` objects I thought it would be better just to hold 8 bytes rather than 16 per `SendData` instance. I even thought of just holding a u16 (probably a timeout of 65535 would be enough for our purposes) but since all milliseconds values inside std::time were expressed as a u64 I left it that way.

bin_tests/src/bin/crashtracker_bin_test.rs
```diff
@@ -46,6 +46,7 @@ mod unix {
             Some(ddcommon::Endpoint {
                 url: ddcommon::parse_uri(&output_url)?,
                 api_key: None,
```
bantonsson, line 48: I guess that all occurrences of these could be removed.

bantonsson (approved):

bin_tests/src/bin/crashtracker_bin_test.rs
hoolioh: you're right. I missed some of them.

pawelchcki (approved):
Generally looks good to me. I really liked the async mocked http tests !

Great addition thanks Julio! :bow: 

Added one optional suggestion taking a look at `thiserror` :) 

Though more importantly - when we plan to connect through world wide internet. 
We should consider giving our connector a bit more smartness with regards to timeout granularity.
e.g. copy impl from https://github.com/hjr3/hyper-timeout/blob/master/src/lib.rs

Basically in this PR we add more less one coarse timeout. 

And as I had to debug bugs in glibc sourced DNS resolution lock ups in ruby some time ago, I'm pretty paranoid. 

In ideal world I would love to see:
1. separate timeout and caching for dns resolution (esp if we're connecting to wide internet)
2. timeout on connect
3. separate timeouts on reads and writes.

This could give our users more visibility into what is happening in the background. 

^ just some thoughts for the future.

This PR -> LGTM!

trace-utils/src/send_data/mod.rs
```diff
@@ -853,17 +889,119 @@ mod tests {
...
+                timeout: MILLIS_PER_SECOND,
             },
         );
 
         let res = data.send().await;
 
         assert!(res.last_result.is_err());
-        assert_eq!(res.errors_timeout, 0);
-        assert_eq!(res.errors_network, 1);
+        match std::env::consts::OS {
```
VianneyRuhlmann, line 899: Why do we expect a timeout for windows instead of a network error ?

hoolioh: Oh, that drove me crazy. I'm not really familiar with Windows TCP/IP stack but it seems that, at hyper level, it returns a timeout rather than a connection refused error despite not having a listening socket on that port (4321).

danielsn:
There are several APIs that currently take timeouts and endpoints in their API.  e.g.
https://github.com/DataDog/libdatadog/blob/main/profiling/src/exporter/mod.rs#L173
https://github.com/DataDog/libdatadog/blob/main/crashtracker/src/configuration.rs#L29

These should be updated to only take the new endpoint

ddcommon-ffi/src/endpoint.rs
```diff
@@ -10,9 +10,12 @@ use std::str::FromStr;
...
 #[must_use]
 pub extern "C" fn ddog_endpoint_from_url(url: crate::CharSlice) -> Option<Box<Endpoint>> {
-    parse_uri(url.to_utf8_lossy().as_ref())
-        .ok()
-        .map(|url| Box::new(Endpoint { url, api_key: None }))
+    parse_uri(url.to_utf8_lossy().as_ref()).ok().map(|url| {
+        Box::new(Endpoint {
+            url,
+            ..Default::default()
+        })
```
danielsn, line 17: This pattern happens enough we should just have a constructor `Endpoint::from_url()`

ddcommon/src/lib.rs
danielsn: At the very least, this should document the unit, e.g. `timeout_ms`.  I'd argue for the `Duration` type as more self-documenting: do we have numbers for how many bytes this actually saves?

ddcommon/src/lib.rs
```diff
@@ -133,6 +144,9 @@ pub fn decode_uri_path_in_authority(uri: &hyper::Uri) -> anyhow::Result<PathBuf>
 }
 
 impl Endpoint {
+    /// Default value for the timeout field in milliseconds.
+    pub const DEFAULT_TIMEOUT: u64 = 10_000;
```
danielsn, line 148: how did we come up with this number?

ddtelemetry/examples/tm-send-sketch.rs
```diff
@@ -116,6 +116,7 @@ async fn async_main() {
             "https://instrumentation-telemetry-intake.datad0g.com/api/v2/apmtelemetry",
         ),
         api_key: Some(Cow::Owned(std::env::var("DD_API_KEY").unwrap())),
+        ..Default::default()
```
danielsn, line 119: why is this needed?

ddtelemetry/src/config.rs
```diff
@@ -228,7 +228,11 @@ impl Config {
             restartable: false,
         };
         if let Ok(url) = parse_uri(&trace_agent_url) {
-            let _res = this.set_endpoint(Endpoint { url, api_key });
+            let _res = this.set_endpoint(Endpoint {
+                url,
+                api_key,
+                ..Default::default()
```
danielsn, line 234: why is this needed?

ddtelemetry/src/worker/mod.rs
```diff
@@ -183,6 +183,9 @@ mod serialize {
 }
 
 impl TelemetryWorker {
+    // Default request timeout in milliseconds.
+    const DEFAULT_TIMEOUT: u64 = 10_000;
```
danielsn, line 187: This could be imported from the Endpoint timeout, avoid duplicating the same constant

profiling/src/exporter/config.rs
```diff
@@ -28,7 +28,10 @@ pub fn agent(base_url: Uri) -> anyhow::Result<Endpoint> {
     };
     parts.path_and_query = p_q;
     let url = Uri::from_parts(parts)?;
-    Ok(Endpoint { url, api_key: None })
+    Ok(Endpoint {
+        url,
+        ..Default::default()
+    })
```
danielsn, line 34: As before, would be clearer as `Endpoint::from_url(...)`

danielsn (dismissed):

trace-utils/src/send_data/mod.rs
```diff
@@ -458,6 +471,7 @@ mod tests {
     use httpmock::MockServer;
     use std::collections::HashMap;
 
+    const MILLIS_PER_SECOND: u64 = 1_000;
```
ekump, line 474: I'm not sure this is the best description of the value (even though it is technically true). Maybe `TIMEOUT_MS` is more correct?

sidecar/src/config.rs
```diff
@@ -182,6 +182,7 @@ pub fn get_product_endpoint(subdomain: &str, endpoint: &Endpoint) -> Endpoint {
         Endpoint {
             url: hyper::Uri::from_parts(parts).unwrap(),
             api_key: Some(api_key.clone()),
+            ..Default::default()
```
ekump, line 185: This also applies to the data-pipeline implementation...Is the current default of 10s appropriate? And do we want the same timeout value for both agent and agentless communication? I would think we would want a much lower value for talking to a trace agent? I'm not sure what appropriate values are, maybe we could check with trace agent folks?

ddcommon/src/lib.rs
hoolioh: @danielsn: Yeah, that'll make things clearer. Didn't have time to do benchnmark it. 
@ekump: `SendData` instances are eventually merged together (up to a maximum payload size) when the `TraceFlusher` tries flush them out so not every timeout is converted to a `Duration` that's why it seemed a waste of space.

trace-utils/src/send_data/mod.rs
VianneyRuhlmann: Why am I not surprised 😩. Maybe add a comment to avoid confusion

data-pipeline/src/trace_exporter.rs
```diff
@@ -310,21 +310,17 @@ impl TraceExporterBuilder {
...
             .enable_all()
             .build()?;
 
         Ok(TraceExporter {
-            endpoint,
+            endpoint: Endpoint::from_slice(
+                self.url
+                    .to_owned()
+                    .unwrap_or("http://127.0.0.1:8126/".to_string())
+                    .as_str(),
```
danielsn, line 322:
```suggestion
                self.url
                    .as_ref()
                    .map(|s| s.as_str())
                    .unwrap_or("http://127.0.0.1:8126/"),
```

sidecar/src/dogstatsd.rs
```diff
@@ -185,7 +185,7 @@ mod test {
                 .as_str()
                 .parse::<Uri>()
                 .unwrap(),
-            api_key: None,
+            ..Default::default()
```
danielsn, line 188:
```
Endpoint::from_slice(socket
                .local_addr()
                .unwrap()
                .to_string()
                .as_str())
```

sidecar/src/dogstatsd.rs
```diff
@@ -250,15 +241,14 @@ mod test {
     fn test_create_client_unix_domain_socket() {
         let res = create_client(Some(Endpoint {
             url: "unix://localhost:80".parse::<Uri>().unwrap(),
-            api_key: None,
+            ..Default::default()
```
danielsn, line 244: `Endpoint::from_slice`

sidecar/src/config.rs
bwoebi: This code should inherit the endpoint from input endpoint var.

bwoebi: I think it's the responsibility of the code setting the agentless mode to also decide on the timeout. But yes, I have no idea what proper timeouts are. Probably `Endpoint::from_url()` may default to a lower 3 sec timeout?

bwoebi (approved):
Looks good to me now, thanks!

bin_tests/src/bin/crashtracker_bin_test.rs
```diff
@@ -37,15 +37,15 @@ mod unix {
...
         let wait_for_receiver = true;
 
         let endpoint = if output_url.is_empty() {
             None
         } else {
-            Some(ddcommon::Endpoint {
-                url: ddcommon::parse_uri(&output_url)?,
-                api_key: None,
+            Some(Endpoint {
+                url: parse_uri(&output_url).unwrap(),
```
danielsn, line 46: `Endpoint::from_slice`

crashtracker/src/api.rs
```diff
@@ -142,8 +141,9 @@ fn test_crash() -> anyhow::Result<()> {
     let output_url = format!("file://{dir}{time}.txt");
 
     let endpoint = Some(Endpoint {
-        url: parse_uri(&output_url).unwrap(),
-        api_key: None,
+        url: ddcommon::parse_uri(&output_url).unwrap(),
```
danielsn, line 144: `Endpoint::from_slice`

data-pipeline/src/trace_exporter.rs
```diff
@@ -310,21 +310,12 @@ impl TraceExporterBuilder {
...
-            )?,
-            api_key: None,
-        };
         let runtime = tokio::runtime::Builder::new_current_thread()
             .enable_all()
             .build()?;
 
         Ok(TraceExporter {
-            endpoint,
+            endpoint: Endpoint::from_slice(self.url.as_deref().unwrap_or("http://127.0.0.1:8126")),
```
danielsn, line 318: 💯

examples/ffi/exporter.cpp
```diff
@@ -140,10 +140,11 @@ int main(int argc, char *argv[]) {
   ddog_CharSlice info_example = DDOG_CHARSLICE_C_BARE(
       "{\"application\": {\"start_time\": \"2024-01-24T11:17:22+0000\"}, \"platform\": {\"kernel\": \"Darwin Kernel 22.5.0\"}}");
 
+  ddog_prof_Exporter_set_timeout(exporter, 30000);
```
danielsn, line 143: I wonder if we even want this?  Tho as an example, maybe its good to show off the new API

profiling-ffi/src/exporter.rs
```diff
@@ -192,6 +192,20 @@ fn ddog_prof_exporter_new_impl(
...
+/// Sets the timeout for the exporte's timeout.
+/// # Arguments
+/// * `exporter` - ProfileExporter instance.
+/// * `timeout_ms` - timeout in milliseconds.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_prof_Exporter_set_timeout(
+    exporter: Option<&mut ProfileExporter>,
+    timeout_ms: u64,
+) {
+    if let Some(ptr) = exporter {
```
danielsn, line 204: Should we return an error in this case?

profiling/tests/form.rs
```diff
@@ -35,7 +35,8 @@ fn multipart(
     let start = now.sub(chrono::Duration::seconds(60));
     let end = now;
 
-    let timeout = std::time::Duration::from_secs(10);
+    let timeout: u64 = 10;
```
danielsn, line 38: This turns a 10 **second** timeout to 10 **ms**

trace-utils/src/send_data/mod.rs
```diff
@@ -977,7 +1120,7 @@ mod tests {
 
         let target_endpoint = Endpoint {
             url: server.url("").to_owned().parse().unwrap(),
-            api_key: None,
+            ..Default::default()
```
danielsn, line 1123: `Endpoint::from_slice("")`

danielsn (approved):
Good to go, modulo some small comments.

PR #521, opened 2024-07-08
Fix sidecar logging

cataphract:
Broken by ce2afd981e994d9f3aa5b4c12e35cab9facee4e3, since when parse_uri starts returning a bogus uri, with the path put in the authority in hexadecimal form and the actual path empty. Sidestep that mess by not using parse_uri.

### DIFF
--- a/sidecar/src/config.rs
+++ b/sidecar/src/config.rs
@@ -5,7 +5,7 @@ use http::uri::{PathAndQuery, Scheme};
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, path::PathBuf, time::Duration};
 
-use ddcommon::{parse_uri, Endpoint};
+use ddcommon::Endpoint;
 use spawn_worker::LibDependency;
 
 const ENV_SIDECAR_IPC_MODE: &str = "_DD_DEBUG_SIDECAR_IPC_MODE";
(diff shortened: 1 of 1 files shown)


sidecar/src/config.rs
```diff
@@ -126,16 +129,13 @@ impl FromEnv {
...
-                        None
-                    }
-                })
-                .unwrap_or_default(),
+            method if method.starts_with("file:///") => {
+                let fixed_uri = format!("file://./{}", &method[7..]);
+                Uri::from_str(&fixed_uri)
+                    .map(|u| LogMethod::File(PathBuf::from(&u.path())))
+                    .unwrap_or_default()
+            }
```
paullegranddc, line 137: Why not just this instead of the roundtrip to url?
```suggestion
            method if method.starts_with("file://") => {
                let path = method.strip_prefix("file://").unwrap();
                LogMethod::File(PathBuf::from(path))
            }
```

cataphract: It's because the specification for the path is in the form `file:///my/path`. However, even though this is a valid URI (with an empty host), see RFC 3986:

```
  URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]

      hier-part   = "//" authority path-abempty
                  / path-absolute
                  / path-rootless
                  / path-empty

      authority   = [ userinfo "@" ] host [ ":" port ]

      host        = IP-literal / IPv4address / reg-name
      reg-name    = *( unreserved / pct-encoded / sub-delims )

   If the URI scheme defines a default for host, then that default
   applies when the host subcomponent is undefined or when the
   registered name is empty (zero length).  For example, the "file" URI
   scheme is defined so that no authority, an empty host, and
   "localhost" all mean the end-user's machine, whereas the "http"
   scheme considers a missing authority or empty host invalid.
```

the http package rejects the uri for not having a host. So add a bogus host (.).

paullegranddc: Going through the `Uri` struct is not a good idea. If there are non-alphanumeric characters, they would need to be URL-encoded for this to work. This value comes from an env variable that can be manually supplied so encoding is painful to do.

We already do not follow the URI standard for unix sockets URL (since the logic is just `unix:// + path on disk without any special encoding` ) so I don't see the point of arbitrarily following it for this.

cataphract: ah sorry I didn't read that right. I can't just strip the prefix because it may have urlencoded characters

paullegranddc (approved):

PR #527, opened 2024-07-11
Make flush interval in telemetry configurable via sidecar

cataphract:
Environment variables read via `environ` when sidecar is spawned are not necessarily available (see https://gist.github.com/cataphract/78ca7cabdded79360d790a8bc249e93d ). Add telemetry flush interval as a session config entry instead.


bwoebi (approved):
Looks good :-)

PR #531, opened 2024-07-15
ekump/APMSP-1279 benchmark trace exporting

ekump:
# What does this PR do?

1)  Move some logic that resided in the sidecar for building a `TracerPayloadCollection` from msgpack bytes and `TracerHeaderTags` into the `trace_utils` crate as this logic can / should be used anywhere `TracerPayloadCollection` is used. No changes to the logic were made. This logic includes the deserialization of traces in msgpack format into an internal representation of traces (currently pb). Previously, this functionality wasn't directly covered under unit tests. Now it is.
2) Add a micro-benchmark test that covers the creation of `TracerPayloadCollections` for v04 traces. 

# Motivation

There is a known problem with excessive allocations when processing traces through libdatadog that is currently blocking the sidecar from being turned on by default for PHP customers. This problem also blocks adoption of data-pipeline by .NET and Node libraries. These benchmarks should help establish a baseline for forthcoming work to improve performance. By moving the deserialization of msgpack closer to the TracerPayloadCollection implementation it ensures common encapsulated logic for both data-pipeline and SIdecar.   

# Additional Notes

This implementation currently only works for v04. Work to support v05 and v07 should follow.

# How to test the change?

Unit tests added.
benchmarks can be run via `cargo bench --package datadog-trace-utils`


trace-utils/src/tracer_payload.rs
```diff
@@ -106,14 +105,140 @@ impl TracerPayloadCollection {
     }
 }
 
+pub type TracerPayloadChunkProcessor<'a> = Box<dyn Fn(&mut pb::TraceChunk, usize) + 'a>;
```
bwoebi, line 108: The proper way to implement this is having a `trait TracerPayloadChunkProcessor { fn process(&mut pb::TraceChunk, usize); }` and adding a `PhantomData<T>` to TracerPayloadParams with `T: TracerPayloadChunkProcessor`.

bwoebi: I meant _allows_, not avoids.

Saving an extra memory allocation and call mostly. Nothing major, just idiomatic and a tiny bit better.

hoolioh (approved):

trace-utils/src/tracer_payload.rs
```diff
@@ -106,14 +105,138 @@ impl TracerPayloadCollection {
...
+/// * `encoding_type`: The encoding type of the trace data, determining how the data should be
+///   deserialized and processed.
+pub struct TracerPayloadParams<'a> {
+    /// A byte slice containing the serialized msgpack data.
+    data: &'a [u8],
+    /// Reference to `TracerHeaderTags` containing metadata for the trace.
+    tracer_header_tags: &'a TracerHeaderTags<'a>,
+    /// A boxed closure that processes each `TraceChunk`, allowing for custom logic to be applied
+    /// during the conversion process.
+    process_chunk: TracerPayloadChunkProcessor<'a>,
```
VianneyRuhlmann, line 129: Do we often use a ChunkProcessor ? I not it feels a bit heavy to have to create an empty closure every time we call it maybe it can be an Option or provide an alternative constructor which creates the empty closure.

VianneyRuhlmann (approved):

trace-utils/src/tracer_payload.rs
pawelchcki: Probably better to prefer static dispatch. Using boxed fn is a bit like Virtual (dynamic) method dispatch in C++ - useful when polymorphism is required but adding overhead. 

Since this is part of trace chunk processing that can happen in a tight loop. 

In addition to extra instructions it will have higher chance to incur a page fault whenever processor is called. 
Ideally we should keep all code needed to do chunk processing within the CPU cache when possible.

trace-utils/src/tracer_payload.rs
```diff
@@ -181,4 +304,95 @@ mod tests {
...
+            service: "test-service".to_string(),
+            name: "test-service-name".to_string(),
+            resource: "test-service-resource".to_string(),
+            trace_id: 111,
+            span_id: 222,
+            parent_id: 100,
+            start: 1,
+            duration: 5,
+            error: 0,
+            meta: HashMap::new(),
```
pawelchcki, line 334: Not related to this PR. But we whould generally avoid using Hashmaps for data which doesn't require it.

e.g. in many tracers we replaced HashMap for meta and metrics with a list of key,values - for internal representation - because it avoid a lot of unnecessary memory operations. And is generally alot faster since the map is very rarely looked up.

bwoebi: I don't know about that, but at least PHP does quite a bunch of meta/metrics lookups at serialization time.
Also, e.g. sampling can be tag based.
And there are also quite a bunch of tags which are tied to state, like sampling, decision makers etc. which end up requiring lookups.
I doubt that a key-value list would bring tangible benefits over a solid HashTable impl.

trace-utils/src/tracer_payload.rs
ekump: @bwoebi Ah ok, that makes more sense. 

I agree, using dynamic dispatch is not optimal. I was just trying to limit the blast radius of the refactor and maintain the same function signature for `trace_utils::collect_trace_chunks`.

I didn't want to commit the changes to this branch yet as I still need to do a bunch of cleanup as well as fix tests. [But is this what you had in mind](https://github.com/DataDog/libdatadog/pull/534/files)? It uses a trait as you suggested and avoids the need for any dynamic dispatch. There is no need for `PhantomData` though. If this makes sense, I'll incorporate it into this PR.

bwoebi: Yes, that works too :-)

tools/docker/Dockerfile.build
```diff
@@ -157,4 +159,4 @@ RUN ./build-profiling-ffi.sh /build/output
 
 FROM scratch as ffi_build_output
 
-COPY --from=ffi_build /build/output/ ./
+COPY --from=ffi_build /build/output/ ./
```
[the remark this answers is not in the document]
ekump: I'll get my IDE configured correctly one of these days

trace-utils/src/tracer_payload.rs
```diff
@@ -106,14 +105,138 @@ impl TracerPayloadCollection {
...
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::test_utils::create_test_span;
-    use datadog_trace_protobuf::pb::TraceChunk;
+    use datadog_trace_protobuf::pb::{Span, TraceChunk};
```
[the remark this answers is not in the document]
ekump: That was not intentional, will fix.

trace-utils/src/tracer_payload.rs
ekump: That's a fair point. It currently only works for `v07` and is only used by the mini-agent. For now, I created a `DefaultTraceChunkProcessor` that can be passed so we don't have to implement the trait just to do a no-op. The compiler _should_ optimize away the call to the processor in `collect_trace_chunks` in this situation. 

I'm inclined to leave it as is for now. I think we may want to do some further refactoring to TracerPayloads in the not to distant future and we may want to think of ways to streamline how some of this works.

trace-utils/Cargo.toml
```diff
@@ -1,14 +1,18 @@
...
 license.workspace = true
+autobenches = false
 
 [lib]
 bench = false
 
+[[bench]]
+name = "main"
+harness = false
+
```
bwoebi, line 15:
```suggestion
path = "benches/main.rs"

```

That allows not shipping the benchmarks in php and have cargo not complain.

bwoebi (approved):

PR #535, opened 2024-07-17
Init CrashTracker support in the sidecar

iamluc:
# What does this PR do?

This PR introduces a CrashTracker receiver (via a Unix socket) in the sidecar.

As the sidecar uses the Tokio async framework, I had to convert a few functions from the `crashtracker` crate into async functions.

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/crashtracker/src/lib.rs
+++ b/crashtracker/src/lib.rs
@@ -65,8 +65,8 @@ pub use crash_info::*;
 
 #[cfg(all(unix, feature = "receiver"))]
 pub use receiver::{
-    receiver_entry_point_stdin, receiver_entry_point_unix_socket,
-    async_receiver_entry_point_unix_socket,
+    async_receiver_entry_point_unix_socket, receiver_entry_point_stdin,
+    receiver_entry_point_unix_socket,
 };
 
 #[cfg(all(unix, any(feature = "collector", feature = "receiver")))]
--- a/crashtracker/src/lib.rs
+++ b/crashtracker/src/lib.rs
@@ -64,7 +64,10 @@ pub use collector::{
 pub use crash_info::*;
 
 #[cfg(all(unix, feature = "receiver"))]
-pub use receiver::{receiver_entry_point_stdin, receiver_entry_point_unix_socket};
+pub use receiver::{
+    receiver_entry_point_stdin, receiver_entry_point_unix_socket,
+    async_receiver_entry_point_unix_socket,
+};
 
 #[cfg(all(unix, any(feature = "collector", feature = "receiver")))]
 pub use shared::configuration::{
--- a/crashtracker/src/lib.rs
+++ b/crashtracker/src/lib.rs
@@ -64,7 +64,7 @@ pub use collector::{
 pub use crash_info::*;
 
 #[cfg(all(unix, feature = "receiver"))]
-pub use receiver::{receiver_entry_point_stdin, reciever_entry_point_unix_socket};
+pub use receiver::{receiver_entry_point_stdin, receiver_entry_point_unix_socket};
 
 #[cfg(all(unix, any(feature = "collector", feature = "receiver")))]
 pub use shared::configuration::{
--- a/crashtracker/src/receiver.rs
+++ b/crashtracker/src/receiver.rs
@@ -5,7 +5,9 @@
 use super::*;
 use crate::shared::constants::*;
 use anyhow::Context;
-use std::{io::BufReader, os::unix::net::UnixListener};
+use tokio::io::AsyncBufReadExt;
+use tokio::io::BufReader;
+use tokio::net::UnixListener;
 
 pub fn resolve_frames(
     config: &CrashtrackerConfiguration,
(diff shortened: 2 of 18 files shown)


crashtracker/src/receiver.rs
```diff
@@ -32,17 +34,38 @@ pub fn get_unix_socket(socket_path: impl AsRef<str>) -> anyhow::Result<UnixListe
...
-    receiver_entry_point(stream)
+    let stream = BufReader::new(tokio::io::stdin());
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()?;
+    rt.block_on(receiver_entry_point(stream))?;
+    Ok(())
+}
+
+pub async fn async_receiver_entry_point_unix_socket(
```
danielsn, line 55: do we have a test for this

sidecar-ffi/src/lib.rs
```diff
@@ -709,3 +710,18 @@ pub extern "C" fn ddog_sidecar_reconnect(
 ) {
     transport.reconnect(|| unsafe { factory() });
 }
+
+/// Return the path of the crashtracker unix domain socket.
+#[no_mangle]
+#[allow(clippy::missing_safety_doc)]
+pub unsafe extern "C" fn ddog_sidecar_get_crashtracker_unix_socket_path() -> ffi::CharSlice<'static>
```
danielsn, line 717: Who is the owner of this string, and is expected to free it?

sidecar-ffi/src/lib.rs
```diff
@@ -709,3 +710,18 @@ pub extern "C" fn ddog_sidecar_reconnect(
...
+{
+    let socket_path = crashtracker_unix_socket_path();
+    let str = socket_path.to_str().unwrap_or_default();
+
+    let size = str.len();
```
danielsn, line 722: Would it be cleaner to take a trip through https://doc.rust-lang.org/std/primitive.str.html#method.into_boxed_bytes

crashtracker/src/crash_info.rs
```diff
@@ -243,10 +243,12 @@ impl CrashInfo {
     /// SIGNAL SAFETY:
     ///     I believe but have not verified this is signal safe.
     pub fn to_file(&self, path: &Path) -> anyhow::Result<()> {
-        let file =
-            File::create(path).with_context(|| format!("Failed to create {}", path.display()))?;
+        let binding = path.as_os_str().to_string_lossy();
+        let path = binding.strip_prefix("file://").unwrap_or(&binding);
```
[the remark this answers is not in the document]
iamluc: Removed, it was not needed.
I think I did that because I copied `ddog_Endpoint_file(to_slice_c_char("file://tmp/test"))` from `examples/ffi/crashinfo.cpp`. 
But I think the correct code is `ddog_Endpoint_file(to_slice_c_char("/tmp/test"))`.

I updated the example.

sidecar-ffi/src/lib.rs
bwoebi: The caller. I mean, it returns libc::malloc, so the caller can just free() it.

sidecar-ffi/src/lib.rs
bwoebi: At the cost of not being able to free() it in C code. This impl allows avoiding to need extra drop routines / relying on the specific allocator rust currently uses.

crashtracker/src/crash_info.rs
```diff
@@ -258,14 +260,39 @@ impl CrashInfo {
                     .context("crash output file was not correctly formatted")?;
                 self.to_file(&path)?;
             }
+            // FIXME return?
+            return Ok(());
         }
-        self.upload_to_telemetry(config)
+
+        let rt = tokio::runtime::Builder::new_current_thread()
```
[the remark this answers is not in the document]
iamluc: I removed the comment because I didn't know what to say...
If you have a suggestion, I'm all ears.

crashtracker/src/receiver.rs
iamluc: It's tested by the "bin_tests" as both sync and async "entry_point_unix_socket" use this code path

crashtracker/src/lib.rs
```diff
@@ -68,5 +68,8 @@ pub use counters::{begin_profiling_op, end_profiling_op, reset_counters, Profili
 pub use crash_handler::{update_config, update_metadata};
 pub use crash_info::*;
 #[cfg(unix)]
-pub use receiver::{receiver_entry_point_stdin, reciever_entry_point_unix_socket};
+pub use receiver::{
+    async_receiver_entry_point_unix_socket, receiver_entry_point_stdin,
+    receiver_entry_point_unix_socket,
```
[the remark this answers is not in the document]
bwoebi: If we use them only for tests I'd update them to use the async version too.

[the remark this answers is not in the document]
iamluc: After checking again, it's also used by the `ddog_crasht_receiver_entry_point_unix_socket` FFI function.

bwoebi (approved):

PR #536, opened 2024-07-18
[Crashtracker] Track active spans and traces

danielsn:
# What does this PR do?

Provides a mechanism for clients to record active spans/traces, and reports them in the crashreport.

# Motivation

Error-tracking said this would be useful to help corrlate crashes with the traces they affected.

# Additional Notes

# How to test the change?


crashtracker/src/spans.rs
```diff
@@ -0,0 +1,265 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+use portable_atomic::{AtomicU128, AtomicUsize};
+use rand::Rng;
+use std::io::Write;
+use std::sync::atomic::Ordering::SeqCst;
+
+static ACTIVE_SPANS: AtomicU128Set<2048> = AtomicU128Set::new();
+static ACTIVE_TRACES: AtomicU128Set<2048> = AtomicU128Set::new();
```
taegyunkim, line 10: Where did 2048 come from?

taegyunkim (approved):

PR #538, opened 2024-07-18
[profiling] add ffi for crossbeam::ArrayQueue

taegyunkim:
# What does this PR do?

Adds ffi for `crossbeam::ArrayQueue<*mut c_void>`

# Motivation

Python profiler allocates a new `Sample` whenever a it collects an event (stack/memory/lock).   Then, the Sample is freed right after. Code for `new/delete Sample` is [here](https://github.com/DataDog/dd-trace-py/blob/d1db2005a8e301816f7618ee2b14ab9acfec3185/ddtrace/internal/datadog/profiling/dd_wrapper/src/sample_manager.cpp#L30-L40). This results in calling `new/delete` a lot and creates some memory pressure to the system and shows up on native profiler. Frequent calls to `new/delete` can be avoided if we use a pool of `Sample`s. And this `ArrayQueue` will be used to implement the pool. 

Another option we considered is using mutex guarded vector in C++ in dd-trace-py. However, it would be little tricky to handle `fork()` if we use a lock. So we decided to use lock-free list/queue/stack.

# Additional Notes

# How to test the change?


profiling-ffi/src/lib.rs
```diff
@@ -9,6 +9,7 @@ use std::time::SystemTime;
 
 use chrono::{DateTime, TimeZone, Utc};
 
+mod array_queue;
```
danielsn, line 12: Would this make sense in `ddcommon-ffi` since there is nothing particularly profiling about it?

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    pub fn take(&mut self) -> Option<Box<crossbeam_queue::ArrayQueue<*mut c_void>>> {
+        let raw = std::mem::replace(&mut self.inner, std::ptr::null_mut());
+
+        if raw.is_null() {
+            None
+        } else {
+            let queue =
+                unsafe { Box::from_raw(raw as *mut crossbeam_queue::ArrayQueue<*mut c_void>) };
```
danielsn, line 40: safety comment

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,285 @@
...
+/// The pointer is null or points to a valid memory location allocated by array_queue_new.
+unsafe fn ddog_array_queue_ptr_to_inner<'a>(
+    queue_ptr: *mut ArrayQueue,
+) -> anyhow::Result<&'a mut crossbeam_queue::ArrayQueue<*mut c_void>> {
+    match queue_ptr.as_mut() {
+        None => anyhow::bail!("queue_ptr is null"),
+        Some(queue) => match queue.inner.as_mut() {
+            None => anyhow::bail!("queue.inner is null"),
+            Some(inner) => {
+                Ok(&mut *(inner as *mut c_void as *mut crossbeam_queue::ArrayQueue<*mut c_void>))
```
danielsn, line 81: safety comment

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,296 @@
...
+pub enum ArrayQueueNewResult {
+    Ok(ArrayQueue),
+    Err(Error),
+}
+
+/// Creates a new ArrayQueue with the given capacity and item_delete_fn.
+/// The item_delete_fn is called when an item is dropped from the queue.
+#[no_mangle]
+#[must_use]
+pub extern "C" fn ddog_array_queue_new(
```
ivoanjo, line 71: Hope you don't mind the tiny nitpick, but for consistency the naming we've adopted calls tries to match the rust struct name on the C side, e.g. 
`ArrayQueue.new` => `ddog_ArrayQueue_new` :)

(This applies to the other public functions as well)

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,296 @@
...
+            },
+            Err(err) => ArrayQueuePushResult::Err(err.into()),
+        }
+    }
+}
+
+/// Pushes an element into the queue, replacing the oldest element if necessary.
+/// # Safety
+/// The pointer is null or points to a valid memory location allocated by array_queue_new. The value
+#[no_mangle]
```
ivoanjo, line 160: This comment seems to have been cut-off?

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,296 @@
...
+    if !queue_ptr.is_null() {
+        drop((*queue_ptr).take());
+    }
+}
+
+#[allow(unused)]
+#[repr(C)]
+pub enum ArrayQueuePushResult {
+    Ok,
+    Full(*mut c_void),
```
ivoanjo, line 112: Minor: It took me a while to understand why it was useful to return the value back, since for `push` it's somewhat redundant. Maybe include here a comment saying returning a value is mostly for `force_push`?

ivoanjo (approved):
This looks quite reasonable! I've been holding off on buffering things on the Ruby library (instead try to make everything immediate) in part because of the complexity of getting this kind of queue going in a C codebase, but I guess after this is in I should probably consider using it in a few places :)

My rust-foo isn't great so I recommend waiting for @danielsn's final pass/approval as well, but here sir, take my :+1:

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,296 @@
...
+    let ffi_queue = ArrayQueue::new(capacity, item_delete_fn);
+    ArrayQueueNewResult::Ok(ffi_queue)
+}
+
+/// Converts a *mut ArrayQueue to a &mut crossbeam_queue::ArrayQueue<*mut c_void>.
+/// # Safety
+/// The pointer is null or points to a valid memory location allocated by array_queue_new.
+unsafe fn ddog_array_queue_ptr_to_inner<'a>(
+    queue_ptr: *mut ArrayQueue,
+) -> anyhow::Result<&'a mut crossbeam_queue::ArrayQueue<*mut c_void>> {
```
paullegranddc, line 84: If this function is not exported (it cannot really since it returns non C types) it probably shouldn't follow the `ddog_array_queue` prefix, because it's not very idiomatic from the rust side.

Maybe make it an associated function on the type?
```rust
impl ArrayQueue {
  fn into_inner(queue_ptr: *mut ArrayQueue) -> anyhow::Result<&'a mut crossbeam_queue::ArrayQueue<*mut c_void>> {
}

// then call like
let queue = ArrayQueue::into_inner(queue_ptr);
```
}

paullegranddc (commented):
As a general comment, all of the `extern "C"` functions in this PR take an `*ArrayQueue`  and cast it to an `&'a crossbeam::ArrayQueue`,  but the thing is, the lifetime is never bound, so you could very well get an `&'static crossbeam::ArrayQueue` and have it escape the fucntion.


I think it'd be better to take an `&'a ArrayQueue` in the ffi fucntions, which is translated to the same `ArrayQueue*` in the C header, which would allow to derive the `&'a crossbeam::ArrayQueue` lifetime safely and ensure it doesn't escape.

ddcommon-ffi/src/array_queue.rs
paullegranddc: Also, this function casts the ArrayQueue to an `&mut`. So, it's undefined behavior to call it concurrently. Most functions on the crossbeam::ArrayQueue only require &self anyway, so IMO you should probably have to cast functions, one creating a ref for push/pop and another one getting a ref_mut for the drop.

taegyunkim: Done. see `as_inner_ref()` and `take()`

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,294 @@
...
+        let internal_queue: crossbeam_queue::ArrayQueue<*mut c_void> =
+            crossbeam_queue::ArrayQueue::new(capacity);
+        let inner = Box::into_raw(Box::new(internal_queue)) as *mut c_void;
+        Self {
+            inner,
+            item_delete_fn,
+        }
+    }
+
+    pub fn take(&mut self) -> Option<Box<crossbeam_queue::ArrayQueue<*mut c_void>>> {
```
taegyunkim, line 33: To reviewers, @ivoanjo @danielsn @paullegranddc 

Not sure whether it would be clearer to have `take()` or put this in `drop()`. Any suggestions?

examples/ffi/array_queue.cpp
```diff
@@ -0,0 +1,100 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
```
taegyunkim, line 1: I've only tested running this with `valgrind`(`leaks` on Mac). Any other suggestions for testing?

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,294 @@
...
+
+#[repr(C)]
+// A simple wrapper around crossbeam_queue::ArrayQueue<*mut c_void>, which is a lock free
+// bounded multi-producer and multi-consumer (MPMC) queue.
+pub struct ArrayQueue {
+    // The actual type here should be *mut crossbeam_queue::ArrayQueue<*mut c_void>.
+    // However, cbindgen does not use the module name crossbeam_queue to generate the C header.
+    // So we use *mut c_void here and cast it to the correct type in the FFI functions.
+    inner: *mut c_void,
+    item_delete_fn: unsafe extern "C" fn(*mut c_void) -> c_void,
```
danielsn, line 16: Can this be null?

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,294 @@
...
+use std::ffi::c_void;
+
+#[repr(C)]
+// A simple wrapper around crossbeam_queue::ArrayQueue<*mut c_void>, which is a lock free
+// bounded multi-producer and multi-consumer (MPMC) queue.
+pub struct ArrayQueue {
+    // The actual type here should be *mut crossbeam_queue::ArrayQueue<*mut c_void>.
+    // However, cbindgen does not use the module name crossbeam_queue to generate the C header.
+    // So we use *mut c_void here and cast it to the correct type in the FFI functions.
+    inner: *mut c_void,
```
danielsn, line 15: Can this be null?

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,294 @@
...
+    item_delete_fn: unsafe extern "C" fn(*mut c_void) -> c_void,
+}
+
+impl ArrayQueue {
+    pub fn new(
+        capacity: usize,
+        item_delete_fn: unsafe extern "C" fn(*mut c_void) -> c_void,
+    ) -> Self {
```
danielsn, line 23: This should return `Result` and check if item_delete_fn is null

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,294 @@
...
+    // The actual type here should be *mut crossbeam_queue::ArrayQueue<*mut c_void>.
+    // However, cbindgen does not use the module name crossbeam_queue to generate the C header.
+    // So we use *mut c_void here and cast it to the correct type in the FFI functions.
+    inner: *mut c_void,
+    item_delete_fn: unsafe extern "C" fn(*mut c_void) -> c_void,
+}
+
+impl ArrayQueue {
+    pub fn new(
+        capacity: usize,
```
danielsn, line 21: 0 capacity is illegal

ddcommon-ffi/src/array_queue.rs
danielsn: I'd put it in `drop`.  Having `take` do anything other than just replacing the value with default violates principal of least suprise.

ddcommon-ffi/src/array_queue.rs
danielsn, line 15: I wonder if we can just make this a `Box<ArrayQueue>` https://github.com/rust-lang/unsafe-code-guidelines/issues/157

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,294 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::Error;
+use anyhow::Context;
+use std::ffi::c_void;
+
+#[repr(C)]
+// A simple wrapper around crossbeam_queue::ArrayQueue<*mut c_void>, which is a lock free
+// bounded multi-producer and multi-consumer (MPMC) queue.
+pub struct ArrayQueue {
```
danielsn, line 11: Since you implement drop, you might need a phantom data

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,294 @@
...
+) -> ArrayQueueNewResult {
+    let ffi_queue = ArrayQueue::new(capacity, item_delete_fn);
+    ArrayQueueNewResult::Ok(ffi_queue)
+}
+
+/// Drops the ArrayQueue.
+/// # Safety
+/// The pointer is null or points to a valid memory location allocated by array_queue_new.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_ArrayQueue_drop(queue: &mut ArrayQueue) {
```
danielsn, line 97: do we want to alert on the case of double free?

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,294 @@
...
+            Ok(value) => match value {
+                Ok(()) => ArrayQueuePushResult::Ok,
+                Err(value) => ArrayQueuePushResult::Full(value),
+            },
+            Err(err) => ArrayQueuePushResult::Err(err.into()),
+        }
+    }
+}
+
+/// Pushes an item into the ArrayQueue.
```
danielsn, line 125: describe the return value

examples/ffi/array_queue.cpp
danielsn: fuzzing?

gleocadie:
Can you add description and what motivates this PR please ? (as describe in the template)
It's difficult to understand why we are adding this and if it would benefit us (other libraries)
Thx

ddcommon-ffi/src/array_queue.rs
taegyunkim: Changed to use `NonNull`

ddcommon-ffi/src/array_queue.rs
taegyunkim: Changed to use `NonNull` with comments

taegyunkim: Changed to use `NonNull<ArrayQueue>`, thanks for pointing this out. And added a comment why `NonNull` instead of `Box`

ddcommon-ffi/src/array_queue.rs
taegyunkim: Would miri complain about this?

ddcommon-ffi/src/array_queue.rs
taegyunkim: Added comment

ddcommon-ffi/src/array_queue.rs
taegyunkim: Here the ArrayQueue only performs push and pop, don't have peek() which could have required me to have a phantom data.

ddcommon-ffi/src/array_queue.rs
taegyunkim: Also, changed the functions to take `&ArrayQueue` instead of `&mut`

examples/ffi/array_queue.cpp
taegyunkim: Added fuzz testing on rust side.

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,474 @@
...
+        }
+        let item_delete_fn = item_delete_fn.context("item_delete_fn must be non-null")?;
+
+        let internal_queue: crossbeam_queue::ArrayQueue<*mut c_void> =
+            crossbeam_queue::ArrayQueue::new(capacity);
+        // # Safety: internal_queue must be non-null If the memory allocation had failed, the
+        // program would panic.
+        let inner = unsafe {
+            NonNull::new_unchecked(Box::into_raw(Box::new(internal_queue)) as *mut ArrayQueue)
+        };
```
danielsn, line 41: This happens so rarely, it would be cleaner to just 
```suggestion
        let inner = NonNull::new(Box::into_raw(Box::new(internal_queue)) as *mut ArrayQueue)?;
```

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,474 @@
...
+/// Returns the length of the ArrayQueue.
+/// # Safety
+/// The pointer is null or points to a valid memory location allocated by array_queue_new.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_ArrayQueue_len(queue_ptr: &ArrayQueue) -> ArrayQueueUsizeResult {
+    (|| {
+        let queue = ArrayQueue::as_inner_ref(queue_ptr)?;
+        anyhow::Ok(queue.len())
+    })()
+    .context("array_queue_len failed")
```
danielsn, line 272: Use the same case as the function

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,474 @@
...
+                            panic!("pop failed");
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    #[test]
+    fn fuzz_with_threads() {
```
danielsn, line 434: 💯

danielsn (approved):
Modulo a few  ⛏️  LGTM

ddcommon-ffi/src/array_queue.rs
```diff
@@ -0,0 +1,470 @@
...
+        capacity: usize,
+        item_delete_fn: Option<unsafe extern "C" fn(*mut c_void) -> c_void>,
+    ) -> anyhow::Result<Self, anyhow::Error> {
+        anyhow::ensure!(capacity > 0, "capacity must be greater than 0");
+        let item_delete_fn = item_delete_fn.context("item_delete_fn must be non-null")?;
+
+        let internal_queue: crossbeam_queue::ArrayQueue<*mut c_void> =
+            crossbeam_queue::ArrayQueue::new(capacity);
+        // # Safety: internal_queue must be non-null.
+        // If the memory allocation had failed, the program would panic.
```
danielsn, line 36:
```suggestion
```

There is no unsafe block anymore

PR #539, opened 2024-07-18
[crashtracker] Make C example emit a file

danielsn:
# What does this PR do?

What it says on the tin

# Motivation

The current C example was hard to follow, since you needed an instrumented agent to see where the output went.  This is cleaner. 

# Additional Notes

# How to test the change?


morrisonlevi (approved):
Looks good on my side.

pawelchcki (approved):

PR #540, opened 2024-07-19
Use serde_bytes for meta_struct

bwoebi:
Otherwise meta_struct will be serialized as an array of integers instead of a byte string, breaking the expected representation.

Appsec is trying to use it, but breaking when using the sidecar trace sender.


bantonsson (commented):
Well, since we're not going to send v0.7 (or at least not optimize or test that) from the sidecar, I don't know how ASM will be able to use `meta_struct`.

I'm not saying that the fix is wrong, but I can only see `meta_struct` being used in an agentless setting, when sending directly to intake.

PR #541, opened 2024-07-19
[crashtracker] Gate compilation using cargo features

danielsn:
# What does this PR do?

Adds two new cargo features: `collector` and `receiver`. Refactors the code to make it easy to use those `cfg` to define what gets built at compile time.

# Motivation

The current implementation of the library always builds in all code.  This is not great, because it bloats binary size etc.  We have users who

1. (.Net) independently collect core-dumps, and then format them into `CrashInfo` objects, which they upload. They have no need or use for the `receiver` or `collector` portions of this code
2. (Ruby, Python) Use the collector to gather info on crashes, and the receiver to format a `CrashInfo` object and send it on. They need all parts of the code.
3. (PHP). Propose to collect the info on crashes in the main binary, then use a sidecar to format a `CrashInfo` object and send it on.  The first binary needs the `collector` but not the `receiver`; the sidecar needs the opposite.  

# Additional Notes

* Currently, we don't have support when building the ffi library for feature selection.  This is a first step towards making that worthwhile.

* I did some refactoring as I moved code around, but there is still more refactoring that could be done.
 
# How to test the change?

This is a refactor, existing tests cover the functionality


taegyunkim (approved):

taegyunkim:
Looks like this also needs changes to ffi

crashtracker/src/collector/emitters.rs
```diff
@@ -1,12 +1,13 @@
-// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
```
sanchda, line 1: Silly question:  do these copyright notices indicate the year of first existence for the noted file, or the last update?

crashtracker/src/collector/emitters.rs
```diff
@@ -45,7 +46,7 @@ pub unsafe fn emit_backtrace_by_frames(
 
             let mut first = true;
             // This can give multiple answers in the case of inlined functions
-            // https://docs.rs/backtrace/latest/backtrace/fn.resolve.html
+            // https://docs.rs/backtrace/latest/backtrace/pub(crate) fn.resolve.html
```
sanchda, line 49: Broken link?

crashtracker/src/collector/emitters.rs
```diff
@@ -94,6 +95,89 @@ pub unsafe fn emit_backtrace_by_frames(
...
+    emit_config(pipe, config_str)?;
+    emit_siginfo(pipe, signum)?;
+    emit_procinfo(pipe)?;
+    pipe.flush()?;
+    emit_counters(pipe)?;
+    pipe.flush()?;
+    emit_spans(pipe)?;
+    pipe.flush()?;
+    emit_traces(pipe)?;
+    pipe.flush()?;
```
sanchda, line 115: Why is it that the emission of some data is immediately followed by a flush operation sometimes, but not always?

sanchda (approved):
LGTM--pretty reasonable-looking refactor with minimal changes to how the code operates.

crashtracker/src/collector/emitters.rs
danielsn: Good question, it probably should always.  For a future PR.

danielsn: https://github.com/DataDog/libdatadog/issues/546

crashtracker/src/collector/emitters.rs
danielsn: fixed

crashtracker/src/collector/emitters.rs
danielsn: I assume first year of existence.  At Amazon we stopped using years, and went to a SPDX header which was much cleaner

PR #545, opened 2024-07-22
Add basic decoding for v0.4 msgpack traces

ekump:
# What does this PR do?
Deserialize msgpack payloads to PB without using serde.

Originally authored by @hoolioh this PR replaces the use of serde to decode msgpack v04 trace payloads
- Adds methods to decode string and &strs
- Adds number abstraction layer in order to decode integer and floats.
- Implements a decoder for Spans
- Implements a decoder for meta attributes
- Implements a decoder for metrics attributes
- Implements a decoder for SpanLinks

# Motivation

This is part of a larger effort to reduce memory allocations when sending traces through the data-pipeline. After we stop using serde we will move on to improving the internal representation of traces to use references rather than allocating new strings. 

# Additional Notes

The introduction of the deserialization process in trace-utils integrations tests will be introduced in a follow-up PR.

# How to test the change?

unit and fuzz tests included.


trace-utils/src/msgpack_decoder/v04/number.rs
```diff
@@ -0,0 +1,236 @@
...
+
+pub enum Number {
+    U8(u8),
+    U32(u32),
+    U64(u64),
+    I8(i8),
+    I32(i32),
+    I64(i64),
+    F64(f64),
+}
```
paullegranddc, line 16: I don't really see why you need this runtime representation of how the message pack is encoded?

You could make the code simpler by having two read function
* read_number_signed<T: TryFrom<i64>>() ->  Result<T> 
* read_number_unsigned<T: TryFrom<u64>>() ->  Result<T> 

What would these functions do is:
* Widening the decoded numbers to i64 or u64, since these type would be big enough to represent any signed or unsigned number using the existing [TryFrom<number>  trait implementation](https://doc.rust-lang.org/std/convert/trait.TryFrom.html#implementors) for these, then call T::try_from to cast to the desired output.

See this playground for a small implementation of this idea https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2dfcc2b984bcebb4395179a62087c97f

paullegranddc (commented):
I don't quite understand what the custom deserializer brings.
The goal is to capture `&str` instead of `String` for span fields in the future right?

But from what I know [this can be done with serde](https://serde.rs/lifetimes.html#borrowing-data-in-a-derived-impl), using `&'a str` references for strings, and #[serde::borrow] annotations right?

hoolioh:
> I don't quite understand what the custom deserializer brings. The goal is to capture `&str` instead of `String` for span fields in the future right?
> 
> But from what I know [this can be done with serde](https://serde.rs/lifetimes.html#borrowing-data-in-a-derived-impl), using `&'a str` references for strings, and #[serde::borrow] annotations right?

The idea is to to have this a a building block to gradually introduce future performance optimizations. We're evaluating different mechanisms for reducing allocation overhead: string interning, implementing a string arena, memory pools, etc. Also we would want to test different ways of representing the data. In essence this will give us the flexibility to test those different approaches easily.

paullegranddc:
> The idea is to to have this a a building block to gradually introduce future performance optimizations. We're evaluating different mechanisms for reducing allocation overhead: string interning, implementing a string arena, memory pools, etc. Also we would want to test different ways of representing the data. In essence this will give us the flexibility to test those different approaches easily.

That makes sense

trace-utils/src/msgpack_decoder/v04/number.rs
ekump: If I understand your suggestion correctly we'd change `Number` to be:
```rust
pub enum Number {
    Signed(i64),
    Unsigned(u64),
    Float(f64),
}
```

The existing `read_number` function wouldn't be simplified much because we still need to handle all the different `Marker` types. It would wind up looking like:

```rust
pub fn read_number(buf: &mut &[u8]) -> Result<Number, DecodeError> {
    match rmp::decode::read_marker(buf)
        .map_err(|_| DecodeError::InvalidFormat("Unable to read marker for number".to_owned()))?
    {
        Marker::FixPos(val) => Ok(Number::Unsigned(val as u64)),
        Marker::FixNeg(val) => Ok(Number::Signed(val as i64)),
        Marker::U8 => Ok(Number::Unsigned(
            buf.read_data_u8().map_err(|_| DecodeError::IOError)? as u64,
        )),
        Marker::U16 => Ok(Number::Unsigned(
            buf.read_data_u16().map_err(|_| DecodeError::IOError)? as u64,
        )),
        Marker::U32 => Ok(Number::Unsigned(
            buf.read_data_u32().map_err(|_| DecodeError::IOError)? as u64,
        )),
        Marker::U64 => Ok(Number::Unsigned(
            buf.read_data_u64().map_err(|_| DecodeError::IOError)?,
        )),
        Marker::I8 => Ok(Number::Signed(
            buf.read_data_i8().map_err(|_| DecodeError::IOError)? as i64,
        )),
        Marker::I16 => Ok(Number::Signed(
            buf.read_data_i16().map_err(|_| DecodeError::IOError)? as i64,
        )),
        Marker::I32 => Ok(Number::Signed(
            buf.read_data_i32().map_err(|_| DecodeError::IOError)? as i64,
        )),
        Marker::I64 => Ok(Number::Signed(
            buf.read_data_i64().map_err(|_| DecodeError::IOError)?,
        )),
        Marker::F32 => Ok(Number::Float(
            buf.read_data_f32().map_err(|_| DecodeError::IOError)? as f64,
        )),
        Marker::F64 => Ok(Number::Float(
            buf.read_data_f64().map_err(|_| DecodeError::IOError)?,
        )),
        _ => Err(DecodeError::InvalidType("Invalid number type".to_owned())),
    }
}
```

A bigger problem arises when we implement the `TryFrom` traits and need to downcast.

The following will silently truncate a u64 value to u32:

```rust
impl TryFrom<Number> for u32 {
    type Error = DecodeError;
    fn try_from(value: Number) -> Result<Self, Self::Error> {
        match value {
            Number::Unsigned(val) => Ok(val as u32),
            _ => Err(DecodeError::InvalidConversion(format!(
                "unable to convert {} to u32",
                value
            ))),
        }
    }
}
```

To raise an error when we can't downcast we wind up with:

```rust
impl TryFrom<Number> for u32 {
    type Error = DecodeError;
    fn try_from(value: Number) -> Result<Self, Self::Error> {
        match value {
            Number::Unsigned(val) if val <= u32::MAX as u64 => Ok(val as u32),
            _ => Err(DecodeError::InvalidConversion(format!(
                "unable to convert {} to u32",
                value
            ))),
        }
    }
}
```

I lean slightly towards preferring the existing implementation as it seems a bit clearer to me. @hoolioh you originally implemented this. Do you have any thoughts?

And, apologies if I misunderstood your suggestion.

trace-utils/src/msgpack_decoder/v04/decoder/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+    let v = read_number(buf)?.try_into()?;
+
+    Ok((k, v))
+}
+
+fn read_map_strs(buf: &mut &[u8]) -> Result<HashMap<String, String>, DecodeError> {
+    match decode::read_marker(buf)
+        .map_err(|_| DecodeError::InvalidFormat("Unable to read marker for map".to_owned()))?
+    {
+        Marker::FixMap(len) => {
```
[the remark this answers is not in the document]
ekump: That's a good question. I'm not aware of anything that says v04 can't have maps with > 16 entries. @hoolioh - Are you aware of anything preventing larger maps from being sent from the client libraries? 

I've updated the decoder to also support Map16.

trace-utils/src/msgpack_decoder/v04/decoder/mod.rs
```diff
@@ -0,0 +1,551 @@
...
+        DecodeStringError::TypeMismatch(marker) => {
+            DecodeError::InvalidType(format!("Type mismatch at marker {:?}", marker))
+        }
+        DecodeStringError::InvalidUtf8(_, e) => DecodeError::Utf8Error(e.to_string()),
+        _ => DecodeError::IOError,
+    })
+}
+
+#[inline]
+fn read_string(buf: &mut &[u8]) -> Result<String, DecodeError> {
```
bantonsson, line 102: Can't this function just use `read_string_ref`, and then bump the `buf` like the code in `fill_span`?

trace-utils/src/msgpack_decoder/v04/decoder/span.rs
```diff
@@ -0,0 +1,159 @@
...
+    let (key, value) = read_string_ref(buf)?;
+
+    // Go to the value
+    *buf = value;
+
+    let key = key.parse::<SpanKey>()?;
+
+    match key {
+        SpanKey::Service => {
+            let (value, next) = read_string_ref(buf)?;
```
bantonsson, line 96: Could this use `read_string`?

trace-utils/src/msgpack_decoder/v04/decoder/span.rs
```diff
@@ -0,0 +1,159 @@
...
+            span.name = String::from_str(value).unwrap();
+            *buf = next;
+        }
+        SpanKey::Resource => {
+            let (value, next) = read_string_ref(buf)?;
+            span.resource = String::from_str(value).unwrap();
+            *buf = next;
+        }
+        SpanKey::TraceId => span.trace_id = read_number(buf)?.try_into()?,
```
bantonsson, line 110: Maybe not necessary for this PR, but I think it would be good if the `read_number`, `read_string`, etc. had versions where the _field name_ was provided so the error messages contained what field we were trying to decode when things went wrong.

bantonsson (commented):
This looks good to me overall, and the performance numbers are promising (-20% even without much in the way of optimizations)

trace-utils/src/msgpack_decoder/v04/decoder/mod.rs
ekump: Yes, that works. Good idea!

trace-utils/src/msgpack_decoder/v04/decoder/span.rs
ekump: I agree. I'll create a ticket to address this later.

trace-utils/src/msgpack_decoder/v04/decoder/span.rs
```diff
@@ -0,0 +1,156 @@
...
+                format!("Invalid span key: {}", s).to_owned(),
+            )),
+        }
+    }
+}
+
+fn fill_span(span: &mut Span, buf: &mut &[u8]) -> Result<(), DecodeError> {
+    // field's key won't be held so no need to copy it in a buffer.
+    let key = read_string(buf)?;
+    let key = key.parse::<SpanKey>()?;
```
bantonsson, line 89: This code actually needs to use `read_string_ref` since we don't want to materialize a `String` just to throw it away. It has a big performance impact.

trace-utils/src/msgpack_decoder/v04/decoder/span.rs
bantonsson: This and all the other ones below that directly convert the `&str` to a `String` can use `read_string`.

bantonsson (commented):
The performance regression is real and can be fixed.

ekump:
> The performance regression is real and can be fixed.

@bantonsson - Not only fix, but looks like improve upon: 

> #### scenario:benching deserializing traces from msgpack to their internal representation
> * 🟩 `execution_time` [-1.102µs; -1.099µs] or [-30.072%; -29.998%]
> 
> #### scenario:tags/replace_trace_tags
> * 🟥 `execution_time` [+130.551ns; +133.498ns] or [+3.022%; +3.090%]

I think the `replace_trace_tags` regression is a false positive. The changes shouldn't influence that benchmark at all from what I can tell.

bantonsson (approved):
Now we're back on track. The criterion results are a bit too sensitive to other things happening on the system, but there are definite performance improvements for the serialization.

PR #548, opened 2024-07-25
[Serverless Mini Agent] Add _dd.mini_agent_version tag to all spans for Azure Functions, Google Cloud Functions, and Azure Spring Apps

duncanpharvey:
# What does this PR do?

Adds `_dd.mini_agent_version` span tag to all spans from Azure Functions, Google Cloud Functions, and Azure Spring Apps.

# Motivation

https://datadoghq.atlassian.net/browse/SVLS-5167

# Additional Notes

Previously there was an `aas.environment.mini_agent_version` span tag added for Azure Functions. This tag is being replaced for a span tag that will be applied in all relevant environments: Azure Functions, Google Cloud Functions, and Azure Spring Apps.

Set `DD_MINI_AGENT_VERSION` environment variable to the package version of `datadog-serverless-trace-mini-agent`. Then reference this environment variable in the packages that are run from `datadog-serverless-trace-mini-agent`.

Set the version of `datadog-trace-mini-agent` to match the version of libdatadog so the Serverless Mini Agent version is only managed in a single package moving forward.

# How to test the change?

**Azure Spring Apps**

- Build mini agent locally
- Build an Azure Spring App with persistent storage
- ssh into the Azure Spring App instance and upload the Datadog Java Tracer and Serverless Mini Agent
- Add `-javaagent:/persistent/dd-java-agent.jar` to the JVM options
- Set `DD_TRACE_TRACER_METRICS_ENABLED` to `true`
- Start the Serverless Mini Agent

**Azure Functions / Google Cloud Functions**

- Build mini agent locally
- Deploy Azure Function or Google Cloud Function with mini agent binary in root path
- Set `DD_MINI_AGENT_PATH`
  - Azure: `/home/site/wwwroot/datadog-serverless-trace-mini-agent`
  - Google: `/workspace/datadog-serverless-trace-mini-agent`

### DIFF
--- a/trace-utils/src/trace_utils.rs
+++ b/trace-utils/src/trace_utils.rs
@@ -10,6 +10,7 @@ use rmpv::decode::read_value;
 use rmpv::{Integer, Value};
 use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
+use std::env;
 
 pub use crate::send_data::send_data_result::SendDataResult;
 pub use crate::send_data::SendData;
@@ -453,10 +454,21 @@ pub enum EnvironmentType {
     LambdaFunction,
 }
 
-#[derive(Clone, Default, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct MiniAgentMetadata {
     pub gcp_project_id: Option<String>,
     pub gcp_region: Option<String>,
+    pub version: Option<String>,
+}
+
+impl Default for MiniAgentMetadata {
+    fn default() -> Self {
+        MiniAgentMetadata {
+            gcp_project_id: Default::default(),
+            gcp_region: Default::default(),
+            version: env::var("DD_MINI_AGENT_VERSION").ok(),
+        }
+    }
 }
 
 pub fn enrich_span_with_mini_agent_metadata(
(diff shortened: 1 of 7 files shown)


trace-utils/src/trace_utils.rs
```diff
@@ -471,9 +483,15 @@ pub fn enrich_span_with_mini_agent_metadata(
         span.meta
             .insert("location".to_string(), gcp_region.to_string());
     }
+    if let Some(mini_agent_version) = &mini_agent_metadata.version {
+        span.meta.insert(
+            "_dd.mini_agent_version".to_string(),
```
duncanista, line 488: Has this tag been approved? I'm not sure if we should continue using the name `mini_agent` when this is trace agent.

duncanista: Ah, I jus read `Azure Functions, Google Cloud Functions, and Azure Spring Apps.`

But still, maybe we should switch to a better naming 🤔

duncanpharvey: Any ideas for a better name? The packages are all named some variation of "mini agent" so at least it's consistent for now. And the tag will only be set when the environment variable `DD_MINI_AGENT_VERSION` is set, which should not be the case for Bottlecap.

jcstorms1: I think we need to sort this out at a later point.

jcstorms1 (approved):

PR #552, opened 2024-07-27
[ddcommon] Add token bucket rate limiter implementation

brettlangdon:
# What does this PR do?

Add an implementation of a token bucket rate limiter to `ddcommon`.

# Motivation

Meant as a reference implementation of the rate limiter intended to be used by tracing clients for their sampling implementation.

# Additional Notes

# How to test the change?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -682,7 +682,7 @@ version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "53397bfda19ccb48527faa14025048fc4bb76f090ccdeef1e5a355bfe4a94467"
 dependencies = [
- "proc-macro-crate",
+ "proc-macro-crate 1.3.1",
  "proc-macro2",
  "quote",
  "syn 1.0.109",
@@ -1665,6 +1665,7 @@ dependencies = [
  "maplit",
  "pin-project",
  "regex",
+ "rstest",
  "rustls 0.23.10",
  "rustls-native-certs 0.6.3",
  "serde",
@@ -2137,6 +2138,12 @@ dependencies = [
  "pin-project",
  "pin-utils",
 ]
+
+[[package]]
+name = "futures-timer"
+version = "3.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f288b0a4f20f9a56b5d1da57e2227c661b7b16168e2f72365f57b63326e29b24"
 
 [[package]]
 name = "futures-util"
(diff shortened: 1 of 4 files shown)


bwoebi:
Hm. A couple notes:
- the time resolution on windows is ... rather milliseconds than nanoseconds. Not a fundamental problem, but it makes not much sense to have interval in anything but at least milliseconds, I'd say
- the time is using system time, which is susceptible to ... system time changes, i.e. not monotonic.

I do already have an implementation of a rate limiter https://github.com/DataDog/libdatadog/blob/012dc9a33750008d2c4961e256ea7101a7fb2403/sidecar/src/shm_limiters.rs
I hadn't needed it outside of sidecar yet ... but probably should have pushed it to ddcommon.

That implementation is based on atomics.
I very much need atomics as the purpose of the limiter is being able to work across processes via shared memory, something I can't just Mutex.
Also my use case involves the ability to change the rate limit on an existing limiter. (i.e. being friendly to that scenario).

Do you want to keep your implementation, or should we rather take my limiter implementation and put the `LocalLimiter` I have into ddcommon (and possibly the `ShmLimiter` into ipc crate)?

brettlangdon:
@bwoebi my original implementation was using `clocksource` as a way to avoid the issues with `SystemTime`, that was something I wanted to revisit here.


I don't care too much about whose implementation is taken, but having a shared token bucket rate limiter implementation in `ddcommon` would be good.

brettlangdon:
> That implementation is based on atomics.
> I very much need atomics as the purpose of the limiter is being able to work across processes via shared memory, something I can't just Mutex.

Is there a reason you need it to work across processes?

For most of the language libraries the sampling rate limiter is per-process.

bwoebi:
@brettlangdon The primary problem is that PHP tends to have many processes in some configurations, with autoscaling the number of processes depending on load, making the sample rate ... unpredictable.
Which is why shared rate limiting is needed - you just set it and it's valid for your host.

brettlangdon:
> @brettlangdon The primary problem is that PHP tends to have many processes in some configurations, with autoscaling the number of processes depending on load, making the sample rate ... unpredictable. Which is why shared rate limiting is needed - you just set it and it's valid for your host, which is a perfectly known quantity.

What is the performance overhead of using a shared memory rate limiter vs an in-process one? I'd expect it to be quite heavy?

bwoebi:
@brettlangdon Why would there be any extra overhead (apart from the initial mmap call)? It's just pointing to some memory.

brettlangdon:
> @brettlangdon Why would there be any extra overhead (apart from the initial mmap call during process init)? It's just pointing to some memory.

Wouldn't we now be creating contention between processes on shared memory updates?

bwoebi:
That's why the implementation is based on atomics, to avoid any lock contention.

bwoebi:
Superseeded by #560.

PR #554, opened 2024-07-30
[crashtracker] RFC for structured log format

danielsn:
# What does this PR do?

Defines a structured log format for crash reports

# Motivation

If we want to interoperate with other tools/services, we need an agreed on format.

# Additional Notes

# How to test the change?


rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+An example is given in Appendix A, and the schema is given in Appendix B.
+
+### Required fields
+- `errortype`:
+    Currently, the only value is "crash", but this allows for extension to capture unhandled soft-errors, e.g. "panic", "uncaught exception", etc.
+- `incomplete`:
+    Boolean `false` if the crashreport is complete (i.e. contains all intended data), `true` if there is expected missing data.
+    This can happen becasue the crashtracker is architected to stream data to an out of process receiver, allowing a partial crash report to be emitted even in the case where the crashtracker itself crashed during stack trace collection.
+    This MUST be set to `true` if any required field is missing.
+- `metadata`:
```
OmerRaviv, line 31: For consistency's sake, I propose matching the fields used here to align 1-to-1 with what we already use in instrumentation-telemetry, see [this](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/0cb13e7181acea60d4009d17c0a95fe6b6956e35/GeneratedDocumentation/ApiDocs/v2/development.md?plain=1#L89-L92) and [this](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/0cb13e7181acea60d4009d17c0a95fe6b6956e35/GeneratedDocumentation/ApiDocs/v2/development.md?plain=1#L105-L113). WDYT?

OmerRaviv: Also, should these (i.e. `library_name`, `library_version`, `os_info` even be part of the structured json, given that they are already defined in the instrumentation-telemetry API which we are using to upload the crash reports? (see the `application` and `host` sections of the [required top level json keys](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/development.md#required-top-level-json-keys) in the instrumentation-telemetry api spec)

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+    TODO: What format do users expect here?
+
+### Extensibility
+Future versions of the crashtracker MAY add additional fields.
+Parsers MUST accept unexpected optional fields, either by ignoring them, or by displaying them as additional data.
+The version number SHOULD be incremented for important optional fields, and MUST be incremented when a required field is added or removed.
+
+### Stacktraces
+Different languages and language runtimes have different representations of a stacktrace.
+The representation below attempts to collect as much information.
```
OmerRaviv, line 86:
```suggestion
The representation below attempts to collect as much information as possible.
```

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+  Note that this is a best effort collection: for optimized code, it may be difficult to associate a given instruction back to file, line and column.
+  Also note that a given stack frame may have more than one associated name, e.g. if function inlining has occurred.
+  - `colno`:
+    The column number in the given file where the symbol was defined.
+  - `filename`:
+    The file name where this function was defined.
+  - `lineno`
+    The line number in the given file where the symbol was defined.
+  - `name`
+    The name of the function.
```
OmerRaviv, line 131: Should we consider renaming all of these so that they are consistent with the [names the Error Tracking stack trace parser uses](https://github.com/DataDog/dd-source/blob/main/domains/error-tracking/libs/stacktrace/stacktrace.go#L29-L80)? 

Reasoning:
1. Less confusing for engineers on the Error Tracking team as that's the naming scheme they already have in place
2. Makes sure that we are naming things consistently across client library, backend and frontend and so we have a common language when things break down

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+    The OS + processor architecture on which the crash occurred.
+- `stacktrace`: 
+    This represents the stack of the crashing thread.
+    See below for more details on how stacktraces are formatted.
+- `timestamp`:
+    The time at which the crash occurred, in ISO 8601 format.
+- `uuid`:
+    A UUID which uniquely identifies the crash.
+- `version_id`:
+    A Semver compatible ID for this format. [TODO, should it be semver?]
```
kevingosse, line 49: Semver is probably overkill, I think a simple number is enough

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+
+- `additional_stacktraces`:
+    This field contains a `Map<ThreadId, Stacktrace>`.
+    In a multi-threaded program, the collector SHOULD collect the stacktraces of all active threads, and report them here.
+- `counters`:
+    The crashtracker offers a mechanism for programs to register counters to track which operations were active at the time of the crash.
+    At present, this is only used by the profiler, but this may be extended in the future.
+- `files`:
+    The collector MAY collect useful files, such as `/proc/self/maps` or `/proc/meminfo`, and include them here.
+    Files are stored as an array of plain text strings, one per line.
```
kevingosse, line 64: Shouldn't it be a map? (<filename, contents>)

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+- `proc_info`: 
+    Currently, just tracks the PID of the crashing process.  
+    In the future, this may record additional info about the crashing process.
+- `siginfo`:
+    The name and signal number of the crashing signal (on UNIX systems)
```
kevingosse, line 69: I wonder if we should extend this. Signal is a very unix-specific concept. Applications can also crash because of an unhandled exception, which we have no normalized way to store right now (in .NET I'm currently storing that as a tag)

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+    A Semver compatible ID for this format. [TODO, should it be semver?]
+
+### Optional fields
+Any field not listed as "Required" is optional.
+In order to minimize logging overhead, producers SHOULD NOT emit anything for an optional field.
+Consumers MUST accept json with elided optional fields.
+
+- `additional_stacktraces`:
+    This field contains a `Map<ThreadId, Stacktrace>`.
+    In a multi-threaded program, the collector SHOULD collect the stacktraces of all active threads, and report them here.
```
bwoebi, line 58: There's value in not directly tying it to ThreadId only. Like, you might want to print the managed _and_ the unmanaged stacktrace (i.e. the C-stack trace and the actual stacktrace within the runtime).

kevingosse: For .NET I merged the managed and native stacktraces (sorting the frames by stackpointer) but I don't know if that makes sense for all languages

bwoebi: That makes only sense if both are tied together, which isn't the case in e.g. PHP, which has two separate stacks for PHP and C, with no really discernible connection.

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+      Examples would include "hostname", "service", and any configuration information the system wishes to track.
+- `os_info`: 
+    The OS + processor architecture on which the crash occurred.
+- `stacktrace`: 
+    This represents the stack of the crashing thread.
+    See below for more details on how stacktraces are formatted.
+- `timestamp`:
+    The time at which the crash occurred, in ISO 8601 format.
+- `uuid`:
+    A UUID which uniquely identifies the crash.
```
pawelchcki, line 47: which kind of UUID?
which uniqueness guarantees it should uphold ?

rfc/0001-crashtracker-user-visable-log.md
pawelchcki: Not all stack traces are the same.
Instead of 1 format handling all - lets make sure we accommodate all languages - current and future.

e.g. we can version a stacktrace schema.

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+    Currently, just tracks the PID of the crashing process.  
+    In the future, this may record additional info about the crashing process.
+- `siginfo`:
+    The name and signal number of the crashing signal (on UNIX systems)
+- `span_ids`: 
+    A vector of 128 bit numbers, representing the active span ids at the time of program crash.
+    The collector SHOULD collect as many as it can, but MAY cap the number of spans that it tracks.
+    TODO: What format do users expect here?
+- `trace_ids:`
+    A vector of 128 bit numbers, representing the active span ids at the time of program crash.
```
pawelchcki, line 75: eventualy we might be able to extract trace id active in specific thread.

When we don't have to lets treat span and trace_ids opaquely - as Strings.

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+  "timestamp": "2024-07-19T16:52:16.422378Z",
+  "uuid": "a42add90-0e60-4799-b9f7-cbe0ebec4f27"
+}
+```
+</details>
+
+## Appendix B: Json Schema
+
+<details>
+  <summary>Json Schema</summary>
```
pawelchcki, line 272: Schema should be its own file- so that we can reference it in other repositories or tools used for generation of example code. 

Same for example json.s

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+- **Relative Addresses**
+  Addresses expressed as an offset into a given library or executable.
+  Can be used by backend symbolication to generate debug names etc.
+  These follow the [blazezym](https://github.com/libbpf/blazesym) format for normalized addresses.
+  - `file_offset`: 
+    The relative offset of the symbol, in the base file
+  - `meta`:
+    Metadata to allow the backend symbolizer to identify the file that symbol is in.
+    Currently, this includes the file type: "Apk", "Elf" or "Unknown", as well as the `path` and `build_id` identifying the file.
+- **Debug information (e.g. "names")**
```
milanboers, line 120: Not sure how well the "ip" -> multiple "names" generalizes to managed stack traces, where you may not have the instruction pointer. Perhaps a flat array would work better, simply duplicating the "ip" field.

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+    Metadata about the system in which the crash occurred:
+    - `library_name`
+    - `library_version`
+    - `family`
+    - `tags`:
+      A set of key:value pairs, representing any tags the crashtracking system wishes to associate with the crash.
+      Examples would include "hostname", "service", and any configuration information the system wishes to track.
+- `os_info`: 
+    The OS + processor architecture on which the crash occurred.
+- `stacktrace`:
```
milanboers, line 41: I think it would make sense to move this under the `error` attribute, `additional_stacktraces` too, and also add:
- `"error.is_crash": true` to indicate that this error is a crash (adds a marker and allows to filter for crashes in the Error Tracking UI)
- `error.kind` and `error.message` for the type and message to show respectively (can be made up, on iOS crashes we base this on the signal; for example `SIGABRT (#0)` for kind and `Application crash: SIGABRT (Abort trap)` for message).
- Perhaps an `"error.stacktrace_format": "crashtracker"` so that we will know the format of the stacktrace and how to parse it. (if we can't determine this from the `source` or another field already)

So it becomes closer to [how we recommend customers](https://docs.datadoghq.com/logs/log_collection/?tab=host#attributes-for-stack-traces) do this.

Adding these attributes can start including events into Error Tracking already, fingerprinted based on these attributes (but maybe that's a happy accident? only the stack trace would not show in the UI; you'd only see the kind and message for now)

@GianlucaBortoli

GianlucaBortoli: Totally agree w/ Milan.

> Perhaps an "error.stacktrace_format": "crashtracker" so that we will know the format of the stacktrace and how to parse it. (if we can't determine this from the source or another field already)

if it's a "proper" logs integration, we could definitely use `source` for this instead of yet another attribute IMO

rfc/0001-crashtracker-user-visable-log.md
dougqh: spanIds & traceIds are separate?
I assuming you need the traceId to be able to contextualize the spanId.

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+- `siginfo`:
+    The name and signal number of the crashing signal (on UNIX systems)
+- `span_ids`: 
+    A vector of 128 bit numbers, representing the active span ids at the time of program crash.
+    The collector SHOULD collect as many as it can, but MAY cap the number of spans that it tracks.
+    TODO: What format do users expect here?
+- `trace_ids:`
+    A vector of 128 bit numbers, representing the active span ids at the time of program crash.
+    The collector SHOULD collect as many as it can, but MAY cap the number of spans that it tracks.
+    TODO: What format do users expect here?
```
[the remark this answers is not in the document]
dougqh: Yeah, my thinking to place them in to just store as primitives in a well-known location.
In Java, we need to do that for CSW (kernel sec) anyway.

rfc/0001-crashtracker-user-visable-log.md
danielsn: done

rfc/0001-crashtracker-user-visable-log.md
danielsn: done

rfc/0001-crashtracker-user-visable-log.md
danielsn: done

rfc/0001-crashtracker-user-visable-log.md
danielsn: fixed

rfc/0001-crashtracker-user-visable-log.md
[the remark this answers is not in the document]
danielsn: There is a mechanism in libdatadog to stash them as atomic ints. 
 It requires the tracer to play along by calling the `insert` and `remove` functions but that shouldn't be hard to implement https://github.com/DataDog/libdatadog/blob/main/crashtracker-ffi/src/collector/spans.rs#L60

rfc/0001-crashtracker-user-visable-log.md
danielsn: done

rfc/0001-crashtracker-user-visable-log.md
danielsn: Extended.

rfc/0001-crashtracker-user-visable-log.md
danielsn: clarified that its v4

rfc/0001-crashtracker-user-visable-log.md
danielsn: Great idea, will do that

rfc/0001-crashtracker-user-visable-log.md
danielsn: done

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,292 @@
...
+    See below for more details on how stacktraces are formatted.
+- `is_crash`:
+    Boolean true if the error was a crash, false otherwise.
+- `kind`:
+    The kind of error that occurred.
+    For a crash, a non-exhaustive set of options include "SigAbort", "SigBus", "SigSegv".
+    This field MAY be extended to include options such at "Panic", "UnhandledException" etc.
+- `message`:
+    A human readable string containing an error message associated with the stack trace.
+- `source`:
```
milanboers, line 65:
```suggestion
- `source_type`:
```
to make this consistent with the naming used on the RUM SDKs

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,292 @@
...
+    It may or may not be demangled (e.g. "_ZNSt28__atomic_futex_unsigned_base26_M_futex_wait_until_steadyEPjjbNSt6chrono8durationIlSt5ratioILl1ELl1EEEENS2_IlS3_ILl1ELl1000000000EEEE" vs "std::__atomic_futex_unsigned_base::_M_futex_wait_until_steady")
+
+### Other data
+
+## Appendix A: Example output
+<details>
+  <summary>Example Output</summary>
+
+```json
+{
```
milanboers, line 164: (this doesn't seem to be updated yet)

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,292 @@
...
+    See below for more details on how stacktraces are formatted.
+- `timestamp`:
+    The time at which the crash occurred, in ISO 8601 format.
+- `uuid`:
+    A UUID v4 which uniquely identifies the crash.
+- `version_id`:
+    An integer version id.
+
+### ErrorData
+- `additional_stacks`:
```
milanboers, line 52: Are these always threads? And are the `Identifier`s strings (of thread id's / names)?

If so, I think it would make sense to use the same format as RUM SDK's do where possible. An array `threads` of objects with attribute:
- `name` for the thread name (same as Identifier here? Can just be "Thread #<id>" which is what we do on iOS)
- `stack` for the stack trace (can be the structured format for the crash tracker)
- (optional, can leave out for the crash tracker format) `crashed` boolean for whether this thread crashed
- (optional, can leave out for the crash tracker format) `state` for thread state like RUNNING, BLOCKED etc. (different on each platform)

[example of an org2 iOS crash](https://app.datadoghq.com/error-tracking/unified?query=&issueId=616bbd8c-825d-11ee-9764-da7ad0900002&source=all&view=spans)

```
            "threads": {
              "type": "array",
              "description": "Description of each thread in the process when error happened.",
              "items": {
                "type": "object",
                "description": "Description of the thread in the process when error happened.",
                "required": ["name", "crashed", "stack"],
                "properties": {
                  "name": {
                    "type": "string",
                    "description": "Name of the thread (e.g. 'Thread 0').",
                    "readOnly": true
                  },
                  "crashed": {
                    "type": "boolean",
                    "description": "Tells if the thread crashed.",
                    "readOnly": true
                  },
                  "stack": {
                    "type": "string",
                    "description": "Unsymbolicated stack trace of the given thread.",
                    "readOnly": true
                  },
                  "state": {
                    "type": "string",
                    "description": "Platform-specific state of the thread when its state was captured (CPU registers dump for iOS, thread state enum for Android, etc.).",
                    "readOnly": true
                  }
                }
              }
            },
```

image

danielsn: Done

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: done

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,724 @@
...
+    The base address of the module to which the frame belongs
+- **Relative Addresses**
+  Addresses expressed as an offset into a given library or executable.
+  Can be used by backend symbolication to generate debug names etc.
+  These follow the [blazezym](https://github.com/libbpf/blazesym) format for normalized addresses.
+  - `file_offset`: 
+    The relative offset of the symbol, in the base file
+  - `meta`:
+    Metadata to allow the backend symbolizer to identify the file that symbol is in.
+    Currently, this includes the file type: "Apk", "Elf" or "Unknown", as well as the `path` and `build_id` identifying the file.
```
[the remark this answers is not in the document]
danielsn: Correct, build_id is per frame

rfc/0001-crashtracker-user-visable-log.md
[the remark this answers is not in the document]
danielsn: When you say "grouped together" what do you mean?

rfc/0001-crashtracker-user-visable-log.md
danielsn: Modified as per the comment below.

rfc/0001-crashtracker-user-visable-log.md
danielsn: done

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: done

rfc/0001-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,703 @@
...
+## Summary
+This document describes version 1 of the crashinfo data format.
+
+## Motivation
+The `libdatadog` crashtracker detects program crashes.
+It automatically collects information relevant to the characterizing and debugging the crash, including stack-traces, the crash-type (e.g. SIGSIGV, SIGBUS, etc) crash, the library version, etc.
+This RFC establishes a standardized logging format for reporting this information.
+
+### Why structured json
+As a text-based format, json can be written to standard logging endpoints.
```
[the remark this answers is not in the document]
milanboers: I think you can try in #logs-core :)

milanboers (approved):

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+- `error`:
+    A structure of type `ErrorData`, described below.
+- `incomplete`:
+    Boolean `false` if the crashreport is complete (i.e. contains all intended data), `true` if there is expected missing data.
+    This can happen becasue the crashtracker is architected to stream data to an out of process receiver, allowing a partial crash report to be emitted even in the case where the crashtracker itself crashed during stack trace collection.
+    This MUST be set to `true` if any required field is missing.
+- `metadata`:
+    Metadata about the system in which the crash occurred:
+    - `library_name`
+    - `library_version`
```
jbachorik, line 34: Just to clarify - eg. for Java the crashtracking is bundled with dd-trace-java. Should the library name be `dd-trace-java` then?

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    - `family`
+    - `tags`:
+      A set of key:value pairs, representing any tags the crashtracking system wishes to associate with the crash.
+      Examples would include "hostname", "service", and any configuration information the system wishes to track.
```
jbachorik, line 38: Apparently, there is an expectation to use `severity:crash` tag for crash reports.
Is the tag to be set here? If it is to be set here, isn't it redundant? We are already sending a crash report so we know this is a crash. What extra information is conveyed by setting the `severity:crash` tag?

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+- `os_info`: 
+    The OS + processor architecture on which the crash occurred.
+- `stacktrace`: 
+    This represents the stack of the crashing thread.
+    See below for more details on how stacktraces are formatted.
```
jbachorik, line 43: This will be duplicated with the `stack` field in `ErrorData` (and potentially by the stacktrace of a thread with `crashed` set to `true`, if `threads` field is provided)?
Why is this duplication necessary?

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+- `timestamp`:
+    The time at which the crash occurred, in ISO 8601 format.
+- `uuid`:
+    A UUID v4 which uniquely identifies the crash.
```
jbachorik, line 47: This is an arbitrary UUID, correct? Eg. I can just generate the UUID at the time I intercept the crash and use it to upload the crash data?

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+- `version_id`:
```
jbachorik, line 48: What is this version related to? It's not the crash tracking library version, is it?
Is this the crashtracking data schema version?

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    Currently, just tracks the PID of the crashing process.  
+    In the future, this may record additional info about the crashing process.
+- `siginfo`:
+    The name and signal number of the crashing signal (on UNIX systems)
+- `span_ids`: 
+    A vector of string identifiers, representing the active span ids at the time of program crash.
+    The collector SHOULD collect as many as it can, but MAY cap the number of spans that it tracks.
+- `trace_ids:`
+    A vector of string identifiers, representing the active span ids at the time of program crash.
+    The collector SHOULD collect as many as it can, but MAY cap the number of spans that it tracks.
```
jbachorik, line 96: Could the structure be changed to put all the top-level fields into the same paragraph, either marking each field as required/optional or creating required and optional subgroups?
The current structure is a bit confusing - one reads first about required fields, then jumps into ErrorData where some fields are also marked as optional and then gets back to optional fields.

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    A human readable string containing an error message associated with the stack trace.
+- `source_type`:
+    The string "crashtracking".
+- `stack`:
+    This represents the stack of the crashing thread.
+    See below for more details on how stacktraces are formatted.
+
+### Optional fields
+Any field not listed as "Required" is optional.
+In order to minimize logging overhead, producers SHOULD NOT emit anything for an optional field.
```
jbachorik, line 77: I am not sure I completely understand this sentence - does it mean that we have fields that should never be filled? Or I am just reading it wrong?

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    A vector of string identifiers, representing the active span ids at the time of program crash.
+    The collector SHOULD collect as many as it can, but MAY cap the number of spans that it tracks.
+- `trace_ids:`
+    A vector of string identifiers, representing the active span ids at the time of program crash.
+    The collector SHOULD collect as many as it can, but MAY cap the number of spans that it tracks.
+
+### Extensibility
+Future versions of the crashtracker MAY add additional fields.
+Parsers MUST accept unexpected optional fields, either by ignoring them, or by displaying them as additional data.
+The version number SHOULD be incremented for important optional fields, and MUST be incremented when a required field is added or removed.
```
jbachorik, line 101: What is an 'important optional field'? This sounds rather vaguely and can be misinterpreted, I am afraid.

What about using semver with minor bump for changes to optional fields and major bump for changes in required fields? 
The reasoning is that a receiver would be able to process any data with the same major version because only optional fields are affected and the compatibility is not broken by any change to the required fields.

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+  - `column`:
+    The column number in the given file where the symbol was defined.
+  - `file`:
+    The file name where this function was defined.
+  - `line`
+    The line number in the given file where the symbol was defined.
+  - `function`
+    The name of the function.
+    This may or may not include module information.
+    It may or may not be demangled (e.g. "_ZNSt28__atomic_futex_unsigned_base26_M_futex_wait_until_steadyEPjjbNSt6chrono8durationIlSt5ratioILl1ELl1EEEENS2_IlS3_ILl1ELl1000000000EEEE" vs "std::__atomic_futex_unsigned_base::_M_futex_wait_until_steady")
```
jbachorik, line 159: Is this a free-text field?
For Java I have already pre-formatted frames in the crash report. There are several types of frames (Native, VM, Compiled Java and Interpreted Java) and each of them has slightly different format. Therefore I would really need to be able to provide an opaque representation of the frame.

docs/RFCs/0002-crashtracker-user-visable-log.md
kevingosse: `severity:crash` was used by .NET before we actually decide what to use. Moving forward we will use `is_crash: true` because this is recognized by error tracking

kevingosse:
In the current implementation there are two ways to add tags (on crashinfo and on metadata). I _think_ they mean a different thing: I would say the ones on metadata are used to classify the crash, whereas the ones on crashinfo are used to add arbitrary debug information that can't be stored otherwise. If that interpretation is correct, I believe they should be given different names to avoid confusion.

kevingosse:
Also, the tags in crashinfo can be easily searched on in the UI:

image

Whereas the ones in metadata can't:

image

We probably should change the format of the tags accordingly.

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+
+A stacktrace consists of 
+  - `format`:
+    An identifier describing the format of the stack trace.
+    Allows for extensibility to support different stack trace formats.
+    The format described below is identfied using the string "CrashTrackerV1"
+  - `frames`:
+    An array of `StackFrame`, described below
+
+#### StackFrame
```
bwoebi, line 119: I'd like to request some data about the actual functions. Especially in C code there's information to be had "this variable was a null pointer - this looks like freed memory - probably valid address".

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    This field MAY be extended to include options such at "Panic", "UnhandledException" etc.
+- `message`:
+    A human readable string containing an error message associated with the stack trace.
+- `source_type`:
+    The string "crashtracking".
+- `stack`:
+    This represents the stack of the crashing thread.
+    See below for more details on how stacktraces are formatted.
+
+### Optional fields
```
bwoebi, line 75: Could we please include the address on which the access failed? si_info contains it trivially - e.g. "was it a NULL ptr"?

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    The base address of the module to which the frame belongs
+- **Relative Addresses**
+  Addresses expressed as an offset into a given library or executable.
+  Can be used by backend symbolication to generate debug names etc.
+  These follow the [blazezym](https://github.com/libbpf/blazesym) format for normalized addresses.
+  - `file_offset`: 
+    The relative offset of the symbol, in the base file
+  - `meta`:
+    Metadata to allow the backend symbolizer to identify the file that symbol is in.
+    Currently, this includes the file type: "Apk", "Elf" or "Unknown", as well as the `path` and `build_id` identifying the file.
```
gleocadie, line 146: in case of Windows, we could have the type `PE`, the path, a hash and a timestamp.
We could also reuse the `build-id` field and store the hash and timestamp in a special format.

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+- **Debug information (e.g. "names")**
+  Human readable debug information representing the location of the stack frame in the high-level code.
+  Note that this is a best effort collection: for optimized code, it may be difficult to associate a given instruction back to file, line and column.
+  - `column`:
+    The column number in the given file where the symbol was defined.
+  - `file`:
+    The file name where this function was defined.
```
tlhunter, line 153: Is this an absolute path or relative to, say, the project directory?

tlhunter: Normalizing might help associate the same error regardless of build / execution location:

```
/Users/daniel.schwartznarbonne/foo/src/backtrace/libunwind.rs
/Users/thomas.hunter/foo/src/backtrace/libunwind.rs
```

docs/RFCs/0002-crashtracker-user-visable-log.md
r1viollet: what would you call it ? Ivo has a PR opened here https://github.com/DataDog/libdatadog/pull/663/files to add this in.

docs/RFCs/0002-crashtracker-user-visable-log.md
nsavoire, line 146: `meta` seems to be closely tied to blazesym implementation, not sure that `Apk` support is something that we need (and `Apk` is just like `Elf` without a buildID).
Perhaps it would make sense to use a more generic and easier to parse structure (one whose fields do not change depending on the type).
Instead of:
```
"meta": {
	"Elf" : {
 		"path": "/usr/share/dotnet/shared/Microsoft.NETCore.App/8.0.0/libcoreclr.so",
		"build_id": [...]
	}
}
```
something like:
```
"meta": {
	"path": "...",
	"build_id": "...",
	"type": "elf" / "pdb",
	"age": xxx,
}
```

nsavoire: Could `age` be appended to `guid` and the result stored into `build_id` ?

nsavoire, line 146: It would be nicer to represent build id as a string instead of a byte array.
GNU build IDs are usually represented as hex string  (eg. "9944168df12b0b9b152113c4ad663bc27797fb15").

kevingosse:
As we start to look into symbolication, I realize that adding all the module information at each stackframe is really inefficient and makes the crash report much bigger than it should. Shall we move the modules to a separate section, and only put the module base address in the stackframe?

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+This address may be given as an absolute address, or a `NormalizedAddress`, which can be used by backend symbolication.
+We follow the [blazezym](https://github.com/libbpf/blazesym) format for normalized addresses.
+
+A stacktrace consists of 
+  - `format`:
+    An identifier describing the format of the stack trace.
+    Allows for extensibility to support different stack trace formats.
+    The format described below is identfied using the string "CrashTrackerV1"
+  - `frames`:
+    An array of `StackFrame`, described below
```
nsavoire, line 117: Are inlined functions expected to be represented by separate `StackFrames` with repeated ip/sp/... ?
It might be useful to explicitly mention it.

docs/RFCs/0002-crashtracker-user-visable-log.md
gleocadie: Yes, we could have `build_id` being a concatenation of the `guid` and the `age` (using a well-known separator)

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    This MUST be set to `true` if any required field is missing.
+- `metadata`:
+    Metadata about the system in which the crash occurred:
+    - `library_name`
+    - `library_version`
+    - `family`
+    - `tags`:
+      A set of key:value pairs, representing any tags the crashtracking system wishes to associate with the crash.
+      Examples would include "hostname", "service", and any configuration information the system wishes to track.
+- `os_info`:
```
nsavoire, line 39: It seems there is some redundancy between `os_info` and top level `host` field in telemetry message.

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    At present, this is only used by the profiler, but this may be extended in the future.
+- `files`:
+    The collector MAY collect useful files, such as `/proc/self/maps` or `/proc/meminfo`, and include them here.
+    Files are stored as a `Map<filename, contents>` where `contents` is an array of plain text strings, one per line.
+- `proc_info`: 
+    Currently, just tracks the PID of the crashing process.  
+    In the future, this may record additional info about the crashing process.
+- `siginfo`:
+    The name and signal number of the crashing signal (on UNIX systems)
+- `span_ids`:
```
nsavoire, line 91: span / trace ids should be associated to thread id if possible

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    See below for more details on how stacktraces are formatted.
+- `is_crash`:
+    Boolean true if the error was a crash, false otherwise.
+- `kind`:
+    The kind of error that occurred.
+    For a crash, a non-exhaustive set of options include "SigAbort", "SigBus", "SigSegv".
+    This field MAY be extended to include options such at "Panic", "UnhandledException" etc.
+- `message`:
+    A human readable string containing an error message associated with the stack trace.
+- `source_type`:
```
nsavoire, line 69: It would be nice to have this information at the top level in telemetry payload. I noticed that there is a [Origin](https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/header.go#L87) field that could be used for this.

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: clarified

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    Metadata about the system in which the crash occurred:
+    - `library_name`
+    - `library_version`
+    - `family`
+    - `tags`:
+      A set of key:value pairs, representing any tags the crashtracking system wishes to associate with the crash.
+      Examples would include "hostname", "service", and any configuration information the system wishes to track.
+- `os_info`: 
+    The OS + processor architecture on which the crash occurred.
+- `stacktrace`:
```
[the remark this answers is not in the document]
danielsn: done

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: Exactly.

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: done

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+- `counters`:
+    The crashtracker offers a mechanism for programs to register counters to track which operations were active at the time of the crash.
+    At present, this is only used by the profiler, but this may be extended in the future.
+- `files`:
+    The collector MAY collect useful files, such as `/proc/self/maps` or `/proc/meminfo`, and include them here.
+    Files are stored as a `Map<filename, contents>` where `contents` is an array of plain text strings, one per line.
+- `proc_info`: 
+    Currently, just tracks the PID of the crashing process.  
+    In the future, this may record additional info about the crashing process.
+- `siginfo`:
```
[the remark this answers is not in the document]
danielsn: done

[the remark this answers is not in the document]
danielsn: added

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: changed to semver

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: done

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+  - `symbol_address`:
+    The starting symbol address of the frame of this function.
+    This will attempt to rewind the instruction pointer returned by ip to the start of the function, returning that value.
+    In some cases, however, backends will just return ip from this function.
+  - `module_base_address`:
+    The base address of the module to which the frame belongs
+- **Relative Addresses**
+  Addresses expressed as an offset into a given library or executable.
+  Can be used by backend symbolication to generate debug names etc.
+  These follow the [blazezym](https://github.com/libbpf/blazesym) format for normalized addresses.
```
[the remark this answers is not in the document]
danielsn: done

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: noted

docs/RFCs/0002-crashtracker-user-visable-log.md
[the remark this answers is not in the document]
danielsn: This file describes a prospective future format.

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: Clarified

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: Clarified that this may be absolute or relative paths

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: https://github.com/DataDog/libdatadog/issues/727

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    The kind of error that occurred.
+    For a crash, a non-exhaustive set of options include "SigAbort", "SigBus", "SigSegv".
+    This field MAY be extended to include options such at "Panic", "UnhandledException" etc.
+- `message`:
+    A human readable string containing an error message associated with the stack trace.
+- `source_type`:
+    The string "crashtracking".
+- `stack`:
+    This represents the stack of the crashing thread.
+    See below for more details on how stacktraces are formatted.
```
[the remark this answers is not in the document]
danielsn: fixed

docs/RFCs/0002-crashtracker-user-visable-log.md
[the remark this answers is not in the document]
danielsn: added TODO

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+In order to minimize logging overhead, producers SHOULD NOT emit anything for an optional field.
+Consumers MUST accept json with elided optional fields.
+
+- `counters`:
+    The crashtracker offers a mechanism for programs to register counters to track which operations were active at the time of the crash.
+    At present, this is only used by the profiler, but this may be extended in the future.
+- `files`:
+    The collector MAY collect useful files, such as `/proc/self/maps` or `/proc/meminfo`, and include them here.
+    Files are stored as a `Map<filename, contents>` where `contents` is an array of plain text strings, one per line.
+- `proc_info`:
```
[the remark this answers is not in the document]
danielsn: updated

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+    The starting symbol address of the frame of this function.
+    This will attempt to rewind the instruction pointer returned by ip to the start of the function, returning that value.
+    In some cases, however, backends will just return ip from this function.
+  - `module_base_address`:
+    The base address of the module to which the frame belongs
+- **Relative Addresses**
+  Addresses expressed as an offset into a given library or executable.
+  Can be used by backend symbolication to generate debug names etc.
+  These follow the [blazezym](https://github.com/libbpf/blazesym) format for normalized addresses.
+  - `file_offset`:
```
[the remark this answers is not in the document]
danielsn: fixed

docs/RFCs/0002-crashtracker-user-visable-log.md
```diff
@@ -0,0 +1,169 @@
...
+
+### ErrorData
+- `threads`:
+    This **optional** field contains an array of `Thread` objects.
+    In a multi-threaded program, the collector SHOULD collect the stacktraces of all active threads, and report them here.
+    A `Thread` object has the following fields:
+    - `crashed`: a boolean which tells if the thread crashed.
+    - `name`: Name of the thread (e.g. 'Thread 0').
+    - `stack`: The `StackTrace` of the thread
+    - `state`: **Optional**. Platform-specific state of the thread when its state was captured (CPU registers dump for iOS, thread state enum for Android, etc.).
```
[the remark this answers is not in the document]
danielsn: yes, for now

docs/RFCs/0002-crashtracker-user-visable-log.md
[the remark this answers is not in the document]
danielsn: fixed

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: updated

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: done

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: https://github.com/DataDog/libdatadog/issues/728

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: You mean like the actual values of local variables on the stack?

0005-crashtracker-structured-log-format.md
```diff
@@ -0,0 +1,203 @@
...
+- `trace_ids:`: **[optional]**
+  A vector representing active span ids at the time of program crash.
+  The collector MAY cap the number of spans that it tracks.
+  - `thread_id`: **[optional]**
+    A string containing the thread_id for the given trace.
+  - `trace_id`: **[required]**
+    A string containing the trace id.
+- `uuid`: **[required]**
+  A UUID v4 which uniquely identifies the crash.
+  This will typically be generated at crash-time, and then associated with the uploaded crash.
```
sanchda, line 126: After this, should add a top-level `fingerprint` field which is an arbitrary string used by the system to fingerprint crashes.

0005-crashtracker-structured-log-format.md
```diff
@@ -0,0 +1,203 @@
...
+  - `architecture`: **[required]**
+  - `bitness`: **[required]**
+  - `os_type`: **[required]**
+  - `version`: **[required]**
+- `proc_info`: **[optional]**
+  A place to store information about the crashing process.
+  In the future, this may have additional optional fields as more data is collected.
+  - `pid`: **[required]**
+    The PID of the crashing process.
+- `sig_info`: **[optional]**
```
sanchda, line 100: As discussed, might be good to harmonize the names of these with the field names used in the siginfo_t.  In particular, we'll rename faulting_address -> si_addr

docs/RFCs/0002-crashtracker-user-visable-log.md
danielsn: Yes

0005-crashtracker-structured-log-format.md
danielsn: done

0005-crashtracker-structured-log-format.md
danielsn: done

PR #568, opened 2024-08-01
[profiling] implement file exporter

taegyunkim:
# What does this PR do?

- Refactor `ddtelemetry::worker::http_client::HttpClient` to `ddcommon::http_client::HttpClient`
- Change `datadog_profiling::exporter::Exporter` to use above

# Motivation

We'd like to export to pprof file using libdatadog 

# Additional Notes

# How to test the change?

### DIFF
--- a/crashtracker/src/crash_info/telemetry.rs
+++ b/crashtracker/src/crash_info/telemetry.rs
@@ -162,7 +162,7 @@ impl TelemetryCrashUploader {
                 count: 1,
             }]),
         };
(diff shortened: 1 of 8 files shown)


sanchda (approved):
Overall, I think this looks pretty good.

It isn't clear to me from the called URI processing code.  I presume we follow the URI [RFC](https://datatracker.ietf.org/doc/html/rfc8089), but would `file:///foo/bar/baz` be interpreted as the local path `foo/bar/baz` or the absolute path `/foo/bar/baz` (do you know)?

PR #570, opened 2024-08-02
Add concentrator

VianneyRuhlmann:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


trace-utils/src/trace_utils.rs
```diff
@@ -410,6 +412,13 @@ pub fn compute_top_level_span(trace: &mut [pb::Span]) {
     }
 }
 
+pub fn has_top_level(span: &pb::Span) -> bool {
```
ekump, line 415: These public functions should have unit tests

data-pipeline/src/concentrator/mod.rs
```diff
@@ -0,0 +1,1158 @@
...
+            ),
+            buffer_len: 2,
+            peer_tags_aggregation,
+            compute_stats_by_span_kind,
+            peer_tag_keys,
+        }
+    }
+
+    pub fn add_span(&mut self, span: &pb::Span) -> Result<()> {
+        if !(trace_utils::has_top_level(span)
```
ekump, line 92: minor: I think this if statement is complex enough that it should be moved into a separate function for readability / maintainability purposes. Something along the lines of:
```rust
fn should_ignore_span(span: &Span, compute_stats_by_span_kind: bool) -> bool {
    !(trace_utils::has_top_level(span)
        || trace_utils::is_measured(span)
        || (self.compute_stats_by_span_kind && compute_stats_for_span_kind))
        || trace_utils::is_partial_snapshot(span)
}

File: data-pipeline/src/concentrator/aggregation.rs [hunk 3]
```diff
@@ -0,0 +1,497 @@
...
+
+const TAG_STATUS_CODE: &str = "http.status_code";
+const TAG_SYNTHETICS: &str = "synthetics";
+const TAG_SPANKIND: &str = "span.kind";
+const TAG_ORIGIN: &str = "_dd.origin";
+
+/// This struct represent the key used to group spans together to compute stats.
+#[derive(Debug, Hash, PartialEq, Eq, Clone, Default)]
+pub struct AggregationKey {
+    pub resource_name: String,
```
bantonsson, line 19: I don't think that any of these fields need to be `pub`. The same goes for the `struct` that can be `pub(crate)`.

data-pipeline/src/concentrator/aggregation.rs
```diff
@@ -0,0 +1,497 @@
...
+    pub service_name: String,
+    pub operation_name: String,
+    pub span_type: String,
+    pub span_kind: String,
+    pub http_status_code: u32,
+    pub is_synthetics_request: bool,
+    pub peer_tags: Vec<Tag>,
+    pub is_trace_root: bool,
+}
```
bantonsson, line 28: This is not for this PR, but maybe the `AggregationKey` should cache a precomputed hash value (since it's immutable anyway), and implement `Hash` by itself.

data-pipeline/src/concentrator/aggregation.rs
```diff
@@ -0,0 +1,497 @@
...
+        let peer_tags = if peer_tags_aggregation && client_or_producer(&span_kind) {
+            get_peer_tags(span, peer_tag_keys)
+        } else {
+            vec![]
+        };
+        Self {
+            resource_name: span.resource.clone(),
+            service_name: span.service.clone(),
+            operation_name: span.name.clone(),
+            span_type: span.r#type.clone(),
```
bantonsson, line 54: It feels a bit costly to _always_ (at least for relevant spans) construct this whole Key by cloning a number of things, only to use it to look up things, and eventually throw it away. Something to keep in mind when we start measuring this code.

bantonsson (approved):
Looks good in general.

data-pipeline/src/concentrator/aggregation.rs
VianneyRuhlmann: I agree, I think https://docs.rs/hashbrown/latest/hashbrown/hash_map/struct.HashMap.html#method.raw_entry_mut would be the way to go. I added APMSP-1331 to track this.

bantonsson: Great

data-pipeline/src/concentrator/aggregation.rs
VianneyRuhlmann: I made the struct pub(super) since it's only used in the Concentrator struct

PR #576, opened 2024-08-06
Stabilize criterion benchmark results

bantonsson:
# What does this PR do?

This PR tries to stabilize the criterion micro benchmarks results. The reason why the benchmark result comment is so full of changes is that the batching and how the benchmarks are run has changed compared to `main`.

# Motivation

There is soo much noise in the results that they trigger false positives on almost every PR.

# Additional Notes

# How to test the change?

I have run the the benchmarks repeatedly on a separate PR #577 (that has no code changes), and only occasionally will there be a change in a benchmark result.

### DIFF
--- a/trace-normalization/benches/normalization_utils.rs
+++ b/trace-normalization/benches/normalization_utils.rs
@@ -1,12 +1,19 @@
 // Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
-use criterion::{criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
+use criterion::measurement::WallTime;
+use criterion::Throughput::Elements;
+use criterion::{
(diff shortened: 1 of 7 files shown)


trace-normalization/benches/normalization_utils.rs
```diff
@@ -15,50 +21,63 @@ fn normalize_service_bench(c: &mut Criterion) {
...
+            ),
             *case,
             |b, case| {
                 b.iter_batched_ref(
                     || case.to_owned(),
-                    datadog_trace_normalization::normalize_utils::normalize_name,
-                    BatchSize::NumIterations(100000),
+                    |mutable| {
+                        (0..elements).for_each(|_| {
+                            mutable.insert_str(0, case);
```
bantonsson, line 71: This is not ideal, having a memcopy in the measured part, but we need to measure over multiple calls to minimize the noise 🤷🏼‍♂️

bantonsson: I'm fixing this to allocate the Strings up front.

ekump (approved):

PR #578, opened 2024-08-06
Support x-datadog-test-session-token for parallel tests with the test agent

bwoebi:
I've decided to put it into Endpoint, as it has to be used at basically any place Endpoint is used.


pawelchcki (approved):

pawelchcki:
> scenario:credit_card/is_card_number/37828224631000521389798
> 🟥 execution_time [+11.565ns; +12.044ns] or [+10.646%; +11.087%]
> 🟥 throughput [-918917.691op/s; -883968.291op/s] or [-9.982%; -9.602%]

This is benchmark is propably sensing larger binary, or just depends too much memory layout. We should make sure its not that sensitive.

PR #579, opened 2024-08-06
Add document describing the new build process.

hoolioh:
# What does this PR do?

Describe a new build system to be implemented in libdatadog.

# Motivation

Improve current the current one so it's less error prone, more maintainable and cand handle features as well as versioning.


docs/RFCs/build-process.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Use cargo-make
+cargo-make enables to define and configure sets of tasks and run them as a flow. A task is a command, script, rust
+code, or other sub tasks to execute. Tasks can have dependencies which are also tasks that will be executed before the
+task itself.
+
+### Use cargo-post
+A cargo wrapper which runs a post build script whenever the current build runs successfully.
+
+## Proposed solution
```
pawelchcki, line 27: I've noticed in other projects a similar "trope" 

https://github.com/cross-rs/cross/blob/main/xtask/src/main.rs

They define a custom https://github.com/cross-rs/cross/blob/main/.cargo/config.toml alias. That essentially makes a local crate execute a custom cargo command:

This way - they get things like:

```
cargo xtask changelog
cargo xtask test
```
as the main interface for interacting with the repository that is shared both with CI and local dev workflow.

IMO this looks neat and avoid having to deal with additional limitations of Makefiles etc. 

it can also be used together with build.rs 

But to me it would be good to have common tooling that can be executed explicitly at will.

docs/RFCs/0002-build-process.md
```diff
@@ -0,0 +1,86 @@
+# Building process proposal
```
danielsn, line 1:
```suggestion
# RFC 0002: Building process proposal
```

docs/RFCs/0002-build-process.md
```diff
@@ -0,0 +1,86 @@
+# RFC 0002: Building process proposal
+
+## Context
+As described in the [RFC](https://github.com/DataDog/libdatadog/blob/main/docs/RFCs/delivering-different-features.md)
```
danielsn, line 4: that RFC has been renamed

docs/RFCs/build-process.md
hoolioh: Agree, when I reviewed cargo-post I thought exactly the same. Once the build system is in place we can tackle this.

pawelchcki (approved):

docs/RFCs/0002-build-process.md
```diff
@@ -0,0 +1,86 @@
+# RFC 0002: Building process proposal
+
+## Context
+As described in the [RFC](https://github.com/DataDog/libdatadog/blob/main/docs/RFCs/0001-delivering-different-features.md)
+about delivering different features the current build process depends heavily on shell scripts which makes it too 
+difficult in terms of managning the process itself like handling errors, dependendencies and features. Also shell
+scripts by themselves poses some disadvantages like portability and maintanibilty.
+
+This solution aims to get rid of those problems so the whole procedure depends as much as possible of built-in features 
+of cargo making it less error prone and also improving its maintanibilty.
```
gleocadie, line 10: Agree🤔 
I would say that shell script allows to have the same command everywhere instead of searching for the command in the yaml file.

Do not get me wrong, I'm not against just saying that they are useful (in some case).

Now, I see that we can build, generate binding in one-line command thanks for that🙇‍♂️

docs/RFCs/0002-build-process.md
```diff
@@ -0,0 +1,86 @@
...
+├─ bin/
+│  ├─ libdatadog-crashtracking-receiver
+├─ include/
+│  ├─ datadog/
+│     ├─ common.h
+│     ├─ profiling.h
+│     ├─ telemetry.h
+│     ├─ data-pipeline.h
+├─ lib/
+│  ├─ libdatadog.a
```
gleocadie, line 79: the `.so` file will be there too ?

docs/RFCs/0002-build-process.md
gleocadie: > Now, I see that we can build, generate bindings
    ^-------------------------- am I correct ?

hoolioh: Yep, in this case the cons were clearly outbalancing the pros.

PR #584, opened 2024-08-09
[APMSP-1013] Add stats exporter

VianneyRuhlmann:
# What does this PR do?

This PR add stats computation in the data pipeline crate by:
- Using a SpanConcentrator to compute stats on received spans in the trace exporter
- Introducing a StatsExporter to send stats on a every time a stats bucket expires
- Dropping spans that are not going to be sampled by the trace agent
- Add support for client-side stats http header
- Update ffi bindings to allow stats computation 

# Motivation
The support for client side stats will allow using client side stats by default. This will reduce the number of span received by the agent.

# Additional Notes
This PR does not allow fetching the list of peer tags from the /info endpoint of the agent.

# How to test the change?


bantonsson:
You need to rebase on `main` if you want the benchmarks numbers to make sense...

data-pipeline/src/stats_exporter.rs
```diff
@@ -0,0 +1,362 @@
...
+    cancellation_token: CancellationToken,
+}
+
+impl StatsExporter {
+    /// Return a new StatsExporter
+    ///
+    /// - `meta` is used when sending the ClientStatsPayload to the agent
+    /// - `cfg` is the configuration for the stats exporter
+    ///
+    /// Returns a result to have the same signature as the blocking implementation.
```
bantonsson, line 56: I don't understand this comment. It returns a `StatsExporter` and not a `result`, or am I missing something?

data-pipeline/src/stats_exporter.rs
```diff
@@ -0,0 +1,362 @@
...
+            meta,
+            sequence_id: AtomicU64::new(0),
+            client: hyper::Client::builder().build(connector::Connector::default()),
+            cancellation_token,
+        }
+    }
+
+    /// Flush the stats stored in the concentrator and send them
+    ///
+    /// If the stats flushed from the concentrator, contain at least one time bucket, the stats are
```
bantonsson, line 77:
```suggestion
    /// If the stats flushed from the concentrator contain at least one time bucket the stats are
```

data-pipeline/src/stats_exporter.rs
VianneyRuhlmann: My bad, this doc is outdated since I removed the blocking call. I updated it.

data-pipeline/src/trace_exporter.rs
```diff
@@ -191,10 +276,27 @@ impl TraceExporter {
             })
     }
 
+    /// Add all spans from the given iterator into the stats concentrator
+    fn add_spans_to_stats<'a>(&self, spans: impl Iterator<Item = &'a pb::Span>) {
+        // TODO: How do we want to react if we have an error i.e. another thread panicked with
```
ajgajg1134, line 281: hmm good question, under what situation can a thread panic with the lock? ie. if it somehow fails to serialize a payload? or an OS level failure to lock?

data-pipeline/src/trace_exporter.rs
```diff
@@ -310,23 +457,110 @@ impl TraceExporterBuilder {
...
+    /// Set the header indicating the tracer has already computed stats.
+    /// This should not be used when stats computation is enabled.
+    pub fn set_client_computed_stats(mut self) -> Self {
+        self.client_computed_stats = true;
+        self
+    }
+
+    /// Enable stats computation on traces sent through this exporter
+    pub fn enable_stats(mut self, bucket_size: time::Duration) -> Self {
+        self.stats_bucket_size = Some(bucket_size);
```
ajgajg1134, line 475: Ah so if we have `Some` bucket_size that's how we know stats is enabled?

data-pipeline/src/trace_exporter.rs
```diff
@@ -310,23 +457,110 @@ impl TraceExporterBuilder {
...
+                        env: self.env,
+                        version: self.version,
+                        lang: self.language.clone(),
+                        tracer_version: self.tracer_version.clone(),
+                        runtime_id: uuid::Uuid::new_v4().to_string(),
+                        service: self.service,
+                        ..Default::default()
+                    },
+                    Endpoint::from_url(stats_exporter::stats_url_from_agent_url(
+                        self.url.as_deref().unwrap_or("http://127.0.0.1:8126"),
```
ajgajg1134, line 528: it seems this string is duplicated in at least one other place, should we make a constant for it? (or is that hard to do because of the borrow checker? :( )

ajgajg1134 (commented):
Overall this looks good to me just a few small comments!

Is there anyway to run this with a real test application / real traces+stats and make sure they show up in the datadog UI? The unit tests look good but it's always nice to have a smoke test or equivalent

data-pipeline/src/trace_exporter.rs
VianneyRuhlmann: Serialization errors should not cause panic imo. It would probably be something critical and since the data can be left in an undefined state it seems reasonable to me to propagate the panic to the other threads of the `trace_exporter`

data-pipeline/src/trace_exporter.rs
VianneyRuhlmann: Yes that's it. I can add a comment to make it explicit

data-pipeline/src/trace_exporter.rs
```diff
@@ -113,20 +179,39 @@ pub struct TraceExporter {
...
+    /// Send msgpack serialized traces to the agent
     pub fn send(&self, data: &[u8], trace_count: usize) -> Result<String, String> {
         match self.input_format {
             TraceExporterInputFormat::Proxy => self.send_proxy(data, trace_count),
             TraceExporterInputFormat::V04 => self.send_deser_ser(data),
         }
     }
 
+    /// Safely shutdown the TraceExporter and all related tasks
+    pub fn shutdown(self) -> Result<String, String> {
```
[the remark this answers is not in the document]
VianneyRuhlmann: It is indeed future proofing, at one point we talked about adding some timeout to it which could return an error

data-pipeline/src/trace_exporter.rs
ajgajg1134: Most of the tracers define a default for the agent URL so it seems reasonable to me here? (Although I think _most_ default to UDS if the socket exists before trying HTTP?)

data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -20,6 +20,10 @@ use std::{ffi::c_char, ptr::NonNull};
 /// * `language` - The language of the client library.
 /// * `language_version` - The version of the language of the client library.
 /// * `language_interpreter` - The interpreter of the language of the client library.
+/// * `hostname` - The hostname of the application, used for stats aggregation
```
ekump, line 23: non-blocking as we can address this later with the planned data-pipeline refactor efforts - but can we consider introducing encapsulating some of these options in objects so every single one isn't a top level function param like it is now? For example there could be a stats struct that encapsulates things like hostname and service?

data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -35,25 +39,38 @@ pub unsafe extern "C" fn ddog_trace_exporter_new(
...
+        .set_service(service.to_utf8_lossy().as_ref())
         .set_input_format(input_format)
         .set_output_format(output_format)
-        .set_response_callback(Box::new(callback_wrapper))
-        .build()
-        .unwrap();
+        .set_response_callback(Box::new(callback_wrapper));
+    if compute_stats {
+        builder = builder.enable_stats(Duration::from_secs(10))
+        // TODO: Enable peer tags aggregation and stats by span_kind based on agent
```
ekump, line 70: non-blocking nit: I prefer putting the ticket ID right after the TODO.

data-pipeline/examples/send-traces-with-stats.rs
```diff
@@ -0,0 +1,63 @@
+use std::{
+    collections::HashMap,
+    time::{Duration, SystemTime, UNIX_EPOCH},
+};
+
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
```
ekump, line 6: The header should be at the top of the file

data-pipeline/src/stats_exporter.rs
```diff
@@ -0,0 +1,360 @@
...
+        env: meta.env,
+        lang: meta.lang,
+        version: meta.version,
+        runtime_id: meta.runtime_id,
+        tracer_version: meta.tracer_version,
+        service: meta.service,
+        container_id: meta.container_id,
+        git_commit_sha: meta.git_commit_sha,
+        tags: meta.tags.into_iter().map(|t| t.to_string()).collect(),
+
```
ekump, line 176: minor: You probably don't want these extra newlines?

data-pipeline-ffi/src/trace_exporter.rs
VianneyRuhlmann: That's a good idea although I'm not sure it would work well with a builder pattern since adding new fields to the struct would break existing implementations. I added a comment about it on the API design doc.

PR #586, opened 2024-08-13
Add build system infrastructure.

hoolioh:
# What does this PR do?

Adds a new build system as described in RFC 0002.

# How to test the change?

The builder crate has been excluded from the default members in the workspace so it's not built when 'build' command is issued with no specific package.

```
export LIBDD_OUTPUT_FOLDER=<path_to_release>
cargo build -p builder--features profiling,data-pipeline,symbolizer,crashtracker --release 
```


docs/RFCs/0002-build-process.md
```diff
@@ -0,0 +1,86 @@
+# RFC 0002: Building process proposal
```
r1viollet, line 1: great doc ! thanks

docs/RFCs/0002-build-process.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Context
+As described in the [RFC](https://github.com/DataDog/libdatadog/blob/main/docs/RFCs/delivering-different-features.md)
```
r1viollet, line 4: minor, link was updated

.github/workflows/test.yml
```diff
@@ -128,13 +134,10 @@ jobs:
...
+        env:
+          RUSTFLAGS: "${{ matrix.flags }}"
         run: |
-           ./build-profiling-ffi.sh -f data-pipeline-ffi ${OUTPUT_FOLDER}/profiling
-
-      - name: "Generate Telemetry FFI"
-        shell: bash
-        run: |
-           ./build-telemetry-ffi.sh ${OUTPUT_FOLDER}/telemetry
+           cargo build -p builder --features profiling,telemetry,data-pipeline,symbolizer,crashtracker --release -vv
```
pawelchcki, line 140: <3

Cargo.toml
```diff
@@ -3,6 +3,38 @@
...
+  "spawn_worker",
+  "tests/spawn_from_lib",
+  "serverless",
+  "bin_tests",
+  "data-pipeline",
+  "data-pipeline-ffi",
+  "ddsketch",
+]
+
+default-members = [
```
pawelchcki, line 37: :exploding_head:

.github/workflows/test.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         if: success() || failure() 
         uses: mikepenz/action-junit-report@v4
         with:
-          report_paths: "target/nextest/ci/junit.xml"
+          report_paths: "target/nextestgci/junit.xml"
```
ekump, line 72: Just curious why this changed?

builder/src/bin/dummy.rs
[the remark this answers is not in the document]
ekump: +1, what is the purpose of this?

ekump (approved):
Not my area of expertise, but LGTM. Nice work!

builder/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "builder"
+build = "build/main.rs"
+rust-version.workspace = true
+edition.workspace = true
+version.workspace = true
+license.workspace = true
+
+[features]
+default = ["profiling"]
```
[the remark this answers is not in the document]
hoolioh: Not really. Since telemetry and data-pipeline symbols are reexported through profiling I assumed that prof was kind of the default feature. Should we leave it out?

builder/src/bin/dummy.rs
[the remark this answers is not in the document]
hoolioh: That's the result of abusing from cargo. Cargo needs one of the following elements in the crate manifest: src/lib.rs, src/main.rs or [[bin]]. Since it wasn't a library nor a binary I chose to add a [[bin]] section just in case it could come in handy execute post-actions after the building. However turns out that for [[bin]] sections is mandatory to have a name and it can't be empty so I named it that way because it does nothing. Anyway maybe it could be named post-action (or something similar) although it currently does nothing. WDYT?

.github/workflows/test.yml
hoolioh: Good catch, probably some ill replace command. I'll revert that.

build-profiling-ffi.sh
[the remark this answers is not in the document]
hoolioh: Yes at first I wanted to also make the proper changes in libddprof-build but I think I should bring it back and, when the build system is stable enough, do the changes in the build repo and delete them.

builder/build/main.rs
```diff
@@ -0,0 +1,207 @@
...
+        ar.append_dir_all("lib", self.target_lib.as_ref())?;
+        ar.append_dir("bin", self.target_bin.as_ref())?;
+        ar.append_dir_all("include", self.target_include.as_ref())?;
+
+        Ok(ar.finish().expect("Failed to write the tarfile"))
+    }
+}
+
+fn main() {
+    let mut path = env::var("OUT_DIR").unwrap();
```
r1viollet, line 152: NIT: should we add defaults and logs on the output behaviour?

builder/build/module.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
+
+pub trait Module {
```
VianneyRuhlmann, line 6: I think it would be useful to have some documentation on this trait, and on how to add a new crate to the build system. Also would it make sense to define a `new` method so that modules have the same interface, or would that be too rigid ?

builder/build/main.rs
```diff
@@ -0,0 +1,207 @@
...
+    if let Ok(libdd_path) = env::var("LIBDD_OUTPUT_FOLDER") {
+        path = libdd_path;
+    }
+
+    let profile = env::var("PROFILE").unwrap();
```
r1viollet, line 157: I had a determine path function in the build-common parts to try and have a common way of defining where we should output files.
What do you think of the logic that is available there ?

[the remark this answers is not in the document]
hoolioh: Good idea. I will add it.

builder/build/module.rs
hoolioh: Agree on the documentation. Could you elaborate a little bit about the reasoning on adding a new method?

builder/build/main.rs
```diff
@@ -0,0 +1,255 @@
...
+        Ok(ar.finish().expect("Failed to write the tarfile"))
+    }
+}
+
+fn main() {
+    // Rerun build script if any of the env vars change.
+    println!("cargo:rerun-if-env-changed=LIBDD_OUTPUT_FOLDER");
+    println!("cargo:rerun-if-env-changed=PROFILE");
+
+    let (_, source_path) = determine_paths();
```
hoolioh, line 199: @r1viollet WDYT?

r1viollet: looks good, thanks!

Cargo.toml
bantonsson, line 37: So this makes `cargo tree` report a number of dependencies as _separate_ instead of the same, ie. `bytes v1.6.0` instead of `bytes v1.6.0 (*)`. I truly don't understand how `cargo` works, and what this setting means for the final artifacts, but have we checked if this affects the sizes?

bantonsson: While looking at the benchmark time changes, I also noticed that this `default-members` thing has other side effects like not running `clippy` on `builder` from the root directory unless you add `--workspace`. What is the real reason for having this?

builder/Cargo.toml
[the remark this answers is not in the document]
bantonsson: This `default` feature for builder is the thing that makes the benchmark times change. Don't ask me to explain how.

Cargo.toml
hoolioh: The reason for not including the builder in the default-members was to save time and resources in scenarios that don't need to build the artifacts, like binding projects or test scenarios.

bantonsson: Removing `--workspace` made no difference to the run-time of the benchmark task. It took 9 minutes 53 seconds, and the last one with `--workspace` took 9 minutes and 50 seconds.

builder/build/module.rs
VianneyRuhlmann: This would make `main.rs` less module specific since adding a module would always be the same 
```
builder.add_module(Box::new(MyModule::new(...));
```
But that would make it more rigid having to pass useless params for some Module and making it more complicated to add new params. So probably not worth it.

builder/build/main.rs
hoolioh: I'm not sure if I'm following you there. Here `OUT_DIR` (which is always exported by cargo) act as the default directory in absence of `LIBDD_OUTPUT_FOLDER`. 
What do you mean by adding logs? Using `cargo::warning`?

builder/Cargo.toml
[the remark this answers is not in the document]
hoolioh: @morrisonlevi any more thoughts on this?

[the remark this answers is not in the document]
morrisonlevi: As a contributing member to the profiling subtree, I am happy to not ship it as default. We don't need special treatment, as far as I am aware.

builder/build/main.rs
r1viollet: Maybe a small doc is what I was looking for. 
When I initially tried running the build, I was unsure of the different knobs I should be trying out.

tools/docker/Dockerfile.build
```diff
@@ -154,12 +156,12 @@ COPY --from=ffi_build_platform_agnostic_cache /root/.cargo /root/.cargo/
...
 # cache debug dependency build
-RUN cargo build --lib --all
+RUN cargo build --lib --workspace --exclude builder
 # cache release dependency build
-RUN cargo build --release --lib --all
+RUN cargo build --release --lib --workspace --exclude builder
 
 COPY ./ ./
-RUN ./build-profiling-ffi.sh /build/output
+RUN LIBDD_OUTPUT_FOLDER=/build/output cargo build -p builder --features profiling,telemetry,data-pipeline,symbolizer,crashtracker --release
```
r1viollet, line 164: This is because we need to have everything built.
NIT, in the future we could introduce an `all` target.

r1viollet:
Happy to merge this when you think we have reached parity with today's build. This is already a leap forward! Some of these can be addressed as follow ups.

builder/build/main.rs
hoolioh: Oh, you're right I include some documentation in the code but totally forgot to include in the README file.

tools/docker/Dockerfile.build
hoolioh: Godd idea. There will be a second stage for the builder so I think it will be a good moment to add this.

hoolioh:
> Happy to merge this when you think we have reached parity with today's build. This is already a leap forward! Some of these can be addressed as follow ups.

@r1viollet during my testing the binaries generated are on par with what we're building now with the scripts. Also the examples are running fine as well as the tests in all the platforms we're currently using. Originally I wanted to modify the gitlab pipeline to use the builder crate but there were some problems with the Windows build. I talked with Gregory about that and we decided that we will make the changes in another separate PR, that's why the build scripts are still in place. Therefore if you're okay with the latest changes I think we can move forward, merge this one and start working on the next iteration.

bantonsson (approved):
Even the benchmarks approve 😉

r1viollet:
Ok, let's go! Please give a heads up to everyone when merging and we will look out for any potential issues.

PR #587, opened 2024-08-15
Add NoAllocString for decoding v04 traces

ekump:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/trace-utils/src/no_alloc_string.rs
+++ b/trace-utils/src/no_alloc_string.rs
@@ -1,0 +1,62 @@
+use bytes::Bytes;
+use serde::ser::{Serialize, Serializer};
+use std::borrow::Borrow;
+
+pub struct BufferWrapper {
+    buffer: Bytes,
+}
+
+impl BufferWrapper {
+    pub fn new(buffer: Bytes) -> Self {
+        BufferWrapper { buffer }
+    }
+
+    pub fn create_no_alloc_string(&self, slice: &[u8]) -> NoAllocString {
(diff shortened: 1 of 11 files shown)


trace-utils/src/no_alloc_string.rs
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+    // Creates a NoAllocString from a Bytes instance (does not copy the data)
+    pub fn from_bytes(bytes: Bytes) -> NoAllocString {
+        NoAllocString { bytes }
+    }
+    // TODO: Perhaps this should return the error instead of panicking
+    pub fn as_str(&self) -> &str {
+        std::str::from_utf8(&self.bytes).expect("Invalid UTF-8")
+    }
```
paullegranddc, line 47: I think we'd have better error locality (and an easier time handling errors) if we asserted the bytes are utf-8 on `NoAllocString` construction, rather than on serialization.
This would allow to return an error on Span construction instead of panicking latter.

Also, we'd pay the price of utf8 validation only once rather than every time we want to get a &str which might be more than once.

bantonsson: Looking through the code, the validation is actually done by `read_string_ref` that is used to get the `&str` that is then converted to `&[u8]` or `Bytes`, and then put in the `NoAllocString`. Let's just change the APIs and functions to accept `&str` instead and then we don't have to validate multiple times.

PR #588, opened 2024-08-15
improve crashtracker config deserialization

rochdev:
# What does this PR do?

Improve crashtracker config deserialization:

- A string can now be passed as an endpoint URL when deserializing instead of having to know about Hyper.
- Fields with a default value are no longer required.

# Motivation

When using NAPI-RS, JavaScript types are converted directly to Rust types through serialization/deserialization. This meant that a lot of unneeded options were required to be passed, and the URL had to be passed with the internal structure as well.

# Additional Notes

# How to test the change?

Good question 😅 I'm not familiar with tests in libdatadog, and I don't know if serialization/deserialization is even tested at all right now.

### DIFF
--- a/crashtracker/src/crash_info/metadata.rs
+++ b/crashtracker/src/crash_info/metadata.rs
@@ -3,7 +3,7 @@
 use ddcommon::tag::Tag;
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
 pub struct CrashtrackerMetadata {
     pub library_name: String,
     pub library_version: String,
--- a/crashtracker/src/shared/configuration.rs
+++ b/crashtracker/src/shared/configuration.rs
@@ -12,7 +12,8 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum StacktraceCollection {
     /// Stacktrace collection occurs in the
-    #[default] Disabled,
+    #[default]
+    Disabled,
     WithoutSymbols,
     EnabledWithInprocessSymbols,
     EnabledWithSymbolsInReceiver,
--- a/crashtracker/src/shared/configuration.rs
+++ b/crashtracker/src/shared/configuration.rs
@@ -9,16 +9,17 @@ use serde::{Deserialize, Serialize};
 /// We recommend fully enabling stacktrace collection, but having an environment
 /// variable to allow downgrading the collector.
 #[repr(C)]
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum StacktraceCollection {
     /// Stacktrace collection occurs in the
-    Disabled,
+    #[default] Disabled,
     WithoutSymbols,
     EnabledWithInprocessSymbols,
     EnabledWithSymbolsInReceiver,
 }
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(default)]
 pub struct CrashtrackerConfiguration {
     // Paths to any additional files to track, if any
     pub additional_files: Vec<String>,
@@ -28,7 +29,8 @@ pub struct CrashtrackerConfiguration {
     pub wait_for_receiver: bool,
 }
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(default)]
 pub struct CrashtrackerReceiverConfig {
     pub args: Vec<String>,
     pub env: Vec<(String, String)>,
--- a/ddcommon/src/lib.rs
+++ b/ddcommon/src/lib.rs
@@ -66,7 +66,7 @@ struct SerializedUri<'a> {
 #[serde(untagged)]
 enum StringOrSerializedUri<'a> {
     String(String),
-    SerializedUri(SerializedUri<'a>)
+    SerializedUri(SerializedUri<'a>),
 }
 
 fn serialize_uri<S>(uri: &hyper::Uri, serializer: S) -> Result<S::Ok, S::Error>
--- a/ddcommon/src/lib.rs
+++ b/ddcommon/src/lib.rs
@@ -62,6 +62,13 @@ struct SerializedUri<'a> {
     path_and_query: Option<Cow<'a, str>>,
 }
 
+#[derive(Serialize, Deserialize)]
+#[serde(untagged)]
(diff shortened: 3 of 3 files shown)


crashtracker/src/crash_info/metadata.rs
```diff
@@ -3,7 +3,7 @@
 use ddcommon::tag::Tag;
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
 pub struct CrashtrackerMetadata {
```
r1viollet, line 7: I do not think we would want this to have empty defaults. Any reason to allow this ?

crashtracker/src/shared/configuration.rs
```diff
@@ -9,16 +9,18 @@ use serde::{Deserialize, Serialize};
 /// We recommend fully enabling stacktrace collection, but having an environment
 /// variable to allow downgrading the collector.
 #[repr(C)]
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum StacktraceCollection {
     /// Stacktrace collection occurs in the
+    #[default]
```
r1viollet, line 15: Is without symbols a reasonable default? Or would users want this disabled?

r1viollet:
Most of the testing would be through unit tests. Though there are some end to end FFI tests that could be leveraged for crashtracking.

crashtracker/src/crash_info/metadata.rs
rochdev: Why not if there are already defaults outside of Serde and they work 🤔

crashtracker/src/shared/configuration.rs
rochdev: I don't really know what this option does to begin with, but it seemed reasonable to have it disabled as a default. A different value might make more sense.

rochdev:
> Most of the testing would be through unit tests. Though there are some end to end FFI tests that could be leveraged for crashtracking.

It doesn't look like there are existing tests for crashtracking to begin with. And FFI tests are something else completely as Node cannot use FFI.

crashtracker/src/shared/configuration.rs
r1viollet: Ok, I would have assumed WithoutSymbols would be the expected default, but this is not my component.

crashtracker/src/crash_info/metadata.rs
r1viollet: What I meant is that you would want the user to fill some of these fields (like family). Feel free to ignore I'm not sure how you would make that field mandatory.

ddcommon/src/lib.rs
```diff
@@ -82,19 +89,22 @@ fn deserialize_uri<'de, D>(deserializer: D) -> Result<hyper::Uri, D::Error>
...
-    if let Some(v) = uri.authority {
-        builder = builder.authority(v.deref());
-    }
-    if let Some(v) = uri.scheme {
-        builder = builder.scheme(v.deref());
-    }
-    if let Some(v) = uri.path_and_query {
-        builder = builder.path_and_query(v.deref());
+    match StringOrSerializedUri::deserialize(deserializer)? {
+        StringOrSerializedUri::String(str) => str.parse().map_err(Error::custom),
```
r1viollet, line 93: are there ways to add a message here to differentiate between the errors ?

r1viollet:
> It doesn't look like there are existing tests for crashtracking to begin with. And FFI tests are something else completely as Node cannot use FFI.

If adding a rust unit test is difficult, ping me, I can approve to unblock. I agree the ffi tests won't help.
If you are able to add a test just to cover the new parts, that is great.

crashtracker/src/crash_info/metadata.rs
rochdev: That's the main thing. I think a configuration object struct should accept basically anything, and it should be up to the implementation using it to validate if the fields have values that are valid for whatever class using the object. In that sense, I think having a default value for every field makes sense, and simplifies usage by language-specific bindings.

crashtracker/src/shared/configuration.rs
rochdev: It's also not my component and I don't know what default would be the most reasonable, I just went with the one with value `0` as it seemed like it would have been intended to be the default but that could be wrong.

rochdev:
> If adding a rust unit test is difficult, ping me, I can approve to unblock. I agree the ffi tests won't help.

Is there an example anywhere in the project where serialization/deserialization is tested that I can look at?

ddcommon/src/lib.rs
rochdev: I think the original error message will be properly remapped onto the new custom error, or at least that is my understanding of what this is doing"

crashtracker/src/shared/configuration.rs
r1viollet: This is fine, it is a safe default. We can get this logic reviewed by @danielsn when he is back

crashtracker/src/crash_info/metadata.rs
paullegranddc: I disagree, in this case the default value is not valid and we should make it invalid by construction rather than validation.
For instance `family` really should be an enum, although right now it's more practical for it to be a String.
People should think about what they want in their metadata and deriving Default automatically goes against that

ddcommon/src/lib.rs
```diff
@@ -62,6 +62,13 @@ struct SerializedUri<'a> {
     path_and_query: Option<Cow<'a, str>>,
 }
 
+#[derive(Serialize, Deserialize)]
+#[serde(untagged)]
+enum StringOrSerializedUri<'a> {
+    String(String),
+    SerializedUri(SerializedUri<'a>),
+}
```
paullegranddc, line 70: There is a reason why this is serialized like that and not just as a String (which it used to in the past).

The reason is that non http uris (unix sockets unix:// , windows pipes, or file uris) are not correctly deserialized by the hyper::Uri from string implementation.

Could you add unit tests to check what happens when you try to deserialize non http Uris?
I'm interested in particular in paths starting with `/`, `./` and containing spaces as they have caused problems

crashtracker/src/crash_info/metadata.rs
rochdev: I was replying basically for all 3 objects, but for `CrashtrackerMetadata` I tend to agree that most fields are mandatory, I added `Default` basically because I was doing it also for the other ones. I'll change that one so that only tags are optional.

ddcommon/src/lib.rs
rochdev: Because this is passed to Hyper directly, I assumed it was just a shortcut to serialize `Url` which is not serializable by default. If there is a good reason to pass a more complex object, I'm fine with also passing that structure instead of relying on a string and removing this change. However, I tend to prefer handling this kind of common use cases internally behind the API instead of externally, thus exposing Hyper limitations to the caller.

rochdev: Actually, how do I even get the tests to pass? Whatever I do it seems there are a lot of tests that are failing before I even touch anything, and I'm unable to make them pass.

PR #590, opened 2024-08-20
Add a tiny bytes::Bytes like implementation

bantonsson:
# What does this PR do?

Adds a tiny `bytes::Bytes` like implementation for having multiple byte slices pointing into the same slice.

# Motivation

The implementation in `bytes::Bytes` does not allow us to use the mapped memory types in the `ipc` crate, and implementing support for such types in the original crate is a not an option, since it is non trivial and there is a large amount of functionality that doesn't map well to the mapped memory.

# Additional Notes

# How to test the change?

There are tests.


tinybytes/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+
+/// Since the Bytes type is immutable, and UnderlyingBytes is `Send + Sync``, it is safe to share
+/// `Bytes` across threads.
+unsafe impl Send for Bytes {}
+unsafe impl Sync for Bytes {}
+
+impl Bytes {
+    /// Creates empty `Bytes`.
+    #[inline]
+    pub fn new() -> Self {
```
bwoebi, line 31: Maybe call it `empty()`?

bantonsson: Indeed, `empty` it is...

tinybytes/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+}
+
+impl Ord for Bytes {
+    fn cmp(&self, other: &Bytes) -> cmp::Ordering {
+        self.as_slice().cmp(other.as_slice())
+    }
+}
+
+impl hash::Hash for Bytes {
+    // TODO should we cache the hash since we know the bytes are immutable?
```
bwoebi, line 223: I suppose we can do that. I suppose the overhead in size is minimal a couple bytes, which are likely anyway within padding (the struct is 32 bytes, and system allocator will give you 16 byte aligned memory...)

tinybytes/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /// Returns a slice of self that is equivalent to the given `subset`, if it is a subset.
+    ///
+    /// When processing a `Bytes` buffer with other tools, one often gets a
+    /// `&[u8]` which is in fact a slice of the `Bytes`, i.e. a subset of it.
+    /// This function turns that `&[u8]` into another `Bytes`, as if one had
+    /// called `self.slice()` with the range that corresponds to `subset`.
+    ///
+    /// This operation is `O(1)`.
```
ekump, line 117:
```suggestion
    /// This operation is `O(1)`.
    ///
    /// # Examples
    ///
    /// ```
    /// use tinybytes::Bytes;
    ///
    /// let bytes = Bytes::copy_from_slice(b"hello world");
    /// let subset = &bytes.as_ref()[0..5];
    /// let slice = bytes.slice_ref(subset).unwrap();
    /// assert_eq!(slice.as_ref(), b"hello");
    ///
    /// let subset = &bytes.as_ref()[6..11];
    /// let slice = bytes.slice_ref(subset).unwrap();
    /// assert_eq!(slice.as_ref(), b"world");
    ///
    /// let invalid_subset = b"invalid";
    /// assert!(bytes.slice_ref(invalid_subset).is_none());
    /// ```    
```

tinybytes/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    /// Returns a slice of self for the provided range.
+    ///
+    /// This will return a new `Bytes` handle set to the slice, and will not copy the underlying
+    /// data.
+    ///
+    /// This operation is `O(1)`.
+    ///
+    /// # Panics
+    ///
+    /// Slicing will panic if the range does not conform to  `start <= end` and `end <= self.len()`.
```
ekump, line 72:
```suggestion
    /// Slicing will panic if the range does not conform to  `start <= end` and `end <= self.len()`.
    ///
    /// # Examples
    ///
    /// ```
    /// use tinybytes::Bytes;
    ///
    /// let bytes = Bytes::copy_from_slice(b"hello world");
    /// let slice = bytes.slice(0..5);
    /// assert_eq!(slice.as_ref(), b"hello");
    ///
    /// let slice = bytes.slice(6..11);
    /// assert_eq!(slice.as_ref(), b"world");
    /// ```    
```

ekump (approved):
A couple non-blocking comments but LGTM!

tinybytes/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "tinybytes"
```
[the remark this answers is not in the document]
bantonsson: I honestly think that the `datadog-` prefixing in this project is weird, and that not having the same name on the directory and the package should be illegal.

tinybytes/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub fn slice(&self, range: impl RangeBounds<usize>) -> Self {
```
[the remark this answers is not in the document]
bantonsson: I think it is ok to panic if you do an out of bounds slice. It's what slice of `str` and other types do.

PR #598, opened 2024-08-26
reduced allocation v04 span representation

ekump:
# What does this PR do?

Deserialize V.04 Spans into a new internal representation that attempts to minimize new String allocations. Using the existing protobuf definitions leads to a level of performance that tracer libraries are/will be uncomfortable with. 

Introduce a `BytesString` that borrows from a tinybytes::Bytes buffer slice and a `BufferWrapper` for keeping track of where in the bytes buffer we are slicing from.

This will impact only v04 Spans at the moment. 

# Motivation

The PHP Language Platform team observed an unacceptable level of increased memory footprint when testing the use of the Sidecar for sending traces to the agent. 

# Additional Notes

The use of non-static `&str` or `Cow` types was not possible as once spans are decoded they may need to be passed to async code paths in `SendData` on the way to the agent and this caused several lifetime issues. The use of `tinybytes` is necessary as `Bytes` doesn't support `&[u8]` input, which is necessary when we can't own the data (i.e. shared memory from the tracer using IPC).

# How to test the change?

Unit and integration tests have been updated.


trace-utils/src/no_alloc_string.rs
```diff
@@ -0,0 +1,149 @@
...
+        })
+    }
+    // Creates a NoAllocString from a Bytes instance (does not copy the data)
+    pub fn from_bytes(bytes: tinybytes::Bytes) -> Result<NoAllocString, std::str::Utf8Error> {
+        std::str::from_utf8(&bytes)?;
+        Ok(NoAllocString { bytes })
+    }
+
+    /// Creates a NoAllocString from a Bytes instance without validating the bytes are UTF-8.
+    pub fn from_bytes_unchecked(bytes: tinybytes::Bytes) -> NoAllocString {
```
bwoebi, line 58: The proper way would be probably to mark this unsafe and assert or assume safety on caller sites.

trace-utils/src/msgpack_decoder/v04/decoder/mod.rs
```diff
@@ -99,54 +105,81 @@ fn read_string_ref(buf: &[u8]) -> Result<(&str, &[u8]), DecodeError> {
...
+    let mut map = HashMap::with_capacity(len.try_into().expect("Unable to cast map len to usize"));
+    for _ in 0..len {
+        let (key, next) = read_string_ref(buf)?;
+        *buf = next;
 
-    Ok((k, v))
+        let (val, next) = read_string_ref(buf)?;
+        *buf = next;
+
+        let key = buf_wrapper.create_no_alloc_string_unchecked(key.as_bytes());
```
VianneyRuhlmann, line 125: I think it would be better to have a function for calling `read_string_ref` and converting the slice with `unchecked` this way we have only one function dealing with potentially unsafe code.

trace-utils/src/msgpack_decoder/v04/decoder/span.rs
```diff
@@ -26,103 +26,77 @@ use std::str::FromStr;
 /// This function will return an error if:
 /// - The map length cannot be read.
 /// - Any key or value cannot be decoded.
-#[inline]
-pub(crate) fn decode_span(buf: &mut &[u8]) -> Result<Span, DecodeError> {
+pub fn decode_span<'a>(
+    buffer: &'a tinybytes::Bytes,
```
VianneyRuhlmann, line 30: Can't we replace `buf` with `buffer`completely ? And have all the function decoding msgpack return slice of `buffer` which would also be Bytes.

trace-utils/src/no_alloc_string.rs
```diff
@@ -0,0 +1,149 @@
+use serde::ser::{Serialize, Serializer};
+use std::borrow::Borrow;
+use tinybytes;
+
+pub struct BufferWrapper {
+    buffer: tinybytes::Bytes,
+}
+
+impl BufferWrapper {
```
VianneyRuhlmann, line 9: I'm not sure I understand the advantages of BufferWrapper compared to calling directly the NoAllocString methods with a buffer.

bwoebi:
I wonder whether we should put `NoAllocString` (Maybe call it `BytesString`?) into tinybytes crate directly. Seems like an useful wrapper for usage with tinybytes in general.

tinybytes/src/bytes_string.rs
```diff
@@ -0,0 +1,259 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `bytes` - A `tinybytes::Bytes` instance that will be converted into a `BytesString`.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it assumes the bytes are valid UTF-8. If the bytes are not
+    /// valid UTF-8, the behavior is undefined.
+    pub fn from_bytes_unchecked(bytes: Bytes) -> BytesString {
```
bwoebi, line 146: It's confusing here. Somewhat. We probably should either have unchecked on input _or_ on output. But currently from_bytes() checks utf8 _and_ as_str() checks it again.

My proposal would be:
- `fn from_bytes(bytes: Bytes) -> Result<BytesString, std::str::Utf8Error>`
- `unsafe fn from_bytes_unchecked(bytes: Bytes) -> BytesString`
- `fn as_str() -> &str`

And we drop `as_str_unchecked()`.

ekump: Why not offer both and let the user decide where / if any checks occur?

bwoebi: You technically _could_ do that in an ideal world, but you're also going to run the risk of double-checking (bad for perf) or not checking at all because each side assumes the other side does it (violating invariants).

bwoebi: Also, if it's not yet checked, you probably are better off moving bare `Bytes` around, rather than `ByteString`.
The intrinsic value of `ByteString` really is "This `Bytes` instance is valid UTF-8".

ekump: But for our only current use-case we don't want to check at all. [The bytes are already checked to be valid utf-8 by rmp](https://github.com/3Hren/msgpack-rust/blob/master/rmp/src/decode/str.rs#L176) when decoding msgpack. So at a minimum we need to expose `_unchecked` functions on both sides.

Ordinarily, I'd agree with you that we should try to limit potential foot guns via API design. I think it's ok in this instance since the use of tinybytes is an intentional decision to add complexity in favor of performance and all the `_unchecked` functions are marked `unsafe` so the caller should be cognizant of the risks. The documentation will be pretty good (IMO), and benchmarks and testing should mitigate most risk.

trace-utils/src/msgpack_decoder/v04/decoder/span.rs
ekump: The `rmp` library code expects &mut [u8], so we need to either pass it around or do a conversion every time from tinybytes to slice. I can include the underlying &mut [u8] in the `BufferWrapper` struct to make it more ergonomic.

tinybytes/src/bytes_string.rs
bwoebi: I mean, you need to only provide an unchecked on the input side. And on the output side you only provide `as_str()` which  _itself_ calls the unchecked function, but the `as_str()` function does not need to be marked unsafe. The contract of the BytesString struct is guaranteeing that any BytesString is valid UTF-8. So you just can safely use `unsafe { std::str::from_utf8_unchecked(&self.bytes) }` then.

tinybytes/src/bytes_string.rs
```diff
@@ -0,0 +1,263 @@
...
+use crate::Bytes;
+#[cfg(all(feature = "bytes_string", feature = "serde"))]
+use serde::ser::{Serialize, Serializer};
+use std::borrow::Borrow;
+use std::str::Utf8Error;
+
+#[cfg(feature = "bytes_string")]
+pub struct BufferWrapper<'a, 'b> {
+    buffer: Bytes,
+    pub underlying: &'a mut &'b [u8],
```
bwoebi, line 13: Why do we need this - instead of a simple `fn as_slice(&self) -> &[u8]`? and possibly `unsafe fn as_mut_slice(&self) -> &mut [u8]`, forwarding to the Bytes instance?

bwoebi:
@ekump (and reviewers) Please consider https://github.com/DataDog/libdatadog/commit/386a6844d571eacd37180a1f567b404406a34017 first. It is quite a refactor on the ByteWrapper approach, which hopefully increases soundness and is probably also a bit more straightforward.
It implements what I've been commenting before (albeit at a larger scale than I initially anticipated).

To view the full diff of this PR plus my commit, please look at https://github.com/DataDog/libdatadog/compare/bob/refactor-bytes-wrapper-away.

tinybytes/src/bytes_string.rs
bantonsson: I'm not sure how that would work @bwoebi. The `underlying` is really `position` of the  message pack decoding, and the `buffer` should not be modified. Maybe the names should change to reflect this in some way? (Naming is hard™)

tinybytes/src/bytes_string.rs
bantonsson: I agree with @bwoebi here. We only check at creation (or avoid it using `unsafe`), and then `as_str` knows that the `Bytes` are _safe_ and use `unsafe { std::str::from_utf8_unchecked(&self.bytes) }`

bantonsson:
There is a reason why `Bytes` is immutable @bwoebi, so it's `Sync` and `Send` and I don't want to introduce a way to mutate it.

bwoebi:
@bantonsson What actually needs to be immutable is the underlying content, which it still is. This commit _just_ allows manipulating the range of the underlying slice. Which in itself doesn't violate Sync/Send constraints (as you still need a `&mut` reference to even call `Bytes.as_mut_slice()`).

bantonsson:
@bwoebi You're indeed right about `Sync` and `Send`, and for decoding MessagePack this way we will only create new slices within the _existing_ slice before we advance it.

bwoebi:
Okay, brought my commit to this branch after talking to Edmund :-)

paullegranddc:
> This commit just allows manipulating the range into the underlying slice

@bwoebi I think there is a better than this to use Bytes to read from rpm serde. You can look at my last comment on the code review.

tinybytes/src/lib.rs
```diff
@@ -51,13 +46,13 @@ impl Bytes {
...
-        self.len
+        self.slice.len()
     }
 
     /// Returns `true` if the `Bytes` is empty.
     #[inline]
     pub const fn is_empty(&self) -> bool {
-        self.len == 0
+        self.slice.is_empty()
     }
```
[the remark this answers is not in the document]
morrisonlevi: Except they are not const through deref, so that would be one reason to keep them (doubt that it helps here, this is merely an educational mention).

tinybytes/src/lib.rs
```diff
@@ -10,8 +10,7 @@ use std::{
 /// Immutable bytes type with zero copy cloning and slicing.
 #[derive(Clone)]
 pub struct Bytes {
-    ptr: *const u8,
-    len: usize,
+    slice: &'static [u8],
```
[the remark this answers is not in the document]
bwoebi: I do prefer (ptr, len), but it doesn't really work with `as_mut_slice()`.

tinybytes/src/lib.rs
```diff
@@ -154,38 +149,49 @@ impl Bytes {
...
+    /// # Safety
+    ///
+    /// Callers of that function must make sure that they only put subslices of the slice into the
+    /// returned reference.
+    /// They also need to make sure to not persist the slice reference for longer than the struct
+    /// lives.
+    #[inline]
+    pub unsafe fn as_mut_slice(&mut self) -> &mut &'static [u8] {
+        &mut self.slice
+    }
```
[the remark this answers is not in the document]
bwoebi: It's not just as simple as that, In particular with the string reading:
```
    read_string_ref_nomut(unsafe { buf.as_mut_slice() }).map(|(str, newbuf)| {
        let string = BytesString::from_bytes_slice(buf, str);
        *unsafe { buf.as_mut_slice() } = newbuf;
        string
    })
```
there's a dance around the ordering of executions: _first_, read the string. Then fetch the string slice. Then adjust the size.
Because `Bytes` asserts that any read from it is within its bounds. So we'd then have to craft an unsafe method for _that_ instead.

Also, your `Read` impl is actually copying every single byte. Passing a slice directly does not incur that overhead. It's maybe a bit more elegant to use `Read`, but actually not as performant. And this PR here is supposed to solve ... performance problems.

bwoebi:
@paullegranddc The main point of using `io::Read` is for reading from an ephemeral buffer which you don't have fully access to. Which has copying overhead.

tinybytes/src/lib.rs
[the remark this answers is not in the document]
bwoebi: I've just been profiling the whole patch. There's currently still one copy happening (the whole data buffer at once) in `SidecarServer::send_trace_v04`. That single copy accounts for 35% of the whole overhead. So copying here is most likely unacceptable.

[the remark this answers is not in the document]
paullegranddc: Yeah for sure we don't want to copy the strings we read, but for that you can make cleaner abstractions than using `read_string_ref_nomut`.


We could very well rewrite it to use the following primitives to encapsulate the unsafe manipulation in the Bytes implementation

```rust
impl Bytes {
  // make the bytes instance point to original_buffer[n..]
  // returns None if the buffer is smaller than n
  fn advance((&mut self, n: usize) -> Option<()> {
    if n > self.len { return None }
    self.ptr += n;
    self.len -= n;
    Ok(())
  }
  
  // make the bytes instance point to original_buffer[..n]
  // returns None if the buffer is smaller than n
  fn shrink((&mut self, n: usize) -> Option<()> {
    if n > self.len { return None }
    self.len = n;
    Ok(())
  }

  // Yield the first n bytes and advance the offset from as much 
  fn split_front_ref(&mut self, n: usize) -> Option<&[u8]> {
    let start = self.ptr;
    self.advance(n)?
    // Safe because the lifetime of the slice is smaller than the Bytes
    // and n smaller than length
    Ok(unsafe { slice:: from_raw_parts(start, n) })
  }
  
  fn split_front(&mut self, n: usize) -> Option<Bytes> {
    if n > self.len { return None }
    let mut front = self.clone();
    self.advance(n)?;
    front.shrink(n)?;
    Ok(front)
  }

}

fn read_string_ref(bytes: &'a mut Bytes) -> Result<&'a str, DecodeError> {
  let len = rpm::decode::read_str_len(bytes)?;
  let str_ref = self.split_front_ref(len).map_err(|| DecodeStringError::BufferSizeTooSmall(len))?;
  str::from_utf8(str_ref).map_err(|_| DecodeStringError::InvalidUtf8())
}

fn read_string_bytes(bytes: &'a mut Bytes) -> Result<&'a str, DecodeError> {
  let len = rpm::decode::read_str_len(bytes)?;
  let str_bytes = self.split_front(len).map_err(|| DecodeStringError::BufferSizeTooSmall(len))?;
  BytesString::from_bytes(str_bytes).map_err(...)
}
```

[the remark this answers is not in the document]
paullegranddc: By the way, I'm curious can your share the profile?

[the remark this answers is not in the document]
bwoebi: I personally feel you're trying to introduce complexity where trivial unsafe is keeping it simple.

Also, fwiw - your read_string_ref does an extra utf8 check, so you'd use the unsafe method to convert to &str with _unchecked.

And we still rely on io::Read - which makes it easy to accidentally use it in contexts it's not supposed to be used (because it would copy).

Alternatively you could use something like:
```
struct RmpBuffer {
    pub bytes: Bytes,
    rmp: decode::Bytes<'static>,
}

impl RmpBuffer {
    pub fn new(mut bytes: Bytes) -> RmpBuffer {
        RmpBuffer {
            rmp: decode::Bytes::new(unsafe { bytes.as_mut_slice() }),
            bytes,
        }
    }
}
```
Basically the fast Bytes abstraction from Rmp itself. But it works on a slice, so you'll have to carry Bytes around to slice from it. But it would reduce the whole thing to a single unsafe {}.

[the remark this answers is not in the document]
bwoebi: And sure, the profile looks like this: 
image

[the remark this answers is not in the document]
paullegranddc: > Also, fwiw - your read_string_ref does an extra utf8 check, so you'd use the unsafe method to convert to &str with _unchecked.

It doesn't. `rpm::decode::read_str_len` doesn't do any validation and just read the length header. So when the  Bytes instance gets sliced, it's necessary to check that the slice you get is a valid string.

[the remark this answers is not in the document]
bwoebi: Oh, my mistake.

hoolioh (approved):
LGTM. Definitely a step forward in direction we want to go. Probably the most relevant thing is the copy in `send_trace_v04` but IMHO I would address that in a separate PR.

tinybytes/src/lib.rs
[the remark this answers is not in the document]
ekump: @bwoebi and @paullegranddc - Is this conversation still ongoing? Any changes either of you require here? I don't have a strong opinion, so happy to implement whatever the group agrees on.

tinybytes/src/lib.rs
```diff
@@ -225,6 +203,9 @@ impl UnderlyingBytes for Vec<u8> {}
 impl UnderlyingBytes for Box<[u8]> {}
 impl UnderlyingBytes for String {}
 
+#[cfg(feature = "ipc")]
+impl UnderlyingBytes for MappedMem<ShmHandle> {}
```
bantonsson, line 207: I think that `ipc` should implement this instead, so `TinyBytes` can continue to be _standalone_.

tinybytes/src/lib.rs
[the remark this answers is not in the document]
bwoebi: Feel free to implement @paullegranddc suggestion with io::Read and the additional methods to Bytes.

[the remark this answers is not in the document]
paullegranddc: On my end, I don't have a strong opinion. The code currently works, so I think we can merge it, but it'd be nice if you have time to refactor it. Either in this PR or latter

tinybytes/src/lib.rs
ekump: That was my original intention, but I ran into some compilation issues initially. Now when I put the impl in IPC it works fine so ¯\_(ツ)_/¯

tinybytes/src/lib.rs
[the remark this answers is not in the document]
ekump: @paullegranddc @bwoebi - I do like the `impl io::Read` solution as it seems to contain less sharp edges, but benchmarks wind up being 20% slower overall for the PR vs 10% faster with the `as_mut_slice` implementation. It's possible I misunderstood the suggestion and implemented something incorrectly here. But if neither of you see any glaring errors I will revert to the `as_mut_slice` implementation for now and create a ticket to follow up on this in another PR.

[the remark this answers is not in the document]
bwoebi: This does not surprise me too much, at least it would need much more optimization by the compiler. So, if it's actually 30% slower, then please revert it.

tinybytes/Cargo.toml
```diff
@@ -14,3 +14,14 @@ once_cell = "1.8"
...
+tinybytes = { path = ".", features = ["bytes_string", "serialization"] }
+
+[dependencies]
+serde = { version = "1.0.209", optional = true }
+#datadog-ipc = { path = "../ipc", optional = true }
+
+[features]
+bytes_string = []
+serialization = ["serde"]
+#ipc = ["datadog-ipc"]
```
morrisonlevi, line 27: Was this meant to be committed?

bwoebi (approved):
Looks good now, thanks a lot for the effort in optimizing this!

bantonsson (approved):

PR #599, opened 2024-08-27
Fix accidental blocking socket in crashtracking

bwoebi:
Rust stdlib UnixListener is not async by default unlike the tokio one. It did apparently not fail during manual testing, but CI showed it sometimes failing.

Explicitly making it set_nonblocking(true).


bantonsson (approved):
Nice catch

r1viollet (approved):
LGTM

PR #601, opened 2024-08-28
Improve miri job runtime

VianneyRuhlmann:
# What does this PR do?

Improve the miri job runtime by skipping some slow tests.

### DIFF
--- a/ddcommon/src/entity_id/unix/container_id.rs
+++ b/ddcommon/src/entity_id/unix/container_id.rs
@@ -51,6 +51,7 @@ mod tests {
     use super::*;
     use maplit::hashmap;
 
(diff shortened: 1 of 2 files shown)


bantonsson (approved):
👍🏻

From the diff, it's clear that you reverted the parallel running of the tests, so it would be great if you had removed those commits as well.

It's easy to back up X commits:

```
git reset --hard HEAD~X
git push -f
```

VianneyRuhlmann:
Yes I figured those would get squashed anyway

bantonsson:
Yes, they will get squashed. It was only a bit confusing to see when I opened the PR.

PR #603, opened 2024-08-28
Add benchmark for SpanConcentrator

VianneyRuhlmann:
# What does this PR do?

Add benchmark to concentrator to allow future otpimizations

# Motivation

The concentrator is lacking some optimizations, this will make it easier to evaluate improvement.

# Additional Notes

# How to test the change?

### DIFF
--- a/data-pipeline/benches/span_concentrator.rs
+++ b/data-pipeline/benches/span_concentrator.rs
@@ -1,0 +1,71 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+use std::{
+    collections::HashMap,
+    time::{self, Duration, SystemTime},
+};
+
+use criterion::{criterion_group, Criterion};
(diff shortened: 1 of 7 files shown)


data-pipeline/benches/span_concentrator.rs
```diff
@@ -0,0 +1,71 @@
...
+        true,
+        vec!["db_name".to_string(), "bucket_s3".to_string()],
+    );
+    let mut spans = vec![];
+    for trace_id in 1..100 {
+        for span_id in 1..100 {
+            spans.push(get_span(now, trace_id, span_id));
+        }
+    }
+    c.bench_function("benching adding span to the SpanConcentrator", |b| {
```
bantonsson, line 57: I think it's better to use groups and names that look like this `tags/replace_trace_tags`, since you can then easily run single benchmarks by adding the  group/name last, ie `cargo bench -- ... tags/replace_trace_tags`

bantonsson (approved):
LGTM Only one comment about benchmark _name_.

hoolioh (approved):

PR #604, opened 2024-08-29
Add defensive ddcommon_ffi checks for Slice

morrisonlevi:
# What does this PR do?

Adds defensive checks in `Slice` in `ddcommon_ffi` and reroutes some code to use it.

# Motivation

It seems it is prudent to add checks here. At some point you need to be able to trust the FFI caller to uphold needed invariants, but it's a bit too easy to zero-initialize stuff in C and not fix it up for these Slices.

# Additional Notes

@AlexJF had some of their time wasted due to not early detecting this mistake..

# How to test the change?

Regular testing is fine. This accepts more valid inputs, but if you had invalid inputs for some edge cases, it might crash now.
1 . If you provide a null pointer but non-zero length will now panic.
2. If you provide a pointer that is not aligned correctly for the slice type, it will now panic.


ivoanjo:
I'm not a fan of this approach :sweat:. From our discussion on slack, the underlying assumption here is that we would catch all such crashes in development. And... I'm slightly suspicious that such an assumption would always hold.

The penalty of getting the assumption wrong is blowing up the client application, which is quite high... :boom: 

My suggestion is to change this to reporting back an API failure: This would equally be caught in development (as we're supposed to check errors on all datadog API calls already!), and in the off chance we never trigger the exact set of circumstances that would cause this error to show up, the penalty in production is that the profiler would get an error, which again we should already be handling.

---

(*If I'll be honest, I'd actually prefer to go with option 3: Libdatadog should correctly handle `NULL`s in this situation.*

*A `NULL` is a very natural thing for C-like languages, and I think we're leaking Rust details back to the API callers by forcing this to not be NULL. E.g. consider this: if `Slice::from_raw_parts` stated in its docs that a (NULL, 0) is fine for a slice, would we still force this constraint on our ffi? If not, does that not point to rust details leaking back, rather than actual intent in API design?*)

morrisonlevi:
I experimented with getting rid of most of the existing API and replaying it with a Result return.

```sh
% git diff --shortstat                                         
 22 files changed, 860 insertions(+), 507 deletions(-)
 ```

I still have pages of errors to fix.

I am okay if we go down this route, but it's not a good idea to do that without more discussion. It touches crash tracking, telemetry, profiling, and data pipeline. Maybe more, since I'm not finished. I can open this as a draft PR, if people want.

ivoanjo:
Fair enough, I agree the error approach "sounds good" has a big ripple effect in a lot of code/APIs.

How much do you dislike my suggestion of having the ffi code take care of handling nulls so the resulting API is more friendly to ffi callers?

I don't mind taking it up myself and writing such a PR.

If we have to, I believe we can live with this PR as-is but... I really do fear it's a really big sharp edge of the "oops you forgot you couldn't do this very natural/straightforward thing (from a caller POV) in this corner case and we blew up in production" kind :sweat_smile:

morrisonlevi:
> How much do you dislike my suggestion of having the ffi code take care of handling nulls so the resulting API is more friendly to ffi callers?

The PR handles slices with null pointers when the length is zero. It panics when:

1. The pointer is null and length is non-zero.
2. The pointer is mis-aligned for the slice type.
3. The length is larger than `isize::MAX` (I added this point after you last reviewed).

So it "handles" null pointers for the only case it could be reasonable. These checks will slow down the code, of course.

ivoanjo (approved):
Ack, I think the current version is quite reasonable. :pray: 

LGTM :+1:

PR #606, opened 2024-08-30
Add DogStatsD Package

duncanpharvey:
# What does this PR do?

Adds the package `dogstatsd`, a DogStatsD server which uses [Saluki](https://github.com/DataDog/saluki) for distribution metrics.

Moved upstream from https://github.com/DataDog/datadog-lambda-extension/tree/main/bottlecap/src/metrics

# Motivation

Add a DogStatsD server that will allow the Serverless Mini Agent to send runtime and custom metrics from Azure Spring Apps and Azure Functions (to be done in an upcoming PR).

https://datadoghq.atlassian.net/browse/SVLS-5177

# Additional Notes

- Updated Rust version from `1.71.1` to `1.76.0` -> **changes moved to https://github.com/DataDog/libdatadog/pull/612**
  *  `error[E0445]: crate-private trait MapFieldAccessor in public interface` on `1.71.1`, `1.72.0`, `1.73.0`
     - Resolved by upgrading Rust version
  * `error[E0658]: use of unstable library feature 'round_ties_even'` on `1.74.0`, `1.75.0`, `1.76.0`
    - Resolved by forking Datadog/saluki and replacing `round_ties_even()` with `round()`
      * https://github.com/DataDog/saluki-backport
 - Use `reqwest 0.12.4`
   * Error introduced in several other libdatadog crates when dependencies are pulled in for `reqwest 0.12.5` => `no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point`
- Updates to get this passing in CI
  * Add copyright headers
  * Apply formatting
  * Update license
  * Install Protoc Binary for `lint`, `test`, and `miri` workflows
    - https://github.com/protocolbuffers/protobuf/releases
    - Use `bash` shell when running test commands so `PATH` is picked up properly in Windows
  *  `dogstatsd` added to Alpine Dockerfile
  * Alpine docker image version updated to `3.19.3` to work with Rust `1.76.0`
  * Update benchmark image
    - Update to Rust 1.76.0: https://github.com/DataDog/benchmarking-platform/pull/96
    - Install unzip: https://github.com/DataDog/benchmarking-platform-tools/pull/16
  * Use clippy argument `invalid_reference_casting` instead of `clippy::cast_ref_to_mut`
  * Disable miri for integration test due to error
```
error: unsupported operation: can't call foreign function `socket` on OS `linux`
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mio-0.8.11/src/sys/unix/net.rs:28:18
   |
28 |     let socket = syscall!(socket(domain, socket_type, 0))?;
```

# How to test the change?

* Already present unit tests
* Integration test to confirm happy path works


bwoebi:
Currently the sidecar crate is using cadence for its dogstatsd needs. Maybe, as a first use case, and to show the crate working properly, the sidecar might be switched over to this crate?

duncanpharvey:
@bwoebi The original motivation for this is to use this dogstatsd crate in the [serverless](https://github.com/DataDog/libdatadog/tree/main/serverless) so I've done some initial testing to confirm that it works there. But the sidecar can definitely be switched over to use this crate as well!

I should have this PR ready to review soon once I get CI passing.

bwoebi:
Ah, I'm seeing you're using 1.77.2. We're currently locked in php to use rust 1.76.0 until the next alpine release.

duncanpharvey:
I'd like to avoid bumping the rust version but I'm running into these errors on versions below `1.77.x`. Any ideas for how to resolve these errors without upgrading rust?

  *  `error[E0445]: crate-private trait MapFieldAccessor in public interface` on `1.71.1`, `1.72.0`, `1.73.0`
  * `error[E0658]: use of unstable library feature 'round_ties_even'` on `1.74.0`, `1.75.0`, `1.76.0`

dogstatsd/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition.workspace = true
+version.workspace = true
+license.workspace = true
+
+[lib]
+bench = false
+
+[dependencies]
+datadog-protos = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki-backport/" }
+ddsketch-agent = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki-backport/" }
```
VianneyRuhlmann, line 13: We already have a ddsketch implementation in libdatadog, there is one in [mp-rs](https://github.com/DataDog/mp-rs/blob/020d8f72540ebfce2091ada1ed72e8eb5e97fd29/libs/dist_vocab/src/ddsketch.rs#L20) and there is [one on crates.io](https://crates.io/crates/sketches-ddsketch) that is not owned by DD. We should probably all agree on a single implementation.

duncanpharvey: Interesting, this package uses the same implementation used by the Datadog Lambda Extension (owned by the Serverless team). Since I intend to use this DogStatsD package for a Serverless solution in Azure I'm inclined to stay consistent with the implementations.

https://github.com/DataDog/datadog-lambda-extension/tree/main/bottlecap/src/metrics

tools/docker/Dockerfile.build
```diff
@@ -1,4 +1,4 @@
-ARG ALPINE_BASE_IMAGE="alpine:3.18"
+ARG ALPINE_BASE_IMAGE="alpine:3.19"
```
gleocadie, line 1: The image used in gitlab is 3.19.3.
https://github.com/DataDog/libddprof-build/blob/main/.gitlab-ci.yml#L10
Shouldn't use the same version here too ?

dogstatsd/Cargo.toml
alexgallotta: I tried to remove the duplication but I read that this one is slightly different and optimized for the agent. So it might be not possible
https://github.com/DataDog/saluki/blob/main/lib/ddsketch-agent/src/lib.rs#L149
>>> /// This implementation is subtly different from the open-source implementations of `DDSketch`, as Datadog made some
/// slight tweaks to configuration values and in-memory layout to optimize it for insertion performance within the
/// agent.

tools/docker/Dockerfile.build
duncanpharvey: Updated to use 3.19.3 ✅ 

https://github.com/DataDog/libdatadog/blob/5d181a07cec2fc66c9489d04c764f5cf2faac93e/tools/docker/Dockerfile.build#L1

dogstatsd/Cargo.toml
morrisonlevi: > slight tweaks to configuration values and in-memory layout to optimize it for insertion performance within the agent.

I'm not familiar with sketches, but optimizing in-memory layout and optimizing for insertion performance also sound relevant to a library, if it is feasible. Any insight as to whether it's appropriate or not?

duncanpharvey:
> Does this spawn any threads on the client side of things? I see a test server which probably does, and that's okay. I scanned through it, but it's a big PR so I may have missed something.
> 
> Most languages need to have control over threads so they can mask off appropriate signals, register them with the runtime, etc, so the libraries generally need to design APIs that don't spawn threads.

@morrisonlevi Correct, the only thread spawned is in the test.

https://github.com/DataDog/libdatadog/blob/5d181a07cec2fc66c9489d04c764f5cf2faac93e/dogstatsd/tests/integration_test.rs#L83-L85

dogstatsd/Cargo.toml
alexgallotta: I am afraid I did not understand your question, can you elaborate more?

tobz: 👋🏻 

I'm the original author of the ported DDSketch implementation being used here. As the doc comments allude to, the main goal of the port was to maintain fidelity with the Agent's own DDSketch implementation, which at the time (and to this day) is not based on the open-source Go implementation of DDSketch.

I didn't write the original Agent-specific implementation, so I can't speak to all of the decisions made there, but I also tried to port over some of the code comments to, again, maintain fidelity with the Agent implementation.

While it probably behooves us, in a general sense, to have One True Implementation in Rust... maintaining compatibility with the Datadog Agent in the most important factor in this case.

alexgallotta (approved):

PR #607, opened 2024-09-02
[PROF-9476] Managed string storage for interning over several profiles

AlexJF:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

[PROF-9476]: https://datadoghq.atlassian.net/browse/PROF-9476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ivoanjo (commented):
This looks quite nice!

The blast radius is not very big, and to be honest, 99% of it seems to be the change to make apis take both an id or a string.

I think if we can solve that one, then the diff for this feature would be quite small + it'll be very close to zero cost for libdatadog clients that aren't using it, which seems like a really nice property.

profiling/src/collections/string_storage.rs
```diff
@@ -0,0 +1,208 @@
...
+            None => {
+                let (index, _inserted) = self.set.insert_full(item.clone());
+                // This wouldn't make any sense; the item couldn't be found so
+                // we try to insert it, but suddenly it exists now?
+                debug_assert!(_inserted);
+                index
+            }
+        };
+        StringId::from_offset(index)
+    }
```
ivoanjo, line 59: It occurred to me that while ideally all this structure would be lock-free and whatnot, inserting and claiming an index is actually the part that would be really nice to be able to do in a non-blocking manner.

I was googling and didn't find a lot of interesting solid/production pre-existing data structure implementations with the kinds of characteristics we'd want (which is super disappointing -- I miss you java.util.concurrent), but yea, may be worth keeping an eye out if we find something.

ivoanjo: Maybe https://crates.io/crates/dashmap ? But we'd need to solve the index assignment (easy if we don't do index reuse, really annoying if we want to do that)

ivoanjo:
Replaced by #725

PR #608, opened 2024-09-03
[ruby] Prepare gem release for version 12

AlexJF:
# What does this PR do?

The ruby gem for libdatadog was still only setup for version 11. This PR prepares the release for version 12.

# Motivation

Keep things up to date.

# Additional Notes

N/A

# How to test the change?

We'll trigger the gem release process once this gets merged and will soon push a PR in dd-trace-rb to make it use the new gem.

### DIFF
--- a/ruby/lib/libdatadog/version.rb
+++ b/ruby/lib/libdatadog/version.rb
@@ -2,11 +2,11 @@
 
 module Libdatadog
   # Current libdatadog version
-  LIB_VERSION = '12.0.0'
+  LIB_VERSION = "12.0.0"
 
-  GEM_MAJOR_VERSION = '1'
-  GEM_MINOR_VERSION = '0'
-  GEM_PRERELEASE_VERSION = '' # remember to include dot prefix, if needed!
+  GEM_MAJOR_VERSION = "1"
+  GEM_MINOR_VERSION = "0"
+  GEM_PRERELEASE_VERSION = "" # remember to include dot prefix, if needed!
   private_constant :GEM_MAJOR_VERSION, :GEM_MINOR_VERSION, :GEM_PRERELEASE_VERSION
 
   # The gem version scheme is lib_version.gem_major.gem_minor[.prerelease].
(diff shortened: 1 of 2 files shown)


ruby/lib/libdatadog/version.rb
```diff
@@ -2,11 +2,11 @@
...
   # Current libdatadog version
-  LIB_VERSION = "11.0.0"
+  LIB_VERSION = '12.0.0'
 
-  GEM_MAJOR_VERSION = "1"
-  GEM_MINOR_VERSION = "0"
-  GEM_PRERELEASE_VERSION = "" # remember to include dot prefix, if needed!
+  GEM_MAJOR_VERSION = '1'
+  GEM_MINOR_VERSION = '0'
+  GEM_PRERELEASE_VERSION = '' # remember to include dot prefix, if needed!
```
ivoanjo, line 9:
```suggestion
  GEM_MAJOR_VERSION = "1"
  GEM_MINOR_VERSION = "0"
  GEM_PRERELEASE_VERSION = "" # remember to include dot prefix, if needed!
```

ruby/lib/libdatadog/version.rb
```diff
@@ -2,11 +2,11 @@
 
 module Libdatadog
   # Current libdatadog version
-  LIB_VERSION = "11.0.0"
+  LIB_VERSION = '12.0.0'
```
ivoanjo, line 5: The gem is using standardrb as well, which defaults strings to `"`.

This will break once we try to run `docker-compose run push_to_rubygems` since it runs the auto-formatter first which will undo this and then complain "There are files that need to be committed first."

```suggestion
  LIB_VERSION = "12.0.0"
```

ivoanjo (approved):
Thanks for picking this up! Slight "nitpick" with the format as standardrb won't like it ;)

ruby/lib/libdatadog/version.rb
AlexJF: Uh weird, this was autoformatted. Let me check why the autoformat was not using standard

PR #610, opened 2024-09-06
improve assert failure message for send_data retry tests and increase tolerance for time measurements

ekump:
# What does this PR do?
It has been observed that some of the send_data retry strategy logic can be flaky on the macOS github actions runners. This PR adds more time to the retry observation tolerances for these tests. Also, a bit more info was added to the relevant test assertions when a failure does occur. 

# Motivation
@bwoebi


bwoebi (approved):
Thanks, looks good.

PR #616, opened 2024-09-09
[Serverless Mini Agent] Use DogStatsD in Serverless

duncanpharvey:
# What does this PR do?

Starts a DogStatsD server when running the Serverless Mini Agent.

# Motivation

Allow runtime metrics and custom metrics to be sent from Azure Spring Apps and Azure Functions.

- https://datadoghq.atlassian.net/browse/SVLS-5177
- https://datadoghq.atlassian.net/browse/SVLS-5521

# Additional Notes

- Updated `main` method of Serverless to be asynchronous.
- Allow custom DogStatsD port with `DD_DOGSTATSD_PORT`
- Allow DogStatsD to be disabled with `DD_USE_DOGSTATSD`

# How to test the change?

* Build mini agent and deploy to Azure Spring App and Azure Function
* Confirm runtime metrics are sent successfully to Datadog

### DIFF
--- a/serverless/src/main.rs
+++ b/serverless/src/main.rs
@@ -2,17 +2,40 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use env_logger::{Builder, Env, Target};
-use log::{error, info};
-use std::{env, sync::Arc};
+use log::{debug, error, info};
+use std::{env, sync::Arc, sync::Mutex};
+use tokio::time::{sleep, Duration};
 
 use datadog_trace_mini_agent::{
     config, env_verifier, mini_agent, stats_flusher, stats_processor, trace_flusher,
     trace_processor,
 };
 
-pub fn main() {
+use dogstatsd::{
+    aggregator::Aggregator as MetricsAggregator,
+    constants::CONTEXTS,
+    dogstatsd::{DogStatsD, DogStatsDConfig},
+    flusher::{build_fqdn_metrics, Flusher},
+};
+
+use tokio_util::sync::CancellationToken;
+
+const DOGSTATSD_FLUSH_INTERVAL: u64 = 10;
+
+#[tokio::main]
+pub async fn main() {
     let env = Env::new().filter_or("DD_LOG_LEVEL", "info");
     Builder::from_env(env).target(Target::Stdout).init();
+
+    let dd_api_key: Option<String> = env::var("DD_API_KEY").ok();
+    let dd_dogstatsd_port: u16 = env::var("DD_DOGSTATSD_PORT")
+        .ok()
+        .and_then(|port| port.parse::<u16>().ok())
+        .unwrap_or(8125);
+    let dd_site = env::var("DD_SITE").unwrap_or_else(|_| "datadoghq.com".to_string());
+    let dd_use_dogstatsd = env::var("DD_USE_DOGSTATSD")
+        .map(|val| val != "false")
+        .unwrap_or(true);
 
     info!("Starting serverless trace mini agent");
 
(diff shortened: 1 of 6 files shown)


serverless/src/main.rs
```diff
@@ -44,7 +67,61 @@ pub fn main() {
...
+    if !dd_use_dogstatsd {
+        return;
     }
+
+    let metrics_aggr = Arc::new(Mutex::new(
+        MetricsAggregator::new(Vec::new(), CONTEXTS).expect("Failed to create metrics aggregator"),
+    ));
+
+    info!("Starting DogStatsD");
+    let _ = start_dogstatsd(dd_dogstatsd_port, &metrics_aggr).await;
```
astuyve, line 86: Probably okay for now not to use the cancel token but at some point you'll want to ensure you flush before the final shutdown of the container.

serverless/src/main.rs
```diff
@@ -44,7 +67,61 @@ pub fn main() {
...
-        error!("Error when starting serverless trace mini agent: {e}");
+    tokio::spawn(async move {
+        let res = mini_agent.start_mini_agent().await;
+        if let Err(e) = res {
+            error!("Error when starting serverless trace mini agent: {e:?}");
+        }
+    });
+
+    if !dd_use_dogstatsd {
+        return;
```
astuyve, line 78: I usually don't love an early return like this, I generally prefer if you'd just wrap the rest in this conditional just so somebody doesn't add more behavior after this line and not realize it isn't executing if statsd is disabled

astuyve (approved):
Couple notes, nothing major

serverless/src/main.rs
duncanpharvey: Yeah it works well enough for a beta just flushing every 10 seconds. But I still need to do some research for how to listen for shutdown events. Already have a card in our backlog for this one!

serverless/src/main.rs
```diff
@@ -2,18 +2,41 @@
...
     Builder::from_env(env).target(Target::Stdout).init();
 
+    let dd_api_key: Option<String> = env::var("DD_API_KEY").ok();
+    let dd_dogstatsd_port: u16 = env::var("DD_DOGSTATSD_PORT")
+        .ok()
+        .and_then(|port| port.parse::<u16>().ok())
+        .unwrap_or(8125);
+    let dd_site = env::var("DD_SITE").unwrap_or_else(|_| "datadoghq.com".to_string());
+    let dd_use_dogstatsd = env::var("DD_USE_DOGSTATSD")
+        .map(|val| val != "false")
```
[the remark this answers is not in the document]
duncanpharvey: Updated: https://github.com/DataDog/libdatadog/blob/01dcdf8002a611fc404e6c423ad1fc73f27dd514/serverless/src/main.rs#L36-L38

serverless/src/main.rs
duncanpharvey: Makes sense, I was trying to reduce the level of nesting but I agree it's more robust to not have the early return.

serverless/src/main.rs
```diff
@@ -44,7 +67,76 @@ pub fn main() {
...
+        }
+    });
+
+    let mut metrics_flusher = if dd_use_dogstatsd {
+        let metrics_aggr = Arc::new(Mutex::new(
+            MetricsAggregator::new(Vec::new(), CONTEXTS)
+                .expect("Failed to create metrics aggregator"),
+        ));
+
+        info!("Starting dogstatsd");
```
duncanista, line 83: Should this be `debug` logs instead of `info?`

serverless/src/main.rs
```diff
@@ -44,7 +67,76 @@ pub fn main() {
...
+            }
+            None => {
+                error!("DD_API_KEY not set, won't flush metrics");
+                None
+            }
+        }
+    } else {
+        info!("dogstatsd disabled");
+        None
+    };
```
duncanista, line 104: All of this could be abstracted into the `start_dogstatsd` method

serverless/src/main.rs
```diff
@@ -44,7 +67,76 @@ pub fn main() {
...
+    flush_interval.tick().await; // discard first tick, which is instantaneous
+
+    loop {
+        flush_interval.tick().await;
+
+        if let Some(metrics_flusher) = metrics_flusher.as_mut() {
+            debug!("Flushing dogstatsd metrics");
+            metrics_flusher.flush().await;
+        } else {
+            debug!("dogstatsd disabled, not flushing metrics");
```
duncanista, line 116: You don't are already logging this, might not be worth having this

trace-mini-agent/src/config.rs
```diff
@@ -41,6 +42,10 @@ impl Config {
             anyhow::anyhow!("Unable to identify environment. Shutting down Mini Agent.")
         })?;
 
+        let dd_dogstatsd_port: u16 = env::var("DD_DOGSTATSD_PORT")
+            .ok()
+            .and_then(|port| port.parse::<u16>().ok())
+            .unwrap_or(8125);
```
duncanista, line 48: magic number

serverless/src/main.rs
duncanpharvey: Refactored to move the metrics aggregator and flusher into `start_dogstatsd`

serverless/src/main.rs
duncanpharvey: Changed this and the `Starting serverless trace mini agent` log to both be `debug` logs.

trace-mini-agent/src/config.rs
duncanpharvey: Added an environment variable `DEFAULT_DOGSTATSD_PORT` for this as well as an `AGENT_HOST` environment variable.

serverless/src/main.rs
duncanpharvey: https://github.com/DataDog/libdatadog/blob/55201cf0181104a17e7eefc13baa7804ab1de595/serverless/src/main.rs#L102-L143

PR #617, opened 2024-09-09
Upstream dogstatsd refactors

alexgallotta:
- remove extra structs not needed
- fixed false positive wrong tag format
- optimization: avoid extra sorting for id
- optimization: use ustr everywhere for tags


taegyunkim (changes requested):
Please fix lint and 3rd party license checks.

dogstatsd/rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "1.78.0"
```
taegyunkim, line 2: libdatadog repository supports rust 1.76.0 or above. 
https://github.com/DataDog/libdatadog#build-dependencies
This line might require us to updating the whole repo and build pipeline. Also, this has some implication to php folks who AFAIK build libdatadog from sources.

taegyunkim: And this might be the reason build jobs are failing

https://gitlab.ddbuild.io/DataDog/apm-reliability/libddprof-build/-/jobs/634108221

alexgallotta: The reason why I pushed to 1.78 is that there was an unsupported instruction in saluki.
Right now I can't downgrade to 1.76 because
```
 cargo test
   Compiling datadog-protos v0.1.0 (https://github.com/DataDog/saluki-backport/#a3e224b2)
error: unsupported output in build script of `datadog-protos v0.1.0 (https://github.com/DataDog/saluki-backport/#a3e224b2)`: `cargo::rerun-if-changed=proto/datadog/api/v1/api.proto`
Found a `cargo::key=value` build directive which is reserved for future use.
Either change the directive to `cargo:key=value` syntax (note the single `:`) or upgrade your version of Rust.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```

I will check with @duncanpharvey tomorrow, since saluki is using 1.80

alexgallotta: I removed the different toolchain @taegyunkim 
is it ok to merge now?

taegyunkim (approved):

PR #618, opened 2024-09-10
move dogstatsd-client to separate crate so it can be shared

ajgajg1134:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ members = [
   "data-pipeline",
   "data-pipeline-ffi",
   "ddsketch",
-  "tinybytes",
+  "tinybytes", "dogstatsd-client",
 ]
 
 default-members = [
--- a/dogstatsd-client/Cargo.toml
+++ b/dogstatsd-client/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "dogstatsd-client"
+rust-version.workspace = true
+edition.workspace = true
+version.workspace = true
+license.workspace = true
+
+[dependencies]
+ddcommon = { path = "../ddcommon" }
+datadog-trace-protobuf = { path = "../trace-protobuf" }
+datadog-trace-utils = { path = "../trace-utils" }
+datadog-trace-normalization = { path = "../trace-normalization" }
+datadog-ddsketch = { path = "../ddsketch"}
+cadence = "1.3.0"
+serde = { version = "1.0", features = ["derive", "rc"] }
+tracing = { version = "0.1", default-features = false }
+anyhow = { version = "1.0" }
+hyper = { version = "0.14", features = ["client"], default-features = false }
+http = "0.2"
--- a/dogstatsd-client/src/lib.rs
+++ b/dogstatsd-client/src/lib.rs
@@ -1,4 +1,4 @@
-// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
 use ddcommon::tag::Tag;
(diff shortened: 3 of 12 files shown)


Cargo.toml
```diff
@@ -33,7 +33,7 @@ members = [
   "data-pipeline",
   "data-pipeline-ffi",
   "ddsketch",
-  "tinybytes",
+  "tinybytes", "dogstatsd-client",
```
ekump, line 36: consistency nit: This should be on a separate line.

dogstatsd-client/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "dogstatsd-client"
+rust-version.workspace = true
+edition.workspace = true
+version.workspace = true
+license.workspace = true
+
+[dependencies]
+ddcommon = { path = "../ddcommon" }
+datadog-trace-protobuf = { path = "../trace-protobuf" }
+datadog-trace-utils = { path = "../trace-utils" }
```
ekump, line 11: This dependency is not necessary.

dogstatsd-client/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
```
ekump, line 1: I have no clue what the correct answer is, but should we be modifying the license headers like this? I'm not sure what is legally correct.

data-pipeline/src/trace_exporter.rs
```diff
@@ -213,23 +231,36 @@ impl TraceExporter {
             return Ok(String::from("{}"));
         }
 
+        // todo: do we need to modify the client to allow for &str to avoid allocating a String?
```
ekump, line 234: The client should probably have some sort of `ActionType` enum implemented instead of passing strings like this. Partially for the allocation cost, and partially to reduce the likelihood of typos.

data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -89,7 +89,7 @@ pub unsafe extern "C" fn ddog_trace_exporter_free(handle: Box<TraceExporter>) {
 /// * `trace_count` - The number of traces to send to the Datadog Agent.
 #[no_mangle]
 pub unsafe extern "C" fn ddog_trace_exporter_send(
-    handle: &TraceExporter,
+    handle: &mut TraceExporter,
```
ekump, line 92: We can avoid making the entire TraceExporter mutable by leveraging something like [RefCell](https://doc.rust-lang.org/std/cell/struct.RefCell.html) to achieve [interior mutability](https://doc.rust-lang.org/book/ch15-05-interior-mutability.html). 

[This commit compiles](https://github.com/DataDog/libdatadog/commit/16b4bcc0c01b5779e7464aebf4103c6162d2095e).

There is a minor performance hit to using RefCell, although I'm not sure we'd notice it. 

It _feels right_ to me for the TraceExporter to be immutable, but I'm not sure that's sufficient justification? I'm curious what other people think.

data-pipeline/src/trace_exporter.rs
```diff
@@ -249,11 +280,21 @@ impl TraceExporter {
...
+                                }
                                 Ok(String::from("{}"))
                             }
                         },
                         Err(err) => {
                             error!("Error sending traces: {err}");
+                            if let Some(ref mut flusher) = self.dogstatsd {
+                                flusher.send(vec![DogStatsDAction::Count(String::from("datadog.libdatadog.send.errors"),
+                                                                         1,
+                                                                         vec![])]);
```
ekump, line 296: `Vec::default()` is probably preferable here? I believe vec![] is going to allocate memory here, whereas Vec::new() will not allocate memory until elements get added to it. Plus, it more clearly shows your intention of passing an empty vec.

dogstatsd-client/src/lib.rs
ajgajg1134: iirc it doesn't strictly matter as long as the header is there when the file is created

PR #619, opened 2024-09-10
[APMSP-1317] Fetch agent info endpoint

VianneyRuhlmann:
# What does this PR do?

Add AgentInfoFetcher to fetch the agent info endpoint and keep it updated

# Motivation

The info endpoint stores some essential parameters for client side stats

# Additional Notes

# How to test the change?


agent-info/src/schema.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Hash of the info
+    pub state_hash: String,
+    /// Info response from the agent
+    pub info: AgentInfoStruct,
+}
+
+/// Schema of an agent info response
+#[allow(missing_docs)]
+#[derive(Clone, Deserialize, Default, Debug, PartialEq)]
+pub struct AgentInfoStruct {
```
knusbaum, line 19: This should be updated with the latest changes to the RFC and here:
https://github.com/DataDog/datadog-agent/pull/28861

knusbaum (changes requested):
Looks mostly good. Just need to update the info struct to reflect the changes we're making regarding some of the fields, namely `span_kinds_stats_computed`

agent-info/src/fetcher.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+/// Fetch the info endpoint and update an ArcSwap keeping it up-to-date.
+///
+/// Once the fetcher has been created you can get an Arc of the config by calling `get_info`.
+/// You can then start the run method, the fetcher will update the AgentInfoArc based on the
+/// given refresh interval.
+///
+/// # Example
+/// ```no_run
```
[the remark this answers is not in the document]
VianneyRuhlmann: I think the `no_run` doc tests are still compiled just not run (which is not possible since it does http calls). But I can add some test with the test agent.

agent-info/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
```
ekump, line 1: Why do we need this to be in a dedicated crate?

VianneyRuhlmann: I figured several other crates might need it in the future and it seemed a bit to complex to be in ddcommon.

agent-info/src/schema.rs
VianneyRuhlmann: done

agent-info/src/fetcher.rs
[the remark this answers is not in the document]
VianneyRuhlmann: I added some integration tests with the test agent. We need to wait for https://github.com/DataDog/dd-apm-test-agent/pull/193 to get merged since the test agent does not send the agent state header. I also added a `require_test_agent` module to the trace-utils integration test to filter out all tests that use it in the pipeline.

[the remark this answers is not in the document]
ekump: Ah, TIL. I incorrectly assumed `no_run` was completely ignored by the compiler. In that case, I have no objection to using it.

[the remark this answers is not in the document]
ekump: > I also added a require_test_agent module to the trace-utils integration test to filter out all tests that use it in the pipeline.

I'm not sure that's necessary. We already run integration tests separately on CI. I think the correct separation for tests is the type (unit, integration, fuzz, etc). Also, by definition I believe all trace-utils integrations should be hitting the test agent anyway.

agent-info/Cargo.toml
ekump: Unless I'm misunderstanding. the complexity exists solely because of our specific use-case and implementation of client-side stats. 

I can see the value in having the info endpoint's schema along with some possible constants for routes, etc in `ddcommon` (or somewhere else appropriate for common code) to reduce boilerplate around making http requests.  

I can also see value in a crate that offers some sort of generic event-driven solution for polling and notifying consumers of HTTP endpoint changes (i'm surprised this doesn't already exist). 

But this PR feels very un-generic. I'm not suggesting we have to come up with a different solution, but I think it should just live in data-pipeline, at least for now.

VianneyRuhlmann: So something like, moving the `fetch_info` function to ddcommon maybe behind a feature, and the `AgentInfoFetcher` into `data_pipeline` ?

agent-info/src/fetcher.rs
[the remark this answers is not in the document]
VianneyRuhlmann: We need a way to filter out the tests using the test agent in `.github/workflows/test.yml` since it requires docker, right ?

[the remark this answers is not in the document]
ekump: Yes. We currently [do that](https://github.com/DataDog/libdatadog/blob/main/.github/workflows/test.yml#L57) with the `-E '!test(tracing_integration_tests::)'` flag.

agent-info/Cargo.toml
ekump: I am ok with all of it living in `data-pipeline` for now since this is the only current use-case I'm aware of and refactoring later wouldn't be a big enough of an effort to violate [the rule of threes](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

But, if you did want to put something in ddcommon I'd suggest just the schema.

agent-info/src/fetcher.rs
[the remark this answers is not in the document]
VianneyRuhlmann: Can we assume all integration tests will require the test agent ?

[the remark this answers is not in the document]
ekump: I think we can assume all integration tests may (but not necessarily) require external entities like containers running the test-agent or other services. I think we also want to assume integration tests should run separately from unit tests and other workflows in CI. These assumptions are why I think a less specific filter of just `integration_tests` is preferable to filtering on whether or not they use the test-agent. 

For data-pipeline or trace-utils specifically - yes I think we would assume the test-agent is being used. But I wouldn't necessarily say the same for other crates in libdatadog that may wish to write integration tests.

data-pipeline/src/agent_info/schema.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub state_hash: String,
+    /// Info response from the agent
+    pub info: AgentInfoStruct,
+}
+
+/// Schema of an agent info response
+#[allow(missing_docs)]
+#[derive(Clone, Deserialize, Default, Debug, PartialEq)]
+pub struct AgentInfoStruct {
+    /// Version of the agent
```
ekump, line 20: Any reason why some of the public fields have doc comments and others don't?

VianneyRuhlmann: The fields are not documented in the agent (or at least I couldn't find it). So I commented the ones that were obvious to me, admittedly not the most useful documentation.

ekump:
FYI, this branch needs a Jira ticket associated with it.

data-pipeline/tests/test_fetch_info.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+#[cfg(test)]
+mod require_test_agent {
```
ekump, line 5: These tests don't seem to pass yet. Are you still waiting on changes from the test-agent to be deployed?

data-pipeline/src/agent_info/fetcher.rs
```diff
@@ -0,0 +1,349 @@
...
+use super::{schema::AgentInfo, AgentInfoArc};
+use anyhow::{anyhow, Result};
+use arc_swap::ArcSwapOption;
+use ddcommon::{connector::Connector, Endpoint};
+use hyper::{self, body::Buf, header::HeaderName};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::sleep;
+
+#[allow(clippy::declare_interior_mutable_const)]
```
ekump, line 15: If you wanted to avoid this allow you could use `once_cell`. Something like:

```rust
use once_cell::sync::Lazy;

static DATADOG_AGENT_STATE: Lazy<HeaderName> = Lazy::new(|| HeaderName::from_static("datadog-agent-state"));

File: data-pipeline/src/agent_info/fetcher.rs [hunk 7]
```diff
@@ -0,0 +1,349 @@
...
+    /// Start fetching the info endpoint with the given interval.
+    ///
+    /// # Warning
+    /// This method does not return and should be called within a dedicated task.
+    pub async fn run(&self) {
+        loop {
+            let current_info = self.info.load();
+            let current_hash = current_info.as_ref().map(|info| info.state_hash.as_str());
+            let res = fetch_info_with_state(&self.info_endpoint, current_hash).await;
+            if let Ok(FetchInfoStatus::NewState(new_info)) = res {
```
ekump, line 144: what about the not OK state? At a minimum, we should probably log something, no?

data-pipeline/src/agent_info/fetcher.rs
```diff
@@ -0,0 +1,349 @@
...
+                self.info.store(Some(Arc::new(*new_info)));
+            }
+            sleep(self.refresh_interval).await; // Wait 5 min between each call to /info
```
ekump, line 147: Comment should probably remove the `5 min` part?

data-pipeline/src/agent_info/fetcher.rs
```diff
@@ -0,0 +1,349 @@
...
+            refresh_interval,
+        }
+    }
+
+    /// Start fetching the info endpoint with the given interval.
+    ///
+    /// # Warning
+    /// This method does not return and should be called within a dedicated task.
+    pub async fn run(&self) {
+        loop {
```
ekump, line 140: Do we need some sort of graceful shutdown procedure here? What happens if the host tracer initiates a graceful shutdown?

agent-info/src/fetcher.rs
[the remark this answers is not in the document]
VianneyRuhlmann: I'm okay with running the integrations tests in a separated ci step. But we still need to be able to run the tests using the test agent separately as they only Linux (at least we [only run them on linux in the CI](https://github.com/DataDog/libdatadog/blob/09d1ef506d59c65cbe58f0a2e153d493716efaa3/.github/workflows/test.yml#L64)) whereas the other integration tests run on all platform.

data-pipeline/src/agent_info/fetcher.rs
VianneyRuhlmann: The lint is a [false positive from clippy](https://github.com/rust-lang/rust-clippy/issues/9776) which [has been fixed in rust 1.80](https://github.com/rust-lang/rust-clippy/pull/12691) so I it is probably ok to just allow it here

data-pipeline/src/agent_info/fetcher.rs
VianneyRuhlmann: I not sure on how tokio will behave on shutdown. Since the host tracer created the runtime will it get shutdown has well ? The fetcher itself doesn't require any graceful shutdown as long it is shutdown as well.

VianneyRuhlmann:
Requires https://github.com/DataDog/libddprof-build/pull/52 to be merged

data-pipeline/src/agent_info/fetcher.rs
ekump: We don't need to address it as part of this PR, but we should probably create a ticket to confirm shutdown behavior for the data pipeline.

agent-info/src/fetcher.rs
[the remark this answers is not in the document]
ekump: If you are worried about operating systems wouldn't it be better to just ignore tests on OSes we don't want to use with `#[cfg(not(target_os = "x"))]`?

PR #624, opened 2024-09-12
Fix CI failure: update `libdatadog_profiling.so` `SONAME`

gleocadie:
# What does this PR do?

Update the `SONAME` in the `libdatadog_profiling.so` file.

# Motivation

Fix CI.

Cargo produces the `libdatadog_profiling_ffi.so` then late renames it into `libdatadog_profiling.so`. But the `SONAME` in `libdatadog_profiling.so` is `libdatadog_profiling_ffi.so` (same as the original).
So when building the symbolizer (for example) and linking against `libdatadog_profiling.so`, the linker will register `libdatadog_profiling_ffi.so` as `NEEDED`

image

But it's not its name anymore, so we have to patch `libdatadog_profiling.so` to change its `SONAME` into `libdatadog_profiling.so`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -183,13 +183,12 @@ jobs:
         if: matrix.platform != 'windows-latest'
         run: |
           set -e
-          rm -rf examples/ffi/build
           mkdir examples/ffi/build
           cd examples/ffi/build
           # Add BUILD_SYMBOLIZER variable only for Linux platforms
           if [[ "${{ matrix.platform }}" == "ubuntu-latest" ]]; then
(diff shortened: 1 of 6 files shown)


sanchda (approved):
I think it'd be ideal to move away from patching binaries once we produce them, but if we're already renaming the file, then we absolutely need to patch it in order for the soname and file name to be in conformance (otherwise, `-l` won't work properly at all).

sanchda:
Thank you for fixing this

morrisonlevi:
As discussed in Slack, I think we should do things the "right" way if we can, but that's more work that may have unknowns. Let's get this incrementally fixed and try to do things better in another PR.

morrisonlevi (approved):

PR #628, opened 2024-09-13
Fix SONAME for artifact produced in Gitlab

gleocadie:
# What does this PR do?

Fix `SONAME` for artifacts produced in Gitlab.

# Motivation

Since in GitLab we do not use the new builder, we have to update the `build-profiling.sh` shell script in order to port this [change](https://github.com/DataDog/libdatadog/pull/624).

# Additional Notes
🤷 

# How to test the change?

Download the artifact from Gitlab and run `readelf -d libdatadog_profiling.so`
image

### DIFF
--- a/build-profiling-ffi.sh
+++ b/build-profiling-ffi.sh
@@ -169,6 +169,10 @@ if [[ "$fix_macos_rpath" -eq 1 ]]; then
     install_name_tool -id @rpath/${shared_library_name} "$destdir/lib/${shared_library_name}"
 fi
 
+if command -v patchelf > /dev/null && [[ "$target" != "x86_64-pc-windows-msvc" ]]; then
+    patchelf --set-soname ${shared_library_name}  "$destdir/lib/${shared_library_rename}"
+fi
+
 # objcopy might not be available on macOS
 if command -v objcopy > /dev/null && [[ "$target" != "x86_64-pc-windows-msvc" ]]; then
     # Remove .llvmbc section which is not useful for clients


build-profiling-ffi.sh
```diff
@@ -169,6 +169,25 @@ if [[ "$fix_macos_rpath" -eq 1 ]]; then
...
+    # In this specific case, we download the latest version of patchelf for aarch64 and use it
+    # https://github.com/NixOS/patchelf/pull/117
+    if [[ "$target" == "aarch64-unknown-linux-gnu" ]]; then
+        mkdir patchelf
+        cd patchelf
+        curl -OL https://github.com/NixOS/patchelf/releases/download/0.18.0/patchelf-0.18.0-aarch64.tar.gz
+        tar zxvf patchelf-0.18.0-aarch64.tar.gz
+        patchelf_bin=`pwd`"/bin/patchelf"
+        cd ..
+    fi
```
ivoanjo, line 186: Minor: Maybe add here a `echo "Downloading patchelf to workaround bug in centos patchelf"` in case someone gets surprised by this?

ivoanjo (approved):
:+1: LGTM Thanks for getting to the bottom of this whole soname thing :)

build-profiling-ffi.sh
gleocadie: I updated the Docker image instead so we only have the call to `patchelf`

ivoanjo (approved):
:+1: :tada:

PR #638, opened 2024-09-18
Health Metrics for data-pipeline + Dogstatsd-Client Crate

ajgajg1134:
# What does this PR do?

Add some health metrics to the datapipeline-crate to assist with debugging any issues in customer environments. To support this the dogstatsd-client code from the sidecar has been promoted to a separate crate where it can be utilized by both.

# Motivation

health metrics are good for debugging, we've used them successfully in other tracers.

# Additional Notes

Unfortunately there is a bit of new code duplication in the dogstatsd-client crate. This was to get around some very difficult to read rust syntax in trying to write a `DogStatsDAction` type that could accept `Vec<Tag>`, `&[&Tag]` and a chained iterator of `&Tag`. Doing this turned into a very difficult task that seems harder than it's worth, if there are any rust experts reviewing this who want to take a stab please feel free to do so, just ensure that your resulting code is able to be used by BOTH the sidecar and data-pipeline as these crates both use the client with different types.

# How to test the change?

Unit tests here provide good coverage for these implementations.

Ticket: APMSP-1248


pierotibou (commented):
Thanks a lot. And very cool first contribution. I tried to do a quick first review, focusing on functionals obviously but It's late here so not able to go through all of it. 
Please link the Jira ticket in the PR desc.

data-pipeline/src/trace_exporter.rs
```diff
@@ -197,13 +235,29 @@ impl TraceExporter {
             })
     }
 
+    /// Emit a health metric to dogstatsd
+    fn emit_metric(&self, metric: HealthMetric, custom_tags: Option<Vec<&Tag>>) {
+        let tags = match custom_tags {
```
pierotibou, line 240: I don't understand anything to rust but is there a perf impact (even minimal) to create tags if we don't use them? Should this be in the `if` below?

data-pipeline/src/trace_exporter.rs
```diff
@@ -117,6 +132,9 @@ pub struct TraceExporter {
     // TODO - do something with the response callback - https://datadoghq.atlassian.net/browse/APMSP-1019
     _response_callback: Option<Box<dyn ResponseCallback>>,
     runtime: Runtime,
+    // None if dogstatsd is disabled
+    dogstatsd: Option<Flusher>,
```
pierotibou, line 136: IIUC, this code isn't yet called and when we'll call it, we'll pass an flusher if the tracer have health metrics enabled. Is that right?

data-pipeline/src/trace_exporter.rs
```diff
@@ -38,6 +41,13 @@ pub enum TraceExporterOutputFormat {
     V07,
 }
 
+// internal health metrics
```
pierotibou, line 44: It seems I'e lost that comment in the first review. I expected the following metric names (maybe some customers or TEEs have dashboards using them) with maybe an extra tag to differntiate the source. Is there a particular reason you didn't go that way (I'm fine with it, just want to understand)?
datadog.tracer.queue.dropped_traces
datadog.tracer.queue.dropped_spans
datadog.tracer.queue.enqueued_traces
datadog.tracer.queue.enqueued_spans
datadog.tracer.queue.dequeued_traces
datadog.tracer.queue.dequeued_spans
datadog.tracer.api.requests
datadog.tracer.api.responses
datadog.tracer.api.errors

ajgajg1134: I had reviewed the original list and there is definitely some overlap of the metrics here but a number of these existing .NET metrics aren't common to other tracers and aren't relevant for some of the functionality being measured here (for example, the `queue` metrics don't make a ton of sense since libdatadog doesn't have a "queue" of traces / spans). 

Along that line it felt relevant that libdatadog itself isn't really a "tracer" so instead of referring to the metrics as `tracer` metrics it felt more appropriate to call out that it's coming from `libdatadog`. This also helps segment the health metrics from health metrics that the tracer itself is emitting, since I imagine that there will be metrics  coming from both simultaneously which will help in debugging if an issue exists in the tracer, or within libdatadog. (e.g. the `queue` metrics from .NET, or `datadog.tracer.spans_started` from the go tracer)

data-pipeline/src/trace_exporter.rs
ajgajg1134: Yep! By default there's no dogstatsd flusher, but when building the trace exporter if one is provided it enables health metrics

data-pipeline/src/trace_exporter.rs
ajgajg1134: You're right that this can be moved into the `if` below (and I'll do so for clarity!)

From a performance perspective I'm not sure of the full effect, the variable itself can live on the stack (which will be needed no matter what), But there's likely a minor improvement here where we don't need to do a `match` and allocate an `Either` or an `iter::Chain`

sidecar/Cargo.toml
```diff
@@ -26,6 +26,7 @@ datadog-trace-utils = { path = "../trace-utils" }
 datadog-trace-normalization = { path = "../trace-normalization" }
 datadog-remote-config = { path = "../remote-config" }
 datadog-crashtracker = { path = "../crashtracker" }
+dogstatsd-client = { path = "../dogstatsd-client" }
```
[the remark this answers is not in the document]
ajgajg1134: dogstatsd crate implements a dogstatsd server, this is just a dogstatsd-client to communicate with a dogstatsd server. To me it makes sense to have separate crates as it should be rare for a project to need BOTH a client and server (and this allows users to only import the one they need).

dogstatsd-client/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
```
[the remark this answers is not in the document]
ajgajg1134: yes, I'll add a readme and description good catch!

sidecar/src/service/sidecar_server.rs
```diff
@@ -897,9 +897,10 @@ impl SidecarInterface for SidecarServer {
         session.modify_trace_config(|cfg| {
             update_cfg(cfg.endpoint.take(), |e| cfg.set_endpoint(e), &token);
         });
-        session.configure_dogstatsd(|cfg| {
-            update_cfg(cfg.endpoint.take(), |e| cfg.set_endpoint(e), &token);
-        });
+        // TODO: the dogstatsd-client doesn't support test_session tokens yet
```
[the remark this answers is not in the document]
ajgajg1134: added!

data-pipeline/src/trace_exporter.rs
```diff
@@ -108,6 +118,11 @@ impl<'a> From<&'a TracerTags> for HashMap<&'static str, String> {
     }
 }
 
+enum HealthMetric {
+    Count(&'static str, i64),
+    //TODO: Add more DogStatsDAction as we need them
```
ekump, line 123: nit: This TODO comment seems unnecessary? When the time is right to add to the enum with new DogStatsDActions someone will just do it. There is no tech-debt here to address later.

data-pipeline/src/trace_exporter.rs
pierotibou: > aren't common to other tracers
:sadpanda: Betrayal from the .NET tracer :p
Ok thanks for the explanation. Could have been good to have this detailed in the PR description (unless you had mentionned it at standup actually, in case my bad, and 🤐 :p 

If you've started listing discrepencies, should we log somewhere the various metrics emitted per language? 🤔  I mean, if we say we want to improve the OOTB dasbhoard to list tracer emitted metrics, we could be adding metrics from all languages in one graph maybe? Could be part of the other story for the OOTB dasboard improvements.

pierotibou (commented):
LGTM from a functional standpoint following @ajgajg1134 's answers.

data-pipeline/src/trace_exporter.rs
```diff
@@ -38,6 +41,13 @@ pub enum TraceExporterOutputFormat {
     V07,
 }
 
+// internal health metrics
+const STAT_SEND_TRACES: &str = "datadog.libdatadog.send.traces";
```
ekump, line 45: These consts could probably be put into either an Enum or a separate module named `HealthMetrics` to better define what they represent. As it is right now, it's unclear for someone looking at the data-pipeline for the first time what they mean.

data-pipeline/src/trace_exporter.rs
```diff
@@ -117,6 +132,9 @@ pub struct TraceExporter {
     // TODO - do something with the response callback - https://datadoghq.atlassian.net/browse/APMSP-1019
     _response_callback: Option<Box<dyn ResponseCallback>>,
     runtime: Runtime,
+    // None if dogstatsd is disabled
```
ekump, line 135: This should probably be a rustdoc comment instead.

data-pipeline/src/trace_exporter.rs
```diff
@@ -175,20 +193,40 @@ impl TraceExporter {
                     .await
                 {
                     Ok(response) => {
-                        if response.status() != 200 {
+                        let response_status = response.status();
+                        if response_status != http::StatusCode::OK {
```
ekump, line 197: you could do `if !response_status.is_success()`

data-pipeline/src/trace_exporter.rs
```diff
@@ -175,20 +193,40 @@ impl TraceExporter {
...
                             let body_bytes = hyper::body::to_bytes(response.into_body()).await?;
                             let response_body =
                                 String::from_utf8(body_bytes.to_vec()).unwrap_or_default();
+                            self.emit_metric(
+                                HealthMetric::Count(STAT_SEND_TRACES_ERRORS, 1),
+                                Some(vec![
+                                    &Tag::new("response_code", response_status.as_str()).unwrap()
```
ekump, line 204: Are we confident that we'll never panic on this unwrap?

bantonsson (approved):

dogstatsd-client/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+    /// and PHP implementation uses float or string (https://github.com/DataDog/php-datadogstatsd/blob/0efdd1c38f6d3dd407efbb899ad1fd2e5cd18085/src/DogStatsd.php#L251)
+    Set(T, i64, V),
+}
+
+/// A dogstatsd-client that flushes stats to a given endpoint.
+/// The default value has no address and is thus disabled, use `new_flusher` or `set_endpoint` to
+/// configure an endpoint.
+#[derive(Default)]
+pub struct Flusher {
+    client: Option<StatsdClient>,
```
VianneyRuhlmann, line 75: Isn't the Option to represent a disabled stats redundant with the Option used on the Flusher in `trace_exporter`. You could force `client` to be defined and let users of the crate use an Option if they want to disable stats.

data-pipeline/src/trace_exporter.rs
```diff
@@ -197,13 +235,29 @@ impl TraceExporter {
             })
     }
 
+    /// Emit a health metric to dogstatsd
+    fn emit_metric(&self, metric: HealthMetric, custom_tags: Option<Vec<&Tag>>) {
```
VianneyRuhlmann, line 239: Not sure if it is worth it performance wise. But using a closure to lazily create the metric and tags can avoid computation and allocation when stats are disabled.

VianneyRuhlmann (approved):
LGTM just a few suggestions on design alternatives

Cargo.toml
```diff
@@ -34,6 +34,7 @@ members = [
   "data-pipeline-ffi",
   "ddsketch",
   "tinybytes",
+  "dogstatsd-client",
```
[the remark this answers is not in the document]
bantonsson: Nice catch! This is exactly why I was afraid of adding this _duplication_ to the project.

data-pipeline/src/trace_exporter.rs
morrisonlevi: Docs say `response_status`'s `as_str` is a numeric string:

> The return value only includes a numerical representation of the status code. The canonical reason is not included.

And so yes, it won't panic, since `response_code:{response_status}` is going to be a valid tag name.

ajgajg1134: I'm like 99% confident since this can only really fail if we generate an invalid tag (so like an empty tag, or end with a `:`). The Status code itself is just a `NonZeroU16` so I think functionally speaking, it should be impossible for this to panic.

But since I'm not 100% sure here I'll just add a quick little check and an error log if it ever happens :P

Cargo.toml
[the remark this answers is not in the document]
bantonsson: I think we can remove the `default-members` #644

dogstatsd-client/src/lib.rs
ajgajg1134: Good idea, it required some changes to the sidecar to support it but I think it's a bit clearer that the client could be disabled

data-pipeline/src/trace_exporter.rs
ekump: @morrisonlevi That makes sense, thanks!

sidecar/src/service/sidecar_server.rs
```diff
@@ -901,9 +901,10 @@ impl SidecarInterface for SidecarServer {
         session.modify_trace_config(|cfg| {
             update_cfg(cfg.endpoint.take(), |e| cfg.set_endpoint(e), &token);
         });
-        session.configure_dogstatsd(|cfg| {
-            update_cfg(cfg.endpoint.take(), |e| cfg.set_endpoint(e), &token);
-        });
+        // TODO(APMSP-1377): the dogstatsd-client doesn't support test_session tokens yet
+        // session.configure_dogstatsd(|cfg| {
+        //     update_cfg(cfg.endpoint.take(), |e| cfg.set_endpoint(e), &token);
+        // });
```
iamluc, line 907: Do you plan to fix this before merging? 
Otherwise, it will prevent dd-trace-php from upgrading, as it will break our tests.

dogstatsd-client/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+    #[allow(missing_docs)]
+    Histogram(T, f64, V),
+    /// Cadence only support i64 type as value
+    /// but Golang implementation uses string (https://github.com/DataDog/datadog-go/blob/331d24832f7eac97b091efd696278fe2c4192b29/statsd/statsd.go#L230)
+    /// and PHP implementation uses float or string (https://github.com/DataDog/php-datadogstatsd/blob/0efdd1c38f6d3dd407efbb899ad1fd2e5cd18085/src/DogStatsd.php#L251)
+    Set(T, i64, V),
+}
+
+/// A dogstatsd-client that flushes stats to a given endpoint. Use `new_flusher` to build one.
+pub struct Flusher {
```
iamluc, line 71: Nit: Should we take the opportunity to rename it to `Client` or something similar?

sidecar/src/service/sidecar_server.rs
ajgajg1134: I commented this out because it was always broken and as part of migrating I didn't want to _also_ add new functionality. Maybe I misunderstood how it works though so is there a way I can test those tests with this branch before merging?

iamluc: What do you mean by "it was always broken"? :thinking: 

But anyway, you're right it seems unused.
I thought it was used to configure the test_session_token that should be added as a tag to the metric, but actually it's done [on dd-trace-php's side](https://github.com/DataDog/dd-trace-php/pull/2802/files#diff-643c589592ade49e8f68a3c034cfddb87ee2fb6a79452207062e173e3634b8e2R205-R207).

data-pipeline/src/trace_exporter.rs
ekump: How likely are metrics to be disabled? If there is a decent chance we definitely should avoid any extra computation and allocation.

data-pipeline/Cargo.toml
```diff
@@ -16,14 +16,16 @@ hyper = {version = "0.14", features = ["client"], default-features = false}
 log = "0.4"
 rmp-serde = "1.1.1"
 bytes = "1.4"
+either = "1.13.0"
```
ekump, line 19: non-blocking: We can change this later, but is this really worth the overhead of adding a new crate (honest question)? Seems like just creating an Enum with two types would be sufficient? But perhaps there is some nuance I'm missing.

ekump (approved):
My only remaining concern is the one @VianneyRuhlmann raised. We shouldn't be calculating metrics at all if the functionality is disabled...assuming there may be a non-trivial amount of situations where metrics are disabled. But that can be addressed by a future PR.

dogstatsd-client/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+    let mut tag_opt = tags_iter.next();
+    while tag_opt.is_some() {
+        builder = builder.with_tag_value(tag_opt.unwrap().as_ref());
+        tag_opt = tags_iter.next();
+    }
+    builder.try_send()?;
+    Ok(())
+}
+
+fn create_client(endpoint: &Endpoint) -> anyhow::Result<StatsdClient> {
```
hoolioh, line 180: Nit: as the functions has multiple points for returning an error and these errors can bubble up in the stack frame I would suggest to add more context creating some explicit errors with `anyhow` so it's easier to debug which method failed.

hoolioh (approved):

data-pipeline/Cargo.toml
ajgajg1134: Unfortunately it's not _just_ an enum with two types, it's also an enum with two types with some additional code to support "IntoIterator". For context, the reason I import this is that `tags` can be two different types depending on if some additional custom tags are provided, I don't expect `Either` (or any different "custom" enum we provide instead) to be used outside this context.

There's a couple different options here we can take but this seemed to be a pretty easy option to follow:
1. Keep it as is
2. Implement our own "Either" enum here including implementing [into_iter](https://docs.rs/either/latest/src/either/lib.rs.html#537) on it like the Either crate does
3. Remove the local variable and have some code duplication so no enum wrapping each type is required.

data-pipeline/src/trace_exporter.rs
```diff
@@ -298,6 +345,21 @@ impl TraceExporter {
...
+        if let Some(flusher) = &self.dogstatsd {
+            let tags = match custom_tags {
+                None => Either::Left(&self.common_stats_tags),
+                Some(custom) => Either::Right(self.common_stats_tags.iter().chain(custom)),
+            };
```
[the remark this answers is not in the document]
ajgajg1134: Yeah I definitely don't _love_ having to use it here but I couldn't find a way around it. I'm also not a huge fan of a more descriptive type since that type would only be used in exactly this one place, and would only exist to get around the type system so I can have `tags` be something that implements "Into Iterator" and avoid needing dynamic dispatch. (more context here: https://github.com/DataDog/libdatadog/pull/638#discussion_r1773354431)

data-pipeline/src/trace_exporter.rs
ajgajg1134: I'm not sure I follow how one would do this / where the performance benefit would be? If the client is `None` then this function does nothing and just returns? 🤔 (or do you mean allocating the `HealthMetric` type?)

data-pipeline/Cargo.toml
bantonsson: I think that we should keep using the crate. Maybe I'm completely corrupted by my Scala days, but I think every language should have an `Either` type 😉

data-pipeline/src/trace_exporter.rs
VianneyRuhlmann: Yes the HealthMetric and Vec still need to be allocated. Another option would be to use a macro to add a `if flusher.is_some() {...}` everywhere we create metrics

dogstatsd-client/src/lib.rs
ajgajg1134: good call out, I've just added additional `anyhow` in the places it wasn't used before!

PR #645, opened 2024-09-23
Fix rustfmt error

VianneyRuhlmann:
# What does this PR do?
The new nightly version of rustfmt changed some formatting rule

# Motivation

# Additional Notes

# How to test the change?


bantonsson (approved):
Maybe not for this PR, but I think that we should lock down the nightly version, e.g. use `nightly-2024-09-22`, to avoid these kinds of unexpected breakage.

PR #647, opened 2024-09-23
Implement deadlock watchdog for sidecar

bwoebi:
If the sidecar hangs, it can cause every process trying to write to it to be much slower.

Better kill it once in doubt.

### DIFF
--- a/sidecar/src/watchdog.rs
+++ b/sidecar/src/watchdog.rs
@@ -49,21 +49,23 @@ impl Watchdog {
         let still_alive = Arc::new(AtomicU32::new(0));
         let still_alive_thread = still_alive.clone();
 
+        const SHUTDOWN: u32 = u32::MAX;
+
         let interval = self.interval.period();
         std::thread::spawn(move || {
-            let mut repeats = false;
+            let mut maybe_stuck = false;
             let mut last = 0;
             loop {
                 std::thread::sleep(interval);
                 let current = still_alive_thread.load(Ordering::Relaxed);
                 if last != current {
-                    if current == u32::MAX {
+                    if current == SHUTDOWN {
                         return;
                     }
                     last = current;
-                    repeats = false;
+                    maybe_stuck = false;
                 } else {
-                    if repeats {
+                    if maybe_stuck {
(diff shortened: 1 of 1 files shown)


cataphract (approved):

sidecar/src/watchdog.rs
```diff
@@ -46,12 +46,44 @@ impl Watchdog {
         let mem_usage_bytes = Arc::new(AtomicUsize::new(0));
         let handle_mem_usage_bytes = mem_usage_bytes.clone();
 
+        let still_alive = Arc::new(AtomicU32::new(0));
+        let still_alive_thread = still_alive.clone();
+
+        let interval = self.interval.period();
+        std::thread::spawn(move || {
+            let mut repeats = false;
```
bantonsson, line 54: I can see what the code does and this variable is named after that, but maybe call it `stuck` or `maybe_stuck`?

sidecar/src/watchdog.rs
```diff
@@ -46,12 +46,44 @@ impl Watchdog {
...
+            let mut last = 0;
+            loop {
+                std::thread::sleep(interval);
+                let current = still_alive_thread.load(Ordering::Relaxed);
+                if last != current {
+                    if current == u32::MAX {
```
bantonsson, line 60: A comment about this constant or naming it `SHUTDOWN` would make things clearer.

bantonsson (approved):

sidecar/src/watchdog.rs
```diff
@@ -46,12 +46,44 @@ impl Watchdog {
...
+                        return;
+                    }
+                    last = current;
+                    repeats = false;
+                } else {
+                    if repeats {
+                        std::thread::spawn(move || {
```
pierotibou, line 67: Should we have a telemety metric as well?

bwoebi: If it's stuck, it'll be unable to flush that one though. So emitting it would be non-trivial...

pierotibou: Ok. Thanks

PR #651, opened 2024-09-26
[APMSP-1391] Wait for spawned children

hoolioh:
# What does this PR do?

Actively waiting for spawned children in the builder in order to avoid concurrency problems.

# Motivation

Letting child process on their own could lead to issues if the parent process finish before them or another subsequent command is added that operates on the same resources.


ekump (approved):
LGTM - But can you just associate a JIRA ticket to this PR before merging?

gleocadie (approved):
LGTM

PR #654, opened 2024-09-27
Fix dogstatsd aggregator

alexgallotta:
- allow creation of Metric outside this crate, setting id public and with constructor
- make it easier to generate metrics


dogstatsd/src/metric.rs
```diff
@@ -148,9 +176,9 @@ pub fn parse(input: &str) -> Result<Metric, ParseError> {
 
             let tags;
             if let Some(tags_section) = caps.name("tags") {
-                tags = SortedTags::parse(tags_section.as_str())?;
+                tags = Some(SortedTags::parse(tags_section.as_str())?);
             } else {
-                tags = EMPTY_TAGS;
+                tags = None;
```
astuyve, line 181: nit: you could initialize tags as an Option with None here on 177, no?

astuyve (approved):

dogstatsd/src/metric.rs
astuyve: fair enough

dogstatsd/src/aggregator.rs
```diff
@@ -19,6 +19,7 @@ use ustr::Ustr;
 impl MetricValue {
     fn aggregate(&mut self, metric: Metric) {
         // safe because we know there's at least one value when we parse
+        // TODO aggregating different types should return error
```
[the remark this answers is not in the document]
alexgallotta: No, it's an implementation details. It's already guarded in the code but this is could be a better place. We are not tracking in Jira all minor improvements to avoid being slowed down

alexgallotta:
> I don't get what the fix is since there is no description about what was wrong, how it's fixed etc.
> 
> Adding a test that ensures that whatever happened, doesn't happen again would be good as well.

Good point, I added it in the description

bantonsson (approved):

PR #656, opened 2024-09-30
Use info config for stats computation

VianneyRuhlmann:
# What does this PR do?
Fetch the info endpoint from the trace exporter to use configuration from the agent for client-side stats. The trace exporter will also check that the agent support client side stats before enabling it.

# Motivation

The agent info contains essential configuration for peer tags aggregation and span kinds eligibility to stats.

# Additional Notes
## Mutability concerns
Since we check the agent info on each call to `send` to update `StatsComputationStatus` I used an ArcSwap (which we already use for the info struct) to wrap the stats computation status. This allows sending traces without holding a mutable reference. ArcSwap is optimized for read-mostly scenarios which is the case here. This should reduce the impact of having to lock the status on every call to `send` to check if stats should be computed.

If you have any thoughts about this solution please let me know, fyi here's the other solutions I considered
### Other solutions considered
- Let `send` take a &mut and use a Mutex on the whole `TraceExporter` this is not viable as send block on the http calls which means holding the lock for a long time
- Using a `Mutex` or a `RwLock` a bit expensive considering it would be on the critical path of send

# How to test the change?


data-pipeline/src/span_concentrator/mod.rs
```diff
@@ -69,10 +68,8 @@ pub struct SpanConcentrator {
     oldest_timestamp: u64,
     /// bufferLen is the number stats bucket we keep when flushing.
     buffer_len: usize,
-    /// flag to enable aggregation of peer tags
-    peer_tags_aggregation: bool,
-    /// flag to enable computation of stats through checking the span.kind field
-    compute_stats_by_span_kind: bool,
+    /// span.kind field eligible for stats computation
```
ajgajg1134, line 71:
```suggestion
    /// span.kind fields eligible for stats computation
```

data-pipeline/src/trace_exporter.rs
```diff
@@ -214,32 +253,173 @@ impl TraceExporter {
...
+                    let _ = exporter_handle.await;
+                }
+            });
+            Ok(())
+        }
+    }
+
+    /// Start the stats exporter and enable stats computation
+    ///
+    /// Should only be used if the agent enabled stats computation
```
ajgajg1134, line 300: Should we add a note here that this won't work if stats computation is disabled at the client level?

data-pipeline/src/trace_exporter.rs
```diff
@@ -214,32 +253,173 @@ impl TraceExporter {
...
+            self.client_side_stats
+                .store(Arc::new(StatsComputationStatus::DisabledByAgent {
+                    bucket_size,
+                }));
+        }
+    }
+
+    /// Check for a new state of agent_info and update the trace exporter if needed
+    fn check_agent_info(&self) {
+        dbg!(self.agent_info.load().clone());
```
ajgajg1134, line 363: do you want to keep this `dbg!` in?

data-pipeline/src/trace_exporter.rs
```diff
@@ -214,32 +253,173 @@ impl TraceExporter {
...
-                        Err(_) => Err("Shutdown timed out".to_string()),
-                    }
-                } else {
-                    self.runtime.block_on(async {
-                        cancelation_token.cancel();
+        if let Some(timeout) = timeout {
+            match self.runtime.block_on(async {
+                tokio::time::timeout(timeout, async {
+                    let stats_status: Option<StatsComputationStatus> =
+                        std::sync::Arc::<StatsComputationStatus>::into_inner(
```
ajgajg1134, line 260:
```suggestion
                       Arc::<StatsComputationStatus>::into_inner(
```

data-pipeline/src/trace_exporter.rs
```diff
@@ -214,32 +253,173 @@ impl TraceExporter {
...
+                })
+                .await
+            }) {
+                Ok(()) => Ok(()),
+                Err(_) => Err("Shutdown timed out".to_string()),
+            }
+        } else {
+            self.runtime.block_on(async {
+                let stats_status: Option<StatsComputationStatus> =
+                    std::sync::Arc::<StatsComputationStatus>::into_inner(
```
ajgajg1134, line 281:
```suggestion
                   Arc::<StatsComputationStatus>::into_inner(
```

ajgajg1134 (commented):
Overall this makes sense to me, just a few small clean up comments (probably want another rustacean to take a peek at this before merging though).

Some of the syntax here is a bit silly looking, but I think it's just kinda "required" from what I can tell (like `&**self.client_side_stats.load()`). Maybe this is just me still getting used to reading rust syntax again :P

data-pipeline/src/trace_exporter.rs
```diff
@@ -1006,15 +1200,20 @@ mod tests {
...
         }];
 
-        trace_utils::compute_top_level_span(&mut trace_chunk);
-
         let data = rmp_serde::to_vec_named(&vec![trace_chunk]).unwrap();
 
+        // Wait for the info fetcher to get the config
+        while mock_info.hits() == 0 {
+            exporter.runtime.block_on(async {
+                sleep(Duration::from_millis(100)).await;
```
ekump, line 1213: since this is in a loop, would it make sense to lower the duration?

data-pipeline/src/trace_exporter.rs
```diff
@@ -17,13 +20,17 @@ use dogstatsd_client::{new_flusher, Client, DogStatsDAction};
...
+use log::{error, info};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use std::{borrow::Borrow, collections::HashMap, str::FromStr, time};
 use tokio::{runtime::Runtime, task::JoinHandle};
 use tokio_util::sync::CancellationToken;
 
+const DEFAULT_STATS_ELIGIBLE_SPAN_KINDS: [&str; 4] = ["client", "server", "producer", "consumer"];
+const STATS_ENDPOINT: &str = "/v0.6/stats";
+const INFO_ENDPOINT: &str = "/info";
```
ekump, line 32: I wonder if we're not properly encapsulating the stats computation logic? It's not obvious to me why all of these changes should live in the trace exporter. These consts , the `StatsComputationStatus` enum, and the stats specific start/shutdown logic are all examples.

data-pipeline/src/trace_exporter.rs
```diff
@@ -214,32 +253,173 @@ impl TraceExporter {
...
+            let bucket_size = stats_concentrator.lock().unwrap().get_bucket_size();
+            self.client_side_stats
+                .store(Arc::new(StatsComputationStatus::DisabledByAgent {
+                    bucket_size,
+                }));
+        }
+    }
+
+    /// Check for a new state of agent_info and update the trace exporter if needed
+    fn check_agent_info(&self) {
```
ekump, line 362: Is it possible that this (and other large functions as part of this PR) be broken up into smaller ones to make them easier to read and maintain?

data-pipeline/src/trace_exporter.rs
VianneyRuhlmann: I tried a lower duration with the info fetcher but this sometimes failed probably due to a race condition where the loop ended because the server received the request but the response had not yet been received by the client.

data-pipeline/src/trace_exporter.rs
```diff
@@ -214,32 +253,173 @@ impl TraceExporter {
...
+                                agent_info
+                                    .info
+                                    .span_kinds_stats_computed
+                                    .clone()
+                                    .unwrap_or_default(),
+                            );
+                            concentrator.set_peer_tags(
+                                agent_info.info.peer_tags.clone().unwrap_or_else(|| {
+                                    DEFAULT_STATS_ELIGIBLE_SPAN_KINDS.map(String::from).to_vec()
+                                }),
```
bantonsson, line 412: So why does this code differ from the code above, where the span kinds use this default, but here it's added to the peer tags 🤷🏼‍♂️

bantonsson (approved):
LGTM. Only one weird thing with the peer tag default fallback.

data-pipeline/src/trace_exporter.rs
VianneyRuhlmann: I agree, yet this would probably require a larger refactor. I'll try to improve it without refactoring too much.

knusbaum (approved):

ekump (approved):
LGTM. As discussed IRL we're going to consider a bigger refactor of the trace exporter and don't want to block on that right now.

data-pipeline/src/trace_exporter.rs
VianneyRuhlmann: I think it's fine if the the client didn't enable stats computation there's no point for the exporter to start it and since it is private I don't think there's a risk of misusing it

data-pipeline/src/trace_exporter.rs
VianneyRuhlmann: That's just me going crazy, I fixed it.

PR #657, opened 2024-09-30
Support for remote config in the appsec helper

cataphract:
Adds some ASM constants, some functions to work with RC shared memory paths (rather than only ConfigInvariants and Targets, which are more difficult to pass around), and an in-memory notification mechanism.

In support of https://github.com/DataDog/dd-trace-php/pull/2864

### DIFF
--- a/sidecar/src/shm_remote_config.rs
+++ b/sidecar/src/shm_remote_config.rs
@@ -21,7 +21,7 @@ use std::cmp::Reverse;
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::default::Default;
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::hash::{Hash, Hasher};
 use std::io;
 #[cfg(windows)]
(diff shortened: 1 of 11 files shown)


sidecar/src/shm_remote_config.rs
```diff
@@ -37,6 +37,26 @@ use zwohash::ZwoHasher;
 pub struct RemoteConfigWriter(OneWayShmWriter<NamedShmHandle>);
 pub struct RemoteConfigReader(OneWayShmReader<NamedShmHandle, CString>);
 
+/// Function to dump the invariants and target, and the corresponding path.
+/// This is useful for debugging purposes.
+///
+/// # Safety
+/// Pointers should be valid and non-null.
+#[no_mangle]
+pub unsafe extern "C" fn debug_dump_inv_tar(
```
bwoebi, line 46: Does anything prevent you from putting the ffi functions into actual sidecar-ffi crate?

bwoebi: Sounds fine, thanks.

bwoebi (approved):
The changes are simple enough and look fine to me :-)

PR #658, opened 2024-09-30
Support http-proxy for trace agent, remove proxy from dsd

astuyve:
# What does this PR do?
reqwest supports system proxy by default, no code changes needed (just `HTTPS_PROXY` set).
hyper client doesn't seem to, but this crate works and I'm able to see the proxied traffic logged and traces show up

squid proxy:
image
Traces:
image

# Motivation
Support folks in isolated VPCs without privatelink access.

# Additional Notes

# How to test the change?


serverless/src/main.rs
```diff
@@ -30,8 +30,6 @@ pub async fn main() {
     Builder::from_env(env).target(Target::Stdout).init();
 
     let dd_api_key: Option<String> = env::var("DD_API_KEY").ok();
-    let dd_http_proxy: Option<String> = env::var("DD_HTTP_PROXY").ok();
```
astuyve, line 33: This was always wrong. `DD_HTTP(S)_PROXY` is used for proxying requests made by the agent for checks or integrations. To proxy the agents connection to Datadog, we've required `HTTPS_PROXY` be set.

bwoebi:
The hyper-proxy dependency probably should be either behind a feature flag, or we support proxying in general, at which point it likely should become part of ddcommon.

astuyve:
These tests are very slow in libdatadog:
image

But when they were in the Lambda extension repo, they ran in thousandths of a second:
image
https://github.com/DataDog/datadog-lambda-extension/actions/runs/10458673752/job/28961015054#step:9:620
Anyone know why?

bwoebi:
@astuyve These are the miri tests. Miri does things and is slow by design, and some codes can make it even slower... They are still fast in e.g. `Test / cargo test --workspace #ubuntu-latest`.

astuyve:
Thanks @bwoebi!! Disabling Miri in another PR: https://github.com/DataDog/libdatadog/pull/661

morrisonlevi (approved):

PR #664, opened 2024-10-03
Fix the trace payload size hint

bwoebi:
The incoming buffer may be much larger than the actual contained data. The decoder knows where the payload starts and ends, but the sidecar send_trace_v04 method does not.

This ensures proper trace coalescing, leading to less individual http requests.

### DIFF
--- a/trace-utils/src/msgpack_decoder/v04/decoder/mod.rs
+++ b/trace-utils/src/msgpack_decoder/v04/decoder/mod.rs
@@ -47,49 +47,56 @@ use tinybytes::{Bytes, BytesString};
 /// };
 /// let encoded_data = to_vec_named(&vec![vec![span]]).unwrap();
 /// let encoded_data_as_tinybytes = tinybytes::Bytes::from(encoded_data);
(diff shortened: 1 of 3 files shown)


trace-utils/src/msgpack_decoder/v04/decoder/mod.rs
```diff
@@ -47,49 +47,56 @@ use tinybytes::{Bytes, BytesString};
 /// };
 /// let encoded_data = to_vec_named(&vec![vec![span]]).unwrap();
 /// let encoded_data_as_tinybytes = tinybytes::Bytes::from(encoded_data);
-/// let decoded_traces = from_slice(encoded_data_as_tinybytes).expect("Decoding failed");
+/// let (decoded_traces, _) = from_slice(encoded_data_as_tinybytes).expect("Decoding failed");
```
ekump, line 50:
```suggestion
/// let (decoded_traces, _payload_size) = from_slice(encoded_data_as_tinybytes).expect("Decoding failed");
```
I think it's helpful to at least name the variable, even if we don't use it in the example.

PR #667, opened 2024-10-07
Refactor proxy handling.

hoolioh:
# What does this PR do?

This PR tries to clean a bit the proxy handling by creating a new function to send requests through a proxy passed as a parameter rather than holding the proxy url in each SendData structure.

# Motivation

Decouple SendData from proxy handling so future planned work on SendData will be easier.


ekump (approved):
much cleaner, thanks for doing this!

PR #672, opened 2024-10-11
[Serverless Mini Agent] Add Span Tags for Azure Spring Apps

duncanpharvey:
# What does this PR do?

Adds `azurespringapp.name` and `azurespringapp.hostname` span tags to spans from Azure Spring apps.

# Motivation

Add span tags for Azure Spring apps to facilitate billing.

https://datadoghq.atlassian.net/browse/SVLS-5643

# Additional Notes

Remove `functionname` span tag for Azure Spring apps.

# How to test the change?

See https://datadoghq.atlassian.net/wiki/spaces/SLS/pages/2977497119/Serverless+Mini+Agent#Testing

### DIFF
--- a/trace-mini-agent/src/env_verifier.rs
+++ b/trace-mini-agent/src/env_verifier.rs
@@ -5,6 +5,7 @@ use async_trait::async_trait;
 use hyper::{Body, Client, Method, Request, Response};
 use log::{debug, error};
 use serde::{Deserialize, Serialize};
+use std::env;
 use std::fs;
 use std::path::Path;
 use std::process;
@@ -113,6 +114,9 @@ impl ServerlessEnvVerifier {
             }
         };
         trace_utils::MiniAgentMetadata {
+            azure_spring_app_hostname: trace_utils::MiniAgentMetadata::default()
+                .azure_spring_app_hostname,
+            azure_spring_app_name: trace_utils::MiniAgentMetadata::default().azure_spring_app_name,
             gcp_project_id: Some(gcp_metadata.project.project_id),
             gcp_region: Some(get_region_from_gcp_region_string(
                 gcp_metadata.instance.region,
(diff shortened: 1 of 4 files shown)


trace-mini-agent/src/env_verifier.rs
```diff
@@ -140,9 +144,13 @@ impl EnvVerifier for ServerlessEnvVerifier {
...
                     .await;
             }
-            trace_utils::EnvironmentType::AzureSpringApp => {
-                trace_utils::MiniAgentMetadata::default()
-            }
+            trace_utils::EnvironmentType::AzureSpringApp => trace_utils::MiniAgentMetadata {
+                azure_spring_app_hostname: env::var("HOSTNAME").ok(),
+                azure_spring_app_name: env::var("ASCSVCRT_SPRING__APPLICATION__NAME").ok(),
+                gcp_project_id: trace_utils::MiniAgentMetadata::default().gcp_project_id,
+                gcp_region: trace_utils::MiniAgentMetadata::default().gcp_region,
```
jcstorms1, line 151: Is it correct for GCP be part of `AzureSpringApp`? Not sure if I'm missing context.

trace-mini-agent/src/trace_processor.rs
```diff
@@ -41,12 +41,12 @@ impl TraceChunkProcessor for ChunkProcessor {
...
         trace_utils::set_serverless_root_span_tags(
             &mut chunk.spans[root_span_index],
-            self.config.function_name.clone(),
+            self.config.app_name.clone(),
             &self.config.env_type,
         );
         for span in chunk.spans.iter_mut() {
             trace_utils::enrich_span_with_mini_agent_metadata(span, &self.mini_agent_metadata);
-            trace_utils::enrich_span_with_azure_metadata(span);
+            trace_utils::enrich_span_with_azure_function_metadata(span);
```
jcstorms1, line 49: I'm a little confused why were are changing everything to app name but still using function metadata? Is this because springs apps is built on top of functions?

duncanpharvey: Originally the mini agent was written just for functions. So the name of the traced service was always a function, hence the property `function_name` in the mini agent config.

Now that Azure Spring Apps, which are unrelated to functions, are also traced with the mini agent I wanted to rename this property to something more generic. I figured `app_name` was generic enough and could represent either a function or a spring app. Let me know if you have another idea for this property name!

In the code you highlighted I just renamed the method to be clear it's only being used for Azure Functions since the old name, `enrich_span_with_azure_metadata` sounded like it could be used for Azure Spring Apps (and it is not used for those).

trace-mini-agent/src/env_verifier.rs
duncanpharvey: It's correct for how this was originally designed but I agree it's confusing to be setting properties that are ultimately not used in some environments.

This `MiniAgentMetadata` struct previously had properties for GCP related fields and the mini agent version. I added the Azure Spring properties to it since that was the most convenient place to add this metadata. So we do need to initialize those GCP properties in the Azure Spring App and Azure Function environments even though they're not actually added to any span tags.

I'm planning on doing a refactor once Google Cloud Functions are supported with the sidecar agent and we can remove all of the GCP related logic in the mini agent.

https://github.com/DataDog/libdatadog/blob/8266d3801119ff45e0678b6cef6f8b9669adebe9/trace-utils/src/trace_utils.rs#L480-L486

jcstorms1 (approved):

PR #675, opened 2024-10-14
[NO-TICKET] Upgrade "datadog-protos"/"ddketch-agent" to unlock tonic 0.12.3

ivoanjo:
# What does this PR do?

This PR upgrades the libdatadog dependencies coming in from the "saluki-backport" repository: datadog-protos and ddsketch-agent.

These unlock the upgrade of tonic to 0.12.3 which will solve the following security advisory:
https://github.com/DataDog/libdatadog/security/dependabot/22 .

# Motivation

Get rid of outdated dependencies.

# Additional Notes

The upgrade of stuff from "saluki-backport" brings other changes; I'll sync with @duncanpharvey to make sure this is good to go.

# How to test the change?

Not sure -- hopefully existing test coverage is enough? I'll sync on this with @duncanpharvey as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -540,7 +540,7 @@ dependencies = [
  "bitflags 2.5.0",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
+ "itertools 0.10.5",
  "lazy_static",
  "lazycell",
  "log",
@@ -1577,13 +1577,13 @@ dependencies = [
 [[package]]
 name = "datadog-protos"
 version = "0.1.0"
-source = "git+https://github.com/DataDog/saluki-backport/#a3e224b2da67f5e4a65e33703aaaa469a3bd3924"
+source = "git+https://github.com/DataDog/saluki-backport/#3c5d87ab82dea4a1a98ef0c60fb3659ca35c2751"
 dependencies = [
  "bytes",
  "prost 0.13.1",
  "protobuf",
  "protobuf-codegen",
- "tonic 0.12.1",
+ "tonic 0.12.3",
  "tonic-build",
 ]
 
@@ -1864,7 +1864,7 @@ dependencies = [
 [[package]]
 name = "ddsketch-agent"
 version = "0.1.0"
-source = "git+https://github.com/DataDog/saluki-backport/#a3e224b2da67f5e4a65e33703aaaa469a3bd3924"
+source = "git+https://github.com/DataDog/saluki-backport/#3c5d87ab82dea4a1a98ef0c60fb3659ca35c2751"
 dependencies = [
  "datadog-protos",
  "float_eq",
@@ -2741,7 +2741,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.5.6",
+ "socket2 0.4.10",
  "tokio",
  "tower-service",
  "tracing",
@@ -3036,15 +3036,6 @@ name = "itertools"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b1c173a5686ce8bfa551b3563d0c2170bf24ca44da99c7ca4bfdab5418c3fe57"
-dependencies = [
- "either",
-]
-
-[[package]]
-name = "itertools"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ba291022dbbd398a455acf126c1e341954079855bc60dfdda641363bd6922569"
 dependencies = [
  "either",
 ]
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -540,7 +540,7 @@ dependencies = [
  "bitflags 2.5.0",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
+ "itertools 0.10.5",
```
bantonsson, line 543: Interesting downgrade...

Cargo.lock
```diff
@@ -2741,7 +2741,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.5.6",
+ "socket2 0.4.10",
```
bantonsson, line 2744: Another downgrade...

Cargo.lock
ivoanjo: Hmmm interesting... It's not entirely a downgrade. Specifically, we already had three versions in our `Cargo.lock` before:

```
[[package]]
name = "itertools"
version = "0.10.5"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "b0fd2260e829bddf4cb6ea802289de2f86d6a7a690192fbe91b3f46e0f2c8473"
dependencies = [
 "either",
]

[[package]]
name = "itertools"
version = "0.11.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "b1c173a5686ce8bfa551b3563d0c2170bf24ca44da99c7ca4bfdab5418c3fe57"
dependencies = [
 "either",
]

[[package]]
name = "itertools"
version = "0.12.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "ba291022dbbd398a455acf126c1e341954079855bc60dfdda641363bd6922569"
dependencies = [
 "either",
]
```

but for some reason cargo decided to move some of the deps to the older versions... Let me see if I can fix that.

bantonsson (commented):
I'm a bit at a loss as to what this PR is supposed to do. Should it update something in a `Cargo.toml` file as well?

ivoanjo:
Thanks @bantonsson for the eagle-eyed review. Not sure why cargo would randomly pick the older versions when given multiple versions of a dependency on the same project, but fixed via a combination of a few more updates + changing the lock file manually and checking that cargo did not revert my changes.

> I'm a bit at a loss as to what this PR is supposed to do. Should it update something in a Cargo.toml file as well?

Maybe? The thing is, we're pulling in `dogstatsd/Cargo.toml` the latest version of the this repo:

```
[dependencies]
datadog-protos = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki-backport/" }
ddsketch-agent = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki-backport/" }
```

So the version is only being maintained in the `Cargo.lock` file. (These are the dependencies that were locking us from upgrading `tonic`).

Should we also move to specify the commit hash in the `.toml` file...? @duncanpharvey any thoughts on doing that?

duncanpharvey:
> Thanks @bantonsson for the eagle-eyed review. Not sure why cargo would randomly pick the older versions when given multiple versions of a dependency on the same project, but fixed via a combination of a few more updates + changing the lock file manually and checking that cargo did not revert my changes.
> 
> > I'm a bit at a loss as to what this PR is supposed to do. Should it update something in a Cargo.toml file as well?
> 
> Maybe? The thing is, we're pulling in `dogstatsd/Cargo.toml` the latest version of the this repo:
> 
> ```
> [dependencies]
> datadog-protos = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki-backport/" }
> ddsketch-agent = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki-backport/" }
> ```
> 
> So the version is only being maintained in the `Cargo.lock` file. (These are the dependencies that were locking us from upgrading `tonic`).
> 
> Should we also move to specify the commit hash in the `.toml` file...? @duncanpharvey any thoughts on doing that?

@ivoanjo Ideally the versions in saluki would be bumped then we reference those versions libdatadog whenever a change is made. But it looks like the version has never been bumped in the saluki datadog-protos package since it was created 6 months ago (which I think makes sense as it's still just an experimental package that is experiencing rapid development).

In this case I think it's ok to pin the saluki version dependencies in `dogstatsd/Cargo.toml` using the commit hash.

https://github.com/DataDog/saluki/blob/c162adbfc91b0d98b0d71b85c7629298025bc50f/lib/datadog-protos/Cargo.toml#L3

ivoanjo:
Pushed https://github.com/DataDog/libdatadog/pull/675/commits/7b14466aba9f2bd3c6927548e92f57d63300d9e2 to explicitly lock the specific git hash for those dependencies in the `dogstatsd/Cargo.toml`

alexgallotta (approved):
As Duncan wrote, pinning the version in the toml seems a clearer way. Looks good now!

ivoanjo:
Thanks folks, merging away :pray:

PR #680, opened 2024-10-16
[PROF-10680] Use ELF virtual address instead of file offset

nsavoire:
# What does this PR do?

Use ELF virtual address instead of file offset for normarlized addresses.

# Motivation

When computing normalized address, use elf virtual addresses instead of file offsets because file offset alone is not enough to perform symbolization with split debug information.

# Additional Notes

`file_offset` should probably be renamed, this will be addressed in a later PR.


gleocadie (approved):
LGTM

PR #681, opened 2024-10-16
Enable code coverage.

hoolioh:
# What does this PR do?

Code coverage started to malfunction due to some compatibility issue between current Rust  and llvm-cov/nextest versions. This PR bumps llvm-cov and nextest to bring back code coverage.


r1viollet (approved):
LGTM

PR #683, opened 2024-10-21
Stabilize credit_cards_bench inlining

bantonsson:
# What does this PR do?

Tells the `credit_cards_bench` to not inline the `is_credit_card` method to try to stabilize results a bit.

# Motivation

The results fluctuate due to _external_ factors like dependency changes that seem to affect the benchmarking harness and the inlining decisions.

# Additional Notes

The benchmark results will differ from `main` since this PR affects the generated code, but hopefully (and according to my tests), it should stabilize them a bit going forward.


VianneyRuhlmann (approved):
🤞

PR #686, opened 2024-10-22
Lock down nightly version for reproducible builds

bantonsson:
# What does this PR do?

Locks down the nightly version of rust so that builds are reproducible and don't break accidentally.

* Use nightly-2024-10-21
* Fix clippy lints
* Change nextest version to 0.9.81 in all places

# Motivation

Changes in the nightly version of rust can break formatting, clippy linting and coverage, among other things. That is not a good developer experience.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -24,7 +24,7 @@ jobs:
           swap-storage: true
       - uses: actions/checkout@v4
       - name: Install Rust
-        run: rustup install nightly && rustup default nightly
+        run: rustup install nightly-2024-10-21 && rustup default nightly-2024-10-21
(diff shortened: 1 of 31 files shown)


morrisonlevi (commented):
Was a specific strategy for bumping the nightly version also chosen? In practice, it needs to be bumped periodically or it will break for a different reasons :-)

ekump (approved):

bantonsson:
@morrisonlevi no strategy for bumping was chosen. Maybe move nightly when we move the supported minimum? Do you have any other ideas?

hoolioh (approved):

mellon85 (approved):

PR #688, opened 2024-10-22
add stale PR GH workflow

ekump:
# What does this PR do?

Introduce the stale action for PRs.
- PRs with no activity for 90 days will be marked with a `stale` label and a comment will be added to the PR
- After being stale for 14 days the PR will be automatically closed
- Issues are not included in this (but can be later, if we want)
- Branches will not be auto-deleted (but they can be, if we want)
- You can avoid this action on a PR with the `keep-open` label

# Motivation

Good repository hygiene suggested by @hoolioh  

# Additional Notes

`stale` and `keep-open` labels have been added to the repo.

# How to test the change?

I don't think we can really test the action until it is merged into main.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -15,7 +15,10 @@ jobs:
     steps:
       - uses: actions/stale@v8
         with:
-          stale-pr-message: "This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs."
+          stale-pr-message: >
+            This pull request has been automatically marked as stale because it has not had recent activity.
+            It will be closed if no further activity occurs. To override this behavior, add the keep-open
+            label or update the PR.
           days-before-issue-stale: -1
           days-before-issue-close: -1
           days-before-pr-stale: 90
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,24 @@
+name: "Mark stale pull requests"
+
+on:
+  schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - uses: actions/stale@v8
+        with:
+          stale-pr-message: "This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs."
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
+          days-before-pr-stale: 90
+          days-before-pr-close: 14
+          stale-pr-label: "stale"
+          exempt-pr-labels: "keep-open"
```
ivoanjo, line 24: Minor: It may be worth including a note in the stale pr message that "you can tag this pr as keep-open to avoid auto-closing" or something like that?

ivoanjo (approved):
This seems quite reasonable (and [it's quite similar to what we're doing for dd-trace-java](https://github.com/DataDog/dd-trace-java/blob/master/.github/workflows/prune-old-pull-requests.yaml)). I'm not sure if there's a way of testing it out before actually merging it in, but it doesn't seem to be overly problematic to merge and then check after the first few runs that everything is in good shape.

ekump:
> I'm not sure if there's a way of testing it out before actually merging it in

@ivoanjo I couldn't find a way to test an action before it is initially merged to main. I may be able to do a dry-run test after the merge but before the cron job kicks off at 0:00 GMT.

PR #689, opened 2024-10-23
[APMSP-1458] Fix single-span sampling

VianneyRuhlmann:
# What does this PR do?

Several fixes to sampling:
- Single-span sampling should not be made based on priority but based on sampling mechanism
- Analyzed tag is `_dd1.sr.eausr`
- Use const for keys

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/data-pipeline/src/trace_exporter.rs
+++ b/data-pipeline/src/trace_exporter.rs
@@ -110,25 +110,20 @@ fn drop_chunks(traces: &mut Vec<Vec<pb::Span>>) {
                 // We send chunks containing an error
                 return true;
             }
-
             // PrioritySampler and NoPrioritySampler
             let priority = span.metrics.get(SAMPLING_PRIORITY_KEY);
             if has_top_level(span) && (priority.is_none() || priority.is_some_and(|p| *p > 0.0)) {
                 // We send chunks with positive priority or no priority
                 return true;
             }
-            // SingleSpanSampler
+            // SingleSpanSampler and AnalyzedSpansSampler
             else if span
                 .metrics
                 .get(SINGLE_SPAN_SAMPLING_MECHANISM)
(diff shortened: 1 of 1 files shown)


data-pipeline/src/trace_exporter.rs
```diff
@@ -98,25 +103,28 @@ fn add_path(url: &Uri, path: &str) -> Uri {
...
-                    // We send chunks with positive priority
-                    return true;
-                }
-                // We send single spans with positive priority
-                sampled_indexes.push(index);
-            } else if priority.is_none() && has_top_level(span) {
-                // We send chunks with no priority
+            // PrioritySampler and NoPrioritySampler
+            let priority = span.metrics.get(SAMPLING_PRIORITY_KEY);
+            if has_top_level(span) && (priority.is_none() || priority.is_some_and(|p| *p > 0.0)) {
```
ekump, line 116: minor: Do you think `priority.map_or(true, |p| *p > 0.0)` is clearer than `(priority.is_none() || priority.is_some_and(|p| *p > 0.0))`?

ekump: Also, do we have sufficient unit tests covering this logic here? I didn't see any changes to tests.

ekump (approved):
a couple of minor comments, but LGTM

data-pipeline/src/trace_exporter.rs
VianneyRuhlmann: There are tests [here](https://github.com/DataDog/libdatadog/blob/321e98ecd1be000873ec1ba8831977cc6787683f/data-pipeline/src/trace_exporter.rs#L989). The sampling logic has not changed so no test update required.

The `(priority.is_none() || priority.is_some_and(|p| *p > 0.0))` syntax seems clearer to me

PR #692, opened 2024-10-24
Crashtracker receiver is spawned on crash

sanchda:
# What does this PR do?

Changes the way the crashtracker receiver works
* Uses an anonymous AF_UNIX socketpair for communication instead of a stdio pipe
* Eliminates some of the public API, such as being able to initialize with a named UDS
* Binds the lifetime of the receiver to the signal handler, being very careful to manage certain signals
* Increases the altstack size to accommodate the slightly higher stack utilization
* Tries really hard to reap the receiver exit status, so other PIDs do not
* Other miscellanea

This PR isn't yet complete, as it requires more tests and an RFC.  But please feel free to comment on what you see!

# Motivation

Some runtimes actually call `waitpid(-1,...` anticipating an ECHILD, which is incompatible with the idea of a long-lived receiver process.  We also wanted to improve some of the safety around how signals work.

# Additional Notes

# How to test the change?

[APMLP-286]

[APMLP-286]: https://datadoghq.atlassian.net/browse/APMLP-286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


danielsn:
How does this interact with the PHP sidecar?

sanchda:
> How does this interact with the PHP sidecar?

It should not.  In a future PR I think many of us hope to harmonize the two implementations, but there are a few questions about how packaging would work in that universe.  Right now I think the priority is to fix the outstanding defects, provide a sane assessment of risk for the runtimes which will rely on this implementation of crashtracking, and then harmonize over a more relaxed timeline.

sanchda:
Curiously, I see here that vfork availability isn't what I thought it was on macos.  Will check that out after I finish writing some documents and tests.

crashtracker/src/lib.rs
```diff
@@ -9,26 +9,24 @@
...
 //!    previous crash handler (if any) to execute, maintaining the customer
 //!    experience as much as possible.
-//! 2. The receiver process runs in the background, listening on `stdin`, which is connected by a
-//!    pipe to the parent process.  When a crash occurs, the receiver gathers the information from
-//!    the pipe, adds additional data about the system state (e.g. /proc/cpuinfo and /proc/meminfo),
-//!    formats it into a crash report, uploads it to the backend, and then exits. The receiver also
-//!    exits if the pipe is closed without a crash report, to avoid leaving a zombie process if the
-//!    parent exits normally.
+//! 2. The receiver process, which is spawned by the signal handler.  It is connected by an
+//!    anynomous AF_UNIX `socketpair()` to the parent process. When a crash occurs, the receiver
```
pawelchcki, line 25: https://man7.org/linux/man-pages/man2/pipe.2.html Pipe pair also could be an option.

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,268 @@
...
+            }
+            _ => {
+                return Err(anyhow::anyhow!("Error waiting for child process to exit"));
+            }
+        }
+    }
+}
+
+/// Wrapper around the child process that will run the crash receiver
+/// TODO the execve arg coersion requires allocation, which we should avoid from a signal handler.
```
r1viollet, line 177: I was going to comment on this, but you have identified it :+1: 
We are in the child process however this is still a signal handler :-/

crashtracker/src/collector/crash_handler.rs
```diff
@@ -41,37 +274,49 @@ enum ReceiverType {
...
+    .map(|(a, b)| (a.into_raw_fd(), b.into_raw_fd()))?;
+
+    // We need to spawn a process without calling atfork handlers, since this is happening inside
+    // of a signal handler.  Moreover, preference is given to multiplatform-uniform solutions.
+    // Although `vfork()` is deprecated, the alternatives have limitations
+    // * `fork()` calls atfork handlers
+    // * There is no guarantee that `posix_spawn()` will not call `fork()` internally
+    // * `clone()`/`clone3()` are Linux-specific
+    // Accordingly, use `vfork()` for now
+    match unsafe { vfork() } {
```
r1viollet, line 302: so here the trade-off is we sacrifice macos compatibility
again thanks for the good comments

sanchda: Well, originally I chose `vfork()` purely because I understood that it was deprecated in macos, but still available.  It seems that it's dropped in macos from `nix`.  I need to do some investigation here.

If `vfork` really isn't available in macos, then I'll probably rewrite this implementation to spawn using the `clone()` syscall directly (I think it amounts to the same thing in this precise case, but `vfork()` generates warnings in Rust because the compiler does not have stable `#[ffi_returns_twice]`).

sanchda: But to be clear, the safety guarantees we currently strive for on Linux will not necessarily extend to macos.

ivoanjo: (Should we care a lot for macOS, at least as a first pass? SSI doesn't support macOS, so we could choose to enable crashtracker only on Linux for now)

sanchda: I don't think it needs to be a top-level concern, but a lot of folks do their initial testing on macos and it's nice for them to be able to see that their code-level changes worked in the happy case.

I don't anticipate it being a big deal.  Either I can pull up `vfork()` and everything works out-of-the-box, or I can't and I just call `fork()` on macos and `clone()` on linux.  :shrug:

sanchda: (ofc this would create a situation where the macos implementation is less correct, and possibly even responsible for spurious failures if we looked at it too hard, but I consider that an acceptable compromise for the developer ergonomics)

((these messages are written speculatively because I do not develop on macos and while I am sensitive to those who expect a common systems interface, I do not think we or anyone can provide such a guarantee within such a fundamentally weird use-case, so I do think there is some amount of risk/error in even trying this on macos in the first place))

sanchda: (((this message brought to you by the Lots of Isolated Silly Parenthesis society)))

ivoanjo: ((((I approve of this message. Also I approve of LISP))))

crashtracker/src/lib.rs
sanchda: Indeed.  I was thinking of sticking to AF_UNIX simply because I need to have a named unix socket for some other use-cases.  I can't think of a scenario in which `socketpair()` would be blocked (e.g., by `seccomp()`) but a pipe pair wouldn't.  I also don't really think the two are all-to-distinct for this specific application, except that a streaming AF_UNIX is owned by the kernel networking code and has some slightly better message passing semantics (which we don't use).

Can you think of a good reason to prefer a pipe over a socketpair here?

crashtracker/src/collector/crash_handler.rs
danielsn: As someone who develops on osx and then does a final round of testing on linux, I'd prefer OSX to continue working for that workflow :).

ivoanjo: I'm just sayin': supporting for our own development is not the same as production-level bulletproof for customers :)

ivoanjo: (Or at least, not necessarily, depending on the effort required)

sanchda: > As someone who develops on osx and then does a final round of testing on linux, I'd prefer OSX to continue working for that workflow :).
Yup, sounds like we're on the same page.

> I'm just sayin': supporting for our own development is not the same as production-level bulletproof for customers :)
And it sounds like we're also on the same page.

Not sure how many folks can be on a single page, but given that I'm a 2-bit developer whose opinion doesn't matter 1 bit, that leaves at least 4095 bytes for the two of you, so I think we're fine.

rochdev: If this doesn't run on macOS, it blocks integrators (which pretty much all use macOS) from testing the crashtracker as a native extension in their languages without adding a lot of custom tooling. It would make development a lot more painful.

sanchda:
:wave: adding @bwoebi and @kevingosse as subject matter experts.  I'd really appreciate a review of the systems interactions in `crash_handler.rs` to make sure we're doing what we said we would on the tin.

crashtracker/src/collector/crash_handler.rs
bwoebi: Yep, nothing wrong with preparing the args at signal handler install time instead, stuffing it into a global and freeing in atexit or such.

crashtracker/src/collector/crash_handler.rs
```diff
@@ -324,75 +442,88 @@ fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Re
...
+    let res: Result<(), anyhow::Error>;
+    {
+        // `_guard` is a lexically-scoped object whose instantiation blocks/suppresses signals and
+        // whose destruction restores the original state
+        let _guard = SaGuard::<2>::new(&[signal::SIGCHLD, signal::SIGPIPE])?;
+
+        // Launch the receiver process
+        let receiver = make_receiver(receiver_config)?;
+
+        // Creating this tream means the underlying RawFD is now owned by the stream, so
```
kevingosse, line 483:
```suggestion
        // Creating this stream means the underlying RawFD is now owned by the stream, so
```

crashtracker/src/collector/crash_handler.rs
sanchda: Fixed -- these are now heap-allocated beforehand

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,227 @@
...
+        (
+            &receiver_args.binary_path,
+            &receiver_args.args_ptrs,
+            &receiver_args.env_vars_ptrs,
+        )
+    };
+
+    // Print the first string in the environment variables to stderr
+    // This is a way to check that the environment variables are being passed correctly
+    println!("Crashtracker receiver environment variable: {:?}", unsafe {
```
kevingosse, line 187: Where will this message be printed? Isn't there a risk of leaking some information?

sanchda: Sorry, this is wrong--debug statement.  Removed.

kevingosse (approved):

crashtracker-ffi/src/collector/datatypes.rs
```diff
@@ -79,16 +82,28 @@ impl<'a> TryFrom<Config<'a>> for datadog_crashtracker::CrashtrackerConfiguration
...
             vec
         };
+        // Verify that the timeout is within the bounds of an i32
+        let timeout_ms = if value.timeout_ms > i32::MAX as u32 {
+            return Err(anyhow::anyhow!(
+                "Timeout value {} is too large to fit in an i32",
+                value.timeout_ms
+            ));
+        } else if value.timeout_ms == 0 {
+            10_000 // We hardcode a magical default of 10s here
```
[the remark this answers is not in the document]
sanchda: Uh, two problems here
* This check should be deeper in the API, not the FFI.  Fixed it.
* Yeah, 10s is an arbitrarily high number.  I'm not sure what kind of default we should shoot for.  I changed it to 5s.

crashtracker/src/collector/crash_handler.rs
sanchda: This is an abomination; stricken

crashtracker-ffi/src/collector/datatypes.rs
```diff
@@ -80,15 +83,17 @@ impl<'a> TryFrom<Config<'a>> for datadog_crashtracker::CrashtrackerConfiguration
             vec
         };
         let create_alt_stack = value.create_alt_stack;
+        let use_alt_stack = value.use_alt_stack;
         let endpoint = value.endpoint.cloned();
         let resolve_frames = value.resolve_frames;
-        let wait_for_receiver = value.wait_for_receiver;
+        let timeout_ms = value.timeout_ms as u32;
```
danielsn, line 89: Shouldn't this check that it fits in an i32

danielsn: Ahh, its checked in the `::new`

crashtracker/src/collector/api.rs
```diff
@@ -146,12 +110,13 @@ fn test_crash() -> anyhow::Result<()> {
     let endpoint = Some(Endpoint::from_slice(&output_url));
 
     let path_to_receiver_binary =
-        "/tmp/libdatadog/bin/libdatadog-crashtracking-receiver".to_string();
+        "/home/ubuntu/dev/libdatadog_ct/target/release/build/builder-5add66ea14bef616/out/build/libdatadog-crashtracking-receiver".to_string();
```
danielsn, line 113: We need a way to standardize this

crashtracker/src/collector/saguard.rs
```diff
@@ -0,0 +1,86 @@
...
+
+// Provides a lexically-scoped guard for signals
+// During execution of the signal handler, it cannot be guaranteed that the signal is handled
+// without SA_NODEFER, thus it also cannot be guaranteed that signals like SIGCHLD and SIGPIPE will
+// _not_ be emitted during this handler as a result of the handler itself. At the same time, it
+// isn't known whether it is safe to merely block all signals, as the user's own handler will be
+// given the chance to execute after ours. Thus, we need to prevent the emission of signals we
+// might create (and cannot be created during a signal handler except by our own execution) and
+// defer any other signals.
+// TODO this forces dynamic allocation in the signal handler
```
danielsn, line 13: fix

crashtracker/src/collector/saguard.rs
```diff
@@ -0,0 +1,86 @@
...
+pub struct SaGuard<const N: usize> {
+    old_sigactions: [(signal::Signal, signal::SigAction); N],
```
danielsn, line 15: is this going to use a bunch of our stack space?

crashtracker/src/collector/saguard.rs
```diff
@@ -0,0 +1,86 @@
...
+        let mut old_sigactions = [(
+            signal::Signal::SIGINT,
+            SigAction::new(
+                SigHandler::SigDfl,
+                SaFlags::empty(),
+                signal::SigSet::empty(),
+            ),
+        ); N];
+
+        // Set SIG_IGN for the specified signals and save old handlers
```
danielsn, line 48: document ign/block/default

bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -114,6 +126,19 @@ fn test_crash_tracking_bin(
...
+
+    // Crashtracking signal handler chaining tests, as well as other tests, might only be able to
+    // influence system state after the main application has crashed, and has therefore lost the
+    // ability to influence the outcome of the test.  Those tests should create an "INVALID" file
+    // in the output directory.
+    // - If the file exists and contains only a single 'O' character, the test passes
+    // - Likewise, if the file does not exist, the test passes
+    // - Tests are free to output additional information in the file in case of a failure; it'll be
+    //   read here
+    let invalid_path = format!("{0}/INVALID", fixtures.output_dir.display());
```
danielsn, line 138: it might be more ideomatic to use pathbuf

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
+#[allow(dead_code)]
+struct PreparedExecve {
+    binary_path: CString,
+    args_cstrings: Vec<CString>,
+    args_ptrs: Vec<*const i8>,
+    env_vars_cstrings: Vec<CString>,
+    env_vars_ptrs: Vec<*const i8>,
+}
+
+impl PreparedExecve {
```
danielsn, line 65: there is heap allocation, but it happens before the signal, so its ok

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
+            .expect("Failed to convert binary path to CString");
+
+        // Allocate and store arguments
+        let args_cstrings: Vec<CString> = config
+            .args
+            .iter()
+            .map(|s| CString::new(s.as_str()).expect("Failed to convert argument to CString"))
+            .collect();
+        let mut args_ptrs: Vec<*const i8> = args_cstrings.iter().map(|arg| arg.as_ptr()).collect();
+        args_ptrs.push(std::ptr::null()); // Null-terminate the argument list
```
danielsn, line 78: chain an iterator that adds a null before we collect

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
+            env_vars_ptrs,
+        }
+    }
+}
+
+/// Opens a file for writing (in append mode) or opens /dev/null
+/// * If the filename is provided, it will try to open (creating if needed) the specified file.
+///   Failure to do so is an error.
+/// * If the filename is not provided, it will open /dev/null Some systems can fail to provide
+///   `/dev/null` (e.g., chroot jails), so this failure is also an error.
```
danielsn, line 107: 🤯

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
+/// * If the child is still alive, or some other error occurs, return an error Either way, after
+///   this returns, you probably don't have to do anything else.
+fn reap_child_non_blocking(pid: Pid, timeout_ms: u32) -> anyhow::Result<bool> {
+    let timeout = Duration::from_millis(timeout_ms as u64);
+    let start_time = Instant::now();
+
+    loop {
+        match waitpid(pid, Some(WaitPidFlag::WNOHANG)) {
+            Ok(WaitStatus::StillAlive) => {
+                if Instant::now().duration_since(start_time) > timeout {
```
danielsn, line 136: is this guaranteed signal safe?

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
+                    return Err(anyhow::anyhow!("Timeout waiting for child process to exit"));
```
danielsn, line 137: `anyhow::bail`

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
+                return Ok(true);
+            }
+            Err(nix::Error::ECHILD) => {
+                // Non-availability of the specified process is weird, since we should have
+                // exclusive access to reaping its exit, but at the very least means there is
+                // nothing further for us to do.
+                return Ok(true);
+            }
+            _ => {
+                return Err(anyhow::anyhow!("Error waiting for child process to exit"));
```
danielsn, line 150: not for today, but anyhow probably allocates when you have a string

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
+            }
+        }
+    }
+}
+
+/// Wrapper around the child process that will run the crash receiver
+fn run_receiver_child(uds_parent: RawFd, uds_child: RawFd, stderr: RawFd, stdout: RawFd) -> ! {
+    // File descriptor management
+    unsafe {
```
danielsn, line 159: This is all getting around the fact you can't just use the `Process` module

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
+        execve(
+            binary_path.as_ptr(),
+            args_ptrs.as_ptr(),
+            env_vars_ptrs.as_ptr(),
+        );
+    }
+
+    // If we reach this point, execve failed, so just exit
+    unsafe {
+        libc::_exit(-1);
```
danielsn, line 196: exit or abort?

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
+fn run_receiver_parent(_uds_parent: RawFd, uds_child: RawFd, _stderr: RawFd, _stdout: RawFd) {
+    let _ = close(uds_child);
+}
+
+fn wait_for_pollhup(target_fd: RawFd, timeout_ms: i32) -> anyhow::Result<bool> {
+    // Need to convert the RawFd into a BorrowedFd to satisfy the PollFd prototype
+    let target_fd = unsafe { BorrowedFd::borrow_raw(target_fd) };
+    let poll_fd = PollFd::new(&target_fd, PollFlags::POLLHUP);
+
+    match poll(&mut [poll_fd], timeout_ms)? {
```
danielsn, line 209: poll is signal safe?

danielsn: In fact, lets just comment a link to the relevent man page on every libc call we make

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,221 @@
...
-    c_void, mmap, sigaltstack, siginfo_t, MAP_ANON, MAP_FAILED, MAP_PRIVATE, PROT_NONE, PROT_READ,
-    PROT_WRITE, SIGSTKSZ,
+    c_void, dup2, execve, mmap, sigaltstack, siginfo_t, MAP_ANON, MAP_FAILED, MAP_PRIVATE,
+    PROT_NONE, PROT_READ, PROT_WRITE, SIGSTKSZ,
 };
-use nix::sys::signal;
-use nix::sys::signal::{SaFlags, SigAction, SigHandler};
-use std::fs::File;
-use std::io::Read;
+use nix::poll::{poll, PollFd, PollFlags};
```
danielsn, line 16: for each import, comment why we believe its signal safe

crashtracker/src/collector/crash_handler.rs
danielsn, line 209: Lets talk more about timeouts on the RFC

crashtracker/src/collector/crash_handler.rs
```diff
@@ -304,18 +410,24 @@ extern "C" fn handle_posix_sigaction(signum: i32, sig_info: *mut siginfo_t, ucon
 }
 
 fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Result<()> {
+    // If this is a SIGSEGV signal, it could be called due to a stack overflow. In that case, since
```
danielsn, line 413: Thanks for commenting this.

crashtracker/src/collector/crash_handler.rs
```diff
@@ -324,75 +436,88 @@ fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Re
...
+    );
+    let receiver_config = unsafe { receiver_config.as_ref().context("receiver config")? };
+
+    // Since we've gotten this far, we're going to start working on the crash report. This
+    // operation needs to be mindful of the total walltime elapsed during handling. This isn't only
+    // to prevent hanging, but also because services capable of restarting after a crash experience
+    // crashes as probabalistic queue-holding events, and so crash handling represents dead time
+    // which makes the overall service increasingly incompetent at handling load.
+    let timeout_ms = config.timeout_ms;
+    let start_time = Instant::now(); // This is the time at which the signal was received
```
danielsn, line 452: is this signal safe?

crashtracker/src/collector/crash_handler.rs
```diff
@@ -324,75 +436,88 @@ fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Re
...
+        // Either the receiver is done, it timed out, or something failed.
+        // In any case, can't guarantee that the receiver will exit.
+        // SIGKILL will ensure that the process ends eventually, but there's
+        // no bound on that time.
+        // We emit SIGKILL and try to reap its exit status for the remaining time, then just give
+        // up. Since the state of the process is checked through `waitpid()`, no check is
+        // made for the return of `kill()` (which would also reveal the equivalent of
+        // ECHILD).
+        unsafe {
+            libc::kill(receiver.receiver_pid, libc::SIGKILL);
```
danielsn, line 512: use `kill(0)` to avoid waitpid

crashtracker/src/collector/crash_handler.rs
```diff
@@ -414,17 +540,21 @@ fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Re
...
 ///     this function executing.
-pub fn register_crash_handlers(create_alt_stack: bool) -> anyhow::Result<()> {
+pub fn register_crash_handlers() -> anyhow::Result<()> {
     if !OLD_HANDLERS.load(SeqCst).is_null() {
         return Ok(());
     }
 
+    let config_ptr = CONFIG.load(SeqCst);
+    anyhow::ensure!(!config_ptr.is_null(), "No crashtracking config");
+    let (config, _config_str) = unsafe { config_ptr.as_ref().context("config ptr")? };
```
danielsn, line 550: could this have a race condition if this is called twice?

danielsn: Or on the other side of a `fork`

danielsn: Lets add some tests for that (apparantly we already do in python, but not in libdatadog)

crashtracker/src/collector/crash_handler.rs
```diff
@@ -414,17 +540,21 @@ fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Re
...
+
     unsafe {
-        if create_alt_stack {
+        if config.create_alt_stack {
             set_alt_stack()?;
```
danielsn, line 554: should that be renamed `create_alt_stack`

danielsn (approved):
Approved with comments

bin_tests/src/bin/crashtracker_bin_test.rs
```diff
@@ -28,29 +28,28 @@ mod unix {
...
+        let stdout_filename = format!("{output_dir}/out.stdout");
 
         let endpoint = if output_url.is_empty() {
             None
         } else {
             Some(Endpoint::from_slice(&output_url))
         };
 
-        let config = CrashtrackerConfiguration {
+        let mut config = CrashtrackerConfiguration {
```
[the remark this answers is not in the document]
sanchda: Discussed how this is related to the testing harness and that each test runs in its own process.

bin_tests/src/modes/behavior.rs
```diff
@@ -0,0 +1,78 @@
...
+use crate::modes::unix::*;
+
+/// Defines the additional behavior for a given crashtracking test
+pub trait Behavior {
+    fn setup(&self, output_dir: &str, config: &mut CrashtrackerConfiguration) -> Result<()>;
+    fn pre(&self, output_dir: &str) -> Result<()>;
+    fn post(&self, output_dir: &str) -> Result<()>;
+}
+
+pub fn does_file_contain_msg(filepath: &str, contents: &str) -> anyhow::Result<bool> {
```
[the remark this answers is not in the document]
sanchda: Agree, refactored to `file_content_equals` and modified all pathlike string thingies to be Path

bin_tests/src/modes/behavior.rs
```diff
@@ -0,0 +1,78 @@
...
+        .open(filepath)
+        .with_context(|| format!("Failed to open file: {filepath}"))?;
+    file.write_all(contents.as_bytes())
+        .with_context(|| format!("Failed to write to file: {filepath}"))?;
+    Ok(())
+}
+
+pub fn atom_to_clone<T: Clone>(atom: &AtomicPtr<T>) -> Result<T> {
+    let ptr = atom.load(Ordering::SeqCst);
+    if ptr.is_null() {
```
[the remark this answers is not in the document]
sanchda: 👍 fixed

bin_tests/src/modes/behavior.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+pub fn remove_file_permissive(filepath: &String) {
+    // Removes the file if it exists.  If it doesn't exist, it's not an error or anything.
+    if !filepath.is_empty() {
+        let _ = std::fs::remove_file(filepath);
+    }
+}
+
+pub fn get_behavior(mode_str: &str) -> Box<dyn Behavior> {
```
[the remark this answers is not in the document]
sanchda: Hard to say since it has to pass between processes as an execve string, but if there's a path forward here, I agree

crashtracker/src/collector/crash_handler.rs
sanchda: Good question.  It is, but more importantly this should be asserted and documented more carefully.  I've added a section at the top of the file that indicates which of the syscalls we use are async-signal safe and provides a link.

Of course, we do use some convenience wrappers through `nix`, so if `nix` calls free on something then that might be problematic.  I hope we can isolate and remove these over time.

crashtracker/src/collector/crash_handler.rs
sanchda: Indeed--done.  This should make us more correct on macos where `waitpid()` from a signal handler is not allowed.

crashtracker/src/collector/crash_handler.rs
sanchda: Agree, fixed

bin_tests/src/modes/unix/test_001_sigpipe.rs
```diff
@@ -0,0 +1,121 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
```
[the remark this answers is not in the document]
sanchda: Added top-level comments to all these test--good call

crashtracker/src/collector/crash_handler.rs
sanchda: Tried to clean this up, added a fork test.

bin_tests/src/bin/crashtracker_bin_test.rs
```diff
@@ -14,13 +14,16 @@ mod unix {
...
     use std::env;
+    use std::path::Path;
 
     use datadog_crashtracker::{
         self as crashtracker, CrashtrackerConfiguration, CrashtrackerMetadata,
         CrashtrackerReceiverConfig,
     };
     use ddcommon::{tag, Endpoint};
 
+    const TEST_COLLECTOR_TIMEOUT: u32 = 10_000;
```
danielsn, line 25:
```suggestion
    const TEST_COLLECTOR_TIMEOUT_MS: u32 = 10_000;
```

bin_tests/src/modes/unix/test_007_chaining.rs
```diff
@@ -1,38 +1,46 @@
 // Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
+//
+// This is a simple test to make sure that crashtracking can be chained properly by other signals.
+// This covers the case where the user code sets a signal handler after crashtracking has been
+// enabled.
+// Note that in this test, the user code doesn't do anything outrageous, like trying to call the
+// signal handler after doing a lonjump back into their normal stack, or doing absurd th ings to
```
danielsn, line 8: longjump

danielsn (commented):
nice

bin_tests/src/modes/unix/test_007_chaining.rs
sanchda: Good spotting--fixed

bin_tests/src/bin/crashtracker_bin_test.rs
sanchda: yoink

crashtracker/src/collector/crash_handler.rs
sanchda: Builds on mac, tests pass on mac.

crashtracker/src/collector/crash_handler.rs
```diff
@@ -414,17 +562,24 @@ fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Re
 ///     will not yet be stored.  This would lead to unexpected behaviour for the
 ///     user.  This should only matter if something crashes concurrently with
 ///     this function executing.
-pub fn register_crash_handlers(create_alt_stack: bool) -> anyhow::Result<()> {
+pub fn register_crash_handlers() -> anyhow::Result<()> {
+    static RUNNERS: AtomicU64 = AtomicU64::new(0);
+    let _ = AtomGuard::new(&RUNNERS)?; // Will return from this function if it's currently being
+                                       // called
```
bwoebi, line 568: I'm confused, how would that function ever be called recursively / simultaneously? I.e. what scenario are we guarding against here?

sanchda: Ah, I was thinking about a hypothetical scenario which is unlikely.  Almost all users are going to get this as part of their top-level call into the crashtracker.  I'll remove this bit.

crashtracker/src/collector/crash_handler.rs
```diff
@@ -2,35 +2,243 @@
...
+
+// The args_cstrings and env_vars_strings fields are just storage.  Even though they're
+// unreferenced, they're a necessary part of the struct.
+#[allow(dead_code)]
+struct PreparedExecve {
+    binary_path: CString,
+    args_cstrings: Vec<CString>,
+    args_ptrs: Vec<*const i8>,
+    env_vars_cstrings: Vec<CString>,
+    env_vars_ptrs: Vec<*const i8>,
```
bwoebi, line 75: Shouldn't these Vecs be `*const libc::c_char`? Different architectures have different signedness for char:
https://github.com/rust-lang/libc/blob/81f0cd3d9715e579c92063f78d0d14da85a31dd1/src/unix/linux_like/linux/gnu/b64/aarch64/mod.rs#L3
https://github.com/rust-lang/libc/blob/81f0cd3d9715e579c92063f78d0d14da85a31dd1/src/unix/linux_like/linux/musl/b64/x86_64/mod.rs#L1

sanchda: Ah, yes.  I'm surprised I didn't get an error here, since the aliases are clearly different, but this is definitely something that's broken things before.  Good catch--thank you.

sanchda: Fixed.  I suspect this would have broken ARM builds entirely due to how the conversions work, which makes me think I ought to build on an arm-linux machine before merging in case there are other issues.

crashtracker/src/collector/crash_handler.rs
```diff
@@ -324,75 +457,88 @@ fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Re
...
+            libc::kill(receiver.receiver_pid, libc::SIGKILL);
         }
-    }
-    // Calling "free" in a signal handler is dangerous, so we just leak the
-    // objects we took (receiver, metadata, config, etc)
+        let receiver_pid_as_pid = Pid::from_raw(receiver.receiver_pid);
+        let reaping_allowed_ms = timeout_ms.saturating_sub(start_time.elapsed().as_millis() as u32);
+        let _ = reap_child_non_blocking(receiver_pid_as_pid, reaping_allowed_ms)
+            .context("Failed to reap receiver process")?;
+    } // Drop the guard
```
bwoebi, line 539: Nit: Is there a specific reason we need this lexical scope and the `let res` declaration ahead of time? function end should drop anyway.

crashtracker/src/collector/saguard.rs
bwoebi: For the code we own we most likely are, but I don't think we can make that assumption for the stack walking.

crashtracker/src/collector/saguard.rs
```diff
@@ -0,0 +1,88 @@
...
+        // Save the current signal mask and block all signals except the suppressed ones
+        let mut old_sigmask = signal::SigSet::empty();
+        signal::sigprocmask(
+            signal::SigmaskHow::SIG_BLOCK,
+            Some(&suppressed_signals),
+            Some(&mut old_sigmask),
+        )?;
+
+        // Initialize array for saving old signal actions
+        // Except for SigChld and SigPipe, which are instantiated to SigIgn by default on all
```
bwoebi, line 35: According to `man 7 signal` sigpipe is _not_ Ign by default (sigchld is).

crashtracker/src/collector/crash_handler.rs
sanchda: Good point, the lexical scope is no longer necessary.

sanchda: Fixed, thank you

crashtracker/src/collector/saguard.rs
sanchda: Almost all of the comments in this are wrong.  Rewriting--thank you for spotting.

crashtracker/src/collector/saguard.rs
sanchda: Yeah, stack walking is going to be broken in a bunch of ways.

FWIW, we recently broke the 8192 byte default limit in a different PR before unwinding ever happened 😭

crashtracker/src/collector/crash_handler.rs
```diff
@@ -41,37 +248,52 @@ enum ReceiverType {
...
+    // NB -- on macos the underlying implementation here is actually `fork()`!  See the top of this
+    // file for details.
+    match unsafe { vfork() } {
+        0 => {
+            // Child (noreturn)
+            run_receiver_child(uds_parent, uds_child, stderr, stdout)
+        }
+        pid if pid > 0 => {
+            // Parent
+            run_receiver_parent(uds_parent, uds_child, stderr, stdout);
```
bwoebi, line 286: Minor nit: I find that code confusing ... calling a function with 4 args, just uses one and then it's an one-liner.
```suggestion
            _ = close(uds_child);
```
Trivial.

bwoebi (approved):
The strategy looks sound to me.

crashtracker/src/collector/crash_handler.rs
sanchda: Agree, refactored

PR #693, opened 2024-10-24
Bump crashtracking altstack size

sanchda:
# What does this PR do?

Some downstream PRs are affected by the too-small default altstack size used in crashtracking.  This is supposed to be a harmless fix.

I think creating a configurable stack size is irrelevant, since we should probably not have an overly-dynamic use of the stack.   Please chime in if you have a strong feeling on an ideal altstack size.

In particular, I think reviewers should check whether they feel like we should be pinned on a multiple of page size, or if we should use a number of bytes which we then round up to the nearest page (e.g., if the CPU MMU has a different page size configured, which may make this value really big).

### DIFF
--- a/crashtracker/src/collector/crash_handler.rs
+++ b/crashtracker/src/collector/crash_handler.rs
@@ -464,10 +464,15 @@ unsafe fn set_alt_stack() -> anyhow::Result<()> {
         return Ok(());
     }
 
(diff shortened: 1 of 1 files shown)


gleocadie (approved):
LGTM. Rubber ducking here since I'm not using this part of the code.
Having multiple page size looks good to me.

danielsn (approved):
:shipit:

PR #696, opened 2024-10-25
Design RFC for crashtracker

danielsn:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+For all languages other than .Net and Java, this is a signal handler.
+The collector interacts with the system in the following ways:
+
+1.  Registers a signal handler (`sigaction`).
+    The old handler is saved into a global variable to enable changing.
+    - Risks to normal operation
+      - Chaining handlers is not atomic.
+        It is possible for concurrent races to occur.
+        Mitigation: doing this operation once, early during system initialization.
+      - If the system expects and handles `sigsegv`, `sigbus` and `sigabort` as part of its ordinary functioning, this can violate principle 1.
```
ivoanjo, line 31: I've noticed that here + in some other comments in the crashtracker sources we claim to handle sigabort but we're actually only handling sigsegv + sigbus.

Does that mean we need to correct the docs, or handle sigaborts?

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        This is not common for most languages, but Java and .Net turn some crashes into recoverable user exceptions.
```
ivoanjo, line 32: I believe this is not the reason for Java; afaik Java (or more specifically the hotspot VM) is using SEGVs for internal VM mechanisms (e.g. GC optimizations and the like).

Not that it matters for this doc! But I would suggest being a bit more generic here to avoid being slightly incorrect -- e.g. stating that these VMs require direct control of these signals for implementation reasons and that we can't touch them there.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        Operating inside a crashing process is inherently risky.
+        We do our best to limit the operations we perform, and to move as much as possible across a process boundary, but there will always be a risk here.
+        Provide an environment variable to enable customers to disable the crashtracker.
+3.  Backtrace collected.
+    - Risks to normal operation
+      - N/A
+    - Risks during a crash
+      - If the stack is corrupted, attempting to collect the backtrace can crash the crashing process.
+        This prevents chaining the old signal handler and can break customer debugging workflows.
+        Mitigation: provide an environment variable allowing customers to turn off backtrace collection
```
ivoanjo, line 69: May be worth mentioning here that this downside only happens when doing stack trace collection in-process, not for the out-of-process option

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+      - If the stack is corrupted, attempting to collect the backtrace can crash the crashing process.
+        This will prevent the collection and reporting of any additional crash info.
+        Mitigation:
+        Collect the backtrace last, allowing less risky operations to complete first.
+        As a stretch mitigation, we could do out-of-process unwinding.
+        This would increase the complexity of the crashtracker, but would avoid this issue.
```
ivoanjo, line 75: Wait, isn't this what we're already doing? The "could" here seems to suggest it's something we could look into implementing, not something we already have as a possibility ready to use.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+    - Risks during a crash
+      - In some cases (e.g. stack overflow), invoking a signal handler can itself fail.
+        Mitigation: support the use of `sigaltstack`
+      - The signal handler could crash.
+        This would prevent the chaining of signal handlers, and could interfere with the users existing debugging workflow.
+        Mitigation:
+        Operating inside a crashing process is inherently risky.
+        We do our best to limit the operations we perform, and to move as much as possible across a process boundary, but there will always be a risk here.
+        Provide an environment variable to enable customers to disable the crashtracker.
+      - The signal handler could hang.
```
sanchda, line 56: IMHO, this is the worst possible outcome.  I've actually observed this to occur in a situation where libSegFault.so is layered with crashtracking and the underlying process does some _something_ (I don't yet know).

It isn't clear to me how to prevent this, since it seems that all of the solutions come down to exiting the process or just avoiding responsibility by mutating the sigaction under some condition.  I think the best idea I personally have is to use the equivalent of TSC to estimate the amount of off-signal work done between invocations of the handler, and if it's below some threshold then `_exit()` if the faulting IP matched the previous one.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        As a stretch mitigation, we could do out-of-process unwinding.
+        This would increase the complexity of the crashtracker, but would avoid this issue.
+4.  Results written to a pipe (unix socket)
+    - Risks to normal operation
+      - Maintaining the open pipe requires a file-descriptor.
+      - Maintaining a unix socket requires a synchronization point, typically a file in the file system.
+    - Risks during a crash
+      - If the pipe/socket is not drained, the collector will stall when attempting to write.
+        This could hang the crashing process.
+        Mitigation: Minimize the amount of data sent.
```
ivoanjo, line 83: > Mitigation: Minimize the amount of data sent.

It's true we _could_ do this, but are we doing it right now? :eyes:

We could look into making the receiver read things ASAP rather than limiting the collector (e.g. have a thread always reading data, that kinda thing?)

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+      - N/A
+    - Risks during a crash
+      - If the stack is corrupted, attempting to collect the backtrace can crash the crashing process.
+        This prevents chaining the old signal handler and can break customer debugging workflows.
+        Mitigation: provide an environment variable allowing customers to turn off backtrace collection
+      - If the stack is corrupted, attempting to collect the backtrace can crash the crashing process.
+        This will prevent the collection and reporting of any additional crash info.
+        Mitigation:
+        Collect the backtrace last, allowing less risky operations to complete first.
+        As a stretch mitigation, we could do out-of-process unwinding.
```
sanchda, line 74: Indeed, although out-of-process unwinding isn't generally as well-supported as in-process.  I know that blazesym and gimli have made some progress here, and I think that `PTRACE_ME` will trump PTRACE_YAMA and allow for e.g., `process_vm_readv()` etc., but this all requires some research.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        This was originally believed to be harmless.
+        However, it turns out that some frameworks assume that they are the only thing that can spawn workers, and do a `waitpid` on all children.
+        The existence of an additional child process hangs the framework.
+        Mitigation: Either do lazy initialization, or use a proper daemonized sidecar.
+    - Risks during a crash
+      - Lazy initialization consumes a PID and a slot in the process table.
+        This can cause either initialization to fail, or block the creation of another process on the customer system.
+        Mitigation: This is unlikely to occur.
+      - Forking inside a signal handler is unsafe (see Notes on [signal-safety(7) \- Linux manual page](https://man7.org/linux/man-pages/man7/signal-safety.7.html)).
+        Mitigation: Do cursed heroics to avoid triggering `at_fork` handlers.
```
ivoanjo, line 124: :rofl:

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        Mitigation:
+        Collect the backtrace last, allowing less risky operations to complete first.
+        As a stretch mitigation, we could do out-of-process unwinding.
+        This would increase the complexity of the crashtracker, but would avoid this issue.
+4.  Results written to a pipe (unix socket)
+    - Risks to normal operation
+      - Maintaining the open pipe requires a file-descriptor.
+      - Maintaining a unix socket requires a synchronization point, typically a file in the file system.
+    - Risks during a crash
+      - If the pipe/socket is not drained, the collector will stall when attempting to write.
```
sanchda, line 81: Good point.  The IPC client on both sides should probably use strictly nonblocking reads and bail out if it detects an error in the underlying state machine.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+    Both options have issues.
+    - Risks to normal operation
+      - Eager initialization consumes a PID and a slot in the process table.
+        This can cause either initialization to fail, or block the creation of another process on the customer system.
+        Mitigation: This is unlikely to occur.
+      - Eager initialization can lead to an additional child process.
+        This was originally believed to be harmless.
+        However, it turns out that some frameworks assume that they are the only thing that can spawn workers, and do a `waitpid` on all children.
+        The existence of an additional child process hangs the framework.
+        Mitigation: Either do lazy initialization, or use a proper daemonized sidecar.
```
ivoanjo, line 118: The words "or use a proper daemonized sidecar" are doing a lot of work here. I have a vague notion of what this is from past discussions, but on a reading of this RFC it's a bit unclear what this is, why it solves the problem, and why we're not considering it below (where we only assume the lazy initialization solution?)

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+    - Risks to normal operation
+      - May consume a small amount of resources.
+        Mitigation: None.
+    - Risks during a crash
+      - If the pipe is not drained fast enough, the crashtracker collector will hang.
+        Mitigation:
+        Currently, none, other than careful coding to limit work done while draining the socket.
+        In the future we might wish to play with process niceness to increase the chance of the receiver running and draining the socket.
+      - The collector might crash or be terminated, truncating the message.
+        Mitigation: If an unexpected input is received, including EOF, make a best effort attempt to format and send a partial crash report.
```
ivoanjo, line 135: This is somewhat duplicate from the above; I guess because it's a concern on both sides of the pipe/socket/etc? It may be worth to add a quick call-out between them, to make the connection more explicit.

sanchda (approved):
Left some comments, mostly speculative around future design.  I do not think my comments should necessarily change the RFC as it is written, I'm just adding them here as a record for future discussions.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+3.  Transmits the message to the backend.
+    - Risks to normal operation
+      - NA
+    - Risks during a crash
+      - The endpoint may be inaccessible.
+      Mitigation: None.
```
ivoanjo, line 141: May be worth explicitly stating "drop the report"? Also, I guess we could try to emit a metric to say we failed? Since it's a different backend.

ivoanjo (approved):
Thanks for putting this together! I think this RFC will be a very useful resource to make sure everyone is on the same page when discussing the crashtracker.

I've left a bunch of suggestions, but I think overall this is pretty good :+1:

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+      - The signal handler could crash.
+        This would prevent the chaining of signal handlers, and could interfere with the users existing debugging workflow.
+        Mitigation:
+        Operating inside a crashing process is inherently risky.
+        We do our best to limit the operations we perform, and to move as much as possible across a process boundary, but there will always be a risk here.
+        Provide an environment variable to enable customers to disable the crashtracker.
+      - The signal handler could hang.
+        This is potentially worse for the customer, since a crashed process will be restarted by the kubernetes controller, while a hanging process will remain alive until reaped.
+        This consumes resources and reduces the ability of the customer application to handle load.
+        Mitigation:
```
bwoebi, line 59: We might also impose a timeout upon crashtracking signal handler entering?

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+3.  Backtrace collected.
+    - Risks to normal operation
+      - N/A
+    - Risks during a crash
+      - If the stack is corrupted, attempting to collect the backtrace can crash the crashing process.
+        This prevents chaining the old signal handler and can break customer debugging workflows.
+        Mitigation: provide an environment variable allowing customers to turn off backtrace collection
+      - If the stack is corrupted, attempting to collect the backtrace can crash the crashing process.
+        This will prevent the collection and reporting of any additional crash info.
+        Mitigation:
```
bwoebi, line 72: We might also just replace the sigsegv handler within the crashtracking handler, with recovery via sigsetjmp and siglongjmp.
That way we probably can do a clean shutdown in any case.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+    - Risks during a crash
+      - Due to user security controls, these files may not be available to read.
+        Mitigation: The crashtracker receiver will still transmit partial crash reports.
+      - The files may be large.
+        Mitigation:
+        None currently.
+        In the future, we could add a file size limit and only send the first `n` bytes.
+      - These files could contain sensitive data
+        Mitigation: only send a select set of files which do not contain sensitive data.
+
```
bwoebi, line 100: 6. Resulting core dumps are sometimes a bit corrupted / contain the crashtracker handler.
 - This can be confusing when viewing the core dumps and registers are not quite what they were before the handler was invoked.
Mitigation: Unset the crashtracker handler and return from the signal handler, so that the OS will fall back to the default action of creating a core dump ... at the original sigsegv location.

docs/RFCs/0004-crashtracker-design.md
bwoebi: We might also just set SO_RCVBUFSIZ on the other side to _big enough_ to basically just have enough space to write it all at once.

docs/RFCs/0004-crashtracker-design.md
bwoebi: I suppose lazy init is really the simplest way to handle that...

pawelchcki: Sidecar will not help with init 1 problem on its own. 

If we don't go the lazy initializaiton route - we will need to add additional mitigations:
e.g. sidecar becomes an actual init 1 - or we spawn sidecar as threads within sibling processes. etc


To simplify existing solution - lazy initialization is the way to go.

docs/RFCs/0004-crashtracker-design.md
bwoebi: This just means invoking the syscall directly, bypassing any libc stuff, right?

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        Mitigation: This is unlikely to occur.
+      - Eager initialization can lead to an additional child process.
+        This was originally believed to be harmless.
+        However, it turns out that some frameworks assume that they are the only thing that can spawn workers, and do a `waitpid` on all children.
+        The existence of an additional child process hangs the framework.
+        Mitigation: Either do lazy initialization, or use a proper daemonized sidecar.
+    - Risks during a crash
+      - Lazy initialization consumes a PID and a slot in the process table.
+        This can cause either initialization to fail, or block the creation of another process on the customer system.
+        Mitigation: This is unlikely to occur.
```
pawelchcki, line 122: This is super highly unlikely to occur. In my n==too_many_years of software development, I only run into issues with process spawning - when I run a shell - ForkBomb :) 

I would expect infinite more problems with seccomp or other permission type things.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+3.  When a crash does occur, the crashtracker should **make a best-effort attempt to minimize impact** on the crashing process.
+
+## Architectural model
+
+### Collector
+
+For all languages other than .Net and Java, this is a signal handler.
+The collector interacts with the system in the following ways:
+
+1.  Registers a signal handler (`sigaction`).
```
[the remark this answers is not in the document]
natitsechanski: I actually implemented something simple like you describe for https://datadoghq.atlassian.net/browse/APMON-1332 back in August, but at some point "we" decided we 'won't do' that 🥴 

image

docs/RFCs/0004-crashtracker-design.md
natitsechanski, line 31: This can probably be mitigated by hooking `sigaction` (monitoring for any new signal handler registrations) storing the new registered handler and manually call it when we're done doing our stuff.

docs/RFCs/0004-crashtracker-design.md
natitsechanski: For `ptrace` you'll probably need to add custom capability `SYS_PTRACE` on docker.
And you might have problems with SElinux.
And I thought we want to be done ASAP, this will incur "some" overhead.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+      - Maintaining a unix socket requires a synchronization point, typically a file in the file system.
+    - Risks during a crash
+      - If the pipe/socket is not drained, the collector will stall when attempting to write.
+        This could hang the crashing process.
+        Mitigation: Minimize the amount of data sent.
+      - If the pipe/socket is not available or is closed early, data could not be sent, or the crashtracker could hang.
+        Mitigation:
+        Data is sent to the receiver in priority order.
+        If the channel is closed, the receiver will send as much data as possible, as well as a flag indicating that an error occurred.
+5.  Files collected from `/proc`
```
natitsechanski, line 88: what and why are we collecting `/proc`?

bwoebi: /proc/maps for instruction pointer to object correlation.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+      - N/A
+    - Risks during a crash
+      - Due to user security controls, these files may not be available to read.
+        Mitigation: The crashtracker receiver will still transmit partial crash reports.
+      - The files may be large.
+        Mitigation:
+        None currently.
+        In the future, we could add a file size limit and only send the first `n` bytes.
+      - These files could contain sensitive data
+        Mitigation: only send a select set of files which do not contain sensitive data.
```
natitsechanski, line 99: this should be well defined and not left to a developers discretion 😅

docs/RFCs/0004-crashtracker-design.md
pawelchcki: avoiding libc things is a separate concern. this just avoids calling `posix_spawn` or raw `fork(2)`

I was looking at the existing wip implementation and it uses `vfork`

I would argue we should use the cascade - depending on what is available:
`clone3` then `clone` then `vfork` on non linux systems.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        However, it turns out that some frameworks assume that they are the only thing that can spawn workers, and do a `waitpid` on all children.
+        The existence of an additional child process hangs the framework.
+        Mitigation: Either do lazy initialization, or use a proper daemonized sidecar.
+    - Risks during a crash
+      - Lazy initialization consumes a PID and a slot in the process table.
+        This can cause either initialization to fail, or block the creation of another process on the customer system.
+        Mitigation: This is unlikely to occur.
+      - Forking inside a signal handler is unsafe (see Notes on [signal-safety(7) \- Linux manual page](https://man7.org/linux/man-pages/man7/signal-safety.7.html)).
+        Mitigation: Do cursed heroics to avoid triggering `at_fork` handlers.
+2.  Listens for a crashreport a socket/pipe.
```
natitsechanski, line 125: why should we implement this socket/pipe overhead in the case of 
1. collecting the crashes in the process context and 
2.  (lazy initialization?) starting a "receiver" (maybe it should be renamed to `transmitter` following this comment 😅 ) with the crash report as an argument and let it just do a best effort to transmit it before everything collapses.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        Mitigation: None.
+    - Risks during a crash
+      - If the pipe is not drained fast enough, the crashtracker collector will hang.
+        Mitigation:
+        Currently, none, other than careful coding to limit work done while draining the socket.
+        In the future we might wish to play with process niceness to increase the chance of the receiver running and draining the socket.
+      - The collector might crash or be terminated, truncating the message.
+        Mitigation: If an unexpected input is received, including EOF, make a best effort attempt to format and send a partial crash report.
+3.  Transmits the message to the backend.
+    - Risks to normal operation
```
natitsechanski, line 137: This could fail
This could hang
This could crash.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+      - Eager initialization can lead to an additional child process.
+        This was originally believed to be harmless.
+        However, it turns out that some frameworks assume that they are the only thing that can spawn workers, and do a `waitpid` on all children.
+        The existence of an additional child process hangs the framework.
+        Mitigation: Either do lazy initialization, or use a proper daemonized sidecar.
+    - Risks during a crash
+      - Lazy initialization consumes a PID and a slot in the process table.
+        This can cause either initialization to fail, or block the creation of another process on the customer system.
+        Mitigation: This is unlikely to occur.
+      - Forking inside a signal handler is unsafe (see Notes on [signal-safety(7) \- Linux manual page](https://man7.org/linux/man-pages/man7/signal-safety.7.html)).
```
pawelchcki, line 123: Other than at_fork handlers we need to be very careful about the code we call from within the crash handler.

We should avoid calling anything libc or any shared resource related - e.g. any internal library mutex'es must be avoided etc.

This includes most operations that allocate memory on heap- no Vec.new .etc

Anything we need to execute the handler process. must be prepared ahead of time.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+3.  Transmits the message to the backend.
+    - Risks to normal operation
+      - NA
+    - Risks during a crash
+      - The endpoint may be inaccessible.
+      Mitigation: None.
+      - It may take a signifcant amount of time to transmit the crashreport.
+      This could cause the user process to hang.
+      Mitigation: a configurable timeout on transmission (default 3s).
+      - The message could be modified or corrupted in transit
```
pawelchcki, line 145: We're using Datadog Agent as the proxy here- and we communicate via TCP. 

To keep the brevity of this document, I would consider removing this concern.

docs/RFCs/0004-crashtracker-design.md
ivoanjo: Actually, I realize now that my comment is incorrect, since we collect the stack in the process, I was mixing it up with the symbolication part.

docs/RFCs/0004-crashtracker-design.md
ivoanjo: Actually, I realize now that my comment is incorrect, since we collect the stack in the process, I was mixing it up with the symbolication part.

docs/RFCs/0004-crashtracker-design.md
bwoebi: I wish you good luck with collecting stacktraces without allocations. I mean, if there's a way to do so, I'll very gladly embrace it. In fact most crashes in the crashtracker (in PHP) are just within allocations.

sanchda: > I wish you good luck with collecting stacktraces without allocations

Indeed.  Conceptually, one can imagine an unwinding strategy that uses only pre-allocated buffers.  In reality
* DWARF specifies instructions to execute in an interpreted sandbox.  It's a VM...
* Organizationally, we've gone over and over and over the idea of building an in-house remote unwinder and it keeps getting shot down.  It's a tremendous burden.  Like, pick and hand-rolled ELF parser and throw the elfutils test suite at it just to see how it crashes and burns--and that's the _easy_ part of the problem.

docs/RFCs/0004-crashtracker-design.md
natitsechanski: So we need to be extremely careful when we pass this information to another process, as this can potentially be an information leak that could break [ASLR](https://en.m.wikipedia.org/wiki/Address_space_layout_randomization)

docs/RFCs/0004-crashtracker-design.md
natitsechanski: A transmitter like that could also prove useful for things like the injector to pass telemetry from the constructor

docs/RFCs/0004-crashtracker-design.md
sanchda: Agree, although the entity responsible for framing such a position is unknown to me...

docs/RFCs/0004-crashtracker-design.md
sanchda: > could break ASLR
Fair point.  Indeed, there are organizations that disable vsyscall (among many, many other things) for exactly this reason.

A better option for the future may be to harvest `/proc/self/maps` and normalize all PC values to offset/file tuples and transmit that information as part of the stack trace (naturally, we would then not emit any values representative of process virtual memory layout).

I do not know how this plays with timing.  I'm just saying that the information we currently extract from `/proc/self/maps` could be reduced to a non-identifying digest of comparable-or-better diagnostic value.

I do wonder how actionable this leakage is, in practice, but I also recognize that this isn't for us to determine.  A user might have zero tolerance, no matter how circumstantial, for forms of leakage which have been the basis for impactful and damaging CVEs.

docs/RFCs/0004-crashtracker-design.md
danielsn: Added to the doc

docs/RFCs/0004-crashtracker-design.md
pawelchcki: https://github.com/chromium/crashpad/blob/b8937c6cb4b38c1ca06b46791c84b31632895f1f/handler/linux/exception_handler_server.cc#L169

Crashpad is a very good source for this - they show a cascade of recommended options

tl;dr; they attempt to use Ptrace if its available - because its safer this way. Then they will try to use a ptracebroker process - then with locked YAMA and no PCAP - they will default to inprocess collection.

docs/RFCs/0004-crashtracker-design.md
pawelchcki: This is what I'm worried about:
https://github.com/lattera/glibc/blob/master/malloc/malloc.c#L3212

There are plenty of locks throughout libc - and just like with fork(2) - accessing those from a signal handler of a crashing process is... potentially interesting.

docs/RFCs/0004-crashtracker-design.md
pawelchcki: We definitely need to be careful. But if I were to bet - ASLR map leak from a crashed process is not very helpful for an exploit. Assuming the process will exit as soon as we're done inspecting its memory.

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        Mitigation: only send a select set of files which do not contain sensitive data.
+
+### Receiver
+
+A signal handler is fundamentally limited in what operations it can perform (see [signal-safety(7) \- Linux manual page](https://man7.org/linux/man-pages/man7/signal-safety.7.html)).
+This is made worse by the fact that the collector operates in the context of a crashing process whose state may be corrupted, causing even seemingly safe operations to crash/  
+The receiver interacts with the system in the following ways:
+
+1.  Forks a new process.
+    This can either occur eagerly, at process initialization or lazily, as part of crash handling.
```
r1viollet, line 108: Or within a privileged container (agent). Though I understand the concerns of having to ship matching libraries on both sides.

docs/RFCs/0004-crashtracker-design.md
r1viollet: I can also see the benefit of starting by de-registering ourselves. In case we also crash within the signal handler.
sigaction is in the [signal safe table](https://man7.org/linux/man-pages/man7/signal-safety.7.html).

docs/RFCs/0004-crashtracker-design.md
kevingosse: I think that's a strong argument in favor of collecting all the information from the receiver process, instead of in-process.

@sanchda why building our own remote unwinder instead of "just" deploying libunwind, possibly forking and customizing it for our needs?

docs/RFCs/0004-crashtracker-design.md
kevingosse: @sanchda  This has the additional benefit of making crash reports easier to compare and therefore to triage.

docs/RFCs/0004-crashtracker-design.md
kevingosse: Is libsegfault something that can be realistically deployed in a production environment? If so, we might want to add some testing around that 🤔

docs/RFCs/0004-crashtracker-design.md
pawelchcki: >I think that's a valid idea. In fact, why not just fork (in a manner that suppresses atfork handlers) and do everything in the fork (exiting the fork without calling atexit handlers), then continuing to chain in the handler after a return?

More or less 0 this is what is done in breakpad impl.

docs/RFCs/0004-crashtracker-design.md
danielsn: Documented in the RFC

docs/RFCs/0004-crashtracker-design.md
danielsn: Documented

docs/RFCs/0004-crashtracker-design.md
```diff
@@ -0,0 +1,146 @@
...
+        Mitigation: doing this operation once, early during system initialization.
+      - If the system expects and handles `sigsegv`, `sigbus` and `sigabort` as part of its ordinary functioning, this can violate principle 1.
+        This is not common for most languages, but Java and .Net turn some crashes into recoverable user exceptions.
+        A crashtracker signal handler risks breaking that mechanism.
+        Mitigation: not using signal handler based crashtracking for these languages.
+    - Risks during a crash
+      - If user code installs signal handlers, these can conflict with the crashtracker.
+        This can occur either intentionally by the user, or unexpectedly as the result of installing a framework which does so.
+        Mitigation:
+        If the user installs their handler after the crashtracker is initialized, then the crashtracker will likely not work unless the user was careful to chain signal handlers.
```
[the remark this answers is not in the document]
danielsn: Documented

docs/RFCs/0004-crashtracker-design.md
danielsn: Documented

docs/RFCs/0004-crashtracker-design.md
danielsn: documented

docs/RFCs/0004-crashtracker-design.md
danielsn: Documented

docs/RFCs/0004-crashtracker-design.md
danielsn: done

docs/RFCs/0004-crashtracker-design.md
natitsechanski: > But if I were to bet - ASLR map leak from a crashed process is not very helpful for an exploit

https://googleprojectzero.blogspot.com/2020/08/mms-exploit-part-5-defeating-aslr-getting-rce.html#:~:text=I%20decided%20to,of%20the%20attack

now the fact WE don't know how to exploit a vulnerability doesn't mean it's not a threat and that we can break security mechanisms.

I want to remind you that 20 years ago most programmers weren't too worried about reading strings of arbitrary length into a stack allocated bounded buffer "200 characters should be enough for a 'full name', right?" 😈

docs/RFCs/0004-crashtracker-design.md
danielsn: documented

docs/RFCs/0004-crashtracker-design.md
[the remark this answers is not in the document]
danielsn: documented

kevingosse (approved):

docs/RFCs/0004-crashtracker-design.md
danielsn: We are currently not handling `sigabort`, but intend to.

docs/RFCs/0004-crashtracker-design.md
danielsn: documented

docs/RFCs/0004-crashtracker-design.md
danielsn: noted

docs/RFCs/0004-crashtracker-design.md
danielsn: We do already have a thread always reading data on the receiver.  We could try playing with the niceness / increasing the buffer size.  I've updated the doc with those ideas.

docs/RFCs/0004-crashtracker-design.md
danielsn: noted

docs/RFCs/0004-crashtracker-design.md
danielsn: fixed

docs/RFCs/0004-crashtracker-design.md
danielsn: Good idea. Added.

sanchda: (FYI, this is implemented in #692 ([here](https://github.com/DataDog/libdatadog/pull/692/files#diff-116d1150f3fe727b2b50d527b061c112c22a6e8cf5b3c29523f30ce6e90310d4R408-R414))

docs/RFCs/0004-crashtracker-design.md
danielsn: Added to the doc

docs/RFCs/0004-crashtracker-design.md
danielsn: noted

docs/RFCs/0004-crashtracker-design.md
danielsn: Implementation issue for #692

docs/RFCs/0004-crashtracker-design.md
danielsn: https://github.com/DataDog/libdatadog/issues/700 to track

docs/RFCs/0004-crashtracker-design.md
danielsn: During normal operation, nothing is transmitted.

docs/RFCs/0004-crashtracker-design.md
danielsn: noted

PR #701, opened 2024-10-31
Collect agent info from sidecar

bwoebi:
The client needs to know about the env default (to properly apply it).
This PR provides the ability to store the agent info in shared memory (polling every 60 sec).


sidecar/src/service/agent_info.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+pub struct AgentInfoFetcher {
+    last_update: Instant,
+    rc: u32,
+    infos: Shared<ManualFuture<AgentInfoStruct>>,
+}
+
+impl AgentInfoFetcher {
```
bantonsson, line 73: This file has no comments or tests 😞 I also find it confusing that this has the same name `AgentInfoFetcher` as the `AgentInfoFetcher` in `data-pipeline`. Yes, I know that there are differences, but couldn't the `break` check, and the _processing_ be moved behind a `trait` and there only be one impl of the loop?

bantonsson (changes requested):
Please add comments and tests.

sidecar/src/service/agent_info.rs
bwoebi: I wasn't intentionally choosing the same name. Maybe I can call it AgentInfoWriter, but not sure if that helps?

You theoretically could do it, but the actual amount of shared code would be quite low & the boilerplate for the sharing of code quite large in comparison? I feel like fetch_info_with_state is the right level of abstraction for this.

I can though add at least a test for the happy path here.

bantonsson: Fine, let's not try to share more code then.

Please add a happy path test and some comments. I mean at least something like:

```rust
//! This file contains code for fetching and sharing the info from the Datadog Agent
//! ... etc
```

bwoebi: Yes, I will do, just haven't gotten round to it yet.

bwoebi: @bantonsson Done :-)

bantonsson (approved):

PR #702, opened 2024-10-31
[crashtracker] Enable timeout on receiver

danielsn:
# What does this PR do?

Adds a timeout to the crashtracker receiver, enabling it to post a partial crash report if the collector hangs.  Previously, a collector hang could cause the receiver to also hang.

# Motivation

This implements a mitigation from the [RFC](https://github.com/DataDog/libdatadog/blob/main/docs/RFCs/0003-crashtracker-design.md).
# Additional Notes

There is an open question about how to configure this timeout.  Currently, we're offering an environment variable.  We could take the value from the config, but that is passed on the socket, so its odd for a timeout on the socket to depend on a value that came on the socket.  

# How to test the change?

Added two new tests.

### DIFF
--- a/crashtracker/src/receiver.rs
+++ b/crashtracker/src/receiver.rs
@@ -425,6 +425,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg_attr(miri, ignore)]
     async fn test_receive_report_short_timeout() -> anyhow::Result<()> {
         let (sender, receiver) = tokio::net::UnixStream::pair()?;
 
@@ -445,6 +446,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg_attr(miri, ignore)]
     async fn test_receive_report_long_timeout() -> anyhow::Result<()> {
         let (sender, receiver) = tokio::net::UnixStream::pair()?;
 
--- a/crashtracker/src/receiver.rs
+++ b/crashtracker/src/receiver.rs
@@ -5,6 +5,8 @@
 use super::*;
 use crate::shared::constants::*;
 use anyhow::Context;
+use std::time::Duration;
+use std::time::Instant;
 use tokio::io::AsyncBufReadExt;
 use tokio::io::BufReader;
 use tokio::net::UnixListener;
@@ -64,12 +66,22 @@ pub fn receiver_entry_point_unix_socket(socket_path: impl AsRef<str>) -> anyhow:
     // Dropping the stream closes it, allowing the collector to exit if it was waiting.
 }
 
+pub fn receiver_timeout() -> Duration {
+    if let Ok(s) = std::env::var("DD_CRASHTRACKER_RECEIVER_TIMEOUT_MS") {
+        if let Ok(v) = s.parse() {
+            return Duration::from_millis(v);
+        }
+    }
+    // Default value
+    Duration::from_millis(5000)
+}
+
 pub fn receiver_entry_point_stdin() -> anyhow::Result<()> {
     let stream = BufReader::new(tokio::io::stdin());
     let rt = tokio::runtime::Builder::new_current_thread()
         .enable_all()
         .build()?;
-    rt.block_on(receiver_entry_point(stream))?;
+    rt.block_on(receiver_entry_point(receiver_timeout(), stream))?;
     Ok(())
 }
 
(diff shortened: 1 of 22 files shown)


crashtracker/src/receiver.rs
```diff
@@ -64,12 +66,22 @@ pub fn receiver_entry_point_unix_socket(socket_path: impl AsRef<str>) -> anyhow:
     // Dropping the stream closes it, allowing the collector to exit if it was waiting.
 }
 
+pub fn receiver_timeout() -> Duration {
+    if let Ok(s) = std::env::var("DD_CRASHTRACKER_RECEIVER_TIMEOUT_MS") {
```
sanchda, line 70: Right--this one is hard.

Two failed ideas from my side.
1. The receiver code actually doesn't have control over the acquisition of its own args and it'd be a weird change to the entrypoint to make a new args struct required.  So not a great choice of arg propagation.
2. Relying on the receiver configuration means waiting a few steps into the IPC sequence, which is not only weird but incorrect.  I think I suggested this at one point, but I see how naive that was.

That really leaves the environment as the dominant way of passing this configuration.

With that in mind, some thoughts.  I think the first two are constructive and the last is speculative.

First, we should probably by the naming that has been determined elsewhere.  All the runtimes are using the `CRASHTRACKING `label instead of `CRASHTRACKER`.

Second, how do you feel about marking this "private' (prepending one or two `_` to the name)?  End-users probably shouldn't be passing this.  They should be using whatever configuration levers are provided by their library, and the library should pass this.

 Third, here's an idea for tightening the interface.  We could take a new option in the ReceiverConfig and then propagate it to the receiver's environment in `make_receiver` as part of the Prepared Args.

To sketch out that idea,
1. `impl PreparedExecve` would have a new `fn new_with_env(config: &CrashtrackerReceiverConfig, env: Option<Vec<String, String>>)` interface
2. Merge `config.env` and `env` (this other vec) in the PreparedExecve `new_with_env`, with the `env` coming last (winner-take-all in case of conflict).

crashtracker/src/receiver.rs
```diff
@@ -64,12 +66,22 @@ pub fn receiver_entry_point_unix_socket(socket_path: impl AsRef<str>) -> anyhow:
...
+        if let Ok(v) = s.parse() {
+            return Duration::from_millis(v);
+        }
+    }
+    // Default value
+    Duration::from_millis(5000)
```
sanchda, line 76: Right now this default is identical to the default we use for the collector timeout.  I think, ideally, this timeout will be shorter than the other timeout, so we have time to send a message somewhere saying what's going to happen.

(in a future PR, that other timeout will be enforced from outside of the collector, so I don't think we can necessarily assume the collector isn't responding to its own timeouts if IPC stalls)

crashtracker/src/receiver.rs
```diff
@@ -283,30 +296,64 @@ enum CrashReportStatus {
...
+    timeout: Duration,
     stream: impl AsyncBufReadExt + std::marker::Unpin,
 ) -> anyhow::Result<CrashReportStatus> {
     let mut crashinfo = CrashInfo::new();
     let mut stdin_state = StdinState::Waiting;
     let mut config = None;
 
     let mut lines = stream.lines();
+    let mut deadline = None;
+    // Start the timeout counter when the deadline when the first crash message is recieved
```
sanchda, line 308: Yup, agree.  This harmonizes the different receiver types.

crashtracker/src/receiver.rs
```diff
@@ -283,30 +296,64 @@ enum CrashReportStatus {
...
+            deadline = Some(Instant::now() + timeout);
+            remaining_timeout = timeout;
+        }
     }
 
     if !crashinfo.crash_seen() {
         return Ok(CrashReportStatus::NoCrash);
     }
 
+    // Without a config, we don't even know the endpoint to transmit to.  Not much to do to recover.
```
sanchda, line 356: We _could_ encode this in the args or the environment, as per my earlier comment about the prepared args.

That said, I think there's not much to do in this case anyway and the likelihood is low anyway.

crashtracker/src/receiver.rs
```diff
@@ -327,3 +374,91 @@ async fn receive_report(
         ))
     }
 }
+
+#[cfg(test)]
```
sanchda, line 378: Very nice, great idea.

sanchda (approved):
LGTM; I added some comments for discussion, but none of them block this PR.

crashtracker/src/receiver.rs
```diff
@@ -5,6 +5,8 @@
 use super::*;
 use crate::shared::constants::*;
 use anyhow::Context;
+use std::time::Duration;
+use std::time::Instant;
```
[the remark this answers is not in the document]
danielsn: done

crashtracker/src/receiver.rs
danielsn: https://github.com/DataDog/libdatadog/issues/717

PR #703, opened 2024-11-01
Update licenses in preparation for v14.0.0

sanchda:
This updates the licenses


sanchda:
Closing because it makes more sense to just do this in the bump PR, since it looks like some licenses changed with the deps.

PR #706, opened 2024-11-01
Updates blazesym to 0.1.0rc1

sanchda:
# What does this PR do?

Seems like a prudent thing to do

# Motivation

Plus this upgrade happens when trying to cargo update, so why not?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
- "gimli 0.31.1",
+ "gimli",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


morrisonlevi (approved):
I see no issue with an rc0 to rc1 update.

peterg17 (approved):
Upgrading the `blazesym-c` bindings to `0.1.0-rc.1` makes sense. I took a quick look at the new APIs and your changes check out

danielsn (approved):

sanchda:
> Shouldn't [crashtracker blazesym version](https://github.com/DataDog/libdatadog/blob/main/crashtracker/Cargo.toml?rgh-link-date=2024-11-08T13%3A49%3A08Z#L28) be updated too ?

Yeah, I have a PR to consolidate these again.  I actually think this might have affected crashtracking unwinding on Alpine for some reason (glibc systems seem fine???)

bwoebi:
@sanchda fwiw - there's also rc.2 by now, so if you update crashtracker blazesym, just update both too?

PR #712, opened 2024-11-05
[APMSP-1512] Add metadata headers for stats

VianneyRuhlmann:
# What does this PR do?

- Add `Datadog-Meta-...` headers to stats
- Add `git_commit_sha` to TraceExporter meta for ClientStatsPayload
- Add `interpreter_vendor` to TraceExporter meta
- Rename the `version` field to `app_version` to be more explicit

This PR can be reviewed commit-by-commit

# Motivation

These headers are required by design doc

# Additional Notes

# How to test the change?

### DIFF
--- a/data-pipeline/src/stats_exporter.rs
+++ b/data-pipeline/src/stats_exporter.rs
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
+    borrow::Borrow,
+    collections::HashMap,
     sync::{
         atomic::{AtomicU64, Ordering},
         Arc, Mutex,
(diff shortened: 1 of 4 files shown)


data-pipeline/src/stats_exporter.rs
```diff
@@ -158,24 +152,24 @@ impl StatsExporter {
...
+        runtime_id: meta.runtime_id.clone(),
+        tracer_version: meta.tracer_version.clone(),
+        service: meta.service.clone(),
         sequence,
         stats: buckets,
-        // Agent-only field
+        git_commit_sha: meta.git_commit_sha.clone(),
+        // These fields will be set by the Agent
+        container_id: String::new(),
+        tags: Vec::new(),
```
ichinaski, line 172: 👍 From trace-agent perspective

ekump (approved):
LGTM...but I noticed we don't have integration tests hitting the test-agent for sending stats to validate we're sending the right headers and payloads. Should we? Does the test-agent even support the stats endpoint right now?

VianneyRuhlmann:
> LGTM...but I noticed we don't have integration tests hitting the test-agent for sending stats to validate we're sending the right headers and payloads. Should we? Does the test-agent even support the stats endpoint right now?

It has support for stats endpoint but I don't think it allows to check headers in snapshots

PR #713, opened 2024-11-06
[crashtracking] fix SA_ONSTACK

sanchda:
# What does this PR do?

In my last crashtracking PR, I split up the configuration for _using_ and _enabling_ the altstack.  However, since this was really tricky to test, it wasn't tested, so of course there was a typo (`use` is unused and `create` implies both)

# Motivation

Fixing crashtracking

# Additional Notes

# How to test the change?

Tests are implemented now for all variants
* use, but do not create
* create and use
* do not use, but create (paradox!)
* do not use, do not create

### DIFF
--- a/crashtracker/src/collector/api.rs
+++ b/crashtracker/src/collector/api.rs
@@ -161,3 +161,426 @@ fn test_crash() -> anyhow::Result<()> {
     assert_eq!(q, 3);
     Ok(())
 }
+
+#[test]
+fn test_altstack_paradox() -> anyhow::Result<()> {
+    use crate::StacktraceCollection;
+    use chrono::Utc;
+    use ddcommon::Endpoint;
+
+    let time = Utc::now().to_rfc3339();
+    let dir = "/tmp/crashreports/";
+    let output_url = format!("file://{dir}{time}.txt");
+
+    let endpoint = Some(Endpoint::from_slice(&output_url));
+
+    let create_alt_stack = true;
+    let use_alt_stack = false;
+    let resolve_frames = StacktraceCollection::EnabledWithInprocessSymbols;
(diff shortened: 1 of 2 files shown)


crashtracker/src/collector/crash_handler.rs
```diff
@@ -604,7 +604,7 @@ unsafe fn register_signal_handler(
     //   operation it is necessary to CREATE and USE the altstack.
     // - There are no known cases where it is useful to crate but not use the altstack--this case
     //   handled in `new()` for CrashtrackerConfiguration.
-    let extra_saflags = if config.create_alt_stack {
+    let extra_saflags = if config.use_alt_stack {
```
danielsn, line 607: 💯

crashtracker/src/collector/api.rs
```diff
@@ -161,3 +161,432 @@ fn test_crash() -> anyhow::Result<()> {
...
+        vec![],
+        create_alt_stack,
+        use_alt_stack,
+        endpoint,
+        resolve_frames,
+        timeout_ms,
+    );
+
+    assert!(
+        config.is_err(),
```
danielsn, line 193: Should we check what the particular error is?

danielsn (approved):

crashtracker/src/collector/crash_handler.rs
sanchda: 😿 😿 🐱  🚀

crashtracker/src/collector/api.rs
sanchda: Yes, good call.  Otherwise we can't guarantee we're testing the precise effect we desired.

PR #714, opened 2024-11-06
Add names to memfds

bwoebi:
I was confused while debugging what role was pertaining to what fd.


ipc/src/platform/unix/mem_handle.rs
```diff
@@ -44,17 +44,17 @@ static ANON_SHM_ID: AtomicI32 = AtomicI32::new(0);
 
 impl ShmHandle {
     #[cfg(target_os = "linux")]
-    fn open_anon_shm() -> anyhow::Result<OwnedFd> {
-        if let Ok(memfd) = memfd::MemfdOptions::default().create("anon-shm-handle") {
+    fn open_anon_shm(name: &str) -> anyhow::Result<OwnedFd> {
```
bantonsson, line 47: This is not really a blocker, but I'm wondering why these methods contain the word `anon` in them, since they all have paths, and now even a `name`, which doesn't feel anonymous.

bantonsson (approved):

ipc/src/platform/unix/mem_handle.rs
bwoebi: The only way you can access them is basically via procfs (the names are really just hints you see in the procfs readlink). Whereas the non-anon ones are properly existing in /dev/shm and accessible by other processes (if permissions allow).

PR #716, opened 2024-11-06
[crashtracker] RFC to mitigate hangs

danielsn:
# What does this PR do?

Adds an RFC mitigating an open issue from #696  

# Motivation

We want to limit the potential impact of crashtracking on users. 

# Additional Notes

# How to test the change?

This is a documentation only change

### DIFF
--- a/docs/RFCs/0004-crashtracker-prevent-hangs.md
+++ b/docs/RFCs/0004-crashtracker-prevent-hangs.md
@@ -27,7 +27,7 @@ Care should be taken during this process to avoid
 This child process (referred to as the **collector** in this RFC) is responsible for collecting the data required for crashtracking, while the parent operates as a watchdog (and is hence referred to as the **watchdog** process in the RFC).
 The **receiver** process remains as in RFC 3, either running as an [execve](https://man7.org/linux/man-pages/man2/execve.2.html) fork of the watchdog, or as an independent sidecar.
 If the child process completes successfully, the parent process reaps its PID, waits for the receiver process to finish uploading, and then chains the next signal handler, as it does currently.
-If the new child process, or the receiver process exceed their timeout budget, the watchdog [kills](https://man7.org/linux/man-pages/man2/kill.2.html) the offending child process, and then and then chains the next signal handler.
+If the collector or the receiver exceed their timeout budget, the watchdog [kills](https://man7.org/linux/man-pages/man2/kill.2.html) the offending child process, and then chains the next signal handler.
 
 ### Timeouts
 
--- a/docs/RFCs/0004-crashtracker-prevent-hangs.md
+++ b/docs/RFCs/0004-crashtracker-prevent-hangs.md
@@ -56,12 +56,12 @@ Each component has a total timeout budget, which it must enforce.
 Since each component conducts a series of operations, the total budget will typically be subdivided among those operations, in a hierarchical fashion.
 This increases the probability that some operations will succeed, maximizing the total amount of actionable data that makes it to the backend.
 The values for these timeout budgets SHOULD be configurable, either through an explicit configuration object, or through documented environment variables.
-This RFC is probably the best place to document such variables: implementers who add a configuration variable SHOULD make an amendment to this RFC to document it,
+This RFC is probably the best place to document such variables: implementers who add a configuration variable SHOULD make an amendment to this RFC to document it.
 
 #### Watchdog
 
 The watchdog is the fundamental defence against process hangs in this scheme.
-The recommended default overall timeout budget is ?.
+The recommended default overall timeout budget is 5 seconds.
 
 It MUST ensure that the crashing process cleanly calls the next signal-handler/default/abort when the overall crashtracker timeout budget is exceeded.
 The watchdog SHOULD make a best effort attempt to cleanup its child processes before exiting.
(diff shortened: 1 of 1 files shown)


docs/RFCs/0004-crashtracker-prevent-hangs.md
```diff
@@ -0,0 +1,81 @@
...
+
+As documented in [RFC 3: Crashtracker design](https://github.com/DataDog/libdatadog/blob/main/docs/RFCs/0003-crashtracker-design.md#design-priorities), the crashtracker has the following design priorities:
+
+1.  When the system is functioning normally, the crashtracker should, to the greatest extent possible, **have no impact on the functioning of a non-crashing system**.
+2.  When a crash does occur, the crashtracker should **prioritize the reliable collection and reporting of crash information**.
+3.  When a crash does occur, the crashtracker should **make a best-effort attempt to minimize impact** on the crashing process.
+
+In the [current architecture](https://github.com/DataDog/libdatadog/blob/main/docs/RFCs/0003-crashtracker-design.md#architectural-model), the crashtracker collector operates within the context of a crashing process.
+As a baseline requirement, crashtracker code should aim to be [signal safe](https://man7.org/linux/man-pages/man7/signal-safety.7.html).
+This is difficult, since some operations (e.g. stack walking), may do non signal safe operations such as `malloc`.
```
sanchda, line 15: This is true.  In absolute terms I suppose we could relax this requirement by building the entire stack ourselves and auditing its implementation intensely, but doing so would require a pretty high investment.

docs/RFCs/0004-crashtracker-prevent-hangs.md
```diff
@@ -0,0 +1,81 @@
...
+Even if the code does avoid all non-signal-safe operations, the program itself may be in such a corrupted state that the crashtracker will itself crash or hang.
+
+In the context of a crashing customer process, a crash within the crashtracker is bad because it can disrupt existing debugging workflows.
```
sanchda, line 18: This could be a frustrating and impactful situation for end-users.  Imagine that you have a crashing application and it turns out that it's due to a dependency, but you have absolutely no means by which to deduce this fact.  You might waste an extreme amount of engineering time trying to figure out why it's crashing.

Moreover, if we crash, then we're not getting a crash report, so _our_ ability to intervene on the behalf of such a user is limited.  Bad situation.

docs/RFCs/0004-crashtracker-prevent-hangs.md
```diff
@@ -0,0 +1,81 @@
...
+However, since the process was crashing anyway, it does not affect the overall availability of the service.
```
sanchda, line 19: I think it's possible to dream up counter-examples, such as segfault recovery, but within the scope of the runtimes we really care about for this implementation I think it's unlikely to be an issue.  So, I agree.

docs/RFCs/0004-crashtracker-prevent-hangs.md
```diff
@@ -0,0 +1,81 @@
...
+A hang is worse, because it can prevent the process/pod from being restarted for a significant amount of time, potentially affecting the availability of the entire service.
+This RFC proposes an architectural mechanism to mitigate the risk of crashtracker hangs.
+
+## Proposed Architecture
+
+In the proposed architecture, the crashing process [clones](https://man7.org/linux/man-pages/man2/clone.2.html) a child process.
+Care should be taken during this process to avoid
```
sanchda, line 26: to avoid `malloc` and also atfork handlers, I take it?  Looks like the sentence got chopped off.

docs/RFCs/0004-crashtracker-prevent-hangs.md
```diff
@@ -0,0 +1,81 @@
...
+Even if the code does avoid all non-signal-safe operations, the program itself may be in such a corrupted state that the crashtracker will itself crash or hang.
+
+In the context of a crashing customer process, a crash within the crashtracker is bad because it can disrupt existing debugging workflows.
+However, since the process was crashing anyway, it does not affect the overall availability of the service.
+A hang is worse, because it can prevent the process/pod from being restarted for a significant amount of time, potentially affecting the availability of the entire service.
+This RFC proposes an architectural mechanism to mitigate the risk of crashtracker hangs.
+
+## Proposed Architecture
+
+In the proposed architecture, the crashing process [clones](https://man7.org/linux/man-pages/man2/clone.2.html) a child process.
```
sanchda, line 25: There are actually a few solutions here, ranging from
* vfork
* the fork syscall itself, which is still widely available on unices/BSD
* the clone syscall on Linux
* the clone3 syscall on Linux

Personally, I'm leaning toward using `fork()` just because it does exactly what we want, it's general, and there's little chance for confusion.  Other folks have very good reasons for preferring `clone()` or even `clone3()`.  For the purpose of this document, I propose that the precise choice is an implementation detail.

docs/RFCs/0004-crashtracker-prevent-hangs.md
```diff
@@ -0,0 +1,81 @@
...
+Care should be taken during this process to avoid
+This child process (referred to as the **collector** in this RFC) is responsible for collecting the data required for crashtracking, while the parent operates as a watchdog (and is hence referred to as the **watchdog** process in the RFC).
+The **receiver** process remains as in RFC 3, either running as an [execve](https://man7.org/linux/man-pages/man2/execve.2.html) fork of the watchdog, or as an independent sidecar.
+If the child process completes successfully, the parent process reaps its PID, waits for the receiver process to finish uploading, and then chains the next signal handler, as it does currently.
+If the new child process, or the receiver process exceed their timeout budget, the watchdog [kills](https://man7.org/linux/man-pages/man2/kill.2.html) the offending child process, and then and then chains the next signal handler.
```
sanchda, line 30:
```suggestion
If the collector or the receiver exceed their timeout budget, the watchdog [kills](https://man7.org/linux/man-pages/man2/kill.2.html) the offending child process, and then chains the next signal handler.
```

docs/RFCs/0004-crashtracker-prevent-hangs.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Timeouts
+
+Each component has a total timeout budget, which it must enforce.
+Since each component conducts a series of operations, the total budget will typically be subdivided among those operations, in a hierarchical fashion.
```
sanchda, line 35: Good call.  This actually implies a tension between
* Speed
* Emitting partial data to the agent/proxy/backend frequently

I believe we are correct to underspecify this at present, since the document only leaves open the possibility without creating an imperative.

docs/RFCs/0004-crashtracker-prevent-hangs.md
```diff
@@ -0,0 +1,81 @@
...
+This increases the probability that some operations will succeed, maximizing the total amount of actionable data that makes it to the backend.
+The values for these timeout budgets SHOULD be configurable, either through an explicit configuration object, or through documented environment variables.
+This RFC is probably the best place to document such variables: implementers who add a configuration variable SHOULD make an amendment to this RFC to document it,
+
+#### Watchdog
+
+The watchdog is the fundamental defence against process hangs in this scheme.
+The recommended default overall timeout budget is ?.
```
sanchda, line 43: I don't have a firmer estimate.  It's hard to frame one because we actually don't know about the variance in time for things like unwinding, and because the IPC mechanism we have is mediated over syscalls.  This might need to be an empirically derived value with a sane ballpark made for this RFC.  For instance, I don't actually think 5s should be too low (it might be too high).

sanchda (approved):
Left some comments, but they are very much for discussion and in no way impact the substance of this RFC.

Looks pretty good to me.

docs/RFCs/0004-crashtracker-prevent-hangs.md
danielsn: Added to the RFC

docs/RFCs/0004-crashtracker-prevent-hangs.md
danielsn: noted

docs/RFCs/0004-crashtracker-prevent-hangs.md
danielsn: fixed

PR #722, opened 2024-11-08
[crashtracking]: add named socket support back in

sanchda:
# What does this PR do?

In the patches leading up to v14, I had tried to consolidate some of the receiver code paths.  In so doing, I accidentally removed the entire ability for a collector to speak with an async receiver.

Sorry, PHP!

Rather than adding the same `ddog_crasht_init_with_unix_socket` as before, this patch adds a new `ddog_crasht_init_without_receiver` interface.  All this does is check that the passed `config` has a valid `unix_socket_path`.

# Motivation

Fixing past mistakes.


crashtracker/src/collector/crash_handler.rs
```diff
@@ -492,13 +538,23 @@ fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Re
...
+        if receiver_config.is_null() {
+            return Err(anyhow::anyhow!("No receiver config"));
+        }
+        let receiver_config = unsafe { receiver_config.as_ref().context("receiver config")? };
+        make_receiver(receiver_config)?
+    };
 
-    // Creating this stream means the underlying RawFD is now owned by the stream, so
-    // we shouldn't close it manually.
+    // No matter how the receiver was creatd, attach to its stream
```
bwoebi, line 557:
```suggestion
    // No matter how the receiver was created, attach to its stream
```

bwoebi (approved):
Looks good to me, thanks for adding it back :-)

sanchda:
Don't merge yet, need to verify the thing works end-to-end with a native test

sanchda:
Looking good, thank you @bwoebi!

I did some local tests and found that it's possible for the collector to lock up indefinitely if the receiver should ever hit a condition where it doesn't drain the unix domain socket (the write will block in the syscall), but I only discovered that by accident (custom receiver implementation with a bug).

Except for freezing the receiver, I can't really think of a condition which would allow this to arise in normal code.  I have an incoming PR which will push the coordinator into its own forked process, allowing the original process to keep track of its time budget.  That will fix the cause of these (speculative!!!) hangs.

PR #725, opened 2024-11-11
[PROF-9476] Add experimental profiling managed string storage

ivoanjo:
# What does this PR do?

This PR builds on the work started by @alexjf on #607 to introduce "managed string storage" for profiling.

The idea is to introduce another level of string storage for profiling that is decoupled in lifetime from individual profiles, and that is managed by the libdatadog client.

At its core, managed string storage provides a hashtable that stores strings and returns ids. These ids can then be provided to libdatadog instead of `CharSlice`s when recording profiling samples.

For FFI users, this PR adds the following APIs to manage strings:

* `ddog_prof_ManagedStringStorage_new`
* `ddog_prof_ManagedStringStorage_intern(String)`
* `ddog_prof_ManagedStringStorage_unintern(id)`
* `ddog_prof_ManagedStringStorage_advance_gen`
* `ddog_prof_ManagedStringStorage_drop`
* `ddog_prof_ManagedStringStorage_get_string`

A key detail of the current implementation is that each `intern` call with the same string will increase an internal usage counter, and `unintern` call with reduce the counter.

Then at `advance_gen` time, if the counter is zero, we get rid of the string.

Then to interact with profiles, there's a new `ddog_prof_Profile_with_string_storage` API to create a profile with a given `ManagedStringStorage`, and all structures that make up a `Sample` (`Mapping`, `Function`, `Label`) etc have been extended so that they either take a `CharSlice` or a `ManagedStringId`.

Thus, after interning all strings for a sample, it's possible to add a sample to a profile entirely by referencing strings by ids, rather than `CharSlice`s.

# Motivation

The initial use-case is to support heap profiling -- "samples" related to heap profiling usually live across multiple profiles (as long as a given object is alive) and so this data must be kept somewhere.
Previously for Ruby we were keeping this on the Ruby profiler side, but having libdatadog manage this instead presents a few optimization opportunities.

We also hope to replace a few other "string tables" that other profilers had to build outside of libdatadog for similar use-cases.

This topic was also discussed in the following two documents (Datadog-only, sorry!):

* [Ruby heap stacktrace storage overhaul RFC](https://docs.google.com/document/d/1O-YhTrMgDi_kLGj5dBfC_lnxxraNma0pxCxyM9qfiro/edit?tab=t.0#heading=h.rnd972k0hiye)
* [Ruby heap profling with managed string storage review](https://docs.google.com/document/d/1NvipWBAwVPEv56TeDJDaksEa7vK7in7nYpySPHWcvyE/edit?tab=t.0#heading=h.cvqer9nx7oxp)

# Additional Notes

In keeping with the experimental nature of this feature, I've tried really hard to not disturb existing profiling API users with the new changes.

That is -- I was going for, if you're not using managed string storage, you should NOT be affected AT ALL by it -- be it API changes or overhead.

(This is why on the pure-Rust profiling crate side, I ended up duplicating a bunch of structures and functions. I couldn't think of a great way to not disturb existing API users other than introducing alternative methods, but to be honest the duplication is all in very simple methods so I don't think this substantially increases complexity/maintenance vs trying to be smarter to bend Rust to our will.)

There's probably a lot of improvements we can make, but with this PR I'm hoping to have something in a close to "good enough" state, that we can merge this in and then start iterating on master, rather than have this continue living in a branch for a lot longer.

This doesn't mean we shouldn't fix or improve things before merging, but I'll be trying to identify what needs to go in now and what can go in as separate, follow-up PRs.

As an addendum, there's still a bunch of `expect`s sprinkled that should be turned into proper errors. I plan to do a pass on all of those. (But again, none of the panics affect existing code, so they're harmless and inert unless you're experimenting with the new APIs)

# How to test the change?

The branch in https://github.com/DataDog/dd-trace-rb/tree/ivoanjo/prof-9476-managed-string-storage-try2 is where I'm testing the changes on the Ruby profiler side.

It may not be entirely up-to-date with the latest ffi changes on the libdatadog side (I've been prettying up the API), but it shows how to use this concept, while passing all the profiling unit/integration tests, and has shown improvements in memory and latency in the reliability environment.


profiling-ffi/src/profiles.rs
```diff
@@ -128,9 +131,11 @@ pub struct Period<'a> {
 #[derive(Copy, Clone, Default)]
 pub struct Label<'a> {
     pub key: CharSlice<'a>,
+    pub key_id: ManagedStringId,
 
     /// At most one of the following must be present
     pub str: CharSlice<'a>,
+    pub str_id: ManagedStringId,
```
nsrip-dd, line 138: Something I wondered about while I was trying this out for the Python profiler: I assume the caller would need to check that this is nonzero before calling unintern on it, right? Or in any other case where a string might be optional. Maybe unintern should do nothing with managed strings with value 0, so it's safe to just always call unintern when you're done with a managed string ID?

profiling-ffi/src/string_storage.rs
```diff
@@ -0,0 +1,204 @@
...
+use std::{ffi::CStr, rc::Rc, sync::RwLock};
+
+#[repr(C)]
+pub struct ManagedStringId {
+    pub value: u32,
+}
+
+#[repr(C)]
+pub struct ManagedStringStorage {
+    // This may be null, but if not it will point to a valid Profile.
```
gleocadie, line 14: typo ? Profile ? 🤔

profiling-ffi/src/string_storage.rs
```diff
@@ -0,0 +1,204 @@
...
+        anyhow::Ok(string)
+    })()
+    .context("ddog_prof_ManagedStringStorage_get_string failed")
+    .into()
+}
+
+#[must_use]
+#[no_mangle]
+/// TODO: @ivoanjo Should this take a `*mut ManagedStringStorage` like Profile APIs do?
+pub unsafe extern "C" fn ddog_prof_ManagedStringStorage_advance_gen(
```
gleocadie, line 142: what does this function ?
It's not clear from the name what advance gen means. Does it mean increase the reference counter/generation number so it's not collected ?

nsrip-dd:
Forgot to comment sooner, but I tried out using this in the Python profiler. [Commit here](https://github.com/DataDog/dd-trace-py/commit/547f1fe1362179f7b7ee4a01159dacde91f8f916#diff-1af87bc2fc4ba5a4e8f4b172318d261b0d53039758557a2a4730c4def9ee42ca), very WIP because I was trying out other approaches as well at the time. The biggest issue I ran into was around forking. If the program forks while the `RwLock` in the table is held, the child process will deadlock trying to access the table. I think we'd need to just make a new table on fork. Not the end of the world, but I guess just something to look out for.

profiling-ffi/src/string_storage.rs
morrisonlevi: I would appreciate some more exposition here too. This basically seems like a re-implementation of reference counted strings? Do we not use those std lib types for some specific reason?

profiling-ffi/src/profiles.rs
ivoanjo: Makes sense, added in 6a658b4ea76407a3ff74a765ccd67b82634e48a6

profiling-ffi/src/string_storage.rs
ivoanjo: Fixed in 00a76f2e042e101292725e64f598f884bf1c7621

profiling-ffi/src/string_storage.rs
ivoanjo: This is an excellent question. [Quoting Alex's original notes](https://docs.google.com/document/d/1NvipWBAwVPEv56TeDJDaksEa7vK7in7nYpySPHWcvyE/edit?disco=AAABUjZ_77A):

> String storage cleanup in the PoC is based on usage counts.
>
> The initial implementation was not based on usage counts but on `last_usage_gen`: if we didn't use a string while building the current profile, we can drop it for the next profile. This led to crashes when we implemented the optimization to not report objects with age == 0 though (the interned strings associated with those objects would only be used after a GC but if a profile flush occurred in between the 2 events, the interned strings would become invalidated).
>
> A trivial change to that would have been to only clean up when unused for x generations. But this felt brittle (what if we have a usecase where we'll skip objects with age < 10) and memory wasteful.

So yeah the original intent was to automatically clean up unused strings based on "was this used in the last profile or not".

This (almost) matches really well with heap profiling, because heap profiling is all about repeating a sample on every profile until the object gets collected, so this was a somewhat natural fit.

But, as Alex pointed out, this becomes a bit thornier as we have an optimization on the Ruby heap profiler to not report objects that haven't at least survived a single GC cycle.

Thus, in the current version, this is purely done based on the caller doing all the tracking work, rather than the generational approach.

I'd say the code is a bit... weird right now because it's not very confident on what to do here. So here's my questions to y'all:

* Does the purely reference-counted mechanism work for you?
* Would the usage-in-generation one work for you?

I'm thinking we could even easily support both (as a setting when the managed string table gets constructed), but I don't want to get too feature-happy if nobody's interested in it yet.

(I think once we settle this, it does make sense to re-examine if we can clean up the code as much as possible, as Levi pointed out -- I'm happy to have as little custom fancy stuff as possible)

ivoanjo:
> If the program forks while the RwLock in the table is held, the child process will deadlock trying to access the table. I think we'd need to just make a new table on fork. Not the end of the world, but I guess just something to look out for.

Yeap, this is a good point. I think even without the lock, it's probably not a great idea to do anything other than throw away the table after the fork.

(It would, on the other hand, be cool to have a lock-free implementation for this, but I fear the cost would probably not be great).

Would having a way to clear the table ignoring the lock work as a temporary solution for Python? How do y'all handle this issue for the regular profile data structures? Reset on fork as well?

r1viollet:
Would everyone agree with merging this ? 
- It does not impact other user's workflows.
- It allows us to experiment / measure with this approach.

nsrip-dd:
> Would everyone agree with merging this ?

Yeah. I followed up on this with @ivoanjo on slack but forgot to update here. It turns out we don't need this right now for Python so nothing about this needs to change on our account :)

profiling-ffi/build.rs
```diff
@@ -5,6 +5,7 @@ extern crate build_common;
 use build_common::generate_and_configure_header;
 
 fn main() {
+    println!("cargo:rerun-if-changed=src");
```
morrisonlevi, line 8: Do you remember why you added this? Seems like an odd thing to have to add, it should already know to rebuild if one of its rust source files changes, right?

profiling-ffi/src/profiles.rs
morrisonlevi: What about id=0 always being the empty string? In many string and string table things I've messed around with in the past, this has been a desirable behavior.

ivoanjo: Yeap, that's what this PR is already doing. Even the FFI short-circuits 0 without going further and (safely) assumes that zero is empty and is not supposed to be interned/uninterned.

LMK if I've missed a spot!

profiling-ffi/build.rs
ivoanjo: My past self didn't include enough info in the commit message, boo past self! >_>

I seem to remember there was something I was changing and the headers weren't properly being updated. But... I don't remember the exact situation.

I'll push a commit to remove this for now -- either we'll discover it again, and properly document it, or we'll find out it's actually not needed. Either way, it avoids cargo-culting some voodoo because "it looked useful" some time in the past :)

ivoanjo: Removed in 172a0a20655f5a1e871a09614575f792b8c1ac65

profiling-ffi/src/string_storage.rs
```diff
@@ -0,0 +1,209 @@
...
+pub unsafe extern "C" fn ddog_prof_ManagedStringStorage_new() -> ManagedStringStorageNewResult {
+    let storage = InternalManagedStringStorage::new();
+
+    ManagedStringStorageNewResult::Ok(ManagedStringStorage {
+        inner: Rc::into_raw(Rc::new(RwLock::new(storage))) as *const c_void,
+    })
+}
+
+#[no_mangle]
+/// TODO: @ivoanjo Should this take a `*mut ManagedStringStorage` like Profile APIs do?
```
morrisonlevi, line 41: Ugh, this is something even in C I go back and forth on. It's one of those "do I trust the user or do I be more defensive?" things. Setting the C pointer to null makes it easier to debug when things go wrong, and can sometimes even prevent further things from going wrong. But it's theoretically wholly wasted work because nobody should use the thing after it's been dropped...

profiling-ffi/src/string_storage.rs
```diff
@@ -0,0 +1,209 @@
...
+#[repr(C)]
+#[allow(dead_code)]
+pub enum ManagedStringStorageInternResult {
+    Ok(ManagedStringId),
+    Err(Error),
+}
+
+#[must_use]
+#[no_mangle]
+/// TODO: Consider having a variant of intern (and unintern?) that takes an array as input, instead
```
morrisonlevi, line 57: Do you have a use case for this in your PoC for Ruby? If so, I'd do it, and if not, I'd pass.

profiling-ffi/src/string_storage.rs
```diff
@@ -0,0 +1,209 @@
...
+#[repr(C)]
+#[allow(dead_code)]
+pub enum StringWrapperResult {
+    Ok(StringWrapper),
+    Err(Error),
+}
+
+#[must_use]
+#[no_mangle]
+/// TODO: @ivoanjo It's not clear to me if the string pointer we return here is the exact one from
```
morrisonlevi, line 121: At least as written, it's a copy which is not owned/managed by the string storage.

profiling/src/collections/string_storage.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// The empty string is guaranteed to have an id of [StringId::ZERO].
+    fn intern(&mut self, item: Rc<str>) -> StringId;
+    fn get_string(&self, id: StringId) -> Rc<str>;
+    fn len(&self) -> usize;
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+    fn into_iter(self: Box<Self>) -> Box<dyn Iterator<Item = Rc<str>>>;
+    fn clone_empty(&self) -> Box<dyn StringStorage>;
+}
```
morrisonlevi, line 24: It seems like in the current PR, this is only implemented by stuff in `experimental`. Is that intended?

profiling-ffi/src/string_storage.rs
ivoanjo: I do! We're literally interning in a loop when I need to consume a whole stack. (Note: `intern_or_raise` here is just a nice helper to call `ddog_prof_ManagedStringStorage_intern` and check if there's an error in the result)

```c
heap_stack* heap_stack_new(heap_recorder *recorder, ddog_prof_Slice_Location locations) {
  uint16_t frames_len = locations.len;
  if (frames_len > MAX_FRAMES_LIMIT) {
    // This should not be happening anyway since MAX_FRAMES_LIMIT should be shared with
    // the stacktrace construction mechanism. If it happens, lets just raise. This should
    // be safe since only allocate with the GVL anyway.
    rb_raise(rb_eRuntimeError, "Found stack with more than %d frames (%d)", MAX_FRAMES_LIMIT, frames_len);
  }
  heap_stack *stack = ruby_xcalloc(1, sizeof(heap_stack) + frames_len * sizeof(heap_frame));
  stack->frames_len = frames_len;
  for (uint16_t i = 0; i < stack->frames_len; i++) {
    const ddog_prof_Location *location = &locations.ptr[i];
    stack->frames[i] = (heap_frame) {
      .name = intern_or_raise(recorder->string_storage, location->function.name),
      .filename = intern_or_raise(recorder->string_storage, location->function.filename),
      // ddog_prof_Location is a int64_t. We don't expect to have to profile files with more than
      // 2M lines so this cast should be fairly safe?
      .line = (int32_t) location->line,
    };
  }
  return stack;
}
```

(from my working branch which is based off of https://github.com/DataDog/dd-trace-rb/pull/3628 ).

My thinking is that, unlike most other libdatadog APIs where either a) Are very small but we don't call them very often (e.g. setup and reporting); b) We do a big chunk of work on every call (profile add), this API does c) Both very little work and gets called many times.

Thus, it seems like a prime candidate to turn C into B -> by having a more coarse-grained call that lowers the overhead cost of the ffi and locking.

profiling-ffi/src/string_storage.rs
ivoanjo: Ah, you're right! I had not seen/used the `StringWrapper` before and missed it was a feature that was already added in ddcommon-ffi and that there's an explicit `ddog_StringWrapper_drop` that goes along with it.

Since I had not seen a drop in the profiling ffi I was a bit confused about how this memory was being managed.

I've pushed d4194a2e78e47bae68d565d54fdb641a4b46d2cd to fix/update this comment.

profiling/src/collections/string_storage.rs
ivoanjo: Yup, indeed none of this is needed or in use right now. In the name of this PR having less unused leftover appendages I've just gone ahead and removed it in 28782b5d2d6ec3508e234792ec94fd1ab8fa4b10 .

profiling/src/collections/string_storage.rs
```diff
@@ -0,0 +1,148 @@
...
+                debug_assert_eq!(old_value, None);
+                id
+            }
+        }
+    }
+
+    pub fn unintern(&self, id: u32) {
+        if id == 0 {
+            panic!("For performance reasons, unintern should not be called with id == 0. Please hardcode a fast path check in the caller to avoid this call")
+        }
```
morrisonlevi, line 98: This can be removed right?

profiling/src/collections/string_storage.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        let data = self.get_data(id);
+        let usage_count = &data.usage_count;
+        usage_count.set(usage_count.get() - 1);
+    }
+
+    pub fn get_seq_num(&self, id: u32, profile_strings: &mut StringTable) -> StringId {
+        if id == 0 {
+            panic!("For performance reasons, get_set_num should not be called with id == 0. Please hardcode a fast path check in the caller to avoid this call")
+        }
```
morrisonlevi, line 108: This panic shouldn't exist either, right? I think avoiding it in the caller is probably okay since it's a `0` to `0` mapping without side effects, though.

profiling/src/collections/string_storage.rs
```diff
@@ -0,0 +1,148 @@
...
+
+use super::identifiable::StringId;
+use super::string_table::StringTable;
+
+#[derive(PartialEq, Debug)]
+struct ManagedStringData {
+    str: Rc<str>,
+    last_usage_gen: Cell<Option<u32>>, // TODO: This is not actually being used; maybe remove?
+    cached_seq_num_for: Cell<Option<*const StringTable>>,
+    cached_seq_num: Cell<Option<StringId>>,
```
morrisonlevi, line 18: We are storing the cache on each string, but aren't these all added in a batch to the same string table? I think you said that you add all these managed strings to the Profile's string table just before serialization, right? Couldn't we perform a larger batch operation and store the cache there? That way the memory is only used on serialization rather than kept around but largely not being used.

profiling/src/collections/string_storage.rs
ivoanjo: Actually, no!

This check is not about the empty string existing or not, it's about pushing callers to do the `if id == 0 return` checks as close to the client code as possible, rather than "eating" all the overhead of coming all the way up here (including all of the lock acquisition that it may entail) just to go "oh, it's a zero, nothing to do!".

So it's the equivalent of putting a big banner on the lobby saying "don't bother going up the stairs all the way to the top if you have an empty string, save yourself some work".

profiling/src/collections/string_storage.rs
ivoanjo: This is here for the same reason as above: not for correctness, but for performance -- callers must do their own checks for zero rather then letting it get to the string table.

This being rust, we can probably in the future be a bit fancier and encode this in a type that can never be zero ;)

profiling/src/collections/string_storage.rs
ivoanjo: As I as looking at this, I did a small tweak to store these as a tuple, rather than as separate entries, as they're related anyway -- 1f2b953cd3fd90df72f6f98a460a682420947b22 .

Your suggestion is interesting, but I'm curious how far were you thinking about the "large batch operation". In particular, were you thinking of moving the cache entirely away from the string table, to the profile? Or even to the caller of the profile?

profiling/src/collections/string_storage.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.next_id = self
+            .next_id
+            .checked_add(1)
+            .ok_or_else(|| anyhow::anyhow!("Ran out of string ids!"))?;
+        let old_value = self.str_to_id.insert(str.clone(), id);
+        debug_assert_eq!(old_value, None);
+        let old_value = self.id_to_data.insert(id, data);
+        debug_assert_eq!(old_value, None);
+        Ok(id)
+    }
```
morrisonlevi, line 93: What's the API level difference between these two methods? Also need to document when errors happen, and what maybe to do with them.

ivoanjo: Ah, nice catch, `intern_new` was fully 100% intended not to be public, yet somehow the `pub` ended up there.

It's not intended to be exposed (too dangerous!) -- I'll remove it.

ivoanjo: Fixed in 20b946994fa2047d415429d8d65122d1b64a22f5

morrisonlevi (commented):
Not quite finished but publishing what I have.

profiling-ffi/src/profiles.rs
```diff
@@ -244,6 +245,22 @@ impl<'a> TryFrom<&'a Mapping<'a>> for api::Mapping<'a> {
     }
 }
 
+impl<'a> TryFrom<&'a Mapping<'a>> for api::StringIdMapping {
```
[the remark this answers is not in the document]
ivoanjo: Fixed in f4df25400154373e915287e6341f269adbdf3e4b

profiling-ffi/src/profiles.rs
```diff
@@ -278,13 +295,18 @@ impl<'a> TryFrom<&'a Function<'a>> for api::Function<'a> {
     }
 }
 
-impl<'a> TryFrom<&'a Line<'a>> for api::Line<'a> {
+impl<'a> TryFrom<&'a Function<'a>> for api::StringIdFunction {
```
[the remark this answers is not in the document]
ivoanjo: Fixed in f4df25400154373e915287e6341f269adbdf3e4b

profiling-ffi/src/profiles.rs
```diff
@@ -327,6 +364,33 @@ impl<'a> TryFrom<&'a Label<'a>> for api::Label<'a> {
     }
 }
 
+impl<'a> TryFrom<&'a Label<'a>> for api::StringIdLabel {
```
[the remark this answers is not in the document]
ivoanjo: Fixed in f4df25400154373e915287e6341f269adbdf3e4b

profiling-ffi/src/profiles.rs
```diff
@@ -352,6 +416,31 @@ impl<'a> TryFrom<Sample<'a>> for api::Sample<'a> {
     }
 }
 
+impl TryFrom<Sample<'_>> for api::StringIdSample {
```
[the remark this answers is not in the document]
ivoanjo: Fixed in f4df25400154373e915287e6341f269adbdf3e4b

profiling-ffi/src/string_storage.rs
```diff
@@ -0,0 +1,215 @@
...
+    })()
+    .context("ddog_prof_ManagedStringStorage_advance_gen failed");
+
+    match result {
+        Ok(_) => MaybeError::None,
+        Err(e) => MaybeError::Some(e.into()),
+    }
+}
+
+pub unsafe fn get_inner_string_storage(
```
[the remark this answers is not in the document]
ivoanjo: Fixed in 557633173dc1c315c8b296b9333a63d6562fe1da

profiling-ffi/src/string_storage.rs
morrisonlevi: I still don't have the "answer" to this one. All I can say is that I made the others re-assign the pointer null on drop when it was feasible, so I clearly thought at the time that it was a good idea.

ivoanjo:
> Is this still considered "experimental"? What's the rough plan to either revert it or make it non-experimental?

Good question. I guess my current thinking about calling it "experimental" is because it's a feature we may end up iterating a lot on, including breaking APIs and whatnot, so if anyone's building on top of this, I was trying to get that across.

Since the early benchmarks we had for Ruby showed good results for this I wouldn't expect this to be entirely reverted; at most maybe we could throw it in a Ruby-specific folder if it turns out this use-case is too specific and nobody else needs it.

> Asking mostly because it it does intrude, albeit only a little, onto the FFI structs of the "main" thing.

My thinking is that it's still harmless, even if ffi callers don't properly zero out those fields, for two reasons:

1. The code that picks whether to use ids at the ffi level goes "is there a regular string? yes, then use that and don't look at the id"
2. If somehow we missed a spot and somehow the garbage ids get accidentally used instead of strings, then, because there's no string table installed (e.g. because only Ruby will have one), the FFI will report a clear "there's no string table" error, and we can use that to spot the issue.

To be honest I'm somewhat more annoyed about the duplication in the profile implementation, but there, because I think the right solution is to have separate structures (thus enforcing that you either have ids OR strings), the only way of avoiding the duplication would be to introduce a bunch of macros to "hide" the fact that we'd have two versions of the code with only the very minor differences. (At least given the current API....)

> What's the rough plan [...]

I'd say:

1. Get this out for use in a stable libdatadog version
2. If we spot any issues with other libraries caused by these APIs, change them
3. Consider it no longer experimental once we've validated it and put out the first Ruby profiler version that's using this for heap profiling (although still very open for feedback/changes, if anyone also wants to use it)

> Approved in general, I've blocked this long enough I think ^_^

cc @danielsn any concerns with me going ahead and merging the PR?

profiling/src/internal/profile/mod.rs
```diff
@@ -137,6 +192,28 @@ impl Profile {
...
+        let string_id = self.string_storage
+            .as_ref()
+            // Safety: We always get here through a direct or indirect call to add_string_id_sample,
+            // which already ensured that the string storage exists.
+            .ok_or_else(|| anyhow::anyhow!("Current sample makes use of ManagedStringIds but profile was not created using a string table"))?
+            .read()
+            // Safety: This failure is unlikely as it only happens if the lock is poisoned (and for
+            // the lock to become poisoned, another unlikely failure already happened
+            // before)
+            .expect("acquisition of read lock on string storage should succeed")
```
danielsn, line 211: Still might as well return an error rather than kill the process

danielsn (approved):

profiling-ffi/src/profiles.rs
```diff
@@ -141,35 +146,29 @@ pub struct Label<'a> {
     /// units and units like "seconds" and "nanoseconds" as time units,
     /// and apply appropriate unit conversions to these.
     pub num_unit: CharSlice<'a>,
+    pub num_unit_id: ManagedStringId,
```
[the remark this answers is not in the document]
ivoanjo: We spoke about this directly, but capturing here that right now it was an explicit decision to not make this change: we can indeed use an enum (and probably pack it tighter!) but for this PR we're making the trade-off of avoiding this breaking change due to the experimental "we're still feeling our way towards exactly where we want to end up, API-wise" vibe of this PR.

profiling-ffi/src/profiles.rs
```diff
@@ -327,6 +358,31 @@ impl<'a> TryFrom<&'a Label<'a>> for api::Label<'a> {
...
 
+impl<'a> From<&'a Label<'a>> for api::StringIdLabel {
+    fn from(label: &'a Label<'a>) -> Self {
+        let key = ManagedStringId::new(label.key_id.value);
+        let str = label.str_id.value;
+        let str = if str == 0 {
+            None
+        } else {
+            Some(ManagedStringId::new(str))
+        };
```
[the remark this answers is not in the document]
ivoanjo: This is because in the `profiling` the API is

```
pub struct Label<'a> {
    pub key: &'a str,

    /// At most one of the following must be present
    pub str: Option<&'a str>,
    pub num: i64,

    /// Should only be present when num is present.
    /// Specifies the units of num.
    /// Use arbitrary string (for example, "requests") as a custom count unit.
    /// If no unit is specified, consumer may apply heuristic to deduce the unit.
    /// Consumers may also  interpret units like "bytes" and "kilobytes" as memory
    /// units and units like "seconds" and "nanoseconds" as time units,
    /// and apply appropriate unit conversions to these.
    pub num_unit: Option<&'a str>,
}
```

(the variant with ids replaces the `str` with an id).

That is, while at the ffi level we use "empty string" to mean no str is provided, the `profiling` crate actually uses an `Option` to make this cleaner. Thus, we need here to make this conversion between both APIs.

profiling-ffi/src/string_storage.rs
```diff
@@ -0,0 +1,215 @@
...
+    Ok(ManagedStringId),
+    Err(Error),
+}
+
+#[must_use]
+#[no_mangle]
+/// TODO: Consider having a variant of intern (and unintern?) that takes an array as input, instead
+/// of just a single string at a time.
+/// TODO: @ivoanjo Should this take a `*mut ManagedStringStorage` like Profile APIs do?
+pub unsafe extern "C" fn ddog_prof_ManagedStringStorage_intern(
```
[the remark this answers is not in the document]
ivoanjo: I'm not sure they'd be especially helpful here.

In particular, the only comment I can think of is "the charslice needs to be valid or empty, and the managed string storage needs to be valid or null", which... IDK... seems to describe every function in our ffi that takes a pointer? :eyes:

profiling-ffi/src/string_storage.rs
```diff
@@ -0,0 +1,215 @@
...
+#[no_mangle]
+/// Returns a string given its id.
+/// This API is mostly for testing, overall you should avoid reading back strings from libdatadog
+/// once they've been interned and should instead always operate on the id.
+/// Remember to `ddog_StringWrapper_drop` the string once you're done with it.
+/// TODO: @ivoanjo Should this take a `*mut ManagedStringStorage` like Profile APIs do?
+pub unsafe extern "C" fn ddog_prof_ManagedStringStorage_get_string(
+    storage: ManagedStringStorage,
+    id: ManagedStringId,
+) -> StringWrapperResult {
```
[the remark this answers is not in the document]
ivoanjo: Cleaning up this string becomes the responsibility of the caller; there's a comment above saying:

> ``/// Remember to `ddog_StringWrapper_drop` the string once you're done with it.``

profiling/src/collections/string_storage.rs
```diff
@@ -0,0 +1,144 @@
...
+
+#[derive(PartialEq, Debug)]
+struct ManagedStringData {
+    str: Rc<str>,
+    cached_seq_num: Cell<Option<(*const StringTable, StringId)>>,
+    usage_count: Cell<u32>,
+}
+
+pub struct ManagedStringStorage {
+    next_id: u32,
```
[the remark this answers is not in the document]
ivoanjo: We can -- the `intern` operation of the managed string storage takes care of turning those events into errors:

```
        self.next_id = self
            .next_id
            .checked_add(1)
            .ok_or_else(|| anyhow::anyhow!("Ran out of string ids!"))?;
```

Currently it's up to the library to decide what to do without help from libdatadog; in the future my thinking is we could have a function to create a new managed string storage from the old one (starting the ids from 0 again), plus some function that would map the old ids to the new ids.

(This would be basically a once-in-a-blue-moon manual compaction step).

profiling/src/internal/profile/mod.rs
ivoanjo: Fixed in 432c12e0455e2f3784dd63aec7ed4aba26742fba

profiling-ffi/src/string_storage.rs
```diff
@@ -0,0 +1,215 @@
...
+#[no_mangle]
+/// TODO: Consider having a variant of intern (and unintern?) that takes an array as input, instead
+/// of just a single string at a time.
+/// TODO: @ivoanjo Should this take a `*mut ManagedStringStorage` like Profile APIs do?
+pub unsafe extern "C" fn ddog_prof_ManagedStringStorage_intern(
+    storage: ManagedStringStorage,
+    string: CharSlice,
+) -> ManagedStringStorageInternResult {
+    // Empty strings always get assigned id 0, no need to check.
+    if string.len() == 0 {
```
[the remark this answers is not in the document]
ivoanjo: Fixed in 5c75aa7b2cb13ef0893e262fc042493537059def

profiling-ffi/src/profiles.rs
```diff
@@ -380,6 +459,30 @@ pub unsafe extern "C" fn ddog_prof_Profile_new(
...
+    let period = period.map(Into::into);
+    let string_storage = match get_inner_string_storage(string_storage, true) {
+        Ok(string_storage) => string_storage,
+        Err(err) => return ProfileNewResult::Err(err.into()),
+    };
+
+    let internal_profile =
+        internal::Profile::with_string_storage(start_time, &types, period, string_storage);
+    let ffi_profile = Profile::new(internal_profile);
+    ProfileNewResult::Ok(ffi_profile)
```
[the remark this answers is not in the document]
ivoanjo: Fixed in 42d9408a6a904a9b82fbee30b9d6b55e8f581e41

profiling-ffi/src/profiles.rs
```diff
@@ -352,6 +408,29 @@ impl<'a> TryFrom<Sample<'a>> for api::Sample<'a> {
     }
 }
 
+impl From<Sample<'_>> for api::StringIdSample {
+    fn from(sample: Sample<'_>) -> Self {
+        let mut locations: Vec<api::StringIdLocation> = Vec::with_capacity(sample.locations.len());
+
+        for location in sample.locations.as_slice().iter() {
+            locations.push(location.into())
+        }
```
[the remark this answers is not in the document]
ivoanjo: Yup, fixed in 98cf24a6edbdc17bc4e10c4e7c2505d127eee1f1

profiling-ffi/src/profiles.rs
```diff
@@ -244,6 +245,20 @@ impl<'a> TryFrom<&'a Mapping<'a>> for api::Mapping<'a> {
     }
 }
 
+impl<'a> From<&'a Mapping<'a>> for api::StringIdMapping {
+    fn from(mapping: &'a Mapping<'a>) -> Self {
+        let filename = ManagedStringId::new(mapping.filename_id.value);
+        let build_id = ManagedStringId::new(mapping.build_id_id.value);
```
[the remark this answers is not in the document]
ivoanjo: In hindsight the conversion wasn't even needed, so I cleaned it up in 0cb84570b7ac6f08159fb7f329aef2fc75eb5ca4 . Thanks for poking me, I had missed it ;)

profiling-ffi/src/profiles.rs
```diff
@@ -244,6 +245,20 @@ impl<'a> TryFrom<&'a Mapping<'a>> for api::Mapping<'a> {
     }
 }
 
+impl<'a> From<&'a Mapping<'a>> for api::StringIdMapping {
+    fn from(mapping: &'a Mapping<'a>) -> Self {
+        let filename = ManagedStringId::new(mapping.filename_id.value);
```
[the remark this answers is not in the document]
ivoanjo: Actually -- this "conversion" was even not needed -- we're already getting a `ManagedStringId` so we were converting a struct to itself. I've simplified it in 0cb84570b7ac6f08159fb7f329aef2fc75eb5ca4 .

Still, my previous answer I think is still relevant -- we're still not checking the managed ids here, whatever we get as ain input, but always check them later on.

ivoanjo:
I think the commit history for this PR is worth preserving. I'm preparing to merge this to master with a regular merge commit (not with a squash). As part of it, I'll push force a rebase so that there's no more "merge from master" commits.

PR #726, opened 2024-11-11
Add more siginfo_t context 

sanchda:
# What does this PR do?

* Standardize code paths that took a signum to take a siginfo_t
* Add tests to ensure serialization works as intended
* Add libc status code -> string helpers

This adds some more siginfo_t context to crashes.  In addition to the SIGBUS codes, I added some logic to make sense of things like SIGSYS, since I would like to add support for that soon.

One possible point of contention that I was unable to resolve:  currently, we emit a `faulting_address`, which is a representation of the `si_addr` from the  `siginfo_t` only when the fault is a segfault.  This makes a lot of sense in the current product, which does no special normalization in the telemetry backend.  However, from a data transmission perspective, I think it makes more sense to represent the struct-as-it-is-written and defer to the backend for normalization.  Moreover, `si_addr` has context outside of `SIGSEGV`.

This decision duplicates an address for now.  Oh well.

# Motivation

While exploring the behavior of crashtracking for system tests, and as part of my experiments in other branches, I started relying on `strace`'s comprehension for `siginfo_t`.  We also had a recent bug that could really only have been comprehended with content from `siginfo_t` (stack overflow by hitting the crashtracking guard page).

I think this information is really important.


crashtracker/src/collector/crash_handler.rs
```diff
@@ -509,8 +501,7 @@ fn handle_posix_signal_impl(signum: i32, sig_info: *mut siginfo_t) -> anyhow::Re
         config,
         config_str,
         metadata_string,
-        signum,
-        faulting_address,
+        unsafe { *siginfo },
```
danielsn, line 504: Calling by value not reference?

bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -121,9 +121,13 @@ fn test_crash_tracking_bin(crash_tracking_receiver_profile: BuildProfile, mode:
     );
     assert_eq!(
         serde_json::json!({
-          "signum": 11,
-          "signame": "SIGSEGV",
+          "code": 1,
+          "codename": "SEGV_MAPERR",
           "faulting_address": 0,
+          "pid": 0,
```
danielsn, line 127: Only collect the fields that are relevent for the given signal https://man7.org/linux/man-pages/man2/sigaction.2.html

danielsn:
This should be synchronized with RFC5.

PR #732, opened 2024-11-12
increase the volume of traces for deserialization benchmarks

ekump:
# What does this PR do?

Previously, we were performing benchmarks for msgpack deserialization on 2 spans which is both unrealistic, and contributed to variance in benchmarks. This PR changes the payload to approximately 10mb of data (~40k spans). 

# Additional Notes

The benchmark report for this PR can be ignored since the baseline was for just 2 spans.

### DIFF
--- a/trace-utils/benches/deserialization.rs
+++ b/trace-utils/benches/deserialization.rs
@@ -6,39 +6,64 @@ use datadog_trace_utils::tracer_header_tags::TracerHeaderTags;
 use datadog_trace_utils::tracer_payload::{
     DefaultTraceChunkProcessor, TraceEncoding, TracerPayloadCollection, TracerPayloadParams,
 };
-use serde_json::json;
+use serde_json::{json, Value};
+
+fn generate_spans(num_spans: usize, trace_id: u64) -> Vec<Value> {
+    let mut spans = Vec::with_capacity(num_spans);
+    let root_span_id = 100_000_000_000 + (trace_id % 1_000_000);
+
+    for i in 0..num_spans {
+        // If it's the first span make it the root
+        let span_id = if i == 0 {
+            root_span_id
+        } else {
+            root_span_id + i as u64 + 1
+        };
+
+        // if it's not the root, then give it the root as a parent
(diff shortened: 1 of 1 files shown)


trace-utils/benches/deserialization.rs
```diff
@@ -6,39 +6,62 @@ use datadog_trace_utils::tracer_header_tags::TracerHeaderTags;
...
-        "parent_id": 100,
-        "start": 1,
-        "duration": 5,
-        "error": 0,
-        "meta": {},
-        "metrics": {},
-    }]);
+fn generate_spans(num_spans: usize, trace_id: u64) -> Vec<Value> {
+    let mut spans = Vec::with_capacity(num_spans);
+    let root_span_id = 100000000000 + (trace_id % 1000000);
```
ajgajg1134, line 13: nit
```suggestion
    let root_span_id = 100_000_000_000 + (trace_id % 1_000_000);
```

trace-utils/benches/deserialization.rs
```diff
@@ -6,39 +6,62 @@ use datadog_trace_utils::tracer_header_tags::TracerHeaderTags;
...
-        "span_id": 333,
-        "parent_id": 100,
-        "start": 1,
-        "duration": 5,
-        "error": 1,
-        "meta": {},
-        "metrics": {},
-    }]);
+    for i in 0..num_chunks {
+        let trace_id = 100000000000 + i as u64;
```
ajgajg1134, line 54: nit
```suggestion
        let trace_id = 100_000_000_000 + i as u64;
```

ajgajg1134 (approved):

bantonsson (changes requested):
So the fact that you've massively increased the data we are deserializing, and there were no measurement changes even though the test doesn't measure per chunk or per span made me suspicious, and the test is now flagged as _unstable_ and excluded. I don't like that the comparison doesn't show which tests are _unstable_. I'll post info on how to look at the raw results in slack.

bantonsson (approved):
I ran the benchmark twice with this branch as the base, and it was marked as stable and had no performance changes.

PR #738, opened 2024-11-13
[crashtracker] Implement RFC 0005

danielsn:
# What does this PR do?

* Adds a new datatype which implements RFC 0005.
* Adds a translation from the old to the new data-format.
* the `file://` output type now outputs RFC0005 format as well as the old format.

# Motivation

This is step 1 in implementing the RFC.

# Additional Notes

This is step 1 of a multi-step program to move crashtracker over to RFC5.  
For now, both dataformats will co-exist.
Future PRs will switch various parts of the API over to use the new format.
Once this is done, we can retire the old format.

# How to test the change?

### DIFF
--- a/crashtracker/src/rfc5_crash_info/error_data.rs
+++ b/crashtracker/src/rfc5_crash_info/error_data.rs
@@ -1,0 +1,64 @@
+use std::collections::HashMap;
+
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+use super::stacktrace::StackTrace;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
+pub struct ErrorData {
+    pub is_crash: bool,
+    pub kind: ErrorKind,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub message: Option<String>,
+    pub source_type: SourceType,
+    pub stack: StackTrace,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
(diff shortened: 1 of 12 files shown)


crashtracker/src/rfc5_crash_info/error_data.rs
```diff
@@ -0,0 +1,64 @@
+use std::collections::HashMap;
```
gleocadie, line 1: just out of curiosity, shouldn't this be after the header ?

crashtracker/src/rfc5_crash_info/stacktrace.rs
```diff
@@ -0,0 +1,178 @@
...
+                    ),
+                    crate::NormalizedAddressMeta::Elf { path, build_id } => (
+                        byte_vec_as_hex(build_id),
+                        Some(BuildIdType::GNU),
+                        Some(FileType::ELF),
+                        Some(path.to_string_lossy().to_string()),
+                    ),
+                    crate::NormalizedAddressMeta::Pdb { path, guid, age } => (
+                        Some(format!(
+                            "{}:{age}",
```
gleocadie, line 44: you can remove the ":".
From what I saw, Windbg will collect the guid and age and concatenate them together without a separator. The request to symbol server is done with the result.

gleocadie (approved):
Overall looks good to me

crashtracker/src/rfc5_crash_info/stacktrace.rs
danielsn: https://github.com/DataDog/libdatadog/issues/739

crashtracker/src/rfc5_crash_info/error_data.rs
danielsn: Yes.  OOPS

PR #744, opened 2024-11-15
Add header so we get real http status codes from the trace agent (APMSP-1418)

ajgajg1134:
# What does this PR do?

Set the `Datadog-Send-Real-Http-Status` header to indicate to the agent that we are able to handle 429 responses and will appropriately back off and retry. Due to legacy compatibility reasons the default behavior of the trace agent for trace payloads is to always respond with a `200` status code even if the trace-agent detects it is overwhelmed and will drop the payload. By setting this header the trace-agent will instead respond with a 429 giving us an opportunity to retry and decrease the likelihood of lost data.

# Motivation
We have retry logic but we aren't using it :(

# Additional Notes

# How to test the change?

This change is well covered by the unit tests


ekump (approved):
LGTM, I wish we could validate the headers in our test-agent snapshot test, but I don't think that functionality is supported at the moment.

PR #746, opened 2024-11-18
Allow dynamic linking on Windows

gleocadie:
# What does this PR do?

Update nuget package to be able to link .NET profiler dynamically against libdatadog.

# Motivation

The package contained the dll but not the `lib` required to link dynamically in VS. The other `lib` file was used to link statically.

# How to test the change?

The package is tested in gitlab CI. If the build is green, it means the linking was correct.
I also download the nuget packages and tested it on my machine.

### DIFF
--- a/windows/libdatadog.csproj
+++ b/windows/libdatadog.csproj
@@ -39,6 +39,8 @@
       PackagePath="include\native\datadog\data-pipeline.h" />
 
     <None Include="$(LibDatadogBinariesOutputDir)\x86_64-pc-windows-msvc\debug\datadog_profiling_ffi.lib"
+      Pack="true" PackagePath="build\native\lib\x64\debug\static\datadog_profiling_ffi.lib" />
+    <None Include="$(LibDatadogBinariesOutputDir)\x86_64-pc-windows-msvc\debug\datadog_profiling_ffi.dll.lib"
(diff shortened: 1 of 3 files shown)


hoolioh (approved):

windows/libdatadog.csproj
```diff
@@ -39,27 +39,35 @@
       PackagePath="include\native\datadog\data-pipeline.h" />
 
     <None Include="$(LibDatadogBinariesOutputDir)\x86_64-pc-windows-msvc\debug\datadog_profiling_ffi.lib"
+      Pack="true" PackagePath="build\native\lib\x64\debug\static\datadog_profiling_ffi.lib" />
+    <None Include="$(LibDatadogBinariesOutputDir)\x86_64-pc-windows-msvc\debug\datadog_profiling_ffi.dll.lib"
```
ganeshnj, line 43: /question

what's the intention behind the extension change? Name collision?

ganeshnj (approved):

windows/libdatadog.csproj
gleocadie: The `dll.lib` is produced by cargo.
Since the static lib `*.lib` is produced in the same folder as the dynamic ones `*.lib + *.dll`, so yes name collision avoiding.

Issue #727, opened 2024-11-11
[crashtracker] add source_type to top level telemetry data


danielsn:
It would be nice to have this information at the top level in telemetry payload. I noticed that there is a [Origin](https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/header.go#L87) field that could be used for this.

_Originally posted by @nsavoire in https://github.com/DataDog/libdatadog/pull/554#discussion_r1796679150_

PR #747, opened 2024-11-19
[telemetry] Add missing origin field

danielsn:
# What does this PR do?

Adds the `origin` field, [defined in the backend](https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/header.go#L88) but not currently implemented in libdatadog.

# Motivation

#727 We'd like to use this field in the crashtracker.

# Additional Notes

* I put placeholder values in for the different uses.  LMK if I should change them?
* The `ddcommon` request builder currently just sets "unknown".  I could plumb through an option to set a value, but this might be a breaking change.  Thoughts?

# How to test the change?

### DIFF
--- a/ddtelemetry/src/worker/mod.rs
+++ b/ddtelemetry/src/worker/mod.rs
@@ -636,7 +636,7 @@ impl TelemetryWorker {
             runtime_id: &self.runtime_id,
             seq_id,
             host: &self.data.host,
-            origin: "unknown",
+            origin: None,
             application: &self.data.app,
             payload,
         };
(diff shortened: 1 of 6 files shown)


sanchda (approved):

ekump (approved):

nsavoire (approved):

ddtelemetry/src/worker/mod.rs
```diff
@@ -636,6 +636,7 @@ impl TelemetryWorker {
             runtime_id: &self.runtime_id,
             seq_id,
             host: &self.data.host,
+            origin: "unknown",
             application: &self.data.app,
```
paullegranddc, line 640: Could you leave it empty if it's not available?

danielsn: Done

PR #748, opened 2024-11-19
refactor: ddcommon-net+profiling with hyper 1

morrisonlevi:
# What does this PR do?

1. There is a new crate `ddcommon-net`, naming suggestions welcome. It has bits of `ddcommon` networking code which have been upgraded to hyper 1. Note it's not a direct 1:1 because hyper 1 changed a lot.
2. The `datadog-profiling` crate has been updated to use it. Currently its tests pass for me locally. The `datadog-profiling-ffi` is currently broken.
3. Other crates are not touched

# Motivation

The main motivation is to upgrade hyper from v0.14 to v1. There are a _lot_ of changes when moving to v1. Some things do not have direct replacements, so upgrading all crates at once is difficult. I decided (with input from a few others) that it seemed sensible to try making a `ddcommon-net` crate that uses hyper 1, and then crates can migrate one-by-one from `ddcommon` to `ddcommon-net`.

# Additional Notes

Please tell me if you have concerns about the approach, technical or otherwise. Here are some of mine:

1. Since `datadog-profiling` is the only crate using this for now, is `ddcommon-net` _really_ going to be common? Maybe we should make a `net` module inside of `datadog-profiling` instead and only extract a crate when there is at least one other crate migrated to hyper 1?
2. There is some of copy/paste in `ddcommon-net::compat` and `ddcommon`, mostly around Uri and Endpoint related things. Probably okay, not sure.

# How to test the change?

Don't. This is a WIP, I'll update this when it's ready for testing.


profiling/src/exporter/config.rs
```diff
@@ -1,76 +1,48 @@
...
-    api_key: IntoCow,
-) -> anyhow::Result<Endpoint> {
-    let intake_url: String = format!("https://intake.profile.{}/api/v2/profile", site.as_ref());
-
-    Ok(Endpoint {
-        url: Uri::from_str(intake_url.as_str())?,
-        api_key: Some(api_key.into()),
-        ..Default::default()
-    })
+pub trait ProfilingEndpoint {
```
danielsn, line 10: Not for this PR, but why does this need to be its own thing different from the `ddcommon::Endpoint`

danielsn, line 10: Why is this a trait?

ddcommon-net/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Note that currently it will still use a thread pool for certain operations
+/// which block.
+pub mod rt {
+    use std::io;
+    use tokio::runtime;
+    use tokio_util::sync::CancellationToken;
+
+    /// Creates a tokio runtime for the current thread. This is the expected
+    /// way to create a runtime used by this crate.
+    pub fn create_current_thread_runtime() -> io::Result<runtime::Runtime> {
```
danielsn, line 33: Should we stash this in a thread_local?

ddcommon-net/src/http.rs
```diff
@@ -0,0 +1,287 @@
...
+    sender.send_request(request).await
+}
+
+pub fn parse_path_from_uri(uri: &Uri) -> io::Result<path::PathBuf> {
+    // This _should_ be a redundant check, caller should only call this if
+    // they expect it's a unix domain socket or windows named pipe.
+    if uri.scheme_str() != Some("unix") || uri.scheme_str() != Some("windows") {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "URI scheme must be unix or windows",
```
danielsn, line 260:
```suggestion
            format!("URI scheme must be unix or windows: was {}", uri.scheme_str()),
```

ddcommon-net/src/http.rs
```diff
@@ -0,0 +1,287 @@
...
+    B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+{
+    let uri = request.uri();
+
+    // This _should_ be a redundant check, caller should only call this if
+    // they expect it's an https connection to begin with.
+    if uri.scheme_str() != Some("https") {
+        return Err(Error::Io(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "URI scheme must be https",
```
danielsn, line 209:
```suggestion
            format!("URI scheme must be https, was {}", uri.scheme_str()),
```

ddcommon-net/src/http.rs
```diff
@@ -0,0 +1,287 @@
...
+    B::Data: Send,
+    B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+{
+    let uri = request.uri();
+    match uri.scheme() {
+        None => Err(Error::UnsupportedScheme(String::new())),
+        Some(scheme) => match scheme.as_str() {
+            "http" => send_http(request).await,
+            "https" => send_https(request).await,
+            #[cfg(unix)]
```
danielsn, line 133: Do we eventually want to support this on Windows? https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

ddcommon-net/src/compat.rs
```diff
@@ -0,0 +1,135 @@
...
+            url: uri,
+            api_key: None,
+            timeout_ms: 0,
+            test_token: None,
+        }
+    }
+}
+
+pub mod header {
+    #![allow(clippy::declare_interior_mutable_const)]
```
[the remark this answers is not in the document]
morrisonlevi: compat stuff is generally just copied from the `ddcommon` package, as I didn't want to introduce a dependency there. I think maybe we could, but not sure. Thoughts?

ddcommon-net/src/lib.rs
morrisonlevi: For now, I'll say no. At least currently the profiler stores an owned one, and haven't needed it anywhere else.

profiling/src/exporter/config.rs
morrisonlevi: It's the "extension" trait pattern to add new methods to existing types. In this case, `Endpoint::profiling_agent`. We don't _have_ to do it this way. Is there something else you would prefer?

ddcommon-net/src/compat.rs
```diff
@@ -0,0 +1,135 @@
...
+
+impl Endpoint {
+    /// Default value for the timeout field in milliseconds.
+    pub const DEFAULT_TIMEOUT: u64 = 3_000;
+
+    /// Return a request builder with the following headers:
+    /// - User agent
+    /// - Api key
+    /// - Container Id/Entity Id
+    pub fn into_request_builder(
```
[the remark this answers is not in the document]
morrisonlevi: I've fixed this by renaming it `to_` instead.

profiling/src/exporter/config.rs
morrisonlevi: I renamed this to `EndpointExt` to hopefully be more clear about that.

ddcommon-net/src/http.rs
morrisonlevi: Since `scheme_str` is actually `Option<&str>`, my commit is a bit different but I improved these, thanks!

PR #749, opened 2024-11-20
Bump v14.2.0

gleocadie:
Bump to v14.2.0. It does not look like there is interface changes. Nuget package is updated but not the interface used.


morrisonlevi (approved):
I checked the commit log and this also looks good to me 👍🏻

PR #753, opened 2024-11-21
Initial move to use SendData in trace exporter (APMSP-1586)

ajgajg1134:
# What does this PR do?
Moves TraceExporter to use trace utils send data functionality.

# Motivation
We want access to the retry functionality of trace utils as well as to start supporting TinyBytes so we have good performance here

# Additional Notes

There are a number of things to do after this is merged that have been commented out explicitly so that we can unblock others working in this area. (e.g. client stats)

# How to test the change?

Covered with unit and integration tests for now

### DIFF
--- a/data-pipeline/src/health_metrics.rs
+++ b/data-pipeline/src/health_metrics.rs
@@ -7,7 +7,8 @@ pub(crate) const STAT_SEND_TRACES: &str = "datadog.libdatadog.send.traces";
 pub(crate) const STAT_SEND_TRACES_ERRORS: &str = "datadog.libdatadog.send.traces.errors";
 pub(crate) const STAT_DESER_TRACES: &str = "datadog.libdatadog.deser_traces";
 pub(crate) const STAT_DESER_TRACES_ERRORS: &str = "datadog.libdatadog.deser_traces.errors";
-// pub(crate) const STAT_SER_TRACES_ERRORS: &str = "datadog.libdatadog.ser_traces.errors";
+#[allow(dead_code)] // TODO (APMSP-1584) Add support for health metrics when using trace utils
+pub(crate) const STAT_SER_TRACES_ERRORS: &str = "datadog.libdatadog.ser_traces.errors";
 
 pub(crate) enum HealthMetric {
     Count(&'static str, i64),
--- a/data-pipeline/src/health_metrics.rs
+++ b/data-pipeline/src/health_metrics.rs
@@ -7,7 +7,7 @@ pub(crate) const STAT_SEND_TRACES: &str = "datadog.libdatadog.send.traces";
 pub(crate) const STAT_SEND_TRACES_ERRORS: &str = "datadog.libdatadog.send.traces.errors";
 pub(crate) const STAT_DESER_TRACES: &str = "datadog.libdatadog.deser_traces";
 pub(crate) const STAT_DESER_TRACES_ERRORS: &str = "datadog.libdatadog.deser_traces.errors";
-pub(crate) const STAT_SER_TRACES_ERRORS: &str = "datadog.libdatadog.ser_traces.errors";
+// pub(crate) const STAT_SER_TRACES_ERRORS: &str = "datadog.libdatadog.ser_traces.errors";
 
 pub(crate) enum HealthMetric {
     Count(&'static str, i64),
--- a/data-pipeline/src/trace_exporter.rs
+++ b/data-pipeline/src/trace_exporter.rs
@@ -7,7 +7,6 @@ use crate::{
 };
 use arc_swap::{ArcSwap, ArcSwapOption};
 use bytes::Bytes;
-// use datadog_trace_protobuf::pb;
 use datadog_trace_utils::trace_utils::{self, SendData, TracerHeaderTags};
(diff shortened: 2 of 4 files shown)


data-pipeline/src/health_metrics.rs
```diff
@@ -7,7 +7,7 @@ pub(crate) const STAT_SEND_TRACES: &str = "datadog.libdatadog.send.traces";
 pub(crate) const STAT_SEND_TRACES_ERRORS: &str = "datadog.libdatadog.send.traces.errors";
 pub(crate) const STAT_DESER_TRACES: &str = "datadog.libdatadog.deser_traces";
 pub(crate) const STAT_DESER_TRACES_ERRORS: &str = "datadog.libdatadog.deser_traces.errors";
-pub(crate) const STAT_SER_TRACES_ERRORS: &str = "datadog.libdatadog.ser_traces.errors";
+// pub(crate) const STAT_SER_TRACES_ERRORS: &str = "datadog.libdatadog.ser_traces.errors";
```
ekump, line 10: Super minor nit, but I slightly prefer this. If you agree, you could do the same for the constants in `trace_exporter`.

```suggestion
#[warn(dead_code)]
pub(crate) const STAT_SER_TRACES_ERRORS: &str = "datadog.libdatadog.ser_traces.errors";
```

data-pipeline/src/trace_exporter.rs
```diff
@@ -99,59 +97,59 @@ fn add_path(url: &Uri, path: &str) -> Uri {
     Uri::from_parts(parts).unwrap()
 }
 
-struct DroppedP0Counts {
-    pub dropped_p0_traces: usize,
-    pub dropped_p0_spans: usize,
-}
+// struct DroppedP0Counts {
```
ekump, line 100: I would recommend for any blocks of code that we're temporarily commenting out we have a `// TODO: APMSP-XXXX re-enable CSS` comment, or something along those lines to make sure we don't miss anything when we add back the CSS functionality. 

Also you can use `/*` and `*/` for block comments instead of each line being a `//`. It's a little bit cleaner.

data-pipeline/src/trace_exporter.rs
```diff
@@ -254,7 +253,9 @@ impl TraceExporter {
         self.check_agent_info();
         match self.input_format {
             TraceExporterInputFormat::Proxy => self.send_proxy(data, trace_count),
-            TraceExporterInputFormat::V04 => self.send_deser_ser(data),
+            TraceExporterInputFormat::V04 => {
+                self.send_deser_ser(tinybytes::Bytes::copy_from_slice(data))
+            } //todo do we want to copy?
```
ekump, line 258:
```suggestion
// TODO: APMSP-1582 - Refactor data-pipeline-ffi so we can leverage a type that implements tinybytes::UnderlyingBytes trait to avoid copying                self.send_deser_ser(tinybytes::Bytes::copy_from_slice(data))
            } 
```

data-pipeline/src/trace_exporter.rs
```diff
@@ -1187,70 +1192,71 @@ mod tests {
...
-        }
-
-        exporter.send(data.as_slice(), 1).unwrap();
-        exporter
-            .shutdown(Some(Duration::from_millis(500)))
-            .unwrap_err(); // The shutdown should timeout
-
-        mock_traces.assert();
-    }
+    // TODO: bring this test back with client side stats
```
ekump, line 1198: Not the end of the world since we need to revisit CSS ASAP, but generally we should have a jira ticket assigned for all TODOs. It doesn't guarantee the TODO gets addressed, but it increases the chances.

ekump (approved):
Rough, to use your words but LGTM since we're addressing all the issues imminently.

ekump:
Please associate a Jira ticket to this PR before merging.

data-pipeline/src/trace_exporter.rs
ajgajg1134: Ahh yeah, I'm just so used to the rust rover cmd+/ to just line comment whole sections :P

data-pipeline/src/trace_exporter.rs
ajgajg1134: Yeah good call out, just added one

data-pipeline/src/trace_exporter.rs
```diff
@@ -254,7 +254,12 @@ impl TraceExporter {
         self.check_agent_info();
         match self.input_format {
             TraceExporterInputFormat::Proxy => self.send_proxy(data, trace_count),
-            TraceExporterInputFormat::V04 => self.send_deser_ser(data),
+            TraceExporterInputFormat::V04 => {
+                self.send_deser_ser(tinybytes::Bytes::copy_from_slice(data))
+                // TODO: APMSP-1582 - Refactor data-pipeline-ffi so we can leverage a type that
+                // implements tinybytes::UnderlyingBytes trait to avoid copying
+                // self.send_deser_ser(tinybytes::Bytes::copy_from_slice(data))
```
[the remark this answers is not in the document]
ajgajg1134: ahh whoops, yeah just pushed a fix

hoolioh (approved):

PR #754, opened 2024-11-22
[crashtracking] improve poll waiting logic

sanchda:
# What does this PR do?

The original implementation accidentally had a mutable _array_ with immutable objects, causing the interface to always throw errors.  Since this part of the code is in the critical path for handling zombie processes, this condition had an adverse side-effect on customer infrastructure.

This code also used a `BorrowedFd`, which is supposed to track an `OwnedFd`.  This was problematic in some conditions, since the underlying implementation would use `prctl()` to check file descriptor liveness and panic in some edge-cases.  The code has been ported to libc, using exclusively `RawFd`, in order to prevent this condition.

Finally, this patch grants some additional time to the act of reaping a PID.  When a receiver process exceeds its timeout budget, it's sent a `SIGKILL`.  However, the old behavior was to `SIGKILL`, the immediately `waitpid( pid, ..., WNOHANG)`.  On a saturated system (i.e., precisely the kind of system where a timeout might be necessary!), it may take some time for the receiver PID to respond to the `SIGKILL`.

In general, there's no way to provided a bounded guarantee for the duration of this reap operation, so an arbitrary number of scheduler slices is chosen as the maximum reaping wait duration.

# Motivation

Fix zombies

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -187,6 +187,7 @@ jobs:
             cmake -S .. -DDatadog_ROOT=$LIBDD_OUTPUT_FOLDER -DBUILD_SYMBOLIZER=true
             cmake --build .
             ./symbolizer
+            ./crashinfo
           else
             cmake -S .. -DDatadog_ROOT=$LIBDD_OUTPUT_FOLDER
             cmake --build .
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1313,6 +1313,7 @@ dependencies = [
  "rmp-serde",
  "serde",
  "serde_json",
+ "tinybytes",
  "tokio",
  "tokio-util",
  "uuid",
--- a/crashtracker-ffi/src/crash_info/datatypes.rs
+++ b/crashtracker-ffi/src/crash_info/datatypes.rs
@@ -247,6 +247,20 @@ impl<'a> TryFrom<SigInfo<'a>> for datadog_crashtracker::SigInfo {
 }
 
 #[repr(C)]
+pub struct ProcInfo {
+    pub pid: u32,
+}
+
+impl TryFrom<ProcInfo> for datadog_crashtracker::ProcessInfo {
+    type Error = anyhow::Error;
+
+    fn try_from(value: ProcInfo) -> anyhow::Result<Self> {
+        let pid = value.pid;
+        Ok(Self { pid })
+    }
+}
+
+#[repr(C)]
 pub struct Metadata<'a> {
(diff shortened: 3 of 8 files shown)


crashtracker/src/collector/crash_handler.rs
```diff
@@ -228,20 +229,29 @@ fn run_receiver_child(uds_parent: RawFd, uds_child: RawFd, stderr: RawFd, stdout
...
-        0 => Ok(false),
-        _ => match poll_fd
-            .revents()
-            .ok_or_else(|| anyhow::anyhow!("No revents found"))?
-        {
-            revents if revents.contains(PollFlags::POLLHUP) => Ok(true),
-            _ => Err(anyhow::anyhow!("poll returned unexpected result")),
-        },
+    let mut poll_fds = [pollfd {
+        fd: target_fd,
```
pawelchcki, line 233: +1 - BorrowedFd prefferably should be use in conjuction with OwnedFd. borrow_raw - without any guarantees of FD lifetime is problematic. 

Probably the safest option would be to dup the fd - and own it within the context of this function. 

Otherwise the code looks like correct but "C'ish" rust :)

crashtracker/src/collector/crash_handler.rs
```diff
@@ -484,7 +494,10 @@ fn receiver_finish(receiver: Receiver, start_time: Instant, timeout_ms: u32) {
         }
 
         let receiver_pid_as_pid = Pid::from_raw(receiver.receiver_pid);
-        let reaping_allowed_ms = timeout_ms.saturating_sub(start_time.elapsed().as_millis() as u32);
+        let reaping_allowed_ms = std::cmp::min(
+            timeout_ms.saturating_sub(start_time.elapsed().as_millis() as u32),
+            DD_CRASHTRACK_MINIMUM_REAP_TIME_MS,
+        );
 
         let _ = reap_child_non_blocking(receiver_pid_as_pid, reaping_allowed_ms);
```
pawelchcki, line 502: >https://github.com/DataDog/libdatadog/blob/25079f5ce6205a50e413dd952db4a6dd6baca008/crashtracker/src/collector/crash_handler.rs#L159

In sidecar - we send kill and term. When the timeout ends.

And it looks that - we're not doing that here either way - so a non 0 timeout will only reduce the incidence of zombies. Not prevent them.

danielsn: Don't we do that just above? https://github.com/DataDog/libdatadog/blob/main/crashtracker/src/collector/crash_handler.rs#L483

PR #755, opened 2024-11-22
ekump/poc byteslice to tinybytes

ekump:
**DO NOT MERGE. This PR is just for illustration purposes.**

# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/data-pipeline-ffi/src/trace_exporter.rs
+++ b/data-pipeline-ffi/src/trace_exporter.rs
@@ -111,8 +111,11 @@ pub unsafe extern "C" fn ddog_trace_exporter_send(
     trace_count: usize,
 ) -> MaybeError {
     // TODO - handle errors - https://datadoghq.atlassian.net/browse/APMSP-1095
+    let static_trace: ByteSlice<'static> = std::mem::transmute(trace);
+    let tinybytes_trace = tinybytes::Bytes::from(static_trace);
+
     handle
-        .send(trace.as_bytes(), trace_count)
+        .send(tinybytes_trace, trace_count)
         .unwrap_or(String::from(""));
     MaybeError::None
 }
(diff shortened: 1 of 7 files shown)


data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -111,8 +111,11 @@ pub unsafe extern "C" fn ddog_trace_exporter_send(
     trace_count: usize,
 ) -> MaybeError {
     // TODO - handle errors - https://datadoghq.atlassian.net/browse/APMSP-1095
+    let static_trace: ByteSlice<'static> = std::mem::transmute(trace);
+    let tinybytes_trace = tinybytes::Bytes::from(static_trace);
+
```
hoolioh, line 116: I don't know what would be more idiomatic using transmute to modify the lifetime or wrapping the pointer in as struct with a lifetime marker as 
in:
```
let data = ByteSlice::from_raw_parts(trace.as_mut_ptr(), trace.len());
let tinybytes_trace = tinybytes::Bytes::from(data);
```

data-pipeline/examples/send-traces-with-stats.rs
```diff
@@ -55,6 +55,6 @@ fn main() {
         traces.push(trace);
     }
     let data = rmp_serde::to_vec_named(&traces).unwrap();
-    exporter.send(&data, 100).unwrap();
+    exporter.send(tinybytes::Bytes::from(data), 100).unwrap();
```
hoolioh, line 58: Myabe I'm under the wrong assumption but wouldn't it be better to hide the tinybytes handling from the user so the signature of the public function would be like this:
```
pub fn send(&self, data: &[u8], trace_count: usize) -> Result<String, String>
``` 
This way we will handle the tinybytes wrapper inside the function and also avoid the dependency in the ffi layer.

PR #764, opened 2024-11-26
Implement UnderlyingBytes for ByteSlice so it can be borrowed as underlying for TinyBytes::Bytes

ekump:
# What does this PR do?

1. Implement necessary traits so that `tinybytes::Bytes` can be created from a static `ByteSlice` without performing a copy of the bytes. 
1. Modify the data-pipeline send function signature to accept anything that implements the `UnderlyingBytes` trait and switch from using `tinybytes::Bytes::copy_from_slice()` to `from` to avoid the aforementioned copying of data and instead borrow it.

# Motivation

Prepping trace exporter for its first integration with a tracer. Having to allocate memory for spans is very expensive. 
 
# Additional Notes

# How to test the change?

Unit tests have been updated

### DIFF
--- a/data-pipeline-ffi/src/trace_exporter.rs
+++ b/data-pipeline-ffi/src/trace_exporter.rs
@@ -111,11 +111,13 @@ pub unsafe extern "C" fn ddog_trace_exporter_send(
     trace_count: usize,
 ) -> MaybeError {
     // TODO - handle errors - https://datadoghq.atlassian.net/browse/APMSP-1095
+
+    // necessary that the trace be static for the life of the handle. Memory should be freed when
+    // the handle is dropped.
     let static_trace: ByteSlice<'static> = std::mem::transmute(trace);
-    let tinybytes_trace = tinybytes::Bytes::from(static_trace);
 
     handle
-        .send(tinybytes_trace, trace_count)
+        .send(static_trace, trace_count)
         .unwrap_or(String::from(""));
     MaybeError::None
 }
--- a/data-pipeline-ffi/src/trace_exporter.rs
+++ b/data-pipeline-ffi/src/trace_exporter.rs
@@ -111,8 +111,11 @@ pub unsafe extern "C" fn ddog_trace_exporter_send(
     trace_count: usize,
 ) -> MaybeError {
     // TODO - handle errors - https://datadoghq.atlassian.net/browse/APMSP-1095
+    let static_trace: ByteSlice<'static> = std::mem::transmute(trace);
+    let tinybytes_trace = tinybytes::Bytes::from(static_trace);
+
     handle
-        .send(trace.as_bytes(), trace_count)
+        .send(tinybytes_trace, trace_count)
         .unwrap_or(String::from(""));
     MaybeError::None
 }
--- a/data-pipeline-ffi/src/trace_exporter.rs
+++ b/data-pipeline-ffi/src/trace_exporter.rs
@@ -102,7 +102,8 @@ pub unsafe extern "C" fn ddog_trace_exporter_free(handle: Box<TraceExporter>) {
 ///
 /// * `handle` - The handle to the TraceExporter instance.
 /// * `trace` - The traces to send to the Datadog Agent in the input format used to create the
-///   TraceExporter.
+///   TraceExporter. The memory for the trace must be valid for the life of the call to this
+///   function.
 /// * `trace_count` - The number of traces to send to the Datadog Agent.
 #[no_mangle]
 pub unsafe extern "C" fn ddog_trace_exporter_send(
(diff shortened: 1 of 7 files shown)


data-pipeline/src/trace_exporter.rs
```diff
@@ -1344,9 +1344,10 @@ mod tests {
             stats_socket.local_addr().unwrap().to_string(),
         );
 
-        let _result = exporter
-            .send(b"some_bad_payload", 1)
-            .expect("failed to send trace");
+        // let bad_payload = tinybytes::Bytes::copy_from_slice(b"some_bad_payload".as_ref());
```
VianneyRuhlmann, line 1347: nit: Should this be removed ?

ddcommon-ffi/src/slice.rs
```diff
@@ -53,6 +54,19 @@ pub type CharSlice<'a> = Slice<'a, c_char>;
 /// Use to represent bytes -- does not need to be valid UTF-8.
 pub type ByteSlice<'a> = Slice<'a, u8>;
 
+// Send, Sync, UnderlyingBytes, and AsRef<[u8]> are all implemented to support using ByteSlice as
```
VianneyRuhlmann, line 57: What's the difference between, using a ByteSlice<'static> as underlying and calling `Bytes::from_static` on the `&'static [u8]`

data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -111,8 +111,13 @@ pub unsafe extern "C" fn ddog_trace_exporter_send(
     trace_count: usize,
 ) -> MaybeError {
     // TODO - handle errors - https://datadoghq.atlassian.net/browse/APMSP-1095
+
+    // necessary that the trace be static for the life of the handle. Memory should be freed when
```
VianneyRuhlmann, line 115: Does this means that the caller has to make sure the `trace` memory is not freed or garbage collected for the life of the handle (quite similar to the life of the program). Should this be added to the doc of the function ?
Is the free handled by libdatadog ?

hoolioh (approved):

ddcommon-ffi/src/slice.rs
ekump: We can't call `Bytes:from_static` in `ddog_trace_exporter_send` because we want `trace_exporter::send` to accept a type that implements `UnderlyingBytes`, not the Bytes themselves.

I guess we could implement `UnderlyingBytes` for `&'static [u8]`. I slightly prefer implementing it, and the associated traits for `ByteSlice` as it's a bit more intentional, but I don't feel strongly about it.

data-pipeline/src/trace_exporter.rs
```diff
@@ -253,15 +254,14 @@ impl TraceExporter {
 
     /// Send msgpack serialized traces to the agent
     #[allow(missing_docs)]
-    pub fn send(&self, data: &[u8], trace_count: usize) -> Result<String, String> {
+    pub fn send<T>(&self, data: T, trace_count: usize) -> Result<String, String>
+    where
+        T: UnderlyingBytes,
```
bantonsson, line 259: I think that this code should just accept `tinybytes::Bytes` and it's up to the caller to figure out how to provide that. It would also mean that we don't have to implement `UnderlyingBytes` for the `ByteSlice`

ekump: > It would also mean that we don't have to implement UnderlyingBytes for the ByteSlice

Because we would just call `tinybytes::Bytes::from_static` in `ddog_trace_exporter_send` like @VianneyRuhlmann [was asking](https://github.com/DataDog/libdatadog/pull/764#discussion_r1860570571)? 

I guess at the end of the day `send` being generic doesn't really save the caller anything. I can make that change.

data-pipeline-ffi/src/trace_exporter.rs
ekump: This is a good callout and we should reword this comment (and update the doc comment). From the Trace Exporter's perspective we want `trace` to be static for the life of this function call, not necessarily for the entire life of the handle. If the caller wishes to re-use memory and swap new traces in to existing memory instead of re-allocating that is up to their implementation.

data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -111,8 +111,16 @@ pub unsafe extern "C" fn ddog_trace_exporter_send(
...
+    // the handle is dropped.
+    let static_trace: ByteSlice<'static> = std::mem::transmute(trace);
```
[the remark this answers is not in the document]
ekump: Discussed IRL. Going to leave as is for now and address properly ASAP.

ddcommon-ffi/src/slice.rs
ekump: Removed this and went with using `Bytes::from_static`

data-pipeline/src/trace_exporter.rs
ekump: done

PR #768, opened 2024-11-27
Refator API for Trace Exporter FFI

hoolioh:
# What does this PR do?

This PR tries to improve the FFI for the trace exporter API by:
* Use a builder approach by encapsulating configuration in a new structure.
* Forward agent responses by introducing a new data type called `Rates` which can be queried to get the sampling rate for a specific service name. The value will be passed to the library user by a new output parameter in `ddog_trace_exporter_send`.
* Improve error handling in the API by returning a new data type called `ddog_TraceExporterError` which contains new error codes and debug information.

### DIFF
--- a/data-pipeline-ffi/src/error.rs
+++ b/data-pipeline-ffi/src/error.rs
@@ -1,0 +1,47 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+/// Represent error codes that `Error` struct can hold
+#[repr(C)]
+#[derive(Debug, PartialEq)]
+pub enum TraceExporterErrorCode {
+    AddressInUse,
+    ConnectionAborted,
+    ConnectionRefused,
+    ConnectionReset,
+    HttpBodyFormat,
+    HttpBodyTooLong,
+    HttpClient,
+    HttpParse,
+    HttpServer,
+    HttpWrongStatus,
+    InvalidArgument,
+    InvalidData,
+    InvalidInput,
+    InvalidUrl,
+    IoError,
+    NetworkUnknown,
+    Serde,
+    TimedOut,
+}
+
+/// Stucture that contains error information that `TraceExporter` API can return.
+#[repr(C)]
+#[derive(Debug, PartialEq)]
+pub struct TraceExporterError {
+    pub code: TraceExporterErrorCode,
+}
+
+impl From<TraceExporterErrorCode> for TraceExporterError {
+    fn from(value: TraceExporterErrorCode) -> Self {
+        TraceExporterError { code: value }
+    }
+}
+
+/// Free
+#[no_mangle]
+pub unsafe extern "C" fn ddog_trace_exporter_error_free(
+    _error: ddcommon_ffi::Option<&mut TraceExporterError>,
+) {
+    // TODO: Placeholder function to remove dynamically allocated properties in Error.
+}
--- a/data-pipeline-ffi/src/error.rs
+++ b/data-pipeline-ffi/src/error.rs
@@ -1,10 +1,17 @@
 // Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
+use data_pipeline::trace_exporter::error::{
+    BuilderErrorKind, NetworkErrorKind, TraceExporterError,
+};
+use std::ffi::{c_char, CString};
+use std::fmt::Display;
+use std::io::ErrorKind as IoErrorKind;
+
 /// Represent error codes that `Error` struct can hold
 #[repr(C)]
-#[derive(Debug, PartialEq)]
-pub enum TraceExporterErrorCode {
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum ExporterErrorCode {
     AddressInUse,
     ConnectionAborted,
     ConnectionRefused,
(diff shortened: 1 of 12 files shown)


examples/ffi/trace_exporter.c
```diff
@@ -57,7 +50,9 @@ int main(int argc, char** argv)
     }
 
     ddog_ByteSlice buffer = { .ptr = NULL, .len=0 };
-    TRY(ddog_trace_exporter_send(trace_exporter, buffer, 0));
+    ddog_AgentResponse response;
+
+    TRY(ddog_trace_exporter_send(trace_exporter, buffer, 0, &response));
```
ganeshnj, line 55: /suggestion

since we are not sending the error information in return, we should update the example to check the error and freeing the error if returned.

data-pipeline-ffi/src/error.rs
```diff
@@ -0,0 +1,161 @@
...
+    NetworkUnknown,
+    Serde,
+    TimedOut,
+}
+
+impl Display for ExporterErrorCode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::AddressInUse => write!(f, "Address already in use"),
+            Self::ConnectionAborted => write!(f, "Connection aborted"),
```
ganeshnj, line 39: /question

do you think it would make sense to return the actual message, when the root cause of the error is not us? This would help in debugging and reduce the number of mapping we own.

Those error could be even more informative.

ganeshnj (approved):
Very nice work :tada

data-pipeline-ffi/src/error.rs
hoolioh: We're actually sending the proper message when the error is originated internally. In the `from` trait [implementation](https://github.com/DataDog/libdatadog/pull/768/files#diff-46620090d8d466154fc006ef9bfff44f8144f70c353356664e866b817790f179R77-R117) you'll see that we're translating to string the `TraceExporterError` contents which means, in the case of Io and Hyper errors, we'll use their own [display representation](https://github.com/DataDog/libdatadog/pull/768/files#diff-1d490aa446c998ae0124ca12a84ec4a2094fc80f9714bb705351599399d63cd4R58-R62).
This trait implementation is only used when the error is caused in the ffi layer as [in parsing arguments](https://github.com/DataDog/libdatadog/pull/768/files#diff-6b9976c838fe0d73c7e4b54e1bb1c3dbee58eb65441f46ff61b6ca8364d468c1R267-R268)

ganeshnj: Perfect

data-pipeline-ffi/src/error.rs
```diff
@@ -0,0 +1,161 @@
...
+                NetworkErrorKind::ConnectionClosed => ExporterErrorCode::ConnectionReset,
+                NetworkErrorKind::MessageTooLarge => ExporterErrorCode::HttpBodyTooLong,
+                NetworkErrorKind::Parse => ExporterErrorCode::HttpParse,
+                NetworkErrorKind::TimedOut => ExporterErrorCode::TimedOut,
+                NetworkErrorKind::Unknown => ExporterErrorCode::NetworkUnknown,
+                NetworkErrorKind::WrongStatus => ExporterErrorCode::HttpWrongStatus,
+            },
+            TraceExporterError::Request(e) => {
+                let status: u16 = e.status().into();
+                if (400..500).contains(&status) {
```
ekump, line 107: I know it _shouldn't_ happen, but Is it possible that we wind up with a `Request` error with a non-error status code? This if statement will return `exporterErrorCode::HttpServer` if the status code is < 400. If so, should we have a `ExporterErrorCode::Unknown` variant?

data-pipeline-ffi/src/error.rs
```diff
@@ -0,0 +1,161 @@
...
+                    ExporterErrorCode::HttpServer
+                }
+            }
+            TraceExporterError::Serde(_) => ExporterErrorCode::Serde,
+        };
+        ExporterError::new(code, &value.to_string())
+    }
+}
+
+/// Free
```
ekump, line 133: Does this warrant a more detailed docstring comment? Or should it be self-explanatory?

data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -1,88 +1,230 @@
...
+}
+
+/// Sets tracer's version to be included in the headers request.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_trace_exporter_config_set_tracer_version(
+    config: ddcommon_ffi::Option<&mut TraceExporterConfig>,
+    version: CharSlice,
+) -> ddcommon_ffi::Option<Error> {
+    if let ddcommon_ffi::Option::Some(handle) = config {
+        handle.tracer_version = Some(version.to_utf8_lossy().to_string());
```
ekump, line 76: Should we be using `to_utf8_lossy()` for stuff like this? `to_utf8_lossy` replaces invalid utf8 characters with a � so the string is technically valid utf8, but the value is basically corrupted for its intended purposes. Would it be better to use `try_to_utf8()` and return an error here? 

This feels like one of those edgecases that if it does become a bug, will be hard to track down later. Or maybe I'm just being paranoid. (If you agree, we can address in a follow-up PR).

ddcommon-ffi/src/option.rs
```diff
@@ -19,6 +19,13 @@ impl<T> Option<T> {
             Option::None => None,
         }
     }
+
```
ekump, line 22: any value in starting to add some tests to this file? I noticed some of the other files in this crate have them now.

data-pipeline/src/trace_exporter/error.rs
```diff
@@ -0,0 +1,163 @@
...
+            msg: msg.to_owned(),
+        }
+    }
+
+    pub fn status(&self) -> StatusCode {
+        self.code
+    }
+}
+
+/// TraceExporterError holds different types of errors occurred when handling traces.
```
ekump, line 93:
```suggestion
/// TraceExporterError holds different types of errors that occur when handling traces.
```

ekump (approved):
some questions / suggestions, but nothing blocking. LGTM

data-pipeline-ffi/src/error.rs
hoolioh: Yep, by design it shouldn't but if there's some bug in the network library I think there's no harm in being prepared for that. I'll add another case in order to handle a generic error.

ddcommon-ffi/src/option.rs
hoolioh: If you agree maybe I could do it in a separate PR. I'm talking with Daniel about the possibility of porting/moving error handling to the common module so I could add more tests then.

ekump: No problem. The more I think about it the less concerned I am about a unit test for `as_mut`. If it didn't work correctly then the code wouldn't compile.

PR #769, opened 2024-11-27
Reenable Client-side stats in TraceExporter

VianneyRuhlmann:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


trace-utils/src/span_v04/trace_utils.rs
```diff
@@ -0,0 +1,81 @@
...
+
+fn set_top_level_span(span: &mut Span, is_top_level: bool) {
+    if !is_top_level {
+        if span.metrics.contains_key(TOP_LEVEL_KEY) {
+            span.metrics.remove(TOP_LEVEL_KEY);
+        }
+        return;
+    }
+    span.metrics.insert(TOP_LEVEL_KEY.into(), 1.0);
+}
```
ganeshnj, line 26: /improvement

you can simplify

```suggestion
   if is_top_level {
        span.metrics.insert(TOP_LEVEL_KEY.into(), 1.0);
    } else {
        span.metrics.remove(TOP_LEVEL_KEY);
    }
}
```

remove works fine if the key doesn't exist https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.remove, just the return value will be None which don't care here.

data-pipeline/src/trace_exporter.rs
```diff
@@ -1026,119 +1031,119 @@ mod tests {
...
-    // }
+
+    #[test]
+    fn test_drop_chunks() {
+        let chunk_with_priority = vec![
+            Span {
+                span_id: 1,
+                metrics: HashMap::from([
+                    (SAMPLING_PRIORITY_KEY.into(), 1.0),
+                    ("_dd.top_level".into(), 1.0),
```
ganeshnj, line 1042:
```suggestion
                    (TRACER_TOP_LEVEL_KEY.into(), 1.0),
```

ganeshnj (changes requested):
Looks good, some questions and suggestions but nothing critical.

ganeshnj (approved):

tinybytes/src/bytes_string.rs
```diff
@@ -106,6 +106,13 @@ impl BytesString {
         // SAFETY: We assume all BytesStrings are valid UTF-8.
         unsafe { std::str::from_utf8_unchecked(&self.bytes) }
     }
+
```
ekump, line 109: As trivial as it may seem, we should really have a unit test for any public methods.

trace-utils/src/span_v04/trace_utils.rs
ekump, line 1: Non-blocking: I understand we want to maintain some level of parity with the existing functions for `pb` spans and we don't want to spend too much time re-designing this functionality. But, could we not call this trace_utils? I would prefer this file not turn into another catch-all file.

trace-utils/src/span_v04/trace_utils.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// Return true if the span has a top level key set
+pub fn has_top_level(span: &Span) -> bool {
+    span.metrics
+        .get(TRACER_TOP_LEVEL_KEY)
+        .is_some_and(|v| *v == 1.0)
+        || span.metrics.get(TOP_LEVEL_KEY).is_some_and(|v| *v == 1.0)
+}
+
+// Returns true if a span should be measured (i.e., it should get trace metrics calculated).
```
ekump, line 67:
```suggestion
/// Returns true if a span should be measured (i.e., it should get trace metrics calculated).
```
unless that was intentional, then ignore.

trace-utils/src/span_v04/trace_utils.rs
```diff
@@ -0,0 +1,81 @@
...
+    span.metrics.insert(TOP_LEVEL_KEY.into(), 1.0);
+}
+
+/// Updates all the spans top-level attribute.
+/// A span is considered top-level if:
+///   - it's a root span
+///   - OR its parent is unknown (other part of the code, distributed trace)
+///   - OR its parent belongs to another service (in that case it's a "local root" being the highest
+///     ancestor of other spans belonging to this service and attached to it).
+pub fn compute_top_level_span(trace: &mut [Span]) {
```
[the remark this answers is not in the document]
ekump: +1, all public functions should have their own unit tests. If for whatever reason that is not practical then we should have a comment indicating that.

ekump (approved):

PR #770, opened 2024-11-27
Add _dd.gcrfx.resource_name to serverless compatibility layer  for Google Cloud Functions

nina9753:
# What does this PR do?

Adds `gcrfx.resource_name` span tag to cloud run services needed for trace mapping by creating a unified tag of the necessary fields i.e `projects/{}/locations/{}/functions/{}` to the _top level span only_
Updates tests to look for the new `gcrfx.resource_name` span tag

# Motivation

https://datadoghq.atlassian.net/browse/SVLS-6008

# Additional Notes
relates to this [PR #31434](https://github.com/DataDog/datadog-agent/pull/31434)
billing is now done in [this PR](https://github.com/DataDog/dd-go/pull/156600)


trace-utils/src/trace_utils.rs
```diff
@@ -526,6 +525,30 @@ pub fn enrich_span_with_mini_agent_metadata(
     }
 }
 
+pub fn enrich_span_with_google_cloud_function_metadata(
```
duncanpharvey, line 528: If we're adding a new method to add metadata specifically for google cloud functions can you move the code for setting the `project_id` and `location`?

https://github.com/DataDog/libdatadog/blob/373b7783730a8191bfb6a1f0135791782cdcc22d/trace-utils/src/trace_utils.rs#L513-L519

That should also allow for using the source data directly to construct the resource name instead of setting the span metadata in one function then fetching it again in another.

trace-utils/src/trace_utils.rs
```diff
@@ -526,6 +525,30 @@ pub fn enrich_span_with_mini_agent_metadata(
...
+    span: &mut pb::Span,
+    env_type: &EnvironmentType,
+) {
+    if let EnvironmentType::CloudFunction = env_type {
+        let function = span.meta.get("functionname");
```
duncanpharvey, line 533: What do you think about passing the functionname in `self.config.app_name.clone()` as a parameter to the `enrich_span_with_google_cloud_function_metadata` method instead of fetching it from the span?

trace-utils/src/trace_utils.rs
```diff
@@ -526,6 +525,30 @@ pub fn enrich_span_with_mini_agent_metadata(
...
+        if function.is_some() && region.is_some() && project.is_some() {
+            let resource_name = format!(
+                "projects/{}/locations/{}/functions/{}",
+                project.unwrap(),
+                region.unwrap(),
+                function.unwrap()
+            );
+
+            span.meta.insert(
+                "_dd.gcrfx.resource_name".to_string(),
```
duncanpharvey, line 545: Is there a reason this resource span tag needs to be prefixed with `_dd`? For Azure App Services and Functions the resource name is visible to end users (ie. not hidden by using `_dd`).

duncanpharvey (changes requested):
Nice work! Just a few minor comments

nina9753:
Closing PR since the code for billing will be consolidated internally in [this PR](https://github.com/DataDog/dd-go/pull/156600) for the serverless compatibility layer

trace-utils/src/trace_utils.rs
nina9753: Thanks for the comment. I added it

trace-utils/src/trace_utils.rs
nina9753: Thanks, I added this

trace-utils/src/trace_utils.rs
nina9753: nevermind! After talking about it again in person, I updated the tag to exclude the `_dd` prefix

trace-utils/src/trace_utils.rs
```diff
@@ -526,6 +517,34 @@ pub fn enrich_span_with_mini_agent_metadata(
...
+    span: &mut pb::Span,
+    mini_agent_metadata: &MiniAgentMetadata,
+    function: Option<String>,
+) {
+    let Some(region) = &mini_agent_metadata.gcp_region else {
+        return;
+    };
+    let Some(project) = &mini_agent_metadata.gcp_project_id else {
+        return;
+    };
```
duncanpharvey, line 530: Since we're short circuiting here on either the `gcp_region` or the `gcp_project_id` being empty there's a possibility that one of the span tags is not set even when the values is available. I'll approve since I think that occurrence should be rare but just wanted to point that out in case it's not what was intended!

duncanpharvey (approved):
LGTM!

trace-utils/src/trace_utils.rs
nina9753: `let Some().... else {return;};` should check/handle the error if the value is empty and should return before adding the resource_name span tag.

duncanpharvey: Yes but my point is that there could be a scenario where `region` has a value but `project` is empty. With this implementation the `region`, `project`, and `resource_name` span tags will all be empty, even though there is a value available for `region`. Before this change the `region` span tag would be set even though the value of `project` is empty.

So it is a change in behavior but it's not enough to warrant a refactor as long as you're ok with this implication.

nina9753: good point this has been updated to todo  instead of return so that we will still add either tag if the value is available

PR #772, opened 2024-11-30
fix(alloc): potentially dangling temporary

morrisonlevi:
# What does this PR do?

 1. Fixes a warning `dangling_pointers_from_temporaries`.
 2. Skips some tests on miri for MacOS.
 3. Fixes a test that doesn't fail for the right reason when running locally (because it can't figure out the cloud env it's running in, because it's not in a cloud env).

# Motivation

I was trying out the Rust 2024 edition which required a new nightly. There's now a warning for `dangling_pointers_from_temporaries`. This could be a false positive but I am not certain. It
seems prudent to fix it and figure out whether it's truly dangling later, if we have time.

The other fixes are just annoyances that happen when developing locally on a Mac.

# Additional Notes

Here's the warning:

```
warning: a dangling pointer will be produced because the temporary `UnsafeCell<ChainNodePtr<A>>` will be dropped
   --> alloc/src/chain.rs:293:30
    |
291 | /                         core::ptr::addr_of!((*non_null.as_ptr()).prev)
292 | |                             .read()
    | |___________________________________- this `UnsafeCell<ChainNodePtr<A>>` is deallocated at the end of the statement, bind it to a variable to extend its lifetime
293 |                               .get()
    |                                ^^^ this pointer will immediately be invalid
    |
    = note: pointers do not have a lifetime; when calling `get` the `UnsafeCell<ChainNodePtr<A>>` will be deallocated at the end of the statement because nothing is referencing it as far as the type system is concerned
    = help: for more information, see <https://doc.rust-lang.org/reference/destructors.html>
    = note: `#[warn(dangling_pointers_from_temporaries)]` on by default
```

# How to test the change?

No changes should be necessary.

### DIFF
--- a/alloc/src/chain.rs
+++ b/alloc/src/chain.rs
@@ -288,10 +288,9 @@ impl<A: Allocator + Clone> Drop for ChainAllocator<A> {
                     // moved to the stack before the chunk is dropped, so it's
                     // alive and valid after the chunk is dropped below.
                     chain_node_ptr = unsafe {
-                        core::ptr::addr_of!((*non_null.as_ptr()).prev)
-                            .read()
-                            .get()
-                            .read()
+                        // Save to variable to avoid a dangling temporary.
+                        let unsafe_cell = core::ptr::addr_of!((*non_null.as_ptr()).prev).read();
+                        unsafe_cell.get().read()
                     };
 
                     // SAFETY: the chunk hasn't been dropped yet, and the
--- a/data-pipeline/src/trace_exporter.rs
+++ b/data-pipeline/src/trace_exporter.rs
@@ -918,6 +918,7 @@ mod tests {
     use tinybytes::BytesString;
     use tokio::time::sleep;
 
+    #[cfg_attr(all(miri, target_os = "macos"), ignore)]
     #[test]
     fn new() {
         let builder = TraceExporterBuilder::default();
@@ -951,6 +952,7 @@ mod tests {
         assert!(!exporter.metadata.client_computed_stats);
     }
 
+    #[cfg_attr(all(miri, target_os = "macos"), ignore)]
     #[test]
     fn new_defaults() {
         let builder = TraceExporterBuilder::default();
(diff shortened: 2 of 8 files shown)


alloc/src/chain.rs
```diff
@@ -288,10 +288,9 @@ impl<A: Allocator + Clone> Drop for ChainAllocator<A> {
...
                     // alive and valid after the chunk is dropped below.
                     chain_node_ptr = unsafe {
-                        core::ptr::addr_of!((*non_null.as_ptr()).prev)
-                            .read()
-                            .get()
-                            .read()
+                        // Save to variable to avoid a dangling temporary.
+                        let unsafe_cell = core::ptr::addr_of!((*non_null.as_ptr()).prev).read();
+                        unsafe_cell.get().read()
                     };
```
ivoanjo, line 294: Wow, very interesting catch. With my knowledge from other programing languages and things that compilers can do :tm: I'd fully expect the old version of the code to be equivalent to the new one, but I guess Rust has other ideas ;) 

On the other hand, as long as the compiler points out the sharp edge, I think I'm fine with Rust being a bit quirky on this :D

ivoanjo (approved):
:+1: Thanks for the nice fixes :)

alloc/src/chain.rs
morrisonlevi: This could be a false positive. It's probably worth time to think about it and report upstream if necessary, except that this is already a tangent from my main work. But we're firmly in unsafe code here, and unsafe Rust can be subtle, so it's certainly possible that this is an issue. I believe miri is running on this code--hence the skips of failures for miri as I double-checked--and it didn't trigger anything. So... not sure who is right.

ivoanjo: Reading the error message and https://doc.rust-lang.org/reference/destructors.html#drop-scopes I'm leaning towards this being a real issue.

It looks like destruction order and scopes are very precisely defined, rather than the "if the compiler sees you're not using it, it may just clean it up much earlier" than other languages I'm used to.

So by introducing the local variable, it looks like the compiler will be forced to keep `unsafe_cell` alive until the end of the scope it's defined on, even though at first sight it looked like it wouldn't need to.

(I guess this means that if the Rust compiler wants to optimize further it'll need to prove that the earlier drop can happen in a safe way, which it probably won't even try in the presence of `unsafe`)

PR #774, opened 2024-12-02
[APMSP-1331] Reduce allocations in client-side stats

VianneyRuhlmann:
# What does this PR do?

Improve SpanConcentrator performance by enabling lookup with a BorrowedAggregationKey. This means that the span fields included in the AggregationKey are only cloned if the key is not already present in the bucket.

# Motivation

Performance of the send method of the TraceExporter is critical for  adoption of data-pipeline

# Additional Notes

Using a different key type for lookup than the one stored (here different lifetime), requires the stored type to be borrowable as the lookup type. The idea used here comes from this [blog post](https://idubrov.name/rust/2018/06/01/tricking-the-hashmap.html) which uses a trait to define a common type. 

# How to test the change?

### DIFF
--- a/ddtelemetry/src/worker/mod.rs
+++ b/ddtelemetry/src/worker/mod.rs
@@ -241,10 +241,10 @@ impl TelemetryWorker {
             Lifecycle(Start) => {
                 if !self.data.started {
                     self.deadlines
-                        .schedule_event(LifecycleAction::FlushData)
+                        .schedule_event(LifecycleAction::FlushMetricAggr)
                         .unwrap();
                     self.deadlines
-                        .schedule_event(LifecycleAction::FlushMetricAggr)
+                        .schedule_event(LifecycleAction::FlushData)
                         .unwrap();
                     self.data.started = true;
                 }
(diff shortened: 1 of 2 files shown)


data-pipeline/src/span_concentrator/aggregation.rs
```diff
@@ -15,65 +16,166 @@ const TAG_ORIGIN: &str = "_dd.origin";
...
 /// This struct represent the key used to group spans together to compute stats.
 #[derive(Debug, Hash, PartialEq, Eq, Clone, Default)]
-pub(super) struct AggregationKey {
-    resource_name: String,
-    service_name: String,
-    operation_name: String,
-    span_type: String,
-    span_kind: String,
+pub(super) struct AggregationKey<'a> {
+    resource_name: Cow<'a, str>,
```
bantonsson, line 20: 👍🏼

data-pipeline/src/span_concentrator/aggregation.rs
```diff
@@ -15,65 +16,166 @@ const TAG_ORIGIN: &str = "_dd.origin";
...
         } else {
             vec![]
         };
         Self {
-            resource_name: span.resource.copy_to_string(),
-            service_name: span.service.copy_to_string(),
-            operation_name: span.name.copy_to_string(),
-            span_type: span.r#type.copy_to_string(),
-            span_kind,
+            resource_name: span.resource.as_str().into(),
```
bantonsson, line 123: Nice

bantonsson (approved):

hoolioh (approved):

ekump (approved):

PR #777, opened 2024-12-02
to not log error on unsupported metric type

alexgallotta:
# What?

Stops logging an error when a metric is of the unsupported type.

Dogstatsd metric types supported by the agent are 6 https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/server/parse_metrics.go#L27-L32

The lambda extension right now supports only g|c|d

- if the parser detects the g|s|ms it will debug log the issue
- if any other invalid string is passed, it will log an error


duncanista (approved):

ekump (approved):

dogstatsd/src/metric.rs
```diff
@@ -14,8 +14,8 @@ use ustr::Ustr;
 pub const EMPTY_TAGS: SortedTags = SortedTags { values: Vec::new() };
 
 lazy_static! {
-    static ref METRIC_REGEX: regex::Regex = Regex::new(
-        r"^(?P<name>[^:]+):(?P<values>[^|]+)\|(?P<type>[cgd])(?:\|@(?P<sample_rate>[\d.]+))?(?:\|#(?P<tags>[^|]+))?(?:\|c:(?P<container_id>[^|]+))?$",
+    static ref METRIC_REGEX: Regex = Regex::new(
+        r"^(?P<name>[^:]+):(?P<values>[^|]+)\|(?P<type>[a-zA-Z]+)(?:\|@(?P<sample_rate>[\d.]+))?(?:\|#(?P<tags>[^|]+))?(?:\|c:(?P<container_id>[^|]+))?$",
```
bantonsson, line 18: I think this should still distinguish between valid metric types `c,g,s,d,h,ms` and garbage.

alexgallotta: Thanks for keeping me honest! I took some time and changed the behavior and improved the error and logging. I will update the description!

bantonsson (approved):

PR #778, opened 2024-12-04
New API for Trace Exporter FFI

hoolioh:
# What does this PR do?

This PR tries to improve the FFI for the trace exporter API by:
* Use a builder approach by encapsulating configuration in a new structure.
* Forward agent responses by introducing a new data type called `Rates` which can be queried to get the sampling rate for a specific service name. The value will be passed to the library user by a new output parameter in `ddog_trace_exporter_send`.
* Improve error handling in the API by returning a new data type called `ddog_TraceExporterError` which contains new error codes and debug information.

### DIFF
--- a/data-pipeline-ffi/src/error.rs
+++ b/data-pipeline-ffi/src/error.rs
@@ -1,0 +1,47 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+/// Represent error codes that `Error` struct can hold
+#[repr(C)]
+#[derive(Debug, PartialEq)]
+pub enum TraceExporterErrorCode {
+    AddressInUse,
+    ConnectionAborted,
+    ConnectionRefused,
+    ConnectionReset,
+    HttpBodyFormat,
+    HttpBodyTooLong,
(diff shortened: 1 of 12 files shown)


data-pipeline-ffi/src/error.rs
```diff
@@ -0,0 +1,167 @@
...
+            Self::Serde => write!(f, "Serialization/Deserialization error"),
+            Self::TimedOut => write!(f, "Operation timed out"),
+        }
+    }
+}
+
+/// Stucture that contains error information that `TraceExporter` API can return.
+#[repr(C)]
+#[derive(Debug, PartialEq)]
+pub struct ExporterError {
```
danielsn, line 63: There are already error types in ddcommon.  Maybe either use those, or move this there?

hoolioh: Hey @danielsn made a mistake with this PR, I accidentally duplicated it. If you have further comments please do it in this [one](https://github.com/DataDog/libdatadog/pull/768). Sorry for the inconvenience.
Regarding your question, we wanted to move forward quickly so we decided to handle errors internally and later, in a separate PR, try to port this to the common module improving the current `MaybeError` data type.

PR #784, opened 2024-12-06
Avoid using lock unnecessarily

cataphract:
For to push these changes before merging https://github.com/DataDog/libdatadog/pull/515

### DIFF
--- a/ddtelemetry/src/worker/mod.rs
+++ b/ddtelemetry/src/worker/mod.rs
@@ -435,7 +435,7 @@ impl TelemetryWorker {
                     payloads.push(data::Payload::MessageBatch(obsevability_events));
                 }
 
-                let self_arc = Arc::new(tokio::sync::RwLock::new(self));
+                let self_arc = Arc::new(tokio::sync::RwLock::new(&mut *self));
                 let futures = payloads.into_iter().map(|payload| {
                     let self_arc = self_arc.clone();
                     async move {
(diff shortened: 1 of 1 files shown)


ddtelemetry/src/worker/mod.rs
```diff
@@ -454,10 +454,9 @@ impl TelemetryWorker {
                 });
                 future::join_all(futures).await;
 
-                let mut self_lock = self_arc.write().await;
-                self_lock.data.started = false;
-                if !self_lock.config.restartable {
```
bantonsson, line 459: I can be mistaken, but I thought that the use of the the `RwLock` here was about ordering and visibility guarantees. But maybe that's being taken care of by the `await`.

cataphract: The `RwLock` is about the fact that both futures may try to write into self, and they may run in parallel (if there was a simple loop instead of `join_all` this wouldn't be the case, because futures run only when polled). If not, this would have sufficed (with the creation of a shared reference `self2`, which is `Copy`, unlike the mut ref `self`)

```
let self2: &TelemetryWorker = &*self;
let futures = payloads.into_iter().map(|payload| {
	async move {
		match self2.send_payload(&payload).await {
			Ok(()) => (),
			Err(err) => self2.log_err(&err),
		}
	}
});
```

I'm pretty sure there are no visibility problems; this would even work with plain threads because of the synchronization guarantees of a join.

bantonsson: Yes, I know what the `RwLock` is used for inside the futures (and how `rust` futures work). I just wasn't sure what guarantees the `join` will give you, but since any writes happens before joining (which needs to make things visible) it sounds fine.

bantonsson (approved):

PR #787, opened 2024-12-06
doc: restrictions on bumping MSRV

morrisonlevi:
# What does this PR do?

Documents the restrictions that the PHP project imposes on bumping the MSRV.

# Motivation

I've been wanting to do this for a while, but wasn't unsure if Remi Collet would be comfortable with me putting this in the docs. I asked him on an issue the last time we bumped MSRV, and he said that's okay.

Alpine Linux released a new version, so that's specifically what prompted to open the PR _right now_.

# How to test the change?

No changes are made, this is just documentation.


ivoanjo (approved):
:+1: LGTM Thanks a lot for documenting this

PR #796, opened 2024-12-10
Allow send empty arrays through the trace exporter

hoolioh:
# What does this PR do?

Allow sending empty arrays due to some tracers send this kind of payloads to check agent's availability.

# Motivation

Ensure compatibility with that feature.

### DIFF
--- a/data-pipeline/src/trace_exporter/mod.rs
+++ b/data-pipeline/src/trace_exporter/mod.rs
@@ -590,13 +590,6 @@ impl TraceExporter {
             }
         };
 
-        if traces.is_empty() {
-            error!("No traces deserialized from the request body.");
-            return Err(TraceExporterError::Io(std::io::Error::from(
-                std::io::ErrorKind::InvalidInput,
-            )));
-        }
-
         let num_traces = traces.len();
 
         self.emit_metric(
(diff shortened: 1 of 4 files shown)


data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -1531,6 +1524,41 @@ mod tests {
         assert_eq!(result, AgentResponse::from(0.8));
     }
 
+    #[test]
+    #[cfg_attr(miri, ignore)]
```
ekump, line 1528: Are you skipping miri here because you were seeing an access violation error with the tinybytes payload? If so, you can create the bytes from static like [here](https://github.com/DataDog/libdatadog/blob/962589d7316094430b90327c90dd60e4095642ae/trace-utils/src/msgpack_decoder/v04/decoder/mod.rs#L327).

ekump (approved):

ganeshnj (approved):

data-pipeline/src/trace_exporter/mod.rs
hoolioh: I'm skipping those tests because it appears that `libc::socket` call is not emulated in Miri, at least in the nightly version we're currently using.

PR #73, opened 2022-11-18
Rename FFI types, functions and enums in preparation for 1.0

ivoanjo:
# What does this PR do?

This PR updates the names used in the libdatadog API as discussed in the 1.0 discussion meeting.

In particular, the following concerns were addressed:

* Namespace names -- `ddog_` for common-ffi, `ddog_prof_` for profiling-ffi. This avoids having something like `ddog_Line` which is very specific to profiling be at the top of the namespace.

* Fixed inconsistency in destructors: we had `free`, `delete` and `drop`. We now use `drop` everywhere.

* Updated namespacing for nested things. E.g. the result of `ddog_prof_Exporter_new` is a `ddog_prof_Exporter_NewResult`.

* Used case to separate structs (CamelCase) from methods (snake_case)

One departure from what was discussed in the meeting is how containers/generics are named. I had proposed `ddog_VecTag` and `ddog_SliceU8` but cbindgen really doesn't like it. Rather than fight cbindigen about it, I settled for `ddog_Vec_tag` and `ddog_Slice_U8`. Hopefully that is acceptable.

 # Motivation

As our API grows, having a more consistent naming will hopefully help people navigate it. We still have a bit of a weird mix of what goes on `common.h` vs `profiling.h`, but I'll leave changing that for a later PR.

Furthermore, it looked to me like 0.9-to-1.0 was a good time to address these concerns.

 # Additional Notes

To the best of my knowledge, here's the complete list of all types changed:

| **Kind** | **0.9**                              | **1.0**                                      |
|:--------:|--------------------------------------|----------------------------------------------|
|     Type | ddog_Endpoint_ddog_Agentless_Body    | ddog_Endpoint_ddog_prof_Agentless_Body       |
|     Type | ddog_Slice_c_char                    | ddog_Slice_CChar                             |
|     Type | ddog_Slice_i64                       | ddog_Slice_I64                               |
|     Type | ddog_Slice_u8                        | ddog_Slice_U8                                |
|     Type | ddog_Vec_tag                         | ddog_Vec_Tag                                 |
| Function | ddog_Vec_tag_drop                    | ddog_Vec_Tag_drop                            |
| Function | ddog_Vec_tag_new                     | ddog_Vec_Tag_new                             |
| Function | ddog_Vec_tag_parse                   | ddog_Vec_Tag_parse                           |
|     Type | ddog_ParseTagsResult                 | ddog_Vec_Tag_ParseResult                      |
[...]

PR #797, opened 2024-12-10
[profiling] Use the new FFI macros and types

danielsn:
# What does this PR do?

Uses the macros introduced in https://github.com/DataDog/libdatadog/pull/752 in the profiler-ffi

# Motivation

DRY: Don't Repeat Yourself ;)

# Additional Notes

This has breaking FFI changes (changes some types and type names) but the upgrade should be pretty mechanical clientside. 
# How to test the change?

This is a refactor change


morrisonlevi (commented):
This has breaking FFI changes, which I didn't expect with a "refactor" title, but that's okay. Just wanted to call a bit of attention there, and we should write something about in this PR description.

I think I need to open this up in an editor to see some of it, diff-view isn't great sometimes.

examples/ffi/profiles.c
```diff
@@ -14,15 +14,15 @@ int main(void) {
   const ddog_prof_Slice_ValueType sample_types = {&wall_time, 1};
   const ddog_prof_Period period = {wall_time, 60};
 
-  ddog_prof_Profile_NewResult new_result = ddog_prof_Profile_new(sample_types, &period, NULL);
-  if (new_result.tag != DDOG_PROF_PROFILE_NEW_RESULT_OK) {
+  ddog_prof_Result_HandleProfile new_result = ddog_prof_Profile_new(sample_types, &period, NULL);
+  if (new_result.tag != DDOG_PROF_RESULT_HANDLE_PROFILE_OK_HANDLE_PROFILE) {
```
morrisonlevi, line 18: I don't write C API these days, but this looks worse. Can we use renames and/or aliases to improve this? Examples:

- `ddog_prof_Profile_NewResult` aliased to `ddog_prof_Result_HandleProfile`
- `ddog_prof_Profile` aliased to `ddog_prof_Handle_Profile`

Not sure what to do about the enum. Maybe @ivoanjo has thoughts?

profiling-ffi/src/datatypes.rs
morrisonlevi, line 1: I like these being moved into their own file 👍🏻

examples/ffi/profiles.c
ivoanjo: Hmmm yeah I think Levi has a good point here, the "handle" change does look a bit like libdatadog internal implementation details being leaked across the FFI.

E.g. looking at it from the pure API client side, I'm not sure it's relevant for client to care what a `ddog_prof_Profile` actually is: is it a profile? a handle? something else?

I also know that (at least in the past) some of our types were a bit less what we wanted because of cbindgen limitations, and I think it's still a reasonable trade-off to have the types uglier if the rust code becomes less error-prone/etc. But ideally having both (clear naming for ffi and better rust types) would be great.

> ddog_prof_Profile aliased to ddog_prof_Handle_Profile

TL;DR: :+1: on this one :)

---

On the `Result` change...

We've been using the convention of calling `Module_operation` returning `Module_OperationResult` since [this PR](https://github.com/DataDog/libdatadog/pull/73). (And [we did explore a few more options at the time](https://docs.google.com/spreadsheets/d/1iuuSCJ0NcSOpo2RVvkQ8f8ugz2XlEbbpFZLVOg6WioE/edit?gid=0#gid=0))

One advantage of this new convention is that it may result (no pun intended) in fewer `...Result` types, because they are now named by their payload's type, rather than the operation's type.

I still think I slightly favor more the current convention, but I think both seem quite reasonable. Here's a few examples for a comparison:

| **Current convention**                   | **with this PR's convention**                                    | **Notes** |
|----------------------------------        |----------------------------------                                |-----------|
| `ddog_prof_Profile_NewResult`            | `ddog_prof_Result_HandleProfile`                                 | |
| `ddog_prof_Exporter_NewResult`           | `ddog_prof_Result_Exporter`                                      | |
| `ddog_prof_Exporter_Request_BuildResult` | `ddog_prof_Result_Request`? `ddog_prof_Result_Exporter_Request`? | How to handle the deeper namespace?
| `ddog_prof_Exporter_SendResult`          | `ddog_prof_Result_HttpStatus` ? `ddog_Result_HttpStatus` ?       | How to handle when the type in the result is not named after the operation? |
| `ddog_prof_Profile_Result`               | `ddog_prof_ResultBool`? `ddog_ResultBool`?                       | (Similar to above) |
| `ddog_prof_Profile_SerializeResult`      | `ddog_prof_Result_EncodedProfile` ||

(There's also a few more results in ddcommon-ffi)

**Not wanting to discourage changes** (_I think one of the key ingredients of libdatadog's massive success so far has been having "no sacred cows", and instead always striving to leave things better, even if in the very short term it causes a bit of migration pain_) I'll suggest that if we plan to introduce the new convention, I don't think we should mix both -- I think we should rip-off the band-aid and convert everything to the new convention.

PR #798, opened 2024-12-12
Change profiling build to first generate the new header in ./target,and then run deduplication

paullegranddc:
# What does this PR do?

* Change profiling build to first generate the new header in ./target,and then run deduplication
* Remove the `data-pipeline-ffi` feature from the default features used in the shell build script

# Motivation

The way we currently do header generation is not great.
The build.rs directly emit the headers in the final output directory of the script. They are then modified in place to deduplicate headers.

This has multiple issues on unclean builds:
* If we delete the final output dir, but haven't modified the rust code, the headers are not re-generated and the build fails
* If we run the script with multiple features, we will leave previous header files there which creates error

The changes makes it so that:
* The bindgen build.rs scripts always output to ./target/inlcude
* We then copy the headers that we care about for the current set of feature flags to the final output directory
* We run header deduplication on the copied header

Also somehow the data-pipeline–ffi symbols were built by default in the profiling lib, but the headers for the symbols were only generated when the ARG_FEATURES flag was passed. This doesn't make sense because we're adding size and extra symbols to a shared lib that cannot be used because the headers are lacking.
I assumed this is an error and removed data-pipeline–ffi from the default features we build with. The crate feature can still be added mannualy though

# Additional Notes

# How to test the change?


pawelchcki (approved):
👍

bantonsson (approved):

build-common/src/cbindgen.rs
```diff
@@ -9,12 +9,19 @@ use std::str;
 
 pub const HEADER_PATH: &str = "include/datadog";
 
+pub struct OutPaths {
+    // Path to the default `./target directory where cargo outputs the result of
+    // it's build process `
+    pub cargo_target_dir: PathBuf,
+
+    // Directory identified by the DESTDIR env variable.
+    // It is taken realtive to the cargo project root.
```
hoolioh, line 18: Nit: relative

build-profiling-ffi.sh
```diff
@@ -224,15 +223,25 @@ echo "Building tools"
...
-        HEADERS="$HEADERS $destdir/include/datadog/data-pipeline.h"
+        HEADERS="$HEADERS data-pipeline.h"
         ;;
 esac
+DEST_HEADERS=""
+for header in $HEADER_FILES; do
+    cp "$CARGO_TARGET_DIR/include/datadog/$header" "$destdir/include/datadog/$header"
+    DEST_HEADERS="$DEST_HEADERS $destdir/include/datadog/$header"
+done
+# Copy the blazesym header separately because they don't need to be duplicated
```
hoolioh, line 240: deduplicated?

hoolioh (approved):

build-profiling-ffi.sh
```diff
@@ -140,7 +140,6 @@ FEATURES=(
     "cbindgen"
     "crashtracker-collector"
     "crashtracker-receiver"
-    "data-pipeline-ffi"
```
ganeshnj, line 143: We would like to keep this as default feature shipped. It seems like headers issue is unrelated to what is included in the features.

build-profiling-ffi.sh
paullegranddc: Sorry, this is not clear. The blazesym header isn't auto-generated by cbindgen so we don't need to move remove definitions that are already present in `common.h`

build-profiling-ffi.sh
paullegranddc: Re-added

build-profiling-ffi.sh
paullegranddc: Fixed the comment

PR #799, opened 2024-12-12
[trace-mini-agent] change logging crate

duncanista:
# What does this PR do?

Changes the logging crate from `log` to `tracing`

# Motivation

So instrumentation can be controlled using `tracing-subscriber`

# Additional Notes

# How to test the change?


duncanpharvey (approved):
LGTM!

bantonsson (approved):
LGTM

Would love to get some more info on this comment though
> So instrumentation can be controlled using `tracing-subscriber`

I see logging in the code, but what _instrumentations_ are we talking about?

duncanista:
> LGTM
> 
> I see logging in the code, but what _instrumentations_ are we talking about?

Any other crate/library/code using this as a dependency, should be able to control the logs from this crate through a subscriber's configuration.

This crate is sometimes noisy as it logs with `info`, and this might be an issue for customers when we are just accepting some traces.

Although the subscriber works with `log`, through experimentation at least, I'd rather have crate-consistency in all projects.
And, since also `tracing` is the [de-facto blessed crate](https://blessed.rs/crates#section-logging) for logging. Although we could debate about which one we really want to stick with!

bantonsson:
@duncanista I have absolutely no objections to the switch to `tracing` (god that name is confusing). I was more commenting on the use of the wording `So instrumentation can be controlled` in the description of the PR, which to me is about enabling/disabling/controlling an `instrumentation` which this seems completely unrelated to since it's about logging.

PR #800, opened 2024-12-12
Svls 6036 set timeouts

alexgallotta:
Allow passing a timeout to the client


duncanista (approved):

serverless/src/main.rs
```diff
@@ -160,6 +160,7 @@ async fn start_dogstatsd(
                 Arc::clone(&metrics_aggr),
                 build_fqdn_metrics(dd_site),
                 https_proxy,
+                Duration::from_secs(5),
```
ekump, line 163: nitpick: I typically like to put these sorts of values in consts so it's clearer what their purpose is.

dogstatsd/src/datadog.rs
```diff
@@ -20,8 +21,13 @@ pub struct DdApi {
 
 impl DdApi {
     #[must_use]
-    pub fn new(api_key: String, site: String, https_proxy: Option<String>) -> Self {
-        let client = match Self::build_client(https_proxy) {
+    pub fn new(
```
ekump, line 24: nitpick: I think it's ok right now, but if we go beyond four parameters here we should consider implementing a Dogstatsd builder.

ekump (approved):

serverless/src/main.rs
duncanista: On it

dogstatsd/src/datadog.rs
duncanista: Agree, I'll do it in another PR tho!

PR #802, opened 2024-12-16
Move telemetry client implementation to data pipeline.

hoolioh:
* Create a new builder for the TelemetryClient object.
* Add implementation for the TelemetryClient.
* Move metrics to data pipeline.

# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/data-pipeline/src/lib.rs
+++ b/data-pipeline/src/lib.rs
@@ -13,4 +13,6 @@ pub mod span_concentrator;
 #[allow(missing_docs)]
 pub mod stats_exporter;
 #[allow(missing_docs)]
+pub mod telemetry;
+#[allow(missing_docs)]
 pub mod trace_exporter;
--- a/data-pipeline/src/telemetry/mod.rs
+++ b/data-pipeline/src/telemetry/mod.rs
@@ -1,0 +1,286 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
+use datadog_trace_utils::trace_utils::SendDataResult;
+use ddcommon::tag;
+use ddcommon::tag::Tag;
+use ddtelemetry::data::metrics::{MetricNamespace, MetricType};
+use ddtelemetry::metrics::ContextKey;
+use ddtelemetry::worker::{
+    LifecycleAction, TelemetryActions, TelemetryWorkerBuilder, TelemetryWorkerHandle,
+};
+use std::collections::HashMap;
+use std::future::Future;
+use std::time::Duration;
+use tokio::select;
+use tokio::task::JoinHandle;
+
+pub trait CancellationObject {
+    fn shutdown(&self) -> impl Future<Output = ()>;
+}
+
+#[derive(Default)]
+pub struct Metrics {
+    pub active_sessions: usize,
+    pub api_requests: u64,
+    pub api_responses_count_per_code: HashMap<u16, u64>,
+    pub api_errors_timeout: u64,
+    pub api_errors_network: u64,
+    pub api_errors_status_code: u64,
+    pub bytes_sent: u64,
+    pub chunks_sent: u64,
+    pub chunks_dropped: u64,
+    pub submitted_payloads: u64,
+    pub memory_usage: usize,
+}
+
+impl Metrics {
+    pub fn update(&mut self, result: &SendDataResult) {
+        self.api_requests += result.requests_count;
+        self.api_errors_timeout += result.errors_timeout;
+        self.api_errors_network += result.errors_network;
+        self.api_errors_status_code += result.errors_status_code;
+        self.bytes_sent += result.bytes_sent;
+        self.chunks_sent += result.chunks_sent;
+        self.chunks_dropped += result.chunks_dropped;
+
+        for (status_code, count) in &result.responses_count_per_code {
+            *self
+                .api_responses_count_per_code
+                .entry(*status_code)
+                .or_default() += count;
+        }
+    }
+}
+
+struct SelfMetric(MetricType, MetricNamespace);
+
+#[derive(Default)]
+pub struct TelemetryClientBuilder {
+    service_name: Option<String>,
+    language: Option<String>,
+    language_version: Option<String>,
+    tracer_version: Option<String>,
+    config: ddtelemetry::config::Config,
+    metrics: HashMap<String, SelfMetric>,
+    interval: u64,
+    url: Option<String>,
+}
+
+impl TelemetryClientBuilder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn set_service_name(mut self, name: &str) -> Self {
+        self.service_name = Some(name.to_string());
+        self
+    }
+
+    pub fn set_language(mut self, lang: &str) -> Self {
+        self.language = Some(lang.to_string());
+        self
+    }
+
+    pub fn set_language_version(mut self, version: &str) -> Self {
+        self.language_version = Some(version.to_string());
+        self
+    }
+
+    pub fn set_tracer_version(mut self, version: &str) -> Self {
+        self.tracer_version = Some(version.to_string());
+        self
+    }
+
+    pub fn add_metric(
+        mut self,
+        name: &str,
+        metric_type: MetricType,
+        namespace: MetricNamespace,
+    ) -> Self {
+        self.metrics
+            .insert(name.to_string(), SelfMetric(metric_type, namespace));
+        self
+    }
+
+    pub fn set_interval(mut self, msecs: u64) -> Self {
+        self.interval = msecs;
+        self
+    }
+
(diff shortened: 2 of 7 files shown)


data-pipeline/src/telemetry/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+use std::time::Duration;
+use tokio::select;
+use tokio::task::JoinHandle;
+
+#[derive(Debug, PartialEq)]
+struct SelfMetric(MetricType, MetricNamespace);
+
+/// Structure to build a Telemetry client.
+///
+/// It will hold partial data until `build` method is called which result in a new
```
ekump, line 24:
```suggestion
/// Holds partial data until the `build` method is called which results in a new
```

data-pipeline/src/telemetry/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+    tracer_version: Option<String>,
+    metrics: HashMap<String, SelfMetric>,
+    heartbeat: u64,
+    interval: u64,
+    url: Option<String>,
+}
+
+impl TelemetryClientBuilder {
+    /// Creates a new empty builder.
+    pub fn new() -> Self {
```
ekump, line 40: I don't feel strongly about this, but if we're just returning a default `Self`, do we need a `new()` method at all?

data-pipeline/src/lib.rs
```diff
@@ -13,4 +13,6 @@ pub mod span_concentrator;
 #[allow(missing_docs)]
 pub mod stats_exporter;
 #[allow(missing_docs)]
```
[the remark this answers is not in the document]
hoolioh: Agree. I added it at the beginning at the project just to start playing around with the implementation and I forgot to remove it.

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -640,6 +647,9 @@ impl TraceExporter {
                 let send_data = SendData::new(size, tracer_payload, header_tags, &endpoint);
                 self.runtime.block_on(async {
                     let send_data_result = send_data.send().await;
+                    if let Some(telemetry) = &self.telemetry {
+                        telemetry.send(&send_data_result).await;
+                    }
```
paullegranddc, line 652: Are you sure we want to do this synchronously with the trace export?
I think that in term of qos, traces should not be lagging if telemetry metrics are, for any reasons and I'd rather see us use the non async, non blocking methods on the TelemetryWorkerHandle (which might drop points if there is backpressure)

data-pipeline/src/telemetry/mod.rs
```diff
@@ -0,0 +1,568 @@
...
+            if let Some(key) = self.metrics.get(API_REQUEST_STR) {
+                futures.push(self.add_point(data.requests_count as f64, *key, vec![]));
+            }
+        }
+        if data.errors_network > 0 {
+            if let Some(key) = self.metrics.get(API_ERRORS_STR) {
+                futures.push(self.add_point(
+                    data.errors_network as f64,
+                    *key,
+                    vec![tag!("type", "network")],
```
paullegranddc, line 128: if you are hardcoding tags, it'll be more efficient to add them when registering the metric contexts instead if on every points

data-pipeline/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,121 @@
...
+    Metric {
+        name: CHUNKS_DROPPED_STR,
+        metric_type: MetricType::Count,
+        namespace: MetricNamespace::Tracers,
+    },
+];
+
+/// Structure to accumulate partial results coming from sending traces to the agent.
+#[derive(Debug)]
+pub struct Metrics(HashMap<String, ContextKey>);
```
paullegranddc, line 65: I'm not a fan of using hashmaps to lookup harcoded keys, for performance and correctness,  but I don't really have a good counter proposition that'd be shorter to write...
Maybe associate an index to each metric name and hold Contexts in a Vec/array?

data-pipeline/src/trace_exporter/mod.rs
ekump: +1, I thought we discussed that we'd be sending telemetry async, unless I misunderstood something.

data-pipeline/src/telemetry/mod.rs
ekump: +1. Perhaps we could also use consts if we're hardcoding strs?

data-pipeline/src/trace_exporter/mod.rs
hoolioh: Just to give some context to the discussion, the 'send' method just adds the points to the corresponding bucket inside the telemetry client. I think this is done through a tokio channel. The sending itself it's done in a separate tokio task which is spawned at the trace exporter instantiation and is triggered at a certain interval which is configurable through the trace exporter configuration. Having said that, Paul are you referring to the fact that we can overwhelm the tokio, running out of its capacity and trigger some waiting on the current task?

data-pipeline/src/telemetry/metrics.rs
hoolioh: I thought about that when coding it but chose the easy path. Since the input is known ahead of time we could save some time from the hashing function. I would try some implementation and run some benchmarks.

ekump: No objection to starting off with a simpler implementation. But would it be a big deal to add a benchmark test to measure this? It'll be easier for us to observe performance changes in the future.

data-pipeline/src/trace_exporter/mod.rs
paullegranddc: > Paul are you referring to the fact that we can overwhelm the tokio, running out of its capacity and trigger some waiting on the current task?

Yes, if I remember correctly (and it hasn't changed), the telemetry sender is structured as a dispatch loop pulling data (like points) and events (like flush metrics) from the channel.
There is a lot of buffering for the channel (5K messages) but of the http requests are lagging a lot it might still be full and block traces.

What I'm sugesting is that instead of using the async methods on the TelemetySender, which wait for space to be available in the channel, we use the sync ones (like `TelemetryWorkerHandle::add_point`, which use try_send and will drop points  if the channel is full)

data-pipeline/src/telemetry/metrics.rs
paullegranddc: Yeah, of course, it shouldn't be that expensive

hoolioh: I did it anyway. When hardcoding the tags for the `trace_api.error` metrics I realized I could squeeze it without a major refactor.

data-pipeline/src/telemetry/mod.rs
```diff
@@ -0,0 +1,540 @@
...
+    pub async fn build(self) -> Result<TelemetryClient, TelemetryError> {
+        let (worker, handle) = TelemetryWorkerBuilder::new_fetch_host(
+            self.service_name.unwrap(),
+            self.language.unwrap(),
+            self.language_version.unwrap(),
+            self.tracer_version.unwrap(),
+        )
+        .spawn_with_config(self.config)
+        .await
+        .map_err(|_| TelemetryError::Builder)?;
```
ekump, line 81: Is there value in including more error details for the Builder error?

data-pipeline/src/telemetry/mod.rs
```diff
@@ -0,0 +1,540 @@
...
+            self.worker
+                .add_point(data.chunks_dropped as f64, key, vec![])?;
+        }
+        if !data.responses_count_per_code.is_empty() {
+            let key = self.metrics.get(metrics::MetricKind::ApiResponses);
+            for (status_code, count) in &data.responses_count_per_code {
+                self.worker.add_point(
+                    *count as f64,
+                    key,
+                    vec![Tag::new("status_code", status_code.to_string().as_str()).unwrap()],
```
ekump, line 145: Is unwrapping here ok? Should we be handling potential errors?

ekump (approved):
couple of questions / suggestions but LGTM

data-pipeline/src/telemetry/metrics.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn index(&self, index: MetricKind) -> &Self::Output {
+        &self.0[index as usize]
+    }
+}
+
+impl Metrics {
+    /// Creates a new Metrics instance
+    pub fn new(worker: &TelemetryWorkerHandle) -> Self {
+        let mut keys = Vec::new();
+        for metric in METRICS {
```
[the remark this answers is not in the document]
hoolioh: I think it's an interesting idea. I think metric names and tag names need to be known in advance so it makes sense to make static. We could do the same as in the tags where the reference to the name is behind a `Cow`.

data-pipeline/src/telemetry/mod.rs
hoolioh: Tag creation only fails if the tag name begins or ends with a `:` or is empty. That said there's little chances that this will panic but I think we need to avoid unwrap as much as possible in production code so I will refactor it.

PR #808, opened 2024-12-17
[fix] Add back set_cgroup_file function.

paullegranddc:
# What does this PR do?

Add back set_cgroup_file function. This adds a static variable used to override the default cgroup path.

# Motivation

This code was removed in #795, but this function is unfortunately directly imported and used in ddtrace-php here https://github.com/DataDog/dd-trace-php/blob/0926c77c3b8e0d41dd79774f89cd5f3dbbf2a03c/components-rs/lib.rs#L46-L49 so we cannot remove it.

They also use this feature in their tests

# Additional Notes

# How to test the change?

### DIFF
--- a/ddcommon/src/entity_id/unix/mod.rs
+++ b/ddcommon/src/entity_id/unix/mod.rs
@@ -4,7 +4,6 @@
 use lazy_static::lazy_static;
 use std::error;
 use std::fmt;
-use std::ops::Deref;
 use std::path::Path;
 
 mod cgroup_inode;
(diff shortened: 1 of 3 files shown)


ddcommon/src/entity_id/unix/mod.rs
```diff
@@ -52,8 +56,18 @@ fn compute_entity_id(
         )
 }
 
+/// Set cgroup mount path to mock during tests
+/// # Safety
+/// Must not be called in multi-threaded contexts
+pub fn set_cgroup_file(path: String) {
```
ekump, line 62: This question applies to all of the functions in the PR...Can we gate them behind a feature like `cgroup-test`? I know we are just adding back previous functionality, but I feel uneasy exposing functions intended for tests publicly like this. Forcing a consumer to opt-in is better than nothing.

paullegranddc: This is certainly not ideal. I was not aware that ddtrace php imported these directly...
The issue is that currently ddtrace php cannot compile their rust code in the repo because that function was removed.

ekump (approved):

PR #810, opened 2024-12-20
[Crashtracking] Add missing telemetry headers

kevingosse:
# What does this PR do?

Adds `api_version` and `request_type` headers to the telemetry messages sent by crashtracking.

# Motivation

Those headers are actually marked as required in the specification. I'm not actually sure _why_ it has worked until now (maybe the agent is adding them?) but it messes up with some tests.


pawelchcki (approved):

gleocadie (approved):
LGTM

PR #815, opened 2025-01-06
[WIP][Profiling] Prevent panic from crossing FFI boundaries

gleocadie:
# What does this PR do?

This PR is a communication channel to discuss about preventing panic from crossing FFI boundaries.
The code change here is to discuss, and find the best way to achieve this.

# Motivation

Playing around with the new crashtracker api, we got a crash (https://github.com/DataDog/libdatadog/pull/756) because there were missing information. The example was not modified nor ran which left with an unusable version of libdatadog (crashtracking)
This may be a misusage and the client's code must be changed, but in no case, this should panic. (at the minimum, there should be tests.

# Additional Notes

# How to test the change?

### DIFF
--- a/profiling-ffi/src/exporter.rs
+++ b/profiling-ffi/src/exporter.rs
@@ -11,6 +11,7 @@ use ddcommon::tag::Tag;
 use ddcommon_ffi::slice::{AsBytes, ByteSlice, CharSlice, Slice};
 use ddcommon_ffi::{Error, MaybeError, Timespec};
 use std::borrow::Cow;
+use std::panic::{AssertUnwindSafe, catch_unwind};
 use std::ptr::NonNull;
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


ivoanjo:
This is a complex subject, but I think we've been avoiding addressing it for a bit too long. 

The ideal mode of operation would be to have no aborts, but in the real world sometimes one slips by...

I think having a "last line of defense" on FFI functions that returns a clean error makes sense. Of course in some cases returning an error means the rust state can be in an inconsistent state, but it seems reasonable to put the burden on the API client to properly handle errors by backing off and stopping e.g. profiling, rather than tearing the whole app down. E.g. an abort is an extreme event, so I think stopping datadog stuff and logging an error (possibly via crashtracking...) is a reasonable outcome to such an extreme event (and is a much better alternative than killing off the host app). 

We could look into enabling lints to make it easier for us to catch things that may cause abort, but I suspect that will hinder and make really awkward a lot of the code we want to write. The Linux kernel folks have faced similar challenges, so perhaps there's some learnings there ([I found this link on a quick googling](https://www.reddit.com/r/rust/comments/xvorsd/what_does_rust_in_the_linux_kernel_do_with_bounds/)).

TL;DR: If we could write a nice macro to add this feature to most of our FFI, I think it's a reasonable and worth improvement.

gleocadie:
> This is a complex subject, but I think we've been avoiding addressing it for a bit too long.
> 
> The ideal mode of operation would be to have no aborts, but in the real world sometimes one slips by...
> 
> I think having a "last line of defense" on FFI functions that returns a clean error makes sense. Of course in some cases returning an error means the rust state can be in an inconsistent state, but it seems reasonable to put the burden on the API client to properly handle errors by backing off and stopping e.g. profiling, rather than tearing the whole app down. E.g. an abort is an extreme event, so I think stopping datadog stuff and logging an error (possibly via crashtracking...) is a reasonable outcome to such an extreme event (and is a much better alternative than killing off the host app).
> 
> We could look into enabling lints to make it easier for us to catch things that may cause abort, but I suspect that will hinder and make really awkward a lot of the code we want to write. The Linux kernel folks have faced similar challenges, so perhaps there's some learnings there ([I found this link on a quick googling](https://www.reddit.com/r/rust/comments/xvorsd/what_does_rust_in_the_linux_kernel_do_with_bounds/)).
> 
> TL;DR: If we could write a nice macro to add this feature to most of our FFI, I think it's a reasonable and worth improvement.

That's why this draft is here :) to talk about that subject.
Thanks for your comment :)

PR #816, opened 2025-01-09
Revert rasp products

estringana:
# What does this PR do?

Neither SSRF or LFI are products and instead they are capabilities. They shouldn't have added here. This PR reverts that.

# Motivation

# Additional Notes

# How to test the change?


Anilm3 (approved):
LGTM

PR #819, opened 2025-01-13
[crashtracker] Add incomplete flag to stack traces

danielsn:
# What does this PR do?

Adds an `incomplete` flag to stack traces, and an RFC explaining why.

# Motivation

Sometimes stack traces are incomplete, and the consumer has no way of knowing.

# Additional Notes

# How to test the change?


sanchda (commented):
Yup, makes a lot of sense--nicely done!

sanchda (approved):

PR #820, opened 2025-01-13
APMSP-1711 check CARGO_TARGET_DIR when calculating artifact location for bin tests

ekump:
# What does this PR do?

When bin_tests generate binaries check if `$CARGO_TARGET_DIR` is set, and the target directory is not the default `target/`. If this is the case then use this as the root of the artifact location.

# Motivation

bin_tests fail when attempting to execute them inside a linux docker container that has $CARGO_TARGET_DIR set in order to avoid build conflicts with the host OS.  

# Additional Notes

# How to test the change?

### DIFF
--- a/bin_tests/src/lib.rs
+++ b/bin_tests/src/lib.rs
@@ -5,7 +5,6 @@ pub mod modes;
 
 use std::{collections::HashMap, env, ops::DerefMut, path::PathBuf, process, sync::Mutex};
 
-use anyhow::Ok;
 use once_cell::sync::OnceCell;
 
 /// This crate implements an abstraction over compilation with cargo with the purpose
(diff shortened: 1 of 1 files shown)


bin_tests/src/lib.rs
```diff
@@ -69,6 +69,11 @@ fn inner_build_artifact(c: &ArtifactsBuild) -> anyhow::Result<PathBuf> {
     /// it's directory
     static ARTIFACT_DIR: OnceCell<PathBuf> = OnceCell::new();
     let artifact_dir = ARTIFACT_DIR.get_or_init(|| {
+        // If the CARGO_TARGET_DIR env var is set, then just use that.
+        if let std::result::Result::Ok(env_target_dir) = env::var("CARGO_TARGET_DIR") {
```
danielsn, line 73:
```suggestion
        if let Ok(env_target_dir) = env::var("CARGO_TARGET_DIR") {
```

danielsn (approved):

bin_tests/src/lib.rs
ekump: I had to provide the path because [anyhow::Ok is included](https://github.com/DataDog/libdatadog/blob/014b27e731d42445916090b501190fbff0a11de2/bin_tests/src/lib.rs#L8). But removing that `use` doesn't break anything so I'll do that.

PR #822, opened 2025-01-14
chore: [SVLS-5992] clearer dogstatsd Flusher API

apiarian-datadog:
# What does this PR do?

Adds structure to the dogstatsd `Flusher::new` api, including a better name for the old `site` parameter and a type for it.

# Motivation

The old `site` parameter for `Flusher::new` was misleading since it wasn't supposed to be the `DD_SITE` value but instead the URL prefix for that `DD_SITE`'s api for intake.

# How to test the change?

Ran the test suite.

### DIFF
--- a/dogstatsd/Cargo.toml
+++ b/dogstatsd/Cargo.toml
@@ -11,7 +11,7 @@ bench = false
 [dependencies]
 datadog-protos = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki/", rev = "c89b58e5784b985819baf11f13f7d35876741222" }
 ddsketch-agent = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki/", rev = "c89b58e5784b985819baf11f13f7d35876741222" }
-derive_more = { version = "1.0.0", features = ["full"] }
+derive_more = { version = "1.0.0", features = ["display", "into"] }
 hashbrown = { version = "0.14.3", default-features = false, features = ["inline-more"] }
 protobuf = { version = "3.5.0", default-features = false }
 ustr = { version = "1.0.0", default-features = false }
--- a/dogstatsd/Cargo.toml
+++ b/dogstatsd/Cargo.toml
@@ -11,6 +11,7 @@ bench = false
 [dependencies]
 datadog-protos = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki/", rev = "c89b58e5784b985819baf11f13f7d35876741222" }
 ddsketch-agent = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki/", rev = "c89b58e5784b985819baf11f13f7d35876741222" }
+derive_more = { version = "1.0.0", features = ["full"] }
 hashbrown = { version = "0.14.3", default-features = false, features = ["inline-more"] }
 protobuf = { version = "3.5.0", default-features = false }
 ustr = { version = "1.0.0", default-features = false }
--- a/dogstatsd/src/datadog.rs
+++ b/dogstatsd/src/datadog.rs
@@ -25,7 +25,7 @@ impl Site {
         // Datadog sites are generally domain names. In particular, they shouldn't have any slashes
         // in them. We expect this to be coming from a `DD_SITE` environment variable or the `site`
         // config field.
-        let re = Regex::new(r"^[a-zA-Z0-9._-]+$").expect("invalid regex");
+        let re = Regex::new(r"^[a-zA-Z0-9._:-]+$").expect("invalid regex");
         if re.is_match(&site) {
             Ok(Site(site))
         } else {
@@ -39,7 +39,7 @@ impl Site {
 pub struct UrlPrefixError(String);
 
 fn validate_url_prefix(prefix: &str) -> Result<(), UrlPrefixError> {
-    let re = Regex::new(r"^https?://[a-zA-Z0-9._-]+$").expect("invalid regex");
+    let re = Regex::new(r"^https?://[a-zA-Z0-9._:-]+$").expect("invalid regex");
     if re.is_match(prefix) {
         Ok(())
     } else {
@@ -88,14 +88,16 @@ pub struct MetricsIntakeUrlPrefix(String);
 pub struct MissingIntakeUrlError;
 
 impl MetricsIntakeUrlPrefix {
-    /// # Safety
-    ///
-    /// This function is unsafe because it does no validation on the input string. It also does not
-    /// follow our convention of using the metrics intake url prefix over the site name. This is
-    /// fine for tests, but in production we should be using new instead.
     #[inline]
-    pub unsafe fn new_unchecked(prefix: String) -> Self {
-        Self(prefix)
+    pub fn new(
+        site: Option<Site>,
+        overridden_prefix: Option<MetricsIntakeUrlPrefixOverride>,
+    ) -> Result<Self, MissingIntakeUrlError> {
+        match (site, overridden_prefix) {
+            (None, None) => Err(MissingIntakeUrlError),
+            (_, Some(prefix)) => Ok(Self::new_expect_validated(prefix.into())),
+            (Some(site), None) => Ok(Self::from_site(site)),
+        }
     }
 
     #[inline]
@@ -108,18 +110,6 @@ impl MetricsIntakeUrlPrefix {
     #[inline]
     fn from_site(site: Site) -> Self {
         Self(format!("https://api.{}", site))
-    }
-
-    #[inline]
-    pub fn new(
-        site: Option<Site>,
-        overridden_prefix: Option<MetricsIntakeUrlPrefixOverride>,
-    ) -> Result<Self, MissingIntakeUrlError> {
-        match (site, overridden_prefix) {
-            (None, None) => Err(MissingIntakeUrlError),
-            (_, Some(prefix)) => Ok(Self::new_expect_validated(prefix.into())),
-            (Some(site), None) => Ok(Self::from_site(site)),
-        }
     }
 }
 
--- a/dogstatsd/src/datadog.rs
+++ b/dogstatsd/src/datadog.rs
@@ -44,6 +44,19 @@ impl Site {
     }
 }
 
+#[derive(thiserror::Error, Debug, Clone, PartialEq)]
+#[error("Invalid URL prefix: {0}")]
+pub struct UrlPrefixError(String);
+
+fn validate_url_prefix(prefix: &String) -> Result<(), UrlPrefixError> {
+    let re = Regex::new(r"^https?://[a-zA-Z0-9._-]+$").unwrap();
+    if re.is_match(&prefix) {
+        Ok(())
+    } else {
+        Err(UrlPrefixError(prefix.clone()))
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct DdUrl(String);
 
@@ -54,8 +67,11 @@ impl fmt::Display for DdUrl {
 }
 
 impl DdUrl {
-    pub fn new(prefix: String) -> Self {
-        DdUrl(prefix)
+    pub fn new(prefix: String) -> Result<Self, UrlPrefixError> {
+        match validate_url_prefix(&prefix) {
+            Ok(_) => Ok(Self(prefix)),
+            Err(e) => Err(e),
+        }
     }
 
     pub fn into_string(self) -> String {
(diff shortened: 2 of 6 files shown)


apiarian-datadog:
Can be reviewed commit by commit.

dogstatsd/src/datadog.rs
```diff
@@ -8,22 +8,45 @@ use protobuf::Message;
...
+
+impl IntakeUrlPrefix {
+    pub fn new(prefix: String) -> Self {
+        // Maybe we will validate this in the future, but for now we assume that the prefix is
+        // sensible.
+        IntakeUrlPrefix(prefix)
+    }
+
+    #[inline]
+    pub fn from_site(site: String) -> Self {
```
apiarian-datadog, line 32: We could also write a `from_site_or_dd_url(site: Optional<String>, dd_url: Optional<String>, dd_dd_url: Optional<String>) -> Result<Self, OurError> { ... }` which would correctly handle our `DD_SITE`, `DD_URL`, and `DD_DD_URL` parameter values (not the actual environment/config parsing, but dealing with the results of those, i.e. picking `dd_dd_url` over `dd_url` over a `from_site(site)`).

Would that be the wrong level of abstraction? should that kind of logic live further up the stack somewhere?

apiarian-datadog: ok, i've implemented this in an atomic commit for us to discuss or roll back

apiarian-datadog:
is there a better pattern we should follow for these immutable string type aliases?

dogstatsd/src/datadog.rs
```diff
@@ -8,22 +8,176 @@ use protobuf::Message;
...
+        DdUrl(prefix)
+    }
+
+    pub fn into_string(self) -> String {
+        self.0
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct DdDdUrl(String);
```
alexgallotta, line 54: Do we need this code duplication? Or can we redirect all to DdUrl?

alexgallotta (commented):
I would suggest to do either one of 2 approach:
a- add a validation for every type, so we make sure we don't pass a prefix url as a site
b- have a single type, DdUrl or DdUri, that will accept a string. Depending on the format of the string, it will recognize if it's site/url/other and store a canonical format. Then it can be converted as_site, as_prefix_url, etc... as a normal string

what do you think?

dogstatsd/src/datadog.rs
```diff
@@ -8,22 +8,176 @@ use protobuf::Message;
...
+
+impl DdDdUrl {
+    pub fn new(prefix: String) -> Self {
+        DdDdUrl(prefix)
+    }
+
+    pub fn into_string(self) -> String {
+        self.0
+    }
+}
```
duncanista, line 70: can't this boilerplate just be done with a derive?

dogstatsd/src/datadog.rs
```diff
@@ -8,22 +8,176 @@ use protobuf::Message;
...
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct IntakeUrlPrefix(String);
+
+impl fmt::Display for IntakeUrlPrefix {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
```
duncanista, line 79: Same here

dogstatsd/src/datadog.rs
```diff
@@ -8,22 +8,176 @@ use protobuf::Message;
...
+        site: Option<Site>,
+        dd_url: Option<DdUrl>,
+        dd_dd_url: Option<DdDdUrl>,
+    ) -> Result<Self, &'static str> {
+        match (site, dd_url, dd_dd_url) {
+            (None, None, None) => Err("No intake URL configuration"),
+            (_, _, Some(dd_dd_url)) => Ok(Self::new(dd_dd_url.into_string())),
+            (_, Some(dd_url), None) => Ok(Self::new(dd_url.into_string())),
+            (Some(site), None, None) => Ok(Self::from_site(site.into_string())),
+        }
```
duncanista, line 115: Looks overly complex?

dogstatsd/src/flusher.rs
```diff
@@ -12,23 +13,27 @@ pub struct Flusher {
...
+        let dd_api = datadog::DdApi::new(
+            params.api_key,
+            params.intake_url_prefix,
+            params.https_proxy,
+            params.timeout,
+        );
+        Flusher {
+            dd_api,
+            aggregator: params.aggregator,
+        }
```
duncanista, line 36:
```suggestion
    pub fn new(config: FlusherConfig) -> Self {
        let dd_api = datadog::DdApi::new(
            config.api_key,
            config.intake_url_prefix,
            config.https_proxy,
            config.timeout,
        );
        Flusher {
            dd_api,
            aggregator: config.aggregator,
        }
```

dogstatsd/src/datadog.rs
apiarian-datadog: yup. but somewhat necessarily so? @alexgallotta and i were talking about moving some of this complexity up into the DdUrl to have that take either a DD_URL or a DD_DD_URL and do the right thing between them. so maybe that will make things less complex.

dogstatsd/src/flusher.rs
apiarian-datadog: ah, right, i renamed the type to config but forgot the variable, will fix!

dogstatsd/src/datadog.rs
apiarian-datadog: maybe! like this? https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/procedural-macros.html (this may be my first serious rust pr =] )

dogstatsd/src/datadog.rs
apiarian-datadog: i'll give it a try!

dogstatsd/src/datadog.rs
duncanista: Yup, read on the `Default,Display` derives

apiarian-datadog: wait, do you mean this one? https://docs.rs/derive_more/latest/derive_more/derive.Display.html can we add `derive_more` as a dependency?

duncanista: No – https://doc.rust-lang.org/rust-by-example/hello/print/print_display.html

duncanista: Actually, no, let's not add a new dependency – there should be an easier boilerplate of just getting the string, but idk until I see the new changes

dogstatsd/src/flusher.rs
apiarian-datadog: done!

dogstatsd/src/datadog.rs
apiarian-datadog: i've split this complexity in to a `MetricsIntakeUrlPrefixOverride` which sorts out the `DD_URL` vs `DD_DD_URL` logic from the `MetricsIntakeUrlPrefix` which handles the `DD_SITE` vs `MetricsIntakeUrlPrefixOverride` logic. so... at least each separate match block is a bit smaller?

dogstatsd/src/datadog.rs
```diff
@@ -5,25 +5,262 @@
...
+#[derive(thiserror::Error, Debug, Clone, PartialEq)]
+#[error("Invalid site: {0}")]
+pub struct SiteError(String);
+
+impl Site {
+    pub fn new(site: String) -> Result<Self, SiteError> {
+        // Datadog sites are generally domain names. In particular, they shouldn't have any slashes in
+        // them. We expect this to be coming from a `DD_SITE` environment variable or the `site`
+        // config field.
+        let re = Regex::new(r"^[a-zA-Z0-9._-]+$").unwrap();
```
duncanista, line 34: Don't use `unwrap`, prefer `expect`

dogstatsd/src/datadog.rs
apiarian-datadog: still a lot of boilerplate floating around, but this at least separates our site from metrics url prefix override logic. (we should be able to actually push the site stuff way up the stack in the future, i think.

will investigate deriving more things next.

dogstatsd/src/datadog.rs
apiarian-datadog: we can't say `type DdDdUrl = DdUrl` since they then could be used interchangeably which is the opposite of what we're trying to do here, unfortunately. maybe with some more clever derives we'll be able to eliminate some of this boilerplate, though?

dogstatsd/src/datadog.rs
apiarian-datadog: fixed!

dogstatsd/src/datadog.rs
```diff
@@ -5,25 +5,263 @@
...
+
+impl MetricsIntakeUrlPrefix {
+    /// # Safety
+    ///
+    /// This function is unsafe because it does no validation on the input string. It also does not
+    /// follow our convention of using the metrics intake url prefix over the site name. This is
+    /// fine for tests, but in production we should be using from_site_or_dd_urls instead.
+    #[inline]
+    pub unsafe fn new_unchecked(prefix: String) -> Self {
+        MetricsIntakeUrlPrefix(prefix)
```
duncanista, line 184:
```suggestion
        Self(prefix)
```

dogstatsd/src/datadog.rs
```diff
@@ -5,25 +5,263 @@
...
+    }
+}
+
+impl DdDdUrl {
+    pub fn new(prefix: String) -> Result<Self, UrlPrefixError> {
+        match validate_url_prefix(&prefix) {
+            Ok(_) => Ok(Self(prefix)),
+            Err(e) => Err(e),
+        }
+    }
```
duncanista, line 97:
```suggestion
        validate_url_prefix(&prefix)?
        Self(prefix)
    }
```

dogstatsd/src/datadog.rs
```diff
@@ -5,25 +5,263 @@
...
+        MetricsIntakeUrlPrefix(validated_prefix)
+    }
+
+    #[inline]
+    fn from_site(site: Site) -> Self {
+        MetricsIntakeUrlPrefix(format!("https://api.{}", site))
+    }
+
+    #[inline]
+    pub fn from_site_or_override(
```
duncanista, line 200:
```suggestion
    pub fn new(
```

dogstatsd/src/flusher.rs
```diff
@@ -12,23 +13,27 @@ pub struct Flusher {
...
 }
 
-#[inline]
-#[must_use]
-pub fn build_fqdn_metrics(site: String) -> String {
-    format!("https://api.{site}")
+pub struct FlusherConfig {
+    pub api_key: String,
+    pub aggregator: Arc<Mutex<Aggregator>>,
+    pub metrics_intake_url_prefix: MetricsIntakeUrlPrefix,
```
duncanista, line 19: Either use `datadog::...` or import also `DdApi`, let's not use both things here

dogstatsd/src/datadog.rs
```diff
@@ -5,25 +5,263 @@
 
 use datadog_protos::metrics::SketchPayload;
 use protobuf::Message;
+use regex::Regex;
```
duncanista, line 8: Let's move forward with your suggestion of using `derive_more` crate

duncanista (commented):
left some comments

dogstatsd/src/datadog.rs
apiarian-datadog: ok, with `display_more` a lot of this boilerplate disappears!

dogstatsd/src/datadog.rs
apiarian-datadog: fixed!

dogstatsd/src/datadog.rs
apiarian-datadog: fixed!

dogstatsd/src/flusher.rs
apiarian-datadog: removed the `datadog::` prefixes

dogstatsd/src/datadog.rs
apiarian-datadog: done!

dogstatsd/src/datadog.rs
apiarian-datadog: done!

dogstatsd/src/datadog.rs
apiarian-datadog: this boilerplate is now gone

dogstatsd/Cargo.toml
```diff
@@ -11,6 +11,7 @@ bench = false
 [dependencies]
 datadog-protos = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki/", rev = "c89b58e5784b985819baf11f13f7d35876741222" }
 ddsketch-agent = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki/", rev = "c89b58e5784b985819baf11f13f7d35876741222" }
+derive_more = { version = "1.0.0", features = ["full"] }
```
duncanista, line 14: the features should only be `display,into` as you're only using those

duncanista (approved):

dogstatsd/Cargo.toml
apiarian-datadog: fixed!

morrisonlevi (approved):
Approving from libcore for license and cargo lock changes only.

alexgallotta (approved):

duncanpharvey (approved):
LGTM!

apiarian-datadog:
@alexgallotta i tried changing the unsafe test helper to be test-only but wasn't able to get things to actually compile. if you think of a way to do that, we can make the change in the future!

apiarian-datadog:
> @alexgallotta i tried changing the unsafe test helper to be test-only but wasn't able to get things to actually compile. if you think of a way to do that, we can make the change in the future!

o, but i *was* able to eliminate the need for the unsafe by actually checking the url override!

dogstatsd/src/datadog.rs
```diff
@@ -4,26 +4,138 @@
...
+#[derive(thiserror::Error, Debug, Clone, PartialEq)]
+#[error("Invalid site: {0}")]
+pub struct SiteError(String);
+
+impl Site {
+    pub fn new(site: String) -> Result<Self, SiteError> {
+        // Datadog sites are generally domain names. In particular, they shouldn't have any slashes
+        // in them. We expect this to be coming from a `DD_SITE` environment variable or the `site`
+        // config field.
+        let re = Regex::new(r"^[a-zA-Z0-9._-]+$").expect("invalid regex");
```
apiarian-datadog, line 28: @alexgallotta it seems like in order to pull these out into static refs we'd need the `lazy_static` crate as a new dependency (not being used in libdatadog yet). since these regexes will presumably be checked only a couple of times at app startup, do you feel strongly that we should add the crate and make them static?

apiarian-datadog: ah, no, we do in fact have the dependency, i must have been spelling something wrong in my search. this is done.

PR #826, opened 2025-01-15
fix: support tag that are not key-value pairs

alexgallotta:
Allow tags that are not a pair key:value to be passed. Use an empty string as value and do not serialize a column with empty string in that case

### DIFF
--- a/dogstatsd/src/metric.rs
+++ b/dogstatsd/src/metric.rs
@@ -263,7 +263,7 @@ pub fn id(name: Ustr, tags: &Option<SortedTags>) -> u64 {
     hasher.finish()
 }
 // <METRIC_NAME>:<VALUE>:<VALUE>:<VALUE>|<TYPE>|@<SAMPLE_RATE>|#<TAG_KEY_1>:<TAG_VALUE_1>,
-// <TAG_KEY_2>:<TAG_VALUE_2>,<TAG_NO_VALUE3|T<METRIC_TIMESTAMP>|c:<CONTAINER_ID>
+// <TAG_KEY_2>:<TAG_VALUE_2>,<TAG_NO_VALUE_3>|T<METRIC_TIMESTAMP>|c:<CONTAINER_ID>
 //
 // Types:
 //  * c -- COUNT, allows packed values, summed
--- a/dogstatsd/src/metric.rs
+++ b/dogstatsd/src/metric.rs
@@ -91,7 +91,11 @@ impl SortedTags {
             if v.is_empty() {
                 tags_as_chars.push(Chars::from(k.to_string()));
             } else {
-                tags_as_chars.push(format!("{}:{}", k, v).into());
+                let mut a_tag = String::with_capacity(k.len() + v.len() + 1);
+                a_tag.push_str(k);
+                a_tag.push(':');
+                a_tag.push_str(v);
+                tags_as_chars.push(a_tag.into());
             }
         }
         tags_as_chars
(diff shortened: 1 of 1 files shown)


dogstatsd/src/metric.rs
```diff
@@ -89,15 +88,23 @@ impl SortedTags {
     pub fn to_chars(&self) -> Vec<Chars> {
         let mut tags_as_chars = Vec::new();
         for (k, v) in &self.values {
-            tags_as_chars.push(format!("{}:{}", k, v).into());
+            if v.is_empty() {
+                tags_as_chars.push(Chars::from(k.to_string()));
+            } else {
+                tags_as_chars.push(format!("{}:{}", k, v).into());
```
pawelchcki, line 94: Just a small performance nit - feel free to ignore if this code is not meant to be called often. 

I'm not sure how often this code is called. And I can see its been used previously.

But format! is somewhat inefficient with its allocations https://stackoverflow.com/questions/63690623/why-is-the-format-macro-slower-than-pushing-into-a-string-directly

Pushing to a string is better, bonus poitns for preallocating string with lenght.

pawelchcki (approved):
One small performance nit. Otherwise LGTM

dogstatsd/src/metric.rs
alexgallotta: Thanks for the feedback!
It's good to know! Right now that code is never used but there is no way to use an inefficient formatting.
I will change to push

dogstatsd/src/metric.rs
```diff
@@ -89,15 +88,23 @@ impl SortedTags {
     pub fn to_chars(&self) -> Vec<Chars> {
         let mut tags_as_chars = Vec::new();
         for (k, v) in &self.values {
-            tags_as_chars.push(format!("{}:{}", k, v).into());
+            if v.is_empty() {
+                tags_as_chars.push(Chars::from(k.to_string()));
```
tobz, line 92: For bare tags, you can optimize this a _little_ further:

- `Ustr::as_str` (the type of `k`) gives you a `&'static str`
- `Bytes::from_static` can be constructed from `&'static [u8]`, which you can get from the `&'static str` by calling `.to_bytes()`
- you can then jump from `Bytes` to `Chars` via `Chars::from_bytes` (which _does_ return a `Result<T, E>` but should be safe to call `.expect` on since the safety condition it's validating is that you passed in valid UTF-8, which we know we did since we got our `Bytes` directly from a `&str`, which should always be valid UTF-8`

The end result is that we'll always skip the allocation for bare tags, and the only cost we'll pay is the UTF-8 validation, which is minimal.

dogstatsd/src/metric.rs
tobz: Probably also worth allocating the string with the known necessary capacity upfront to avoid any reallocations:

```rust
let mut formatted = String::with_capacity(k.len() + v.len() + 1);
formatted.push_str(...);
...
```

tobz (approved):

dogstatsd/src/metric.rs
alexgallotta: This second optimization seems a bit less readable, add the `expect` path and requires an extra dependency for [Bytes::from_static](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_static), so I would just use a pre allocated string as in the similar code too, unless there is a strong feeling about optimizing this

PR #827, opened 2025-01-16
Mitigate race conditions on environments where /proc/<pid> does not immediately appear in the procfs

bwoebi:
This is apparently the case in google cloud run environments. We just retry with /proc/self then, and it works there.


morrisonlevi (approved):
Approved. Can we leave a code comment that we retry because we saw issues in google cloud run environments where it doesn't immediately appear?

PR #829, opened 2025-01-16
resources are only dd.internal.resource tags

alexgallotta:
avoid sending duplicates tags as resources when they are not prefixed with `dd.internal.resource`

### DIFF
--- a/dogstatsd/src/metric.rs
+++ b/dogstatsd/src/metric.rs
@@ -104,12 +104,18 @@ impl SortedTags {
 
     pub(crate) fn to_resources(&self) -> Vec<datadog::Resource> {
         let mut resources = Vec::with_capacity(constants::MAX_TAGS);
-        for (kind, name) in &self.values {
-            let resource = datadog::Resource {
-                name: name.as_str(), // val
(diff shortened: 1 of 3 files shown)


dogstatsd/src/metric.rs
```diff
@@ -104,12 +104,18 @@ impl SortedTags {
...
         let mut resources = Vec::with_capacity(constants::MAX_TAGS);
-        for (kind, name) in &self.values {
-            let resource = datadog::Resource {
-                name: name.as_str(), // val
-                kind: kind.as_str(), // key
-            };
-            resources.push(resource);
+        for (key, val) in &self.values {
+            if key == "dd.internal.resource" {
+                //anything coming in via dd.internal.resource:<value> has to be a key/value pair
```
duncanista, line 109: how was this not caught in testing? is there only unit testing and not integration testing?

alexgallotta: The only way to get this in testing would be with a proper json/binary payload to compare, which is not done afaik

pawelchcki (approved):

PR #831, opened 2025-01-17
Fix RC shmem synchronization protocol

cataphract:
Fixes a problem in the current protocol:

            w1. writing <- true (seq cst)
            w2. write data
            w3. gen <- gen + 1 (seq cst)
            w4. writing <- false (seq cst)
            
            r1. read gen (acquire)
            r2. read data
            r3. read writing (seq cst)
            r4. read gen (acquire)
            if writing is true or gen has changed, retry

Partial or full data writes for the following generation are observed in practice on armv8.

This introduces a seqlock implemenation that, even if not [formally correct according to the C++ memory model](https://web.archive.org/web/20240225081806/https://www.hpl.hp.com/techreports/2012/HPL-2012-68.pdf) (the data reads/writes should likely be made with something equivalent to relaxed `__atomic_load/store_n`), this generates reasonable code on armv8:

* Non-atomic write is surrounded by ldaddal and ldaddl on the atomic generation.
* Non-atomic read is preceded by ldapr and followed by dmb ishld before the second load of gen, this time made with a plain ldr.

It's also a bit simpler to reason about, because there is no separate `writing` variable.

### DIFF
--- a/sidecar/src/one_way_shared_memory.rs
+++ b/sidecar/src/one_way_shared_memory.rs
@@ -4,7 +4,7 @@
 use datadog_ipc::platform::{FileBackedHandle, MappedMem, NamedShmHandle, ShmHandle};
 use std::ffi::{CStr, CString};
 use std::io;
-use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::atomic::{fence, AtomicU64, Ordering};
 use std::sync::Mutex;
 
 pub struct OneWayShmWriter<T>
@@ -28,7 +28,6 @@ where
 struct RawMetaData {
     generation: AtomicU64,
     size: usize,
-    writing: AtomicBool,
 }
 
 #[repr(C)]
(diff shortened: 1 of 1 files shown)


sidecar/src/one_way_shared_memory.rs
```diff
@@ -124,6 +123,11 @@ where
                 unsafe { reinterpret_u8_as_u64_slice(handle.as_slice()) }.into();
             let new_generation = source_data.meta.generation.load(Ordering::Acquire);
 
+            if new_generation & 1 == 1 {
+                // mid-write
+                return (false, "".as_bytes());
```
bwoebi, line 128: This should return `(false, skip_last_byte(cur_data.as_slice()));`, not an empty string (i.e. the current valid copy of data.)

bwoebi (approved):
Thanks, the fence is definitely correct instead of the Ordering (which just synchronizes with itself and not generally)

PR #832, opened 2025-01-17
Prepare 15.0.0 release

paullegranddc:
# What does this PR do?

* Bump workspace version to 15.0.0
* Build configuration-ffi by default with the profiling artifacts so we can distribute it in the next release

# Motivation

The major version increase is necessary because of some breaking changes in the crashtracker APIs.

# Additional Notes

# How to test the change?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ resolver = "2"
 [workspace.package]
 rust-version = "1.78.0"
 edition = "2021"
-version = "14.3.1"
+version = "15.0.0"
 license = "Apache-2.0"
 
 [profile.dev]
--- a/build-profiling-ffi.sh
+++ b/build-profiling-ffi.sh
@@ -143,6 +143,7 @@ FEATURES=(
     "data-pipeline-ffi"
     "datadog-profiling-ffi/ddtelemetry-ffi"
     "datadog-profiling-ffi/demangler"
+    "datadog-library-config-ffi"
 )
 if [[ "$symbolizer" -eq 1 ]]; then
     FEATURES+=("symbolizer")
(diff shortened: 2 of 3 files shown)


build-profiling-ffi.sh
```diff
@@ -227,10 +228,10 @@ echo "Generating $destdir/include/libdatadog headers..."
 rm -r $destdir/include/datadog/
 mkdir $destdir/include/datadog/
 
-CBINDGEN_HEADERS="common.h profiling.h telemetry.h crashtracker.h data-pipeline.h"
+CBINDGEN_HEADERS="common.h profiling.h telemetry.h crashtracker.h data-pipeline.h library-config.h"
 # When optional features are added, don't forget to also include the headers here
 case $ARG_FEATURES in
-    datadog-library-config-ffi) CBINDGEN_HEADERS="$CBINDGEN_HEADERS library-config.h"
+    # datadog-library-config-ffi) CBINDGEN_HEADERS="$CBINDGEN_HEADERS library-config.h"
 esac
```
ivoanjo, line 235: Is this being commented out intended?

Cargo.toml
```diff
@@ -58,7 +58,7 @@ resolver = "2"
 [workspace.package]
 rust-version = "1.78.0"
 edition = "2021"
-version = "14.3.1"
+version = "14.4.0"
```
ivoanjo, line 61: I suspect this should be 15.0 given [there's breaking changes in the crashtracking API](https://github.com/DataDog/dd-trace-rb/pull/4255).

build-profiling-ffi.sh
paullegranddc: Yes, I should probably have removed the line completely

Cargo.toml
paullegranddc: Makes sense.
We also increased the minimum rust version to 1.78 which drops support for windows 7 (which none of our products currently support anyway)

paullegranddc: I fixed the PR to bump to 15.0.0

ivoanjo (approved):
:+1: LGTM

PR #834, opened 2025-01-17
[crashtracker] Receiver speaks RFC5

danielsn:
# What does this PR do?

Modifies the crashtracker receiver and emitter to speak RFC5, rather than the old ad-hoc format.

# Motivation

This is the last step before fully tearing out the old format

# Additional Notes

Anything else we should know when reviewing?

# How to test the change?

Existing tests

PR #836, opened 2025-01-21
[crashtracker] Remove the old dataformat

danielsn:
# What does this PR do?

Removes now unused code that supported the old data format, and makes RFC5 the only supported format.

# Motivation

With #834 we no longer use the old format anyway.  Dead code should be removed.

# Additional Notes

# How to test the change?


sanchda (approved):
Looks really good.  Was only able to offer a minor/style suggestion that wasn't even novel to the PR.

I agree with the current posture on holding back failures until the end--partial information is still useful.

ekump (approved):

PR #838, opened 2025-01-22
fix: handle a zero-sized file for `mmap()`ing

bwoebi:
Otherwise we unwrap() with zero in some possible race conditions.


morrisonlevi (approved):
> Size of handle used for mmap() is zero. When used for shared memory this may originate from race conditions between creation and truncation of the shared memory file.

I'm not familiar with the code details but shouldn't we be able to avoid this altogether?

In any case, avoiding the panic and coalescing it into an error is a direct improvement and avoids the panic, so approved regardless.

PR #840, opened 2025-01-22
[Library configuration] Expose path to stable configuration path through ffi

paullegranddc:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


library-config/src/lib.rs
```diff
@@ -331,6 +354,7 @@ fn string_operator_match(op: &Operator, matches: &[u8], value: &[u8]) -> bool {
 pub struct LibraryConfig {
     pub name: LibraryConfigName,
     pub value: String,
+    pub source: LibraryConfigSource,
```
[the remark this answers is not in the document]
BaptisteFoy: Agreed. I supported reading config_id from the YAML in [5d29f21](https://github.com/DataDog/libdatadog/pull/840/commits/5d29f218c038845f98d5477af97076591e5a710f), we'll drop it in from Fleet like we do for the agent.

library-config/src/lib.rs
```diff
@@ -363,28 +396,75 @@ impl Configurator {
...
+            &stable_config_local,
+            LibraryConfigSource::LocalFile,
+            &process_info,
+        );
+
+        let managed_config = self.get_config(
+            &stable_config_managed,
+            LibraryConfigSource::Managed,
+            &process_info,
+        );
```
paullegranddc, line 451: Do we want both configs to be applied? I thought we'd match managed config, and only use the local config if no rule was matching in the managed one

Also if we do it like that, we should probably do some dedup in order to not return the same config twice

library-config/src/lib.rs
```diff
@@ -240,6 +244,27 @@ impl LibraryConfigName {
...
+
+impl LibraryConfigSource {
+    pub fn to_str(&self) -> &'static str {
+        use LibraryConfigSource::*;
+        match self {
+            LocalFile => "local_file",
+            Managed => "fleet_automation",
+        }
+    }
+}
```
paullegranddc, line 266: This is not used anywhere?

BaptisteFoy: It's used in client libraries, exactly like `LibraryConfigName`

library-config/src/lib.rs
BaptisteFoy: Ah, yes, fair my bad.

library-config/src/lib.rs
```diff
@@ -331,6 +357,8 @@ fn string_operator_match(op: &Operator, matches: &[u8], value: &[u8]) -> bool {
 pub struct LibraryConfig {
     pub name: LibraryConfigName,
     pub value: String,
+    pub source: LibraryConfigSource,
+    pub config_id: Option<String>,
```
BaptisteFoy, line 361: tbh i'm not sure whether or not we should put it in the vec or not, this may become quite a lot of duplicated data

but i don't really have a better idea today

raphaelgavache: if/when selectors are introduced they can leverage that field too
ex - remote_{version}_{selector_id}

BaptisteFoy (approved):

raphaelgavache (approved):

PR #842, opened 2025-01-24
[crashtracking] Little improvements to the FFI api

gleocadie:
# What does this PR do?

This PR tries to improve a bit the crashtracker ffi APIs.

# Motivation

The goal is to simplify the crashtracker ffi API and adding code in the `crashinfo.cpp` example.

# Implementation

- When passing an address, instead of asking for a string, pass a `usize` (expecting it to be 64bit on x86_64, and 32bit on x86 architecture) and build a `0x`-prefixed string from it.
- *Experimental* : 
`ddog_crasht_CrashInfoBuilder_with_stack` adds the callstack of the crashing threads but only the callstack.
If we want to add the crashing thread (`ThreadData`), we have to duplicate the callstack and make one calls to `ddog_crasht_CrashInfoBuilder_with_stack` and `ddog_crasht_CrashInfoBuilder_with_thread`.
Instead, we check the `crashed` field, and call into `with_stack` to add a duplicate of the callstack thread (if it's true) and call into `with_thread` with the thread data 
Maybe we could remove the `ddog_crasht_CrashInfoBuilder_with_stack` ?
* Add `XXXNewResult` types to have shorter symbols for the generated code.

# Additional Notes

# How to test the change?

### DIFF
--- a/crashtracker-ffi/src/crash_info/builder.rs
+++ b/crashtracker-ffi/src/crash_info/builder.rs
@@ -5,7 +5,7 @@ use super::{Metadata, OsInfo, ProcInfo, SigInfo, Span, ThreadData};
 use ::function_name::named;
 use datadog_crashtracker::{CrashInfo, CrashInfoBuilder, ErrorKind, StackTrace};
 use ddcommon_ffi::{
-    slice::AsBytes, wrap_with_ffi_result, wrap_with_void_ffi_result, CharSlice, Handle, Result,
+    slice::AsBytes, wrap_with_ffi_result, wrap_with_void_ffi_result, CharSlice, Error, Handle,
     Slice, Timespec, ToInner, VoidResult,
 };
 
@@ -13,13 +13,20 @@ use ddcommon_ffi::{
 //                                              FFI API                                           //
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#[allow(dead_code)]
+#[repr(C)]
+pub enum CrashInfoBuilderNewResult {
+    Ok(Handle<CrashInfoBuilder>),
+    Err(Error),
+}
+
 /// Create a new CrashInfoBuilder, and returns an opaque reference to it.
 /// # Safety
 /// No safety issues.
 #[no_mangle]
 #[must_use]
-pub unsafe extern "C" fn ddog_crasht_CrashInfoBuilder_new() -> Result<Handle<CrashInfoBuilder>> {
-    ddcommon_ffi::Result::Ok(CrashInfoBuilder::new().into())
+pub unsafe extern "C" fn ddog_crasht_CrashInfoBuilder_new() -> CrashInfoBuilderNewResult {
+    CrashInfoBuilderNewResult::Ok(CrashInfoBuilder::new().into())
 }
 
 /// # Safety
@@ -34,15 +41,31 @@ pub unsafe extern "C" fn ddog_crasht_CrashInfoBuilder_drop(builder: *mut Handle<
     }
 }
 
+#[allow(dead_code)]
+#[repr(C)]
+pub enum CrashInfoNewResult {
+    Ok(Handle<CrashInfo>),
+    Err(Error),
+}
+
 /// # Safety
 /// The `builder` can be null, but if non-null it must point to a Builder made by this module,
 /// which has not previously been dropped.
 #[no_mangle]
 #[must_use]
-#[named]
 pub unsafe extern "C" fn ddog_crasht_CrashInfoBuilder_build(
+    builder: *mut Handle<CrashInfoBuilder>,
+) -> CrashInfoNewResult {
+    match ddog_crasht_crash_info_builder_build_impl(builder) {
+        Ok(crash_info) => CrashInfoNewResult::Ok(crash_info),
+        Err(err) => CrashInfoNewResult::Err(err.into()),
+    }
+}
+
+#[named]
+unsafe fn ddog_crasht_crash_info_builder_build_impl(
     mut builder: *mut Handle<CrashInfoBuilder>,
-) -> Result<Handle<CrashInfo>> {
+) -> anyhow::Result<Handle<CrashInfo>> {
     wrap_with_ffi_result!({ anyhow::Ok(builder.take()?.build()?.into()) })
 }
 
--- a/crashtracker-ffi/src/crash_info/builder.rs
+++ b/crashtracker-ffi/src/crash_info/builder.rs
@@ -313,9 +313,13 @@ pub unsafe extern "C" fn ddog_crasht_CrashInfoBuilder_with_stack(
 #[named]
 pub unsafe extern "C" fn ddog_crasht_CrashInfoBuilder_with_thread(
     mut builder: *mut Handle<CrashInfoBuilder>,
-    thread: ThreadData,
+    mut thread: ThreadData,
 ) -> VoidResult {
     wrap_with_void_ffi_result!({
+        if thread.crashed {
+            let stack = (thread.stack.to_inner_mut())?.clone();
+            builder.to_inner_mut()?.with_stack(stack)?;
+        }
         builder.to_inner_mut()?.with_thread(thread.try_into()?)?;
     })
 }
--- a/crashtracker-ffi/src/crash_info/sig_info.rs
+++ b/crashtracker-ffi/src/crash_info/sig_info.rs
@@ -2,22 +2,22 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use datadog_crashtracker::{SiCodes, SignalNames};
-use ddcommon_ffi::{slice::AsBytes, CharSlice};
+use ddcommon_ffi::ToHexStr;
 
 #[repr(C)]
-pub struct SigInfo<'a> {
-    pub addr: CharSlice<'a>,
+pub struct SigInfo {
+    pub addr: usize,
     pub code: libc::c_int,
     pub code_human_readable: SiCodes,
     pub signo: libc::c_int,
     pub signo_human_readable: SignalNames,
 }
 
-impl<'a> TryFrom<SigInfo<'a>> for datadog_crashtracker::SigInfo {
+impl TryFrom<SigInfo> for datadog_crashtracker::SigInfo {
     type Error = anyhow::Error;
-    fn try_from(value: SigInfo<'a>) -> anyhow::Result<Self> {
+    fn try_from(value: SigInfo) -> anyhow::Result<Self> {
         Ok(Self {
-            si_addr: value.addr.try_to_string_option()?,
+            si_addr: Some(value.addr.to_hex_str()),
             si_code: value.code,
             si_code_human_readable: value.code_human_readable,
             si_signo: value.signo,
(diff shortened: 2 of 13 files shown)


crashtracker-ffi/src/crash_info/builder.rs
```diff
@@ -313,9 +313,13 @@ pub unsafe extern "C" fn ddog_crasht_CrashInfoBuilder_with_stack(
 #[named]
 pub unsafe extern "C" fn ddog_crasht_CrashInfoBuilder_with_thread(
     mut builder: *mut Handle<CrashInfoBuilder>,
-    thread: ThreadData,
+    mut thread: ThreadData,
 ) -> VoidResult {
     wrap_with_void_ffi_result!({
+        if thread.crashed {
+            let stack = (thread.stack.to_inner_mut())?.clone();
+            builder.to_inner_mut()?.with_stack(stack)?;
```
gleocadie, line 321: I wonder if `ddog_crasht_CrashInfoBuilder_with_stack` shouldn't be removed ? 🤔

examples/ffi/crashinfo.cpp
```diff
@@ -64,13 +65,29 @@ EXTRACT_RESULT(CrashInfo, DDOG_CRASHT_RESULT_HANDLE_CRASH_INFO_OK_HANDLE_CRASH_I
 EXTRACT_RESULT(StackTrace, DDOG_CRASHT_RESULT_HANDLE_STACK_TRACE_OK_HANDLE_STACK_TRACE)
 EXTRACT_RESULT(StackFrame, DDOG_CRASHT_RESULT_HANDLE_STACK_FRAME_OK_HANDLE_STACK_FRAME)
 
-void add_stacktrace(ddog_crasht_Handle_CrashInfoBuilder *builder) {
-  auto stacktrace = extract_result(ddog_crasht_StackTrace_new(), "failed to make new StackTrace");
+std::optional<std::string> demangle(std::string const& name)
+{
+  // We must keep this call to demangle to check that the symbol is exported.
+  auto result = ddog_crasht_demangle(to_slice_string(name), DDOG_CRASHT_DEMANGLE_OPTIONS_COMPLETE);
+  if (result.tag == DDOG_CRASHT_RESULT_STRING_WRAPPER_OK_STRING_WRAPPER)
```
danielsn, line 72: Not for this pr, but we should probably rename some of these super long enums

examples/ffi/crashinfo.cpp
```diff
@@ -64,13 +65,29 @@ EXTRACT_RESULT(CrashInfo, DDOG_CRASHT_RESULT_HANDLE_CRASH_INFO_OK_HANDLE_CRASH_I
...
+  {
+    // TODO: There is currently no safe way to free the StringWrapper
+    auto stringWrapper = result.ok;
+    return std::string((char*)stringWrapper.message.ptr, stringWrapper.message.len);
+  }
 
+  print_error("Failed to demangled string", result.err);
```
danielsn, line 79:
```suggestion
  print_error("Failed to demangle string", result.err);
```

examples/ffi/crashinfo.cpp
```diff
@@ -64,13 +65,29 @@ EXTRACT_RESULT(CrashInfo, DDOG_CRASHT_RESULT_HANDLE_CRASH_INFO_OK_HANDLE_CRASH_I
...
-  auto stacktrace = extract_result(ddog_crasht_StackTrace_new(), "failed to make new StackTrace");
+std::optional<std::string> demangle(std::string const& name)
+{
+  // We must keep this call to demangle to check that the symbol is exported.
+  auto result = ddog_crasht_demangle(to_slice_string(name), DDOG_CRASHT_DEMANGLE_OPTIONS_COMPLETE);
+  if (result.tag == DDOG_CRASHT_RESULT_STRING_WRAPPER_OK_STRING_WRAPPER)
+  {
+    // TODO: There is currently no safe way to free the StringWrapper
+    auto stringWrapper = result.ok;
+    return std::string((char*)stringWrapper.message.ptr, stringWrapper.message.len);
```
danielsn, line 76: This seems like a problem.  Would a different return type be better here?  Or having a `drop` ffi function

examples/ffi/crashinfo.cpp
```diff
@@ -81,16 +98,18 @@ void add_stacktrace(ddog_crasht_Handle_CrashInfoBuilder *builder) {
...
 
+void add_windows_style_frame(ddog_crasht_Handle_StackTrace* stacktrace) {
   // Windows style frame with normalization
   auto pbd_frame = extract_result(ddog_crasht_StackFrame_new(), "failed to make StackFrame");
-  check_result(ddog_crasht_StackFrame_with_ip(pbd_frame.get(), to_slice_c_char("0xDEADBEEF")),
+  check_result(ddog_crasht_StackFrame_with_ip(pbd_frame.get(), 3735928559), // 0xDEADBEEF
                "failed to add ip");
-  check_result(ddog_crasht_StackFrame_with_module_base_address(pbd_frame.get(),
-                                                               to_slice_c_char("0xABBAABBA")),
+  check_result(ddog_crasht_StackFrame_with_module_base_address(pbd_frame.get(), 2881137594), // 0xABBAABBA"
```
danielsn, line 112: `0x` constants should also work in C++, no? Or was the point to demonstrate that this is clearly an int not a string?

crashtracker-ffi/src/crash_info/sig_info.rs
```diff
@@ -2,22 +2,22 @@
...
 
 use datadog_crashtracker::{SiCodes, SignalNames};
-use ddcommon_ffi::{slice::AsBytes, CharSlice};
+use ddcommon_ffi::ToHexStr;
 
 #[repr(C)]
-pub struct SigInfo<'a> {
-    pub addr: CharSlice<'a>,
+pub struct SigInfo {
+    pub addr: usize,
```
danielsn, line 9: Are we guaranteed to always have one of these?

crashtracker-ffi/src/crash_info/builder.rs
danielsn: If there is only one executing thread, does it make sense to have it twice?

danielsn: But you're the main user of this API, so do what makes sense to you!

examples/ffi/crashinfo.cpp
gleocadie: omg, I forgot this one... good point, will fix it

examples/ffi/crashinfo.cpp
gleocadie: good point, will address this in this PR

examples/ffi/crashinfo.cpp
gleocadie: hum I saw 
`ddog_StringWrapper` and `ddog_crasht_StringWrapper` ....
I will clean it up

examples/ffi/crashinfo.cpp
gleocadie: ✔️

crashtracker-ffi/src/crash_info/stacktrace.rs
```diff
@@ -3,19 +3,26 @@
...
-use ddcommon_ffi::{wrap_with_void_ffi_result, Handle, Result, ToInner, VoidResult};
+use ddcommon_ffi::{wrap_with_void_ffi_result, Error, Handle, ToInner, VoidResult};
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 //                                              FFI API                                           //
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#[allow(dead_code)]
+#[repr(C)]
+pub enum StackTraceNewResult {
```
danielsn, line 14: you can use `ddcommon_ffi::Result` e.g. `ddcommon_ffi::Result<Handle<StackTrace>>`

crashtracker-ffi/src/crash_info/sig_info.rs
gleocadie: good point

crashtracker-ffi/src/crash_info/stacktrace.rs
gleocadie: If I use `ddcommon_ffi::Result<Handle<StackFrame>>` instead of `StackFrameNewResult`
The enum item name is 
`DDOG_CRASHT_STACK_FRAME_NEW_RESULT_OK_HANDLE_STACK_FRAME` (with renaming)
or
`DDOG_CRASHT_RESULT_HANDLE_FRAME_OK_HANDLE_STACK_FRAME`

If I use the new enum `StackFrameNewResult`, I got this
`DDOG_CRASHT_STACK_FRAME_NEW_RESULT_OK` which is shorter

gleocadie: If you know a way to prevent a new enum to have shorter name, I'll be happy :)

crashtracker-ffi/src/crash_info/sig_info.rs
gleocadie: rolled back to the previous code

ddcommon-ffi/src/string.rs
```diff
@@ -72,3 +73,45 @@ pub unsafe extern "C" fn ddog_StringWrapper_message(s: Option<&StringWrapper>) -
...
     }
 }
+
+pub trait ToHexStr {
+    fn to_hex_str(&self) -> String;
+}
+
+impl ToHexStr for usize {
+    fn to_hex_str(&self) -> String {
+        format!("0x{:X}", self)
```
danielsn, line 83: Do we want this to be fixed length? i.e. should `2` be `0x2` or `0x0000000000000002`?

ddcommon-ffi/src/string.rs
```diff
@@ -72,3 +73,45 @@ pub unsafe extern "C" fn ddog_StringWrapper_message(s: Option<&StringWrapper>) -
         Some(s) => CharSlice::from(s.as_ref()),
     }
 }
+
+pub trait ToHexStr {
```
danielsn, line 77: I'm not sure this is the file I'd expect to find this in.

danielsn (commented):
Looks good, just a few more minor comments

ddcommon-ffi/src/string.rs
gleocadie: Where would you put it ?
I thought about this file since it was about string 🤔

ddcommon-ffi/src/string.rs
gleocadie: 🤷 
the specification says that it's a hexadecimal string prefixed with `0x` but nothing about the length.

danielsn (approved):

ddcommon-ffi/src/string.rs
danielsn: Maybe in utils? This file is more about the `CharSlice` FFI implementation of string.

ddcommon-ffi/src/string.rs
danielsn: Agreed that either is standards compliant.  Its just aesthetically, which do we prefer?

paullegranddc (approved):

PR #844, opened 2025-01-27
[PROF-9476] Add intern_all/unintern_all for managed string storage

ivoanjo:
# What does this PR do?

This PR introduces two new operations to the profiling-ffi for use with the managed string storage:

* `intern_all`
* `unintern_all`

These behave similarly to the existing `intern`/`unintern` BUT work for an array of inputs, rather than a single one.

# Motivation

I've been thinking about this optimization since early on in the managed string storage effort, because unlike most other libdatadog APIs we use in profiling, `intern` and `unintern` can get called a lot (e.g. in a loop).

Thus, for these operations any ffi cost will add overhead, whereas for most of our other APIs they're so coarse-grained + called rarely that the ffi cost is negligible.

# Additional Notes

N/A

# How to test the change?

I initially tried to benchmark this from the `profiling-ffi` crate and this was really hard because of `profiling-ffi` having `crate-type = ["staticlib", "cdylib"]` which makes our usual benchmarking setup with criterion fail with really confusing compiler errors about not finding our crate code.

Instead, I decided to include the benchmark in my Ruby branch (which is not yet under PR). [Here's the benchmark](https://github.com/DataDog/dd-trace-rb/commit/a7114845af8469bce518c0a285801e42a16e5149) ([from this branch](https://github.com/DataDog/dd-trace-rb/compare/ivoanjo/prof-9476-managed-string-storage-try2?expand=1)) if you're curious and here's the results from my laptop:

```
$ bundle exec ruby benchmarks/profiler_string_storage_intern.rb 
Current pid is 104113
ruby 3.2.5 (2024-07-26 revision 31d0f1a2e7) [x86_64-linux]
Warming up --------------------------------------
profiler - intern 1000 strings
                         3.106k i/100ms
profiler - intern_all 1000 strings
                         6.728k i/100ms
Calculating -------------------------------------
profiler - intern 1000 strings
                         29.126k (± 2.2%) i/s   (34.33 μs/i) -    291.964k in  10.029188s
profiler - intern_all 1000 strings
                         64.896k (± 1.3%) i/s   (15.41 μs/i) -    652.616k in  10.058208s

Comparison:
profiler - intern_all 1000 strings:    64895.5 i/s
profiler - intern 1000 strings:    29125.5 i/s - 2.23x  slower
```


profiling-ffi/src/string_storage.rs
```diff
@@ -84,6 +82,49 @@ pub unsafe extern "C" fn ddog_prof_ManagedStringStorage_intern(
     .into()
 }
 
+#[no_mangle]
+/// TODO: @ivoanjo Should this take a `*mut ManagedStringStorage` like Profile APIs do?
+pub unsafe extern "C" fn ddog_prof_ManagedStringStorage_intern_all(
```
morrisonlevi, line 87: There are leaks on failure. If we have 20 strings, and we fail on the 11th string, then we leak the first 10 strings because we return an error but do not unintern them.

ivoanjo: Hmm, you're right that any strings that already went in cannot be cleaned one-by-one.

My thinking is: failing to **intern** means something is very wrong -- either the strings have non-utf8 characters, or something even weirder happened.

Thus, in that situation, the only correct way to handle this is to drop the entire string table, or possibly to drop everything and stop profiling, as we've hit a bug.

And if we were to drop the string table, all strings would be freed and no memory would be leaked.

morrisonlevi: >  the strings have non-utf8 characters

I think ideally you wouldn't stop profiling for such a thing. A log, or just replace it lossily with the UTF-8 replacement character � would be better.

> something even weirder happened.

Why does this fail today? Anything aside from UTF-8? I know, ideally I'd just look, but I'm writing this just before bed and don't want to leave you without any feedback at all.

ivoanjo: > I'm writing this just before bed and don't want to leave you without any feedback at all.

Thanks for the help, haha, and definitely don't feel pressured to review code before bed (at least not my code! If it's a cool programming book that seems fine :D )

> Why does this fail today? 

1. Input and output array sizes not matching
2. String storage is null
3. Failed to acquire lock on string storage
4. Internal consistency bug in managed string storage (string -> id and id -> data maps out-of-sync)

All of these I think are "oops something is really wrong here, better not proceed" catastrophic level-events.

> I think ideally you wouldn't stop profiling for such a thing. A log, or just replace it lossily with the UTF-8 replacement character � would be better.

So, yes and no. Specifically -- yea I can see how we could be a bit more lenient here.

But... there is something to be said about this direction. So let me speculate on how it would work. I see two options: either libdatadog reports an error and the caller handles it, or libdatadog just handles it silently.

If libdatadog handles it silently, then it's no longer an error on this API, so we're back to "all other errors are catastrophic-level".

So let's assume libsatadog reports an error. So now this means the caller needs to understand what to do. So now we need to introduce additional error types for the caller to be able to understand which errors are catastrophic and which one is the utf-8 failure. And "which one" is... complicated -- in this API it means which array index, but what about the `add` API which takes locations and also uses `try_to_utf8` in the many strings? Signalling back that the string inside the 3rd location's mapping's build_id is the incorrect one is not impossible, but it seems a lot of code on both sides to recover. (And would we gather all errors? Or report just the first?)

TL;DR I think if we want to be lenient, I think the sane path would be for libdatadog to be lenient transparently, and not expose an error. (Or libdatadog to expose an API for validation, and the caller is responsible for calling it before calling the other APIs? Which for the purposes of this discussion is somewhat equivalent to doing it transparently)

Finally, I'll point out that this "invalid utf-8 stops profiling" is not new -- the `add` API already does the same checks, and returns errors for them, and in particular we haven't had any reports of profilers stopping because of this (and I know for a fact the Ruby profiler has Korean and Japanese customers) so this doesn't seem to be an issue so far and the strings we get from the Ruby VM are in good shape :)

morrisonlevi (approved):
Approved pending CI. I did a review with Ivo over Zoom and we refactored a bit and wrote a test, so theoretically something _could_ fail.

ivoanjo:
Thanks Levi for the feedback && pairing with me on this. I just re-tested the latest version of this PR with the Ruby profiler test suite and it looks like it's in good shape.

I'll continue validating this on the Ruby side so if there's any issue it should come up in our usual integration testing in relenv/benchmarking platform.

PR #847, opened 2025-01-31
[crashtracker] Enable the client to register arbitrary messages

danielsn:
# What does this PR do?

Adds a new capability where the client can temporary add strings which will go on the crash report.  For e.g. a program might register information about its doing when it starts a potentially crashing operation.

# Motivation

Request from the PHP team

# Additional Notes

# How to test the change?

Run `crashtracking.c`, notice how the output now includes

```json
  handle_uintptr_t_result(ddog_crasht_insert_additional_tag(DDOG_CHARSLICE_C("This is a very informative extra bit of info")));
  handle_uintptr_t_result(ddog_crasht_insert_additional_tag(DDOG_CHARSLICE_C("This message will for sure help us debug the crash")));
```


crashtracker/src/collector/atomic_set.rs
```diff
@@ -0,0 +1,440 @@
...
+        if used >= self.set.len() / 2 {
+            // We only fill to half full to get good amortized behaviour
+            self.used.fetch_sub(1, SeqCst);
+            anyhow::bail!("Crashtracker Atomic Set: No space to store {:?}", &value);
+        }
+
+        // Start at a random position.
+        // Since the array is only at most half full, and since we start scanning at random
+        // indicies, every slot should independently have <.5 probability of being occupied.
+        // Long scans become exponentially unlikely, giving amortized constant time insertion.
```
danielsn, line 88: This is linear probing, which should be OK given the high degree of entropy in the rng.  But I could implement a more clever probing strategy if we're worried about getting long linear chains.

danielsn: Changed to do n random scans, and only then a linear poll.

crashtracker/src/receiver/receive_report.rs
```diff
@@ -41,6 +42,16 @@ fn process_line(
     state: StdinState,
 ) -> anyhow::Result<StdinState> {
     let next = match state {
+        StdinState::AdditionalTags if line.starts_with(DD_CRASHTRACK_END_ADDITIONAL_TAGS) => {
+            StdinState::Waiting
+        }
+        StdinState::AdditionalTags => {
+            //DSN TODO fix the input for this
```
danielsn, line 49: remove completed todo

sanchda (approved):

PR #850, opened 2025-02-03
fix: tolerate empty trace payload body

astuyve:
# What does this PR do?
Removes a condition where we return an error to the API client if the trace payload is empty

# Motivation
dotnet sends an empty trace payload as a liveliness check. When we return the error, it causes ddtrace-dotnet to leak file descriptors as per [this issue](https://github.com/DataDog/datadog-lambda-extension/issues/533).
Note that 1024 file descriptors is the maximum allowed for Lambda, and afterwards the function is killed (hence the spikey graph)

Here's a screenshot with this fix and without:
image

# Additional Notes
I believe this change was simply good error handling, and not necessarily an intentional behavior change. We still should investigate the high fd use here for dotnet, and I'm working with that team to identify why we're using so many.

# How to test the change?
My smoke tests should be sufficient, I'm happy to show you the result or help you build your own. We will of course test this fully in self-monitoring before releasing.

### DIFF
--- a/trace-utils/src/trace_utils.rs
+++ b/trace-utils/src/trace_utils.rs
@@ -46,10 +46,6 @@ pub async fn get_traces_from_request_body(
         }
     };
 
-    if traces.is_empty() {
-        anyhow::bail!("No traces deserialized from the request body.")
-    }
-
     Ok((size, traces))
 }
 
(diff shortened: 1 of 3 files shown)


duncanista (approved):

ekump (approved):
LGTM for the tracer_payload and sending of v04 payloads.

ganeshnj (approved):

PR #851, opened 2025-02-03
Svls 6036 respect timeouts

alexgallotta:
this change make the configured flushing time out respected
- Do not keep trying shipping for each batch if a 500 error is met
- send sketches and distributions in parallel

### DIFF
--- a/dogstatsd/src/datadog.rs
+++ b/dogstatsd/src/datadog.rs
@@ -9,6 +9,7 @@ use lazy_static::lazy_static;
 use protobuf::Message;
 use regex::Regex;
 use reqwest;
+use reqwest::{Error, Response};
 use serde::{Serialize, Serializer};
 use serde_json;
 use std::time::Duration;
@@ -119,7 +120,7 @@ impl MetricsIntakeUrlPrefix {
 }
 
 /// Interface for the `DogStatsD` metrics intake API.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct DdApi {
     api_key: String,
     metrics_intake_url_prefix: MetricsIntakeUrlPrefix,
@@ -134,13 +135,10 @@ impl DdApi {
         https_proxy: Option<String>,
         timeout: Duration,
     ) -> Self {
-        let client = match Self::build_client(https_proxy, timeout) {
-            Ok(client) => client,
-            Err(e) => {
-                error!("Unable to parse proxy URL, no proxy will be used. {:?}", e);
-                reqwest::Client::new()
-            }
-        };
+        let client = Self::build_client(https_proxy, timeout).unwrap_or_else(|e| {
+            error!("Unable to parse proxy URL, no proxy will be used. {:?}", e);
+            reqwest::Client::new()
+        });
         DdApi {
             api_key,
             metrics_intake_url_prefix,
@@ -149,40 +147,19 @@ impl DdApi {
     }
 
     /// Ship a serialized series to the API, blocking
-    pub async fn ship_series(&self, series: &Series) {
-        let body = serde_json::to_vec(&series).expect("failed to serialize series");
-        debug!("Sending body: {:?}", &series);
-
+    pub async fn ship_series(&self, series: &Series) -> Result<Response, Error> {
         let url = format!("{}/api/v2/series", &self.metrics_intake_url_prefix);
-        let resp = self
-            .client
-            .post(&url)
-            .header("DD-API-KEY", &self.api_key)
-            .header("Content-Type", "application/json")
-            .body(body)
-            .send()
-            .await;
+        debug!("Sending body: {:?}", &series);
+        let body = serde_json::to_vec(&series).expect("failed to serialize series");
 
-        // match resp {
-        //     Ok(resp) => match resp.status() {
-        //         reqwest::StatusCode::ACCEPTED => {}
-        //         unexpected_status_code => {
-        //             debug!(
-        //                 "{}: Failed to push to API: {:?}",
-        //                 unexpected_status_code,
-        //                 resp.text().await.unwrap_or_default()
-        //             );
-        //         }
-        //     },
-        //     Err(e) => {
-        //         debug!("500: Failed to push to API: {:?}", e);
-        //     }
-        // };
+        self.ship_data(url, body, "application/json").await
     }
 
-    pub async fn ship_distributions(&self, sketches: &SketchPayload) {
+    pub async fn ship_distributions(&self, sketches: &SketchPayload) -> Result<Response, Error> {
         let url = format!("{}/api/beta/sketches", &self.metrics_intake_url_prefix);
         debug!("Sending distributions: {:?}", &sketches);
+        let body = sketches.write_to_bytes().expect("can't write to buffer");
+
         // TODO maybe go to coded output stream if we incrementally
         // add sketch payloads to the buffer
         // something like this, but fix the utf-8 encoding issue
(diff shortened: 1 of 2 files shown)


dogstatsd/src/flusher.rs
```diff
@@ -50,14 +51,46 @@ impl Flusher {
...
+                debug!(
+                    "{}: Failed to push to API: {:?}",
+                    unexpected_status_code,
+                    resp_payload.text().await.unwrap_or_default()
+                );
+                true
+            }
+        },
+        Err(err) => {
+            debug!("500: Failed to push to API: {:?}. Aborting retries", err);
```
astuyve, line 92: not sure if it would help to log an attempt count here

dogstatsd/src/flusher.rs
```diff
@@ -50,14 +51,46 @@ impl Flusher {
...
-            // TODO(astuyve) retry and do not panic
-        }
-        for a_batch in all_distributions {
-            self.dd_api.ship_distributions(&a_batch).await;
+        let dd_api_clone = self.dd_api.clone();
+        tokio::spawn(async move {
+            for a_batch in all_series {
+                let continue_shipping =
+                    should_continue_shipping(dd_api_clone.ship_series(&a_batch).await).await;
+                // TODO(astuyve) retry and do not panic
```
astuyve, line 59: we can probably remove this comment now, ty!

astuyve (approved):

dogstatsd/src/flusher.rs
```diff
@@ -50,14 +51,45 @@ impl Flusher {
...
+            for a_batch in all_series {
+                let continue_shipping =
+                    should_continue_shipping(dd_api_clone.ship_series(&a_batch).await).await;
+                if !continue_shipping {
+                    break;
+                }
+            }
+        });
+        let dd_api_clone = self.dd_api.clone();
+        tokio::spawn(async move {
```
morrisonlevi, line 65: I expected to see a select or something for timeouts. If a single request takes a long time, this still blocks until it resolves (either success or failure), right? Or did I not read this right?

dogstatsd/src/datadog.rs
```diff
@@ -134,13 +135,10 @@ impl DdApi {
...
-            Ok(client) => client,
-            Err(e) => {
-                error!("Unable to parse proxy URL, no proxy will be used. {:?}", e);
-                reqwest::Client::new()
-            }
-        };
+        let client = Self::build_client(https_proxy, timeout).unwrap_or_else(|e| {
+            error!("Unable to parse proxy URL, no proxy will be used. {:?}", e);
+            reqwest::Client::new()
+        });
```
morrisonlevi, line 141: Good simplification.

dogstatsd/src/datadog.rs
```diff
@@ -149,40 +147,19 @@ impl DdApi {
...
-        let resp = self
-            .client
-            .post(&url)
-            .header("DD-API-KEY", &self.api_key)
-            .header("Content-Type", "application/json")
-            .body(body)
-            .send()
-            .await;
+        debug!("Sending body: {:?}", &series);
+        let body = serde_json::to_vec(&series).expect("failed to serialize series");
```
danielsn, line 153:
```suggestion
        let body = serde_json::to_vec(&series).context("failed to serialize series")?;
```

Lets avoid the use of panic as much as possible.

dogstatsd/src/datadog.rs
```diff
@@ -149,40 +147,19 @@ impl DdApi {
...
-            }
-        };
+        self.ship_data(url, body, "application/json").await
     }
 
-    pub async fn ship_distributions(&self, sketches: &SketchPayload) {
+    pub async fn ship_distributions(&self, sketches: &SketchPayload) -> Result<Response, Error> {
         let url = format!("{}/api/beta/sketches", &self.metrics_intake_url_prefix);
         debug!("Sending distributions: {:?}", &sketches);
+        let body = sketches.write_to_bytes().expect("can't write to buffer");
```
danielsn, line 161:
```suggestion
        let body = sketches.write_to_bytes().context("can't write to buffer")?;
```

danielsn (changes requested):

dogstatsd/src/flusher.rs
alexgallotta: The client is already using a timeout, but there is a code path where it doesn't. I will fix that.

Regarding the "blocking issue, the block you are referring to is just the subsequent iteration of the loop, right? If that is the case, I think it makes sense to wait and do it sequentially

dogstatsd/src/datadog.rs
alexgallotta: I just moved the old behavior, but agree, I will handle the error safely

dogstatsd/src/datadog.rs
alexgallotta: I just moved the old behavior, but agree, I will handle the error safely

astuyve (approved):

dogstatsd/src/flusher.rs
```diff
@@ -50,14 +51,54 @@ impl Flusher {
 
         debug!("Flushing {n_series} series and {n_distributions} distributions");
 
-        // TODO: client timeout is for each invocation, so NxM times with N time series batches and
-        // M distro batches
-        for a_batch in all_series {
-            self.dd_api.ship_series(&a_batch).await;
-            // TODO(astuyve) retry and do not panic
+        let dd_api_clone = self.dd_api.clone();
+        tokio::spawn(async move {
```
astuyve, line 55: Do we wait for this to return when we flush? Otherwise we may lose data here

alexgallotta: oh good point, I will add a join

dogstatsd/src/datadog.rs
```diff
@@ -134,11 +136,11 @@ impl DdApi {
...
-            Ok(client) => client,
+        let client = match build_client(https_proxy, timeout) {
+            Ok(client) => Some(client),
             Err(e) => {
-                error!("Unable to parse proxy URL, no proxy will be used. {:?}", e);
-                reqwest::Client::new()
+                error!("Unable to create client {:?}", e);
+                None
             }
         };
```
danielsn, line 145:
```rust
let client = build_client(https_proxy, timeout).inspect_err(|e| error!("Unable to create client {:?}", e)).ok();
```

danielsn (commented):
LGTM rust-wise, with one minor suggestion.  Can't speak to the business logic, so I'll leave someone else to approve that part.

danielsn (approved):

dogstatsd/src/datadog.rs
```diff
@@ -149,40 +149,32 @@ impl DdApi {
...
-            }
-        };
+        if let Ok(safe_body) = serde_json::to_vec(&series) {
+            debug!("Sending body: {:?}", &series);
+            self.ship_data(url, safe_body, "application/json").await
+        } else {
+            Err(ShippingError::Payload(
+                "Failed to serialize series".to_string(),
+            ))
+        }
```
danielsn, line 161:
```suggestion
        let safe_body = serde_json::to_vec(&series).map_err(|_| Err(ShippingError::Payload("Failed to serialize series".to_string())?;
        debug!("Sending body: {:?}", &series);
        self.ship_data(url, safe_body, "application/json").await
```

danielsn (approved):

dogstatsd/src/datadog.rs
danielsn: You could even put the series into the error message if that's useful

dogstatsd/src/datadog.rs
```diff
@@ -192,43 +177,44 @@ impl DdApi {
...
+                .await;
+
+            let elapsed = start.elapsed();
+            debug!("Request to {} took {}ms", url, elapsed.as_millis());
+            resp.map_err(|e| {
+                ShippingError::Destination(e.status(), format!("Cannot reach {}", url))
+            })
+        } else {
+            Err(ShippingError::Destination(None, "No client".to_string()))
         }
```
danielsn, line 206:
```suggestion
        let client = self.client.ok_or_else(|| Err(ShippingError::Destination(None, "No client".to_string())))?;
        let start = std::time::Instant::now();

        let resp = client
            .post(&url)
            .header("DD-API-KEY", &self.api_key)
            .header("Content-Type", content_type)
            .body(body)
            .send()
            .await;

        let elapsed = start.elapsed();
        debug!("Request to {} took {}ms", url, elapsed.as_millis());
        resp.map_err(|e| {
            ShippingError::Destination(e.status(), format!("Cannot reach {}", url))
        })
```

PR #859, opened 2025-02-06
Add function in order to enable telemetry metrics.

hoolioh:
# What does this PR do?

Add a new method to enable telemetry metrics in the trace exporter.

### DIFF
--- a/data-pipeline-ffi/src/trace_exporter.rs
+++ b/data-pipeline-ffi/src/trace_exporter.rs
@@ -4,7 +4,7 @@
 use crate::error::{ExporterError, ExporterErrorCode as ErrorCode};
 use data_pipeline::trace_exporter::agent_response::AgentResponse;
 use data_pipeline::trace_exporter::{
-    TraceExporter, TraceExporterInputFormat, TraceExporterOutputFormat,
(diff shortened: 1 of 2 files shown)


data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -215,6 +223,30 @@ pub unsafe extern "C" fn ddog_trace_exporter_config_set_service(
     }
 }
 
+/// Enables metrics.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_trace_exporter_config_enable_telemetry(
```
ganeshnj, line 228: /nit

I would prefer staying consistent with other APIs and less cognitive load

```suggestion
pub unsafe extern "C" fn ddog_trace_exporter_config_set_telemetry(
```

ganeshnj (approved):

data-pipeline-ffi/src/trace_exporter.rs
ekump: +1 to what @VianneyRuhlmann said. Consistency is important, but I think there is a distinction between "setting" a param and "enabling" functionality.

PR #867, opened 2025-02-11
feat: add API for process discovery

dmehala:
# What does this PR do?

This PR adds a function to store the tracer configuration in an in-memory file.
Resolves [APMAPI-1063]

[APMAPI-1063]: https://datadoghq.atlassian.net/browse/APMAPI-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddcommon-ffi/src/tracer_metadata.rs
dmehala, line 1: Does it work? I don't know. But I tried something 🫠

ddcommon-ffi/src/tracer_metadata.rs
```diff
@@ -0,0 +1,24 @@
...
+            AnonymousFileHandle::Linux(memfd) => Box::as_ptr(memfd) as *mut c_void,
+            #[cfg(not(target_os = "linux"))]
+            AnonymousFileHandle::Other(()) => std::ptr::null() as *mut c_void,
+        }
+    }
+}
+
+#[no_mangle]
+#[must_use]
+pub extern "C" fn ddog_store_tracer_metadata(tracer_metadata: TracerMetadata) -> *mut c_void {
```
bwoebi, line 21: You cannot directly construct TracerMetadata from outside of rust. This needs a struct with `ffi::CharSlice`.

ddcommon-ffi/src/tracer_metadata.rs
```diff
@@ -0,0 +1,24 @@
...
+    let handle = store_tracer_metadata(tracer_metadata).unwrap();
+    return handle.as_raw_pointer();
```
bwoebi, line 23: Also you need to add an an API to drop the metadata memfd. It's a raw pointer... now, the users have to be able to tell rust, here we have a memfd pointer, please cleanup its contents.

ddcommon/Cargo.toml
```diff
@@ -55,6 +58,7 @@ hyper-rustls = { version = "0.27", default-features = false, features = [
     "tls12",
     "aws-lc-rs",
 ] }
+memfd = { version = "0.6" }
```
pawelchcki, line 62: I just checked `memfd` crate and it uses `rustix` as dependency - which means this doesn't bring new libc dependencies unnecessarily 👍 LGTM:

ddcommon/src/tracer_metadata.rs
```diff
@@ -0,0 +1,75 @@
...
+        .allow_sealing(true)
+        .create::<&str>(mfd_name.as_ref())
+        .map_err(|e| format!("Unable to create memfd: {}", e))?;
+
+    let mut buf = Vec::new();
+    data.serialize(&mut Serializer::new(&mut buf).with_struct_map())
+        .unwrap();
+
+    mfd.as_file().write_all(&buf).unwrap();
+    mfd.add_seals(&[
```
pawelchcki, line 62: should inability to seal - be considered failure of the method?

Maybe its only something we should log - but otherwise it could be ignored ?

ddcommon/src/tracer_metadata.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    let mfd = MemfdOptions::default()
+        .close_on_exec(true)
+        .allow_sealing(true)
+        .create::<&str>(mfd_name.as_ref())
+        .map_err(|e| format!("Unable to create memfd: {}", e))?;
+
+    let mut buf = Vec::new();
+    data.serialize(&mut Serializer::new(&mut buf).with_struct_map())
+        .unwrap();
```
pawelchcki, line 59: this unwrap should probably be map to error, that aborts the method.

pawelchcki (changes requested):
Two open questions and one opinion about unnecessary `.unwrap()`

Otherwise looks good to me.

ddcommon/src/tracer_metadata.rs
bwoebi: Yeah, I think sealing is a nice to have, not a hard requirement.

ddcommon/src/tracer_metadata.rs
```diff
@@ -0,0 +1,86 @@
...
+pub enum AnonymousFileHandle {
+    #[cfg(target_os = "linux")]
+    Linux(Box<Memfd>),
+    #[cfg(not(target_os = "linux"))]
+    Other(()),
+}
+
+/// Create the anonymous file storing the tracer metadata.
+#[cfg(target_os = "linux")]
+pub fn store_tracer_metadata(data: TracerMetadata) -> Result<AnonymousFileHandle, String> {
```
paullegranddc, line 53: for dynamic errors like this we usually use `anyhow::Result<AnonymousFileHandle>`
Then you can do this (there are other advantaged other than brevity)
```rust
foo.context("unable t create memfd")?;
// rather than
foo.map_err(|e| format!("unable to create memfd: {}", e))?;
```

ddcommon/src/tracer_metadata.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    let mfd = MemfdOptions::default()
+        .close_on_exec(true)
+        .allow_sealing(true)
+        .create::<&str>(mfd_name.as_ref())
+        .map_err(|e| format!("unable to create memfd: {}", e))?;
+
+    let mut buf = Vec::new();
+    data.serialize(&mut Serializer::new(&mut buf).with_struct_map())
+    .unwrap();
```
paullegranddc, line 69: That's probably not going to fail but 
```suggestion
    data.serialize(&mut Serializer::new(&mut buf).with_struct_map())?;
```

paullegranddc (commented):
LGTM other than a few remarks on coding conventions

ddcommon/src/tracer_metadata.rs
```diff
@@ -0,0 +1,86 @@
+// Copyright 2023-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
```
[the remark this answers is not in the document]
dmehala: addressed in 7729ef21cb445659bb0fcfad997732d06c1c586b

ddcommon/src/tracer_metadata.rs
dmehala: ditto

ddcommon/src/tracer_metadata.rs
```diff
@@ -0,0 +1,91 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub service_env: Option<String>,
+    /// Version of the service being instrumented.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub service_version: Option<String>,
+}
+
+pub enum AnonymousFileHandle {
+    #[cfg(target_os = "linux")]
+    Linux(Box<memfd::Memfd>),
```
bwoebi, line 33: Why is this boxed?

dmehala: addressed in 3604d921baada7ccbc66b6e7b9f866d01d7e4cbf

bwoebi (approved):

PR #870, opened 2025-02-11
[WIP] Fix crash stack trackes on musl

r1viollet:
# What does this PR do?

Check out the new RFC to get a good description.

# Motivation

[The following issue has more context on the possible solutions](https://github.com/rust-lang/backtrace-rs/issues/698#issuecomment-2623916192)

# Additional Notes

The decision to pull in libunwind should be visited

# How to test the change?

Using a crash on alpine with the example provided in crashtracking

### DIFF
--- a/crashtracker/build.rs
+++ b/crashtracker/build.rs
@@ -24,7 +24,6 @@ fn main() {
     if target.contains("musl") {
         // possibly lzma compressed symbol tables. Do we really need it ?
         println!("cargo:rustc-link-lib=static=lzma");
-        println!("cargo:rustc-link-lib=static=unwind-{}", link_lib_arch);
         println!("cargo:rustc-link-lib=static=unwind");
     }
     else {
--- a/crashtracker/build.rs
+++ b/crashtracker/build.rs
@@ -1,0 +1,13 @@
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+
+    // Link the system `libunwind`
+    println!("cargo:rustc-link-lib=unwind");
+
+    // Link `libgcc_s` (needed for `_Unwind` symbols)
+    println!("cargo:rustc-link-lib=dylib=gcc_s");
+    // todo: avoid hard coding these paths
+    // Specify library search paths
+    println!("cargo:rustc-link-search=native=/usr/local/lib");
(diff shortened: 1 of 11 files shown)


crashtracker/build.rs
```diff
@@ -0,0 +1,33 @@
...
+        return;
+    }
+    
+    println!("cargo:rustc-link-search=native=/usr/local/lib");
+    println!("cargo:rustc-link-search=native=/usr/lib");
+    #[cfg(target_arch = "x86_64")]
+    println!("cargo:rustc-link-search=native=/usr/lib/x86_64-linux-gnu");
+    if target.contains("musl") {
+        // possibly lzma compressed symbol tables. Do we really need it ?
+        println!("cargo:rustc-link-lib=static=lzma");
```
r1viollet, line 26: LZMA should not be needed with appropriate libunwind build flags.

crashtracker/src/collector/emitters.rs
```diff
@@ -81,6 +92,66 @@ unsafe fn emit_backtrace_by_frames(
...
+            let mut sp: u64 = 0;
+
+            if unw_get_reg(&mut cursor, UNW_REG_IP, &mut ip) != 0 ||
+               unw_get_reg(&mut cursor, UNW_REG_SP, &mut sp) != 0 {
+                writeln!(w, "{{ \"error\": \"Failed to retrieve registers during stack unwinding\" }}")?;
+                break;
+            }
+
+            if resolve_symbols {
+                let mut func_name = "<unknown>".to_string();
```
danielsn, line 128: does this have to be a string? That risks deadlocking on allocation inside a signal handler

docs/RFCs/0008-custom-unwinder.md
```diff
@@ -0,0 +1,55 @@
+# RFC 0006: Crashtracker unwinding (Version 0.1).
```
danielsn, line 1: 0008

crashtracker/build.rs
r1viollet: This is where you no longer need a standard install of libunwind if you want to optimize for size.
Then should we have the install of libunwind as part of the project ?

PR #871, opened 2025-02-12
Use send_with_retry in the trace exporter

VianneyRuhlmann:
# What does this PR do?

Use `send_with_retry` in trace exporter to remove ownership constraint.

# Motivation

Removing ownership constraint is required to remove incorrect lifetime assumption in the trace exporter.

# Additional Notes

Some minor refactors were needed:
- Replace the Serde error type which was not used anymore to support rmp_serde error
- Move headers used in SendData to ddcommon to be reused in data-pipeline
- Improve the trace exporter errors in case the sending fails (e.g. returning message from hyper error)
- Introduce a common representation for telemetry to handle both SendWithRetryResult and SendDataResult

Commits although not atomic separate the changes into areas affected, so reviewing by commit can help by breaking down the changes.

# How to test the change?

### DIFF
--- a/trace-utils/src/span/v04/span.rs
+++ b/trace-utils/src/span/v04/span.rs
@@ -54,6 +54,7 @@ pub struct Span {
     pub service: BytesString,
     pub name: BytesString,
     pub resource: BytesString,
+    #[serde(skip_serializing_if = "BytesString::is_empty")]
     pub r#type: BytesString,
     pub trace_id: u64,
     pub span_id: u64,
--- a/trace-utils/tests/snapshots/compare_v04_trace_snapshot_test.json
+++ b/trace-utils/tests/snapshots/compare_v04_trace_snapshot_test.json
@@ -17,8 +17,7 @@
       "metrics": {
         "_dd_metric1": 1.0,
         "_dd_metric2": 2.0
-      },
-      "type": ""
+      }
     },
     {
       "service": "test-service",
@@ -34,8 +33,7 @@
         "runtime-id": "test-runtime-id-value",
         "service": "test-service"
       },
-      "metrics": {},
-      "type": ""
+      "metrics": {}
     },
     {
       "service": "test-service",


data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -631,26 +635,38 @@ impl TraceExporter {
...
+                    let result =
+                        send_with_retry(&endpoint, payload, &headers, &strategy, None).await;
+
+                    // Send telemetry for the payload sending
                     if let Some(telemetry) = &self.telemetry {
-                        if let Err(e) = telemetry.send(&send_data_result) {
+                        if let Err(e) = telemetry.send(&SendPayloadTelemetry::from_retry_result(
+                            &result,
+                            payload_len.try_into().unwrap_or(0),
+                            chunks.try_into().unwrap_or(0),
```
hoolioh, line 661: nit: here you can use `as u64`, in this situation is safe to perform an unchecked casting since usize is always 32 or 64 bit depending on the architecture. Besides it'll probably be slightly faster due `try_into` + `unwrap` could result in extra branching but I haven't disassemble the code to check it.

data-pipeline/src/telemetry/mod.rs
```diff
@@ -108,13 +111,83 @@ pub struct TelemetryClient {
...
+    errors_timeout: u64,
+    errors_status_code: u64,
+    bytes_sent: u64,
+    chunks_sent: u64,
+    chunks_dropped: u64,
+    responses_count_per_code: HashMap<u16, u64>,
+}
+
+impl From<&SendDataResult> for SendPayloadTelemetry {
+    fn from(value: &SendDataResult) -> Self {
```
hoolioh, line 129: nit: I think there is no test for this function. Right now it's just assigning and cloning values but if something changes in `SendDataResult` the test might come in useful.

hoolioh (approved):

PR #875, opened 2025-02-12
[crashtracker] Names for SI_CODEs

danielsn:
# What does this PR do?

Adds best effort code to support translating numeric `si_code` into human readable names.

# Motivation

Every different platform has different mappings of numeric codes to meanings.  This saves time on the user to figure out what the code actually was. 

# Additional Notes

This is a best effort code: It currently doesn't handle signals we don't expect to receive. In that case, it soundly says "UNKNOWN".

OSX POSIX support is the worst: https://vorner.github.io/2021/01/03/dark-side-of-posix-apis.html

# How to test the change?

Existing additional tests.


sanchda (approved):
Wonderful!  Just to reaffirm, the numerical relationship between si_code concepts and values is underdefined at the POSIX level, but needs to be a part of the stable API provided by a platform.

Usually (Linux, at least), if there are changes to this type of thing, it tends to be _additive_, so the risk factor is missing an `si_code` if our build-time kernel version is less than the runtime kernel version.  In practice, `si_code` is not frequently changed.

PR #879, opened 2025-02-13
Skip serializing parent-id when 0

VianneyRuhlmann:
# What does this PR do?

Skip serializing parent-id when 0. The agent parses it in the same way wether it is defined or not.

# Motivation

Dotnet snapshots are expecting the field to not be defined

# Additional Notes

# How to test the change?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,12 @@ opt-level = "s"            # optimize for size
 [profile.release.package.datadog-serverless-trace-mini-agent]
 strip = true
 
+[profile.bench]
+codegen-units = 1
+debug = false
+incremental = false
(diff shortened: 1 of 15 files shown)


paullegranddc (approved):
LGTM

trace-utils/src/span_v04/span.rs
```diff
@@ -57,6 +57,7 @@ pub struct Span {
     pub r#type: BytesString,
     pub trace_id: u64,
     pub span_id: u64,
+    #[serde(skip_serializing_if = "is_default")]
```
ganeshnj, line 60: Like we are talking, probably a test case + making it Option<> would be nice.

PR #874, opened 2025-02-12
[WIP] Test benchmarks

bantonsson:

PR #881, opened 2025-02-13
Stabilize benchmarks using opt-level

bantonsson:
# What does this PR do?

Changes the `opt-level` of the `bench` profile to get more stable benchmark results.

# Motivation

The benchmark results are sensitive to strange things like changing the version number of the main project. This has been consistently repeated in a test PR #874. As to the reason why, I can only speculate that optimizing for size can change the placement of code and or fields in unexpected ways. This behavior goes away with the settings in this PR.

# Additional Notes

Nope

# How to test the change?

Was validate in PR #878

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,12 @@ opt-level = "s"            # optimize for size
 [profile.release.package.datadog-serverless-trace-mini-agent]
 strip = true
 
+[profile.bench]
+codegen-units = 1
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
:+1: LGTM, Thanks for improving this!

ganeshnj (approved):

paullegranddc (approved):

PR #884, opened 2025-02-18
Add generic Span

VianneyRuhlmann:
# What does this PR do?

Add a span with a generic string type to be used to support BytesStrings and &str.

# Motivation

Supporting slice based spans is required to support borrowed payload.

# Additional Notes

The functions in `trace-utils/src/span/v04/trace_utils.rs` have been modified to use a generic span to show how to use the new type.
Many files have been changed for renaming but most changes are in `trace-utils/src/span/v04`.
`
# How to test the change?

### DIFF
--- a/trace-utils/src/span/v04/span.rs
+++ b/trace-utils/src/span/v04/span.rs
@@ -53,12 +53,12 @@ impl FromStr for SpanKey {
 
 /// Trait representing the requirements for a type to be used as a Span "string" type.
 /// Note: Borrow<str> is not required by the derived traits, but allows to access HashMap elements
-/// from a static str.
-pub trait SpanValue: Eq + Hash + Borrow<str> {}
-/// Implement the SpanValue trait for any type which satisfies the sub trait.
-impl<T: Eq + Hash + Borrow<str>> SpanValue for T {}
+/// from a static str and check if the string is empty.
+pub trait SpanText: Eq + Hash + Borrow<str> {}
+/// Implement the SpanText trait for any type which satisfies the sub traits.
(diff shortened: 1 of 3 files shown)


trace-utils/src/span/v04/span.rs
```diff
@@ -49,13 +51,40 @@ impl FromStr for SpanKey {
     }
 }
 
+/// Trait representing the requirements for a type to be used as a Span "string" type.
+/// Note: Borrow<str> is not required by the derived traits, but allows to access HashMap elements
+/// from a static str.
+pub trait SpanValue: Eq + Hash + Borrow<str> {}
+/// Implement the SpanValue trait for any type which satisfies the sub trait.
+impl<T: Eq + Hash + Borrow<str>> SpanValue for T {}
```
bantonsson, line 59: Nice. Makes things way more readable.

bantonsson (approved):

trace-utils/src/span/v04/span.rs
```diff
@@ -49,13 +51,40 @@ impl FromStr for SpanKey {
     }
 }
 
+/// Trait representing the requirements for a type to be used as a Span "string" type.
+/// Note: Borrow<str> is not required by the derived traits, but allows to access HashMap elements
+/// from a static str.
+pub trait SpanValue: Eq + Hash + Borrow<str> {}
```
ganeshnj, line 57: `SpanValue` is too broad as name, given this is explicitly targeting the strings, I would prefer `SpanString` to narrow down the scope.

ganeshnj (approved):
Nice work 🎉 

I played a bit with your PR to see if associated types make sense here, but after playing I bit, I would prefer the current approach.

PR #890, opened 2025-02-19
ekump/APMSP-1805 lazily init cadence client

ekump:
# What does this PR do?

Wait until send is called on dogstatsd client to initialize the cadence client. 

# Motivation

After [this PR](https://github.com/DataDog/libdatadog/pull/638) PHP customers in serverless environments started to experience [an issue](https://github.com/DataDog/dd-trace-php/issues/2991) running out of file descriptors / threads. It is believed to be related to the creation of a new cadence client when initializing a new dogstatsd_client. This PR should re-establish previous behavior where the cadence client is lazily initialized when sending stats.

# Additional Notes

We couldn't simply revert to using the previous code in the sidecar that lazily initialized the cadence client because the dogstatsd client is used in the Data Pipeline Trace Exporter inside async code and you can't just make the `send` method `&mut self` to allow us to set the client. So we have to use interior mutability via `Arc` and `Mutex`. 

# How to test the change?

Unit tests added and existing ones continue to work. PHP team manually verified this resolves the thread count issue.


dogstatsd-client/src/lib.rs
```diff
@@ -333,12 +351,56 @@ mod test {
...
         //     std::mem::variant_count::<DogStatsDActionOwned>(),
         //     std::mem::variant_count::<DogStatsDAction<String, Vec<&Tag>>>(),
         //     "DogStatsDActionOwned and DogStatsDAction should have the same number of variants,
         // did you forget to update one?", );
     }
+    use std::sync::Arc;
+    use std::thread;
+
+    #[tokio::test]
+    async fn test_thread_safety() {
```
ekump, line 365: I'm not sure how useful this test is? It is just checking that we don't set a client until after we send. I couldn't think of a better way to unit test this.

dogstatsd-client/src/lib.rs
```diff
@@ -67,62 +68,57 @@ pub enum DogStatsDAction<'a, T: AsRef<str>, V: IntoIterator<Item = &'a Tag>> {
...
+        endpoint: Some(endpoint),
+        ..Default::default()
     })
 }
 
 impl Client {
-    /// Set the destination for dogstatsd metrics, if an API Key is provided the client is disabled
-    /// as dogstatsd is not allowed in agentless mode. Returns an error if the provided endpoint
-    /// is invalid.
-    pub fn set_endpoint(&mut self, endpoint: Endpoint) -> anyhow::Result<()> {
```
ekump, line 88: This function wasn't used anywhere else in libdatadog. I can leave it in as it is technically a breaking API change if we think the crate is being used outside of libdatadog.

dogstatsd-client/src/lib.rs
```diff
@@ -67,62 +68,57 @@ pub enum DogStatsDAction<'a, T: AsRef<str>, V: IntoIterator<Item = &'a Tag>> {
     Set(T, i64, V),
 }
 
-/// A dogstatsd-client that flushes stats to a given endpoint. Use `new_flusher` to build one.
-#[derive(Debug)]
+/// A dogstatsd-client that flushes stats to a given endpoint.
+#[derive(Debug, Default)]
 pub struct Client {
-    client: StatsdClient,
+    client: Arc<Mutex<Option<StatsdClient>>>,
```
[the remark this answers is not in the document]
bantonsson: So thinking a bit more about this, I see that the `DogstatsdClient` isn't `Clone` and what I think that you really want is `client: Mutex<Option<Arc<StatsdClient>>>,`.

dogstatsd-client/src/lib.rs
```diff
@@ -133,31 +129,53 @@ impl Client {
...
+            .map_err(|e| anyhow!("Failed to acquire dogstatsd client lock: {}", e))?;
+
+        if client_guard.is_none() {
+            if let Some(endpoint) = &self.endpoint {
+                *client_guard = Some(create_client(endpoint)?);
+            }
+        }
+
+        Ok(client_guard)
+    }
```
bantonsson, line 178: I think that only protecting the initialization would be best. Something like this.

```suggestion
    fn get_or_init_client<'a>(&self) -> anyhow::Result<Option<Arc<StatsdClient>>> {
        if let Some(endpoint) = &self.endpoint {
            let mut guard = self.client.lock().map_err(|e| anyhow!(e.to_string()))?;
            if let Some(client) = guard.as_ref() {
                return Ok(Some(client.clone()));
            } else {
                let client = Arc::new(create_client(endpoint)?);
                *guard = Some(client.clone());
                return Ok(Some(client));
            }
        }

        Ok(None)
    }
```

dogstatsd-client/src/lib.rs
```diff
@@ -133,31 +129,55 @@ impl Client {
...
-                DogStatsDAction::Gauge(metric, value, tags) => {
-                    do_send(client.gauge_with_tags(metric.as_ref(), value), tags)
-                }
-                DogStatsDAction::Histogram(metric, value, tags) => {
-                    do_send(client.histogram_with_tags(metric.as_ref(), value), tags)
-                }
-                DogStatsDAction::Set(metric, value, tags) => {
-                    do_send(client.set_with_tags(metric.as_ref(), value), tags)
+        let client_opt = match self.get_or_init_client() {
+            Ok(guard) => guard,
```
bantonsson, line 133: Nitpick: variable name `guard` is a bit confusing since it's the client option now.

bantonsson (approved):
👍🏼

dogstatsd-client/src/lib.rs
ekump: ah yes, forgot to update this variable. I'll make a note to update it in another PR.

PR #892, opened 2025-02-20
Crashtracking for Windows

kevingosse:
# What does this PR do?

Add Windows support for crashtracking.

# Motivation

PHP (and probably other languages) have some unique crashes on Windows, so the Linux support isn't enough.

# Additional Notes

Our implementation of crashtracking for Windows relies on WER (Windows Error Reporting).

The way it works is: the process calls `WerRegisterRuntimeExceptionModule` to register the DLL that contains the crash handling code. The DLL must expose three methods:

 - `OutOfProcessExceptionEventCallback`: gives a chance to analyze the crashing process. Optionally, the crash handler can set `ownershipClaimed` to true to "claim" the crash. If it does, then the crash handler provide some metadata that will be attached to the crash report in the Windows event log. Since we are only implementing crash tracking for telemetry, we never claim the crash.
 - `OutOfProcessExceptionEventSignatureCallback`: this method is only called if the crash handler claimed the crash. It's used to provide metadata to help triaging the crash.
 - `OutOfProcessExceptionEventDebuggerLaunchCallback`: this method is only called if the crash handler claimed the crash. It's used to provide the path to a custom debugger that will be used to debug the crashing process.

In addition, the path to the crash handler must be added to the registry, in `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\Windows Error Reporting\RuntimeExceptionHelperModules` (requires administrator permissions). Since Windows 10 20H1, it's possible to use `HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\Windows Error Reporting\RuntimeExceptionHelperModules` instead (doesn't require additional permissions in most environments). The only limitation of using HKCU instead of HKLM is that the crash handler can't claim the crash (which we don't really care about).

Ideally, we should add the key to `HKEY_LOCAL_MACHINE` during installation. To handle scenarios where it's not possible, the `init_crashtracking_windows` method looks for the key, and creates one in `HKEY_CURRENT_USER` if it can't be found.

When the process crashes, Windows suspends it then spawns `WerFault.exe`. WerFault loads the registered crash handlers in order, and calls `OutOfProcessExceptionEventCallback` for each of them until the crash is claimed (or none are left). The crashing process is resumed when WerFault terminates (gracefully or abruptly). In theory, crashing in WerFault has no visible consequence on the system, so this is a very safe way to inspect the crashes.

To inspect the crashing process during `OutOfProcessExceptionEventCallback`, we use the APIs exposed by `dbghelp.dll`. `dbghelp.dll` is present on all Windows systems, however the version of the library can change. Because of this, we must be careful about what APIs we call, and favor the oldest ones. Currently we limit ourselves to `SymInitializeW` (it's unclear what it does exactly, but it's required to be able to walk the callstacks), and `StackWalkEx` (walks the stack of the given thread). The .NET tracer already relies on those methods for its crashtracking implementation, so we know they're safe to call on the supported versions of Windows (2012 R2 and ulterior).

It's very hard to monitor what happens inside of the crash handler loaded in WerFault.exe, because it's a background process with no window or console. To provide a bit of visibility, we call the Win32 API `OutputDebugString`, which logs message that can be listened to using a special system-wide debugger (for instance, [DebugView](https://learn.microsoft.com/en-us/sysinternals/downloads/debugview)).

# How to test the change?

`test_crash` in `collector_windows_tests` launches a test application and checks that a crash report is generated. The same test application can be used to test manually if needed.


crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+        prefix: WER_CONTEXT_PREFIX,
+        ptr: static_slice.as_ptr(),
+        len,
+        suffix: WER_CONTEXT_SUFFIX,
+    };
+
+    Ok(wer_context)
+}
+
+// https://github.com/microsoft/win32metadata/issues/1044
```
danielsn, line 380: Kinda crazy how long that issue has been open

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+    }
+    #[cfg(target_arch = "x86")]
+    {
+        context.ctx.ContextFlags = CONTEXT_FULL_X86;
+    }
+
+    GetThreadContext(thread_handle, &mut context.ctx)?;
+
+    let mut native_frame = STACKFRAME_EX::default();
+    let machine_type: u32;
```
danielsn, line 408: Would it be cleaner to initialize this directly in the `cfg` block, and then we wouldn't need the mut / have an uninitilized variable hanging around?

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+            if let Some(pdb_info) = &module.pdb_info {
+                frame.build_id = Some(format!("{:x}{:x}", pdb_info.signature, pdb_info.age));
+                frame.build_id_type = Some(datadog_crashtracker::BuildIdType::PDB);
+                frame.file_type = Some(datadog_crashtracker::FileType::PE);
+            }
+        }
+
+        stacktrace
+            .push_frame(frame, true)
+            .expect("Failed to add frame");
```
danielsn, line 478:
```suggestion
            .context("Failed to add frame")?;
```

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+        let module = modules.iter().find(|module| {
+            module.base_address <= native_frame.AddrPC.Offset
+                && native_frame.AddrPC.Offset < module.base_address + module.size
+        });
+
+        if let Some(module) = module {
+            frame.module_base_address = Some(format!("{:x}", module.base_address));
+            frame.symbol_address = Some(format!(
+                "{:x}",
+                native_frame.AddrPC.Offset - module.base_address
```
danielsn, line 465: Is this guaranteed not to underflow?

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+
+    let result = ReadProcessMemory(
+        process_handle,
+        address as *const _,
+        value.as_mut_ptr() as *mut _,
+        size,
+        Some(&mut bytes_read),
+    );
+
+    if result.is_err() || bytes_read != size {
```
danielsn, line 561: anyhow::ensure

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+        return Err(anyhow!("Invalid machine type"));
+    }
+
+    let debug_data_dir: IMAGE_DATA_DIRECTORY = if is_pe32 {
+        let nt_headers32: IMAGE_NT_HEADERS32 = read_memory(process_handle, nt_headers_address)?;
+        nt_headers32.OptionalHeader.DataDirectory[IMAGE_DIRECTORY_ENTRY_DEBUG.0 as usize]
+    } else {
+        let nt_headers64: IMAGE_NT_HEADERS64 = read_memory(process_handle, nt_headers_address)?;
+        nt_headers64.OptionalHeader.DataDirectory[IMAGE_DIRECTORY_ENTRY_DEBUG.0 as usize]
+    };
```
danielsn, line 657: Is there documentation for why this is the case?

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+        /* 'RSDS' */
+        {
+            return Ok(PdbInfo {
+                age: cv_info.age,
+                signature: cv_info.guid,
+            });
+        }
+    }
+
+    Err(anyhow!("No CodeView entry found"))
```
danielsn, line 688: anyhow::bail!

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+use ddcommon_ffi::CharSlice;
+use serde::{Deserialize, Serialize};
+use std::ffi::{c_void, OsString};
+use std::fmt;
+use std::mem::MaybeUninit;
+use std::os::windows::ffi::OsStringExt;
+use std::ptr::{addr_of, read_unaligned};
+use windows::core::{w, HRESULT, PCWSTR};
+use windows::Win32::Foundation::{BOOL, ERROR_SUCCESS, E_FAIL, HANDLE, HMODULE, S_OK, TRUE};
+#[cfg(target_arch = "x86_64")]
```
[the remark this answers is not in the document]
kevingosse: In the future maybe, but for now ARM on Windows Server is not a thing (only the desktop version of Windows runs on ARM)

spawn_worker/src/win32.rs
```diff
@@ -85,6 +85,35 @@ fn write_trampoline(process_name: &Option<String>) -> io::Result<(PathBuf, File)
     Ok((path, file))
 }
 
+pub fn write_crashtracking_trampoline(process_name: &String) -> io::Result<(PathBuf, File)> {
+    let mut path = env::temp_dir().join(process_name);
+    path.set_extension("dll");
+
+    // Attempt to move it just in case it already exists
```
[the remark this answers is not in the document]
kevingosse: I reused the logic from the trampoline (since the need is the same): https://github.com/DataDog/libdatadog/blob/main/spawn_worker/src/win32.rs#L48

The filename is made of the user SID and the version number, so if multiple instances of PHP are running they will share the same file. I think this is a good thing for crashtracking because we need to add the path to the registry, and I'm afraid we would add a lot of garbage if the path was random.

spawn_worker/build.rs
```diff
@@ -38,4 +38,17 @@ fn main() {
             .try_compile_shared_lib("ld_preload_trampoline.shared_lib")
             .unwrap();
     }
+
+    if cfg!(target_os = "windows") {
+        cc_utils::ImprovedBuild::new()
+            .file("src/crashtracking_trampoline.cpp") // Path to your C++ file
+            .warnings(true)
+            .warnings_into_errors(true)
+            .flag("/std:c++17") // Set the C++ standard (adjust as needed)
```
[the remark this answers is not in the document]
kevingosse: Probably but the size is still reasonable. The only reason I used C++ is because it has regex support in the stdlib. The DLL size is 160 KB, I believe it's acceptable (it was ~60 KB in C with manual parsing).

spawn_worker/src/crashtracking_trampoline.cpp
```diff
@@ -0,0 +1,140 @@
...
+    {
+        OutputDebugStringW(L"Datadog Crashtracking - OutOfProcessExceptionEventCallback");
+
+        auto process = pExceptionInformation->hProcess;
+
+        DWORD cbNeeded;
+
+        if (!EnumProcessModules(process, nullptr, 0, &cbNeeded))
+        {
+            OutputDebugStringW(L"Failed to enumerate process modules (1st)");
```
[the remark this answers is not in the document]
kevingosse: We call `EnumProcessModules` twice (first to get the number of modules, then to populate them). It's simply to know if we failed in the first or the second call.

spawn_worker/src/crashtracking_trampoline.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+        auto modules = std::vector<HMODULE>(cbNeeded / sizeof(HMODULE));
+
+        if (!EnumProcessModules(process, modules.data(), cbNeeded, &cbNeeded))
+        {
+            OutputDebugStringW(L"Failed to enumerate process modules (2nd)");
+            return E_FAIL;
+        }
+
+        std::wregex php_pattern(LR"(php\d+(ts|nts)\.dll$)", std::regex_constants::icase);
```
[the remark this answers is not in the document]
kevingosse: I made the regex as restrictive as possible to make this unlikely to happen. Still, not impossible.

In the false positive case, we load an arbitrary DLL that can execute code. I don't think that's a concern because that DLL was already loaded in the crashing process, and the WER handler has the same permissions (no privilege escalation). 
Also, loading php_ddtrace.dll will fail to load because we didn't load the right php DLL, so crashtracking will silently fail (no crash report sent).

I don't think a false negative is possible, but the consequence would be the same (silently fails, no crash report).

sidecar/src/windows.rs
```diff
@@ -128,6 +132,29 @@ pub fn setup_daemon_process(listener: OwnedHandle, spawn_cfg: &mut SpawnWorker)
...
+    let result = write_crashtracking_trampoline(&format!(
+        "datadog-crashtracking-{}",
+        primary_sidecar_identifier()
+    ));
+
+    if result.is_ok() {
+        let path = result.unwrap().0.into_os_string().into_string().unwrap();
+
+        unsafe {
+            return ddog_crasht_init_windows(CharSlice::from(path.as_str()), endpoint, metadata);
```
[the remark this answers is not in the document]
kevingosse: `ddog_crasht_init_windows` makes a copy of `path`, so it only needs to stay alive as long as the function. I don't know if there is a way to make this explicit.

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+/// # Atomicity
+///   This function is not atomic. A crash during its execution may lead to
+///   unexpected crash-handling behaviour.
+pub unsafe extern "C" fn ddog_crasht_init_windows(
+    module: CharSlice,
+    endpoint: Option<&Endpoint>,
+    metadata: Metadata,
+) -> bool {
+    let result: Result<(), _> = (|| {
+        let endpoint = endpoint.cloned();
```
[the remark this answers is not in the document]
kevingosse: `TelemetryCrashUploader` supports a null endpoint, in which case it populates the info from the environment variables.

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+            format!("ddog_crasht_init_windows failed: {}", result.err().unwrap()).as_str(),
+        );
+        return false;
+    }
+
+    true
+}
+
+unsafe fn create_registry_key(path: &str) -> Result<()> {
+    // First, check if there is already a key named "path" in SOFTWARE\Microsoft\Windows\Windows
```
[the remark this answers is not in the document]
kevingosse: The value isn't used by WER, only the name of the key (yeah it's weird)

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+    if set_value_result != ERROR_SUCCESS {
+        return Err(anyhow!("Failed to set registry value"));
+    }
+
+    let _ = RegCloseKey(hkey);
+
+    Ok(())
+}
+
+unsafe fn output_debug_string(message: &str) {
```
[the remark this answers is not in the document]
kevingosse: There's a pointer:

```
PCWSTR::from_raw(wstr.as_ptr())
```

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+    let mut wstr: Vec<u16> = message.encode_utf16().collect();
+    wstr.push(0); // Ensure null termination
+    OutputDebugStringW(PCWSTR::from_raw(wstr.as_ptr()));
+}
+
+fn set_error_context(message: &str) {
+    let bytes = message.as_bytes();
+    let boxed_slice = bytes.to_vec().into_boxed_slice();
+    let static_slice = Box::leak(boxed_slice);
```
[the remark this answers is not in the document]
kevingosse: It needs to stay alive for the whole duration of the application. I suppose I could add a `WerContextBytes` static field instead of calling `Box::leak` if you think it's better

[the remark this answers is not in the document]
kevingosse: Actually it looks like this code is illegal:

```
static mut WerContextBytes: Box<[u8]> = Box::new([]);
```

So I'm not sure how to do that

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+#[repr(C)]
+pub struct WerContext {
+    prefix: u64,
+    ptr: *const u8,
+    len: usize,
+    suffix: u64,
+}
+
+#[no_mangle]
+#[cfg(target_os = "windows")]
```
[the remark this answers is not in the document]
kevingosse: Good catch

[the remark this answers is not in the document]
kevingosse: It is needed because I use `target_os` as condition in cbindgen to add `#if defined _WIN32` in the .h. It doesn't seem to work at the module level.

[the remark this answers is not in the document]
kevingosse: I put this in the cbindgen:
```
"target_os = windows" = "_WIN32"
```
but it only works if I add `#[cfg(target_os = "windows")]` on each method, it doesn't seem to work at the module level.

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+    // Copy the raw bytes from the Vec<u8> into the MaybeUninit<WerContext>
+    let ptr = wer_context.as_mut_ptr();
+    std::ptr::copy_nonoverlapping(buffer.as_ptr(), ptr as *mut u8, buffer.len());
+
+    // Validate prefix and suffix
+    let raw_ptr = wer_context.as_ptr();
+
+    // We can't call assume_init yet because ptr hasn't been set,
+    // and apparently (*raw_ptr).prefix before calling assume_init is undefined behavior
+    let prefix = read_unaligned(addr_of!((*raw_ptr).prefix));
```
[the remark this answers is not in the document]
kevingosse: It's fairly easy to test (create a variable, call `read_wer_context` with a pointer to it, check the result). I haven't checked where to put the tests yet.

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+        machine_type = IMAGE_FILE_MACHINE_AMD64.0 as u32;
+        native_frame.AddrPC.Offset = context.ctx.Rip;
+        native_frame.AddrPC.Mode = AddrModeFlat;
+        native_frame.AddrStack.Offset = context.ctx.Rsp;
+        native_frame.AddrStack.Mode = AddrModeFlat;
+        native_frame.AddrFrame.Offset = context.ctx.Rbp;
+        native_frame.AddrFrame.Mode = AddrModeFlat;
+    }
+
+    #[cfg(target_arch = "x86")]
```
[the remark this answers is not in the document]
kevingosse: On Windows at the moment we only support x86 and x64. I don't know what kind of check you have in mind.

crashtracker-ffi/src/collector_windows/api.rs
kevingosse: Yeah I did that originally but it doesn't work. If I declare the variable in the block then its scope is limited to the block.

crashtracker-ffi/src/collector_windows/api.rs
kevingosse: We could read garbage in the registers so nothing is guaranteed. Would Rust panic in case of underflow?

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+    let mut thread_entry = THREADENTRY32 {
+        dwSize: size_of::<THREADENTRY32>() as u32,
+        ..Default::default()
+    };
+
+    if Thread32First(snapshot, &mut thread_entry).is_ok() {
+        loop {
+            if thread_entry.th32OwnerProcessID == pid {
+                thread_ids.push(thread_entry.th32ThreadID);
+            }
```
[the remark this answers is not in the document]
kevingosse: Yes and yes. But as crazy as it sounds, this is the normal way.

In "recent" versions of windows (since 2012 R2) there is an alternative way that doesn't require to enumerate all threads (using something called "process snapshotting", which can be thought as Windows' sane version of `vfork`). However we would need to confirm that it works correctly in the context of WER, so that requires additional research. Our implementation of crashtracking in .NET uses `CreateToolhelp32Snapshot`, so I'd rather rely on this battle-tested solution for now.

crashtracker-ffi/src/collector_windows/api.rs
kevingosse: I don't think there is a good official documentation about the PE format. It's mostly the definitions in the official headers (https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_nt_headers64) and then a bunch of third-party articles: https://learn.microsoft.com/en-us/archive/msdn-magazine/2002/february/inside-windows-win32-portable-executable-file-format-in-detail  https://wiki.osdev.org/PE

For the Rust implementation, I simply converted the C++ code we wrote for crashtracking in .net: https://github.com/DataDog/dd-trace-dotnet/blob/master/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/CrashReportingWindows.cpp#L271
which has proper testing: https://github.com/DataDog/dd-trace-dotnet/blob/master/profiler/test/Datadog.Profiler.Native.Tests/CrashReportingTest.cpp

We probably want to add similar testing in the libdatadog repository.

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+    pub data2: u16,
+    pub data3: u16,
+    pub data4: [u8; 8],
+}
+
+impl fmt::LowerHex for Guid {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "{:08x}{:04x}{:04x}{:02x}{:02x}{:02x}{:02x}{:02x}{:02x}{:02x}{:02x}",
```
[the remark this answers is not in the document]
kevingosse: This is standard guid formatting on Windows, but without the dashes: https://devblogs.microsoft.com/oldnewthing/20220928-00/?p=107221

crashtracker-ffi/src/collector_windows/api.rs
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

spawn_worker/build.rs
```diff
@@ -38,4 +38,17 @@ fn main() {
             .try_compile_shared_lib("ld_preload_trampoline.shared_lib")
             .unwrap();
     }
+
+    if cfg!(target_os = "windows") {
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

sidecar/src/windows.rs
```diff
@@ -128,6 +132,29 @@ pub fn setup_daemon_process(listener: OwnedHandle, spawn_cfg: &mut SpawnWorker)
...
+) -> bool {
+    // Ensure unique process names - we spawn one sidecar per console session id (see
+    // setup/windows.rs for the reasoning)
+    let result = write_crashtracking_trampoline(&format!(
+        "datadog-crashtracking-{}",
+        primary_sidecar_identifier()
+    ));
+
+    if result.is_ok() {
+        let path = result.unwrap().0.into_os_string().into_string().unwrap();
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+        let path = module.try_to_string()?;
+        create_registry_key(&path)?;
+
+        let mut wpath: Vec<u16> = path.encode_utf16().collect();
+        wpath.push(0); // Ensure null termination
+        WerRegisterRuntimeExceptionModule(
+            PCWSTR::from_raw(wpath.as_ptr()),
+            addr_of!(WERCONTEXT) as *const c_void,
+        )?;
+        Ok::<(), anyhow::Error>(())
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+    })();
+
+    if result.is_err() {
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+        REG_OPTION_NON_VOLATILE,
+        KEY_WRITE,
+        None,
+        &mut hkey,
+        None,
+    );
+
+    if create_result != ERROR_SUCCESS {
+        return Err(anyhow!("Failed to create registry key"));
+    }
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+pub unsafe extern "C" fn ddog_crasht_event_signature_callback(
+    _context: *const c_void,
+    _exception_information: *const WER_RUNTIME_EXCEPTION_INFORMATION,
+    _index: i32,
+    _name: *mut u16,
+    _name_size: *mut u32,
+    _value: *mut u16,
+    _value_size: *mut u32,
+) -> HRESULT {
+    output_debug_string("ddog_crasht_event_signature_callback");
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+) -> HRESULT {
+    output_debug_string("ddog_crasht_exception_event_callback");
+
+    let result: Result<(), _> = (|| {
+        let exception_information = *exception_information;
+        SymInitializeW(exception_information.hProcess, PCWSTR::null(), true)?;
+
+        let pid = GetProcessId(exception_information.hProcess);
+        let crash_tid = GetThreadId(exception_information.hThread);
+        let threads = list_threads(pid).expect("Failed to list threads");
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+        let modules = list_modules(exception_information.hProcess).unwrap_or_else(|_| Vec::new());
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+
+            if thread == crash_tid {
+                builder
+                    .with_stack(stack.clone())
+                    .context("Failed to add crashed thread info")?;
+            }
+
+            let thread_data = ThreadData {
+                crashed: thread == crash_tid,
+                name: format!("{}", thread),
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+        let crash_info = builder.build().context("Failed to build crash info")?;
+
+        crash_info
+            .upload_to_endpoint(&error_context.endpoint)
+            .context("Failed to upload crash info")?;
+
+        Ok::<(), anyhow::Error>(())
+    })();
+
+    if result.is_err() {
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+    output_debug_string("ddog_crasht_exception_event_callback succeeded");
+    S_OK
+}
+
+pub unsafe fn read_wer_context(process_handle: HANDLE, base_address: usize) -> Result<WerContext> {
+    let buffer = read_memory_raw(process_handle, base_address as u64, size_of::<WerContext>())?;
+
+    if buffer.len() != size_of::<WerContext>() {
+        return Err(anyhow!("Failed to read the full WerContext, wrong size"));
+    }
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,737 @@
...
+            None,
+            None,
+            None,
+            None,
+            SYM_STKWALK_DEFAULT,
+        );
+
+        if result != TRUE {
+            break;
+        }
```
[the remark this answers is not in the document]
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/ea8e7c31f38e2b3c86a21406ac52785ae29e7797

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,836 @@
...
+}
+
+static mut WERCONTEXT: WerContext = WerContext {
+    prefix: WER_CONTEXT_PREFIX,
+    ptr: std::ptr::null(),
+    len: 0,
+    suffix: WER_CONTEXT_SUFFIX,
+};
+
+pub const WER_CONTEXT_PREFIX: u64 = 0xBABA_BABA_BABA_BABA;
```
danielsn, line 203: Is there meaning to this bitpattern? Or can it be any, and this is just easy to see in a debugger?

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,836 @@
...
+
+    anyhow::ensure!(
+        result.is_ok() && bytes_read == size,
+        "Failed to read memory"
+    );
+
+    Ok(value.assume_init())
+}
+
+unsafe fn read_memory_raw(process_handle: HANDLE, address: u64, size: usize) -> Result<Vec<u8>> {
```
danielsn, line 572:
```rust
// #Safety
``` 
comments on unsafe function

sidecar/src/windows.rs
```diff
@@ -128,6 +132,38 @@ pub fn setup_daemon_process(listener: OwnedHandle, spawn_cfg: &mut SpawnWorker)
...
+) -> bool {
+    // Ensure unique process names - we spawn one sidecar per console session id (see
+    // setup/windows.rs for the reasoning)
+    match write_crashtracking_trampoline(&format!(
+        "datadog-crashtracking-{}",
+        primary_sidecar_identifier()
+    )) {
+        Ok((path, _)) => {
+            if let Ok(path_str) = path.into_os_string().into_string() {
+                unsafe {
```
danielsn, line 148: safety comment

crashtracker-ffi/tests/test_app_lib/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+use windows::Win32::Foundation::HMODULE;
+use windows::Win32::System::LibraryLoader::{
+    GetModuleHandleExW, GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS,
+};
+use windows::Win32::System::ProcessStatus::GetModuleFileNameExW;
+use windows::Win32::System::Threading::GetCurrentProcess;
+
+/// # Safety
+///
+/// This function is unsafe because it relies on Win32 APIs.
```
danielsn, line 23: What are the required preconditions to run this function?

crashtracker-ffi/src/collector_windows/api.rs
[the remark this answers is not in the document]
kevingosse: Using `HSTRING` does indeed make the conversion easier, but it still requires `unsafe`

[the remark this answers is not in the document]
kevingosse: 👍 
In C#, putting `unsafe` in the function is just a way of wrapping the entire function in an unsafe block, I didn't realize there was an added meaning in Rust.

crashtracker-ffi/src/collector_windows/api.rs
kevingosse: No meaning, I just needed a non-zero pattern. In the future it could be repurposed for versioning if needed. Gonna add a comment.

Cargo.toml
```diff
@@ -7,6 +7,8 @@ members = [
   "builder",
   "crashtracker",
   "crashtracker-ffi",
+  "crashtracker-ffi/tests/test_app",
+  "crashtracker-ffi/tests/test_app_lib",
```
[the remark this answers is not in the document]
kevingosse: I need a test app to crash. The lib is because I need a DLL for WER. Maybe it's possible to reference datadog-crashtracker-ffi as a DLL but I didn't find how (when I reference the crate it gets statically linked)

crashtracker-ffi/src/collector_windows/api.rs
kevingosse: Done

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,836 @@
+// Copyright 2023-Present Datadog, Inc. https://www.datadoghq.com/
```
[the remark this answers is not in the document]
kevingosse: Done

crashtracker-ffi/Cargo.toml
```diff
@@ -10,12 +10,19 @@ license.workspace = true
 
 [lib]
 bench = false
+crate-type = ["lib", "cdylib"]
+
+[[bin]]
+name = "test_app"
+path = "tests/test_app/src/main.rs"
```
danielsn, line 17: What happens with this on Linux?

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,119 @@
...
+    if let Err(e) = result {
+        output_debug_string(format!("ddog_crasht_init_windows failed: {}", e).as_str());
+        return false;
+    }
+
+    true
+}
+
+fn output_debug_string(message: &str) {
+    unsafe { OutputDebugStringW(&HSTRING::from(message)) };
```
danielsn, line 50: Safety comment

crashtracker-ffi/src/collector_windows/api.rs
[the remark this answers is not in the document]
danielsn: Got it.

crashtracker-ffi/src/collector_windows/api.rs
```diff
@@ -0,0 +1,119 @@
...
+    _size: *mut u32,
+    _signature_count: *mut u32,
+) -> HRESULT {
+    let result: Result<(), _> = (|| {
+        anyhow::ensure!(
+            !exception_information.is_null(),
+            "exception_information is null"
+        );
+
+        let process_handle = unsafe { (*exception_information).hProcess };
```
danielsn, line 102: Safety comments

crashtracker-ffi/tests/collector_windows_test.rs
```diff
@@ -0,0 +1,78 @@
...
+    if let Ok(env_target_dir) = env::var("CARGO_TARGET_DIR") {
+        return PathBuf::from(env_target_dir);
+    }
+
+    let test_bin_location = PathBuf::from(env::args().next().unwrap());
+    let mut location_components = test_bin_location.components().rev().peekable();
+    loop {
+        let Some(c) = location_components.peek() else {
+            break;
+        };
```
danielsn, line 71: `while let Some(c) = location_components.peek() ...`

danielsn: Also, why is peek() needed at all?  Why can't this be 
```rust
for c in test_bin_location.components().rev()
...
```

Edit: I think I see, you want to leave the "target" string still there

crashtracker-ffi/tests/test_app/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#[cfg(not(windows))]
+fn main() {
+    panic!("This test is only supported on Windows");
+}
+
+#[cfg(windows)]
+fn main() {
+    let args: Vec<String> = std::env::args().collect();
```
danielsn, line 11: // comment showing the expected args, for the next maintainer

crashtracker-ffi/tests/test_app/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+    let exe_dir = exe_path.parent().unwrap();
+    let lib_path = exe_dir.join("datadog_crashtracker_ffi.dll");
+
+    init_crashtracking(crash_path, lib_path.to_str().unwrap());
+
+    // Force a segfault to crash
+    let ptr = std::ptr::null_mut::<i32>();
+    unsafe {
+        *ptr = 42;
+    }
```
danielsn, line 31: You might want to wrap this in `blackbox` like some of the other tests do to avoid the compiler doing something clever

crashtracker-ffi/tests/test_app/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+#[cfg(windows)]
+fn init_crashtracking(crash_path: &str, module_name: &str) -> bool {
+    use datadog_crashtracker_ffi::Metadata;
+    use ddcommon::Endpoint;
+    use ddcommon_ffi::CharSlice;
+    use std::path::Path;
+    use windows::Win32::System::Diagnostics::Debug::{SetErrorMode, THREAD_ERROR_MODE};
+
+    // Make sure WER is enabled
+    unsafe { SetErrorMode(THREAD_ERROR_MODE(0x0001)) };
```
danielsn, line 45: Safety comment

crashtracker/src/collector_windows/api.rs
```diff
@@ -0,0 +1,789 @@
...
+    // Check both HKEY_LOCAL_MACHINE and HKEY_CURRENT_USER
+    for root in [HKEY_LOCAL_MACHINE, HKEY_CURRENT_USER] {
+        let mut hkey = HKEY::default();
+
+        // Try to open the registry key
+        let open_result = unsafe { RegOpenKeyExW(root, subkey, None, KEY_READ, &mut hkey) };
+
+        if open_result == ERROR_SUCCESS {
+            // Check if the value exists
+            let query_result = unsafe { RegQueryValueExW(hkey, &name, None, None, None, None) };
```
danielsn, line 106: safety comments

crashtracker/src/collector_windows/api.rs
```diff
@@ -0,0 +1,789 @@
...
+            None,
+            None,
+            None,
+            None,
+            SYM_STKWALK_DEFAULT,
+        )
+    } {
+        let mut frame = StackFrame::new();
+
+        frame.ip = Some(format!("{:x}", native_frame.AddrPC.Offset));
```
danielsn, line 382: Do you want to specify a width here?

crashtracker/src/collector_windows/api.rs
```diff
@@ -0,0 +1,789 @@
...
+    pub data2: u16,
+    pub data3: u16,
+    pub data4: [u8; 8],
+}
+
+impl fmt::LowerHex for Guid {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "{:08x}{:04x}{:04x}{:02x}{:02x}{:02x}{:02x}{:02x}{:02x}{:02x}{:02x}",
```
danielsn, line 551: comment documenting why this format

danielsn (commented):
Looking pretty good!

crashtracker/src/collector_windows/api.rs
kevingosse: I don't understand, why would I? I'm just converting the ip to hex

crashtracker-ffi/Cargo.toml
kevingosse: The test application will compile but be empty.

crashtracker/src/collector_windows/api.rs
kevingosse: Done in https://github.com/DataDog/libdatadog/pull/892/commits/3c3a9de72faaa3431a4cef62a8c911b5bdd218e3

crashtracker/src/collector_windows/api.rs
bwoebi: I think he means to have consistent padding, like always 16 hex chars.

crashtracker-ffi/tests/collector_windows_test.rs
kevingosse: This is copy/pasted from bin_tests: https://github.com/DataDog/libdatadog/blob/9a4a79118bcf5e62f90eb93b0540ffbf9279ba98/bin_tests/src/lib.rs#L77

crashtracker/src/collector_windows/api.rs
kevingosse: Oh 🤔 Those values are not meant to be interpreted by a human, so I don't think the padding makes sense

crashtracker-ffi/tests/test_app/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+    // Get the directory of the current exe
+    let exe_path = std::env::current_exe().unwrap();
+    let exe_dir = exe_path.parent().unwrap();
+    let lib_path = exe_dir.join("datadog_crashtracker_ffi.dll");
+
+    init_crashtracking(crash_path, lib_path.to_str().unwrap());
+
+    // Force a segfault to crash
+    let ptr = std::ptr::null_mut::<i32>();
+    // SAFETY: Don't worry, we are crashing on purpose
```
danielsn, line 33: 🤣

danielsn (approved):

PR #896, opened 2025-02-25
[PROF-11405] Fix managed string storage causing corrupt profiles

ivoanjo:
# What does this PR do?

This PR fixes a bug with managed string storage's cache invalidation logic that caused corrupt profiles.

Specifically, profiles could be emitted that referenced wrong or even non-existing entries in the string table, which would fail to parse.

# Motivation

Fix managed string storage bug.

# Additional Notes

Prior to this PR, the managed string table was using the profile's `StringTable` pointer as a key for cache invalidation in the `ManagedStringId` to `StringId` mapping.

This is not correct since when a profile gets reset, the `StringTable` pointer **doesn't change**.

This bug lay dormant because Ruby uses two profiles and switches back and forth between them. This meant that in most cases the cache would go back and forth between the two profiles, working correctly... until a cache entry was left behind and got reused incorrectly.

# How to test the change?

I've added a test to make sure we don't regress here.


profiling/src/collections/string_storage.rs
```diff
@@ -32,13 +65,24 @@ impl ManagedStringStorage {
...
         storage.intern_new("").expect("Initialization to succeed");
         storage
     }
 
+    pub fn next_cached_profile_id(&mut self) -> anyhow::Result<CachedProfileId> {
+        let next_id = self.next_cached_profile_id.id;
+        self.next_cached_profile_id = InternalCachedProfileId {
+            id: next_id
+                .checked_add(1)
+                .ok_or_else(|| anyhow::anyhow!("Ran out of cached_profile_ids!"))?,
```
danielsn, line 81: `.context("Ran out of cached_profile_ids!")`

profiling/src/collections/string_storage.rs
```diff
@@ -23,6 +22,40 @@ pub struct ManagedStringStorage {
...
+//
+// **This struct is on purpose not Copy and not Clone to try to make it really hard to accidentally
+// reuse** when a profile gets reset.
+pub struct CachedProfileId {
+    id: u32,
+}
+
+#[derive(PartialEq, Debug, Copy, Clone)]
+struct InternalCachedProfileId {
+    id: u32,
```
danielsn, line 52: Do we want to use an incrementing number? If there are two profiles we switch between, would we get conflicts on this id?

morrisonlevi (approved):
I can't quite put my finger on it right now, but it seems like there should be a better way to solve this issue. But, it seems to work and does guard against at least one real-world bug, so it's progress. Approving so I don't let perfect be the enemy of better.

profiling/src/collections/string_storage.rs
ivoanjo: Fixed in 4ee0c0be1061da71fd48fa433d668ddc3a21a076

PR #898, opened 2025-02-26
Add v05 support in the trace exporter

hoolioh:
# What does this PR do?

Add support for v05 payload format to the trace exporter crate.

# Additional Notes

- v05 deserialization done in `trace_utils::msgpack_decoder::v05` module.
- v05 encoding done `trace_utils::span::v05` module.
- `SpanBytes` used for representing internally both v05 and v04. That avoids code duplication in computing stats since both formats are practically identical.


trace-utils/src/trace_utils.rs
```diff
@@ -594,9 +596,32 @@ pub fn collect_trace_chunks<T: tracer_payload::TraceChunkProcessor>(
     tracer_header_tags: &TracerHeaderTags,
     process_chunk: &mut T,
     is_agentless: bool,
-) -> TracerPayloadCollection {
+    deduplicate_strings: bool,
```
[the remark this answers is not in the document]
ekump: Speaking of the `TracerPayloadParams::TryInto`, I'm a bit confused why we aren't using it in the exporter. I could be mistaken, but I thought the original idea was that tracer_payload.rs would eventually replace collect_trace_chunks. We were just waiting to implement v0.5. Or do we have to wait for @VianneyRuhlmann's PR so that spans support slices?

My main concern is that we're diverging from what we implemented for the sidecar.

trace-utils/src/tracer_payload.rs
```diff
@@ -1,39 +1,132 @@
...
+                    }
+                    let sampled_spans = sampled_indexes
+                        .iter()
+                        .map(|i| std::mem::take(&mut chunk[*i]))
+                        .collect();
+                    *chunk = sampled_spans;
+                    true
+                });
+            }
+            TraceCollection::V07(_) => todo!("drop_chunks not implemented for v07"),
```
ekump, line 114: We haven't formalized a no-panic policy yet, but I think we should at least refrain from adding new panics. If it's easier to address this in a follow-up PR (I assume we'll have to add some new error types and change code to handle those errors, so not a trivial amount of work)that is ok. If that's the case can you add a `TODO` comment with a jira ID anywhere you've added new panics?

trace-utils/src/msgpack_decoder/v05/mod.rs
```diff
@@ -0,0 +1,419 @@
...
+use crate::msgpack_decoder::decode::{
+    map::read_map_len,
+    number::read_number_bytes,
+    string::{handle_null_marker, read_string_bytes},
+};
+use crate::span::SpanBytes;
+use std::collections::HashMap;
+
+const PAYLOAD_LEN: u32 = 2;
+const SPAN_ELEM_COUNT: u32 = 12;
```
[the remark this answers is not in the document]
hoolioh: The spec seems pretty tight but anyway I think it's a good idea to have a comment clarifying it.

trace-utils/src/trace_utils.rs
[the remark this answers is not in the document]
hoolioh: That would work. In any case I didn't want to make more meaningful changes in our data type architecture. I think after v05 is merged it would be good to try to get some ideas to consolidate the implementation between the sidecar and data-pipeline. Also I think it will be easier now that we have more tests in place.

trace-utils/src/tracer_payload.rs
```diff
@@ -1,39 +1,132 @@
...
+    V05,
     /// v0.7 encoding (TracerPayload).
     V07,
 }
 
 /// A collection of traces before they are turned into TraceChunks.
 pub enum TraceCollection {
     V07(Vec<Vec<pb::Span>>),
-    V04(Vec<Vec<SpanBytes>>),
+    TraceChunk(Vec<Vec<SpanBytes>>),
```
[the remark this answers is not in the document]
hoolioh: Yah, I struggled a bit with this. I'll try to clarify that.

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -114,66 +106,14 @@ fn add_path(url: &Uri, path: &str) -> Uri {
         }
         None => PathAndQuery::from_str(path),
     }
+    // TODO: APMSP-18190
```
ekump, line 109: nit: For the `TODO` comments we should also include a sentence describing the todo for those who don't have access to Jira. Something like "Properly handle non-OK states to prevent possible panics" or something like that.

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -280,9 +220,17 @@ impl TraceExporter {
         trace_count: usize,
     ) -> Result<AgentResponse, TraceExporterError> {
         self.check_agent_info();
+
         match self.input_format {
             TraceExporterInputFormat::Proxy => self.send_proxy(data.as_ref(), trace_count),
-            TraceExporterInputFormat::V04 => self.send_deser_ser(data),
+            TraceExporterInputFormat::V04 => match msgpack_decoder::v04::from_slice(data) {
```
ekump, line 226: non-blocking question - Should we eventually encapsulate some of this logic in the decoder? You pass a TraceCollection and the format and it figures out which particular version's decode to call?

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -588,38 +547,33 @@ impl TraceExporter {
...
-                    HealthMetric::Count(health_metrics::STAT_DESER_TRACES_ERRORS, 1),
-                    None,
-                );
-                return Err(TraceExporterError::Deserialization(err));
-            }
-        };
-
-        let num_traces = traces.len();
-
+    fn send_deser_ser(
```
ekump, line 571: We can leave it as is in this PR but this function name is even less accurate after these changes.

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -628,130 +582,168 @@ impl TraceExporter {
...
+                    &header_tags,
+                    &mut tracer_payload::DefaultTraceChunkProcessor,
+                    self.endpoint.api_key.is_some(),
+                    true,
+                )
+                .map_err(|e| {
+                    TraceExporterError::Deserialization(DecodeError::InvalidFormat(e.to_string()))
+                })?,
+                // TODO: APMSP-18190
+                _ => todo!(
```
ekump, line 635: nitpick - since the next PR is going to remove the panics it's not a big deal but `unimplemented!()` or `unreachable!()` are probably more appropriate than `todo!()` for a lot of these. The end result is the same, a panic. But the intention is clearer.

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -1051,6 +1051,17 @@ impl TraceExporterBuilder {
             query_params: self.query_params,
         })
     }
+
+    fn is_mode_allowed(input: TraceExporterInputFormat, output: TraceExporterOutputFormat) -> bool {
```
ekump, line 1055: I think naming this something closer to `is_input_output_formats_compatible` is more descriptive. It will also make others think twice about just adding further validations to this function rather than a more maintainable solution.

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -1743,7 +1655,7 @@ mod tests {
 
     #[test]
     #[cfg_attr(miri, ignore)]
-    fn exporter_metrics() {
+    fn exporter_metrics_v4() {
```
ekump, line 1658:
```suggestion
    fn exporter_metrics_v04() {
```

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -1796,6 +1708,124 @@ mod tests {
         metrics_endpoint.assert_hits(1);
     }
 
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn exporter_metrics_v5() {
```
ekump, line 1713:
```suggestion
    fn exporter_metrics_v05() {
```

ekump, line 1713:
```suggestion
    fn test_exporter_metrics_v05() {
```

data-pipeline/src/trace_exporter/mod.rs
ekump, line 1658:
```suggestion
    fn testexporter_metrics_v04() {
```

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -1448,7 +1347,12 @@ mod tests {
         datagram.trim_matches(char::from(0)).to_string()
     }
 
-    fn build_test_exporter(url: String, dogstatsd_url: String) -> TraceExporter {
+    fn build_test_exporter(
```
ekump, line 1350: Can you use this helper function instead of creating exporters in the [new test](https://github.com/DataDog/libdatadog/pull/898/files#diff-ceb25b8be20e4522d23de0732e538c8c49257277a37edbc8bd30af04de55f4ecR1713) you added?

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -1796,6 +1708,124 @@ mod tests {
...
+            exporter.runtime.block_on(async {
+                sleep(Duration::from_millis(100)).await;
+            })
+        }
+        metrics_endpoint.assert_hits(1);
+    }
+
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn exporter_metrics_v4_to_v5() {
```
ekump, line 1771:
```suggestion
    fn test_exporter_metrics_v04_to_v05() {
```

data-pipeline/tests/test_trace_exporter.rs
```diff
@@ -30,6 +33,37 @@ mod tracing_integration_tests {
...
+            12341,
+            1,
+            false,
+            &mut dict,
+            Some(vec![
+                ("_dd_metric1".to_string(), 1.1),
+                ("_dd_metric2".to_string(), 2.2),
+            ]),
+        );
+        let span_2 = create_test_v05_span(1234, 12343, 12341, 1, false, &mut dict, None);
```
ekump, line 51: For v0.4 tests we serialize json rather than using the same Span struct we deserialize with to avoid masking potential flaws in our model. Can / should we do the same for v0.5?

data-pipeline/tests/snapshots/compare_v05_trace_snapshot_test.json
```diff
@@ -0,0 +1,66 @@
...
+      "parent_id": 12341,
+      "start": 1,
+      "duration": 5,
+      "error": 0,
+      "meta": {
+        "env": "test-env",
+        "service": "test-service",
+        "runtime-id": "test-runtime-id-value"
+      },
+      "metrics": {},
```
ekump, line 39: Do the empty fields for `error`, `metrics` and `type` have to be present for v0.5? They do not for v0.4 and the test-agent doesn't complain about it if you include optional fields. There was a discussion around this [here](https://github.com/DataDog/libdatadog/pull/897#discussion_r1971934009).

trace-utils/src/msgpack_decoder/v05/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+/// let decoded_span = &decoded_traces[0][0];
+/// assert_eq!("", decoded_span.name.as_str());
+/// ```
+pub fn from_slice(mut data: tinybytes::Bytes) -> Result<(Vec<Vec<SpanBytes>>, usize), DecodeError> {
+    let data_elem = rmp::decode::read_array_len(unsafe { data.as_mut_slice() })
+        .map_err(|_| DecodeError::InvalidFormat("Unable to read payload len".to_string()))?;
+
+    if data_elem != PAYLOAD_LEN {
+        return Err(DecodeError::InvalidFormat(
+            "Invalid payload size".to_string(),
```
ekump, line 77: is there value in including the `data_elem` and `PAYLOAD_LEN` values in the error message?

trace-utils/src/msgpack_decoder/v05/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+        ));
+    }
+
+    let dict = deserialize_dict(&mut data)?;
+
+    let trace_count = rmp::decode::read_array_len(unsafe { data.as_mut_slice() })
+        .map_err(|_| DecodeError::InvalidFormat("Unable to read trace len".to_string()))?;
```
ekump, line 84: I realize you are by and large copying behavior from v04, so I don't think this is necessarily blocking the PR. But after this PR is merged we should probably audit all the places where we are mapping errors and not including the original error message and change the behavior to provide more detail in our error messages.

trace-utils/src/span/mod.rs
```diff
@@ -1,4 +1,164 @@
...
+/// or borrowed (e.g. &str). To define a generic function taking any `Span<T>` you can use the
+/// [`SpanValue`] trait:
+/// ```
+/// use datadog_trace_utils::span::{Span, SpanText};
+/// fn foo<T: SpanText>(span: Span<T>) {
+///     let _ = span.meta.get("foo");
+/// }
+/// ```
+#[derive(Clone, Debug, Default, PartialEq, Serialize)]
+pub struct Span<T>
```
ekump, line 82: I think the namespacing here is confusing. trace_utils::span::Span is a v04 span, but trace_utils::span::v05::Span is a v05 span? Presumably we should change this so v04 is trace_utils::span::v04::Span?

trace-utils/src/span/v05/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub parent_id: u64,
+    pub start: i64,
+    pub duration: i64,
+    pub error: i32,
+    pub meta: HashMap<u32, u32>,
+    pub metrics: HashMap<u32, f64>,
+    pub r#type: u32,
+}
+
+pub fn from_span_bytes(span: &SpanBytes, dict: &mut SharedDict) -> Result<Span> {
```
ekump, line 32: It looks like unit tests are missing for this public function.

trace-utils/src/trace_utils.rs
ekump, line 599: @VianneyRuhlmann Agreed. I think my comment could have been more succulently stated as a rhetorical (and unhelpful) "why didn't we clean this all up earlier?".

Not for this PR, but I also wonder if we should have a concrete type that encapsulates some of the logic present here for a V05 `TracerPayloadCollection` rather than using a tuple.

ekump, line 599: Any reason why the input param `deduplicate_strings` shouldn't more directly convey the intent of using v05? The current name is ambiguous, imo.

trace-utils/src/tracer_payload.rs
```diff
@@ -64,6 +173,7 @@ impl TracerPayloadCollection {
                     dest.append(src)
                 }
             }
+            TracerPayloadCollection::V05(_) => todo!("Append for V05 not implemented"),
```
ekump, line 176: forgot the comment for this todo!()

ekump (approved):
The majority of my comments can be addressed in a follow-up PR, the ones that should be addressed in this PR are relatively small and can be merged without further review from me. 

Could you put together a list of all the things we want to follow-up on? We can come up with a plan with @VianneyRuhlmann on how to best tackle some of our ideas after he merges his related changes.

data-pipeline/tests/test_trace_exporter.rs
hoolioh: Oh, with v05 is a problem because there are some incompatibilities between the formats. The one affecting us is the fact that json can't have numbers as keys in fields so meta and metrics can't be represented directly from json unless we implement some custom serialization for that json objects.
One thing that we could do is maybe generate the payloads from a reference implementation and use that both for data-pipeline and the agent.

trace-utils/src/span/mod.rs
hoolioh: I struggled with this a little bit and decided to leave it like that because the other options were even more confusing. The options I evaluated:

- What Vianney suggests, duplicate the structure without the serde annotations but we would need to copy the fields in the v04 span before being serialized which is not optimal. 
- Using one common span struct by defining a wrapper composed of the Span and some fields that could be set at runtime to annotate the format (v04 or v05). Then anotating the optional/particular fields with a custom function so they are conditionally serialized given the format selected. I came up with the wrapper solution because serde function annotations doesn't allow parameters. In the end I discarded this solution because I was unsure it was clearer and it was more code to the PR.
- Another one could be give it another name like `TraceChunkSpan` (dont' know if the terminology is correct) so it's a Span which is valid for all trace chunk based formats. 

WDYT?

data-pipeline/tests/snapshots/compare_v05_trace_snapshot_test.json
hoolioh: Yes, the [spec](https://github.com/DataDog/datadog-agent/blob/579d60cbfc8635559338b19fa855bf59b6d10f00/pkg/trace/api/version.go#L71) states that all elements should be there

PR #899, opened 2025-02-26
feat: add dev container for consistent development environment

ganeshnj:
# What does this PR do?

- ensures consistent development setup
- includes necessary dependencies and configurations
- allows step-in for things like how container id is generated since the whole workspace is inside a container

I went with minimal setup and targeting only one OS, but this can be extended with multiple container (debian, alpine etc)

# Motivation

I have been using it locally by keeping in my stashes, but this is general problem since we are all (or almost) on osx and want to test and debug how the libraries behave on linux. 

# Additional Notes

none.

# How to test the change?

Follow the readme.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,38 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/rust
+{
+    "name": "Rust",
+    // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+    "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
+
+    // Allow access to host machine
+    "runArgs": ["--network=host"],
+
+    // Use 'mounts' to make the cargo cache persistent in a Docker Volume.
+    "mounts": [
+        {
+            "source": "devcontainer-cargo-cache-${devcontainerId}",
+            "target": "/usr/local/cargo",
+            "type": "volume"
+        }
+    ],
+
+    // Features to add to the dev container. More info: https://containers.dev/features.
+    "features": {
+        "ghcr.io/devcontainers/features/rust:1": {},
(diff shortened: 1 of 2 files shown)


ivoanjo (approved):
:+1: LGTM, thanks for the nice addition :)

ekump (approved):

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,38 @@
...
+    // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+    "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
+
+    // Allow access to host machine
+    "runArgs": ["--network=host"],
+
+    // Use 'mounts' to make the cargo cache persistent in a Docker Volume.
+    "mounts": [
+        {
+            "source": "devcontainer-cargo-cache-${devcontainerId}",
```
bantonsson, line 14: So what does this translate to? Is it using Docker `cache` mounts underneath? I mean something like `--mount=type=cache,target=/usr/local/cargo`, or is it specific to dev containers.

ganeshnj: you get a volume which is attached to dev container just like any other volume

```
docker volume ls | grep devcontainer-cargo-cache
local     devcontainer-cargo-cache-1u4p07avg375gi7o172o2cfupr5iafgjj9t7hvog8dscsv7co5df
```
since it is container specific, running the same container would give you same volume and helps avoiding repetitive whole world of internet download.

bantonsson: Ok, so it's manually maintaining what `--mount=type=cache,...` gives you.

Do you know if it has a cached `target` directory as well for the output of the build? I'm just curious.

ganeshnj: I didn't try but I think you can as per spec https://github.com/devcontainers/spec/blob/main/schemas/devContainer.base.schema.json#L711C8-L711C12

you can bring your own arbitrary location if this is what you mean.

PR #900, opened 2025-02-26
Handle EINTR in blocking ipc sender

bwoebi:
Without this change file descriptions were not sent when send_with_fd was interrupted by a signal, causing the connection to become broken.

Also use statm for memory tracking in sidecar (we don't need full accuracy for the watchdog).


morrisonlevi (approved):

morrisonlevi:
I approved, but of course that's dependent on getting jobs to pass ^_^

Using statm should fix one perf issue found with the sidecar, which does affect one customer that is reporting elevated overhead with APM.

PR #901, opened 2025-02-26
feat(dogstatsd): compress metrics

astuyve:
- **feat: use zstd for dogstatsd**
- **feat: licenses**

# What does this PR do?
Uses zstd for dogstatsd. Not optional, always enabled.

# Motivation
should help a small bit with network bytes:
Screenshot 2025-02-26 at 2 52 29 PM


dogstatsd/Cargo.toml
```diff
@@ -25,6 +25,7 @@ tokio = { version = "1.37.0", default-features = false, features = ["macros", "r
 tokio-util = { version = "0.7.11", default-features = false }
 tracing = { version = "0.1.40", default-features = false }
 regex = { version = "1.10.6", default-features = false }
+zstd = { version = "0.13.3", default-features = false }
```
alexgallotta, line 28: should we use gzip or zstd for both logs and metrics? To avoid adding 2 libs?

astuyve: We use zstd for both

astuyve: https://github.com/DataDog/datadog-lambda-extension/blob/main/bottlecap/src/logs/flusher.rs#L13

morrisonlevi (approved):
Looks good to me, now that the content-encoding matches success/failure of compressing.

PR #902, opened 2025-02-27
[PROF-11456] Add extra validations for incorrect profiling location ids

ivoanjo:
# What does this PR do?

This PR adds extra validations for incorrect location ids being generated and/or being present at serialization time.

# Motivation

In incident 35390 (JIRA PROF-11456) we observed invalid `location_id`s being present in emitted profiles.

We've stared hard at the code and can't really see how they could ever happen. By adding these checks, if we see the issue come up again, then we'll be sure that it happened after all the checks passed.

# Additional Notes

N/A

# How to test the change?

Since we don't know how to trigger the incorrect ids -- and they may not happen ever, it's hard to exercise these branches; yet, they're quite simple so I think it's fine to live with them as-is.


profiling/src/collections/identifiable/mod.rs
```diff
@@ -66,13 +67,34 @@ pub trait Dedup<T: Item> {
     /// Panics if the number of items overflows the storage capabilities of
     /// the associated Id type.
     fn dedup(&mut self, item: T) -> <T as Item>::Id;
+
+    /// Deduplicate the Item, and check if the generated Id is valid.
+    fn checked_dedup(&mut self, item: T) -> anyhow::Result<<T as Item>::Id>;
```
danielsn, line 72: Having a panic in the code isn't great.  Depending on how many places dedup is used in, maybe this should be the only version and clients can `?` on it?

profiling/src/internal/profile/mod.rs
```diff
@@ -1062,7 +1080,7 @@ mod api_tests {
     }
 
     #[test]
-    fn lazy_endpoints() -> anyhow::Result<()> {
+    fn lazy_endpoints() {
```
danielsn, line 1083: Why this change?

profiling/src/collections/identifiable/mod.rs
ivoanjo: I hesitated on doing that -- that `?` will need to propagate quite far. Not sure it's worth the extra work given id overflow would mean we have 2**32 items so we'd already be using many GiB's for the profiler?

profiling/src/internal/profile/mod.rs
ivoanjo: It doesn't look correct for a test to exit early if something goes wrong, right: It should blow up?

profiling/src/collections/identifiable/mod.rs
danielsn: Fine for this PR, as we don't change the existing behaviour.

profiling/src/collections/identifiable/mod.rs
```diff
@@ -66,13 +67,34 @@ pub trait Dedup<T: Item> {
...
+    fn checked_dedup(&mut self, item: T) -> anyhow::Result<<T as Item>::Id> {
+        let (id, _) = self.insert_full(item);
+
+        anyhow::ensure!(
+            id < self.len(),
+            "out of bounds id generated {:?}, len was {:?}",
+            id,
+            self.len()
+        );
+        small_non_zero_pprof_id(id).ok_or_else(|| anyhow!("invalid id generated {:?}", id))?;
```
danielsn, line 94: Can use `.with_context()` here.

profiling/src/internal/profile/mod.rs
danielsn: Returning an error causes the test runner to blow up.  But de gustibus non est disputandum.

danielsn (approved):

profiling/src/internal/profile/mod.rs
ivoanjo: Ah! I didn't know that. In that case, these changes aren't relevant. Reverted them in f4dd6ae172ae645f4b729defcedfe7b4958ed846

profiling/src/collections/identifiable/mod.rs
ivoanjo: Fixed in 2ddd704d9e30798ac219c34673ea71e7c41baf64

PR #906, opened 2025-03-04
[APMSP-1816] Add retry logic to the StatsExporter

anais-raison:
# What does this PR do?

Add the same retry logic that we have in the TraceExporter.

# Motivation

Because if the request fail we do nothing and it's not retried. Now it is.

# Additional Notes

I also removed the client field in the StatsExporter struct because the it was no longer needed with the changes I did.

# How to test the change?

By checking that the tests are still working.


data-pipeline/src/stats_exporter.rs
```diff
@@ -82,30 +81,24 @@ impl StatsExporter {
...
+        headers.insert(
+            hyper::header::CONTENT_TYPE.as_str(),
+            ddcommon::header::APPLICATION_MSGPACK_STR.to_string(),
+        );
 
-        let req = req_builder.body(hyper::Body::from(body))?;
+        let strategy = RetryStrategy::default();
 
-        let resp = self.client.request(req).await?;
+        let result = send_with_retry(&self.endpoint, body, &headers, &strategy, None).await;
```
ekump, line 93: nitpick: you can skip declaring strategy above and just pass `&RetryStrategy:default()` in the function call.

data-pipeline/src/stats_exporter.rs
```diff
@@ -82,30 +81,24 @@ impl StatsExporter {
...
 
-        if !resp.status().is_success() {
-            anyhow::bail!(
-                "received {} status code from the agent",
-                resp.status().as_u16()
-            );
+        if let Ok((resp, _)) = result {
```
ekump, line 95: This will return an `Ok` result when `send_with_retry` returns [SendWithRetryError](https://github.com/DataDog/libdatadog/blob/7f40e141a19874a40da892cbd5770f130c408913/trace-utils/src/send_with_retry/mod.rs#L22), which I don't think we want. 

At a minimum, we should be [mapping](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err) the error we get from `send_with_retry` and provide a more detailed error message in the `anyhow::bail`. 

@VianneyRuhlmann - Possibly beyond the scope of this PR, but should the StatsExporter have a proper error type instead of anyhow?

data-pipeline/src/stats_exporter.rs
```diff
@@ -82,30 +81,24 @@ impl StatsExporter {
...
+            if !resp.status().is_success() {
+                anyhow::bail!(
+                    "received {} status code from the agent",
+                    resp.status().as_u16()
+                );
+            }
```
ekump, line 101: The existing unit tests don't cover returning errors. Would it be possible to add one that covers receiving a not Ok status code as well as when a `SendWithRetryError` is received?

Unfortunately, we don't have a good mocking solution in libdatadog to make this easy, so it may not be worth the effort right now.

anais-raison: Just added the test.

data-pipeline/src/stats_exporter.rs
anais-raison: I replace the `if let` by mapping, and added a log for the error.

ekump (approved):
Thanks for addressing comments so quickly, LGTM!

PR #907, opened 2025-03-05
[trace-mini-agent] add trace aggregator

duncanista:
# What does this PR do?

Adds a Trace Aggregator for the Trace Flusher.

# Motivation

There is a memory leak bc the buffer was growing without bounds.

# Additional Notes

There is still an unbounded queue, but its OK bc that will shrink in size on every batch.

# How to test the change?

Added tests for the aggregator.


astuyve (approved):
Mostly copied over from bottlecap, lgtm

duncanpharvey (approved):
Ran a quick smoke test for Node Azure Functions. Traces are coming through successfully. LGTM!

Screenshot 2025-03-05 at 4 29 33 PM

PR #908, opened 2025-03-05
Add poisson upscaling without storing count in the sample types

realFlowControl:
# What does this PR do?

So far our poisson upscaling relied on the count of samples to be part of the sample data. This works fine for sample types where we also collect the count, like memory allocations, where we collect the `alloc-size` and `alloc-count`. In that case we can just use the `alloc-count` from the samples collected. In some cases, like I/O profiling, we add a lot of sample types (socket/file read/write times/bytes) but also adding the count for all of these operations would just bloat the resulting pprof, without a lot of value, so we need a way to inject the count to calculate the scale without having it in a sample type. This PR adds this.

# Motivation

I/O profiling in PHP is such a feature, where we add 8 sample types already. Adding all the counts would add another 8 sample types for no value (besides upscaling, which we can solve differently).

# Additional Notes

- I am planing to release this as `v15.1.0` as the PHP profiler is not yet on `v16` and I do not want to mix to many things together.
- The `clippy` errors are unrelated to the changes in this PR

# How to test the change?

https://github.com/DataDog/dd-trace-php/pull/3118


profiling/src/api.rs
```diff
@@ -153,6 +159,15 @@ impl std::fmt::Display for UpscalingInfo {
...
             ),
+            UpscalingInfo::PoissonNonSampleTypeCount {
+                sum_value_offset,
+                count_value,
+                sampling_distance,
+            } => write!(
+                f,
+                "Poisson = sum_value_offset: {}, count_value: {}, sampling_distance: {}",
+                sum_value_offset, count_value, sampling_distance
+            ),
```
morrisonlevi, line 170: Clippy would probably suggest this:

```suggestion
            } => write!(
                f,
                "Poisson = sum_value_offset: {sum_value_offset}, count_value: {count_value}, sampling_distance: {sampling_distance}"
            ),
```

(or something close, writing code in a text box in a browser is hard)

This kind of thing is repeated a few times in this file (in and out of the PR).

morrisonlevi (approved):
I have a code style suggestion that clippy would probably give too. Aside from that, it looks fine to me, although I did not check the upscaling math--maybe someone more familiar can look at it.

profiling/src/api.rs
realFlowControl: Interestingly clippy does not complain. I'd tend keep it as is for consistency reasons and clean up al of this in another PR, WDYT?

gleocadie (approved):
LGTM

profiling/src/api.rs
realFlowControl: See https://github.com/DataDog/libdatadog/pull/913

PR #912, opened 2025-03-07
[ APMSP-1674] Support Span Events in Trace Exporter

anais-raison:
# What does this PR do?

Add the `span_events` field in the span model in Trace Exporter.

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/trace-utils/src/msgpack_decoder/decode/span_event.rs
+++ b/trace-utils/src/msgpack_decoder/decode/span_event.rs
@@ -6,7 +6,7 @@ use crate::msgpack_decoder::decode::number::read_number_bytes;
 use crate::msgpack_decoder::decode::string::{
     handle_null_marker, read_string_bytes, read_string_ref,
 };
-use crate::span::v04::{AttributeAnyValueBytes, AttributeArrayValueBytes, SpanEventBytes};
+use crate::span::{AttributeAnyValueBytes, AttributeArrayValueBytes, SpanEventBytes};
 use rmp::decode::ValueReadError;
 use rmp::Marker;
 use std::collections::HashMap;
--- a/trace-utils/src/msgpack_decoder/decode/span_event.rs
+++ b/trace-utils/src/msgpack_decoder/decode/span_event.rs
@@ -131,7 +131,6 @@ impl FromStr for AttributeAnyKey {
     }
 }
 
-// TODO : find a way to decode it by checking the `type` thing to get the type of the value
 fn decode_attribute_any(buf: &mut Bytes) -> Result<AttributeAnyValueBytes, DecodeError> {
     let mut attribute: Option<AttributeAnyValueBytes> = None;
     let attribute_size =
@@ -181,7 +180,7 @@ fn decode_attribute_any(buf: &mut Bytes) -> Result<AttributeAnyValueBytes, Decod
             ))
         }
     } else {
-        Err(DecodeError::InvalidFormat("todo".to_owned()))
+        Err(DecodeError::InvalidFormat("Invalid attribute".to_owned()))
     }
 }
 
@@ -300,7 +299,7 @@ fn decode_attribute_array(
             ))
         }
     } else {
-        Err(DecodeError::InvalidFormat("todo".to_owned()))
+        Err(DecodeError::InvalidFormat("Invalid attribute".to_owned()))
     }
 }
 
--- a/trace-utils/src/msgpack_decoder/decode/span_event.rs
+++ b/trace-utils/src/msgpack_decoder/decode/span_event.rs
@@ -1,0 +1,314 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::msgpack_decoder::decode::error::DecodeError;
+use crate::msgpack_decoder::decode::number::read_number_bytes;
+use crate::msgpack_decoder::decode::string::{
+    handle_null_marker, read_string_bytes, read_string_ref,
+};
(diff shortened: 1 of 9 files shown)


trace-utils/src/span/v04/span.rs
```diff
@@ -121,8 +126,117 @@ where
...
+    T: SpanText,
+{
+    pub time_unix_nano: u64,
+    pub name: T,
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    pub attributes: HashMap<T, AttributeAnyValue<T>>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum AttributeAnyValue<T>
```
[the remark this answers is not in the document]
anais-raison: Great idea, thanks, I switched the `AttributeAnyValue` to that.

trace-utils/src/test_utils/mod.rs
```diff
@@ -54,6 +54,7 @@ pub fn create_test_no_alloc_span(
         r#type: BytesString::default(),
         meta_struct: HashMap::new(),
         span_links: vec![],
+        span_events: vec![],
```
ekump, line 57: You probably want to include the span_events key in any other methods in this file that generate v04 spans. For example, our [integration tests](https://github.com/DataDog/libdatadog/blob/9a4a79118bcf5e62f90eb93b0540ffbf9279ba98/data-pipeline/tests/test_trace_exporter.rs#L69) use [create_test_json_span](https://github.com/DataDog/libdatadog/blob/9a4a79118bcf5e62f90eb93b0540ffbf9279ba98/trace-utils/src/test_utils/mod.rs#L296).

trace-utils/src/msgpack_decoder/decode/span_event.rs
```diff
@@ -0,0 +1,411 @@
...
+        }
+    } else {
+        Err(DecodeError::InvalidFormat("Invalid attribute".to_owned()))
+    }
+}
+
+fn type_from_attribute(attribute: &AttributeAnyValueBytes) -> u8 {
+    match attribute {
+        AttributeAnyValueBytes::SingleValue(value) => type_from_attribute_array(value),
+        AttributeAnyValueBytes::Array(_) => 4,
```
ekump, line 179: This should be a named constant.

trace-utils/src/msgpack_decoder/decode/span_event.rs
```diff
@@ -0,0 +1,411 @@
...
+        Err(DecodeError::InvalidFormat("Invalid attribute".to_owned()))
+    }
+}
+
+fn type_from_attribute_array(attribute: &AttributeArrayValueBytes) -> u8 {
+    match attribute {
+        AttributeArrayValueBytes::String(_) => 0,
+        AttributeArrayValueBytes::Boolean(_) => 1,
+        AttributeArrayValueBytes::Integer(_) => 2,
+        AttributeArrayValueBytes::Double(_) => 3,
```
ekump, line 314: These values should be named constants. It complies to the same thing, but is a bit better for clarity and maintainability.

trace-utils/src/msgpack_decoder/decode/span_event.rs
```diff
@@ -0,0 +1,411 @@
...
+            Err(DecodeError::InvalidType(
+                "Type mismatch for attribute".to_owned(),
+            ))
+        }
+    } else {
+        Err(DecodeError::InvalidFormat("Invalid attribute".to_owned()))
+    }
+}
+
+fn type_from_attribute_array(attribute: &AttributeArrayValueBytes) -> u8 {
```
ekump, line 309: Instead of a function here, should we instead implement the [From](https://doc.rust-lang.org/std/convert/trait.From.html) trait for `u8` in `AttributeArrayValueBytes`?

Something like 

```rust
impl From<&AttributeArrayValueBytes> for u8 {
    fn from(attribute: &AttributeArrayValueBytes) -> u8 {
        match attribute {
            AttributeArrayValueBytes::String(_) => 0,
            AttributeArrayValueBytes::Boolean(_) => 1,
            AttributeArrayValueBytes::Integer(_) => 2,
            AttributeArrayValueBytes::Double(_) => 3,
        }
    }
}
```

and then instead of calling this function now like `type_from_attribute_array(&foo)` you could just do `&foo.into()` (implementing `From` gets you `into` automatically.

anais-raison: Done (but for `AttributeAnyValue<T>` and `AttributeArrayValue<T>` instead) !

trace-utils/src/test_utils/mod.rs
anais-raison: I just added it for the `create_test_json_span` since all the others function in that files are made for protobuf or the v05.

trace-utils/src/msgpack_decoder/decode/span_event.rs
ekump: Since the implementation lives in AttributeArrayValueBytes and the numbers don't have to be directly referenced elsewhere I don't think it's necessary.

trace-utils/src/span/mod.rs
```diff
@@ -124,8 +129,127 @@ where
...
+        AttributeArrayValue::Integer(value) => state.serialize_field("int_value", value),
+        AttributeArrayValue::Double(value) => state.serialize_field("double_value", value),
+    }
+}
+
+impl<T> From<&AttributeArrayValue<T>> for u8
+where
+    T: SpanText,
+{
+    fn from(attribute: &AttributeArrayValue<T>) -> u8 {
```
ekump, line 238: I know it's probably covered in your other tests, but it's not a bad idea to add a unit test specifically for this trait impl. It may seem unnecessary for such trivial logic, but could come in handy as part of future refactors.

PR #916, opened 2025-03-11
feat: Impl eq for replacerule so we can parse a yaml file

astuyve:
# What does this PR do?
We support `datadog.yaml` alongside environment variables. PartialEq allows figment to compare config instances

# Motivation
Fixes a yaml parsing issue with datadog.yaml

# Additional Notes

# How to test the change?

### DIFF
--- a/trace-obfuscation/src/replacer.rs
+++ b/trace-obfuscation/src/replacer.rs
@@ -12,7 +12,13 @@ struct RawReplaceRule {
     repl: String,
 }
 
-#[derive(Debug)]
+impl PartialEq for ReplaceRule {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name && self.repl == other.repl && self.re.as_str() == other.re.as_str()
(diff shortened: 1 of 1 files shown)


duncanista (approved):

trace-obfuscation/src/replacer.rs
```diff
@@ -12,7 +12,13 @@ struct RawReplaceRule {
     repl: String,
 }
 
-#[derive(Debug)]
+impl PartialEq for ReplaceRule {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name && self.repl == other.repl && self.re.as_str() == other.re.as_str()
```
astuyve, line 17: Can't derive Eq/PartialEq because rust devs say that two regular expressions should be equal if their state machines are equal, not just the literal string. That's a level deeper than we need, so this is sufficient.

ekump (approved):

PR #933, opened 2025-03-17
chore: update action versions in workflow files

ganeshnj:
# What does this PR do?

Pins all the action versions to specific commit sha.

# Motivation

Tags can be moved and we could end up using action versions which we don't intend to.

# Additional Notes

I bumped the action versions to their minor versions.

# How to test the change?

🤞  all actions pass


gleocadie (approved):
LGTM

ekump (approved):

PR #940, opened 2025-03-18
Push retry strategy into client

astuyve:
# What does this PR do?
The original idea was to make retries happen from the users of this lib, but now that we've added compression this adds the overhead of re-compressing the request on each retry. This PR changes the behavior to retry in the request client instead.

# Motivation
Saves CPU time

### DIFF
--- a/dogstatsd/src/datadog.rs
+++ b/dogstatsd/src/datadog.rs
@@ -261,7 +261,7 @@ impl DdApi {
                         return Ok(resp);
                     }
                     match retry_strategy {
-                         RetryStrategy::LienarBackoff(max_attempts, _)
+                         RetryStrategy::LinearBackoff(max_attempts, _)
                         | RetryStrategy::Immediate(max_attempts)
                             if attempts >= max_attempts =>
                         {
--- a/dogstatsd/src/datadog.rs
+++ b/dogstatsd/src/datadog.rs
@@ -166,12 +166,7 @@ impl DdApi {
         let safe_body = serde_json::to_vec(&series)
             .map_err(|e| ShippingError::Payload(format!("Failed to serialize series: {e}")))?;
         debug!("Sending body: {:?}", &series);
-        self.ship_data(
-            url,
-            safe_body,
-            "application/json",
-        )
-        .await
+        self.ship_data(url, safe_body, "application/json").await
     }
 
     pub async fn ship_distributions(
@@ -183,12 +178,8 @@ impl DdApi {
             .write_to_bytes()
             .map_err(|e| ShippingError::Payload(format!("Failed to serialize series: {e}")))?;
         debug!("Sending distributions: {:?}", &sketches);
-        self.ship_data(
-            url,
-            safe_body,
-            "application/x-protobuf",
-        )
-        .await
+        self.ship_data(url, safe_body, "application/x-protobuf")
+            .await
         // TODO maybe go to coded output stream if we incrementally
         // add sketch payloads to the buffer
         // something like this, but fix the utf-8 encoding issue
--- a/dogstatsd/src/datadog.rs
+++ b/dogstatsd/src/datadog.rs
@@ -251,29 +251,32 @@ impl DdApi {
             attempts += 1;
             match builder.try_clone() {
                 Some(builder) => {
-                    let resp = builder.send().await.map_err(|e| {
-                        ShippingError::Destination(
-                            e.status(),
-                            format!("Failed to send request: {e}"),
-                        )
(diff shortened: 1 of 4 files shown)


dogstatsd/src/datadog.rs
```diff
@@ -219,12 +234,68 @@ impl DdApi {
...
+                                {
+                                    // handle if status code missing like timeout
+                                    return Err(ShippingError::Destination(
+                                        resp.status(),
+                                        "Failed to send request after {max_attempts}".to_string(),
+                                    ));
+                                }
+                                RetryStrategy::LinearBackoff(_, delay) => {
+                                    tokio::time::sleep(Duration::from_secs(delay)).await;
+                                }
```
hghotra, line 278: This will keep re-trying indefinitely until timeout?

hghotra (approved):

dogstatsd/src/datadog.rs
astuyve: So this block will just delay and then at the top of the loop (251) we'll increment the attempts. After max attempts is hit, it'll error.

I can add a better test

dogstatsd/src/datadog.rs
```diff
@@ -219,12 +234,68 @@ impl DdApi {
...
+    ) -> Result<Response, ShippingError> {
+        let mut attempts = 0;
+        loop {
+            attempts += 1;
+            match builder.try_clone() {
+                Some(builder) => {
+                    let resp = builder.send().await;
+                    match resp {
+                        Ok(resp) if resp.status().is_success() => return Ok(resp),
+                        Ok(resp) => {
```
astuyve, line 257: Todo: fix - this should pass into the error block

hghotra (approved):

dogstatsd/src/datadog.rs
```diff
@@ -193,6 +207,7 @@ impl DdApi {
         url: String,
         body: Vec<u8>,
         content_type: &str,
+        retry_strategy: RetryStrategy,
```
duncanista, line 210: `ship_data` has access to `self`, can't you just access to it by doing `self.retry_strategy.clone()` instead of passing it on every other method?

dogstatsd/src/datadog.rs
```diff
@@ -219,12 +234,65 @@ impl DdApi {
...
         let elapsed = start.elapsed();
         debug!("Request to {} took {}ms", url, elapsed.as_millis());
-        resp.map_err(|e| ShippingError::Destination(e.status(), format!("Cannot reach {url}")))
+        resp
     }
+
+    async fn send_with_retry(
+        &self,
+        builder: reqwest::RequestBuilder,
+        retry_strategy: RetryStrategy,
```
duncanista, line 247: `send_with_retry` also has access to `self`, do we really need to pass the `retry_strategy` here?

dogstatsd/src/datadog.rs
```diff
@@ -219,12 +234,65 @@ impl DdApi {
...
+                                Err(err) => err.status(),
+                            };
+                            // handle if status code missing like timeout
+                            return Err(ShippingError::Destination(
+                                status,
+                                "Failed to send request after {max_attempts}".to_string(),
+                            ));
+                        }
+                        RetryStrategy::LinearBackoff(_, delay) => {
+                            tokio::time::sleep(Duration::from_millis(delay)).await;
```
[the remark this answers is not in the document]
astuyve: yes

dogstatsd/src/datadog.rs
astuyve: Yup you're right, a holdover from when I initially passed it around

dogstatsd/src/datadog.rs
astuyve: done

dogstatsd/src/flusher.rs
```diff
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::aggregator::Aggregator;
-use crate::datadog;
-use datadog::{DdApi, MetricsIntakeUrlPrefix};
+use crate::datadog::{self};
+use datadog::{DdApi, MetricsIntakeUrlPrefix, RetryStrategy};
```
duncanista, line 6: Are these two different crates? can't we merge this?

duncanista (commented):
Left some comments

ekump:
CI is blocked on macOS runner issues. This PR is isolated to serverless and should not impact artifact builds. Bypassing checks and merging at @astuyve request.

PR #945, opened 2025-03-18
perf(profiling): always be serializing

morrisonlevi:
# What does this PR do?

Immediately serializes Mappings, Locations, and Functions to their in-wire protobuf format, and dedups based on that instead. See [Always Be Serializing](https://docs.google.com/document/d/1jWcVpdvo4bYErKdOdlWozckXEy3K_Tvd_i3CdDr7hVc/edit?usp=sharing).

# Motivation

- Writing them in-order can help with certain debugging scenarios.
- Saves space when a given language doesn't use certain fields.
- Might save CPU from fewer intermediate representations, but in-wire serialization is not trivial so we'll have to see how it shakes out.

# Additional Notes

I bounced ideas off @ivoanjo and he helped me refine the implementation. He also helped me test the performance.

I didn't delete the now-unused representations to keep the PR more focused, but there are some internal representations that could be removed if this merges.

# How to test the change?

The external APIs are the same. However, some fuzzer tests are failing. I _think_ this is an issue with the fuzzer harness/test, because they look fine when I manually serialize and unserialize the problematic cases. If pursued to production quality, we'd want to get to the root of this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1522,7 +1522,7 @@ dependencies = [
  "futures",
  "futures-core",
  "futures-util",
- "hashbrown 0.14.5",
+ "hashbrown 0.15.2",
  "http 0.2.12",
  "http-body 0.4.6",
  "hyper 0.14.31",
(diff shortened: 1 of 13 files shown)


morrisonlevi:
Closing as "not worth it" at this time. It may be worth it in the future. We've discussed maybe trusting the runtimes with having the ids, so they could cache on their side. This could lead to fewer deduplications being needed, so the cost of serializing at deduplication time would matter less.

PR #955, opened 2025-03-20
ci: use a larger github runner for miri

morrisonlevi:
# What does this PR do?

This uses a larger runner size for the miri job on GitHub Actions.

It also updates nextest from 0.9.81 to 0.9.92. There were some notes in some of the changelogs that seemed like they might be interesting to have.

# Motivation

The miri job is up to 20 minutes, trying to bring it back down.

# Additional Notes

I do not know if we have access to these runners.

# How to test the change?

Nothing to do, this is a CI change only,.

### DIFF
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -4,7 +4,7 @@ on:
 
 jobs:
   run-miri:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-8-cores
     env:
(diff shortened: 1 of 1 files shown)


morrisonlevi:
I'm not sure if it's coincedence, but this took over 16 minutes to even get picked up. Even if it works, getting an 8x speedup but having to wait forever to get a runner is not a good tradeoff.

May investigate doing parallel jobs instead.

PR #968, opened 2025-03-25
Use cached default connector in hyper_migration.rs

bwoebi:
Rebuilding the connector all the time - and especially implicitly refetching root certificates - is very expensive. Use the cached connector.


morrisonlevi (commented):
Can we leave a comment here? If `::default()` has different behavior from `new`, this is a bit "unusual" and we should call that out here that we're specifically looking to cache the root certificates.

PR #979, opened 2025-03-26
Fix REDACTED_NAMES_INITIALIZED

bwoebi:
This flag is only to be set on the first usage of redacted names.
Regression introduced by b39c6ee1 (was noticed when integrating in dd-trace-php).


tylfin (approved):

ekump (approved):

ekump:
bypassing requirements and merging at @bwoebi request. FFI Alpine build check failure is flakiness and unrelated to PR.

PR #980, opened 2025-03-27
fix(trace-utils)!: serialize v0.5 span links and events into meta

paullegranddc:
# What does this PR do?

The v0.5 trace format is a fixed 12-element positional span array with no dedicated slots for span links, span events, or `meta_struct`. The v0.5 conversion previously dropped span links and span events. The Datadog agent/backend read them back from the `meta` field as JSON strings, so this PR serializes them there:

- **Span links** → `meta["_dd.span_links"]`, matching the agent's OTLP `transform.MarshalLinks`: a hex-encoded 128-bit `trace_id` (high 64 bits then low), hex-encoded 64-bit `span_id`, and optional `tracestate`/`attributes`. The v0.4 `flags` field is intentionally omitted to match the agent, the canonical producer of this key.
- **Span events** → `meta["events"]`, matching the agent's `MarshalEvents` (`{time_unix_nano, name, attributes}`, attributes rendered as natural JSON rather than the v0.4 msgpack tagged form).

Both keys are only emitted when non-empty.

[span_links fallback format spec](https://docs.google.com/document/d/1RcTjO8ER5FNweJ0rqVcQEbf6tb1PrlPsQ6coZZww8Xg/edit?tab=t.0#heading=h.pwna8rp9exnf)
[event fallback format spec](https://docs.google.com/document/d/1cVod_VI7Yruq8U9dfMRFJd7npDu-uBpste2IB04GyaQ/edit?tab=t.0#heading=h.78pxz0icbkx)

# Motivation

Ensure the v0.5 format carries the same span data as v0.4. Without these fallbacks span links and events are silently dropped on the v0.5 path.

# Additional Notes

Reworked on top of current `main`, which had refactored the span types to be generic over a `TraceData`/`SpanText` abstraction with a zero-copy borrowed variant.

Carrying links/events requires interning dynamically-built JSON strings, which a borrowed (`&str`) dictionary cannot own. The v0.5 shared dictionary therefore always owns its strings (`SharedDictBytes`), via a new `SpanText::to_bytes_string` conversion: borrowed input text is copied, owned text is reference-counted. Tradeoff: the zero-copy decode path now copies span text into the v0.5 dictionary; owned-`BytesString` callers are unaffected.

`meta_struct` still has no v0.5 representation (arbitrary binary blobs, no agent-side meta-key convention) and is documented as dropped; callers needing it must use the v0.4 output format.

# How to test the change?

```
cargo nextest run -p libdd-trace-utils -p libdd-data-pipeline -E '!test(tracing_integration_tests::)'
```

Unit tests cover: the `_dd.span_links`/`events` JSON shapes; 128-bit hex `trace_id` encoding; conditional key emission; `flags` omission; all scalar attribute types plus arrays and multi-attribute events; the `meta_struct` drop; multi-link and partial-field link serialization; and preserved dictionary interning order.

# Breaking changes

This is an API-breaking change (Rust API; libdatadog callers pin versions). The v0.5 shared dictionary is now always owned (`SharedDictBytes`) so it can intern dynamically-built JSON:
- `v05::from_v04_span` now takes `&mut SharedDictBytes` (was `&mut SharedDict<T::Text>`).
- `TraceChunks::V05` now holds `SharedDictBytes` (was `SharedDict<T::Text>`).


trace-utils/src/span/v05/mod.rs
```diff
@@ -29,24 +32,146 @@ pub struct Span {
...
+        HashMap::with_capacity(span.meta.len()),
+        |mut meta, (k, v)| -> anyhow::Result<HashMap<u32, u32>> {
+            meta.insert(dict.get_or_insert(k)?, dict.get_or_insert(v)?);
+            Ok(meta)
+        },
+    )?;
+    if !span.span_links.is_empty() {
+        let serialized_span_links = serde_json::to_string(&span.span_links)?;
+        meta.insert(
+            dict.get_or_insert(&tinybytes::BytesString::from("span_links"))?,
```
hoolioh, line 152: In the RFC there is some discussion about "span_links" or "_dd.span_links". Did they finally choose the former?

hoolioh (dismissed):
Couple of comments. Additionally it'd be great to add spanlinks and spanevents support to v05 snapshots in data-pipeline/tests/ to have them covered.

trace-utils/src/span/v05/mod.rs
paullegranddc: No, it's a mistake the key should definitely be "_dd.span_links" from other implementations

trace-utils/src/span/v05/mod.rs
```diff
@@ -29,24 +32,146 @@ pub struct Span {
...
+///
+/// It is meant to overrdide the default serialization, so we can serialize attributes
+/// differently from the original impl.
+/// Span events are serialized to JSON and added to "meta" when serializing to v0.5
+///
+/// The main difference with messagepacck serialization is that attributes with any types
+/// are supposed to be mapped to their natural JSON representation.
+///
+/// Sadly, I haven't found a good way of overriding the default Serialize behavior, other
+/// than just doing it for the whole data structures that embed it.
```
ganeshnj, line 45:
```suggestion
/// This structure is a wrapper around a slice of span events
///
/// It is meant to override the default serialization, so we can serialize attributes
/// differently from the original impl.
/// Span events are serialized to JSON and added to "meta" when serializing to v0.5
///
/// The main difference with messagepack serialization is that attributes with any types
/// are supposed to be mapped to their natural JSON representation.
```

you can skip the justification and personalization.

trace-utils/src/span/v05/mod.rs
```diff
@@ -82,8 +209,47 @@ mod tests {
             )]),
             metrics: HashMap::from([(BytesString::from("metrics_field"), 1.1)]),
             meta_struct: HashMap::new(),
-            span_links: vec![],
-            span_events: vec![],
+            span_links: vec![SpanLinkBytes {
```
ganeshnj, line 212: /suggestion

- would be nice to have test case for multiple span links
- a complex attributes (remember we do some flattening for array and maps), I'm not sure at this point we have them already or we have to do ourselves
   eg 
```
{"key": [[1,2], ["3", "4"]]} to {"key.0.0": "1", "key.0.1": 2, "key.1.0": 3, "key.1.1":4}
```

trace-utils/src/span/v05/mod.rs
```diff
@@ -29,24 +32,146 @@ pub struct Span {
...
+    let resource = dict.get_or_insert(&span.resource)?;
+    let mut meta = span.meta.iter().try_fold(
+        HashMap::with_capacity(span.meta.len()),
+        |mut meta, (k, v)| -> anyhow::Result<HashMap<u32, u32>> {
+            meta.insert(dict.get_or_insert(k)?, dict.get_or_insert(v)?);
+            Ok(meta)
+        },
+    )?;
+    if !span.span_links.is_empty() {
+        let serialized_span_links = serde_json::to_string(&span.span_links)?;
```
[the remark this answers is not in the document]
ganeshnj: Yes, anything beyond 25KB limits is dropped. the implementation should try to stuff as many as links possible under the limit and drop the others.

There used to be `dropped_attributes_count` to count the dropped attributes in case of oversized span links which i didn't see our [implementation](https://github.com/search?q=repo%3ADataDog%2Flibdatadog%20dropped_attributes_count&type=code).

ganeshnj (dismissed):
a few minor comments around tests and one major around tag size limit

overall looks good.

trace-utils/src/span/v05/mod.rs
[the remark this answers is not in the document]
bengl: Good question — I looked into it and concluded we shouldn't add a guard here. Reasoning:

The 25 KB cap (`traceutil.MaxMetaValLen = 25000`) lives agent-side in `truncator.go`, applied **uniformly to every meta value** as the final normalization step (`Truncate(span)` in `agent/agent.go`), regardless of which key or source produced it. The agent is the arbiter of this limit — it normalizes whatever it receives.

Given that, a check here:
- **couldn't improve the outcome** — truncating a JSON string to fit 25 KB produces invalid JSON (exactly what the agent's truncator does anyway), and dropping the key would instead diverge from what the agent does with an oversized value (keep it, truncated). Either way the agent re-applies its own 25 KB cap as its last step, so a client-side guard can't change the result;
- **would be inconsistent** — libdatadog caps no other meta value's length anywhere in the v0.4/v0.5 conversion; it defers meta-value length normalization to the agent. Guarding only these two keys would be the lone exception.

So I've left it to the agent's existing uniform truncation. Happy to revisit if you'd rather we drop oversized link/event blobs client-side (a deliberate divergence from the agent's normalization).

trace-utils/src/span/v05/mod.rs
bengl: It's the latter — `_dd.span_links`. Confirmed against the agent as the canonical consumer of this key. That's what this serializes to now.

trace-utils/src/span/v05/mod.rs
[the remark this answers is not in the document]
paullegranddc: Agreed on not doing truncation for now. That could be followup PR, with valid JSON truncation or we could just ignore that since v05 is going to be deprecated in the not so far future

paullegranddc:
Approved, but I cannot approve since this is technically my PR

paullegranddc:
There is some redundant code with the agentless endpoint here https://github.com/DataDog/libdatadog/blob/main/libdd-trace-utils/src/agentless_encoder/mod.rs#L322 but we can do the unification in a followup PR

Aaalibaba42 (approved):
Idk the protocol well, but rust side seems reasonable o/

iunanua (approved):
Second! 😆

libdd-trace-utils/src/span/v05/mod.rs
```diff
@@ -28,46 +32,265 @@ pub struct Span {
...
+            &span.span_events,
+        ))?)
+    };
+
+    let extra_meta =
+        serialized_span_links.is_some() as usize + serialized_span_events.is_some() as usize;
+
+    // Intern fields in the same order as the base conversion to keep dictionary indices
+    // stable; the span links / events keys are appended to `meta` afterwards.
+    let service = dict.get_or_insert(span.service.to_bytes_string())?;
```
ekump: I'm ok with doing this in a follow-up PR (let's make sure to create a Jira ticket if we do) - But should we have something like `SharedDIctBytes::get_or_insert_with()` that takes a `&str` instead so we don't do the ByteString allocation so frequently for what is most often going to be a "hit" and not an "insert" and defer the `to_bytes_string()` to only inserts?

libdd-trace-utils/src/span/v05/mod.rs
```diff
@@ -28,46 +32,265 @@ pub struct Span {
...
+                "attributes",
+                &SortedStrMapSerializerV05::<T>(&link.attributes),
+            )?;
+        }
+        map.end()
+    }
+}
+
+/// Serializes a `HashMap<T::Text, T::Text>` as a JSON object with keys in sorted order, matching
+/// Go's `encoding/json` (which the agent uses) and keeping the output deterministic.
```
ekump: nit: I don't the agent sorts, so perhaps drop the comment here.

paullegranddc: I think the sorting is necessary because of the test agent assertions.
Since rust hashmap order is not determinsitic, the snapshot comparison fails

libdd-trace-utils/src/span/v05/mod.rs
```diff
@@ -28,46 +32,265 @@ pub struct Span {
...
+/// expect under the `_dd.span_links` meta key.
+///
+/// This matches the agent's `transform.MarshalLinks`
+/// (datadog-agent `pkg/trace/transform/transform.go`), the canonical producer of
+/// `_dd.span_links`:
+/// - `trace_id` is the full 128-bit id hex-encoded as 32 lowercase chars (high 64 bits first, then
+///   low 64 bits).
+/// - `span_id` is the 64-bit id hex-encoded as 16 lowercase chars.
+/// - `tracestate` and `attributes` are only emitted when non-empty.
+/// - The v0.4 `flags` field is intentionally not emitted: it is not part of the `_dd.span_links`
```
ekump: I'm slightly confused by this comment and behavior. The RFC states that flags should only be omitted if not set. Is the RFC wrong?

libdd-trace-utils/src/span/v05/mod.rs
ekump: I'm saying remove the comment claiming parity with the agent, not remove the sorting.

libdd-trace-utils/src/span/v05/mod.rs
paullegranddc: This is a topic we have been discussing with @VianneyRuhlmann for a while. Taking references to the v04 span string might not be possible in all cases as it requires keeping the v04 spans alive and pinned to the stack.

The solution would probably to add a `from_string` function to the SpanText trait, which would allow both:
* Create new dynamic SpanText strings for cases like this where we modify meta attributes in the exporter
* not having to convert ByteString and instead taking ownership of existing Spantext objects

but that requires non trivial changes to how we implement the trait in python (because we cannot python string in rust because no GIL).
I think a Jira ticket with refactoring of this after we implement dynamic SpanText creation in the exporter situation is the best course of action

ekump (approved):
I have a question around dropping flags, but LGTM

libdd-trace-utils/src/span/v05/mod.rs
paullegranddc: I'm not sure what my original intention was.
It looks like the agent does not emit flags but that it's a bug in the agent, not an actual thing 🤔

paullegranddc: added flags serialization

libdd-trace-utils/src/span/v05/mod.rs
paullegranddc: Yes of course. I updated the comment

PR #985, opened 2025-03-28
feat: Expose ddcommon memfd_create

anna-git:
# What does this PR do?

In dotnet we're gonna need to expose the FFI 

Related to 
https://datadoghq.atlassian.net/browse/APMAPI-1072
https://github.com/DataDog/libdatadog/pull/867

# Motivation

# Additional Notes

# How to test the change?

Integrated in the dotnet tracer locally, it's working for me under WSL, I can see and decode the linux file descriptor.

### DIFF
--- a/builder/Cargo.toml
+++ b/builder/Cargo.toml
@@ -14,6 +14,7 @@ telemetry = []
 data-pipeline = []
 symbolizer = []
 library-config = []
+ddcommon = []
 
 [lib]
 bench = false
--- a/builder/Cargo.toml
+++ b/builder/Cargo.toml
@@ -14,7 +14,6 @@ telemetry = []
 data-pipeline = []
 symbolizer = []
 library-config = []
-ddcommon = []
 
 [lib]
 bench = false
--- a/ddcommon-ffi/src/tracer_metadata.rs
+++ b/ddcommon-ffi/src/tracer_metadata.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::slice::CharSlice;
+use crate::Result;
 #[cfg(target_os = "linux")]
 use ddcommon::tracer_metadata::AnonymousFileHandle;
 use ddcommon::tracer_metadata::{self, TracerMetadata};
@@ -12,8 +13,6 @@ use std::os::raw::c_int;
 pub struct TracerMemfdHandle {
     /// File descriptor (relevant only on Linux)
     pub fd: c_int,
-    /// Whether the handle is valid
-    pub is_valid: bool,
 }
 
 /// Store tracer metadata to a file handle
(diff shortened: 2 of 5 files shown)


ddcommon-ffi/src/tracer_metadata.rs
```diff
@@ -0,0 +1,85 @@
...
+use ddcommon::tracer_metadata::{self, TracerMetadata};
+use std::os::raw::c_int;
+
+/// C-compatible representation of an anonymous file handle
+#[repr(C)]
+pub struct TracerMemfdHandle {
+    /// File descriptor (relevant only on Linux)
+    pub fd: c_int,
+    /// Whether the handle is valid
+    pub is_valid: bool,
```
bwoebi, line 16: Do we actually need that, or can we just return `c_int` and check for -1 in code?

ddcommon-ffi/src/tracer_metadata.rs
```diff
@@ -0,0 +1,85 @@
...
+        },
+        service_version: if service_version.is_empty() {
+            None
+        } else {
+            Some(service_version.to_string())
+        },
+    };
+
+    // Call the actual implementation
+    match tracer_metadata::store_tracer_metadata(&metadata) {
```
ganeshnj, line 64: This should go to `ddcommon`, ideally `-ffi` crates should only contain FFI related conversions and call to real implementations.

bwoebi: I don't understand the concern? All this function is doing, is collecting the args into the rust struct, forwarding the real implementation and returning it in a way compatible with C API?

builder/Cargo.toml
```diff
@@ -14,6 +14,7 @@ telemetry = []
 data-pipeline = []
 symbolizer = []
 library-config = []
+ddcommon = []
```
hoolioh, line 17: No need to add ddcommon to the features list because ddcommon module is always added to the module list.

ddcommon-ffi/src/tracer_metadata.rs
ganeshnj: my bad - I just realized `TracerMemfdHandle` is a FFI type.

ddcommon-ffi/src/tracer_metadata.rs
anna-git: Good point, I removed it, instead I'm now wrapping [the result here](https://github.com/DataDog/libdatadog/pull/985/commits/df8d0ae1b9cf4d7b69a72b3bb5ad54d8174ecc2f#diff-37c0cf3172b52d455f9826a760faa521726c026954021d9763ec85958c33a0f2) to get the error message on the tracer side and be able to log it, let me know wyt thanks

builder/Cargo.toml
anna-git: removed in https://github.com/DataDog/libdatadog/pull/985/commits/df8d0ae1b9cf4d7b69a72b3bb5ad54d8174ecc2f

ddcommon-ffi/src/tracer_metadata.rs
bwoebi: Sounds good :-)

ganeshnj (approved):

hoolioh (approved):

PR #989, opened 2025-03-31
fix(ddcommon_connector): the crypto provider also needs to be initialized with webpkis disabled

paullegranddc:
# What does this PR do?

*Initialise the crypto provider even if without the `use_webpki_roots` feature

# Motivation

This has fallen though the crask of testing in the CI because running tests on the full workspace enables `use_webpki_roots` though feature unification. But depending only on data-pipeline makes it break.

From bisecting these tests staring being broken in #983, probably because the previous version of bollard we depended on for creating test containers was creating the crypto provider for us during testing...

### DIFF
--- a/ddcommon/src/connector/mod.rs
+++ b/ddcommon/src/connector/mod.rs
@@ -92,7 +92,6 @@ mod https {
 
     use rustls::ClientConfig;
 
-    #[cfg(feature = "use_webpki_roots")]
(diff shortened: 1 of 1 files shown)


ivoanjo:
Do you know if this affects the binaries in https://github.com/DataDog/libdatadog/releases/tag/v17.0.0 ? I was planning on pulling that one into Ruby soon and I'm curious if I can still use or I'll need to wait for this PR to go in.

PR #999, opened 2025-04-03
chore: fix crashtracker segfault tests for 1.86.0

sanchda:
# What does this PR do?

Rust 1.86 changed code emission so more operations are handled safely.  Unfortunately, this made it a little harder to trigger true segmentation faults within the language.

This PR uses some inline assembly to make this situation occur on supported platforms.

# Motivation

Make tests more correct

# Additional Notes

# How to test the change?

### DIFF
--- a/bin_tests/src/bin/crashtracker_bin_test.rs
+++ b/bin_tests/src/bin/crashtracker_bin_test.rs
@@ -28,10 +28,23 @@ mod unix {
     const TEST_COLLECTOR_TIMEOUT_MS: u32 = 10_000;
 
     #[inline(never)]
-    unsafe fn deref_ptr(p: *mut u8) -> u8 {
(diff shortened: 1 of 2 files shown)


taegyunkim (approved):
thank you for quickly updating this!

ekump (approved):

morrisonlevi:
Seems the assembly for ARM gets rejected:

```
error: invalid operand for instruction
  --> bin_tests/src/bin/crashtracker_bin_test.rs:39:30
   |
39 |             std::arch::asm!("ldr x0, [xzr]", options(nostack));
   |                              ^
   |
note: instantiated into assembly here
  --> <inline asm>:1:11
   |
1  |     ldr x0, [xzr]
   |              ^
error: could not compile `bin_tests` (bin "crashtracker_bin_test") due to 1 previous error
```

PR #1003, opened 2025-04-04
fix: fix new stable clippy lints

paullegranddc:
# What does this PR do?

New clippy lints have enabled on rust 1.86
and this breaks in CI.
I ran `cargo +stable clippy --fix` to fix the lints

# Motivation

# Additional Notes

# How to test the change?


paullegranddc:
Already fixed on main

PR #1011, opened 2025-04-09
chore: upgrade hashbrown from 0.14 to 0.15

morrisonlevi:
# What does this PR do?

Updates the version of hashbrown from 0.14 to 0.15. To accomplish this, I needed to update other packages.

# Motivation

I thought it would be a quick and easy thing to prune down a duplicate dependency. I was wrong (see additional notes).

# Additional Notes

Unfortunately, simd-json requires Rust 1.85. I hate that you only learn this kind of thing at the very end 🙃

# How to test the change?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1285,19 +1285,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dashmap"
-version = "5.5.3"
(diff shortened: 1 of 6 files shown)


morrisonlevi:
Closing since it requires Rust 1.85.

PR #1018, opened 2025-04-11
feat(data-pipeline): logging ffi

ganeshnj:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1324,7 +1324,6 @@ dependencies = [
  "httpmock",
  "hyper 1.6.0",
  "hyper-util",
- "log",
  "rand 0.8.5",
  "regex",
  "rmp-serde",
@@ -1334,6 +1333,7 @@ dependencies = [
  "tinybytes",
  "tokio",
  "tokio-util",
+ "tracing",
  "uuid",
 ]
 
@@ -1401,7 +1401,6 @@ dependencies = [
  "ddcommon-ffi",
  "function_name",
  "libc",
- "log",
  "serde",
  "serde_json",
  "symbolic-common",
@@ -1831,7 +1830,6 @@ dependencies = [
  "httpmock",
  "hyper 1.6.0",
  "hyper-http-proxy",
- "log",
  "prost 0.11.9",
  "rand 0.8.5",
  "rmp",
@@ -1843,6 +1841,7 @@ dependencies = [
  "testcontainers",
  "tinybytes",
  "tokio",
+ "tracing",
  "urlencoding",
  "zstd",
 ]
@@ -1866,7 +1865,6 @@ dependencies = [
  "hyper-util",
  "indexmap 2.6.0",
  "libc",
- "log",
  "maplit",
  "memfd",
  "nix",
--- a/crashtracker-ffi/Cargo.toml
+++ b/crashtracker-ffi/Cargo.toml
@@ -45,7 +45,6 @@ function_name = "0.3.0"
 libc = "0.2.167"
 serde_json = "1.0.132"
 serde = { version = "1.0.214", features = ["derive"] }
-log = "0.4.22"
 
 [target.'cfg(windows)'.dependencies]
 windows = { version = "0.59.0", features = ["Win32_System_Diagnostics_Debug", "Win32_System_ErrorReporting"] }
--- a/data-pipeline/Cargo.toml
+++ b/data-pipeline/Cargo.toml
@@ -16,7 +16,7 @@ hyper = { version = "1.6", features = ["http1", "client"] }
 hyper-util = { version = "0.1", features = ["client", "client-legacy"] }
 http = "1.0"
 http-body-util = "0.1"
-log = "0.4"
+tracing = { version = "0.1", default-features = false }
 rmp-serde = "1.1.1"
 serde = "1.0.209"
 serde_json = "1.0.127"
--- a/data-pipeline/src/agent_info/fetcher.rs
+++ b/data-pipeline/src/agent_info/fetcher.rs
@@ -10,10 +10,10 @@ use ddcommon::hyper_migration;
 use ddcommon::Endpoint;
 use http_body_util::BodyExt;
 use hyper::{self, body::Buf, header::HeaderName};
-use log::{error, info};
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::time::sleep;
+use tracing::{error, info};
 
 #[allow(clippy::declare_interior_mutable_const)]
 const DATADOG_AGENT_STATE: HeaderName = HeaderName::from_static("datadog-agent-state");
@@ -157,7 +157,7 @@ impl AgentInfoFetcher {
                     info!("Agent info is up-to-date")
                 }
                 Err(err) => {
-                    error!("Error while fetching /info: {}", err);
+                    error!(?err, "Error while fetching /info");
                 }
             }
             sleep(self.refresh_interval).await;
--- a/data-pipeline/src/stats_exporter.rs
+++ b/data-pipeline/src/stats_exporter.rs
@@ -11,15 +11,14 @@ use std::{
     time,
 };
 
+use crate::{span_concentrator::SpanConcentrator, trace_exporter::TracerMetadata};
 use datadog_trace_protobuf::pb;
 use datadog_trace_utils::send_with_retry::{send_with_retry, RetryStrategy};
 use ddcommon::Endpoint;
 use hyper;
-use log::error;
 use tokio::select;
 use tokio_util::sync::CancellationToken;
-
-use crate::{span_concentrator::SpanConcentrator, trace_exporter::TracerMetadata};
+use tracing::error;
 
 const STATS_ENDPOINT_PATH: &str = "/v0.6/stats";
 
@@ -101,7 +100,7 @@ impl StatsExporter {
         match result {
             Ok(_) => Ok(()),
             Err(err) => {
-                error!("Error with the StateExporter when sending: {err}");
+                error!(?err, "Error with the StateExporter when sending stats");
                 anyhow::bail!("Failed to send stats: {err}");
             }
         }
--- a/data-pipeline/src/trace_exporter/mod.rs
+++ b/data-pipeline/src/trace_exporter/mod.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 pub mod agent_response;
 pub mod error;
+use self::agent_response::AgentResponse;
 use crate::agent_info::{AgentInfoArc, AgentInfoFetcher};
 use crate::telemetry::{SendPayloadTelemetry, TelemetryClient, TelemetryClientBuilder};
 use crate::trace_exporter::error::{RequestError, TraceExporterError};
@@ -26,15 +27,13 @@ use error::BuilderErrorKind;
 use http_body_util::BodyExt;
 use hyper::http::uri::PathAndQuery;
 use hyper::{header::CONTENT_TYPE, Method, Uri};
-use log::{error, info};
 use std::io;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use std::{borrow::Borrow, collections::HashMap, str::FromStr, time};
 use tokio::{runtime::Runtime, task::JoinHandle};
 use tokio_util::sync::CancellationToken;
-
-use self::agent_response::AgentResponse;
+use tracing::{error, info};
 
 const DEFAULT_STATS_ELIGIBLE_SPAN_KINDS: [&str; 4] = ["client", "server", "producer", "consumer"];
 const STATS_ENDPOINT: &str = "/v0.6/stats";
@@ -249,7 +248,10 @@ impl TraceExporter {
         })
         .map_err(|err| {
             if let TraceExporterError::Deserialization(ref e) = err {
-                error!("Error deserializing trace from request body: {e}");
+                error!(
+                    err = e.to_string(),
+                    "Error deserializing trace from request body"
+                );
                 self.emit_metric(
                     HealthMetric::Count(health_metrics::STAT_DESER_TRACES_ERRORS, 1),
                     None,
@@ -500,7 +502,10 @@ impl TraceExporter {
                                 // This should really never happen as response_status is a
                                 // `NonZeroU16`, but if the response status or tag requirements
                                 // ever change in the future we still don't want to panic.
-                                error!("Failed to serialize response_code to tag {}", tag_err)
+                                error!(
+                                    err = tag_err.to_string(),
+                                    "Failed to serialize response_code to tag"
+                                )
                             }
                         }
                         return Err(TraceExporterError::Request(RequestError::new(
@@ -690,7 +695,7 @@ impl TraceExporter {
                     payload_len as u64,
                     chunks as u64,
                 )) {
-                    error!("Error sending telemetry: {}", e.to_string());
+                    error!(?e, "Error sending telemetry");
                 }
             }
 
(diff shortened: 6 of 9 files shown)


r1viollet:
# Artifact Size Benchmark Report
<details >
<summary>aarch64-alpine-linux-musl</summary>

| Artifact | Baseline | Commit | Change |
|----------|-----------|-------------|------------|
| /aarch64-alpine-linux-musl/lib/libdatadog_profiling.so|7.78 MB|7.79 MB|+.01% (+1.57 KB) 🔍 |
| /aarch64-alpine-linux-musl/lib/libdatadog_profiling.a|78.03 MB|78.04 MB|+.01% (+12.04 KB) 🔍 |
| /aarch64-alpine-linux-musl/lib/libdatadog_profiling.so.debug|24.21 MB|24.21 MB|+.02% (+5.91 KB) 🔍 |
</details>
<details >
<summary>aarch64-unknown-linux-gnu</summary>

| Artifact | Baseline | Commit | Change |
|----------|-----------|-------------|------------|
| /aarch64-unknown-linux-gnu/lib/libdatadog_profiling.so.debug|22.82 MB|22.82 MB|+.01% (+2.79 KB) 🔍 |
| /aarch64-unknown-linux-gnu/lib/libdatadog_profiling.so|7.72 MB|7.72 MB|+.01% (+968 B) 🔍 |
| /aarch64-unknown-linux-gnu/lib/libdatadog_profiling.a|72.41 MB|72.42 MB|+.01% (+8.39 KB) 🔍 |
</details>
<details >
<summary>libdatadog-x64-windows</summary>

| Artifact | Baseline | Commit | Change |
|----------|-----------|-------------|------------|
| /libdatadog-x64-windows/debug/dynamic/datadog_profiling_ffi.dll|17.01 MB|17.05 MB|+.20% (+36.00 KB) 🔍 |
| /libdatadog-x64-windows/debug/dynamic/datadog_profiling_ffi.lib|61.83 KB|61.83 KB|0% (0 B) 👌 |
| /libdatadog-x64-windows/debug/dynamic/datadog_profiling_ffi.pdb|114.25 MB|114.40 MB|+.12% (+152.00 KB) 🔍 |
| /libdatadog-x64-windows/debug/static/datadog_profiling_ffi.lib|633.71 MB|634.97 MB|+.19% (+1.26 MB) 🔍 |
| /libdatadog-x64-windows/release/dynamic/datadog_profiling_ffi.dll|5.06 MB|5.06 MB|+.01% (+1.00 KB) 🔍 |
| /libdatadog-x64-windows/release/dynamic/datadog_profiling_ffi.lib|61.83 KB|61.83 KB|0% (0 B) 👌 |
| /libdatadog-x64-windows/release/dynamic/datadog_profiling_ffi.pdb|16.15 MB|16.15 MB|0% (0 B) 👌 |
| /libdatadog-x64-windows/release/static/datadog_profiling_ffi.lib|26.84 MB|26.84 MB|+0% (+450 B) 👌 |
</details>
<details >
<summary>libdatadog-x86-windows</summary>

| Artifact | Baseline | Commit | Change |
|----------|-----------|-------------|------------|
| /libdatadog-x86-windows/debug/dynamic/datadog_profiling_ffi.dll|14.40 MB|14.43 MB|+.21% (+31.00 KB) 🔍 |
| /libdatadog-x86-windows/debug/dynamic/datadog_profiling_ffi.lib|62.78 KB|62.78 KB|0% (0 B) 👌 |
| /libdatadog-x86-windows/debug/dynamic/datadog_profiling_ffi.pdb|116.21 MB|116.38 MB|+.14% (+176.00 KB) 🔍 |
| /libdatadog-x86-windows/debug/static/datadog_profiling_ffi.lib|625.80 MB|626.84 MB|+.16% (+1.04 MB) 🔍 |
| /libdatadog-x86-windows/release/dynamic/datadog_profiling_ffi.dll|3.83 MB|3.83 MB|+.03% (+1.50 KB) 🔍 |
| /libdatadog-x86-windows/release/dynamic/datadog_profiling_ffi.lib|62.78 KB|62.78 KB|0% (0 B) 👌 |
| /libdatadog-x86-windows/release/dynamic/datadog_profiling_ffi.pdb|16.78 MB|16.78 MB|0% (0 B) 👌 |
| /libdatadog-x86-windows/release/static/datadog_profiling_ffi.lib|24.74 MB|24.74 MB|+0% (+236 B) 👌 |
</details>
<details >
<summary>x86_64-alpine-linux-musl</summary>

| Artifact | Baseline | Commit | Change |
|----------|-----------|-------------|------------|
| /x86_64-alpine-linux-musl/lib/libdatadog_profiling.a|67.28 MB|67.29 MB|+.01% (+7.28 KB) 🔍 |
| /x86_64-alpine-linux-musl/lib/libdatadog_profiling.so|8.27 MB|8.27 MB|+.01% (+872 B) 🔍 |
| /x86_64-alpine-linux-musl/lib/libdatadog_profiling.so.debug|23.35 MB|23.35 MB|+0% (+2.22 KB) 👌 |
</details>
<details >
<summary>x86_64-unknown-linux-gnu</summary>

| Artifact | Baseline | Commit | Change |
|----------|-----------|-------------|------------|
| /x86_64-unknown-linux-gnu/lib/libdatadog_profiling.a|68.23 MB|68.23 MB|+0% (+2.09 KB) 👌 |
| /x86_64-unknown-linux-gnu/lib/libdatadog_profiling.so|8.14 MB|8.14 MB|+0% (+616 B) 👌 |
| /x86_64-unknown-linux-gnu/lib/libdatadog_profiling.so.debug|20.98 MB|20.98 MB|+0% (+304 B) 👌 |
</details>

morrisonlevi:
Prefixes I know are already in use:
 - `ddog_prof_*`
 - `ddog_crasht_*`
 - `ddog_trace_exporter_*
 
 Can we name this `ddog_log_*` instead?

ganeshnj:
> The symbol prefixes I know are already in use are:
> 
> * `ddog_prof_*`
> * `ddog_crasht_*`
> * `ddog_trace_exporter_*`
> * `ddog_sidecar_*`
> * `ddog_telemetry_*`
> 
> Can we name this `ddog_log_*` instead of `ddlog_*`?

actually this threw me off a few times, currently we have all sort of prefixes depending on the context (crate name vs dir name vs rust type name vs ffi method names)

 do you mean anything specific like FFI methods or in general?

I agree that atleast ffi methods should be `ddog_log_` since this is only way to namespace them.

morrisonlevi:
There are conventions for the full name too, but they are less standardized. I was specifically talking about the FFI symbol prefixes with this comment.

log/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
```
ekump, line 3: You probably want to include the clippy attributes for warning on panic macros

```suggestion
#![cfg_attr(not(test), deny(clippy::panic))]
#![cfg_attr(not(test), deny(clippy::unwrap_used))]
#![cfg_attr(not(test), deny(clippy::expect_used))]
#![cfg_attr(not(test), deny(clippy::todo))]
#![cfg_attr(not(test), deny(clippy::unimplemented))]

```

log-ffi/src/lib.rs
```diff
@@ -0,0 +1,46 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
```
ekump, line 1: Do you think there is any value in adding something for log-ffi to [examples](https://github.com/DataDog/libdatadog/tree/64fd851081732a8849dcc17dc38298d5ef419f1f/examples/ffi)?

log-ffi/src/lib.rs
```diff
@@ -16,7 +16,7 @@ use ddlog::logger::{logger_init, logger_set_log_level, LogCallback, LogLevel};
 ///
 /// Returns `None` on success, or a boxed `Error` if the update fails.
 #[no_mangle]
-pub extern "C" fn ddog_logger_set_log_level(log_level: LogLevel) -> Option<Box<Error>> {
+pub extern "C" fn ddog_log_logger_set_log_level(log_level: LogLevel) -> Option<Box<Error>> {
```
morrisonlevi, line 19: Since this does not take a "logger" object, you could probably simplify the name to `ddog_log_set_log_level`. Same with `ddog_log_logger_init` to `ddog_log_init`.

log-ffi/src/lib.rs
ganeshnj: good call, a dumb example would be good for people understand the usage.

log-ffi/src/lib.rs
ganeshnj: fair enough.

log-ffi/src/lib.rs
ganeshnj: I didn't add its own example but updated the trace exporter which actually emits some logs
```
➜ ./examples/ffi/build/trace_exporter                                                                                               
ERROR :: Error deserializing trace from request body :: err: "Invalid format: Unable to read array len for trace count"
Operation failed with error: 18, reason: Invalid format: Unable to read array len for trace count
TRACE :: checkout waiting for idle connection: ("http", localhost:8126) :: 
TRACE :: Http::connect; scheme=Some("http"), host=Some("localhost"), port=Some(Port(8126)) :: 
TRACE :: checkout dropped for ("http", localhost:8126) :: 
ERROR :: Error while fetching /info :: err: client error (Connect)
```

docs/RFCs/0009-cross-language-logging-integration.md
```diff
@@ -0,0 +1,210 @@
+# RFC: Cross-Language Logging Integration for libdatadog
```
ekump, line 1: Given that we may want to clarify the RFC a bit, do you maybe want to add this in a separate PR? (Or add it now and change it later, up to you)

datadog-log/src/logger.rs
```diff
@@ -0,0 +1,361 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::writers::{FileWriter, NoopWriter, StdoutWriter};
+use ddcommon_ffi::Error;
+use once_cell::sync::Lazy;
```
hoolioh, line 6: Since we're not currently covering no_std scenarios wouldn't be better to use std::sync::OnceCell. It basically provide the same functionality and we'll remove a dependency.

datadog-log-ffi/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+
+/// Configures the global logger.
+///
+/// # Arguments
+/// * `config` - Configuration for the logger including level and output destination
+///
+/// # Errors
+/// Returns an error if the logger cannot be configured.
+#[no_mangle]
+pub extern "C" fn ddog_logger_configure(config: LoggerConfig) -> Option<Box<Error>> {
```
hoolioh, line 30: Just wondering if it'd be worth hiding the LoggerConfig and just provide a couple of simple functions to just log to fd(stdout)? Do we plan to support many more methods?

ganeshnj: The logger config will have more things such as rate limiting interval, so ideally we want to have a config type for better forward compatibility. Thinking on having each logging type their own methods we have to do something like
- `ddog_logger_disable()`
- `ddog_logger_configure_std(config: StdConfig)`
- `ddog_logger_configure_file(config: FileConfig)`

tbh I'm fine with either since these all have similar limitations.

What I would definitely avoid is passing n-different params for each config which is going to be a bad API from compatibility point of view.

datadog-log/src/logger.rs
ganeshnj: fair call, I was under impression that we are already using `once_cell` as dependency.

datadog-log-ffi/src/lib.rs
hoolioh: I'm in favor of following a constructor pattern or break down the underlying logic into smaller functions. In this case I like the approach you suggested, that way we can isolate the inner workings of the configuration and there is less cognitive load for the library user. I agree that a struct provides better encapsulation and forward compatibility.

datadog-log-ffi/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+
+/// Sets the global log level.
+///
+/// # Arguments
+/// * `log_level` - The minimum level for events to be logged
+///
+/// # Errors
+/// Returns an error if the log level cannot be set.
+#[no_mangle]
+pub extern "C" fn ddog_logger_set_log_level(
```
hoolioh, line 16: I would check that log_level is between the proper range and return an error accordingly. C++ compilers will render an error if the value passed is not within the range of the enum but, unfortunately, with gcc is perfectly possible pass any value as long is an integer.

ganeshnj: I consider it a developer error if out of range values are passed, but I have still added error handling

datadog-log/src/logger.rs
```diff
@@ -0,0 +1,363 @@
...
+            .json();
+
+        let filter = LevelFilter::from_level(level);
+        let (filter_layer, reload_handle) = reload::Layer::new(filter);
+
+        let fmt_layer = fmt::Layer::default()
+            .event_format(format)
+            .with_writer(dynamic_writer.clone());
+
+        tracing_subscriber::registry()
```
ekump, line 163: What, if anything will happen when libdatadog crates are used in other crates, like `dd-trace-rs`? I assume if the other crate registers a subscriber then it should just work? No need to use this logger crate?

docs/RFCs/0010-logging-in-libdatadog.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Overview
+
+**Document Purpose:**
+
+This document outlines the design for adding logging sink capabilities to libdatadog. Currently, logs generated within libdatadog are not visible to the calling systems, making it difficult for Datadog to diagnose issues in customer environments. This document proposes implementing logging directly within libdatadog, with support for multiple output destinations including stdout and file-based output. These logs are intended solely for Datadog's internal troubleshooting purposes and are not meant to be used by customers for their application logging needs.
+
+**Background**
+
+The libdatadog `TraceExporter` is responsible for exporting traces to the Datadog agent. While processing and exporting traces, it generates important diagnostic logs which are currently not visible in production environments. This makes the `TraceExporter` integration backward incompatible with existing Trace SDKs.
```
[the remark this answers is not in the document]
ganeshnj: it refers to `dd-trace-*`

>Why is it not possible to convey information through return values from TraceExporter function calls to dd-trace-* libraries?

This is not possible for all cases, for errors it is fine but logging isn't just about errors. Although, we try best to return as much as possible.

>we'd log to customer environment but don't want them to be sent to us, right?
And we'll ask them to look for these logs when they come to us for escalations?

that's correct, things like tracer flare will help. Also, we can get some automatic telemetry from error returned from direct method calls which we can potentially get through `dd-trace-*` via error telemetry but that's out of scope of this document.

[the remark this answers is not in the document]
taegyunkim: I guess from the Python side, we'd just be calling the logger API using the Rust bindings to use the `TraceExporter` there. Since I'm not that familiar with how tracer works in general, as long as the logged output is identical, I can't really think of problems that would potentially caused by doing so.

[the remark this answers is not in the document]
taegyunkim: could there be any race condition accessing stderr/stdout from this logger and other parts of `dd-trace-*`? or that's out of scope for this PR/RFC?

taegyunkim (approved):

PR #1072, opened 2025-05-27
chore: move memfd config to datadog-library-config

dmehala:
# What does this PR do?

Move shared in-memory tracer configuration from `ddcommon` to `datadog-library-config`.

# Motivation

It's better suited than `ddcommon`.


bantonsson (approved):

ivoanjo:
> @dmehala I actually don't know what those checks do, but I bet @ivoanjo can tell you if they are related.

Thanks for the patience with the Ruby noise, it should be fixed now, let me know if it shows up again.

PR #1081, opened 2025-05-29
Flatten tarpc workspace

hoolioh:
# What does this PR do?

This PR removes tarpc nested workspace by flattening the crates used in datadog-ipc.

# Motivation

Cargo resolver does not work very well in cases where there are nested workspaces. This leads to problems when integrating libdatadog from source.


ivoanjo (approved):
:+1: There's a lot of diff noise but it's actually a very straightforward small move.

PR #1085, opened 2025-06-03
[APMSP-1976] Add check of parsed config and return action needed for tracer flare

anais-raison:
# What does this PR do?

This PR add a function that check the config received by the listener of RemoteConfig. This way it returns, through an enum, which action have to be done by the tracer flare depending of the config.

# Motivation

# Additional Notes

Also I updated the CODEOWNER file because of the crate `datadog-dynamic-configuration` was remove in my previous PR.

# How to test the change?

### DIFF
--- a/datadog-remote-config/src/config/agent_config.rs
+++ b/datadog-remote-config/src/config/agent_config.rs
@@ -1,7 +1,9 @@
 // Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
-use serde::{Deserialize, Serialize};
+use serde::Deserialize;
+#[cfg(feature = "test")]
+use serde::Serialize;
 
 #[derive(Debug, Deserialize)]
 #[cfg_attr(feature = "test", derive(Serialize))]
--- a/datadog-remote-config/src/config/agent_task.rs
+++ b/datadog-remote-config/src/config/agent_task.rs
@@ -1,7 +1,9 @@
 // Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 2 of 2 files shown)


datadog-tracer-flare/src/lib.rs
```diff
@@ -23,59 +23,92 @@ pub enum FlareError {
...
+    fn from(level: &String) -> Self {
+        match level.as_str() {
+            "trace" => ReturnAction::StartTrace,
+            "debug" => ReturnAction::StartDebug,
+            "info" => ReturnAction::StartInfo,
+            "warn" => ReturnAction::StartWarn,
+            "error" => ReturnAction::StartError,
+            "critical" => ReturnAction::StartCritical,
+            "off" => ReturnAction::StartOff,
+            _ => ReturnAction::None,
```
ekump, line 65: If we get an unknown log level is just setting it to None the appropriate action? Or should it be an error?

datadog-tracer-flare/src/lib.rs
```diff
@@ -23,59 +23,92 @@ pub enum FlareError {
...
         }
     }
 }
 
+pub type RemoteConfigFile = std::sync::Arc<RawFile<Result<RemoteConfigData, anyhow::Error>>>;
```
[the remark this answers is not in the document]
anais-raison: It's the type of something returned by the `remote-config` crate so I cannot use the `FlareError` here but if I catch an error, I convert it to a `FlareError` right away.

datadog-tracer-flare/src/lib.rs
anais-raison: I think an error would be the most appropriate. But it could be nice to launch the flare still with a default level.

hoolioh (commented):
I would add the crate to the .codecov file so we can the code coverage report.

ekump (approved):

PR #1089, opened 2025-06-05
fix: correct metric names for chunks sent and dropped in telemetry

ganeshnj:
# What does this PR do?

Matches the metric names as we should.

# Motivation

Some alarms went off due to invalid names of metric events.

# Additional Notes

Our metric names must always match https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-metrics/static/common_metrics.json

# How to test the change?

Tests updated

### DIFF
--- a/data-pipeline/src/telemetry/metrics.rs
+++ b/data-pipeline/src/telemetry/metrics.rs
@@ -23,18 +23,20 @@ pub enum MetricKind {
     ApiBytes,
     /// trace_api.responses metric
     ApiResponses,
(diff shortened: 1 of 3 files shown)


andrewlock (approved):
I'd recommend trying to integrate automated testing against [the allow list](https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-metrics/static/common_metrics.json) but obviously this solves our immediate problem, thanks! 😄

PR #1091, opened 2025-06-05
[Build] Remove unused build-<config/profiling/telemetry>.sh scripts

danielsn:
# What does this PR do?

Removes these scripts since we shouldn't use them anymore.

# Motivation

We have a better way of building FFI now.

# Additional Notes

# How to test the change?


gleocadie:
`build-profiling-ffi.sh` is still used https://github.com/DataDog/libddprof-build/blob/main/.gitlab-ci.yml#L211

taegyunkim (approved):

paullegranddc (approved):
Ok config and telemetry

PR #1005, opened 2025-04-04
[crashtracker] Don't send the same stack trace twice

danielsn:
# What does this PR do?

Currently we send the stacktrace in the log message, and again as part of the `crash_info` payload.  This wastes bytes for no reason.  Instead, just send it in the crash_info payload.

# Motivation

Lets not waste bytes.

# Additional Notes

This will likely require synchronization with the backend.  

# How to test the change?

Describe here in detail how the change can be validated.

PR #1093, opened 2025-06-06
[Crashtracker][bug]Add `is_crash`  field to Log message

gleocadie:
# What does this PR do?

Add `is_crash` to Log message

# Motivation

In this [PR](https://github.com/DataDog/libdatadog/pull/1005)  we do not fill the `Stacktrace` field of the log message to reduce the size of the message.
Except that in the backend, we use that field to recognize crash log and symbolize them.

In the backend code, I saw that the `Log` struct has a `IsCrash/is_crash` field.https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/logs.go#L65

### DIFF
--- a/datadog-crashtracker/src/crash_info/telemetry.rs
+++ b/datadog-crashtracker/src/crash_info/telemetry.rs
@@ -135,6 +135,7 @@ impl TelemetryCrashUploader {
                 tags,
                 is_sensitive: true,
                 count: 1,
+                is_crash: true,
             }]),
             origin: Some("Crashtracker"),
         };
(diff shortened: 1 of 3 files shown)


hoolioh (approved):

datadog-crashtracker/src/crash_info/telemetry.rs
```diff
@@ -268,6 +269,7 @@ mod tests {
         assert_eq!(payload["application"]["service_version"], "bar");
         assert_eq!(payload["request_type"], "logs");
         assert_eq!(payload["tracer_time"], 1568898000);
+        assert_eq!(payload["origin"], "Crashtracker");
```
danielsn, line 272: good catch

PR #1097, opened 2025-06-09
feat(logging): enhance logging in trace exporter

ganeshnj:
# What does this PR do?

Enhances logging in trace exporter.

# Motivation

We want better visibility on failures cases.

# Additional Notes

None

# How to test the change?

Usual logging macro usage, nothing to test except logs are flowing.

### DIFF
--- a/data-pipeline/src/trace_exporter/mod.rs
+++ b/data-pipeline/src/trace_exporter/mod.rs
@@ -527,7 +527,7 @@ impl TraceExporter {
                     match response.into_body().collect().await {
                         Ok(body) => {
                             info!(
-                                trace_count = trace_count,
+                                trace_count,
                                 "Traces sent successfully to agent"
                             );
                             self.emit_metric(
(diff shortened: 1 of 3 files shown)


hoolioh (approved):

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -522,6 +526,10 @@ impl TraceExporter {
                     }
                     match response.into_body().collect().await {
                         Ok(body) => {
+                            info!(
```
pierotibou, line 529: Shouldn't this be a debug log?

data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -522,6 +526,10 @@ impl TraceExporter {
...
+                                trace_count = trace_count,
```
pierotibou, line 530: I know nothing about Rust, so forgive my ignorance, but is it on purpose you assign this variable to itself? I see you do it elsewhere so I assume it's on purpose :)

data-pipeline/src/trace_exporter/mod.rs
ganeshnj: that's a name-value pair, could be shorthand

https://docs.rs/tracing/latest/tracing/#recording-fields

data-pipeline/src/trace_exporter/mod.rs
ganeshnj: that's a name-value pair, could be shorthand

https://docs.rs/tracing/latest/tracing/#recording-fields

PR #1102, opened 2025-06-12
do not include devflow/merge check in all checks as it creates a circular block

ekump:
# What does this PR do?

When trying to use the merge queue there is an action called `devflow/merge` that checks to see if required actions are complete. `all_checks` never completes because it's waiting on `devflow/merge`, and in turn `devflow/merge` never completes because `all_checks` never pass.

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/.github/workflows/all-checks.yml
+++ b/.github/workflows/all-checks.yml
@@ -16,3 +16,4 @@ jobs:
                 delay: '3'
                 retries: '30'
                 polling_interval: '1'
+                checks_exclude: 'devflow/merge'


danielsn (approved):
Thanks

.github/workflows/all-checks.yml
```diff
@@ -16,3 +16,4 @@ jobs:
                 delay: '3'
                 retries: '30'
                 polling_interval: '1'
+                checks_exclude: 'devflow/merge'
```
bantonsson, line 19: 🤣

bantonsson (approved):

PR #1104, opened 2025-06-13
feat(data-pipeline): signal agent response payload version and only send body when changed

paullegranddc:
# What does this PR do?

The datadog agent can take a `datadog-rates-payload-version` header, which signals the version of the agent rates which are sent in the agent response. This allows not refreshing the rates in the sampler of they haven't changed.

This PR adds a feature to store, send and update this value, to return the response body from the trace exporter only if the payload  version has been updated

# Motivation

This is a performance optimization because:
* If the header is unchanged, the agent is not going to bother sending the rates and will answer with an empty JSON object
* The libraries have less work to do to refresh the rates used by the sampler

# Additional Notes

# How to test the change?

### DIFF
--- a/bin_tests/tests/crashtracker_bin_test.rs
+++ b/bin_tests/tests/crashtracker_bin_test.rs
@@ -203,6 +203,9 @@ fn test_crash_tracking_bin(
     let sig_info = &crash_payload["sig_info"];
     assert_siginfo_message(sig_info, crash_typ);
 
+    let error = &crash_payload["error"];
+    assert_error_message(&error["message"], sig_info);
+
     let crash_telemetry = fs::read(fixtures.crash_telemetry_path)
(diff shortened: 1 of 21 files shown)


data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -1096,6 +1148,12 @@ impl TraceExporterBuilder {
         self
     }
 
+    /// Enables storing and checking the agent payload
+    pub fn enable_agent_rates_payload_version_enabled(&mut self) -> &mut Self {
```
bantonsson, line 1152: This feels very strange to `enable...enabled`. I think you should skip the last `enabled`.

bantonsson (approved):
Nice. Only minor nitpick.

hoolioh (approved):

PR #1106, opened 2025-06-13
Workaround in order to fix flaky tests

hoolioh:
# What does this PR do?

Check the file system entry before the file is parsed.

# Motivation

Flaky tests on macos runners.

# Additional Notes

Since macos tests are disabled on regular merges this PR sets a test branch for the gitlab workflow so the macos runners are triggered. Once the fix is validated the changes in workflow should be reverted.

### DIFF
--- a/bin_tests/src/constants.rs
+++ b/bin_tests/src/constants.rs
@@ -1,0 +1,4 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+//
+pub const TEST_COLLECTOR_TIMEOUT_MS: u32 = 10_000;
--- a/bin_tests/src/lib.rs
+++ b/bin_tests/src/lib.rs
@@ -1,7 +1,10 @@
 // Copyright 2023-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
+mod constants;
 pub mod modes;
+
+pub use constants::*;
 
 use std::{collections::HashMap, env, ops::DerefMut, path::PathBuf, process, sync::Mutex};
 
--- a/bin_tests/tests/crashtracker_bin_test.rs
+++ b/bin_tests/tests/crashtracker_bin_test.rs
@@ -7,7 +7,11 @@ use std::collections::HashMap;
 use std::io::{Read, Write};
 use std::path::Path;
 use std::process;
-use std::{fs, path::PathBuf, time::{Duration, Instant}};
+use std::{
+    fs,
+    path::PathBuf,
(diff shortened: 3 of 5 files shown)


bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -7,12 +7,35 @@ use std::collections::HashMap;
...
+    fs,
+    path::PathBuf,
+    time::{Duration, Instant},
+};
 
 use anyhow::Context;
 use bin_tests::{build_artifacts, ArtifactType, ArtifactsBuild, BuildProfile};
 use serde_json::Value;
 
+const CHECK_TIMEOUT_MS: u64 = 20_000;
```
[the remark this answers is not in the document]
hoolioh: I left that value because it was the one Gregory was using in his tests but it should be probably in sync with `const TEST_COLLECTOR_TIMEOUT_MS: u32 = 10_000;`. WDYT?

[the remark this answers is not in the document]
hoolioh: I've just moved the constant to a new module so they can be shared across the tests. @danielsn let me know if it better that way.

bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -184,6 +207,11 @@ fn test_crash_tracking_bin(
     );
     assert_eq!(Ok(""), String::from_utf8(stdout).as_deref());
 
+    assert!(check_file_existence(
+        fixtures.crash_profile_path.as_path(),
+        Duration::from_secs(TEST_COLLECTOR_TIMEOUT_MS as u64)
```
danielsn, line 212: see comment above about making `TEST_COLLECTOR_TIMEOUT_MS` a `Duration`

bin_tests/src/lib.rs
```diff
@@ -1,8 +1,11 @@
 // Copyright 2023-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
+mod constants;
 pub mod modes;
 
+pub use constants::*;
```
danielsn, line 7: I don't really care either way, but any reason that `modes` is `pub mod`, while `constants` is `pub use *`?

bin_tests/src/constants.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+//
+pub const TEST_COLLECTOR_TIMEOUT_MS: u32 = 10_000;
```
danielsn, line 4: Comment explaining how this value was chosen

PR #1111, opened 2025-06-17
[crashtracker] Prevent crashtracker from deadlocking

gleocadie:
# What does this PR do?

Prevent crashtracker from deadlocking when calling `is_being_traced`  function.

# Motivation

We got a recent report with this callstack (shorten):
```
(gdb) bt
#0  0x0000ffffb50ba180 in ?? () from /lib/aarch64-linux-gnu/libc.so.6
#1  0x0000ffffb50557ec in malloc () from /lib/aarch64-linux-gnu/libc.so.6
#2  0x0000ffffb38d57c8 in alloc::raw_vec::RawVecInner<A>::try_allocate_in () from /home/circleci/app/tmp/build_extension/modules/ddtrace.so
#3  0x0000ffffb2e0a614 in alloc::raw_vec::RawVecInner<alloc::alloc::Global>::with_capacity (capacity=8192) at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/alloc/src/raw_vec.rs:153
#4  alloc::raw_vec::RawVec<u8, alloc::alloc::Global>::with_capacity<u8> (capacity=8192) at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/alloc/src/raw_vec.rs:131
#5  alloc::boxed::{impl#2}::new_uninit_slice<u8> (len=8192) at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/alloc/src/boxed.rs:670
#6  0x0000ffffb2e0092c in std::io::buffered::bufreader::buffer::Buffer::with_capacity (capacity=8192) at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/std/src/io/buffered/bufreader/buffer.rs:35
#7  std::io::buffered::bufreader::BufReader<std::fs::File>::with_capacity<std::fs::File> (capacity=8192, inner=...) at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/std/src/io/buffered/bufreader.rs:103
#8  0x0000ffffb2e009d0 in std::io::buffered::bufreader::BufReader<std::fs::File>::new<std::fs::File> (inner=...) at /rustc/e71f9a9a98b0faf423844bf0ba7438f29dc27d58/library/std/src/io/buffered/bufreader.rs:74
#9  0x0000ffffb2dd77f8 in ddcommon::unix_utils::is_being_traced () at src/unix_utils.rs:192
#10 0x0000ffffb2dd7b38 in ddcommon::unix_utils::alt_fork () at src/unix_utils.rs:216
#11 0x0000ffffb268cf38 in datadog_crashtracker::collector::collector_manager::Collector::spawn (receiver=0xfffff20496e8, config=0xaaaac2cf3f40, config_str=..., metadata_str=..., sig_info=0xfffff20498d0, ucontext=0xfffff2049950) at src/collector/collector_manager.rs:31
#12 0x0000ffffb26881f0 in datadog_crashtracker::collector::crash_handler::handle_posix_signal_impl (sig_info=0xfffff20498d0, ucontext=0xfffff2049950) at src/collector/crash_handler.rs:174
#13 0x0000ffffb2687e84 in datadog_crashtracker::collector::crash_handler::handle_posix_sigaction (signum=6, sig_info=0xfffff20498d0, ucontext=0xfffff2049950) at src/collector/crash_handler.rs:95
#14 <signal handler called>
...
```
As you can see, an allocation is done while running in the signal handler. As shown in this case, this causes a deadlock: application gets stuck.

This happens (as seen in the callstack), `BufReader::new`  allocates. This is done [here](https://github.com/DataDog/libdatadog/blob/9ae6183d99cf80b69371b65a7bba06b74eb70e13/ddcommon/src/unix_utils.rs#L192). 

To avoid that, we need a zero-heap-allocation version of this version.

### DIFF
--- a/ddcommon/src/unix_utils/fork.rs
+++ b/ddcommon/src/unix_utils/fork.rs
@@ -31,67 +31,65 @@ const BUFFER_SIZE: usize = 1024;
 
 #[cfg(target_os = "linux")]
 fn is_being_traced_internal(mut file: File) -> io::Result<bool> {
-    let tracer_pid_marker = b"TracerPid:";
+    const TRACER_PID_MARKER: &[u8] = b"TracerPid:";
     let mut buffer = [0u8; BUFFER_SIZE];
     let mut data_len = 0;
-    let mut eof = false;
     let mut offset = 0;
 
-    while !eof {
+    loop {
+        // Shift unread data to the start if needed
         if offset > 0 && offset < data_len {
             let leftover_len = data_len - offset;
             buffer.copy_within(offset..data_len, 0);
             data_len = leftover_len;
             offset = 0;
-        } else if offset == data_len {
-            // All data processed, reset data_len and offset
+
+        // Handle edge case where no newline found in a full buffer,
+        // or all data processed: reset to read fresh data.
+        } else if offset == data_len || (offset == 0 && data_len == BUFFER_SIZE) {
             data_len = 0;
             offset = 0;
         }
 
-        let n = file.read(&mut buffer[data_len..])?;
-        if n == 0 {
-            eof = true;
+        let bytes_read = file.read(&mut buffer[data_len..])?;
+        if bytes_read == 0 {
+            // EOF reached, no more data to read
+            break;
         }
-        data_len += n;
+        data_len += bytes_read;
 
-        while offset < data_len {
-            if let Some(newline_pos) = buffer[offset..data_len].iter().position(|&b| b == b'\n') {
-                let line_end = offset + newline_pos;
-                let line = &buffer[offset..line_end];
+        // Process lines in the buffer
+        while let Some(newline_pos) = buffer[offset..data_len].iter().position(|&b| b == b'\n') {
+            let line_end = offset + newline_pos;
 
-                if line.starts_with(tracer_pid_marker) && line.len() > tracer_pid_marker.len() {
-                    if let Ok(line_str) = std::str::from_utf8(line) {
-                        let tracer_pid = line_str.split_whitespace().nth(1).unwrap_or("0");
-                        return Ok(tracer_pid != "0");
-                    }
-                }
-
-                offset = line_end + 1;
-            } else {
-                if offset == 0 && data_len == BUFFER_SIZE {
-                    // We did not find any newline in the buffer, so force
-                    // reading the full buffer anew
-                    offset = data_len;
-                }
-                // No newline found: partial line, stop processing and read more
-                break;
+            if let Some(result) =
+                check_tracer_pid_line(&buffer[offset..line_end], TRACER_PID_MARKER)?
+            {
+                return Ok(result);
             }
+
+            offset = line_end + 1;
         }
     }
 
-    // search in the remaining data
-    if data_len > offset {
-        let line = &buffer[offset..data_len];
-        if line.starts_with(tracer_pid_marker) {
-            if let Ok(line_str) = std::str::from_utf8(line) {
-                let tracer_pid = line_str.split_whitespace().nth(1).unwrap_or("0");
-                return Ok(tracer_pid != "0");
-            }
+    // Check remaining data without newline at EOF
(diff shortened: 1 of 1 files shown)


morrisonlevi (commented):
Looks good assuming that the thing we want is going to fit in the first 8K of the file. Otherwise we gotta figure out how to loop or something else.

Also, is 8k too much space on the stack?

gleocadie:
> Looks good assuming that the thing we want is going to fit in the first 8K of the file. Otherwise we gotta figure out how to loop or something else.
> 

Yeah, I did not want to make the complex more complex directly. Generally `/proc/self/status` is less that 4K but in some cases they can be a bit bigger. That's why I went for 8k.

> Also, is 8k too much space on the stack?
maybe, it was to take care of the case where the file is big. I can make it 4k if you think it would be safer.

ddcommon/src/unix_utils.rs
```diff
@@ -188,17 +188,21 @@ fn is_being_traced() -> io::Result<bool> {
...
-            return Ok(tracer_pid != "0");
+    // This function may run in signal handler, so we should ensure that we do not allocate
+    // memory on the heap (ex: avoiding using BufReader for exao).
+    let mut file = File::open("/proc/self/status")?;
+    let mut buf = [0u8; 8192];
+    let n = file.read(&mut buf)?;
+    let data = &buf[..n];
+
+    for line in data.split(|&b| b == b'\n') {
+        if let Ok(line_str) = std::str::from_utf8(line) {
```
danielsn, line 199: IF this fails, should we return false, or an error?

ddcommon/src/unix_utils.rs
```diff
@@ -188,17 +188,21 @@ fn is_being_traced() -> io::Result<bool> {
...
-
-    for line in reader.lines() {
-        let line = line?;
-        if line.starts_with("TracerPid:") {
-            let tracer_pid = line.split_whitespace().nth(1).unwrap_or("0");
-            return Ok(tracer_pid != "0");
+    // This function may run in signal handler, so we should ensure that we do not allocate
+    // memory on the heap (ex: avoiding using BufReader for exao).
+    let mut file = File::open("/proc/self/status")?;
+    let mut buf = [0u8; 8192];
```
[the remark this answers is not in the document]
gleocadie: in general it's less than 4KB, a pathological case would be 8-10KB but I've never seen such thing

ddcommon/src/unix_utils.rs
gleocadie: The question would be what should we do on error.
Today, the semantic is:
- `Ok(true)` : use `CLONE_PTRACE` as an extra-flags
- `Ok(false)`: use `0` to specify no clone flags

Should an error have the same behavior as `Ok(false)` or should be stop and no crash report is generated ?
All that said, I would say, we are fine today :)

gleocadie: so keep it like that

ddcommon/src/unix_utils.rs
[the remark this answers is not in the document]
gleocadie: TBH I do not know, my main goal is to avoid calling malloc here as simple as I can.
going for looping would make thing more complex and we sufficient space, I feel like it won't be necessary.
WDYT ?

gleocadie:
> Is there a specific reason this hasn't been merged yet? @gleocadie

sorry, I was on holiday and I have to discuss with @danielsn about this. This should be merged (updated if needed) this week.

ddcommon/src/unix_utils.rs
danielsn, line 194: One concern that just occurred to me here: this is a fair amount of stack space.  I recall we had issues with stack overflows before: it might be worth using a smaller buffer and looping to avoid that.

PR #1138, opened 2025-07-10
Add ddsketch to windows features list.

hoolioh:
# What does this PR do?

Release ddsketch on windows builds.


ivoanjo (commented):
Arguably, Ruby is the first (and only) consumer for this, so we could omit it until actually someone else wants this, since we don't support windows on Ruby libdatadog builds anyway.

windows/build-artifacts.ps1
```diff
@@ -28,7 +28,8 @@ $features = @(
     "datadog-profiling-ffi/ddtelemetry-ffi",
     "datadog-profiling-ffi/demangler",
     "datadog-library-config-ffi",
-    "datadog-log-ffi"
+    "datadog-log-ffi",
```
ekump, line 31: @hoolioh I cherry-picked this into my PR. You can close this one. Thanks!

PR #1152, opened 2025-07-24
Replace DevFlow stale branch auto-closer with MergeGate rule

ekump:
# What does this PR do?

We already have a github action that handles stale PRs that we like. The Devflow one isn't as flexible. The intention of the rule was to require branches not be too stale before merging.  This change gets rid of the old definition that auto-closed branches > 3 months old and adds a MergeGate rule that prevents merging branches that haven't been updated within the last 10 days.

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/repository.datadog.yml
+++ b/repository.datadog.yml
@@ -6,5 +6,7 @@ kind: mergequeue
 merge_method: squash
 ---
 schema-version: v1
-kind: stale-branches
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
:+1: Thanks for getting to the bottom of this.

PR #1162, opened 2025-07-28
bugfix - calc the agent info hash since it isn't provided via response header on the /info endpoint

ekump:
# What does this PR do?

The `AgentInfoFetcher` was incorrectly assuming the `datadog-agent-state` header would be present in the response to `/info` endpoint. This is not the case, so we have to calculate the hash ourselves from the body of `/info`. 
 
# Motivation

# Additional Notes

# How to test the change?


data-pipeline/src/agent_info/fetcher.rs
```diff
@@ -321,7 +335,11 @@ mod single_threaded_tests {
         "peer_tags": ["db.hostname","http.host","aws.s3.bucket"]
     }"#;
 
-    const TEST_INFO_HASH: &str = "8c732aba385d605b010cd5bd12c03fef402eaefce989f0055aa4c7e92fe30077";
+    fn calculate_hash(json: &str) -> String {
+        format!("{:x}", Sha256::digest(json.as_bytes()))
+    }
+
+    const TEST_INFO_HASH: &str = "b7709671827946c15603847bca76c90438579c038ec134eae19c51f1f3e3dfea";
```
ajgajg1134, line 342: I noticed the hash value changed here, is this intentional?

PR #1172, opened 2025-08-05
Fix in-process submission of telemetry

cataphract:
# What does this PR do?

Submission of telemetry logs from the appsec helper happens directly with the TelemetryWorker


bwoebi (approved):
This looks much simpler and should not pose any problems.

PR #1187, opened 2025-08-20
fix(build): Better parsing of C type definitions

paullegranddc:
Header type definitions generated by cbindgen need to be deduplicated between ddcommon and other crates.

Before, the code doing this dedup relied only on regexes,  making it unreliable for nested definitons, and it didn't work for type aliases and unions. Now this should work

# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/tools/src/lib.rs
+++ b/tools/src/lib.rs
@@ -2,22 +2,86 @@
 // SPDX-License-Identifier: Apache-2.0
 
 pub mod headers {
-    use regex::{Match, Regex, RegexBuilder};
+    use regex::{Regex, RegexBuilder};
     use std::collections::HashSet;
     use std::fs::{File, OpenOptions};
     use std::io::{self, BufReader, BufWriter, Read, Seek, Write};
     use std::sync::LazyLock;
 
-    static HEADER_TYPE_DECL_RE: LazyLock<Regex> = LazyLock::new(|| {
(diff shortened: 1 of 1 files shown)


ekump:
the failing integration tests for UDS will be fixed if you rebase main

tools/src/lib.rs
```diff
@@ -79,90 +145,197 @@ pub mod headers {
...
-        let mut base_new_parts = vec![&base_header_content[..base_defs.last().unwrap().end()]];
+        let mut base_new_parts = vec![&base_header_content[..base_defs.last().unwrap().end]];
         for child_def in &unique_child_defs {
             if base_defs_set.contains(child_def.as_str()) {
                 continue;
             }
             base_new_parts.push(child_def);
         }
-        base_new_parts.push(&base_header_content[base_defs.last().unwrap().end()..]);
+        base_new_parts.push(&base_header_content[base_defs.last().unwrap().end..]);
```
ekump, line 157: non-blocking general comment about unwrap() usage: I think we'd like to expand the use of the tools crate to help facilitate building libdatadog in the SDKs. We probably want to at least switch to expect to make debugging build failures easier. @hoolioh maybe this is something we can do as part of the publishing and versioning project?

PR #1189, opened 2025-08-26
[crashtracker] Add test to check callstack

gleocadie:
# What does this PR do?

This PR introduces a `crash_test_app` which does function calls, crashes. The crash tracker will collect the report and make sure that the crashing thread callstack contains certain frames.

# Motivation

Stability and quality: We want to make sure that crash tracker, more specifically the unwinder, accurately collects stack frames.

# Additional Notes

Currently, the test is enabled for:
- x86_64/glibc
- aarch64/glibc
- aarch64/musl-libc

But disabled for:
- x86_64 musl, due to missing CFI (unwinding information) in the libc library.
- macosx: we need to investigate why the report contains messages like `"resolve_names failed with failed to open proc maps file /proc/10282/maps: No such file or directory (os error 2)"`.  It seems that the application already died when we resolve names/normalize ips.

### DIFF
--- a/bin_tests/tests/crashtracker_bin_test.rs
+++ b/bin_tests/tests/crashtracker_bin_test.rs
@@ -129,11 +129,11 @@ fn test_crash_tracking_bin_prechain_sigabrt() {
     test_crash_tracking_bin(BuildProfile::Release, "prechain_abort", "null_deref");
 }
 
-// This test is disabled for now on x86_64 musl.
+// This test is disabled for now on x86_64 musl and macos
 // It seems that on aarch64 musl, libc has CFI which allows
 // unwinding passed the signal frame.
 #[test]
-#[cfg(not(all(target_arch = "x86_64", target_env = "musl")))]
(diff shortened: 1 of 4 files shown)


bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -129,6 +129,94 @@ fn test_crash_tracking_bin_prechain_sigabrt() {
...
+        "crashing_test_app::unix::fn1",
+        "crashing_test_app::unix::main",
+        "crashing_test_app::main",
+    ]
+    .to_vec();
+
+    let crashing_callstack = &crash_payload["error"]["stack"]["frames"];
+    assert!(
+        crashing_callstack.as_array().unwrap().len() >= expected_functions.len(),
+        "crashing thread callstacks ddddd",
```
gleocadie, line 205: make it cleaner

gleocadie: 🤣 it's the assert message, but I forgot to clean it and make it more clear

PR #1190, opened 2025-08-27
(libraryconfig) LibraryConfig logging instead of console stdout

anna-git:
# What does this PR do?

Dont print to console but add messages to a list if debug is activated.
Make this list available through FFI

Interrogation: 
I had to use a RefCell to store debug messages to have a mutable field in `Configurator`. I would have preferred these messages directly being available in the Result type when we call Configurator.get() but this would cause breaking changes with libraries already using the API and expecting a Vector of LibraryConfig... 

# Motivation

Tracers should avoid printing to console, but we still want the logs, especially to be able to debug.

# Additional Notes

# How to test the change?

### DIFF
--- a/datadog-library-config-ffi/src/lib.rs
+++ b/datadog-library-config-ffi/src/lib.rs
@@ -190,4 +190,12 @@ pub extern "C" fn ddog_library_config_local_stable_config_path() -> ffi::CStr<'s
 }
 
 #[no_mangle]
+pub extern "C" fn ddog_library_configurator_get_debug_message(
+    configurator: &Configurator,
+) -> ffi::CString {
+    let messages = configurator.inner.get_debug_messages();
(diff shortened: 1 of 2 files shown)


ivoanjo:
> Hi! so there shouldn't be any breaking changes,

Right, I'm just saying, in reply to your original comment on...

> I would have preferred [...] but this would cause breaking changes 

To say that we've done this often in libdatadog, so that can/should be on the table if it makes sense.

anna-git:
> > Hi! so there shouldn't be any breaking changes,
> 
> Right, I'm just saying, in reply to your original comment on...
> 
> > I would have preferred [...] but this would cause breaking changes
> 
> To say that we've done this often in libdatadog, so that can/should be on the table if it makes sense.

Oh ok, I got it the other around 🤣 
I guess given it's only debug messages and nothing really structural, causing breaking changes might not be worth it in this case (?), but good to know that it's an option for the future, thanks!

PR #1193, opened 2025-08-29
Increase retries from 30 to 45 in all-checks.yml

cbeauchesne:
# What does this PR do?

Per title

# Motivation

Some tests can takes ~30mn : https://github.com/DataDog/libdatadog/actions/runs/17307571226

So a 33 minutes timeout may be hit and cause a false positive : https://github.com/DataDog/libdatadog/actions/runs/17308401804/attempts/1 

# Additional Notes

# How to test the change?


ivoanjo (approved):
:+1: So depressing that CI is this slow but yeah at least let's make the all-checks not flaky because of this >_>

PR #1200, opened 2025-09-03
Fix spurious truncated debugger payloads

bwoebi:
The Body wrapper from hyper_migration is only used in debugger code. It very occasionally leads to a state where there's still a pending message on the Body::channel() stream, but the sender was closed. Then hyper will omit that last frame.


paullegranddc:
That makes sense, good find 😓

PR #1206, opened 2025-09-08
feat(data-pipeline): make errors more descriptive and granular

paullegranddc:
# Motivations

The Display implementation on the TraceExporterError doesn't print what the error variant is, which looses a lot of info. Furthermore some errors are cast to generic errors cases which make hard to go back to their actual sources


hoolioh (approved):
LGTM

PR #1209, opened 2025-09-09
[crashtracking] send initial `crash_ping` message for crash reporting

gyuheon0h:
# What does this PR do?

Currently, when the crashtracker itself fails, we have no way of knowing. Having the crashtracker send a small heartbeat message at the beginning of its operation, could allow us to pair with a successful crash report using the same crash ID, or if there is no pairing, we would know that something had failed.

# Motivation

The decision to send ping right after the first bytes are received in the `Receiver` is a design decision with tradeoffs. 
1. We cannot send network requests within any upstream components, as that is within the env of a signal handler.
2. We should send heartbeat request as early as possible, in order to cover the largest range in which data reporting could fail, to get the most value out of sending heartbeat request.

# Additional Notes

We send a lightweight message instead of a full `CrashInfo`. We create a separate `send_crash_ping` method to send crash ping because ping messages do not need extra tags and enrichment. We spawn a separate thread for sending ping to unblock the main thread.

# How to test the change?

`bin_test`


gyuheon0h:
Question to self: Why not send heartbeat as soon as `Receiver` process is forked?

gyuheon0h:
> I wonder if it's possible to check and test this in the `bin_tests` 🤔

Great idea. I just added some tests in `bin_tests` ^_^

PR #1224, opened 2025-09-17
feat: use docker cli instead of relying on testcontainers

paullegranddc:
# Motivations
Both testcontainers and httpmock have a lot of dependencies which we don't use elsewhere (some of them duplicates with incompatible versions).

This makes building tests quite a bit slower...

Also since testcontainer uses the docker sock, which is not available by default on mac, running the tests on a brand new mac install requires some extra setup.

In total this PR drops 60 tests dependencies on the data-pipeline crare

# What does this PR do?

* testcontainers can be replaced by running a few command with the docker cli directly,
* httpmock has been updated to the latest version, which uses hyper 1.0 instead of 0.14, which means deps are reusable now


ekump (approved):
I suspect more complex container usage is in our future, but we don't need test-containers right now. LGTM!

PR #1225, opened 2025-09-18
Add env to telemetry payload

VianneyRuhlmann:
# What does this PR do?

Add env to telemetry payloads.

# Motivation

Telemetry payloads sent by the exporter are missing env.

# Additional Notes

# How to test the change?


ganeshnj:
> Shouldn't we add the configuration option to the FFI layer?

i wouldn't recommend, this should be trace exporter env.

Long term, we should remove telemetry from trace exporter and make it own independent component supporting all missing properties.

PR #1226, opened 2025-09-18
feat: add support for process tags and container ids.

dmehala:
This PR introduces support for two new metadata fields:
 - `process_tags`: a comma-separated list of tags that will be used to identify a process.
 - `container_id`: the container id computed by the tracing library.

I also took the opportunity to rework the FFI to be easier to interface with, which also provide ABI stability.

[RFC](https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA/edit?tab=t.0#heading=h.fc4dk46k3lmw)

Changes:
  - Rework the FFI to be easier to use and ABI stability over time.
  - Bump the schema version to v2.

#### Usage Example
```c
void* builder = ddog_tracer_metadata_new();
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_TRACER_VERSION, "18.0.0");
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_RUNTIME_ID, "my-runtim-id");
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_TRACER_LANGUAGE, "c");
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_TRACER_VERSION, "0.0.1-rc");
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_HOSTNAME, "dmehala-beast-v2");
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_SERVICE_NAME, "my-service");
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_SERVICE_VERSION, "18.2.4");
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_PROCESS_TAGS, "p1,p2,p3");
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_CONTAINER_ID, "uuwidhd-djqhdiq-djkdq");

ddog_Result_TracerMemfdHandle handle = ddog_tracer_metadata_store(metadata);
// From there keep a ref on `handle` to keep the memfd alive.
```

### DIFF
--- a/datadog-library-config-ffi/src/tracer_metadata.rs
+++ b/datadog-library-config-ffi/src/tracer_metadata.rs
@@ -4,8 +4,9 @@
 #[cfg(target_os = "linux")]
 use datadog_library_config::tracer_metadata::AnonymousFileHandle;
 use datadog_library_config::tracer_metadata::{self, TracerMetadata};
-use ddcommon_ffi::{CharSlice, Result};
-use std::os::raw::c_int;
+use ddcommon_ffi::Result;
+use std::os::raw::{c_int, c_char};
+use std::ffi::CStr;
 
 /// C-compatible representation of an anonymous file handle
 #[repr(C)]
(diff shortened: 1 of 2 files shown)


raphaelgavache (approved):
nice, thanks!

dougqh:
While this is easier to use than what existed before, the part that caused the most problems still exists.
The return type which is a complex struct with unions is actually the hard thing to handle in Java and .NET.
That's why I'd proposed moving to a simpler shared result type.

dmehala:
> The problem here is the introduction of the enum. Since the enum definition cannot be consumed by Java or .NET, the enum creates similar maintenance problems to the struct / union return types but not as serious.

C enum are just plain `int` const. The enum definition don't need to be consumed. Instead, consumers of the FFI needs to provide the value. For example:
```c
ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_TRACER_VERSION, "18.0.0");

// Is equivalent to
ddog_tracer_metadata_set(builder, (int)2, "18.0.0");
```

> Since ABI compatibility allows for adding new functions, a better option would be creating a function for setting each field in the struct rather than using a field enum.

That's a possible solution, however, I don't agree it is particularly better. Contrary to your belief, adding new functions also suffers from maintenance and it's not cheaper than the enum. All in all, there's no silver bullet, you add new functionalities to the API, consumers of the FFI still needs to write the glue code.

dougqh:
> > C enum are just plain `int` const. The enum definition don't need to be consumed. Instead, consumers of the FFI needs to provide the value. For example:
> 
> ```c
> ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_TRACER_VERSION, "18.0.0");
> 
> // Is equivalent to
> ddog_tracer_metadata_set(builder, (int)2, "18.0.0");
> ```

 Yes, I realize that enum is the same as int, but by consume, I mean the calling side needs knowledge of the new definition. Having a magic constant on the calling side isn't great from a maintenance standpoint. That's why I proposed adding function that returns the value of each enum element for discoverability. That could be a good solution to some problems, but then we're already adding another function anyway.

> That's a possible solution, however, I don't agree it is particularly better. Contrary to your belief, adding new functions also suffers from maintenance and it's not cheaper than the enum. All in all, there's no silver bullet, you add new functionalities to the API, consumers of the FFI still needs to write the glue code.
 
My standpoint isn't that it is easier to add a function in libdatadog than adding an enum element. I think the enum element is slightly easier than adding another function. However, I want to focus on ease of use and maintenance for the many callers of libdatadog.

Also, my ultimate suggestion was just add a v2 function, that nicely conveys the new requirements and only needs one call from the host language to native land.

PR #1232, opened 2025-09-19
Migrate from regex to regex-lite

Aaalibaba42:
# What does this PR do?

Migrate from the `regex` crate to `regex-lite`.

# Motivation

The `regex` crate is very fast, but takes up lots of space in the binaries. `regex-lite` might introduce regression in performance, but we should see it's much more optimized in space.

# How to test the change?

There is (I believe so) a benchmark on the size of the artifacts (and the performance). We can evaluate if this change is worth it based on those.


Aaalibaba42:
tools:
- Regex used to find typedef/defines at the beginning of c files (Like #define MY_VAR 35) -> Not trivial but doable without regex without being too expensive I think

datadog-trace-obfuscation:
- ip_address.rs: Just segmentation of protocol vs address, maybe we could do without it but harder to optimize better than regex crates
- replacer.rs: Used to run the clients' regex for obfuscation, we can't do without

ddcommon:
- azure_app_services.rs: parsing and getting the "resource group" of azure. Given the Regex pattern we should be able to it without regexes relatively quickly
- entitiy_id/unix/mod.rs: Only used for testing in this file, I don't even think that it would be in the release crate with the configuration
- entitiy_id/unix/container_id.rs: Used to match cgroup to identify running container id, would be a bit harder to do without

datadog-live-debugger:
- expr_eval.rs: I don't have the full context of this, but when condition is checked for strings, it can be in the form of a regex match. So we couldn't do without them if the pattern is not known before-hand.
- redacted_names.rs: Most of the file uses regex_automata crate, only one time the regex crate to escape regular expression meta characters. Don't know whether we could do without

data-pipeline:
- CAN'T MIGRATE: testing done with `httpmock` implements traits from the `regex` crate that are not the same as the `regex-lite` crate (I think, I did not 100% investigate but compilation message lead me to believe this)
- src/telemetry/mod.rs: Just used for testing, Same as above, I don't even believe this would be present in the final release binary

paullegranddc (commented):
One not about specifying the dependency but otherwise LGTM

PR #1238, opened 2025-09-23
[APMSP-2299] artifact-size-benchmark from PATs to dd-octo-sts

Aaalibaba42:
# What does this PR do?

Use dd-octo-sts to have fine grained control of github token, instead of relying on github PATs.

https://datadoghq.atlassian.net/browse/APMSP-2299

# Motivation

artifact-size-benchmark gitlab workflow stopped working because the PAT was expired, and general policy change towards having more security and control over these kind of automation by going through dd-octo-sts

https://datadoghq.atlassian.net/browse/APMSP-2299

# Additional Notes

I don't know what I'm doing just yet

# How to test the change?

If gitlab's libddprof-build's artifact-size-benchmark PR manages to automagically post a comment on this PR (I'll try to make it work whilst both PRs are not merged) using dd-octo-sts instead of PAT, we will know it works okay-ish

### DIFF
--- a/.github/chainguard/gitlab.comment-artifact-size-in-pr.sts.yml
+++ b/.github/chainguard/gitlab.comment-artifact-size-in-pr.sts.yml
@@ -1,0 +1,21 @@
+# Runner on gitlab so issuer is the gitlab
+issuer: https://gitlab.ddbuild.io
+
+# We are on the target repo, so we ok things coming from there :downarrow:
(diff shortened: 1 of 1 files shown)


Aaalibaba42:
Since the repo emitting the token and the one the token acts upon is not the same, we need org policy, so this change is not required in this repository (if I understood correctly)

PR #1004, opened 2025-04-04
[APMSP-1764] Support borrowed buffer for msgpack decoding

anais-raison:
# What does this PR do?

Use &[u8] in the msgpack decoder instead of Bytes and produce a SpanSlice which contains &str instead of BytesString. This allows to parse non-owned buffer containing traces. It also provides a from_bytes wrapper which uses the same parsing logic but convert the &str to BytesString based on the given Bytes buffer.

# Motivation

The trace exporter needs to support reading traces from borrowed buffer.

# Additional Notes

This PR also refactors the string, map and number decoding logic to be moved to separate modules.

# How to test the change?

Describe here in detail how the change can be validated.

PR #1247, opened 2025-09-29
feat(sidecar)!: introduce TraceData to unify text and binary data

bwoebi:
Also move Span structures to v04, to make space for v1 spans (which will eventually become the new default). TraceData is also going to be used in the V1 implementation, to carry around byte arrays and strings alike, separate from the indexed offsets into the big vector.
This trivially allows directly working on tinybytes data or u8 slices, depending on the input, rather than requiring a copy-conversion at the end for tinybytes input. This is a major performance penalty (benchmarks say about 20% gain - I'm really confused why the regression introduced by #1004 was deemed acceptable).

This also addresses "APMSP-1941 - Replace `Bytes` with a wrapper that borrows the underlying".

Note that this PR partially reverts #1139 by @shreyamalpani - but instead provides the ability to pass an endpoint directly when sending.
This was done by the deliberate choice to remove Clone from Span, which is quite an expensive operation and should simply never be done. Especially the goal is for v1 spans to possibly store the backing storage for strings alongside the actual TracerPayload (which v1 has), at which point such an operation would be prohibitively expensive.

### DIFF
--- a/bin_tests/src/modes/behavior.rs
+++ b/bin_tests/src/modes/behavior.rs
@@ -167,7 +167,7 @@ mod tests {
             }
 
             let sigpipe_action = libc::sigaction {
-                sa_sigaction: sigpipe_handler as usize,
+                sa_sigaction: sigpipe_handler as *const () as usize,
                 sa_mask: sigset,
                 sa_flags: libc::SA_RESTART | libc::SA_SIGINFO,
                 #[cfg(target_os = "linux")]
--- a/bin_tests/src/modes/unix/test_001_sigpipe.rs
+++ b/bin_tests/src/modes/unix/test_001_sigpipe.rs
@@ -106,7 +106,7 @@ pub fn setup(output_dir: &Path) -> anyhow::Result<()> {
 
     // Set up the sigaction struct with sa_sigaction and sa_flags
     let sigpipe_action = libc::sigaction {
-        sa_sigaction: sigpipe_handler as usize,
+        sa_sigaction: sigpipe_handler as *const () as usize,
         sa_mask: sigset,
         sa_flags: libc::SA_RESTART | libc::SA_SIGINFO,
         #[cfg(target_os = "linux")]
--- a/bin_tests/src/modes/unix/test_002_sigchld.rs
+++ b/bin_tests/src/modes/unix/test_002_sigchld.rs
@@ -124,7 +124,7 @@ pub fn setup(output_dir: &Path) -> anyhow::Result<()> {
 
     // Set up the sigaction struct with sa_sigaction and sa_flags
     let sigchld_action = libc::sigaction {
-        sa_sigaction: sigchld_handler as usize,
+        sa_sigaction: sigchld_handler as *const () as usize,
         sa_mask: sigset,
         sa_flags: libc::SA_RESTART | libc::SA_SIGINFO,
         #[cfg(target_os = "linux")]
--- a/bin_tests/src/modes/unix/test_003_sigchld_with_exec.rs
+++ b/bin_tests/src/modes/unix/test_003_sigchld_with_exec.rs
@@ -102,7 +102,7 @@ pub fn setup(output_dir: &Path) -> anyhow::Result<()> {
 
     // Set up the sigaction struct with sa_sigaction and sa_flags
     let sigchld_action = libc::sigaction {
-        sa_sigaction: sigchld_handler as usize,
+        sa_sigaction: sigchld_handler as *const () as usize,
         sa_mask: sigset,
         sa_flags: libc::SA_RESTART | libc::SA_SIGINFO,
         #[cfg(target_os = "linux")]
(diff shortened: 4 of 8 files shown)


datadog-trace-utils/src/span/mod.rs
```diff
@@ -79,340 +80,94 @@ impl SpanText for BytesString {
...
 }
 
-#[derive(Serialize)]
-struct ArrayValueWrapper<'a, T: SpanText> {
-    values: &'a Vec<AttributeArrayValue<T>>,
-}
+/// Trait representing a tuple of (Text, Bytes) types used for different underlying data structures.
+/// Note: The functions are internal to the msgpack decoder and should not be used directly: they're
+/// only exposed here due to the inavailability of min_specialization in stable Rust.
+pub trait TraceData: Default + Debug + Clone + PartialEq + Serialize {
```
bantonsson, line 102: Why the `Clone` and `PartialEq`?

datadog-trace-utils/src/span/mod.rs
```diff
@@ -79,340 +80,94 @@ impl SpanText for BytesString {
...
-        S: serde::Serializer,
-    {
-        let mut state = serializer.serialize_struct("AttributeArrayValue", 2)?;
-        serialize_attribute_array::<S, T>(&mut state, self)?;
-        state.end()
-    }
-}
+/// TraceData implementation using `Bytes` and `BytesString`.
+#[derive(Default, Debug, Clone, PartialEq, Serialize)]
+pub struct TinyData;
```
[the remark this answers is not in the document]
bantonsson: I think `BytesData` is better.

datadog-trace-utils/src/span/mod.rs
```diff
@@ -79,340 +80,94 @@ impl SpanText for BytesString {
...
-                .map(|link| link.try_to_bytes(bytes))
-                .collect::<Option<Vec<SpanLinkBytes>>>()?,
-            span_events: self
-                .span_events
-                .iter()
-                .map(|event| event.try_to_bytes(bytes))
-                .collect::<Option<Vec<SpanEventBytes>>>()?,
+    #[inline]
+    fn read_string(buf: &mut Bytes) -> Result<BytesString, DecodeError> {
+        // Note: we need to pass a &'static lifetime here, otherwise it'll complain
```
bantonsson, line 137: What am I missing? I don't see any `&'static` here.

datadog-trace-utils/src/span/mod.rs
bwoebi: PartialEq is necessary because v04::Span is PartialEq. And yeah, generic types need to have all these derives of the context they are used in, otherwise #[derive] doesn't work.
I have no idea though why I added Clone. Removing.

datadog-trace-utils/src/span/mod.rs
bwoebi: read_string_ref_nomut takes &'static.

datadog-trace-utils/src/span/mod.rs
bwoebi: Yeah, cfg_attr doesn't work anyway, that'd need at least a feature flag. That's ugly.

datadog-trace-utils/src/send_data/mod.rs
```diff
@@ -26,7 +26,7 @@ use std::io::Write;
 #[cfg(feature = "compression")]
 use zstd::stream::write::Encoder;
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
```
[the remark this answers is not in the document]
bwoebi: They are, because it relied on Clone on the Span, and TraceData is not meant to be Clone outside of testing.

datadog-trace-utils/src/msgpack_decoder/decode/buffer.rs
```diff
@@ -0,0 +1,63 @@
...
+            DecodeError::InvalidType(format!("Type mismatch at marker {marker:?}"))
+        }
+        DecodeStringError::InvalidUtf8(_, e) => DecodeError::Utf8Error(e.to_string()),
+        _ => DecodeError::IOError,
+    })
+}
+
+/// Internal Buffer used to wrap msgpack data for decoding.
+/// Provides a couple accessors to extract data from the buffer.
+pub struct Buffer<T: TraceData>(T::Bytes);
```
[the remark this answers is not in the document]
bwoebi: We need to at least store the Bytes reference to construct the Bytes/BytesString object.

bwoebi:
> This also means we will have to reimplement all trace utils for v1 span.

My plan is to have a projection with some functions which is compatible with all versions of the spans. Meaning that adding a float attribute will reroute to metrics, everything else to string.

That's why moving the span to v04 (and add a new span to v1), then we can add a common projection which routes accordingly, agnostic to the actual backend span used. I definitely want to avoid re-implementing everything.

PR #1249, opened 2025-09-30
Reuse connections

hoolioh:
# What does this PR do?
This PR modifies `send_data` module and `send_with_retry` function so they can accept an `HttpClient`. This leads to some benefits:

- Since the http client (based on hyper) maintain a pool of connections they can be reused allowing to avoid unnecessary TCP handshakes with the agent.
- Avoid constantly recreating the client in the send functions.
- Allow to the caller to handle the connection pool and the proxy configuration.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -20,7 +20,9 @@ datadog-remote-config       @Datadog/libdatadog-php @Datadog/libdatadog-apm @Dat
 datadog-sidecar             @Datadog/libdatadog-php @Datadog/libdatadog-apm
 datadog-sidecar-ffi         @Datadog/libdatadog-php @Datadog/libdatadog-apm
 data-pipeline*/             @Datadog/libdatadog-apm
+datadog-tracer-flare        @Datadog/libdatadog-apm
 ddsketch                    @Datadog/libdatadog-apm @Datadog/libdatadog-telemetry
+datadog-ffe                 @Datadog/feature-flagging-and-experimentation-sdk
 
 # Most of the bin_tests are owned by the profiling team, but some are owned by the core team
 bin_tests/                  @Datadog/libdatadog-profiling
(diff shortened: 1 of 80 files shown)


data-pipeline/src/stats_exporter.rs
```diff
@@ -48,6 +50,7 @@ impl StatsExporter {
         meta: TracerMetadata,
         endpoint: Endpoint,
         cancellation_token: CancellationToken,
+        client: Option<HttpClient>,
```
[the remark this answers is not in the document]
hoolioh: I was unsure if there would be any use case scenario in which we wouldn't require the caller to pass it. I you can't think of any I'll remove the Option

datadog-trace-utils/src/send_with_retry/mod.rs
```diff
@@ -223,41 +223,16 @@ pub async fn send_with_retry(
...
     timeout: Duration,
     req: HttpRequestBuilder,
     payload: Bytes,
-    http_proxy: Option<&str>,
 ) -> Result<hyper_migration::HttpResponse, RequestError> {
     let req = req
         .body(hyper_migration::Body::from_bytes(payload))
         .or(Err(RequestError::Build))?;
 
-    #[cfg(feature = "proxy")]
```
bantonsson, line 235: Who was the user of this `proxy` feature? Is the refactor cleared with them?

bantonsson (approved):
LGTM

Has this been cleared with Serverless?

datadog-trace-utils/src/send_with_retry/mod.rs
hoolioh: The proxy feature was introduced by the serverless team. I shared the the PR when it was a draft with AJ and the response was positive. He was going to run some tests, I'm waiting for his feedback before merging.

bantonsson: Sounds like s plan

PR #1251, opened 2025-09-30
Add support for pb spans in span concentrator

VianneyRuhlmann:
# What does this PR do?

Add support for pb spans in span concentrator. The concentrator has been moved to a new crate to make it easier to reuse.

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/.github/chainguard/gitlab.libddprof-comment-pr.sts.yaml
+++ b/.github/chainguard/gitlab.libddprof-comment-pr.sts.yaml
@@ -1,7 +1,0 @@
-issuer: https://token.actions.githubusercontent.com
-
-subject: repo:DataDog/libdatadog
-
-permissions:
-  metadata: read
-  pull_requests: write
--- a/.github/chainguard/gitlab.libddprof-draft-release.sts.yaml
+++ b/.github/chainguard/gitlab.libddprof-draft-release.sts.yaml
@@ -1,7 +1,0 @@
-issuer: https://token.actions.githubusercontent.com
-
-subject: repo:DataDog/libdatadog
-
-permissions:
-  contents: write
-  metadata: read
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1301,6 +1301,7 @@ dependencies = [
  "clap",
  "criterion",
  "datadog-ddsketch",
+ "datadog-log",
  "datadog-trace-protobuf",
  "datadog-trace-stats",
  "datadog-trace-utils",
--- a/builder/src/bin/release.rs
+++ b/builder/src/bin/release.rs
@@ -123,7 +123,7 @@ pub fn main() {
     match res {
         Ok(_) => {
             builder.sanitize_libraries();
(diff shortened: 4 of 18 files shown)


datadog-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -24,26 +25,25 @@ fn align_timestamp(t: u64, bucket_size: u64) -> u64 {
     t - (t % bucket_size)
 }
 
-/// Return true if the span has a span.kind that is eligible for stats computation
-fn compute_stats_for_span_kind<T>(span: &Span<T>, span_kinds_stats_computed: &[String]) -> bool
+/// Return true if the span is eligible for stats computation
+fn span_is_eligible<T>(span: &Span<T>, span_kinds_stats_computed: &[String]) -> bool
```
ekump, line 29: nit: I prefer `is_span_eligible` (a question) over `span_is_eligible` (a statement). functions that return a bool should be named like a question the return value will answer.

datadog-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -146,6 +158,67 @@ impl<'a> BorrowedAggregationKey<'a> {
...
+    /// Return an AggregationKey matching the given span.
+    ///
+    /// If `peer_tags_keys` is not empty then the peer tags of the span will be included in the
+    /// key.
+    pub(super) fn from_pb_span(span: &'a pb::Span, peer_tag_keys: &'a [String]) -> Self {
+        let span_kind = span
+            .meta
+            .get(TAG_SPANKIND)
+            .map(|s| s.as_str())
+            .unwrap_or("");
```
ekump, line 171: This can be `unwrap_or_default()`, no?

ekump (approved):
It would be nice if we could have unified logic for both span types, but not blocking the PR right now in case it's time sensitive, or doing so would be less performant.

datadog-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -146,6 +158,67 @@ impl<'a> BorrowedAggregationKey<'a> {
...
+            .or_else(|| span.meta.get("http.route"))
+            .map(|s| s.as_str())
+            .unwrap_or_default();
+
+        let status_code = if let Some(status_code) = span.metrics.get(TAG_STATUS_CODE) {
+            *status_code as u32
+        } else if let Some(status_code) = span.meta.get(TAG_STATUS_CODE) {
+            status_code.as_str().parse().unwrap_or(0)
+        } else {
+            0
```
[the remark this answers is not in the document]
ekump: leaving it as is is not a big deal, but what is heavy about `u32::default()`? I assume the compiler will optimize and inline it as `0` at the end of the day. 

The reason I suggested a named const, or `u32::default()`even, is so that future readers know we're intentionally setting a default value. If you think `0` correctly implies this then that's fine too.

PR #1422, opened 2025-12-19
feat(profiler): support cancellation tokens in exporter cxx bindings

danielsn:
# What does this PR do?

Adds cancelation support for exporting in the C++ bindings

# Motivation

Forgot to add this when I created the bindings in the first place

# Additional Notes

# How to test the change?


morrisonlevi (approved):
Looks straightforward to me. Does cbindgen automatically make a `send_profile_with_cancellation` method in C++ from this?

PR #1552, opened 2026-02-10
feat(profiling): OTel style API

danielsn:
# What does this PR do?

Adds a new API which takes a single sample type by enum along with a single value.  Currently, we just forward to the existing API, once we're happy with this we can start changing the underlying data model.

# Motivation

1. This matches what OTel expects, first step towards making libdatadog match otel
2. The current vector of values is sparse, we're storing a bunch of zeros.

# Additional Notes

# How to test the change?


ivoanjo (approved):
👍 Seems harmless for testing but I do kinda wonder how much it's worth adding these without a pairing PR on the library side showing how it's used. I think the best libdatadog APIs are the ones co-designed where we did both sides at once and made sure they fit really well, as there's often opportunities to change the API to make the caller side even nicer.

PR #1634, opened 2026-02-26
feat: change cargo patch approach to generic annotations

Aaalibaba42:
# What does this PR do?

Change cargo patch approach to generic annotations

# Motivation

Discussions with the team about the best approach.

# Additional Notes

Currently this is for 1 (one) capability, when we have 2, 3, or more capabilities, all these annations should either:
- Be bundled up into one
- Take multiple generics (that actually will always be all correlated together)

In the former option, we lose the main gain of this approach: explicit denotation of the aspect that can change. In the latter, it's just more ugly than it already is.

But functionally, I agree this is still static dispatch for statically known information, so that's fine by me.

# How to test the change?

For now this is mostly demonstrative, not actually something that ought be merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1568,6 +1568,7 @@ dependencies = [
  "http-body-util",
  "httpmock",
  "libc",
+ "libdd-capabilities-impl",
  "libdd-common",
  "libdd-common-ffi",
  "libdd-crashtracker",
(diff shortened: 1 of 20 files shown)


bantonsson (commented):
So I don't think this looks that terrible, and as you said bundling up the _capabilities_ in some common type will ensure that it doesn't get too unwieldy.

PR #1642, opened 2026-03-03
ci: dynamic pipeline

hoolioh:
# What does this PR do?

Makes all CI workflows **dynamic** — they detect which crates were changed/affected by a PR and only run jobs for those crates, skipping everything on pushes to `main` where nothing changed.
### Key changes

**New `crates-reporter` composite action** (replaces `changed-crates`)
- The old action was a bash-heavy composite action doing crate detection in shell script
- The new one is a compiled Rust binary that uses `cargo-metadata` to detect changed crates **and** compute transitive dependants (affected crates), giving a richer output: `crates`, `affected_crates`, `crates_count`, `affected_crates_count`, `status`
- A new `ci-shared` library crate provides shared logic (`git`, `workspace`, `crate_detection`, `github_output`) reused across multiple action binaries
- A shared `Cargo.toml`/`Cargo.lock` workspace at `.github/actions/` replaces per-action workspaces

**`test.yml`** — unit tests now only run on affected packages (direct + transitive dependants), not the full workspace. The `cross-centos7` job was made dynamic: it builds the nextest command conditionally based on `$PACKAGES`/`$CRASHTRACKER_FEATURE` from the setup stage

**`test-ffi.yml`** (new, split from `test.yml`) — FFI jobs extracted into their own workflow, also driven by the setup stage; only runs when `-ffi` crates are affected

**`lint.yml`** — `rustfmt` now runs only on directly changed crates; `clippy` runs on affected crates (since a change can introduce warnings in dependants)

**`miri.yml`** — Miri now runs only on affected crates on PRs; still runs `--workspace` on pushes to `main`

**`fuzz.yml`** — fuzz jobs now filter to only the fuzz-capable crates that were affected

# Motivation

Previously, every PR triggered CI jobs against the entire workspace regardless of what changed. With the current approach:
- Reduces CI runtime and cost by skipping unaffected crates
- Uses transitive dependency analysis (not just direct changes) so a change in a shared crate still triggers tests for its dependants
- Replaces a fragile bash-based crate detector with a reliable, tested Rust implementation using `cargo-metadata`
- Splits FFI testing into its own workflow for cleaner separation of concerns and run both test jobs (nextests and ffi) in parallel.

# Results
The new pipeline has been tested with two crates. libdd-common which is a foundational crate and has a lot of transitive dependencies (almost all workspace) and libdd-data-pipeline-ffi which only has one transitive dependency.

| Workflow | libddcommon | libdd-data-pipeline-ffi |
|----------|----------------------------|-----------------------------------------|
| Fuzz     | ~22m (95%) | ~13m (92%)  |
| Miri     | ~1m (4%) | **~23m (92%)** |
| Test     | ~1m (4%) | **~12m (43%)** |
| Lint     | -1m (-14%, setup overhead) | ~6m (50%) |
| **Total wall-clock** | **~0m (0%)** | **11m 6s (39%)** |


.github/actions/ci-shared/src/crate_detection.rs
```diff
@@ -35,47 +33,90 @@ fn find_closest_cargo_toml(mut path: &Path) -> Option<PathBuf> {
...
+        .and_then(|v| v.as_str())
+        .with_context(|| format!("No package.name in {}", manifest.display()))?
+        .to_string();
+
+    // version.workspace = true → use "workspace" as the version string
+    let version = if let Some(ver_table) = package.get("version").and_then(|v| v.as_table()) {
+        if ver_table.get("workspace").and_then(|v| v.as_bool()) == Some(true) {
+            "workspace".to_string()
+        } else {
+            "workspace".to_string() // unusual but treat as workspace
```
ekump: both branches are returning the same thing

.github/actions/crates-reporter/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "crates-reporter"
+version = "0.1.0"
+edition = "2021"
+rust-version = "1.84.1"
+description = "GitHub Action for reporting modified crates and transitive dependencies"
+homepage = "https://github.com/DataDog/libdatadog/tree/main/.github/actions/changed-crates"
```
ekump: i don't think these urls are correct anymore

.github/workflows/lint.yml
```diff
@@ -1,10 +1,55 @@
 name: Lint
 on:
+  pull_request:
   push:
+    branches:
+      - main
```
ekump, line 6: I think you need the `mq-working-branch-*` in this file too

.github/workflows/miri.yml
```diff
@@ -1,13 +1,53 @@
 name: Miri test
 on:
+  pull_request:
+    branches-ignore:
+      - release
   push:
+    branches:
+      - main
```
ekump, line 8: `mq-working-branch-*`  is misisng

.github/workflows/fuzz.yml
```diff
@@ -1,12 +1,47 @@
 name: Fuzz test
 on:
+  pull_request:
   push:
+    branches:
+      - main
```
ekump, line 6: `mq-working-branch-*` is missing

.github/workflows/fuzz.yml
```diff
@@ -1,12 +1,47 @@
...
+    - name: Set fuzz directories
+      id: set-dirs
+      env:
+        CHANGED_CRATES_STATUS: ${{ steps.changed-crates.outputs.status }}
+        AFFECTED_CRATES: ${{ steps.changed-crates.outputs.affected_crates }}
+      run: |
+        ALL_FUZZ='["libdd-alloc", "libdd-profiling", "libdd-common-ffi", "libdd-trace-utils"]'
+        if [[ "$CHANGED_CRATES_STATUS" == "success" ]]; then
+          FILTERED=$(echo "$AFFECTED_CRATES" | jq --argjson fuzz "$ALL_FUZZ" \
+            '[.[] | .path | split("/") | last | . as $dir | if ($fuzz | index($dir) != null) then $dir else empty end]')
```
ekump: Now that you're using `affected_crates` is this jq query going to work? Do the objects in `affected_crates` have a path?

hoolioh: good catch

.github/workflows/lint.yml
hoolioh: My reasoning was that if we check the lining in every PR likely a merge will not incur in any linting error. However since the linting process is quite light compared with the miri or fuzz ones, moreover it runs in parallel I think it won't make any difference so I'm adding it.

ekump: Ah, ok. Maybe not in this PR, but I'm open to the idea of not running the linters in the MQ, as long as it also doesn't run on main. The chance of a merge creating clippy errors is low. I don't have a strong opinion, but I think a case could be made that clippy failing on main isn't a big deal as long as the code compiles and tests pass. A valid counterpoint is that when you branch off main to work on a new PR it should be in a good state. I think running clippy on the MQ and main is fine for now and we can revisit it if it starts to become a bottleneck.

ekump (approved):

ekump (approved):
LGTM, but can we add something to the README or contributing doc that calls out PRs will have dynamic tests, and the MQ / main will have the full suite run and that there is a chance the MQ will (hopefully never) catch failures that weren't in the PR?

PR #1696, opened 2026-03-10
feat(profiling)!: Add `BufWriter` for batching calls to `ZSTD_compressStream`

realFlowControl:
# What does this PR do?

This PR wraps the zstd streaming encoder in `TimestampedObservations` with a `BufWriter` sized to `ZSTD_CStreamInSize()` (128 KiB). Each `add_sample` call makes several small writes (4-44 bytes) without buffering. Each one triggers a `ZSTD_compressStream` call. The `BufWriter` batches these into fewer, optimally sized calls to the compressor, reducing per-call FFI overhead. A new benchmark (`profile_add_sample_timestamped_x1000`) is included to measure the impact.      

# Motivation

Lower overhead in profiling.

I found this when checking the `dd-trace-rb` full host profiles, that for every `add_sample` call we spend quite an amount in `ZSTD_compressStream` and wondered why that is the case. A bit of searching made me land at https://github.com/facebook/zstd/blob/d7ee3207cc0db53f78fc6a69babc80747b1b7658/lib/zstd.h#L829-L843 with the TLDR being that it is adviced to call `ZSTD_compressStream*()` with the recommended buffers sizes and limit calls over FFI.

And that is what this PR does.

# Additional Notes

I added a second `Label` to the bench, making the numbers a bit more realistic, in PHP we send at least 5 labels with every sample, IDK about other consumers of libdatadog, so I guess doubling the amount of labels is most likely fine with everyone.

The PR has two commits, the first brings the benchmark, the second the code changes.

# How to test the change?

```bash
$ git checkout HEAD~1
...
$ cargo bench -p libdd-profiling -- profile_add_sample_timestamped_x1000
...
profile_add_sample_timestamped_x1000
                        time:   [747.84 µs 749.31 µs 751.03 µs]
...
$ git checkout florian/buf_writer
...
$ cargo bench -p libdd-profiling -- profile_add_sample_timestamped_x1000
...
profile_add_sample_timestamped_x1000
                        time:   [680.57 µs 681.52 µs 682.50 µs]
                        change: [-9.5874% -9.3376% -9.0913%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Or with `dd-trace-php` at https://github.com/DataDog/dd-trace-php/pull/3697

### DIFF
--- a/libdd-profiling/benches/add_samples.rs
+++ b/libdd-profiling/benches/add_samples.rs
@@ -106,12 +106,20 @@ pub fn bench_add_sample_vs_add2(c: &mut Criterion) {
     let functions = dict.functions();
     let thread_id = get_current_thread_id();
     let thread_id_key: StringId2 = strings.try_insert("thread id").unwrap().into();
-    let labels_api = vec![api::Label {
-        key: "thread id",
-        str: "",
-        num: thread_id,
-        num_unit: "",
-    }];
+    let labels_api = vec![
+        api::Label {
+            key: "thread id",
+            str: "",
+            num: thread_id,
+            num_unit: "",
+        },
+        api::Label {
+            key: "thread name",
+            str: "this thread",
+            num: 0,
+            num_unit: "",
+        },
+    ];
 
     let frames2 = frames.map(|f| {
         let set_id = functions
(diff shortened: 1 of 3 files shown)


ivoanjo (approved):
👍 Very nice improvement and very nicely spotted :)

PR #1698, opened 2026-03-10
chore(sidecar): reorg/remove unsafe code in span FFI

yannham:
# What does this PR do?

This PR makes a pass on the span API to clean up `unsafe` annotations, add missing Safety comments, etc. leaving behavior unchanged and the C API unchanged (see the diff below), although the Rust's side types are richer/more restrictive. Some code is also written in a more idiomatic way, when applicable.

I believe the change from `*mut` to `Option<&mut>` for some return values could actually return immutable references (since the two functions are call `get_xxx`), but this is technically a breaking change (changing `*mut` to `*const` in the FFI) and I don't know if those values can actually be mutated, so I stayed with `mut`. Happy to make them const in the future if it's sufficient.

<details>
<summary>C header diff</summary>

```diff
751a752,755
> /**
>  * The returned slice is an owned allocation that must be properly freed using
>  * [`ddog_free_charslice`].
>  */
753a758,766
> /**
>  * Frees an owned [`CharSlice`]. Note that some functions of this API return borrowed slices that
>  * must NOT be freed. Only a few selected functions return slices that must be freed, and this is
>  * mentioned explicitly in their documentation.
>  *
>  * # Safety
>  *
>  * `slice` must be an owned char slice that has been returned by one of the functions of this API.
>  */
803a817,820
> /**
>  * The return value is an owned array of slices (`Box<[CharSlice]>`) that must be freed explicitly
>  * through [`ddog_span_free_keys_ptr`].
>  */
823a841,844
> /**
>  * The return value is an array of slices (`Box<[CharSlice]>`) that must be freed explicitly
>  * through [`ddog_span_free_keys_ptr`].
>  */
825a847,852
> /**
>  * # Safety
>  *
>  * `keys_ptr` must have been returned by one of the `ddog_xxx_get_keys()` functions, and must not
>  * have been already freed.
>  */
857a885,888
> /**
>  * The returned slice is an owned allocation that must be properly freed using
>  * [`ddog_free_charslice`].
>  */
```
</details>

# Motivation

While reading the code, I realized some of the safety comments were missing, some unsafe annotations weren't required, or `unsafe` attributes on functions were missing. Since unsafe and FFI code must be particularly scrutinized, I think a little cleaning pass doesn't hurt.

# Additional Notes

One could question the approach of making the lifetimes more precise in the FFI, since all that information is erased at the interface (indeed, the C header is identical, on purpose). However, I find that it's an efficient way to convey the constraints that the caller must uphold (most of the time, do not free the original allocation this function's return value borrows from). I originally wrote those conditions as "Safety" comments, but it's really verbose and repetitive.

Maybe both could be combined; make the Rust types more precise, but also propagate that information as text, so that it appears in the C header file?

# How to test the change?

Run `cbindgen` in the `datadog-sidecar-ffi` crate directory to generate the C header file and see that the diff is only comments when compared to main. Run the tests.

### DIFF
--- a/datadog-sidecar-ffi/src/span.rs
+++ b/datadog-sidecar-ffi/src/span.rs
@@ -142,9 +142,14 @@ pub extern "C" fn ddog_get_traces_size(traces: &TracesBytes) -> usize {
 // can return values as `Option<&/&mut>` that will have the same ABI `*const/*mut TraceBytes`, with
 // `None` corresponding to `null`.
 
+// I don't think we need to return a mutable reference here, but when changing the return value
+// from `*mut` to `Option<&mut>`, we wanted to preserve the FFI at first.
 #[no_mangle]
-pub extern "C" fn ddog_get_trace(traces: &mut TracesBytes, index: usize) -> Option<&TraceBytes> {
-    traces.get(index)
+pub extern "C" fn ddog_get_trace(
+    traces: &mut TracesBytes,
+    index: usize,
+) -> Option<&mut TraceBytes> {
+    traces.get_mut(index)
 }
 
 // ------------------ TraceBytes ------------------
@@ -159,9 +164,11 @@ pub extern "C" fn ddog_get_trace_size(trace: &TraceBytes) -> usize {
     trace.len()
 }
 
+// I don't think we need to return a mutable reference here, but when changing the return value
+// from `*mut` to `Option<&mut>`, we wanted to preserve the FFI at first.
 #[no_mangle]
-pub extern "C" fn ddog_get_span(trace: &mut TraceBytes, index: usize) -> Option<&SpanBytes> {
-    trace.get(index)
+pub extern "C" fn ddog_get_span(trace: &mut TraceBytes, index: usize) -> Option<&mut SpanBytes> {
+    trace.get_mut(index)
 }
 
 // ------------------- SpanBytes -------------------
--- a/datadog-sidecar-ffi/src/span.rs
+++ b/datadog-sidecar-ffi/src/span.rs
@@ -12,13 +12,13 @@ use std::ffi::{c_char, CString};
 
 fn convert_char_slice_to_bytes_string(slice: CharSlice) -> BytesString {
     // TODO: Strip the invalid bytes in the tracer instead
-    unsafe {
-        match String::from_utf8_lossy(slice.as_bytes()) {
-            Cow::Owned(s) => s.into(),
-            Cow::Borrowed(_) => {
-                BytesString::from_bytes_unchecked(Bytes::from_underlying(slice.as_bytes().to_vec()))
-            }
-        }
+    match String::from_utf8_lossy(slice.as_bytes()) {
+        Cow::Owned(s) => s.into(),
+        Cow::Borrowed(_) => unsafe {
+            // Safety: if `from_utf8_lossy` returns a borrowed `str`, the latter can't be anything
+            // else than the original slice which is thus valid UTF8.
+            BytesString::from_bytes_unchecked(Bytes::from_underlying(slice.as_bytes().to_vec()))
+        },
     }
 }
 
@@ -31,8 +31,10 @@ fn set_string_field(field: &mut BytesString, slice: CharSlice) {
 }
 
 #[inline]
-fn get_string_field(field: &BytesString) -> CharSlice<'static> {
+fn get_string_field(field: &BytesString) -> CharSlice<'_> {
     let string = field.as_str();
+    // Safety: `field` is a `ByteString`, which guarantees it is backed by valid UTF8 bytes. The
+    // lifetime of the returned slice makes sure it borrows from `field`.
     unsafe { CharSlice::from_raw_parts(string.as_ptr().cast(), string.len()) }
 }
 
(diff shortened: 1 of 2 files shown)


datadog-sidecar-ffi/src/span.rs
```diff
@@ -130,41 +138,44 @@ pub extern "C" fn ddog_get_traces_size(traces: &TracesBytes) -> usize {
...
+// `None` corresponding to `null`.
 
-    unsafe { traces.get_unchecked_mut(index) as *mut TraceBytes }
+// I don't think we need to return a mutable reference here, but when changing the return value
+// from `*mut` to `Option<&mut>`, we wanted to preserve the FFI at first.
+#[no_mangle]
+pub extern "C" fn ddog_get_trace(
+    traces: &mut TracesBytes,
+    index: usize,
+) -> Option<&mut TraceBytes> {
```
yannham, line 148: I don't think we need to return a mutable reference here, but when changing the return value from `*mut` to `Option<&mut>`, I wanted to preserve the FFI at first.

datadog-sidecar-ffi/src/span.rs
```diff
@@ -130,41 +138,44 @@ pub extern "C" fn ddog_get_traces_size(traces: &TracesBytes) -> usize {
...
-    if index >= traces.len() {
-        return std::ptr::null_mut();
-    }
+// Note that per [RFC3391](https://rust-lang.github.io/rfcs/3391-result_ffi_guarantees.html), we
+// can return values as `Option<&/&mut>` that will have the same ABI `*const/*mut TraceBytes`, with
+// `None` corresponding to `null`.
 
-    unsafe { traces.get_unchecked_mut(index) as *mut TraceBytes }
+// I don't think we need to return a mutable reference here, but when changing the return value
+// from `*mut` to `Option<&mut>`, we wanted to preserve the FFI at first.
```
yannham:
```suggestion
// TODO: I don't think we need to return a mutable reference here
```

datadog-sidecar-ffi/src/span.rs
```diff
@@ -130,41 +138,43 @@ pub extern "C" fn ddog_get_traces_size(traces: &TracesBytes) -> usize {
...
+    new_vector_item(traces)
 }
 
 #[no_mangle]
 pub extern "C" fn ddog_get_trace_size(trace: &TraceBytes) -> usize {
     trace.len()
 }
 
+// I don't think we need to return a mutable reference here, but when changing the return value
+// from `*mut` to `Option<&mut>`, we wanted to preserve the FFI at first.
```
yannham:
```suggestion
// TODO:I don't think we need to return a mutable reference here
```

datadog-sidecar-ffi/src/span.rs
```diff
@@ -12,13 +12,13 @@ use std::ffi::{c_char, CString};
...
-                BytesString::from_bytes_unchecked(Bytes::from_underlying(slice.as_bytes().to_vec()))
-            }
-        }
+    match String::from_utf8_lossy(slice.as_bytes()) {
+        Cow::Owned(s) => s.into(),
+        Cow::Borrowed(_) => unsafe {
+            // Safety: if `from_utf8_lossy` returns a borrowed `str`, the latter can't be anything
+            // else than the original slice which is thus valid UTF8.
+            BytesString::from_bytes_unchecked(Bytes::from_underlying(slice.as_bytes().to_vec()))
+        },
```
paullegranddc, line 24: Can you leave something like

```
TODO: replace with [from_utf8_lossy_owned](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy_owned)  once stabilized
```

datadog-sidecar-ffi/src/span.rs
```diff
@@ -130,41 +138,43 @@ pub extern "C" fn ddog_get_traces_size(traces: &TracesBytes) -> usize {
...
 }
 
-#[no_mangle]
-pub extern "C" fn ddog_get_trace(traces: &mut TracesBytes, index: usize) -> *mut TraceBytes {
-    if index >= traces.len() {
-        return std::ptr::null_mut();
-    }
+// Note that per [RFC3391](https://rust-lang.github.io/rfcs/3391-result_ffi_guarantees.html), we
+// can return values as `Option<&/&mut>` that will have the same ABI `*const/*mut TraceBytes`, with
+// `None` corresponding to `null`.
```
paullegranddc: This pattern is used pretty extensively in other ffi crates in libdatadog so I am not sure it needs to be mentioned in a comment here

datadog-sidecar-ffi/src/span.rs
```diff
@@ -173,40 +183,46 @@ pub extern "C" fn ddog_trace_new_span_with_capacities(
...
+        SpanBytes {
+            meta: HashMap::with_capacity(meta_size),
+            metrics: HashMap::with_capacity(metrics_size),
+            ..SpanBytes::default()
+        },
+    )
 }
 
+/// The returned slice is an owned allocation that must be properly freed using
+/// [`ddog_free_charslice`].
```
paullegranddc, line 193: The return type is a bit weird because it's a CharSlice but it's nul terminated, and the null byte is one past the length of the CharSlice... So the allocation is `len + 1` bytes
So I am not sure freeing the it with `ddog_free_charslice` is valid, as we releasing an allocation of `len` bytes

yannham: I think you're right, but the thing is `ddog_span_debug_log` does heap-allocate stuff that needs to be de-allocated. We should either use `len+1` in `free`, but in fact I'm not sure it's guaranteed that `CString` has the same layout as a slice (it has currently, but is it part of the contract?). The safest would be to use `CString::from_raw` instead I believe.

datadog-sidecar-ffi/src/span.rs
bwoebi: Well, it's mut, because ddog_get_span etc. accept &mut too.
So, that's actually right.

bwoebi: Same for the other places where you noted that.

At least bindgen must not produce `*const TraceBytes`.

bwoebi:
Looks good now, except for these todos about the &mut, please remove these comments, then I'll ack.

datadog-sidecar-ffi/src/span.rs
yannham: Thanks for the clarification. Done in latest version

bwoebi (approved):
Thanks, looks indeed a bit cleaner now :-)

PR #1712, opened 2026-03-11
feat: change regex to regex-lite

Aaalibaba42:
# What does this PR do?

1. When using a whole regex engine for trivial parsing, unroll a hand-made parser
2. When trivially convertible to regex-lite, use it instead of regex
3. Figure out what to do for cases that don't fit 1. and 2.

# Motivation

Reduce binary sizes

# Additional Notes

3 places currently are option 3:
- datadog-live-debugger/src/expr_eval.rs
- datadog-ffe/src/rules_based/ufc/models.rs
- libdd-trace-obfuscation/src/replacer.rs

For each of these, the reason is user provided patterns. We should try and figure out if this pattern can contain stuff not supported by regex-lite.

# How to test the change?

Testsuite is comprehensive enough


VianneyRuhlmann (approved):
LGTM, any idea why the benchmark are not showing any size improvement ? Is it because of the remaining "option 3" ?

yannham (commented):
Is the goal of hand-made parsers (1.) (instead of using regex-lite) to get rid entirely of any regex dependency in some crates, is it for performance reasons or both ? While the parsers are reasonably simple, it does have a small cost in term of readability and maintenance IMHO (manual indexing is quite off-by-one-error-prone). Just wondering about the overall trade-off.

datadog-live-debugger/Cargo.toml
```diff
@@ -14,6 +14,7 @@ http-body-util = "0.1"
 bytes =  "1.11.1"
 
 regex = "1.9.3"
+regex-lite = "0.1"
```
paullegranddc: pin to minor version maybe? So we never downgrade by mistake

```suggestion
regex-lite = "0.1.9"
```

yannham (approved):

ekump (changes requested):
Is this a PR just for discussion? If so, it should be moved to draft. If your intention is to merge this, then I don't approve for the following reasons:

1) It doesn't actually remove `regex` as a dependency. This can't be merged unless the dependency can be removed, or the use of `regex` / `regex-lite` can be controlled via a feature, and we have a realistic scenario where someone would opt for using `regex-lite`. I don't see the point in having both `regex` and `regex-lite` in our dependency tree at the same time.
1) When we plug normalization and obfuscation into stats computation I believe [this function](https://github.com/DataDog/libdatadog/blob/2afbf2118494ccb4766e6b68f9312714e937d655/libdd-trace-obfuscation/src/replacer.rs#L116) is going to run on on every span for every tag. I don't think the performance degradation of regex-lite will be acceptable here. 
1) I'm fairly confident that unicode support is needed. I know that metrics intake supports unicode characters on tags. Additionally, the sensitive data scanner supports unicode. That isn't something implemented client-side in libdatadog yet, but implies unicode is supported. The python dogstatsd client supports unicode [here](https://github.com/DataDog/datadogpy/blob/80a5445cc85c7709954b5cb7a1eedfc0d55231af/datadog/util/format.py#L13). Also, [looking at the agent](https://github.com/DataDog/datadog-agent/blob/5a007aeb01ea7f0dce84cb1400867db9bdd62d69/comp/trace/config/impl/setup.go#L739-L755) it doesn't appear to be configured to reject / replace unicode (@ichinaski can you confirm I'm understanding the agent code correctly?)

Aaalibaba42:
#1939 took over

PR #1733, opened 2026-03-16
refactor(profiling): split upload function

KowalskiThomas:
## What does this PR do?

This PR splits the Exporter creation process into two steps (two functions to call):
1. Create the Tokio runtime with `init_runtime`
2. Upload the Profile with `send_blocking` (`send_blocking` calls `init_runtime` if it hasn't been initialised from the outside)

This fixes a TSan-detected race condition where we try to cancel the Upload request (using the Cancellation Token) at the same time the runtime is being initialised.   
Example runtime Profiler fix PR: https://github.com/DataDog/dd-trace-py/pull/16467/s

The PR also updates the existing `exporter.cpp` example to use the new function (tested with `cargo ffi-test --filter exporter`).

## Open questions

Currently, the PR maintains compatibility with existing usages -- not explicitly calling `init_runtime` will make `send_blocking` call it when it needs it. However, this is risky (for the reasons that made us make this PR in the first place).
* Should we prevent people from doing this? _Doing so will break compatibility with current usages._
* ... Or should we instead keep the race? _Doing so will guarantee backwards compatibility but also not force usages to be fixed._

--- We ended up going for the second, keeping the race so that we don't have a breaking API change.

### DIFF
--- a/examples/ffi/exporter.cpp
+++ b/examples/ffi/exporter.cpp
@@ -146,6 +146,18 @@ int main(int argc, char *argv[]) {
   auto cancel = ddog_CancellationToken_new();
   auto cancel_for_background_thread = ddog_CancellationToken_clone(&cancel);
 
+  // Eagerly initialize the tokio runtime. This is optional, but required
+  // to avoid race conditions if another thread might be using the
+  // the cancellation token at the same time as the profile is being sent
+  // (as is the case here).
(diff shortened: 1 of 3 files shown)


r1viollet (approved):
LGTM

taegyunkim (approved):

PR #1770, opened 2026-03-20
feat(library-config)!: make libdd-library-config no_std compatible

pawelchcki:
### What does this PR do?

Make `libdd-library-config` and `libdd-library-config-ffi` usable in `no_std` environments.

### Motivation

Enable library config parsing in contexts without a full std runtime (e.g. injected into processes).

### Key changes

- Replace `serde_yaml` with [`yaml_serde`](https://github.com/yaml/yaml-serde) (no_std-compatible), `HashMap` with `BTreeMap`
- Add `no_std_entry` feature for standalone staticlib builds (allocator, panic handler)
- Add `ddog_library_configurator_get_from_bytes` FFI function (works in both std and no_std)
- Propagate `std` feature to workspace consumers of `libdd-common-ffi`

Uses a fork of yaml_serde pending upstream no_std support: https://github.com/yaml/yaml-serde/pull/7

### How to test

```
cargo check -p libdd-library-config --no-default-features
cargo check -p libdd-library-config-ffi --no-default-features
cargo check -p libdd-library-config-ffi --all-features
cargo check --workspace
```


libdd-library-config/src/lib.rs
```diff
@@ -495,12 +512,17 @@ impl Configurator {
...
         } else {
-            match serde_yaml::from_slice(buf) {
-                Ok(config) => config,
-                Err(e) => return LoggedResult::Err(e.into()),
+            let s = match core::str::from_utf8(buf) {
+                Ok(s) => s,
+                Err(e) => return LoggedResult::Err(anyhow::Error::msg(e)),
+            };
+            match yaml_peg::serde::from_str::<StableConfig>(s) {
+                Ok(mut docs) => docs.remove(0),
```
bantonsson: The call `yaml_peg::serde::from_str::<StableConfig>(s)` will return an empty `Vec` for a comment only yaml file, making this `panic` in `Vec.remove`. The old code returned `StableConfig::default()`.


```suggestion
                Ok(mut docs) => docs.into_iter().next().unwrap_or_default(),
```

bantonsson (dismissed):
I'm not super confident about the changes either. Besides, is there a need for `no_std` right now, or is it just a nice to have?

pawelchcki:
image

libdd-library-config-ffi/src/lib.rs
```diff
@@ -181,42 +240,119 @@ pub extern "C" fn ddog_library_configurator_with_detect_process_info(c: &mut Con
...
+
+/// Parses library configuration from raw YAML bytes (local and fleet configs).
+///
+/// `process_info` must be set on the configurator before calling this function
+/// (via `ddog_library_configurator_with_process_info`).
+///
+/// Available in both std and no_std builds (unlike `ddog_library_configurator_get` which
+/// reads files from disk and requires std).
+#[no_mangle]
+pub extern "C" fn ddog_library_configurator_get_from_bytes(
```
pawelchcki: leftover, removing

pawelchcki:
@copilot fix lint/rusfmt error

pawelchcki:
@copilot merge in latest main branch

libdd-common-ffi/src/slice.rs
```diff
@@ -1,25 +1,50 @@
...
 pub enum SliceConversionError {
+    #[error("length was too large")]
     LargeLength,
+    #[error("null pointer with non-zero length")]
     NullPointer,
+    #[error("pointer was not aligned for the type")]
     MisalignedPointer,
 }
 
+#[cfg(feature = "std")]
```
yannham, line 53: Why does this need to be std-gated?

libdd-library-config-ffi/src/lib.rs
```diff
@@ -1,15 +1,62 @@
...
+            fn abort() -> !;
+        }
+        // SAFETY: abort() is a C standard library function with no preconditions.
+        unsafe { abort() }
+    }
+
+    /// Required by the Rust compiler's exception handling ABI. A no-op is safe because
+    /// unwinding will never occur under `panic = "abort"` (enforced by the compile_error!
+    /// guard above). WARNING: this symbol is globally visible — this library must not be
+    /// linked with other Rust code compiled with `panic = "unwind"`.
```
yannham, line 38: This sounds like a pretty big footgun in practice. I'm not sure there's a way to enforce that/warn in any way, though...

libdd-library-config-ffi/src/lib.rs
```diff
@@ -31,36 +78,38 @@ macro_rules! catch_panic {
...
 pub enum LibraryConfigLoggedResult {
     Ok(OkResult),
     Err(Error),
 }
 
 // TODO: Centos 6 build
-// Trust me it works bro 😉😉😉
+// Trust me it works bro
 // #[cfg(linux)]
 // std::arch::global_asm!(".symver memcpy,memcpy@GLIBC_2.2.5");
```
yannham: I know this is unrelated, but is this relevant anymore? Should we clear this dead code/comment?

libdd-library-config-ffi/src/lib.rs
```diff
@@ -82,29 +131,30 @@ pub struct LibraryConfig {
 }
 
 impl LibraryConfig {
-    fn rs_vec_to_ffi(configs: Vec<lib_config::LibraryConfig>) -> anyhow::Result<ffi::Vec<Self>> {
+    fn vec_to_ffi(
```
yannham: Nitpick: name change seems unrelated to the no_std business

libdd-library-config-ffi/src/lib.rs
```diff
@@ -255,6 +314,7 @@ pub extern "C" fn ddog_library_config_local_stable_config_path() -> ffi::CStr<'s
     })
 }
 
+#[cfg(feature = "std")]
```
yannham: I wonder if we could reduce the clutter by having a `std` module (or another name, if that clashes with `std`) that is cfg-gated once and then `pub use` it instead of repeating the guard. As we do for platform-specific implementations

libdd-library-config/src/lib.rs
```diff
@@ -395,7 +416,7 @@ pub struct LibraryConfig {
 }
 
 #[derive(Debug)]
-/// This struct is used to hold configuration item data in a Hashmap, while the name of
+/// This struct is used to hold configuration item data in a BTreeMap, while the name of
```
yannham: It's an implementation detail after all, as this PR demonstrates 😛 

```suggestion
/// This struct is used to hold configuration item data in a map, while the name of
```

libdd-library-config/src/lib.rs
```diff
@@ -741,7 +732,7 @@ impl Configurator {
                         LibraryConfigVal {
                             value: v,
                             source,
-                            config_id: stable_config.config_id.clone(),
+                            config_id: config_id.clone(),
```
yannham: Isn't this cloning the config_id a second time? Also, I think the original version was less verbose/Ok.

libdd-library-config/src/lib.rs
```diff
@@ -779,7 +771,7 @@ impl Configurator {
                 LibraryConfigVal {
                     value,
                     source,
-                    config_id: stable_config.config_id.clone(),
+                    config_id: config_id.clone(),
```
yannham: Same

yannham (commented):
It seems there are two different motivations mentioned in this PR:

1. Artifact size. I would claim here that `no_std` doesn't make a big difference (it can shave off some formatting code IIRC for errors and whatnot), at least not _per se_. Rather the fact that `no_std` basically implies an API surface that is reduced by 90% is likely to be the main contributor to the size reduction. I personally don't have anything against making the crate `no_std`, but if size the main motivation, similar results could probably be achieved with a `minimal` or whatnot feature gate that just hides most symbols (as does `no_std`) without much other changes, and thus be a smaller/uncontroversial change? Another variant could be to create more fine-grained features, so that we can enable a bunch of selected modules _à la carte_.
2. Actually running in a `no_std` environment where some OS-level primitive aren't available. In that case `no_std` is indeed the only solution.

As far as the code goes, mostly LGTM except that maybe we could reduce the `#[cfg(..)]` clutter with appropriate feature-specific modules.

libdd-common-ffi/src/slice.rs
pawelchcki: Its because FfiSafeErrorMessage is in libdd-common and for now is std only 

I didn't want to touch dd-common no_std- althouth maybe I should ?

pawelchcki: No, the changes are too big, I should do it in followup pr

yannham: > Its because FfiSafeErrorMessage is in libdd-common and for now is std only

Ah, makes sense :+1: 

> No, the changes are too big, I should do it in followup pr

Right, small PRs ftw!

pawelchcki:
@codex review

libdd-library-config-ffi/src/lib.rs
pawelchcki: yeah, I need to experiment a alittle with it, but I was worried it would require too many changes and baloon the PR - beyond the functionality I needed

libdd-library-config/src/lib.rs
```diff
@@ -1,14 +1,29 @@
...
+use alloc::vec;
+use alloc::vec::Vec;
+use core::cell::OnceCell;
+use core::mem;
+use core::ops::Deref;
+
+#[cfg(feature = "std")]
+use std::env;
+#[cfg(feature = "std")]
 use std::path::Path;
```
yannham:
```suggestion
#[cfg(feature = "std")]
use std::{env, path::Path};
```

libdd-library-config/src/lib.rs
```diff
@@ -524,92 +540,55 @@ impl Configurator {
             Ok(_) => {}
             Err(e) => return LoggedResult::Err(e.into()),
         }
-        self.parse_stable_config_slice(utils::trim_bytes(&buffer))
+        self.parse_stable_config_slice(&buffer)
     }
 
+    #[cfg(feature = "std")]
```
yannham: It's a bit less severe for this file, so it's a non-blocking suggestion, but I think the same std-specific-module technique can be used in the pure Rust API (`mod no_std_api` or `mod std_api` and then `pub use` -- especially since in Rust you can put different pieces of `impl SomeStruct` in different places, including submodules).

yannham (commented):
Better separation of std/no_std boilerplate now 👍 I don't approve right now because there's the more general question of should we proceed with making some crates `no_std` and what's the plan here (which isn't only my call), but as far the code goes, LGTM

.github/workflows/test.yml
```diff
@@ -141,6 +141,12 @@ jobs:
             # shellcheck disable=SC2086
             cargo build $PACKAGES --verbose
           fi
+      - name: "[${{ steps.rust-version.outputs.version}}] cargo check no_std (library-config)"
```
yannham, line 144: I think we currently don't have a good way to test sets of features (ideally we would use `cargo hack` or something like that). But just to double check if it's the right place/the right move to add a bespoke step for one specific crate and features combination, or if there's a more principled way to do that @ekump ?

PR #1806, opened 2026-03-26
feat: agent-level HTTP client

yannham:
# What does this PR do?

This PR adds a new crate, `libdd-agent-client`, which is a agent-level http client layer over `libdd-http-client`.

# Motivation

The motivation is similar to the HTTP component (#1624): provide a native, unified, lock-free (in the sense of the interpreter lock) and potentially fast way for SDKs to interact with the Datadog agent. The HTTP component is too low-level for that. The agent client automatically handles headers, compression, etc. to offer simple endpoints to the agent.

# Additional Notes

Please be particularly careful about comments, environment variables and header names, and Datadog-specific assumptions in general. Most of the API and the implementation have been LLM-generated from the current usage in `dd-trace-py` in particular (but also other tracers). Unfortunately, I don't always have sufficient context to challenge every part of it.

# How to test the change?

The new crate comes with comprehensive tests.


libdd-agent-client/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+//! | Language metadata headers | `writer.py:638-644`, `stats.py:113-117`, `datastreams/processor.py:128-133` |
+//! | Container/entity-ID headers | `http.py:32-37`, `container.py:157-183` |
+//! | Retry logic (fibonacci backoff) | `writer.py:245-249`, `stats.py:123-126`, `datastreams/processor.py:140-143` |
+//! | Trace send with `X-Datadog-Trace-Count` | `writer.py:749-752` |
+//! | `rate_by_service` parsing | `writer.py:728-734` |
+//! | Stats send | `stats.py:204-228` |
+//! | Pipeline stats send (always gzip) | `datastreams/processor.py:204-210` |
+//! | Telemetry send with per-request headers | `telemetry/writer.py:111-129` |
+//! | EVP event send | `openfeature/writer.py:114-117` |
+//! | `GET /info` with typed result | `agent.py:17-46` |
```
brettlangdon: probably want to cut this 😅

libdd-agent-client/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+//!
+//! # Quick start
+//!
+//! ```rust,no_run
+//! # async fn example() -> Result<(), libdd_agent_client::BuildError> {
+//! use libdd_agent_client::{AgentClient, LanguageMetadata};
+//!
+//! let client = AgentClient::builder()
+//!     .http("localhost", 8126)
```
brettlangdon, line 15: should we bake in the behaviors for env var parsing and etc so this is all picked up automatically for users, unless they want to manually set?

brettlangdon: same below for api key/intake url, etc

brettlangdon: oh, I see in the builder having an "auto detect" option, I guess that is this. nice

libdd-agent-client/src/builder.rs
```diff
@@ -0,0 +1,280 @@
...
+    /// Point the transport to the intake endpoint via [`AgentClientBuilder::https`].
+    pub fn api_key(self, key: impl Into<String>) -> Self {
+        todo!()
+    }
+
+    /// Set the test session token.
+    ///
+    /// When set, `x-datadog-test-session-token: <token>` is injected on every request.
+    /// Replaces dd-trace-py's `AgentWriter.set_test_session_token` (`writer.py:754-755`).
+    pub fn test_token(self, token: impl Into<String>) -> Self {
```
brettlangdon: 🙌🏻

brettlangdon:
```suggestion
    pub fn test_agent_session_token(self, token: impl Into<String>) -> Self {
```

? not super important, but might make it more clear to users

libdd-agent-client/src/builder.rs
```diff
@@ -0,0 +1,280 @@
...
+
+/// Connection mode for the underlying HTTP client.
+///
+/// # Correctness note
+///
+/// The Datadog agent has a low keep-alive timeout that causes "pipe closed" errors on every
+/// second connection when connection reuse is enabled. [`ClientMode::Periodic`] (the default)
+/// disables connection pooling and is **correct** for all periodic-flush writers (traces, stats,
+/// data streams). Only high-frequency continuous senders (e.g. a streaming profiling exporter)
+/// should opt into [`ClientMode::Persistent`].
```
brettlangdon: oh....... we do keep-alive by default in Python, should we not be?

libdd-agent-client/src/builder.rs
```diff
@@ -0,0 +1,280 @@
...
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub enum ClientMode {
```
brettlangdon: instead of ClientMode, what about just a builder method for `use_keep_alive(enabled: bool)` default of `false`

?

libdd-agent-client/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+//! # Agentless mode
+//!
+//! Set an API key via [`builder::AgentClientBuilder::api_key`] and point the transport to the
+//! intake endpoint:
+//!
+//! ```rust,no_run
+//! # async fn example() -> Result<(), libdd_agent_client::BuildError> {
+//! use libdd_agent_client::{AgentClient, LanguageMetadata};
+//!
+//! let client = AgentClient::builder()
```
brettlangdon: I wonder if we should explicitly split an agentless client from a normal agent client.

Since the requirements for building are probably different, and can we enforce which API endpoints are available and the expected payload formats/responses/etc

we might be able to better control from the builder requirements as well, like agentless MUST have an API key and be using a public intake endpoint.

brettlangdon (commented):
some initial thoughts/questions, in general lgtm

libdd-agent-client/src/builder.rs
yannham: Sounds simpler and as good 👍

libdd-agent-client/src/lib.rs
yannham: Agreed. The fact that this PR is called agent-level HTTP client should imply that the focus is on the agent indeed 😅

libdd-agent-client/src/builder.rs
yannham: Maybe? I don't really know to be honest, but Claude picked this from the following comment in `libdd_common::http_common`: https://github.com/DataDog/libdatadog/blob/91e071354744d5b2a3987a8d6726ec51dad43dff/libdd-common/src/http_common.rs#L118

libdd-agent-client/src/lib.rs
yannham: I'm not sure yet about the API of the auto-detect, but yeah the auto-detect feature makes sense. It should probably be listed in the examples, though

yannham: So, after discussing with @paullegranddc and @VianneyRuhlmann it seems it's a bad idea to do too much here because:

1. there might be different configuration sources (env variables + others), and libdatadog can't know that. It's probability better to shift the "configuration resolution" responsibility to the SDK.
2. Reading environment variable in Rust where embedded into another language is full of traps. Like threads, we want to avoid doing that silently or even at all.

So I removed the whole `auto_config()` business in the end. If I misunderstood what you were proposing, or if some pure helpers (for parsing typically, which would get the config field as a string) could be added, please let me know!

libdd-agent-client/src/builder.rs
ajgajg1134: The trace-agent definitely supports keep-alives 🤔 you can safely use them. I think the "gotcha" here is that you shouldn't set the idle connection timeout longer than 60s as then you'll run into this issue where the server closed the idle connection and when you go to try and use it from a conn pool it will break

ajgajg1134: (EDIT): after some more googling - I think you can use keep-alives just fine, the 90s limit in the go tracer is hypothetically not good, but I think in practice doesn't matter due to periodic span flushing that limits connections actually ever being idle that long. So I would allow keep-alives and just set the limit to something like 30s or 45s

errr maybe this actually doesn't matter... the go tracer uses the default from go of 90s idle timeouts....

yannham: Thanks for the info! So the original comment mentions:

```
/// This client does not keep connections because otherwise we would get a pipe closed
/// every second connection because of low keep alive in the agent.
///
/// This is on general not a problem if we use the client once every tens of seconds.
```

Do you think it's outdated @ajgajg1134 , or just plain wrong? Should we not care about connection pooling in this API?

yannham:
@codex review

libdd-agent-client/src/language_metadata.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Debug, Clone)]
+pub struct LanguageMetadata {
+    /// Value of `Datadog-Meta-Lang`, e.g. `"python"`.
+    pub language: String,
+    /// Value of `Datadog-Meta-Lang-Version`, e.g. `"3.12.1"`.
+    pub language_version: String,
+    /// Value of `Datadog-Meta-Lang-Interpreter`, e.g. `"CPython"`.
+    pub interpreter: String,
+    /// Value of `Datadog-Meta-Tracer-Version`, e.g. `"2.18.0"`.
+    pub tracer_version: String,
```
Aaalibaba42, line 19: Very nit: Just thinking out loud but even though those are effectively static in the tracers' perspective, since we go through the ffi we can't mark them as such, `'a` would be viral and clumbersome, and CoW<'static, str> would add a little verbosity on call sites for 4 small heap allocations so are not worth it

Aaalibaba42 (commented):
I didn't see anything wrong :c

libdd-agent-client/src/builder.rs
```diff
@@ -0,0 +1,338 @@
...
+
+        let mut headers = Vec::new();
+
+        if let Some(container_id) = entity_id::get_container_id() {
+            headers.push(("Datadog-Container-Id".to_string(), container_id.to_owned()));
+        }
+
+        if let Some(entity_id) = entity_id::get_entity_id() {
+            headers.push(("Datadog-Entity-ID".to_string(), entity_id.to_owned()));
+        }
```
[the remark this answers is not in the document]
yannham: Is there effectively a need for `datadog-external-env`? In any case if it's needed we don't want to fetch it from the environment, it must be propagated to by the FFI caller.

libdd-agent-client/src/language_metadata.rs
yannham: Right, especially since language metadata is only built once per client. I suppose most runtimes would just keep build one agent client on the side and keep it around. Another way would be to fetch this from `TracerMetadata` maybe somehow? Not entirely sure

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,343 @@
...
+            TraceFormat::MsgpackV5 => ("/v0.5/traces", "application/msgpack"),
+            TraceFormat::MsgpackV4 => ("/v0.4/traces", "application/msgpack"),
+        };
+
+        let mut request = HttpRequest::new(HttpMethod::Put, format!("{}{}", self.base_url, path))
+            .with_body(payload)
+            .with_headers(self.static_headers.iter().cloned())
+            .with_header("Content-Type", content_type)
+            .with_header("X-Datadog-Trace-Count", trace_count.to_string())
+            .with_header("Datadog-Send-Real-Http-Status", "true");
```
iunanua: should we use `"1"` instead `"true"` to be aligned with [libdd-trace-utils](https://github.com/DataDog/libdatadog/blob/main/libdd-trace-utils/src/send_data/mod.rs#L712)?

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,343 @@
...
+        };
+
+        let container_tags_hash = header("datadog-container-tags-hash");
+        let state_hash = header("datadog-agent-state");
+
+        let info: InfoResponse =
+            from_slice(response.body()).map_err(|e| SendError::Encoding(e.to_string()))?;
+
+        Ok(Some(AgentInfo {
+            endpoints: info.endpoints.unwrap_or_default(),
```
iunanua: should we return an Error instead of an empty list?

Cargo.toml
```diff
@@ -51,6 +51,7 @@ members = [
   "libdd-tinybytes",
   "libdd-dogstatsd-client",
   "libdd-http-client",
+  "libdd-agent-client",
```
iunanua, line 56: do we need to update `.codecov.yml`?

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,343 @@
...
+        opts: TraceSendOptions,
+    ) -> Result<AgentResponse, SendError> {
+        let (path, content_type) = match format {
+            TraceFormat::MsgpackV5 => ("/v0.5/traces", "application/msgpack"),
+            TraceFormat::MsgpackV4 => ("/v0.4/traces", "application/msgpack"),
+        };
+
+        let mut request = HttpRequest::new(HttpMethod::Put, format!("{}{}", self.base_url, path))
+            .with_body(payload)
+            .with_headers(self.static_headers.iter().cloned())
```
iunanua, line 81: `self.static_headers` are cloned several times, could we avoid it?

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,343 @@
...
+        trace_count: usize,
+        format: TraceFormat,
+        opts: TraceSendOptions,
+    ) -> Result<AgentResponse, SendError> {
+        let (path, content_type) = match format {
+            TraceFormat::MsgpackV5 => ("/v0.5/traces", "application/msgpack"),
+            TraceFormat::MsgpackV4 => ("/v0.4/traces", "application/msgpack"),
+        };
+
+        let mut request = HttpRequest::new(HttpMethod::Put, format!("{}{}", self.base_url, path))
```
iunanua, line 79: and build full URL string just once? to avoid `format!("{}{}", self.base_url, path)` on each request

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,343 @@
...
+    /// controls the target intake (injected as `X-Datadog-EVP-Subdomain`); `path` is the
+    /// endpoint on that intake (e.g. `/api/v2/exposures`).
+    pub async fn send_evp_event(
+        &self,
+        subdomain: &str,
+        path: &str,
+        payload: Bytes,
+        content_type: &str,
+    ) -> Result<(), SendError> {
+        let request = HttpRequest::new(HttpMethod::Post, format!("{}{}", self.base_url, path))
```
iunanua: should we validate that `path` has a leading `'/'`?

libdd-agent-client/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+    #[error("HTTP error {status}: {body:?}")]
+    HttpError {
+        /// HTTP status code returned by the server.
+        status: u16,
+        /// Raw response body.
+        body: Bytes,
+    },
+    /// All retry attempts exhausted without a successful response.
+    #[error("retries exhausted: {last_error}")]
+    RetriesExhausted {
```
iunanua: Is `SendError::RetriesExhausted` used?

iunanua (commented):
minor things

libdd-agent-client/src/client.rs
yannham: Totally, thanks for the catch. I don't where did Claude get `true` here, it's `1` everywhere else in the tracers.

Cargo.toml
yannham: Mmh, there are a bunch of missing crates here. What is the rationale to add crates to `.codeconv.yml`? Should they just all be added by default?

libdd-agent-client/src/client.rs
yannham: Currently the underlying interface of `libdd-http-client` stores headers as owned strings. In order to allow for borrowed data, we need to change that interface -- which is fine, but I'd rather do it in a follow up, given that the current PR is already meaty.

libdd-agent-client/src/client.rs
yannham: This, on the other hand, we can do only in libdd-agent-client indeed.

yannham: My bad, in fact we can't either (needs to rework libdd-httl-client). Will include in a follow up

libdd-agent-client/src/error.rs
yannham: Nope, thanks for the catch. The underlying http component will just return the last HTTP error when all the retries failed.

libdd-agent-client/src/builder.rs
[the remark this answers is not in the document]
VianneyRuhlmann: >Is there effectively a need for datadog-external-env

Yes it's used by container origin detection https://datadoghq.atlassian.net/wiki/spaces/CONTP/pages/4424270224/Origin+Detection#Related-RFCs,-Engineering-Briefs-and-Technical-documents

> it must be propagated to by the FFI caller

Right now all requests from libdatadog automatically include it which is nice because tracer don't have to implement it. I think this shouldn't be handled by the client. Is there an issue with fetching env variable from libdatadog ?

libdd-agent-client/src/agent_info.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Types for [`crate::AgentClient::agent_info`].
+
+/// Parsed response from a `GET /info` probe.
+///
+/// Returned by [`crate::AgentClient::agent_info`]. Contains agent capabilities and headers.
+#[derive(Debug, Clone)]
+pub struct AgentInfo {
```
VianneyRuhlmann, line 10: Can you re-use the struct from libdd-data-pipeline::agent_info::schema (or move it there) ?

libdd-agent-client/src/builder.rs
```diff
@@ -0,0 +1,338 @@
...
+        builder.build()
+    }
+
+    fn build_static_headers(
+        language: LanguageMetadata,
+        test_token: Option<String>,
+        extra_headers: Vec<(String, String)>,
+    ) -> Vec<(String, String)> {
+        let user_agent = language.user_agent();
+        let mut headers = vec![
```
VianneyRuhlmann, line 255: You're missing `Datadog-Meta-Lang-Interpreter-vendor`. Also I wonder if you could re-use the `TracerHeaderTags` from trace-utils

libdd-agent-client/src/builder.rs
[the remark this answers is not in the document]
yannham: >  Is there an issue with fetching env variable from libdatadog ?

libc's `setenv` and `getenv` are fundamentally not thread-safe. Rust adds its own lock, so as long as you fully stay within pure Rust, it's fine to read or write the environment. However, in a multi-language context such as in libdatadog, any external non-Rust code (native or XXX language runtime) that calls to `setenv` concurrently as you read the environment is unsafe, and this has happened already, see for example https://github.com/DataDog/libdatadog/pull/1930. A bit like spawning OS threads, we should avoid reading the environment silently at a random point in time as much as possible. One solution on Linux is to read `/proc/self/environ` which is snapshotted at the beginning of the program but isn't subject to concurrent updates, or to read the environment in some init function that is required to run in some setup step, where the risk is low.

Since the SDK will have to read the environment anyway for various Datadog-specific variables, I think it's safer to let them read this one as well, but it's indeed a bit less ergonomic.

libdd-agent-client/src/traces.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/// Per-request options for [`crate::AgentClient::send_traces`].
+#[derive(Debug, Clone, Default)]
+pub struct TraceSendOptions {
+    /// When `true`, appends `Datadog-Client-Computed-Top-Level: yes`.
+    ///
+    /// Signals to the agent that the client has already marked top-level spans, allowing the agent
+    /// to skip its own top-level computation.
+    pub computed_top_level: bool,
```
VianneyRuhlmann, line 26: Should we also include the client_computed_stats header. It's used when the tracer already computed stats with CSS, not sure it's going to be used but since you have a send_stats method it would make sense

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    /// Send span stats (APM concentrator buckets) to `/v0.6/stats`.
+    pub async fn send_stats(&self, payload: Bytes) -> Result<(), SendError> {
+        let request = HttpRequest::new(HttpMethod::Put, format!("{}/v0.6/stats", self.base_url))
+            .with_body(payload)
+            .with_headers(self.static_headers.iter().cloned())
+            .with_header("Content-Type", "application/msgpack");
+
+        let response = self.http.send(request).await?;
```
VianneyRuhlmann, line 124: nit: According to CSS spec we shouldn't use retry on stats payload we currently do in the stats exporter. I wonder if from an API pov we should provide a way to configure the retry strategy for each request.

VianneyRuhlmann (approved):
A few notes but overall LGTM

libdd-agent-client/src/agent_info.rs
yannham: Would that be ok to move it here and make libdd-data-pipeline depends on libdd-agent-client? I'm asking because the former looks much bigger, with a deeper dependency tree than the agent client. If so I would actually do that as a later consolidation step.

libdd-agent-client/src/builder.rs
yannham: Thanks for the pointer. The issue is that `TracerHeaderTags` seem to use borrowed data, and `http` types. Which makes me wonder if we shouldn't have used `http` in the first place, instead of redefining our basic http types in `libdd-http-client`, which are a bit inflexible (rely on String/owned data). For now I think I'll duplicate the logic here (still adding `Datadog-Meta-Lang-Interpreter-vendor`) but I'll consider consolidating with `TracerHeaderTags` and `http` type in a follow-up PR maybe.

libdd-agent-client/src/builder.rs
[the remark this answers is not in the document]
yannham: After discussion, we decided to use `get_env` here _because it's the current behavior of libdatadog anyway, and we don't want to lose features by requiring tracers to provide this data_. However, I'll add something on the backlog to get rid of it because it's a trap waiting to trigger...

libdd-http-client/src/request.rs
```diff
@@ -134,6 +134,18 @@ impl HttpRequest {
...
     }
 
+    /// Append headers to this request.
+    #[inline]
+    pub fn with_headers<K, V>(mut self, it: impl IntoIterator<Item = (K, V)>) -> Self
+    where
+        K: Into<String>,
+        V: Into<String>,
+    {
+        self.headers
```
ekump, line 144: Do we need to deduplicate headers somewhere?

yannham: I tend to think that http client is too "low-level" for that, and this should be handled at call site IMHO. Otherwise that would mean paying a tax for the vast majority of cases where headers aren't going to be duplicated. Additionally, quickly looking at the hyper implementation, it seems they already use a custom data structure for headers that merge multiple values of identical headers, so we would duplicate the work.

libdd-agent-client/src/builder.rs
```diff
@@ -0,0 +1,335 @@
...
+        let transport = self.transport.ok_or(BuildError::MissingTransport)?;
+        let language = self.language.ok_or(BuildError::MissingLanguageMetadata)?;
+        let timeout = self
+            .timeout
+            .unwrap_or(Duration::from_millis(DEFAULT_TIMEOUT_MS));
+        let retry = self.retry.unwrap_or_else(default_retry_config);
+
+        let http =
+            Self::build_http_client(transport, timeout, retry, self.allow_connection_pooling)
+                .map_err(|e| BuildError::HttpClient(e.to_string()))?;
```
ekump: minor: `HttpClientError` implements `std::error::Error`, but we lose the error chain when stringifying it like this. Is that being done intentionally to avoid exposing `HttpClientError` through the agent client's API?

libdd-agent-client/src/error.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/// Errors that can occur when sending a request via [`crate::AgentClient`].
+#[derive(Debug, Error)]
+pub enum SendError {
+    /// Connection refused, timeout, or I/O error.
+    #[error("transport error: {0}")]
+    Transport(#[source] std::io::Error),
+    /// The server returned an HTTP error status. Includes the raw status and body.
+    #[error("HTTP error {status}: {body:?}")]
```
ekump: I don't know how large the body is for a typical agent error response, but do you want the whole thing in the display for an error? If it's accessible via the body field, would it make sense to truncate the body in the display?

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,359 @@
...
+                .map(|(_, v)| v.clone())
+        };
+
+        let container_tags_hash = header("datadog-container-tags-hash");
+        let state_hash = header("datadog-agent-state");
+
+        let info: InfoResponse =
+            from_slice(response.body()).map_err(|e| SendError::Encoding(e.to_string()))?;
+
+        let endpoints = info.endpoints.ok_or(SendError::MissingEndpointsInfo)?;
```
ekump, line 229: Is erroring when there are no endpoints returned by `/info` the correct behavior? I believe the only time it would be empty is if it's an old version of the agent. Would it be preferable to log a warning and use some sane defaults?

libdd-http-client/src/request.rs
ekump: That's a fair point. If we need to handle deduplication somewhere, it should be in the agent client. But I'm ok with not worrying about it at all for now.

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,359 @@
...
+}
+
+/// Parse `rate_by_service` from an agent trace response body.
+fn parse_rate_by_service(body: &Bytes) -> Option<HashMap<String, f64>> {
+    #[derive(serde::Deserialize)]
+    struct TraceResponse {
+        rate_by_service: Option<HashMap<String, f64>>,
+    }
+
+    from_slice::<TraceResponse>(body)
```
ekump: Should we log something (at warning or debug level) if we get malformed rates from the agent?

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    /// Send an event via the agent's EVP (Event Platform) proxy.
+    ///
+    /// The agent forwards the request to `<subdomain>.datadoghq.com<path>`. `subdomain`
+    /// controls the target intake (injected as `X-Datadog-EVP-Subdomain`); `path` is the
+    /// endpoint on that intake (e.g. `/api/v2/exposures`).
+    pub async fn send_evp_event(
+        &self,
+        subdomain: &str,
```
ekump: nit: Should we have a `EvpEventRequest` struct instead of all these positional args?

libdd-agent-client/src/builder.rs
```diff
@@ -0,0 +1,335 @@
...
+use libdd_http_client::RetryConfig;
+
+use crate::{error::BuildError, language_metadata::LanguageMetadata, AgentClient};
+
+/// Default timeout for agent requests.
+pub const DEFAULT_TIMEOUT_MS: u64 = 2_000;
+
+/// Default retry configuration: 2 retries (3 total attempts), 100 ms initial delay,
+/// exponential backoff with full jitter.
+//TODO: Do we really want something different from `RetryConfig::default()` for the agent? The only
```
ekump: I think matching `RetryConfig::default()` is ok? After all, it's just a default, SDKs can set whatever they want.

libdd-agent-client/src/client.rs
```diff
@@ -0,0 +1,359 @@
...
+            body: response.body().clone(),
+        })
+    } else {
+        Ok(())
+    }
+}
+
+/// Gzip-compress `payload` at level 6.
+fn gzip_compress(payload: Bytes) -> Result<Bytes, SendError> {
+    let mut encoder = GzEncoder::new(Vec::new(), Compression::new(6));
```
ekump: nit: maybe make the level a named const?

libdd-agent-client/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+//! ```
+//!
+//! # Fork safety
+//!
+//! The underlying `libdd-http-client` uses the `hickory-dns` DNS resolver by default, which is
+//! in-process and fork-safe.
+
+pub mod agent_info;
+pub mod builder;
+pub mod client;
```
ekump: minor: Should the modules be `pub(crate)`? Seems like we're publicly exposing both module paths and re-exports. Not a big deal, but potentially confusing?

libdd-agent-client/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+//!         "python", "3.12.1", "CPython", "", "2.18.0",
+//!     ))
+//!     .build()?;
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! # Fork safety
+//!
+//! The underlying `libdd-http-client` uses the `hickory-dns` DNS resolver by default, which is
```
ekump: Isn't this only true when using the `reqwest` backend?

libdd-agent-client/src/traces.rs
```diff
@@ -0,0 +1,51 @@
...
+use std::collections::HashMap;
+
+/// Wire format of the trace payload.
+///
+/// Determines both the `Content-Type` header and the target endpoint.
+///
+/// # Format selection
+///
+/// The caller is currently responsible for choosing the format. In practice this means starting
+/// with [`TraceFormat::MsgpackV5`] and downgrading to [`TraceFormat::MsgpackV4`] when the agent
```
ekump: is this accurate? I don't think most tracers default to v05 and fallback to v04. You might want to just to remove that part.

libdd-agent-client/src/traces.rs
```diff
@@ -0,0 +1,51 @@
...
+/// returns 404 or 415 (e.g. on Windows, or when AppSec/IAST is active).
+///
+/// In a future version this negotiation may be moved into the client itself so that format
+/// selection becomes automatic and callers no longer need to track the downgrade state.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TraceFormat {
+    /// `application/msgpack` to `/v0.5/traces`. Preferred format.
```
ekump: similar to my comment above, v0.5 is not "preferred".

ekump (approved):
mostly minor comments, otherwise LGTM

libdd-agent-client/src/builder.rs
yannham: Good point, we're pulling `libdd-http-client` anyway so I don't see a good reason to not expose the more structured underlying error.

libdd-agent-client/src/error.rs
yannham: Well, today I learnt there's a format modifier for this :+1:

libdd-agent-client/src/client.rs
yannham: Good question. I don't have the context to be sure, but since `endpoints` is a vec, maybe it's ok to just let it be empty if there are no endpoints and leave it to the caller to handle that case?

libdd-agent-client/src/client.rs
yannham: Agreed.

libdd-agent-client/src/lib.rs
yannham: I believe it's somehow common to re-export a few selected types at the top-level, but still make the modules public for more symbols. However I'm not sure I like the "re-export symbols at the top-level"; the hierarchy is clear here, so I think I'll keep the `pub mod` and get rid of the `pub use` below.

yannham: Ah, but this is what we do in `libdd-http-client`. Well for consistency I'll rather get rid of the `pub mod` then.

libdd-agent-client/src/traces.rs
yannham: I suppose Claude synthesized this from dd-trace-py. I'll remove it if this isn't accurate.

libdd-agent-client/src/client.rs
yannham: @datadog fix this

libdd-agent-client/src/client.rs
yannham: Would returning a `Result` be ok here? I would be inclined to let the caller handle that (I don't think this library logs anything else anywhere). On the other hand if this is really the default behavior to log something but return `None`, it's arguably also reasonable to include it here. What do you think?

libdd-agent-client/src/client.rs
yannham: Thanks. This has been logged and deferred to the consolidation follow-up PR(s).

libdd-agent-client/src/agent_info.rs
ekump: +1 to @yannham's suggestion. data-pipeline should depend on agent-client, not the other way around.

libdd-agent-client/src/client.rs
ekump: I think that's reasonable. @ajgajg1134 or @ichinaski is my analysis accurate?

libdd-agent-client/src/client.rs
ekump: The logging landscape across libdatadog is inconsistent at the moment. I don't think you need to introduce it right now for consistency's sake. 

I don't mind the idea of returning an error, although that is slightly awkward since the agent is sending a 200 and we successfully submitted the payload.

Thinking about this some more, practically speaking, the only way I can see us getting malformed rates is if the agent changes the payload schema. That's something that would probably only get caught by system tests, and (if the sampling system tests are correctly written) we should see those tests fail due to incorrect behavior. 

I'm ok with leaving this as-is for now.

yannham: > I don't mind the idea of returning an error, although that is slightly awkward since the agent is sending a 200 and we successfully submitted the payload.

It would make sense to return an error for `parse_rate_by_service`, but I see your point that it only moves the problem up, since then `send_traces` still has to decide what to do (return an error but as you say, this isn't a "hard" error). Another solution would be to store a `Result` in the `AgentResponse` for the sampling rate, so that the caller of `send_traces` has more information (empty rates or parsing error). Not entirely sure how idiomatic this is, and how heavy this would be through the FFI.

For the time being, I would say let's log something, actually. It's easy, localized, and we can revisit later if needed, but at least if something go south -- even if highly unlikely -- it won't be entirely silent.

libdd-agent-client/src/client.rs
ajgajg1134: hmm good question, what's the impact of an error here?

I would say that if the agent doesn't return _anything_ for endpoints then the agent version is likely VERY old, a quick search suggests the agent with endpoints /info data was released **April of 2021**.  The serverless bottlecap agent also has been non-empty for this field since they introduced trace support it seems.

So I would say "Crashing" is probably the wrong response here, but if you truly get nothing that would definitely be a really bad sign

libdd-agent-client/src/builder.rs
ajgajg1134: I think this is outdated / wrong, keep-alives should be safe and efficient to use 🤔 You might want to do a quick sanity check though by testing with connection keep alives / pooling enabled and verify with slow traffic / low traffic that it works as expected

libdd-agent-client/src/client.rs
ekump: Thanks @ajgajg1134. In that case I think returning an error is the right move. The SDK can then send that via telemetry logs so it would hopefully get noticed.

libdd-agent-client/src/builder.rs
yannham: Thanks! Since this is how it's handled today in `libdd-common` and requires a double check, I'll merge as it is but add a task for the consolidation to assess if it's not relevant anymore, and in that case, scrap it.

PR #1819, opened 2026-03-30
feat!: integrate obfuscation to the stats exporter [APMSP-2764]

Eldolfin:
# What does this PR do?

Integrate the work that has been done on obfuscate_span into the trace exporter

# Motivation

Let the sdks obfuscate spans before sending them to the agent when possible, reducing the amount of data transfered from the sdks to the agent.

# Additional Notes

# How to test the change?

- [x] system tests
- [x] unit tests

# TODO
- [x] retrieve obfuscation config from /info
- [x] add a test for agent obfuscation version > library obfuscation version (it should not obfuscate)
- [x] ~obfuscate more than just sql and redis~ not for stats


libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -189,32 +210,134 @@ pub(crate) fn handle_stats_disabled_by_agent(
...
+                    span_type, resource, dbms_hint,
+                );
+            let wrapper = ObfuscatedStatSpan {
+                inner: span,
+                obfuscated_resource,
+            };
+            stats_concentrator.add_span(&wrapper);
+        } else {
+            stats_concentrator.add_span(span);
+        }
```
VianneyRuhlmann: I think the obfuscation should be done in the SpanConcentrator in the aggregation key computation. This will allow you to remove the ObfuscatedStatSpan

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -189,32 +210,134 @@ pub(crate) fn handle_stats_disabled_by_agent(
...
         concentrator.set_peer_tags(agent_info.info.peer_tags.clone().unwrap_or_default());
+
+        let new_obfuscation_active = is_obfuscation_active(agent_info);
+        if new_obfuscation_active != current_obfuscation_active {
+            // Drop the concentrator lock before updating the ArcSwap
+            drop(concentrator);
+            client_side_stats.store(Arc::new(StatsComputationStatus::Enabled {
+                stats_concentrator: stats_concentrator.clone(),
+                cancellation_token: cancellation_token.clone(),
+                obfuscation_active: new_obfuscation_active,
```
VianneyRuhlmann: If you want to avoid having to recreate the StatsExporter when obfuscation status is updated, you could use an `Arc<AtomicBool>` or `ArcSwap<AtomicBool>` (since it's not updated often). And share between the trace exporter and the stats exporter.

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -180,31 +212,160 @@ pub(crate) fn handle_stats_disabled_by_agent<H: HttpClientTrait + MaybeSend + Sy
...
+                    sql_obfuscation_mode,
+                };
+                let obfuscated_resource =
+                    libdd_trace_obfuscation::obfuscate::obfuscate_resource_for_stats(
+                        span_type, resource, dbms_hint, config,
+                    );
+                let wrapper = ObfuscatedStatSpan {
+                    inner: span,
+                    obfuscated_resource,
+                };
```
VianneyRuhlmann: The obfuscation logic should live in the SpanConcentrator and be controlled by a `obfuscation_enabled` attribute in the SpanConcentrator (it can use the Arc from the stats exporter if it's easier). I don't think you need the ObfuscatedStatSpan either. This has the benefit of sharing the logic with serverless who also use the SpanConcentrator.

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -216,16 +377,28 @@ pub(crate) fn process_traces_for_stats<T: libdd_trace_utils::span::TraceData>(
...
         }
-        add_spans_to_stats(stats_concentrator, traces);
+        #[cfg(feature = "stats-obfuscation")]
+        let obfuscation_active = if let StatsComputationStatus::Enabled {
+            obfuscation_active, ..
+        } = &**status
+        {
+            obfuscation_active.load(Ordering::Relaxed)
+        } else {
+            unreachable!()
```
VianneyRuhlmann: Use an error! log instead. We want to avoid panicking as much as possible.

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -216,16 +377,28 @@ pub(crate) fn process_traces_for_stats<T: libdd_trace_utils::span::TraceData>(
...
             }
         }
-        add_spans_to_stats(stats_concentrator, traces);
+        #[cfg(feature = "stats-obfuscation")]
+        let obfuscation_active = if let StatsComputationStatus::Enabled {
+            obfuscation_active, ..
+        } = &**status
+        {
+            obfuscation_active.load(Ordering::Relaxed)
+        } else {
```
VianneyRuhlmann: This is a bit convoluted because of the stats-obfuscation feature, I don't have all the background but is it required ?

Eldolfin:
@codex review

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -462,7 +478,13 @@ impl TraceExporterBuilder {
                 shared_runtime,
                 dogstatsd,
                 common_stats_tags: vec![libdatadog_version],
-                client_side_stats: ArcSwap::new(stats.into()),
+                client_side_stats: StatsComputationConfig {
+                    status: ArcSwap::new(stats.into()),
+                    #[cfg(feature = "stats-obfuscation")]
+                    obfuscation_config: Arc::new(ArcSwap::from_pointee(
+                        StatsComputationObfuscationConfig::default(),
```
yannham: Isn't this only imported with the stats-obfuscation feature?

Eldolfin: Yes, `StatsComputationObfuscationConfig` is only needed with the stats-obfuscation feature otherwise it's not relevant. Now that I'm reading it again maybe `StatsComputationObfuscationConfig` is too similar to `StatsComputationConfig` which can make it confusing to read

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -155,47 +185,86 @@ pub(crate) fn stop_stats_computation(
...
+    client_side_stats: &StatsComputationConfig,
+) {
+    if matches!(
+        &**client_side_stats.status.load(),
+        StatsComputationStatus::Enabled { .. }
+    ) {
+        let obfuscation_active = is_obfuscation_active(agent_info);
+        std::borrow::Borrow::<ArcSwap<StatsComputationObfuscationConfig>>::borrow(
+            &client_side_stats.obfuscation_config,
+        )
```
VianneyRuhlmann:
```suggestion
        client_side_stats.obfuscation_config
```
I don't think you need the explicit borrow

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -155,47 +185,86 @@ pub(crate) fn stop_stats_computation(
...
+                .info
+                .config
+                .as_ref()
+                .and_then(|config| {
+                    config
+                        .obfuscation
+                        .as_ref()
+                        .map(|obfuscation_cfg| obfuscation_cfg.sql_obfuscation_mode)
+                })
+                .unwrap_or_default(),
```
VianneyRuhlmann: nit: This can be simplified a bit either by using pattern-matching or a closure and "?" chaining

libdd-data-pipeline/Cargo.toml
```diff
@@ -81,4 +82,8 @@ https = [
     "libdd-trace-utils/https",
     "libdd-dogstatsd-client/https",
 ]
+stats-obfuscation = [
```
VianneyRuhlmann, line 96: I'm not sure we need this to be a cargo-feature, we could probably use a runtime config option for this. If we go with the cargo feature I think it should be a default feature

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -26,7 +26,7 @@ const GRPC_STATUS_CODE_FIELD: &[&str] = &[
 #[derive(Clone, Hash, PartialEq, Eq)]
 /// Represent a stats aggregation key borrowed from span data
 pub(super) struct BorrowedAggregationKey<'a> {
-    resource_name: &'a str,
+    resource_name: String,
     service_name: &'a str,
```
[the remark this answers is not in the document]
VianneyRuhlmann: I'm not sure we need to change this at all you can take a &str from the aggregation key to check if the bucket already exists

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -123,8 +143,28 @@ impl SpanConcentrator {
             if bucket_timestamp < self.oldest_timestamp {
                 bucket_timestamp = self.oldest_timestamp;
             }
+            let mut resource_name = span.resource().to_owned();
+
+            #[cfg(feature = "stats-obfuscation")]
+            if self.obfuscation_config.load().enabled {
+                let dbms_hint: Option<&str> = span.get_meta("db.type");
```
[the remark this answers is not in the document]
VianneyRuhlmann: +1

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -210,7 +210,11 @@ impl<'a> BorrowedAggregationKey<'a> {
     ///
     /// If `peer_tags_keys` is not empty then the peer tags of the span will be included in the
     /// key.
-    pub(super) fn from_span<T: StatSpan<'a>>(span: &'a T, peer_tag_keys: &'a [String]) -> Self {
+    pub(super) fn from_span<T: StatSpan<'a>>(
```
VianneyRuhlmann: It might be better to have a separate `from_obfuscated_span(obfuscated_resource_name, span, peer_tags_key)`

VianneyRuhlmann (approved):

libdd-data-pipeline/Cargo.toml
Eldolfin: For now I think it's better to keep this as both a cargo-feature and runtime option so that we don't add unused code to consumer's binary size that don't need it. For now only one tracer (probably dd-trace-py) will enable this feature for testing and benchmarking and it will later put as default when deemed stable.

PR #1826, opened 2026-03-31
feat(data-pipeline): port dd-trace-rs trace buffer implementation

paullegranddc:
# Motivation

As we implement native spans in ruby in python, we have opportuinities to replace more components of the span pipeline with native code.

One of them i the span buffer which we is before the trace exporter.

# Changes

Add a TraceBuffer implementation.
The TraceBuffer is split in two part, a Sender which allows putting trace chunks in the queue and a background task, spawned on the shared runtime which periodically pulls data from the chunk and submit them using the TraceExporter.


libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,889 @@
...
+        self.sender_notifier.notify_all();
+    }
+}
+
+/// A pluggable export operation for the trace buffer
+///
+/// This allows mapping from the buffered spans to another type, and
+/// calling any method on the trace exporter to send traces
+pub trait Export<T>: Send + Debug {
+    fn export_trace_chunks<'a: 'c, 'b: 'c, 'c>(
```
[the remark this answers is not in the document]
paullegranddc: No, refactored to have only one lifetime

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,851 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
paullegranddc: Added a doc module

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,851 @@
...
+            chunks: Vec::with_capacity(PRE_ALLOCATE_CHUNKS),
+            last_flush: std::time::Instant::now(),
+            span_count: 0,
+            batch_gen,
+            max_buffered_spans,
+        }
+    }
+
+    fn reset(&mut self) {
+        *self = Self::new(self.max_buffered_spans);
```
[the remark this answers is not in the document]
paullegranddc: I changed to reset each field, and `clear` the Vec.
Right now this is not perf sensitive as this method is only used after a process forks but better make the code a bit nicer just in case

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,889 @@
...
+        let chunk_len = chunk.len();
+        if let Err(e @ BatchFullError { spans_dropped }) = state.batch.add_trace_chunk(chunk) {
+            state.metrics.spans_dropped_full_buffer += spans_dropped;
+            return Err(TraceBufferError::BatchFull(e));
+        }
+        state.metrics.spans_queued += chunk_len;
+        let gen = state.batch.batch_gen;
+        if state.batch.span_count() > self.flush_trigger_number_of_spans || self.synchronous_write {
+            state.flush_needed = true;
+            self.waiter.notify_receiver(state);
```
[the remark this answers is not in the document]
paullegranddc: after thinking a bit, I agree that we can spare the notification if flush_needed is already true

bwoebi:
The sidecar has a TraceFlusher - this seems to duplicate this functionality, instead of moving the battle-tested implememtation to some common place and adapting it in a way that it fits both use cases perfectly.

The approach used seems also quite a bit more (unnecessarily?) complicated over simply pushing to a Mutex<Vec> (what the TraceFlusher does) ... and what you seem to be doing as well, a few layers down in Batch. Not sure what the Waiter<T> is about - it feels like what the TraceFlusher impl ultimately solves with a simple ManualFuture.

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+        *span_count = 0;
+
+        *batch_gen = {
+            let mut batch_gen = BatchGeneration::default();
+            batch_gen.incr();
+            batch_gen
+        };
+    }
+
+    fn span_count(&self) -> usize {
```
yannham, line 164: Not very important, but any caller that sees `span_count` is also able to see the field directly, so could access the field instead of going through a getter. I would just access the field directly, FWIW.

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+                spans_dropped: chunk.len(),
+            });
+        }
+        if chunk.is_empty() {
+            return Ok(());
+        }
+
+        let chunk_len = chunk.len();
+        self.chunks.push(chunk);
+        self.span_count += chunk_len;
```
yannham, line 187: Nit: you can move this increment up and avoid storing the length in a variable (well, the only difference is if `push` panics, not sure if it's relevant here).

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+/// will be buffered until:
+/// * The number of spans in the buffer is greater than [`TraceBufferConfig::span_flush_threshold`]
+/// * The time since the last flush is greater than [`TraceBufferConfig::max_flush_interval`]
+/// * [`TraceBuffer::force_flush`] is called. This method triggers a flush, but do not wait for the
+///   flush to be done before returning
+///
+/// # Synchronous mode
+///
+/// If [`TraceBufferConfig::synchronous_writes`] is true and
+/// * Either until the chunks have been flushed the agent
```
yannham, line 220:
```suggestion
/// * Either until the chunks have been flushed to the agent
```

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+/// * Or if `synchronous_writes_timeout` is Some, until the timeout is reached. At which point the
+///   flush might continue in the background
```
yannham, line 222: I don't understand this paragraph. It starts with `If` but it's not clear which part is the `then` part.

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+pub struct TraceBuffer<T> {
+    tx: Sender<T>,
+    /// Enables synchronous exports if Some
```
yannham, line 225: ?
```suggestion
    /// Enables synchronous exports if true
```

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+    ///
+    /// Each batch in the queue will get a generation associated. Generations are strictly
+    /// incremental and processed in order by the background thread.
+    /// When the background thread processes a batch it will increment it's 'last_flushed_batch'
+    /// and an export can wait until the 'last_flushed_batch' is equal to the batch it added it's
+    /// trace chunks to.
```
yannham, line 231: This comment seems out of place

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,851 @@
...
+    }
+}
+
+#[derive(Default)]
+pub struct QueueMetrics {
+    pub spans_dropped_full_buffer: usize,
+    pub spans_queued: usize,
+}
+
+fn channel<T>(
```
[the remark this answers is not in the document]
yannham: After some discussions and experiments, there are peculiarities of the trace buffer that makes it not that much simpler using standard channels (that we trade against atomics to hold additional state, which comes with ordering issues etc.). We can always change this implementation later, but for now, even if it's not the most efficient to have a mutex, we agreed it's a reasonable first step.

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+        &self,
+        flush_gen: BatchGeneration,
+        timeout: Option<Duration>,
+    ) -> Result<(), TraceBufferError> {
+        let cond = |state: &mut SharedState<T>| {
+            state.last_flush_generation < flush_gen && !state.has_shutdown
+        };
+
+        if let Some(timeout) = timeout {
+            if timeout.is_zero() {
```
yannham, line 360: Does this really need to be special-cased? Won't the `wait_timeout_while` handle this already?

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+            let _state = self
+                .waiter
+                .sender_notifier
+                .wait_while(state, cond)
+                .map_err(|_| TraceBufferError::MutexPoisoned)?;
+        }
+        Ok(())
+    }
+
+    fn get_state(&self) -> Result<MutexGuard<'_, SharedState<T>>, TraceBufferError> {
```
yannham, line 383: Very nitpicky-nitpick: maybe this method should be called `lock_state`, or just `state`, but I would not expect a getter to actually block. That being said, the type makes it clear that it's locking a mutex, so it's not very important (same for `get_running_state`)

libdd-data-pipeline/src/trace_buffer/mod.rs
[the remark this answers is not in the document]
yannham: Do you want to leave a comment here for a future TODO, so that it's not lost in limbo? (There's also a similar situation in `trigger_flush` below)

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+struct TraceExporterRunInput<T> {
+    trace_chunks: Vec<TraceChunk<T>>,
+}
+
+pub struct TraceExporterWorker<T> {
+    rx: Receiver<T>,
+    export_operation: Box<dyn Export<T> + Send + Sync>,
+    agent_response_handler: ResponseHandler,
+    config: TraceBufferConfig,
+
```
yannham, line 622:
```suggestion
```

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,930 @@
...
+            // Enable the notify future BEFORE acquiring the lock to avoid lost wakeups:
+            // any notify_waiters() call that fires between enable() and .await is captured.
+            let notified = self.waiter.receiver_notifier.notified();
+            let mut notified = std::pin::pin!(notified);
+            notified.as_mut().enable();
+
+            // The MutexGuard must not be held across .await points
+            let leftover;
+            {
+                let mut state = self.waiter.state.lock().map_err(|_| MutexPoisonedError)?;
```
yannham, line 481: This line comes up a lot. Should we have a smaller helper similar to `get_state` for this?

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,930 @@
...
+        *metrics = QueueMetrics::default();
+        Ok(())
+    }
+
+    async fn receive(&self, timeout: Duration) -> Result<Vec<TraceChunk<T>>, MutexPoisonedError> {
+        loop {
+            // Enable the notify future BEFORE acquiring the lock to avoid lost wakeups:
+            // any notify_waiters() call that fires between enable() and .await is captured.
+            let notified = self.waiter.receiver_notifier.notified();
+            let mut notified = std::pin::pin!(notified);
```
yannham, line 475: Out of curiosity, is there any difference between `tokio::pin!(notified)` and `let mut notifier = std::pin::pin!(notified)`?

libdd-data-pipeline/src/trace_buffer/mod.rs
yannham, line 481: Note: while we do not hold a mutex guard across .await points, this will still block the executor waiting for the mutex, right? Maybe it's ok in this case, but just to make sure this is a conscious decision.

yannham (approved):

libdd-data-pipeline/src/trace_buffer/mod.rs
paullegranddc: In a sense yes, `wait_timeout_while` will create an `Instant::now` and compare the `elapsed` time with the duration we passed to the function.
So if duration is zero, the elapsed will always be greater.
I don't know if the check is really necessary though

libdd-data-pipeline/src/trace_buffer/mod.rs
paullegranddc: The `std::pin::pin! ` macro uses super_let, an unstable feature which allows defining variables with the lifetime of the upper block https://rust-lang.github.io/rust-project-goals/2026/redesigning-super-let.html

Which is why it allows this syntax when `tokio::pin!` doesn't.

Otherwise it's the same, create a variable to have a place for the pinned value, then shadow the binding with a mutable borrow and call Pin::new_unchecked

libdd-data-pipeline/src/trace_buffer/mod.rs
paullegranddc: The comment is not clear but the idea is that all of the critical section protected by this mutex are fairly quick, do not call to blocking code, or get suspended so even if it blocks it should be for a short time. (well under high contention it might take a bit of time) 

Maybe it'd be better to use parking_lot, which Mutex guarantees fairness 🤔

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -0,0 +1,934 @@
...
+
+    /// The maximum number of spans that will be buffered before we drop data
+    pub fn max_buffered_spans(self, max: usize) -> Self {
+        Self {
+            max_buffered_spans: max,
+            ..self
+        }
+    }
+
+    /// The number of spans that will be buffered before we decide to flush
```
brettlangdon, line 73: @paullegranddc 

should this be based on size of payload and not number of spans? since we have payload size limits to the agent, not number of span limits. not all spans are the same size

PR #1873, opened 2026-04-13
feat(capablities)!: sleep & spawn capabilities

Aaalibaba42:
# What does this PR do?

1. Create the capabilities for Sleep and Spawn.
2. Change the terminology of Http Capability stuff to be more in line with current work.
3. Change some things about shared runtime.

# Motivation

Before if we were to spawn tasks on wasm, we would have required a current thread tokio runtime blocking JS event loop until the end of the tokio runtime itself. Now in wasm we delegate tasks to the eventloop itself, making the whole thing non blocking, and enabling stuff like sleeping in a JS compatible way.

# Additional Notes

/

# How to test the change?

https://github.com/DataDog/libdatadog-nodejs/pull/70


libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -11,23 +11,79 @@
...
 }
 
+impl fmt::Debug for WorkerEntry {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("WorkerEntry")
+            .field("id", &self.id)
+            .field("worker", &self.worker)
+            .finish()
+    }
+}
```
VianneyRuhlmann, line 43: How is this different from `derive` ?

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -11,23 +11,79 @@
...
+
+#[cfg(not(target_arch = "wasm32"))]
+impl<T> Future for RuntimeJoinHandle<T> {
+    type Output = T;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<T> {
+        match Pin::new(&mut self.get_mut().0).poll(cx) {
+            Poll::Ready(Ok(val)) => Poll::Ready(val),
+            // JoinError means the task panicked or was aborted. Both unrecoverable.
+            Poll::Ready(Err(e)) => panic!("spawned task failed: {e}"),
```
VianneyRuhlmann, line 64: I think tokio's JoinHandle return a JoinError instead of panicking which is probably better to be able to recover in some cases

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -164,9 +169,9 @@ impl<H: HttpClientTrait, C: FlushableConcentrator> StatsExporter<H, C> {
...
 impl<
-        H: HttpClientTrait + MaybeSend + Sync + Debug + 'static,
-        C: FlushableConcentrator + Send + Debug,
-    > Worker for StatsExporter<H, C>
+        Cap: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static,
+        Con: FlushableConcentrator + Send + Debug,
+    > Worker for StatsExporter<Cap, Con>
 {
     async fn trigger(&mut self) {
         tokio::time::sleep(self.flush_interval).await;
```
bantonsson: What happened here? Is `trigger` now a noop?

libdd-trace-utils/src/send_with_retry/mod.rs
```diff
@@ -130,10 +128,10 @@ pub async fn send_with_retry<H: HttpClientTrait>(
...
         };
 
-        #[cfg(not(target_arch = "wasm32"))]
-        let result = tokio::time::timeout(timeout, client.request(req)).await;
-        #[cfg(target_arch = "wasm32")]
-        let result: Result<Result<_, _>, std::convert::Infallible> = Ok(client.request(req).await);
+        let result = tokio::select! {
+            _ = capabilities.sleep(timeout) => Err(()),
+            r = capabilities.request(req) => Ok(r),
+        };
```
bantonsson, line 135: This behaves differently than `tokio::time::timeout`. On simultaneous completion, `timeout` favors the `request` while `select!` picks one at random.

```suggestion
        let result = tokio::select! {
            biased;
            r = capabilities.request(req) => Ok(r),
            _ = capabilities.sleep(timeout) => Err(()),
        };
```

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
```diff
@@ -135,16 +166,11 @@ impl<T: Worker + MaybeSend + Sync + 'static> PausableWorker<T> {
...
 
-                if let Ok(mut worker) = handle.await {
-                    debug!(?worker, "Worker paused successfully");
-                    worker.on_pause().await;
-                    *self = PausableWorker::Paused { worker };
-                    Ok(())
-                } else {
-                    // The task has been aborted and the worker can't be retrieved.
-                    *self = PausableWorker::InvalidState;
-                    Err(PausableWorkerError::TaskAborted)
```
bantonsson, line 146: Why has the error handIng been removed? Won't the new code panic?

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -298,209 +305,154 @@ impl TraceExporterBuilder {
...
-                AgentInfoFetcher::<H>::new(info_endpoint.clone(), Duration::from_secs(5 * 60));
-            let info_fetcher_handle = shared_runtime.spawn_worker(info_fetcher).map_err(|e| {
+        let _guard = shared_runtime
+            .runtime_handle()
+            .map_err(|e| {
                 TraceExporterError::Builder(BuilderErrorKind::InvalidConfiguration(e.to_string()))
-            })?;
+            })?
+            .enter();
+        let capabilities = C::new_client();
```
bantonsson, line 316: Maybe a comment here that the native `C` is expecting to be in the context of the `tokio` runtime, while the node one doesn't care.

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -139,79 +192,103 @@ impl From<io::Error> for SharedRuntimeError {
...
+
+    /// Spawn a PausableWorker using the provided spawn capability.
     ///
     /// The worker will be tracked by this SharedRuntime and will be paused/resumed
-    /// during fork operations.
+    /// during fork operations (native only).
+    ///
+    /// The runtime handle is extracted internally and passed to [`SpawnCapability::spawn`]
+    /// as the [`SpawnCapability::RuntimeContext`]. On native this constrains the spawner's
+    /// `RuntimeContext` to `tokio::runtime::Handle`; on wasm it is `()`.
```
bantonsson: I think this comment should be on the type `SpawnCapability::RuntimeContext`

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -298,209 +305,154 @@ impl TraceExporterBuilder {
...
 
-            if let Some(bucket_size) = self.stats_bucket_size {
-                stats = StatsComputationStatus::DisabledByAgent { bucket_size };
+        let info_endpoint = Endpoint::from_url(add_path(&agent_url, INFO_ENDPOINT));
+        // On wasm the AgentInfoFetcher is not spawned (it uses tokio::time::sleep
+        // internally), but we still need the ResponseObserver for header checks.
```
bantonsson: This comment is about the block below this one.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -298,209 +305,154 @@ impl TraceExporterBuilder {
...
+        #[cfg(target_arch = "wasm32")]
+        let (_info_fetcher, info_response_observer) =
+            AgentInfoFetcher::<C>::new(info_endpoint, Duration::from_secs(5 * 60));
+
+        #[cfg(not(target_arch = "wasm32"))]
+        if let Some(bucket_size) = self.stats_bucket_size {
+            stats = StatsComputationStatus::DisabledByAgent { bucket_size };
+        }
+
+        #[cfg(all(not(target_arch = "wasm32"), feature = "telemetry"))]
```
bantonsson, line 350: Can't this `#[cfg...]` mess be separated out in some way? This is barely readable.

libdd-trace-stats/src/stats_exporter.rs
bantonsson: Hm... maybe I looked at an old diff... ignore this comment for now.

bantonsson: Shouldn't this use a capability?

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
Aaalibaba42: Tokio's JoinHandle implements `Future<Output = Result<T, JoinError>>`, now I use `Box<dyn Future<Output = T>>` instead for our `WorkerJoinHandle<T>`, so we can't directly use the same pattern.

I could make it so our JoinHandle type also is a Result but `futures::remote_handle()` yields a T and panics on drop-cancel, so this way everything has the same API.

Looking at solutions a little, I guess I can do something along the lines of `std::panic::AssertUnwindSafe(handle).catch_unwind().await` to get back a result here, but I'm not sure it's better.

libdd-data-pipeline/src/trace_exporter/builder.rs
Aaalibaba42: I'm not sure how without making spaghetti, do you have something in mind ?

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -11,23 +11,76 @@
...
+    type JoinHandle<T: MaybeSend + 'static> = RuntimeJoinHandle<T>;
+
+    fn spawn<F, T>(&self, future: F, ctx: &tokio::runtime::Handle) -> RuntimeJoinHandle<T>
+    where
+        F: Future<Output = T> + MaybeSend + 'static,
+        T: MaybeSend + 'static,
+    {
+        RuntimeJoinHandle(ctx.spawn(future))
+    }
+}
```
VianneyRuhlmann, line 82: Should this be replaced by `NativeSpawnCapability`

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
VianneyRuhlmann: > I could make it so our JoinHandle type also is a Result

I think this is the way to go we want to avoid panics and this also allows us to follow through with other workers shutdown rather than panicking and dropping data on all workers.

libdd-shared-runtime/src/shared_runtime/mod.rs
Aaalibaba42: I don't think it should:
1. NativeSpawnCapability lives in libdd-capabilities-impl, which is a leaf crate, libdd-shared-runtime (a core crate) can't depend on it without inverting the layering
2. libdd-capabilities-impl doesn't compile on wasm32, and this module needs to (the non-fork parts are wasm-compatible)
3. We could make the fork methods generic and have the FFI caller pass a spawner in, but it adds API surface for no real benefit, the internal RuntimeSpawner is a 5-line ZST that avoids the dependency while keeping the fork hooks simple

libdd-data-pipeline/src/trace_exporter/builder.rs
yannham: One possibility would be to accept that the two implementations are too different and make two separate versions of `build`. If there's really code duplication maybe some part can be moved into separate helpers?

libdd-capabilities-impl/src/lib.rs
```diff
@@ -8,54 +8,85 @@
...
+use std::time::Duration;
+
+pub use http::NativeHttpClient;
+use libdd_capabilities::http::HttpError;
+pub use libdd_capabilities::HttpClientCapability;
+use libdd_capabilities::MaybeSend;
+pub use libdd_capabilities::SleepCapability;
+pub use libdd_capabilities::SpawnCapability;
+pub use sleep::NativeSleepCapability;
+pub use spawn::{NativeJoinHandle, NativeSpawnCapability};
```
yannham: Nitpick: those could be compacted a bit (`pub use libdd_capabilities::{A, B, C, D}`)

libdd-capabilities-impl/src/lib.rs
```diff
@@ -8,54 +8,85 @@
...
+        Self::new()
     }
+}
 
-    impl HttpClientTrait for NativeCapabilities {
-        fn new_client() -> Self {
-            Self {
-                http: DefaultHttpClient::new_client(),
-            }
+impl NativeCapabilities {
```
yannham, line 42: I wonder if bundling all capabilities together has a cost. For example, does it mean we need to instantiate an http client even if all we want to do is to `sleep`?

libdd-capabilities-impl/src/lib.rs
```diff
@@ -8,54 +8,85 @@
...
+    fn request(
+        &self,
+        req: ::http::Request<bytes::Bytes>,
+    ) -> impl Future<Output = Result<::http::Response<bytes::Bytes>, HttpError>> + MaybeSend {
+        self.http.request(req)
+    }
+}
+
+impl SleepCapability for NativeCapabilities {
+    fn sleep(&self, duration: Duration) -> impl Future<Output = ()> + MaybeSend {
```
yannham, line 75: any reason to not make this an `async` function instead of the explicit `impl Future` ? Is it for the `MaybeSend` bound?

libdd-capabilities-impl/src/spawn.rs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        NativeJoinHandle(ctx.spawn(future))
+    }
+}
+
+/// Newtype wrapping `tokio::task::JoinHandle<T>` that normalises the output to
+/// `T` instead of `Result<T, JoinError>`.
+///
+/// A `JoinError` means the spawned task panicked or was aborted. Workers use
+/// `CancellationToken` for graceful shutdown, so `JoinError` indicates a bug.
```
yannham: It's really a detail, but I think they don't use cancellation tokens any more with the new shared runtime?

libdd-capabilities-impl/src/spawn.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct NativeJoinHandle<T>(JoinHandle<T>);
+
+impl<T> Future for NativeJoinHandle<T> {
+    type Output = T;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<T> {
+        // JoinHandle<T>: Unpin, so Pin::new is safe.
```
yannham: Nit: the safety comment feels a bit unnecessary. There's no unsafe code here, so it wouldn't actually compile if the precondition was false.

```suggestion
```

libdd-data-pipeline/src/trace_exporter/builder.rs
yannham: (or dually move the platform-dependents bit into bespoke functions?)

libdd-shared-runtime/src/shared_runtime/mod.rs
yannham: I agree. We're just wrapping `JoinHandle` which returns a result, so there's no obvious reason to depart from this behavior and introduce panics instead.

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -29,6 +35,45 @@ struct WorkerEntry {
     worker: PausableWorker<BoxedWorker>,
 }
 
+/// Internal spawner for fork recovery paths.
+///
+/// The runtime handle is passed at spawn time via [`SpawnCapability::RuntimeContext`],
+/// not stored, so the spawner is always using the current (potentially rebuilt) runtime.
+#[cfg(not(target_arch = "wasm32"))]
```
yannham, line 25: Instead of all those cfg gates, could we move everything spawner-related to a submodule and gate the submodule once and for all instead? as we do with `xxx::linux` for example elsewhere in libdatadog.

libdd-shared-runtime/src/shared_runtime/mod.rs
yannham: I might be missing something, but when you say:

> NativeSpawnCapability lives in libdd-capabilities-impl, which is a leaf crate, libdd-shared-runtime (a core crate) can't depend on it without inverting the layering

Why should shared_runtime depend on it once we move everything to `libdd-capabilities-impl`? I _think_ (but I'm not sure) @VianneyRuhlmann's question is, or at least the one I have as well, is why this code has to be in `shared_runtime` at all. It naively looks like it's entirely capability and native-specific, so should live in libdd-capabilities-impl. Do you need some private symbols from here?

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -234,6 +299,43 @@ impl SharedRuntime {
...
+        T: Worker + Sync + 'static,
+        S: SpawnCapability<RuntimeContext = ()> + 'static,
+    >(
+        &self,
+        worker: T,
+        restart_on_fork: bool,
+        spawner: &S,
+    ) -> Result<WorkerHandle, SharedRuntimeError> {
+        let boxed_worker: BoxedWorker = Box::new(worker);
+        debug!(?boxed_worker, "Spawning worker on SharedRuntime");
```
yannham, line 429: Is this a left-over ? (might not be, just asking)

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
```diff
@@ -135,16 +166,11 @@ impl<T: Worker + MaybeSend + Sync + 'static> PausableWorker<T> {
...
-                }
+                let mut worker = handle.await;
+                debug!(?worker, "Worker paused successfully");
```
yannham: Same, is this supposed to stay?

yannham (commented):
I have mixed feelings about the complexity and boilerplate added. Part of it is inherent to the capability approach. So the elephant in the room: is it acted and 100% decided that we want to go the WASM route and that we should support that? It has an impact of the whole of libdatadog, and it seems for now that the route for dd-trace-js is not entirely decided yet. It's obviously not one person's decision, but I wonder if we should move forward and merge things if this is not fully committed yet; since it's far-reaching, it might be harder to revert than just a `git revert`, as people will quickly build on it.

Another point is the shared runtime. My understanding was that the whole promise of capabilities (vs a more classical platform-specific swappable module implementation as we do for linux vs windows for example) was to push all of the WASM and node-specific concerns out of libdatadog. However the impression when looking at the shared runtime's changes is that it's full of WASM-specific concerns, which feels like we got the drawbacks of both approaches (complexity + wasm-specific code in libdatadogand cfg gates). Some functions are available on WASM, some others are not. I'm not sure yet what's the conclusion, but I think the interface of the shared runtime should be reworked if we want it to work with capabilities. Maybe the conclusion is that the whole shared runtime should be a capability, since it's offering an API that has fundamentally different implementation on WASM and native (and it seems impossible to delegate that to existing capabilities). Or maybe the answer is that we need to move some of the shared runtime into new capability.ies for async, but can still keep some code in libdd-shared-runtime. In any case, the current situation feels a bit wrong.

libdd-data-pipeline/src/agent_info/fetcher.rs
```diff
@@ -577,7 +580,10 @@ mod single_threaded_tests {
         );
         assert!(agent_info::get_agent_info().is_none());
         let shared_runtime = SharedRuntime::new().unwrap();
-        shared_runtime.spawn_worker(fetcher, true).unwrap();
+        let spawner = NativeCapabilities::new();
```
VianneyRuhlmann: Can we remove the spawner and pass capabilities at shared_runtime init?

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
```diff
@@ -27,6 +35,19 @@ pub enum PausableWorker<T: Worker + MaybeSend + Sync + 'static> {
...
+        match self {
+            Self::Running { .. } => f.debug_struct("PausableWorker::Running").finish(),
+            Self::Paused { worker } => f
+                .debug_struct("PausableWorker::Paused")
+                .field("worker", worker)
+                .finish(),
+            Self::InvalidState => write!(f, "PausableWorker::InvalidState"),
+        }
+    }
+}
```
VianneyRuhlmann, line 67: Why do you need this ?

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
```diff
@@ -54,17 +75,23 @@ impl<T: Worker + MaybeSend + Sync + 'static> PausableWorker<T> {
...
+    ) -> Result<(), PausableWorkerError>
+    where
+        S::JoinHandle<T>: 'static,
+    {
         match self {
             PausableWorker::Running { .. } => Ok(()),
-            PausableWorker::Paused { worker } => {
-                debug!(?worker, "Starting pausable worker");
+            PausableWorker::Paused { .. } => {
+                debug!(?self, "Starting pausable worker");
```
VianneyRuhlmann: Try to avoid (or train your ai to avoid) this kind of unrelated changes which clutter the PR. (If it's actually related disregard)

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
```diff
@@ -119,7 +150,7 @@ impl<T: Worker + MaybeSend + Sync + 'static> PausableWorker<T> {
     /// Pause the worker and wait for it to complete, storing its state for restart.
     ///
     /// # Errors
-    /// Fails if the worker handle has been aborted preventing the worker from being retrieved.
```
VianneyRuhlmann, line 122: No need to update

libdd-shared-runtime/src/lib.rs
```diff
@@ -22,3 +22,16 @@ pub mod worker;
...
+/// On native this is `tokio::runtime::Handle`; on wasm it is `()`.
+/// Generic code that calls [`SharedRuntime::spawn_worker`] should bound
+/// `C: SpawnCapability<RuntimeContext = SpawnRuntimeContext>` so the constraint
+/// is satisfied on every target without `#[cfg]` on individual where clauses.
+#[cfg(not(target_arch = "wasm32"))]
+pub type SpawnRuntimeContext = tokio::runtime::Handle;
+
+/// See the non-wasm variant for documentation.
+#[cfg(target_arch = "wasm32")]
+pub type SpawnRuntimeContext = ();
```
VianneyRuhlmann: Can't you reuse the `NativeSpawnCapability` associated type  ?

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -54,19 +54,24 @@ impl<'a> From<&'a StatsMetadata> for TracerHeaderTags<'a> {
...
+    Con: FlushableConcentrator = SpanConcentrator,
+> {
     flush_interval: time::Duration,
-    concentrator: Arc<Mutex<C>>,
+    concentrator: Arc<Mutex<Con>>,
     endpoint: Endpoint,
     meta: StatsMetadata,
     sequence_id: AtomicU64,
-    client: H,
+    capabilities: Cap,
```
VianneyRuhlmann, line 69: ultra nit: Since we're going to pass the capabilities in a lot of functions it would be nice to have a "standardized" way of passing it e.g. "always the first param" to make it easier to recognize.

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -400,16 +405,17 @@ mod tests {
...
         let buckets_duration = Duration::from_secs(10);
 
+        let caps = NativeCapabilities::new();
         let stats_exporter = StatsExporter::<NativeCapabilities>::new(
             buckets_duration,
             Arc::new(Mutex::new(get_test_concentrator())),
             get_test_metadata(),
             Endpoint::from_url(stats_url_from_agent_url(&server.url("/")).unwrap()),
-            NativeCapabilities::new_client(),
+            caps.clone(),
```
VianneyRuhlmann, line 414: Maybe implementing Copy on NativeCapabilities would make it more ergonomic (since it's zero sized anyway)

libdd-trace-utils/src/send_with_retry/retry_strategy.rs
```diff
@@ -162,9 +136,10 @@ mod tests {
             backoff_type: RetryBackoffType::Constant,
             jitter: None,
         };
+        let sleeper = NativeSleepCapability;
```
VianneyRuhlmann: nit: I think it's better to always call it cap or capabilities than to use a different name depending on the capabilities available

libdd-capabilities-impl/src/lib.rs
Aaalibaba42: I would have expected that to be handled by `cargo fmt`, I'll check if it's something we can enable

libdd-capabilities-impl/src/lib.rs
Aaalibaba42: If a function just needs to sleep, it should not use the bundle, because the NativeHttp trait is not a ZST. Ideally everything should be a ZST, but even in this case, it's not that big a deal since using the bundle is a convenience option and not an obligation, if you just sleep, the function's bound should just be `SleepCapability` and it should just be passed the corresponding ZST.

I think there would be a way to have the HttpCapability a ZST, but that would be out-of-scope here

libdd-capabilities-impl/src/lib.rs
Aaalibaba42: Yes it is for the MaybeSend bound. But given we are in NativeCapabilities I guess MaybeSend will always be Send, and so we could make it explicit async. Do you think it would be better ?

libdd-capabilities-impl/src/lib.rs
Aaalibaba42: The codebase doesn't have any standard on that front from what I can tell, whether you put `crate` or `module` on the rustfmt.toml, there are dozens of files that change.

libdd-data-pipeline/src/agent_info/fetcher.rs
Aaalibaba42: Storing the spawner in SharedRuntime would make it generic (`SharedRuntime<S>`) or require type erasing a non object-safe trait. It's also not straightforward because the fork recovery paths use their own internal RuntimeSpawner, not the caller-provided one. In practice, production callers already have capabilities in scope so passing it to spawn_worker isn't a burden imho

libdd-shared-runtime/src/shared_runtime/mod.rs
Aaalibaba42: The fork methods are impl SharedRuntime and access private fields. They can't move to libdd-capabilities-impl without making those fields public. RuntimeSpawner is only consumed by these methods, so it's co-located with them. Moving just the spawner out wouldn't buy anything since the methods that use it must stay here regardless

libdd-shared-runtime/src/shared_runtime/mod.rs
Aaalibaba42: I did not add this ? I don't know

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
Aaalibaba42: I went from the JoinHandle of `tokio` to ours, and so we can't derive debug automatically, and WorkerEntry needs debug.

libdd-shared-runtime/src/lib.rs
Aaalibaba42: I'm not sure I understand, *this* is the associated type for the NativeSpawnCapabilitiy

libdd-trace-stats/src/stats_exporter.rs
Aaalibaba42: Mmh I think this would be hard to enforce, and it can be conflicting with stuff etc. Generally I agree but I'm not sure it's realistic on a common codebase like this one. But clearly it's far from a hill I wish to die on.

libdd-trace-stats/src/stats_exporter.rs
Aaalibaba42: I was wrong before, the nativehttpcapability has the hyper client to reuse the connection pool iirc. I think that's why this is required.

libdd-capabilities-impl/src/lib.rs
yannham: I don't have a strong opinion, it was mostly a curiosity question :slightly_smiling_face: I would say do what you think makes the most sense to you!

libdd-capabilities-impl/src/lib.rs
yannham: ZST would be one extreme. Another possibility would be to at least avoid heap allocation at instantiation (for example, the client could be an option that is lazily instantiated).

My general sentiment here is if I were a consumer, I would not expect capabilities to actually hold state and heap-allocate, but rather be just a glorified compile-time dispatch. Invitably at some point someone will use `NativeCapabilities` without using the HTTP client (or will be instantiating `NativeCapabilities` on the spot repeatedly thinking it doesn't cost much, etc.). I also takes some control out of the hands of the caller (the client lifecycle).

Another possibility could be to have a bespoke notion of HttpClient in the `HttpCapability` trait - after all the code had to deal with the client abstraction before - so the callsites can decide when to instantiate the client, where to store it, etc. It could even piggy-back onto the new Http client component.

Note that it doesn't have to be in this PR at all, or that people might disagree with my opinion, but I wanted to open the discussion.

libdd-shared-runtime/src/shared_runtime/mod.rs
yannham: Sorry, I don't know why I thought this has been added by this PR. I probably hallucinated

libdd-shared-runtime/src/shared_runtime/mod.rs
yannham: I see, thanks! Now I wonder if we shouldn't just move all of the shared runtime crate into libdd-capabilities 😛 but that's another debate for another day I guess. Current situation sounds reasonable for this PR.

Though, the new refactoring also removed `shutdown` and `block_on` from WASM interface. Is that ok?

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
yannham: Naive question, but would that be simpler to implement a dumb `Debug` on `JoinHandle` instead, and let derive do the rest of the work?

libdd-shared-runtime/src/lib.rs
yannham: I think Vianney meant `NativeSpanCapability::RuntimeContext` instead of duplicating the `tokio::runtime::Handle` definition. Same for WASM, although maybe here you can't because you don't have the WASM cap implementation at hand.

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
```diff
@@ -4,21 +4,29 @@
...
+use std::future::Future;
+use tokio::select;
 use tokio_util::sync::CancellationToken;
 use tracing::debug;
 
+#[cfg(not(target_arch = "wasm32"))]
+type WorkerJoinHandle<T> = Pin<Box<dyn Future<Output = Result<T, SpawnError>> + Send>>;
+
+#[cfg(target_arch = "wasm32")]
+type WorkerJoinHandle<T> = Pin<Box<dyn Future<Output = Result<T, SpawnError>>>>;
```
yannham, line 24: Could the two definitions be merged into one thanks to `MaybeSend`?

libdd-shared-runtime/src/shared_runtime/mod.rs
Aaalibaba42: I think it is, we don't really want to do these since it would amount to blocking the EventLoop (and we can't shut it down). The point is mainly to have a "proxy" for shared runtime when throwing futures on the JS side.

Maybe moving more of the shared_runtime as a capability would make sens, but I don't think we can escape the fact that we would not have the same API on both platforms because they can't do the same things. So except having sub-capabilities for SharedRuntimeCapabilityCapabilities, which is a mouthful (amongst other problems), I don't know how we'd bridge that gap.

libdd-capabilities-impl/src/lib.rs
Aaalibaba42: > ZST would be one extreme. Another possibility would be to at least avoid heap allocation at instantiation (for example, the client could be an option that is lazily instantiated).

Oh that sounds nice indeed, I can do that, it should be fairly straightforward.

And I agree with the rest of your comment, but I also agree it should not be in this PR, it's already not atomic at all lol.

libdd-shared-runtime/src/shared_runtime/mod.rs
Aaalibaba42: I also squashed to make rebasing easier and changed a few thing, maybe before it didn't catch it as a diff or something.

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
Aaalibaba42: Sure I can move that, I don't have strong opinions on it

libdd-shared-runtime/src/lib.rs
Aaalibaba42: Oh okay my bad, yes for native that makes sens and it would be more maintainable. For wasm as yann said it lives outside libdatadog and we can't have a dependancy on libdatadog-nodejs from libdatadog.

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
Aaalibaba42: I tried and no I could not because of reasons chatgpt explained but I forgor. Something about MaybeSend not being a "compiler-blessed Trait" and so it couldn't work here.

libdd-capabilities-impl/src/lib.rs
Aaalibaba42: Okay I made it a `Arc<OnceLock<>>` o/

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
Aaalibaba42: the "official" error: `error[E0225]: only auto traits can be used as additional traits in a trait object`

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
Aaalibaba42: Mmmh up until now the WorkerJoinHandle was a type alias and not a struct itself so I couldn't have implemented Debug on it since it would be not my type neither my trait. I can still make it work by making the WorkerJoinHandle a wrapper around the type, but it might be more of a mouthful since I'd need to wrap it at construction too. What do you think ?

libdd-shared-runtime/src/shared_runtime/mod.rs
yannham: That makes sense. I could see something like having a capability (`AsyncCapability` ? But JS does have async :thinking: ) that would only be implemented by native, but as you say, it doesn't change the state of affairs; it merely pushes the detail into the capability crate (which maybe is still nice to have, but not a fundamental improvement either)

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
yannham: Ah, if it's a type alias I suppose the current situation is fine as well. If it were a wrapper that would just be less code, but I'm not sure it's really worth making a wrapper just to get rid of this smallish implementation... I would say pick the option you like the most, both are fine and relatively unimportant

libdd-capabilities-impl/src/lib.rs
yannham: Sounds good! Maybe we could track the ZST/client abstraction approach in a Jira ticket?

yannham (approved):
There might be some minor comments to be addressed still, but overall it's now in a good state to move forward.

libdd-capabilities-impl/src/spawn.rs
```diff
@@ -0,0 +1,12 @@
...
+
+//! Native spawn implementation is now handled internally by `SharedRuntime`.
+//!
+//! This module is kept for backwards compatibility but the type is no longer
+//! used by capability bundles or consumer code.
+
+/// Marker type retained for backwards compatibility.
+/// Task spawning is now handled internally by `SharedRuntime`.
+#[derive(Clone, Debug)]
+pub struct NativeSpawnCapability;
```
VianneyRuhlmann: Which backwards compatibility ? This was just added

libdd-shared-runtime/Cargo.toml
```diff
@@ -24,5 +24,6 @@ tracing = { version = "0.1", default-features = false }
 libdd-capabilities = { path = "../libdd-capabilities", version = "1.0.0" }
 libdd-common = { version = "4.0.0", path = "../libdd-common", default-features = false }
 
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-tokio = { version = "1.23", features = ["rt-multi-thread"] }
```
VianneyRuhlmann, line 31: You probably want to keep this, this is compiling because of libdd-common pulls it in but it should be explicit

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -14,12 +14,257 @@ use crate::worker::Worker;
...
+            // Lock runtime first to synchronize with before_fork (which does
+            // runtime.take() then workers.lock()), following the documented mutex
+            // lock order. If the runtime has been taken (fork window), skip starting
+            // the worker, after_fork_parent/child will start it on the new runtime.
+            let runtime_handle = self
+                .runtime
+                .lock_or_panic()
+                .as_ref()
+                .map(|rt| rt.handle().clone());
+            let mut workers_guard = self.workers.lock_or_panic();
```
VianneyRuhlmann, line 90: You lost the comment explaining why we keep the workers_guard while starting

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
```diff
@@ -177,6 +198,17 @@ mod tests {
...
+    fn tokio_spawn_fn(
+        handle: &tokio::runtime::Handle,
+    ) -> impl FnOnce(WorkerFuture<Box<dyn Worker + Sync>>) -> WorkerJoinHandle<Box<dyn Worker + Sync>>
+    {
+        let h = handle.clone();
+        move |future| {
+            let jh = h.spawn(future);
+            Box::pin(async { jh.await.map_err(|e| SpawnError::new(e.to_string())) })
+        }
+    }
```
VianneyRuhlmann: nit: Maybe also reuse this function in the SharedRuntime `impl`

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -305,228 +294,164 @@ impl TraceExporterBuilder {
...
+                .spawn_worker(info_fetcher, false)
+                .map_err(|e| {
+                    TraceExporterError::Builder(BuilderErrorKind::InvalidConfiguration(
+                        e.to_string(),
+                    ))
+                })?;
+            (handle, observer)
+        };
+        // On wasm the AgentInfoFetcher is not spawned yet (it requires spawn_local),
+        // but we still need the ResponseObserver for header checks.
```
VianneyRuhlmann: Is it still the case ?

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -14,12 +14,257 @@ use crate::worker::Worker;
...
+
+            // Lock runtime first to synchronize with before_fork (which does
+            // runtime.take() then workers.lock()), following the documented mutex
+            // lock order. If the runtime has been taken (fork window), skip starting
+            // the worker, after_fork_parent/child will start it on the new runtime.
+            let runtime_handle = self
+                .runtime
+                .lock_or_panic()
+                .as_ref()
+                .map(|rt| rt.handle().clone());
```
VianneyRuhlmann: I think we actually need to keep the runtime locked while spawning the worker otherwise, `before_fork` could take ownership of both `runtime` and `workers` once we dropped the runtime guard. This means we would spawn a worker which wouldn't be dropped by `before_fork`.

libdd-data-pipeline/src/trace_exporter/builder.rs
Aaalibaba42: Nop, but now it's blocked by sleeping, I didn't migrate it to use the SleepCapability, I'll do that.

libdd-shared-runtime/src/shared_runtime/mod.rs
Aaalibaba42: Good catch

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -309,29 +309,23 @@ impl TraceExporterBuilder {
...
+        let info_fetcher_handle = shared_runtime
+            .spawn_worker(info_fetcher, false)
+            .map_err(|e| {
+                TraceExporterError::Builder(BuilderErrorKind::InvalidConfiguration(e.to_string()))
+            })?;
+        // The handle is currently only tracked for shutdown on native; on wasm
+        // it is dropped here (the worker keeps running on the JS event loop
+        // until the page/module is torn down).
+        #[cfg(target_arch = "wasm32")]
+        let _ = info_fetcher_handle;
```
VianneyRuhlmann, line 342: Maybe it's ok for JS but just a heads up. In python we sometimes have to recreate the TraceExporter. If the trace exporter doesn't have the handle the worker can't be stopped on exporter shutdown and the worker will keep running in the background

VianneyRuhlmann (approved):
LGTM, congrats on making this work

libdd-data-pipeline/src/trace_exporter/builder.rs
Aaalibaba42: I don't wager that should happen for the wasm module, and if ever the situation arises, this can always be added later on.

PR #1875, opened 2026-04-14
ci(versioning): crates-to-package.sh refactor

iunanua:
# What does this PR do?
Refactor to simplify `crates-to-package.sh`script:

- No more awk parsing: the script now runs cargo metadata --no-deps at two refs (base/head) and compares package.version.
- Uses git worktree (like scripts/major-bumps-level.sh) to run cargo metadata as of each ref.
- Excludes are no longer hardcoded: a crate is included only if it’s a workspace member and publishable per metadata:


hoolioh (approved):
LGTM

PR #1877, opened 2026-04-14
chore(libdd-crashtracker): remove path reference for libdd-libunwind-sys

hoolioh:
# What does this PR do?

Remove `libdd-libunwind-sys` path reference.

# Motivation

Since libdd-libunwind-sys is being moved to another repo the path reference is no longer needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3101,6 +3101,8 @@ dependencies = [
 [[package]]
 name = "libdd-libunwind-sys"
 version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


yannham (commented):
I still see local references of `libdd-libunwind-sys` in the Dockerfile.build file. Is that ok?

hoolioh:
> I still see local references of `libdd-libunwind-sys` in the Dockerfile.build file. Is that ok?

When the migration is fully finished a subsequent PR will follow this one removing libdd-libunwind-sys and their presence in the build scripts.

gleocadie (approved):

yannham (approved):

PR #1878, opened 2026-04-14
feat(crashtracking)!: collect all threads

gyuheon0h:
# What does this PR do?

When a process crashes the crashtracker now captures a full stack trace for every live non-crashing thread and includes them in the crash report under error.threads.

Previously only the crashing thread's stack was reported. Customers had no visibility into what other threads were doing at the time of the crash, making it much harder to diagnose race conditions, deadlocks, and cross-thread interactions.

Thread collection runs entirely in the receiver process after it has finished reading the crash pipe. Because the crashed process stays alive in its signal handler until the receiver calls `finish()`, all threads remain valid ptrace targets for the entire collection window.

The receiver:
1. Enumerates threads from` /proc/<pid>/task/`
2. Attaches to each non-crashing thread with `PTRACE_SEIZE` + `PTRACE_INTERRUPT` 
3. Uses libunwind remote unwinding (`_UPT_create` / `unw_init_remote` / `unw_step_remote`) to walk each thread's stack
4. Thread names and states are read from /proc/<pid>/task/<tid>/stat`
5. Thread collection is opt-in using `CrashtrackerConfiguration::set_collect_all_threads(true)` and bounded by a max of 256. The user can also set a lower boundary using `CrashtrackerConfiguration::set_max_threads()` . The collection timeout is the remaining slice of the receiver's deadline after the crash pipe is fully read, so the total receiver lifetime is always bounded by `config.timeout()`

# Breaking changes
Updated data schema to 1.7; there is now a `Threads` object.

Errors intake upload still sends an array of `ThreadData`, as that is the schema that the enrichment pipeline expects

`CrashtrackerConfig` now has

```
pub struct CrashtrackerConfiguration {
    additional_files: Vec<String>,
    collect_all_threads: bool, <-- NEW
    create_alt_stack: bool,
    demangle_names: bool,
    endpoint: Option<Endpoint>,
    max_threads: usize, <-- NEW 
    resolve_frames: StacktraceCollection,
    signals: Vec<i32>,
    timeout: Duration,
    unix_socket_path: Option<String>,
    use_alt_stack: bool,
}
```

```
config.set_collect_all_threads(true);
config.set_max_threads(128);          // how many threads to collect
```

```
CrashtrackerConfiguration::builder()
    .collect_all_threads(true)
    .max_threads(128)
    .build()?;
```

# Additional notes
One enhancement we can do is to collect even the crashing thread in the receiver along with all threads, since enumerating all threads touches the crashing thread also. We can do this by checking for the crashing thread first in the receiver, then checking all threads. 

However, this will be explored in a future PR/investigation, as this is changing the core crashing thread unwinding logic

# How to test the change?

Run a crash with a program with multiple threads, with all thread collection turned on.

Bin tests also added to test e2e flow

```
{
      "is_crash": true,
      "kind": "UnixSignal",
      "message": "Process terminated with SEGV_MAPERR (SIGSEGV)",
      "thread_name": "crashtracker_bi",
      "source_type": "Crashtracking",
      "stack": {
        "format": "Datadog Crashtracker 1.0",
        "frames": [
          {
            "ip": "0x5d1e89d2adcb",
            "module_base_address": "0x5d1e89cfd000",
            "sp": "0x7ffc9f327330",
            "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
            "build_id_type": "GNU",
            "file_type": "ELF",
            "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
            "relative_address": "0x000000000002ddcb",
            "column": 13,
            "file": "/home/bits/go/src/github.com/DataDog/libdatadog/bin_tests/src/bin/crashtracker_bin_test.rs",
            "function": "cause_segfault",
            "line": 37
          },
         ...
          {
            "ip": "0x5d1e89d29405",
            "module_base_address": "0x5d1e89cfd000",
            "sp": "0x7ffc9f327d70",
            "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
            "build_id_type": "GNU",
            "file_type": "ELF",
            "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
            "relative_address": "0x000000000002c405",
            "function": "_start"
          }
        ],
        "incomplete": false
      },
      "threads": [
        {
          "crashed": false,
          "name": "ct_worker_0",
          "stack": {
            "format": "Datadog Crashtracker 1.0",
            "frames": [
              {
                "ip": "0x5d1e89d34b57",
                "sp": "0x778c6131ed48",
                "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
                "relative_address": "0x0000000000037b57",
                "column": 9,
                "file": "/home/bits/go/src/github.com/DataDog/libdatadog/bin_tests/src/modes/unix/test_017_multi_thread_collection.rs",
                "function": "worker_fn_0",
                "line": 21
              },
              {
                "ip": "0x5d1e89d34b2c",
                "sp": "0x778c6131ed50",
                "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
                "relative_address": "0x0000000000037b2c",
                "column": 18,
                "file": "/home/bits/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs",
                "function": "__rust_begin_short_backtrace<bin_tests::modes::unix::test_017_multi_thread_collection::{impl#0}::post::{closure_env#0}, ()>",
                "line": 158
              },
              {
                "ip": "0x5d1e89d349e4",
                "sp": "0x778c6131ed70",
                "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
                "relative_address": "0x00000000000379e4",
                "column": 5,
                "file": "/home/bits/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs",
                "function": "call_once<std::thread::{impl#0}::spawn_unchecked_::{closure_env#1}<bin_tests::modes::unix::test_017_multi_thread_collection::{impl#0}::post::{closure_env#0}, ()>, ()>",
                "line": 250
              },
              {
                "ip": "0x5d1e89d8223f",
                "sp": "0x778c6131edf0",
                "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
                "relative_address": "0x000000000008523f",
                "column": 17,
                "file": "/rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/std/src/sys/thread/unix.rs",
                "function": "std::sys::thread::unix::Thread::new::thread_start",
                "line": 126
              },
              {
                "ip": "0x778c613c7ac3",
                "sp": "0x778c6131ee20",
                "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
                "relative_address": "0x0000000000094ac3",
                "column": 8,
                "file": "./nptl/pthread_create.c",
                "function": "start_thread",
                "line": 442
              },
              {
                "ip": "0x778c614598c0",
                "sp": "0x778c6131eec0",
                "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
                "relative_address": "0x00000000001268c0",
                "column": 0,
                "file": "./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S",
                "function": "__clone3",
                "line": 83
              }
            ],
            "incomplete": false
          },
          "state": "R"
        },
        {
          "crashed": false,
          "name": "ct_worker_1",
          "stack": {
            "format": "Datadog Crashtracker 1.0",
            "frames": [
              {
                "ip": "0x5d1e89d34894",
                "sp": "0x778c6111dd48",
                "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
                "relative_address": "0x0000000000037894",
                "column": 9,
                "file": "/home/bits/go/src/github.com/DataDog/libdatadog/bin_tests/src/modes/unix/test_017_multi_thread_collection.rs",
                "function": "worker_fn_1",
                "line": 29
              },
              {
                "ip": "0x5d1e89d34869",
                "sp": "0x778c6111dd50",
                "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
                "relative_address": "0x0000000000037869",
                "column": 18,
                "file": "/home/bits/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs",
                "function": "__rust_begin_short_backtrace<bin_tests::modes::unix::test_017_multi_thread_collection::{impl#0}::post::{closure_env#1}, ()>",
                "line": 158
              },
              {
                "ip": "0x5d1e89d34719",
                "sp": "0x778c6111dd70",
                "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
                "relative_address": "0x0000000000037719",
                "column": 5,
                "file": "/home/bits/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs",
                "function": "call_once<std::thread::{impl#0}::spawn_unchecked_::{closure_env#1}<bin_tests::modes::unix::test_017_multi_thread_collection::{impl#0}::post::{closure_env#1}, ()>, ()>",
                "line": 250
              },
              {
                "ip": "0x5d1e89d8223f",
                "sp": "0x778c6111ddf0",
                "build_id": "8d4cd090dde4a270bed5fb7f8168dcc1291051e0",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
                "relative_address": "0x000000000008523f",
                "column": 17,
                "file": "/rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/std/src/sys/thread/unix.rs",
                "function": "std::sys::thread::unix::Thread::new::thread_start",
                "line": 126
              },
              {
                "ip": "0x778c613c7ac3",
                "sp": "0x778c6111de20",
                "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
                "relative_address": "0x0000000000094ac3",
                "column": 8,
                "file": "./nptl/pthread_create.c",
                "function": "start_thread",
                "line": 442
              },
              {
                "ip": "0x778c614598c0",
                "sp": "0x778c6111dec0",
                "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
                "build_id_type": "GNU",
                "file_type": "ELF",
                "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
                "relative_address": "0x00000000001268c0",
                "column": 0,
                "file": "./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S",
                "function": "__clone3",
                "line": 83
              }
            ],
            "incomplete": false
          },
          "state": "R"
        }
      ]
    }
    ```

### CODE REVIEW

**Comment by gyuheon0h** [comment]
* **#1878** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/1878?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/1878?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

File: libdd-crashtracker/src/receiver/receive_report.rs [hunk 1]
```diff
@@ -290,22 +289,6 @@ fn process_line(
             StdinState::StackTrace
         }
 
-        StdinState::ThreadName(thread_name) if line.starts_with(DD_CRASHTRACK_END_THREAD_NAME) => {
```
gyuheon0h, line 293: This was dead code

gyuheon0h:
@codex r

libdd-crashtracker/src/receiver/receive_report.rs
```diff
@@ -529,6 +509,35 @@ pub(crate) async fn receive_report_from_stream(
         }
     }
 
+    // Thread collection is budgeted against the *remaining* receiver timeout;
+    // whatever time is left after reading the crash data from stdin.
+    // This makes sure the total receiver lifetime is bounded by the configured
+    // timeout, and we always emit whatever threads were collected before
+    // the deadline rather than silently discarding them.
+    #[cfg(target_os = "linux")]
+    if config.collect_all_threads() && builder.error.kind != Some(ErrorKind::UnhandledException) {
```
gyuheon0h: Skipping unhandled exceptions, because I am not convinced of the value of native stacks for all threads, when the main error is the application provided Error/Exception stack, which are application runtime stacks.

bin_tests/src/modes/unix/test_018_thread_limit.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    fn pre(&self, _output_dir: &Path) -> anyhow::Result<()> {
+        Ok(())
+    }
+
+    fn post(&self, _output_dir: &Path) -> anyhow::Result<()> {
+        let barrier = Arc::new(Barrier::new(THREAD_COUNT + 1));
+
+        let _: Vec<_> = (0..THREAD_COUNT)
```
yannham: Nit: since the mapped function is effectful, I don't think rustc can elide the whole thing just because the vec is not used. But if it's just a for loop, IMHO the intent is clearer using an actual `for i in 0..THREAD_COUNT` than an unused `(..).map().collect()`.

bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -188,6 +189,155 @@ fn test_crash_tracking_bin_runtime_callback_frame() {
...
+        let error = &payload["error"];
+        let threads = error["threads"]
+            .as_array()
+            .expect("error.threads should be a JSON array");
+
+        // assert!(
+        //     false,
+        //     "{}",
+        //     serde_json::to_string_pretty(error).unwrap_or_default()
+        // );
```
yannham: Leftover?

bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -188,6 +189,155 @@ fn test_crash_tracking_bin_runtime_callback_frame() {
...
+                    .as_str()
+                    .map(|name| name.contains(worker_fn))
+                    .unwrap_or(false)
+            });
+            assert!(
+                has_worker_frame,
+                "{expected} stack should contain a frame for '{worker_fn}' but got: {frames:?}"
+            );
+        }
+
```
yannham, line 298: Nit: should we also check that there's no additional thread, like `assert!(threads.len() == 2)` (unless I've missed it somewhere)?

libdd-crashtracker/src/collector/crash_handler.rs
```diff
@@ -303,6 +303,18 @@ fn handle_posix_signal_impl(
...
     let receiver = Receiver::from_crashtracker_config(config)?;
 
+    // Enable ptrace permissions for receiver if multi-thread collection is enabled
+    #[cfg(target_os = "linux")]
+    if config.collect_all_threads() {
+        if let Some(receiver_pid) = receiver.handle.pid {
+            // Allow the receiver to ptrace this process for thread context collection
+            // SAFETY: prctl is async-signal-safe and we're just setting ptrace permissions
+            unsafe {
+                libc::prctl(libc::PR_SET_PTRACER, receiver_pid as libc::c_ulong, 0, 0, 0);
```
yannham: Doesn't `PR_SET_PTRACER` take [only one variadic argument](https://man7.org/linux/man-pages/man2/pr_set_ptracer.2const.html) ?
```suggestion
                libc::prctl(libc::PR_SET_PTRACER, receiver_pid as libc::c_ulong);
```

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+    let entries = std::fs::read_dir(&task_dir).map_err(PtraceError::Enumeration)?;
+
+    let mut tids = Vec::new();
+    for entry in entries {
+        let entry = entry.map_err(PtraceError::Enumeration)?;
+        if let Ok(name) = entry.file_name().into_string() {
+            if let Ok(tid) = name.parse::<libc::pid_t>() {
+                tids.push(tid);
+            }
+        }
```
yannham, line 87: Nit: you can use let-chains to get rid of one nesting level 🙂 

```suggestion
        if let Ok(name) = entry.file_name().into_string()
        && let Ok(tid) = name.parse::<libc::pid_t>() {
                tids.push(tid);
        }
```

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+        }));
+    }
+
+    Ok(())
+}
+
+/// Attach to a thread using PTRACE_SEIZE + PTRACE_INTERRUPT, then wait for it
+/// to enter ptrace-stop state before returning.
+fn attach_thread(tid: libc::pid_t) -> Result<(), PtraceError> {
+    // SAFETY: PTRACE_SEIZE attaches without stopping the thread
```
yannham: Nit: this is not a safety comment (safety comment would be about a safety pre-condition of ptrace that we justify is true)

```suggestion
    // PTRACE_SEIZE attaches without stopping the thread
```

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+            ptr::null_mut::<libc::c_void>(),
+            ptr::null_mut::<libc::c_void>(),
+        )
+    };
+    if result == -1 {
+        let errno = unsafe { *libc::__errno_location() };
+        return Err(PtraceError::Attach(tid, errno));
+    }
+
+    // SAFETY: PTRACE_INTERRUPT delivers a stop to the seized thread
```
yannham:
```suggestion
    // PTRACE_INTERRUPT delivers a stop to the seized thread
```

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+    // The thread must already be stopped via ptrace for this to succeed.
+    let upt_info = unsafe { _UPT_create(tid) };
+    if upt_info.is_null() {
+        return StackTrace::new_incomplete();
+    }
+
+    // SAFETY: _UPT_accessors is a static accessor table provided by libunwind-ptrace.
+    // byteorder=0 means native byte order.
+    let addr_space: UnwAddrSpaceT =
+        unsafe { unw_create_addr_space(&raw const _UPT_accessors as *mut _, 0) };
```
yannham: Just to double check here, `unw_create_addr_space` doesn't mutate the pointee, even if we pass a mutable pointer, right? If yes, might be worth mentioning in the safety comment, as it would otherwise be UB to derive a mutable pointer from a static and mutate it, I believe.

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+    if addr_space.is_null() {
+        unsafe { _UPT_destroy(upt_info) };
```
yannham: Not necessarily for this PR (and maybe this actually belongs to libdd-unwind-sys), but I wonder if making small RAII wrappers around `UptInfo` and `UnwAdressSpaceT` that clean upon `drop` would help avoid leak and unclutter the `destroy` dance.

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+    for tid in tids {
+        if start_time.elapsed() >= timeout {
+            break;
+        }
+        if tid == crashing_tid {
+            continue;
+        }
+        if processed >= max_threads {
+            break;
+        }
```
yannham:
```suggestion
        if start_time.elapsed() >= timeout || processed >= max_threads {
            break;
        }
        if tid == crashing_tid {
            continue;
        }
```

libdd-crashtracker/src/receiver/receive_report.rs
```diff
@@ -544,6 +553,88 @@ pub(crate) async fn receive_report_from_stream(
...
+    };
+
+    let name_start = match content.find('(') {
+        Some(i) => i,
+        None => return (None, None),
+    };
+    let name_end = match content.rfind(')') {
+        Some(i) => i,
+        None => return (None, None),
+    };
```
yannham:
```suggestion
    let Some(name_start) = content.find('(') else {
        return (None, None);
    };
    let Some(name_end) = content.rfind(')') else {
        return (None, None);
    };
```

yannham (approved):
I'm less familiar with the libunwind part, but overall I was able to follow thanks to good comments and the syscall/ptrace business looks sound AFAICT 👍 just left minor/pedantic comments.

libdd-crashtracker/src/receiver/ptrace_collector.rs
gyuheon0h: True, that is cleaner, but libdatadog's workspace edition is 2021 and let chains are only in 2024, iirc

yannham: Argh, my bad. Sorry for the noise!

libdd-crashtracker/src/receiver/ptrace_collector.rs
gyuheon0h: Good idea!

libdd-crashtracker/src/receiver/receive_report.rs
gleocadie: It would be confusing:
if runtimes configures `with threads` and do no see the other threads in the crash report. This should not be driven by the crash type

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+            type_name: None,
+            mangled_name: None,
+            comments: vec![],
+        };
+
+        // Resolve the function name if symbol resolution is enabled
+        // We don't care whether it is EnabledWithInprocessSymbols or
+        // EnabledWithSymbolsInReceiver since this is happening in the receiver
+        if resolve_frames != crate::StacktraceCollection::Disabled
+            && resolve_frames != crate::StacktraceCollection::WithoutSymbols
```
gleocadie: This must be aligned with the configuration: if the receiver is configured with Disable or WithoutSymbols, the crashing thread won't have the additional data but other threads yes.

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+        return StackTrace::new_incomplete();
+    }
+
+    // SAFETY: _UPT_accessors is a static accessor table provided by libunwind-ptrace.
+    // unw_create_addr_space only reads the accessor struct; it does not mutate it.
+    // The *mut _ cast is required because the C declaration is const-incorrect, but no
+    // mutation occurs, so casting &raw const to *mut is ok here.
+    // byteorder=0 means native byte order.
+    let addr_space: UnwAddrSpaceT =
+        unsafe { unw_create_addr_space(&raw const _UPT_accessors as *mut _, 0) };
```
gleocadie: The same address space can be used for all threads

libdd-crashtracker/src/receiver/receive_report.rs
```diff
@@ -529,6 +509,35 @@ pub(crate) async fn receive_report_from_stream(
...
+    // the deadline rather than silently discarding them.
+    #[cfg(target_os = "linux")]
+    if config.collect_all_threads() && builder.error.kind != Some(ErrorKind::UnhandledException) {
+        if let Some(proc_info) = builder.proc_info.as_ref() {
+            let parent_pid = proc_info.pid;
+            let crashing_tid = proc_info.tid;
+            // remaining_budget: time left on the receiver's clock after stdin.
+            // If we never received a first line (deadline is None) use zero so
+            // collection is skipped; there is nothing to attach to anyway.
+            let remaining_budget = deadline
```
gleocadie, line 524: we must be careful, that due to scheduling, time accuracy, the receiver is not killed while sending the crash report.

libdd-crashtracker/src/shared/configuration/mod.rs
```diff
@@ -30,17 +30,25 @@ pub enum StacktraceCollection {
...
+    max_threads: usize,
     resolve_frames: StacktraceCollection,
     signals: Vec<i32>,
     timeout: Duration,
     unix_socket_path: Option<String>,
     use_alt_stack: bool,
 }
 
+pub const fn default_max_threads() -> usize {
+    2048
```
gleocadie: A bit big, not sure if it would be realistic. 128-256. might be ok.

libdd-crashtracker/src/receiver/receive_report.rs
gleocadie: It would be great if we could have threads count in the report. That way we could adjust the max threads threshold

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+            if let Ok(tid) = name.parse::<libc::pid_t>() {
+                tids.push(tid);
+            }
+        }
+    }
+    Ok(tids)
+}
+
+/// Wait for a thread to enter ptrace-stop after `PTRACE_INTERRUPT`.
+fn wait_for_stop(tid: libc::pid_t) -> Result<(), PtraceError> {
```
gleocadie: you should make this stop time-bounded (overall and per thread): one stop could consume the remaining time and maybe got the receiver killed.

You can do
```
compute deadline = min(now + 50ms, overall_deadline)
loop:
   ret = waitpid(tid, &mut status, __WALL | WNOHANG)
   if ret == tid and WIFSTOPPED(status) → can_collect = true
   else if ret == 0 → sleep a little and retry until deadline
   else (ret == -1) → error → bail/skip
 
if not stopped by deadline → bail/skip
if stopped → collect_thread (unwind) → detach → next
```

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -0,0 +1,466 @@
...
+    let result = unsafe {
+        libc::ptrace(
+            libc::PTRACE_DETACH,
+            tid as libc::c_long,
+            ptr::null_mut::<libc::c_void>(),
+            ptr::null_mut::<libc::c_void>(),
+        )
+    };
+    if result == -1 {
+        let errno = unsafe { *libc::__errno_location() };
```
gleocadie, line 180: in case of ESRCH, we should not stop. This means that the target thread died.

gleocadie:
> Thread names and states are read from /proc/<pid>/task/<tid>/comm

Can you update the PR description, I did not see where you read the thread name from `/proc/<pid>/task/<tid>/comm`

gyuheon0h:
> Can you update the PR description, I did not see where you read the thread name from `/proc/<pid>/task/<tid>/comm`

@gleocadie That was an earlier iteration. fixed

libdd-crashtracker/src/receiver/ptrace_collector.rs
gyuheon0h: Is the current code not aligned with the configuration? If we set `Disabled` or `WithoutSymbols`, we skip `unw_get_proc_name_remote`. 

> if the receiver is configured with Disable or WithoutSymbols, the crashing thread won't have the additional data but other threads yes

To confirm, we _dont_ want this behavior right? We want symbol or not symbols to be consistent between crashing thread and other threads?

gyuheon0h: Synced offline; only do symbol resolution here for inprocess symbols. if out of process is set, then the symbol resolution will happen in stacktrace enrichment step.

gleocadie (approved):
LGTM
super nice :excited:

PR #1879, opened 2026-04-14
feat(profiling-ffi): ProfilesDictionary_get_func

morrisonlevi:
# What does this PR do?

Adds a `ddog_pprof_ProfilesDictionary_get_func` so you can go from a `FunctionId2` to `Function2`. This capability was already available in the Rust component, just missing from FFI.

# Motivation

I may need this in Ruby to adopt `FunctionId2` efficiently.

# Additional Notes

The tests were written with AI, please scrutinize them.

There was already `ddog_pprof_ProfilesDictionary_get_str`, this mirrors that for functions.

# How to test the change?

Regular tests apply, something like `cargo nextest run -p libdd-profiling -p libdd-profiling-ffi`.

### DIFF
--- a/libdd-profiling/src/profiles/datatypes/function.rs
+++ b/libdd-profiling/src/profiles/datatypes/function.rs
@@ -23,8 +23,12 @@ pub struct Function {
 
 /// An FFI-safe version of the Function which allows null. Be sure to maintain
 /// layout-compatibility with it, except that StringId2 may be null.
+///
+/// Equality and hashing are defined by pointer identity, so they are only
+/// meaningful when comparing ids that originate from the same
+/// `ProfilesDictionary`.
 #[repr(C)]
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash)]
 pub struct Function2 {
     pub name: StringId2,
     pub system_name: StringId2,
--- a/libdd-profiling-ffi/src/profiles/profiles_dictionary.rs
+++ b/libdd-profiling-ffi/src/profiles/profiles_dictionary.rs
@@ -242,15 +242,35 @@ pub unsafe extern "C" fn ddog_prof_ProfilesDictionary_get_str(
     let Some(dict) = dict else {
         return ProfileStatus::from(NULL_PROFILES_DICTIONARY);
     };
-    let string_ref = StringRef::from(string_id);
     // SAFETY: It's not actually safe--as indicated in the docs
     // for this function, the caller needs to be sure the string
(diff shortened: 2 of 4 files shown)


libdd-profiling/src/profiles/datatypes/function.rs
```diff
@@ -23,8 +23,12 @@ pub struct Function {
 
 /// An FFI-safe version of the Function which allows null. Be sure to maintain
 /// layout-compatibility with it, except that StringId2 may be null.
+///
+/// Equality and hashing are defined by pointer identity, so they are only
+/// meaningful when comparing ids that originate from the same
+/// `ProfilesDictionary`.
```
gyuheon0h, line 29: This comment was a bit confusing for me, because be `Eq` and `Hash` is on `Function2`, but "they are only meaningful when comparing ids that originate.." is, I assume, referring the to `StringId2` fields of this struct

gyuheon0h (approved):

libdd-profiling/src/profiles/datatypes/function.rs
morrisonlevi: It means that if you have two ProfilesDictionaries, that a `FunctionId2` from one dictionary cannot meaningfully be compared to a `FunctionId2` from the other. Technically, there is some result, but it is defined to be meaningless. This is so that we can have some flexibility in implementation space. The Eq and Hash traits are quite useful so we define it this way.

I suppose _technically_, this might be a case for using `PartialEq` without `Eq`. I'll consider this for a bit when I'm not tired.

gyuheon0h: Yeah, I understand what the comment means and it is a total nit + me just asking so I can get some clarity for myself. I guess the source of confusion was `..comparing ids..` for me, when `Eq` and `Hash` in this context is for comparing the `Function2`s (which technically bubble down to comparing its id fields 🤷🏽). Would something like this be more clear?

```
/// Equality and hashing are derived field-wise over `name`, `system_name`,
/// and `file_name`. Because those fields are `StringId2`s, comparisons are
/// only generally meaningful when the field `StringId2`s originate from the same
/// `ProfilesDictionary`.
```

Also late for me so I might not be thinking 100% here 😅

morrisonlevi: So PartialEq without Eq isn't quite the right idea. It's like saying `a == a` and `a != a` may behave strangely, like `NaN` in floats. I don't think that's the case here. 

Basically, just don't compare FunctionIds or StringIds that come from different dictionaries and you are okay.

PR #1895, opened 2026-04-20
refactor(libdd-capabilities)!: remove transitive dependency

hoolioh:
# What does this PR do?

PR removes libdd-capabilities as a transitive dependency of libdd-common, cleaning up the layering so libdd-capabilities-impl owns the bridge between traits and infra.

# Motivation

Deduplicate dependencies. The current layout leads to this kind of situations:
```
datadog-sidecar                                                                                                                                                                                              
  ├── libdd-common                                                                                                                                                                                             
  │   └── libdd-capabilities        
  ├── libdd-capabilities-impl
  │   ├── libdd-capabilities (*)                                                                                                                                
  │   └── libdd-common (*)                                                                                                                                                                      
  └── ...                                
```


libdd-capabilities-impl/src/http.rs
```diff
@@ -1,7 +1,62 @@
 // Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
-//! Re-exports `DefaultHttpClient` from `libdd-common`, where it lives alongside
-//! the hyper infrastructure it wraps.
+//! Native HTTP client implementation backed by hyper.
 
-pub use libdd_common::DefaultHttpClient;
+#[cfg(not(target_arch = "wasm32"))]
```
Aaalibaba42: the `libdd-capabilities-impl` crate will never be pulled from wasm32, as `libdatadog-nodejs` owns its own implementation of the capabilities. Gating for `wasm32` is not useful, but it's not a problem either I guess, it just might be misleading to have a gate for something that cannot happen.

Aaalibaba42 (approved):
LGTM, I don't remember why I didn't do that in the first place

libdd-capabilities-impl/src/http.rs
hoolioh: I wasn't sure of that. I'll definitely clean it up

PR #1896, opened 2026-04-20
feat!: add encoder from v04 to v1

anais-raison:
# What does this PR do?

Implements the v0.4 to V1 encoder for the Trace Exporter.

Key additions:
- StringTable for streaming string interning: first occurrence writes the string, subsequent ones write an integer
   ID
- Chunk-level attribute extraction (trace_id 128-bit, sampling_priority, origin) promoted from the root span
- Integer keys for all msgpack fields

# Motivation

[APMSP-2808](https://datadoghq.atlassian.net/browse/APMSP-2808)

# Additional Notes

- V04 and V05 formats are untouched, no existing behavior changes
- TraceExporterOutputFormat::V1 is not exposed in the C FFI yet so there is no production impact
- The string interning is scoped per payload (table reset on each to_vec call)

# How to test the change?

Added unit tests.

[APMSP-2808]: https://datadoghq.atlassian.net/browse/APMSP-2808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libdd-data-pipeline/src/trace_exporter/trace_serializer.rs
```diff
@@ -67,13 +67,27 @@ impl<'a> TraceSerializer<'a> {
...
+                let chunks = trace_utils::collect_trace_chunks(traces, false).map_err(|e| {
+                    TraceExporterError::Deserialization(DecodeError::InvalidFormat(e.to_string()))
+                })?;
+                match chunks {
+                    tracer_payload::TraceChunks::V04(traces) => {
+                        Ok(tracer_payload::TraceChunks::V1(traces))
+                    }
+                    other => Ok(other),
+                }
+            }
```
Aaalibaba42: You call `collect_trace_chunks` saying false on `use_v05_format`, so it will always return `TraceChunks::V04(traces)`, the `other` arm should be unreachable iiuc. I don't know even what's the point of this all or if the `collect_trace_chunks` function should change to take the traces and the enum of formats directly, it seems to do the same logic twice somewhat

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    next_id: u32,
+}
+
+impl StringTable {
+    fn new() -> Self {
+        Self {
+            seen: HashMap::new(),
+            next_id: 1,
+        }
+    }
```
paullegranddc, line 57: Could be 
```suggestion
pub(crate) struct StringTable {
    seen: HashMap<String, u32>,
}

impl StringTable {
    fn new() -> Self {
        let mut seen = HashMap::new();
        seen.insert(String::new(), 0);
        Self {
            seen,
        }
    }
} 
```

Then you can remove the special empty string caee in `write_interned` and do
```rust
let id = self.seen.len();
self.seen.insert(s.to_string(), id);
```

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+/// Promoted fields extracted from spans and written at the chunk level.
+struct ChunkAttrs {
+    /// Full 128-bit trace ID (encodes as 16-byte big-endian binary).
+    trace_id: u128,
+    /// Sampling priority from `_sampling_priority_v1` metric on the root span.
+    sampling_priority: Option<i32>,
+    /// Origin tag from `_dd.origin` meta on the root span.
+    origin: Option<String>,
```
paullegranddc: Is it possible to borrow the origin as an &str instead of copying it?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+        // Chunk-level attributes come from the root span (parent_id == 0).
+        if span.parent_id == 0 {
+            // HashMap::get accepts &Q where K: Borrow<Q>; T::Text: Borrow<str> so &str works.
+            if let Some(v) = span.metrics.get("_sampling_priority_v1") {
+                sampling_priority = Some(*v as i32);
+            }
+            if let Some(v) = span.meta.get("_dd.origin") {
+                origin = Some(v.borrow().to_owned());
+            }
+        }
```
paullegranddc: Can we break out of the loop here if we encountered a root span?
This seems to take the "last root span" in the chunk if there are multiple which i don't think is really deliberate

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+pub(crate) struct StringTable {
+    seen: HashMap<String, u32>,
+    next_id: u32,
+}
+
+impl StringTable {
+    fn new() -> Self {
+        Self {
+            seen: HashMap::new(),
+            next_id: 1,
```
Aaalibaba42: You could probably derive that from the size of the hashmap (I'm not saying it would be more readable, or better, but just to have the option thought of)

libdd-data-pipeline/src/trace_exporter/trace_serializer.rs
Aaalibaba42: Plus in terms of "design", passing `true`/`false` to a function is without context is not great, I had to look up the function in my editor to know what/why etc, I'd be keen to make `use_v05_format` of `collect_trace_chunk` an enum anyway (but maybe that's out of scope for this PR idk).

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,310 @@
...
+                write_uint8(writer, 4u8)?; // Int64
+                write_sint(writer, *i)?;
+            }
+            AttributeArrayValue::Double(d) => {
+                write_uint8(writer, AnyValueKey::Double as u8)?;
+                write_f64(writer, *d)?;
+            }
+        }
+        Ok(())
+    }
```
Aaalibaba42, line 212: I'm not against this per-se, but if it needs to be a function and not a lambda, I'd write it out of the function's body. Like to me either it's an auxiliary function and it's not in the body, or it's a lambda and it can be there. Except if you have good reasons not to do it like that.

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,310 @@
...
+/// Encodes a v0.4 span into the V1 msgpack format.
+///
+/// Key differences from v0.4:
+/// - Uses integer keys for all fields.
+/// - `meta` and `metrics` are combined into a single `attributes` map with type-tagged values.
+/// - `meta_struct` bytes are included in `attributes` as `Bytes` values.
+/// - `trace_id` is not encoded in the span (it belongs to the chunk).
+/// - `error` is encoded as a boolean.
+/// - String values use streaming string interning via `StringTable`.
+#[inline(always)]
```
Aaalibaba42: Why ? It doesn't look like your typical function that benefits from being inline, I'd leave that decision to the compiler except if you have good reasons not to.

Aaalibaba42 (approved):
I'm not learnt in the data-flow etc, I mostly just commented on the code, which do have a few smells but nothing blocking I'd say

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    let mut trace_id = 0u128;
+    let mut sampling_priority = None;
+    let mut origin: Option<String> = None;
+
+    for span in spans {
+        // Any span gives us the trace_id.
+        trace_id = span.trace_id;
+
+        // Chunk-level attributes come from the root span (parent_id == 0).
+        if span.parent_id == 0 {
```
paullegranddc: This is not complete.
Span chunks can have a remote parent, in which case the top level is a span where the parent exists but is not in the chunk. In which case we need to promote it's attributes

You can look for `_dd.top_level` I think, as it is tagged before in the pipeline https://github.com/DataDog/libdatadog/blob/main/libdd-trace-utils/src/span/trace_utils.rs#L16


Btw the agent this is not obvious when looking at the agent v04 to v1 conversion in the agent, but that does happen when the converter iterates over all spans.

It put fields to be promoted in a `SpanConvertedFields` and then promotes these fields to chunk level attributes latter.

(Example here on how this work for the sampling priority:

```
// Collect promoted fields
// https://github.com/DataDog/datadog-agent/blob/paullgdc/trace/add_agent_telemetry_batching/pkg/proto/pbgo/trace/idx/span.go#L1651-L1657

// handlePromotedMetricsFields processes promoted metrics fields that have dedicated span fields
// If we fail to parse a value we don't use the promoted value, but the original string will still be in the span attributes
func (s *InternalSpan) handlePromotedMetricsFields(key uint32, val float64, convertedFields *SpanConvertedFields) {
	if s.Strings.Get(key) == "_sampling_priority_v1" {
		convertedFields.SamplingPriority = int32(val)
	}
}

//  promotes them https://github.com/DataDog/datadog-agent/blob/paullgdc/trace/add_agent_telemetry_batching/pkg/proto/pbgo/trace/idx/span.go#L931

// ApplyPromotedFields applies span promoted fields to the chunk, copying chunk promoted fields to chunkConvertedFields
func (c *InternalTraceChunk) ApplyPromotedFields(convertedFields *SpanConvertedFields, chunkConvertedFields *ChunkConvertedFields) {
    // ...
	c.Priority = int32(convertedFields.SamplingPriority)
    // ...
}
```
)

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+            self.next_id += 1;
+            self.seen.insert(s.to_string(), id);
+            write_str(writer, s)?;
+        }
+        Ok(())
+    }
+}
+
+/// Promoted fields extracted from spans and written at the chunk level.
+struct ChunkAttrs {
```
paullegranddc: I think you are missing the sampling mechanism (_dd.p.dm)

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+///   TraceKey::Chunks (11) → Array[Chunk, ...]
+/// }
+/// ```
+fn encode_payload<W: RmpWrite, T: TraceData, S: AsRef<[Span<T>]>>(
+    writer: &mut W,
+    traces: &[S],
+) -> Result<(), ValueWriteError<W::Error>> {
+    let mut table = StringTable::new();
+
+    // Top-level map contains only the chunks array for now.
```
paullegranddc: What about the payload promoted fields?
* env
* apm mode
* hostname
* app version 
* git commit sha

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
paullegranddc: I kind of like this pattern. This keeps the function scoped, not need to wonder if it's called elsewhere, and having an actual function is not much worse than a closure.
The only thing gained by making this a closure would probably be a shorter declaration thanks to inference

libdd-data-pipeline/src/trace_exporter/trace_serializer.rs
```diff
@@ -67,13 +67,27 @@ impl<'a> TraceSerializer<'a> {
...
+                // For V1, collect as V04 spans and wrap in the V1 variant so the
+                // serializer knows to use the V1 msgpack encoder.
+                let chunks = trace_utils::collect_trace_chunks(traces, false).map_err(|e| {
+                    TraceExporterError::Deserialization(DecodeError::InvalidFormat(e.to_string()))
+                })?;
+                match chunks {
+                    tracer_payload::TraceChunks::V04(traces) => {
+                        Ok(tracer_payload::TraceChunks::V1(traces))
+                    }
+                    other => Ok(other),
```
bantonsson: This shouldn't be reachable, right, so it should error.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    let attrs = extract_chunk_attrs(spans);
+
+    let fields = 2u32 // trace_id + spans are always present
+        + attrs.origin.is_some() as u32
+        + attrs.sampling_priority.is_some() as u32;
+
+    write_map_len(writer, fields)?;
+
+    // 128-bit trace ID as 16-byte big-endian binary.
+    write_uint8(writer, ChunkKey::TraceId as u8)?;
```
bantonsson:
```suggestion
    write_uint(writer, ChunkKey::TraceId)?;
```

The `write` methods will use the smallest possible encoding anyway, and you can remove all the casts.

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
Aaalibaba42: I guess it's just a point of style, I agree it's not a problem in any way, it's just feels icky idk

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
Aaalibaba42: I said that first è_é https://github.com/DataDog/libdatadog/pull/1896#discussion_r3117301566

paullegranddc: when I started my review you hadn't posted yours I think 😄

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
yannham: It's a matter of taste, but I kinda agree with @paullegranddc here. If it's a local re-usable block but doesn't actually need to capture any environment, I personally use the `fn` syntax that I find more uniform (you also don't have to worry about making it `mut` or a reference etc.). For that matter I've seen `struct`, `enum` or even `trait` defined locally. I don't think it's wrong in any way, when justified.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
paullegranddc: The casts are there because ChunkKey is a repr(u8) enum I think no, so using write_uint will nto change anything?

I guess I would prefer something like this pattern though. There is no need for an enum because we never match back 
```
mod chunk_key {
  const PRIORITY: u8  = 1;
  // other keys
  const TRACE_ID: u8 = 6;
}

File: libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs (hunk 7 above)
Comment (yannham) [reply libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs]: I agree, especially given that this function is quite big. FWIW, my "runbook" for inline annotations:
 - if it's a very small function (like, getter or setter, no more than a few operations really), and it's a `pub` item that might be re-used from other crates: `#[inline]` (it often doesn't change much the decision of the compiler but enable cross-crate inlining, which is otherwise not possible by default, and is usually the actual motivation for `#inline` annotations)
 - if it's a small-ish function, that is very performance critical (e.g. called repeatedly in a hot loop) _and I have thoroughly measured with benchmarks that inlining it shows clear, reproducible improvement_: `#[inline(always)]` might be considered. But inlining can be a pessimization, whether it is or not can heavily depends on the callsites as well (and change over time), so it's risky business.

File: libdd-trace-utils/src/msgpack_encoder/v1/mod.rs (hunk 5 above)
Comment (yannham) [reply libdd-trace-utils/src/msgpack_encoder/v1/mod.rs]: [EDIT: Ok I get what you meant, sorry for the noise] Are you talking about `next_id`?

File: libdd-trace-utils/src/msgpack_encoder/v1/mod.rs [hunk 13]
```diff
@@ -0,0 +1,337 @@
...
+/// Streaming string intern table.
+///
+/// The first time a string is written, it is emitted as a msgpack `str` and assigned an
+/// incrementing integer ID. On subsequent occurrences only the ID is emitted as a msgpack `uint`.
+/// ID 0 is reserved for the empty string (written as fixint `0`).
+///
+/// The string table is scoped per payload: each `to_vec` / `write_to_slice` call starts with a
+/// fresh table so deduplication is payload-local.
+pub(crate) struct StringTable {
+    seen: HashMap<String, u32>,
```
yannham, line 49: It's probably for much later, I'm just thinking out loud here. I wonder if we can make the string table slightly more performant with the following scheme, if all the strings have the same lifetime:

- allocate the strings in a local bump arena
- have seen be a `Hashmap<&'self str, u32>` (it's self referential but easy to do with something like `ourboros`)

Something like this:

```rust
#[ouroboros::self_referencing]
struct StringTable {
    /// Preallocates space where strings are stored.
    arena: Arena<u8>,

    /// Deduplicate strings.
    #[borrows(arena)]
    #[covariant]
    map: HashMap<&'this str, u32>,
}
```

The gain would be faster string allocation and de-allocation. This is only if the encoding is perf sensitive.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+        if s.is_empty() {
+            write_uint8(writer, 0)?;
+            return Ok(());
+        }
+        if let Some(&id) = self.seen.get(s) {
+            write_uint(writer, id as u64)?;
+        } else {
+            let id = self.next_id;
+            self.next_id += 1;
+            self.seen.insert(s.to_string(), id);
```
yannham:
```suggestion
            self.seen.insert(s.to_string(), self.next_id);
            self.next_id += 1;
```

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+
+    Ok(())
+}
+
+/// Serializes traces into a slice using the V1 msgpack format.
+///
+/// # Errors
+/// Returns a `ValueWriteError` if the underlying writer fails.
+pub fn write_to_slice<T: TraceData, S: AsRef<[Span<T>]>>(
+    slice: &mut &mut [u8],
```
yannham, line 380: Out of curiosity, why does it need to be `&mut &mut` here?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+/// Serializes traces into a `Vec<u8>` with a pre-allocated capacity.
+pub fn to_vec_with_capacity<T: TraceData, S: AsRef<[Span<T>]>>(
+    traces: &[S],
+    capacity: u32,
+) -> Vec<u8> {
+    let mut buf = ByteBuf::with_capacity(capacity as usize);
+    #[allow(clippy::expect_used)]
+    encode_payload(&mut buf, traces).expect("infallible: the error is std::convert::Infallible");
```
yannham: I think there's a way to `unwrap` an infallible without actually needing a panicking function, although the second match is a bit confusing:

```rust
let unwrapped = match result {
    Ok(x) => x,
    Err(e) => match e {},
}
```

It's not uncommon to have a little helper `unwrap_infallible` in some crates to do that for you.

yannham: Additionally, since you don't do anything with the result, you could also do something like:
`let _ = encode_payalod` with a comment explaining why you ignore the result. Might slightly nicer than the allow(clippy) and the `expect` and error message that will never happen.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,337 @@
...
+    buf.into_vec()
+}
+
+/// Returns the number of bytes the V1 payload for `traces` would occupy.
+pub fn to_len<T: TraceData, S: AsRef<[Span<T>]>>(traces: &[S]) -> u32 {
```
yannham: Nitpick: I find the name of this function a bit obscure.

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,310 @@
...
+    writer: &mut W,
+    span_links: &[SpanLink<T>],
+    table: &mut StringTable,
+) -> Result<(), ValueWriteError<W::Error>> {
+    write_uint8(writer, SpanKey::SpanLinks as u8)?;
+    rmp::encode::write_array_len(writer, span_links.len() as u32)?;
+
+    for link in span_links.iter() {
+        let trace_id_128 = ((link.trace_id_high as u128) << 64) | link.trace_id as u128;
+        let link_len = 1 /* trace_id (always) */
```
yannham:
```suggestion
        let link_len = 1 // trace_id (always)
```

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,310 @@
...
+/// Uses integer keys and string interning. Attribute values are type-tagged.
+pub fn encode_span_events<W: RmpWrite, T: TraceData>(
+    writer: &mut W,
+    span_events: &[SpanEvent<T>],
+    table: &mut StringTable,
+) -> Result<(), ValueWriteError<W::Error>> {
+    write_uint8(writer, SpanKey::SpanEvents as u8)?;
+    rmp::encode::write_array_len(writer, span_events.len() as u32)?;
+
+    for event in span_events.iter() {
```
yannham: I think `&[_]` implements `IntoIter<Item = &_>`?

```suggestion
    for event in span_events {
```

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,310 @@
...
+        let event_len = 2 /* time_unix_nano, name */
```
yannham:
```suggestion
        let event_len = 2 // time_unix_nano, name
```

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
ekump: I agree it doesn't have to be in this PR, but I think this is worth exploring deeper. String allocation is traditionally the biggest drag on performance. @anais-raison Can you create a ticket for the epic to investigate later?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
ekump: @ajgajg1134 - Weren't we discussing something similar in the agent recently? A scenario where we encounter multiple root spans?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
ekump: +1 we should try to avoid panics unless absolutely necessary (like acquiring mutex locks, for example)

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
ekump: Just to pile on, I agree with @paullegranddc and @yannham. The scoping here makes the intentions clear.

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
ekump: Agreed. I think our msgpack encoder / decoders had a lot of inline annotations in the past (and probably some still hanging around) and most of them made no difference.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -84,6 +85,7 @@ impl TraceExporterOutputFormat {
             match self {
                 TraceExporterOutputFormat::V04 => "/v0.4/traces",
                 TraceExporterOutputFormat::V05 => "/v0.5/traces",
+                TraceExporterOutputFormat::V1 => "/v1.0/traces",
```
ekump, line 97: I believe the builder is going to fail validation if input is `v0.4` and output is `v1.0`. Is that intentional? With this PR we should have no issue serializing to `v1.0`, right?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -74,6 +74,7 @@ pub enum TraceExporterOutputFormat {
     #[default]
     V04,
     V05,
+    V1,
```
ekump, line 86: Can we include an integration test that does a snapshot test with the test agent? I assume you'll have to bump the version of the test agent to pick up the V1 support.

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
anais-raison: I had mostly kept it because it was used as well in the `v4/span.rs` parallel function but I'm fine with removing it. Thanks for the explanations !

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: I put this name to match what was done in for the v04 but I took the opportunity to fix it there as well !

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: Yep done !

libdd-data-pipeline/src/trace_exporter/trace_serializer.rs
anais-raison: I'm gonna do a follow up PR to change the design with `use_v05_format` because I do agree that it's not the best. But we do in fact return v1 here.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: Without the outer &mut, the slice advances only locally and the caller can't tell how many bytes were written.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
ajgajg1134: huh weird, I missed this notification originally but am seeing it now. So 2 things: 1, we should be "Safe" to drop out iff we find all these fields on that root. There's a separate issue here which is that we should not _only_ rely on the root for these fields, e.g. if we can't find a root (partial flushing is one potential reason) then we should just accept any value we find for these tags.

Note that we should still prefer the root span values for these over non-root spans for backwards compatibility reasons.

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,309 @@
...
+    if !span.r#type.borrow().is_empty() {
+        write_uint8(writer, SpanKey::Type as u8)?;
+        table.write_interned(writer, span.r#type.borrow())?;
+    }
+
+    if has_attributes {
+        let attr_count =
+            span.meta.len() as u32 + span.metrics.len() as u32 + span.meta_struct.len() as u32;
+        write_uint8(writer, SpanKey::Attributes as u8)?;
+        rmp::encode::write_map_len(writer, attr_count)?;
```
ajgajg1134: I think this code doesn't match the spec for serializing attributes? I believe this should be serialized as an array where each "meta" entry ends up as 3 distinct values in the array (so the length is changed too). Double check the "Serializing to v1" example in the tracer implementation guide document

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,482 @@
...
+            if let Some(v) = span.metrics.get("_sampling_priority_v1") {
+                sampling_priority = Some(*v as i32);
+            }
+            if let Some(v) = span.meta.get("_dd.origin") {
+                origin = Some(v.borrow());
+            }
+            // _dd.p.dm is a signed integer sampling mechanism code stored as a string.
+            if let Some(v) = span.meta.get("_dd.p.dm") {
+                if let Ok(dm) = v.borrow().parse::<i32>() {
+                    sampling_mechanism = Some(dm as u32);
```
ajgajg1134: I think this conversion is wrong? e.g. a value of `-4` unwrapped to `u32` will be a big positive number and not `4`. I think this should be an easy fix though and you can just add `unsigned_abs` https://doc.rust-lang.org/std/primitive.i64.html#method.unsigned_abs

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,309 @@
...
+    write_bin, write_bool, write_f64, write_i64, write_sint, write_u64, write_uint, write_uint8,
+    RmpWrite, ValueWriteError,
+};
+use std::borrow::Borrow;
+
+use super::StringTable;
+
+/// Integer keys for V1 span fields.
+#[repr(u8)]
+pub(super) enum SpanKey {
```
ajgajg1134, line 17: it looks like span Kind is missing from this list (and needs to be looked up from the span's existing meta )

ajgajg1134: Same for `component`

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,482 @@
...
+use crate::span::TraceData;
+use rmp::encode::{
+    write_array_len, write_bin, write_map_len, write_sint, write_str, write_uint, write_uint8,
+    ByteBuf, RmpWrite, ValueWriteError,
+};
+use std::borrow::Borrow;
+use std::collections::HashMap;
+
+/// Integer keys for the top-level V1 trace payload map.
+mod trace_key {
```
ajgajg1134, line 17: looks like there's a few other fields missing here: language_name, language_version, tracer_version, and runtimeId

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        if !link.attributes.is_empty() {
+            write_uint8(writer, SpanLinkKey::Attributes as u8)?;
+            rmp::encode::write_map_len(writer, link.attributes.len() as u32)?;
+            for (k, v) in link.attributes.iter() {
+                table.write_interned(writer, k.borrow())?;
+                write_uint8(writer, AnyValueKey::String as u8)?;
+                table.write_interned(writer, v.borrow())?;
+            }
+        }
```
VianneyRuhlmann: nit: Any reason to not put them in the same order as the keys ?

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,309 @@
...
+                write_bool(writer, *b).map_err(ValueWriteError::InvalidDataWrite)?;
+            }
+            AttributeArrayValue::Integer(i) => {
+                write_uint8(writer, 4u8)?; // Int64
+                write_sint(writer, *i)?;
+            }
+            AttributeArrayValue::Double(d) => {
+                write_uint8(writer, AnyValueKey::Double as u8)?;
+                write_f64(writer, *d)?;
+            }
```
VianneyRuhlmann:
```suggestion
            AttributeArrayValue::Double(d) => {
                write_uint8(writer, AnyValueKey::Double as u8)?;
                write_f64(writer, *d)?;
            }
            AttributeArrayValue::Integer(i) => {
                write_uint8(writer, 4u8)?; // Int64
                write_sint(writer, *i)?;
            }
```
nit: to follow key indices

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,309 @@
...
+            AttributeArrayValue::String(s) => {
+                write_uint8(writer, AnyValueKey::String as u8)?;
+                table.write_interned(writer, s.borrow())?;
+            }
+            AttributeArrayValue::Boolean(b) => {
+                write_uint8(writer, AnyValueKey::Bool as u8)?;
+                write_bool(writer, *b).map_err(ValueWriteError::InvalidDataWrite)?;
+            }
+            AttributeArrayValue::Integer(i) => {
+                write_uint8(writer, 4u8)?; // Int64
```
VianneyRuhlmann: Why isn't this in the AnyValueKey enum ?

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,309 @@
...
+        }
+        Ok(())
+    }
+
+    match attribute {
+        AttributeAnyValue::SingleValue(value) => {
+            encode_array_element(writer, value, table)?;
+        }
+        AttributeAnyValue::Array(array) => {
+            write_uint8(writer, 6u8)?; // Array
```
VianneyRuhlmann: Why isn't this in the enum ?

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,309 @@
...
+pub(super) enum SpanEventKey {
+    Time = 1,
+    Name = 2,
+    Attributes = 3,
+}
+
+/// Type discriminants for attribute values.
+/// An attribute value is encoded as [type_uint8][actual_value].
+#[repr(u8)]
+pub(super) enum AnyValueKey {
```
VianneyRuhlmann, line 57: Is it ok to not support `KeyValueList` ? I guess it's fine for now if we don't use it when converting from v04, but calling it out since we will need it to fully support v01 spans.

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -165,20 +144,20 @@ impl std::io::Write for CountLength {
...
 /// let span = SpanSlice {
 ///     name: "test-span",
 ///     ..Default::default()
 /// };
 /// let traces = vec![vec![span]];
-/// let encoded_len = to_len(&traces);
+/// let encoded_len = to_encoded_byte_len(&traces);
 ///
 /// assert!(encoded_len > 0);
 /// ```
```
VianneyRuhlmann, line 158: Part of this doc is also relevant for `CountLength`. Now that they're separated you probably should add a doc to `CountLength`

libdd-data-pipeline/src/trace_exporter/trace_serializer.rs
```diff
@@ -67,13 +67,27 @@ impl<'a> TraceSerializer<'a> {
...
+                // serializer knows to use the V1 msgpack encoder.
+                let chunks = trace_utils::collect_trace_chunks(traces, false).map_err(|e| {
+                    TraceExporterError::Deserialization(DecodeError::InvalidFormat(e.to_string()))
+                })?;
+                match chunks {
+                    tracer_payload::TraceChunks::V04(traces) => {
+                        Ok(tracer_payload::TraceChunks::V1(traces))
+                    }
+                    _ => unreachable!(),
+                }
```
VianneyRuhlmann:
```suggestion                        Ok(tracer_payload::TraceChunks::V1(traces))
```

libdd-trace-utils/src/tracer_header_tags.rs
```diff
@@ -27,6 +27,7 @@ pub struct TracerHeaderTags<'a> {
     pub lang_vendor: &'a str,
     pub tracer_version: &'a str,
     pub container_id: &'a str,
+    pub runtime_id: &'a str,
```
VianneyRuhlmann: I don't think you should add `runtime_id` to `TraceHeaderTags` since the runtime id is not sent as a header the serializer should take the TracerMetadata struct which is the more generic one and includes all metadata you need.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -70,11 +76,27 @@ impl StringTable {
...
+///
+/// These fields are required by the V1 spec: tracers must include them in the payload
+/// (not only in HTTP headers).
+#[derive(Default)]
+pub struct PayloadMetadata<'a> {
+    pub language_name: &'a str,
+    pub language_version: &'a str,
+    pub tracer_version: &'a str,
+    pub runtime_id: &'a str,
+}
```
VianneyRuhlmann: Do you need an intermediate struct here ? Can't you directly pass the `TracerHeaderTags` or `TracerMetadata` struct to the serializer ?

libdd-trace-utils/src/tracer_header_tags.rs
VianneyRuhlmann: I think the runtime_id shouldn't be added here since it's not passed as a header. You should use the TracerMetadata struct instead as it contains all necessary fields.

VianneyRuhlmann (approved):
LGTM

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,679 @@
...
+///   trace_key::CHUNKS       (11) → Array[Chunk, ...]
+/// }
+/// ```
+fn encode_payload<W: RmpWrite, T: TraceData, S: AsRef<[Span<T>]>>(
+    writer: &mut W,
+    traces: &[S],
+    metadata: &TracerMetadata,
+) -> Result<(), ValueWriteError<W::Error>> {
+    let mut table = StringTable::new();
+    let payload_attrs = extract_payload_attrs(traces);
```
ekump: `extract_payload_attrs()` only walks span meta to populate `env`, `hostname`, `app_version`, `git_commit_sha`. But aren't these fields set on the builder once, not in every span and stored in `TracerMetadata`?

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,372 @@
...
+
+/// Maps the `span.kind` string tag (from v0.4 meta) to the OTEL SpanKind uint32.
+fn span_kind_from_str(s: &str) -> Option<u32> {
+    match s {
+        "internal" => Some(1),
+        "server" => Some(2),
+        "client" => Some(3),
+        "producer" => Some(4),
+        "consumer" => Some(5),
+        _ => None,
```
ekump: Is omitting span kinds that don't match intentional? Should there be a default, like [in the agent](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/converter.go#L156)?

libdd-data-pipeline/tests/snapshots/compare_exporter_v04_to_v1_trace_snapshot_test.json
```diff
@@ -0,0 +1,51 @@
+[[
+  {
+    "name": "test_exporter_v04_v1_snapshot_root",
+    "service": "test-service",
+    "resource": "test-resource",
+    "trace_id": 0,
```
ekump, line 6: I know this is how the test agent normally works. It ignores `trace_id` and I think it sets the id based on the index of where the trace is in the payload it receives. 

But for V1, should we be testing the trace_id matches an expected value since we're converting to 128-bit and we want to cover that logic? 

If that requires changes to the test agent then it can be handled in a follow up PR.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,679 @@
...
+            span.parent_id == 0 || span.metrics.get("_dd.top_level").copied().unwrap_or(0.0) == 1.0;
+
+        if is_root {
+            // Root span is authoritative: its values supersede any non-root fallback,
+            // including absence (a field missing on the root should not be filled from non-roots).
+            sampling_priority = span.metrics.get("_sampling_priority_v1").map(|v| *v as i32);
+            origin = span.meta.get("_dd.origin").map(|v| v.borrow());
+            // _dd.p.dm is a signed integer stored as a string; unsigned_abs preserves the
+            // magnitude.
+            sampling_mechanism = span
```
ekump, line 192: Are we testing `_dd.p.dm` is being correctly set at the chunk level anywhere?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,679 @@
...
+                .meta
+                .get("_dd.p.dm")
+                .and_then(|v| v.borrow().parse::<i32>().ok())
+                .map(|dm| dm.unsigned_abs());
+            break;
+        }
+
+        // No root found yet — accumulate fallback values from non-root spans (partial flush).
```
ekump, line 200: I'm not sure we're testing the fallback logic anywhere?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -0,0 +1,679 @@
...
+where
+    T::Text: 'a,
+{
+    let mut trace_id = 0u128;
+    let mut sampling_priority = None;
+    let mut origin = None;
+    let mut sampling_mechanism = None;
+
+    for span in spans {
+        trace_id = span.trace_id;
```
ekump: Could setting `trace_id` be done outside of this for loop? Isn't it the same for every span in a chunk?

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,372 @@
...
+    if !span.resource.borrow().is_empty() {
+        write_uint8(writer, SpanKey::Resource as u8)?;
+        table.write_interned(writer, span.resource.borrow())?;
+    }
+
+    write_uint8(writer, SpanKey::SpanId as u8)?;
+    write_u64(writer, span.span_id)?;
+
+    write_uint8(writer, SpanKey::Start as u8)?;
+    write_u64(writer, span.start as u64)?;
```
ekump: non-blocking: for this and `span.duration` below - are you at all concerned about negative values? I believe as it is now, if we get a negative value from a tracer (not that we ever should) then it'll silently wrap around and produce a garbage value. If I'm reading the agent code correctly, it does the same thing.

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    if span.duration != 0 {
+        write_uint8(writer, SpanKey::Duration as u8)?;
+        write_u64(writer, span.duration as u64)?;
+    }
+
+    if span.error != 0 {
+        write_uint8(writer, SpanKey::Error as u8)?;
+        write_bool(writer, span.error != 0).map_err(ValueWriteError::InvalidDataWrite)?;
```
ekump: non-blocking and a nit: I think the compiler will correctly optimize this, but you could just hardcode true here for `span.error != 0` since we check above.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
ajgajg1134: I see this is marked as resolved but it doesn't appear these fields are here?

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
anais-raison: Yes you're right, done !

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: True, it's updated.

libdd-data-pipeline/tests/snapshots/compare_exporter_v04_to_v1_trace_snapshot_test.json
anais-raison: I added the test, but a follow up PR is on progress for the test agent cause of some change needed there to fully test the V1.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: Added a test for that as well

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: Added a test as well

libdd-trace-utils/src/msgpack_encoder/v1/span_v04.rs
anais-raison: Just added a fallback strategy to 0 for duration and now for start time to be sure to not send wrong data to the agent.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: Done !

ekump (approved):

Aaalibaba42:
I've gotten @anais-raison 's approval for me doing this

PR #1897, opened 2026-04-20
fix(libdd-telemetry-ffi): logic

Aaalibaba42:
# Bug

The FFI functions ddog_telemetry_handle_add_dependency, ddog_telemetry_handle_add_integration, and ddog_telemetry_handle_add_log had an inverted condition when converting optional string fields (version, stack_trace) from C slices to Rust Option<String>.

The code was using .is_empty().then(|| ...), which would populate the Option only when the input was empty — the exact opposite of the intended behavior. Non-empty values were silently dropped, and empty strings were incorrectly converted and passed through.

# Fix

Changed the condition to (!slice.is_empty()).then(|| ...) so that the Option is Some(...) when the input is non-empty and None when it is empty, as intended.

# Additional Notes

Addresses APMSP-2937

# How to test the change?

96844c851a894dd7528dd5ae30bf914c5e16ceb2 show tests with failing stuff, 14b6d1e2b79522274525625515dbc2a04bd27091 show the fix and the tests passing

### DIFF
--- a/libdd-telemetry-ffi/src/lib.rs
+++ b/libdd-telemetry-ffi/src/lib.rs
@@ -119,6 +119,35 @@ mod tests {
     };
     use std::{mem::MaybeUninit, ptr::NonNull};
 
+    /// Spins up a worker backed by a file:// endpoint, returns (handle, temp_file).
+    /// The caller is responsible for stopping the worker and reading the file.
+    unsafe fn start_file_backed_worker() -> (Box<TelemetryWorkerHandle>, tempfile::NamedTempFile) {
+        let mut builder: MaybeUninit<Box<TelemetryWorkerBuilder>> = MaybeUninit::uninit();
+        ddog_telemetry_builder_instantiate(
+            NonNull::new(&mut builder).unwrap().cast(),
+            ffi::CharSlice::from("test-service"),
+            ffi::CharSlice::from("rust"),
+            ffi::CharSlice::from("1.0"),
+            ffi::CharSlice::from("0.0.1"),
+        )
+        .unwrap_none();
+        let mut builder = builder.assume_init();
+
+        let f = tempfile::NamedTempFile::new().unwrap();
+        ddog_telemetry_builder_with_endpoint_config_endpoint(
+            &mut builder,
+            &Endpoint::from_slice(&format!("file://{}", f.path().to_str().unwrap())),
+        )
+        .unwrap_none();
+
+        let mut handle: MaybeUninit<Box<TelemetryWorkerHandle>> = MaybeUninit::uninit();
+        ddog_telemetry_builder_run(builder, NonNull::new(&mut handle).unwrap().cast())
+            .unwrap_none();
+        let handle = handle.assume_init();
+        ddog_telemetry_handle_start(&handle).unwrap_none();
+        (handle, f)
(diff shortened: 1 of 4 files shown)


libdd-telemetry-ffi/src/worker_handle.rs
```diff
@@ -20,9 +20,8 @@ pub unsafe extern "C" fn ddog_telemetry_handle_add_dependency(
     dependency_version: ffi::CharSlice,
 ) -> MaybeError {
     let name = dependency_name.to_utf8_lossy().into_owned();
-    let version = dependency_version
-        .is_empty()
-        .then(|| dependency_version.to_utf8_lossy().into_owned());
+    let version =
```
ekump: There is already something similar [here](https://github.com/DataDog/libdatadog/blob/650c804b170f9bb47ace9a0e8e672851f818b5d7/libdd-common-ffi/src/slice.rs#L117). You could just add another function to support lossy (I'd question why we're ok with lossy here, but that's perhaps beyond the scope of the PR). If the logic lives in libdd-common-ffi, it should theoretically make testing a bit easier too.

libdd-telemetry-ffi/src/worker_handle.rs
```diff
@@ -73,9 +71,7 @@ pub unsafe extern "C" fn ddog_telemetry_handle_add_log(
         id,
         message.to_utf8_lossy().into_owned(),
         level,
-        stack_trace
-            .is_empty()
-            .then(|| stack_trace.to_utf8_lossy().into_owned()),
+        (!stack_trace.is_empty()).then(|| stack_trace.to_utf8_lossy().into_owned()),
```
ekump: I don't think this change is covered by your existing tests.

libdd-telemetry-ffi/src/worker_handle.rs
Aaalibaba42: Looking a bit into it, I agree this probably shouldn't use the lossy version. Other crates in libdatadog don't accept lossy except in the demangler where mangled symbols can genuinely be non-UTF-8. Here, it would mean accepting a version number or something like this which contain non utf-8 compliant bytes, and ending up accepting things with `1.2.�` at the end of the chain, which would be harder to debug etc.

ekump (approved):

libdd-telemetry-ffi/src/lib.rs
```diff
@@ -372,4 +404,218 @@ mod tests {
...
+        if value["request_type"].as_str() == Some(target_type) {
+            results.push(value.clone());
+        }
+        if value["request_type"].as_str() == Some("message-batch") {
+            if let Some(batch) = value["payload"].as_array() {
+                for item in batch {
+                    collect_payloads_recursive(item, target_type, results);
+                }
+            }
+        }
```
iunanua: call `value["request_type"].as_str()` just once?

```suggestion
        let request_type = value["request_type"].as_str();
        if request_type == Some(target_type) {
            results.push(value.clone());
        }
        if request_type == Some("message-batch") {
            if let Some(batch) = value["payload"].as_array() {
                for item in batch {
                    collect_payloads_recursive(item, target_type, results);
                }
            }
        }
```

PR #1909, opened 2026-04-22
chore(skills): add create-release skill

VianneyRuhlmann:
# What does this PR do?

Add a claude skill to create a new release

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/.claude/skills/create-release/SKILL.md
+++ b/.claude/skills/create-release/SKILL.md
@@ -1,0 +1,72 @@
+---
+name: create-release
+description: Bump the Rust workspace version in root Cargo.toml, regenerate the lockfile, and open a draft PR on GitHub. Use this skill whenever the user says something like "create a release", "bump the version", "release vX.Y.Z", "prepare a release branch", or "bump workspace version". Trigger even if they just say "release X.Y.Z" or mention a semver version in a release context.
+---
+
+# Create Release
+
+Automate the mechanical parts of cutting a release: branch, version bump, lockfile update, draft PR.
+
+## Steps
+
+### 1. Get the target version
+
+If the user didn't supply a version already, ask: "What version should I bump to?"
+
+Validate it looks like a semver string (e.g. `32.0.0`, `31.1.0`). Do not add a `v` prefix in Cargo.toml — the file uses bare semver (e.g. `31.0.0`).
+
+### 2. Check working directory is clean
+
+```bash
+git status --porcelain
+```
+
+If there are uncommitted changes, stop and tell the user to commit or stash them first.
+
+### 3. Create and switch to release branch
+
+Branch name: `release/v<version>` (e.g. `release/v32.0.0`). Always branch from `main`, regardless of the current branch.
+
+```bash
+git fetch origin main
+git checkout -b release/v<version> origin/main
+```
+
+### 4. Bump the workspace version
+
+Edit the root `Cargo.toml`. Find the `[workspace.package]` section and update the `version` field to the new value. Only change this one field — do not touch individual crate `Cargo.toml` files, `CHANGELOG.md`, or anything else.
+
+Use the Edit tool to make a targeted replacement — do not rewrite the whole file.
+
+### 5. Regenerate the lockfile
+
+```bash
+cargo build
+```
+
+This may take a moment. Wait for it to finish.
+
+### 6. Commit
+
+```bash
+git add Cargo.toml Cargo.lock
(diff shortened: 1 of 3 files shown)


.claude/skills/create-release/SKILL.md
```diff
@@ -0,0 +1,72 @@
...
+
+If there are uncommitted changes, stop and tell the user to commit or stash them first.
+
+### 3. Create and switch to release branch
+
+Branch name: `release/v<version>` (e.g. `release/v32.0.0`). Always branch from `main`, regardless of the current branch.
+
+```bash
+git fetch origin main
+git checkout -b release/v<version> origin/main
```
Aaalibaba42: I think `git switch -c` is prefered

Aaalibaba42 (commented):
I'm generally uncomfortable with letting AI use git. Why can't this be a script instead of a skill ? Is there some non-deterministic / hard to anticipate aspects ? This looks pretty straight forward

VianneyRuhlmann:
It can be a script, I think the skill helps with discoverability since you don't have to know it exist for claude to use it. I can make it a script and have a skill to make the script discoverable.

.claude/skills/create-release/SKILL.md
VianneyRuhlmann: I never did the switch to switch 😄

.github/CODEOWNERS
```diff
@@ -2,6 +2,7 @@
 # Rules are evaluated in order. Later matches override earlier ones.
 # More specific patterns at the bottom override general patterns above.
 
+.claude/                             @DataDog/libdatadog-core
```
ekump: should this be `apm-common-components-core`?

scripts/create-release.sh
```diff
@@ -0,0 +1,72 @@
...
+
+VERSION="$1"
+
+if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$ ]]; then
+  echo "Error: '$VERSION' is not a valid bare semver (e.g. 32.0.0). Do not include a 'v' prefix." >&2
+  exit 2
+fi
+
+REPO_ROOT="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$REPO_ROOT"
```
Aaalibaba42, line 31: mmh since we have access to git, don't we want to use `git rev-parse --show-toplevel` ?

scripts/create-release.sh
```diff
@@ -0,0 +1,72 @@
...
+
+echo "==> Committing"
+git add Cargo.toml Cargo.lock
+git commit -m "chore: bump workspace version to ${VERSION}"
+
+echo "==> Pushing ${BRANCH}"
+git push -u origin "${BRANCH}"
+
+echo "==> Creating draft PR"
+gh pr create \
```
Aaalibaba42, line 62: We might want to start the script with a `if [ ! -x "$(which gh)" ]; then echo "gh not installed or couldn't be found"; exit 1; fi` or something like this

scripts/create-release.sh
```diff
@@ -0,0 +1,72 @@
...
+
+pattern = re.compile(
+    r'(\[workspace\.package\][^\[]*?\nversion\s*=\s*")[^"]*(")',
+    re.DOTALL,
+)
+new_text, n = pattern.subn(rf'\g<1>{version}\g<2>', text, count=1)
+if n != 1:
+    sys.exit("Error: could not locate version field in [workspace.package]")
+path.write_text(new_text)
+PY
```
Aaalibaba42: That looks like a sed thing

ekump: Yes, but I believe there are minor differences with sed on mac and linux, so you'd just have to account for that. This might do what you need? 

` sed -i.bak '/\[workspace\.package\]/,/^\[/{s/^version = ".*"/version = "'"${VERSION}"'"/;}' Cargo.toml`

scripts/create-release.sh
ekump: I think `gh` is typically available on dd laptops, but good call. Better to fail fast.

ekump (approved):
@Aaalibaba42's suggestions are good. LGTM after you apply them. I think you also have to assign the script a codeowner and add license headers.

Aaalibaba42 (approved):
lgtm

PR #1915, opened 2026-04-23
feat: otel thread ctx FFI

yannham:
# What does this PR do?

This PR adds a basic FFI for the OTel thread-level context feature: create a new context, attach, detach, and update in place.

We also make `ThreadContextRecord` public, or at least exposed in the FFI. The rationale is that:

1. it's imposed by the spec, so it should not be a liability regarding breaking changes: we can't really touch it anyway.
2. as mentioned in the doc of the FFI, there's a potential for SDK updating themselves the contexts without going through libdatadog at all after publication. In this usage mode, the export of the C struct `ThreadContextRecord` is a way to document its expected memory layout.

<details>
<summary>Generated C header</summary>

```c


#ifndef DDOG_OTEL_THREAD_CTX_H
#define DDOG_OTEL_THREAD_CTX_H

#pragma once

#include <stdbool.h>
#include <stddef.h>
#include <stdint.h>

/**
 * In-memory layout of a thread-level context.
 *
 * **CAUTION**: The structure MUST match exactly the OTel thread-level context specification.
 * It is read by external, out-of-process code. Do not re-order fields or modify in any way,
 * unless you know exactly what you're doing.
 *
 * # Synchronization
 *
 * Readers are async-signal handlers. The writer is always stopped while a reader runs.
 * Sharing memory with a signal handler still requires some form of synchronization, which is
 * achieved through atomics and compiler fence, using `valid` and/or the TLS slot as
 * synchronization points.
 *
 * - The writer stores `valid = 0` *before* modifying fields in-place, guarded by a fence.
 * - The writer stores `valid = 1` *after* all fields are populated, guarded by a fence.
 * - `valid` starts at `1` on construction and is never set to `0` except during an in-place
 *   update.
 */
typedef struct ddog_ThreadContextRecord {
  /**
   * Trace identifier; all-zeroes means "no trace".
   */
  uint8_t trace_id[16];
  /**
   * Span identifier.
   */
  uint8_t span_id[8];
  /**
   * Whether the record is ready/consistent. Always set to `1` except during in-place update
   * of the current record.
   */
  uint8_t valid;
  uint8_t _reserved;
  /**
   * Number of populated bytes in `attrs_data`.
   */
  uint16_t attrs_data_size;
  /**
   * Packed variable-length key-value records.
   *
   * It's a contiguous list of blocks with layout:
   *
   * 1. 1-byte `key_index`
   * 2. 1-byte `val_len`
   * 3. `val_len` bytes of a string value.
   *
   * # Size
   *
   * Currently, we always allocate the max recommended size. This potentially wastes a few
   * hundred bytes per thread, but it guarantees that we can modify the context in-place
   * without (re)allocation in the hot path. Having a hybrid scheme (starting smaller and
   * resizing up a few times) is not out of the question.
   */
  uint8_t attrs_data[ddog_MAX_ATTRS_DATA_SIZE];
} ddog_ThreadContextRecord;

#ifdef __cplusplus
extern "C" {
#endif // __cplusplus

/**
 * Allocate and initialise a new thread context.
 *
 * Returns a non-null owned handle that must eventually be released with
 * `ddog_otel_thread_ctx_free`.
 */
struct ddog_ThreadContextRecord *ddog_otel_thread_ctx_new(const uint8_t (*trace_id)[16],
                                                          const uint8_t (*span_id)[8],
                                                          const uint8_t (*local_root_span_id)[8]);

/**
 * Free an owned thread context.
 *
 * # Safety
 *
 * `ctx` must be a valid non-null pointer obtained from `ddog_otel_thread_ctx_new` or
 * `ddog_otel_thread_ctx_detach`, and must not be used after this call. In particular, `ctx`
 * must not be currently attached to a thread.
 */
void ddog_otel_thread_ctx_free(struct ddog_ThreadContextRecord *ctx);

/**
 * Attach `ctx` to the current thread. Returns the previously attached context if any, or null
 * otherwise.
 *
 * # Safety
 *
 * `ctx` must be a valid non-null pointer obtained from this API. Ownership of `ctx` is
 * transferred to the TLS slot: the caller must not drop `ctx` while it is still actively
 * attached.
 *
 * ## In-place update
 *
 * The preferred method to update the thread context in place is [ddog_otel_thread_ctx_update].
 *
 * If calling into native code is too costly, it is possible to update an attached context
 * directly in-memory without going through libdatadog (contexts are guaranteed to have a
 * stable address through their lifetime). **HOWEVER, IF DOING SO, PLEASE BE VERY CAUTIOUS OF
 * THE FOLLOWING POINTS**:
 *
 * 1. The update process requires a [seqlock](https://en.wikipedia.org/wiki/Seqlock)-like
 *    pattern: [ThreadContextRecord::valid] must be first set to `0` before the update and set
 *    to `1` again at the end. Additionally, depending on your language's memory model, you
 *    might need specific synchronization primitives (compiler fences, atomics, etc.), since
 *    the context can be read by an asynchronous signal handler at any point in time. See the
 *    [Otel thread context
 *    specification](https://github.com/open-telemetry/opentelemetry-specification/pull/4947)
 *    for more details.
 * 2. Only update the context from the thread it's attached to. Contexts are designed to be
 *    attached, written to and read from on the same thread (whether from signal code or
 *    program code). Thus, they are NOT thread-safe. Given the current specification, I don't
 *    think it's possible to safely update an attached context from a different thread, since
 *    the signal handler doesn't assume the context can be written to concurrently from another
 *    thread.
 */
struct ddog_ThreadContextRecord *ddog_otel_thread_ctx_attach(struct ddog_ThreadContextRecord *ctx);

/**
 * Remove the currently attached context from the TLS slot.
 *
 * Returns the detached context (caller now owns it and must release it with
 * `ddog_otel_thread_ctx_free`), or null if the slot was empty.
 */
struct ddog_ThreadContextRecord *ddog_otel_thread_ctx_detach(void);

/**
 * Update the currently attached context in-place.
 *
 * If no context is currently attached, one is created and attached, equivalent to calling
 * `ddog_otel_thread_ctx_new` followed by `ddog_otel_thread_ctx_attach`.
 */
void ddog_otel_thread_ctx_update(const uint8_t (*trace_id)[16],
                                 const uint8_t (*span_id)[8],
                                 const uint8_t (*local_root_span_id)[8]);

#ifdef __cplusplus
}  // extern "C"
#endif  // __cplusplus

#endif  /* DDOG_OTEL_THREAD_CTX_H */
```

</details>

# Motivation

OTel thread-level context has been implemented in #1791 in order to provide better interop with the OTel eBPF profiler. The first user is supposed to be dd-trace-rs, but it turns out the dotnet SDK people are interested in using it as well (and eventually other non-Rust SDKs will use it and thus require an FFI).

# Additional Notes

N/A

# How to test the change?

There's a test to check that the TLS symbol is properly handled. For real usage, we plan to check when integrating in dotnet (or whichever is the first SDK to use it).

### DIFF
--- a/libdd-otel-thread-ctx-ffi/build.rs
+++ b/libdd-otel-thread-ctx-ffi/build.rs
@@ -17,11 +17,15 @@ fn main() {
     // Passing our own version script with an explicit `global:` entry for the
     // symbol beats the `local: *` wildcard and prevents that relaxation.
     //
-    // Merging multiple version scripts is not supported by GNU ld, so we also
-    // force lld explicitly.
+    // Merging multiple version scripts is not supported by GNU ld, but lld (or rather rust-lld) is
+    // the default linker picked by Rust since v1.70+. We use to force lld via:
+    // `println!("cargo:rustc-cdylib-link-arg=-fuse-ld=lld")` but this caused extra pain in CI
+    // because it requires a system wide lld install (instead of using rust-lld). Since we have a
+    // test to check that the symbol was properly propagated, we rather let Rust pick lld which is
+    // preferred anyway, and if it's not possible in an exotic build setup, the test should catch
+    // it.
     if target_os == "linux" {
         let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
-        println!("cargo:rustc-cdylib-link-arg=-fuse-ld=lld");
         println!(
             "cargo:rustc-cdylib-link-arg=-Wl,--version-script={manifest_dir}/tls-dynamic-list.txt"
         );
--- a/libdd-otel-thread-ctx-ffi/build.rs
+++ b/libdd-otel-thread-ctx-ffi/build.rs
@@ -3,12 +3,11 @@
 extern crate build_common;
 
 use build_common::generate_and_configure_header;
+use std::env;
 
 fn main() {
-    let header_name = "otel-thread-ctx.h";
-    generate_and_configure_header(header_name);
-
-    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+    generate_and_configure_header("otel-thread-ctx.h");
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
 
     // Export the TLSDESC thread-local variable to the dynamic symbol table so
     // external readers (e.g. the eBPF profiler) can locate it. Rust's cdylib
@@ -21,7 +20,7 @@ fn main() {
     // Merging multiple version scripts is not supported by GNU ld, so we also
     // force lld explicitly.
     if target_os == "linux" {
-        let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+        let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
         println!("cargo:rustc-cdylib-link-arg=-fuse-ld=lld");
         println!(
             "cargo:rustc-cdylib-link-arg=-Wl,--version-script={manifest_dir}/tls-dynamic-list.txt"
@@ -31,8 +30,7 @@ fn main() {
         // locate the cdylib without fragile path-walking.
         // OUT_DIR = <target>/[<triple>/]<profile>/build/<pkg>-<hash>/out
         // Three levels up lands on <target>/[<triple>/]<profile>.
-        let out_dir = std::env::var("OUT_DIR").unwrap();
-        let profile_dir = std::path::PathBuf::from(&out_dir)
+        let profile_dir = std::path::PathBuf::from(&env::var("OUT_DIR").unwrap())
             .ancestors()
             .nth(3)
             .unwrap()
--- a/libdd-otel-thread-ctx-ffi/build.rs
+++ b/libdd-otel-thread-ctx-ffi/build.rs
@@ -17,15 +17,11 @@ fn main() {
     // Passing our own version script with an explicit `global:` entry for the
     // symbol beats the `local: *` wildcard and prevents that relaxation.
     //
-    // Merging multiple version scripts is not supported by GNU ld, but lld (or rather rust-lld) is
-    // the default linker picked by Rust since v1.70+. We use to force lld via:
-    // `println!("cargo:rustc-cdylib-link-arg=-fuse-ld=lld")` but this caused extra pain in CI
-    // because it requires a system wide lld install (instead of using rust-lld). Since we have a
-    // test to check that the symbol was properly propagated, we rather let Rust pick lld which is
-    // preferred anyway, and if it's not possible in an exotic build setup, the test should catch
-    // it.
+    // Merging multiple version scripts is not supported by GNU ld, so we also
+    // force lld explicitly.
     if target_os == "linux" {
         let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+        println!("cargo:rustc-cdylib-link-arg=-fuse-ld=lld");
         println!(
             "cargo:rustc-cdylib-link-arg=-Wl,--version-script={manifest_dir}/tls-dynamic-list.txt"
         );
--- a/libdd-otel-thread-ctx-ffi/build.rs
+++ b/libdd-otel-thread-ctx-ffi/build.rs
@@ -25,18 +25,5 @@ fn main() {
         println!(
             "cargo:rustc-cdylib-link-arg=-Wl,--version-script={manifest_dir}/tls-dynamic-list.txt"
         );
-
-        // Expose the profile output directory to integration tests so they can
-        // locate the cdylib without fragile path-walking.
-        // OUT_DIR = <target>/[<triple>/]<profile>/build/<pkg>-<hash>/out
-        // Three levels up lands on <target>/[<triple>/]<profile>.
-        let profile_dir = std::path::PathBuf::from(&env::var("OUT_DIR").unwrap())
-            .ancestors()
-            .nth(3)
-            .unwrap()
-            .to_str()
-            .unwrap()
-            .to_owned();
-        println!("cargo:rustc-env=CDYLIB_PROFILE_DIR={profile_dir}");
     }
 }
--- a/libdd-otel-thread-ctx-ffi/build.rs
+++ b/libdd-otel-thread-ctx-ffi/build.rs
@@ -1,0 +1,29 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+extern crate build_common;
+
+use build_common::generate_and_configure_header;
+
+fn main() {
+    let header_name = "otel-thread-ctx.h";
+    generate_and_configure_header(header_name);
+
+    // Export the TLSDESC thread-local variable to the dynamic symbol table so
+    // external readers (e.g. the eBPF profiler) can locate it. Rust's cdylib
+    // linker applies a version script with `local: *` that hides all symbols
+    // not explicitly whitelisted, and also causes lld to relax the TLSDESC
+    // access to local-exec (LE), eliminating the dynsym entry entirely.
+    // Passing our own version script with an explicit `global:` entry for the
+    // symbol beats the `local: *` wildcard and prevents that relaxation.
+    //
+    // Merging multiple version scripts is not supported by GNU ld, so we also
+    // force lld explicitly.
+    #[cfg(target_os = "linux")]
+    {
+        let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+        println!("cargo:rustc-link-arg=-fuse-ld=lld");
+        println!(
+            "cargo:rustc-link-arg=-Wl,--version-script={manifest_dir}/tls-dynamic-list.txt"
+        );
+    }
+}
--- a/libdd-otel-thread-ctx-ffi/build.rs
+++ b/libdd-otel-thread-ctx-ffi/build.rs
@@ -8,6 +8,8 @@ fn main() {
     let header_name = "otel-thread-ctx.h";
     generate_and_configure_header(header_name);
 
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+
     // Export the TLSDESC thread-local variable to the dynamic symbol table so
     // external readers (e.g. the eBPF profiler) can locate it. Rust's cdylib
     // linker applies a version script with `local: *` that hides all symbols
@@ -18,12 +20,25 @@ fn main() {
     //
     // Merging multiple version scripts is not supported by GNU ld, so we also
     // force lld explicitly.
-    #[cfg(target_os = "linux")]
-    {
+    if target_os == "linux" {
         let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
-        println!("cargo:rustc-link-arg=-fuse-ld=lld");
+        println!("cargo:rustc-cdylib-link-arg=-fuse-ld=lld");
         println!(
-            "cargo:rustc-link-arg=-Wl,--version-script={manifest_dir}/tls-dynamic-list.txt"
+            "cargo:rustc-cdylib-link-arg=-Wl,--version-script={manifest_dir}/tls-dynamic-list.txt"
         );
+
+        // Expose the profile output directory to integration tests so they can
+        // locate the cdylib without fragile path-walking.
+        // OUT_DIR = <target>/[<triple>/]<profile>/build/<pkg>-<hash>/out
+        // Three levels up lands on <target>/[<triple>/]<profile>.
+        let out_dir = std::env::var("OUT_DIR").unwrap();
+        let profile_dir = std::path::PathBuf::from(&out_dir)
+            .ancestors()
+            .nth(3)
+            .unwrap()
+            .to_str()
+            .unwrap()
+            .to_owned();
+        println!("cargo:rustc-env=CDYLIB_PROFILE_DIR={profile_dir}");
     }
 }
--- a/libdd-otel-thread-ctx-ffi/build.rs
+++ b/libdd-otel-thread-ctx-ffi/build.rs
@@ -4,23 +4,60 @@ extern crate build_common;
 
 use build_common::generate_and_configure_header;
 use std::env;
+use std::path::PathBuf;
+use std::process::Command;
+
+/// Locate the `gcc-ld/` shim directory shipped with the Rust toolchain.
+///
+/// This directory contains an `ld.lld` wrapper that delegates to `rust-lld`.
+/// Passing it via `-B` to the C compiler driver makes it discover rust-lld
+/// before any system-wide lld, which matters when the system lld is too old
+/// (e.g. lld 7 on CentOS 7 cannot handle TLSDESC relocations).
+///
+/// The discovery goes through `rustc --print sysroot` so it works on any
+/// layout (NixOS, Homebrew, rustup, distro packages) without assuming FHS.
+fn find_rust_lld_dir() -> Option<PathBuf> {
+    let rustc = env::var("RUSTC").unwrap_or_else(|_| "rustc".into());
+    let target = env::var("TARGET").ok()?;
+
+    let output = Command::new(&rustc)
+        .arg("--print")
+        .arg("sysroot")
+        .output()
+        .ok()?;
+
+    let sysroot = std::str::from_utf8(&output.stdout).ok()?.trim();
+    let dir = PathBuf::from(sysroot)
+        .join("lib/rustlib")
+        .join(&target)
+        .join("bin/gcc-ld");
+
+    dir.join("ld.lld").exists().then_some(dir)
+}
 
 fn main() {
     generate_and_configure_header("otel-thread-ctx.h");
     let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
 
     // Export the TLSDESC thread-local variable to the dynamic symbol table so
-    // external readers (e.g. the eBPF profiler) can locate it. Rust's cdylib
+    // external readers (e.g. the eBPF profiler) can discover it. Rust's cdylib
     // linker applies a version script with `local: *` that hides all symbols
     // not explicitly allowlisted, and also causes lld to relax the TLSDESC
     // access to local-exec (LE), eliminating the dynsym entry entirely.
     // Passing our own version script with an explicit `global:` entry for the
     // symbol beats the `local: *` wildcard and prevents that relaxation.
     //
-    // Merging multiple version scripts is not supported by GNU ld, so we also
-    // force lld explicitly.
+    // Merging multiple version scripts is not supported by GNU ld, so we need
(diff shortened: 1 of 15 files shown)


libdd-otel-thread-ctx-ffi/build.rs
```diff
@@ -0,0 +1,29 @@
...
+use std::env;
+
+fn main() {
+    generate_and_configure_header("otel-thread-ctx.h");
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+
+    // Export the TLSDESC thread-local variable to the dynamic symbol table so
+    // external readers (e.g. the eBPF profiler) can locate it. Rust's cdylib
+    // linker applies a version script with `local: *` that hides all symbols
+    // not explicitly whitelisted, and also causes lld to relax the TLSDESC
```
ivoanjo:
```suggestion
    // not explicitly allowlisted, and also causes lld to relax the TLSDESC
```

:)

libdd-otel-thread-ctx/src/lib.rs
```diff
@@ -140,7 +140,7 @@ pub mod linux {
     // `valid`. We just use a const assertion in `new()` to avoid surprises and make sure this
     // struct has the right total size.
     #[repr(C)]
-    struct ThreadContextRecord {
+    pub struct ThreadContextRecord {
```
ivoanjo: I'm not quite sure about this part. In particular, I don't think we're giving callers the tools that they need to directly use this without getting "burned" right now: no docs, no validation. :hurtrealbad: 

My "gut feeling" is: if callers haven't asked for this yet, let's not give it to them. 🤔 

If they want to bang at the bits directly, I think in that situation they might as well prefer to reimplement it from scratch, e.g. why half-go through libdatadog if you want maximum control anyway?

ivoanjo (commented):
This looks quite reasonable. Some notes:

1. I would strongly recommend getting one of the dotnet folks (which I believe are intended first users of this API) to give a pass on the PR

2. I think we're missing an end-to-end example in C or C++. In particular, one that sets both the process and thread context, and thus can be picked up correctly by the external reader.

    Right now we're kinda assuming that callers will "just know" how to put this together, and suspect that will cause a bunch of confusion. Thus, it would be very helpful to provide some docs + a fully working example that could be used as a model.

2. As I mentioned, I'm not convinced about the whole "expose the full structure", see my comment on why.

libdd-otel-thread-ctx/src/lib.rs
yannham: > I'm not quite sure about this part. In particular, I don't think we're giving callers the tools that they need to directly use this without getting "burned" right now: no docs, no validation. :hurtrealbad:

I would say that there is some documentation (and warnings!) in the documentation of `attach`. When talking with dotnet people, we thought it might make sense for performance to try to do everything on the SDK side for context switching, without going through native. I agree that then using libdatadog is questionable, but I think you'll still have to manage the whole TLSDESC/elf business. It might be reasonable to use libdatadog to attach the first context and then update it in place. 

Also the structure is defined per the spec anyway, so it's not going to change and is "public" (as is, defined somewhere on the internet).

All of that being said,  I agree that making it visible here is stronger, as we guarantee our FFI functions always return pointer to contexts as per the spec. Additionally, the in-place-update-from-the-sdk was really mentioned as "we can try that later in the future in mandated", so I would be totally fine reverting the changes to `otel-thread-ctx`, removing the bit about raw update, making the record type an entirely opaque pointer, and only discuss this again after there's a need for it, or at least some benchmarks.

Why do you dotnet guys think @gleocadie ?

ivoanjo: > I think you'll still have to manage the whole TLSDESC/elf business.

I was thinking that if we're building a .so from C/C++, it's even easier to not use libdatadog for this part -- no need to fight linker, just declare a public thread-local symbol.

But yeah, as I mentioned above -- my suggestion is "gut feeling we're not gonna need it" -- if y'all want to try this path, we can, but I would suggest avoiding any "let's expose this now just-in-case-for-later" things ;)

yannham: > I was thinking that if we're building a .so from C/C++, it's even easier to not use libdatadog for this part -- no need to fight linker, just declare a public thread-local symbol.

That's true... although you still need the process context part, and re-implementing this in C/C++ starts to be a tad less trivial. But I agree that if you only need to swap the TLS slot, you're probably better off with a thin C/C++ shared lib than messing around with the Rust build :+1:

ivoanjo: Yeap, I agree you'd still probably want to use the libdatadog process context impl, assuming whatever downstream consumer we're talking about is pulling libdatadog already as a dependency anyway.

chrisnas: > I would be totally fine reverting the changes to otel-thread-ctx, removing the bit about raw update, making the record type an entirely opaque pointer, and only discuss this again after there's a need for it, or at least some benchmarks.

We would definitively want to use in-place value change for .NET for performance reason due to the high number of span that can be created over a trace.

This is EXACTLY how it works today: the tracer asked the profiler (equivalent of libdatadog call) ONCE per thread where to set the span id value and then, directly update the value in memory when the span changes for the thread.

libdd-otel-thread-ctx/src/lib.rs
chrisnas, line 168: Is this the place where to store the endpoint?
If this is the case, how to know the string ID corresponding the EndPoint attribute. Since the value seems to also be an interned string ID, it means that each different end must be interned in the libdatadog string table; growing over time since the endpoints could be very different

Don't forget that the endpoint is usually know close to the end of the request. It means that any client would have to "reconciliate" the real endpoint for the samples already created for this thread. More important, there is no notification for the full host to know WHEN a thread gets its endpoint so it looks like impossible to be sure of the right endpoint to use if not already set; i.e. the same thread could have switch request since the last time samples where created.

Also, how to reset the current end point? by setting an "empty string" ID?

gleocadie: What interning API should we use ?

libdd-otel-thread-ctx-ffi/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    /// Free an owned thread context.
+    ///
+    /// # Safety
+    ///
+    /// `ctx` must be a valid non-null pointer obtained from `ddog_otel_thread_ctx_new` or
+    /// `ddog_otel_thread_ctx_detach`, and must not be used after this call. In particular, `ctx`
+    /// must not be currently attached to a thread.
+    #[no_mangle]
+    pub unsafe extern "C" fn ddog_otel_thread_ctx_free(ctx: *mut ThreadContextRecord) {
```
chrisnas: Just to confirm the expected usage:

the first time a thread needs to set the context:
```
pCurrentThreadContext = ddog_otel_thread_ctx_new(...);
ddog_otel_thread_ctx_attach(pCurrentThreadContext)

```

when a span is processed by a thread:
```
// get current thread's context
...

// directly set the value of span id/trace id (in case of reset)
...
```

However, on shutdown, I don't see a way to "detach" all contextes but only for the current thread. Not sure it would be safe to call ddog_otel_thread_ctx_free() on each context without detaching first

libdd-otel-thread-ctx/src/lib.rs
yannham: > . Since the value seems to also be an interned string ID

They are not, only the _keys_ are interned string IDs. I think that what we drafted in our first discussion, that is having a small set of known keys with fixed IDs as a convention (`root_span_id` is 0, `endpoint` or whatever is the end point attribute is `1`, etc.) sounds like the way to go, ideally? So there would be no need for dynamic string interning business at all.

For the endpoint question, I'm not sure to be honest. Even beyond the current spec, given the setup and the constraints of an external eBPF profiler, I don't see an easy way out of having the profiler do some additional work and book-keeping to correlate past samples once the endpoint become known. Maybe @ivoanjo has some thoughts on this.

libdd-otel-thread-ctx-ffi/src/lib.rs
yannham: One possible answer is that you don't do anything, if you think this can be acceptable: if we're shutting down just before the process exit for example, the OS will reclaim the memory anyway, so not sure it's worth doing it manually just before.

Otherwise I admit the thread clean-up story might be non-trivial. If you need to free contexts either all at once, or during the app lifecycle as threads are created and freed, I can see two ways:

- either do it as some kind of exit callback/cleanup phase on the thread before it exits (if doable, easiest)
- or store aliases to the contexts in a separate datastructure (says a map threadid -> contexts, or a queue of to-be-cleaned, or even a pool of to-be-reused contexts), and free or reuse after the corresponding thread exited. Once the original thread is dead, the TLS has been de-allocated anyway so the profiler can't reach it anymore. It's then safe to call `ddog_otel_thread_ctx_free()` from another thread or to repurpose the context.

libdd-otel-thread-ctx/src/lib.rs
ivoanjo: OK so there's two versions of this answer for the "endpoint shows up late" problem.

**The in-process datadog version of this answer**: While we want to adopt this mechanism, we don't have to adopt it **exclusively**. E.g. [like we did for Java](https://github.com/DataDog/java-profiler/pull/347), if it's the datadog tracer talking to the datadog in-process profiler, we can have additional channels beyond just this one to provide extra information if needed. So for sure we don't want to regress in terms of feature-set, and the Java PR is a great example of how we both implement the otel spec and still keep the entire feature-set.

**The out-of-process and pure-otel version of this answer** to me is: I think we'll have to live with best-effort for now. That is -- we set the endpoint name in the context when we get it, and if the eBPF Profiler misses it, such is life. If an otel backend wants an accurate picture, the otel backend will probably need to look at traces/spans for this information.

The whole basis of the thread context sharing mechanism is to be a "loose coordination" kinda thing; I'm hoping to avoid expanding the scope for now because it opens a lot of hard questions that would make the implementation a lot more complex IMO.

libdd-otel-thread-ctx/src/lib.rs
ivoanjo: The question here is -- once an implementation moves to writing the context directly, then I don't think they'll get very much out of libdatadog for the thread context.

Writing the format and changing the pointer is 90% of this (and to be clear, it's 90% of not very much -- we're talking about < 1000 LoC on the libdatadog side) so I personally wonder if using libdd-otel-thread-ctx for the remaining 10% is worth or not.

(And I guess we'll find out the answer once we try to implement? And I'm happy to review :D and provide guidance)

yannham: > so I personally wonder if using libdd-otel-thread-ctx for the remaining 10% is worth or not.

One argument could be: it's code that's already written and built (or will be soon), takes care of the whole annoying business around linking and symbols and hopefully gets memory ordering stuff right (even if only for the initial publish).

So I could revert the question: any reason to _not_ use libdd-otel-thread-ctx? If it's for size, it's also possible to feature-gate the update function, so that you only pay for setup. It gets the native stuff out of the way.

ivoanjo: > any reason to not use libdd-otel-thread-ctx?

It depends on the implementation, but for instance [I don't think what we did for Java](https://github.com/DataDog/java-profiler/pull/347) -- specifically embedding the structure within another structure and precisely controlling initialization -- would be possible when going through libdatadog.

yannham: I guess you meant would be _impossible_? I'm not familiar with what they do in Java that's special to them (and couldn't get that quickly from the PR), but I trust there are custom cases where you'd like to escape from libdatadog. Since dotnet is the first motivation here, I guess it's @chrisnas' question to answer: even in the case of doing the update directly from dotnet, do you still see value in libdatadog for setting up the process context and attach the first thread context? Or do you already need to fiddle with native code in your codebase already, and would be inclined to just go with that entirely?

I would like to move forward with an FFI anyway, because I have some small benchmarks that rely on its existence. However, if it's [edit: NOT so] useful for dotnet in the end, I can trim the PR to make the context private again, which would be minimal first version.

libdd-otel-thread-ctx-ffi/src/lib.rs
chrisnas: The only risk I see comes from "short lived" threads for which it would be better to be able to release the memory. In .NET, we might be notified, from the exiting thread, that it will dies soon and call the `free` API but I don't know for the other runtimes

libdd-otel-thread-ctx/src/lib.rs
ivoanjo: > I guess you meant would be impossible?

I _suspect_ it would be impossible; but I may be missing some detail.

> Since dotnet is the first motivation here, I guess it's @chrisnas' question to answer: even in the case of doing the update directly from dotnet, do you still see value in libdatadog for setting up the process context and attach the first thread context?

To be clear I think process context is very worth using the libdatadog one. The thread context is the one where I think the options are 100% libdatadog or 0% libdatadog -- I don't think 10% libdatadog is going to be very useful and I suspect it'll be more confusing/error-prone than not ;)

> I would like to move forward with an FFI anyway, because I have some small benchmarks that rely on its existence. However, if it's [edit: NOT so] useful for dotnet in the end, I can trim the PR to make the context private again, which would be minimal first version.

Idk, I'm still in the "I don't think it's very useful to add stuff if we don't think we'll need them", but also we can and should evolve the API with what we learn from adding this to dotnet, so from me not a blocker at all.

libdd-otel-thread-ctx-ffi/src/lib.rs
yannham: Ok, good for .NET then! For other runtimes, as I said there are other possibilities (overall map of active contexts/queue/pool). If you can test if a thread is still active, you can maybe do some kind of cleanup in the background or at regular intervals.

libdd-otel-thread-ctx/src/lib.rs
yannham: > I don't think 10% libdatadog is going to be very useful and I suspect it'll be more confusing/error-prone than not ;)

I'm not sure I get why. Once again you need to do some stuff in native anyway, so if you can reuse/offload and pull libdatadog's parts for the process context, it's strictly better to re-use what you can for thread context than re-implementing it yourself in unsafe C/C++ on the side...?

At any rate, I want to insist that the whole debate here is purely on the contract/visibility part of the API. There' absolutely no code or logic involved in making the thread context updatable purely on the .NET runtime side. It's mostly about making some C struct exported or not in the header, but this struct's layout is available in the spec anyway. My personal view is that it doesn't cost much to expose the possibility as long as we have that big CAUTION in the code comments (like: you can do it, but at your own risk). I think @ivoanjo is wary of making this part of the official public API because it's full of traps, which I can also understand.

So here is a proposal: for now, let's get the record struct out of the API and make returned values opaque pointers. Still, the code and the memory layout will be exactly the same as if we didn't. So it's not exposed in the API, but if @chrisnas wants to experiment with in-place update, they can and have everything they need to do it (I don't see any reason why we would change the layout or introduce indirection on the libdatadog side in the forseeable future). Once we see how it turns out, and if .NET think it's worth keeping libdatadog in that setting, we can update the FFI and make it more "official", or do nothing otherwise and keep the opaque version. That would unblock this PR and be forward-compatible with changing our mind.

What do you guys think?

libdd-otel-thread-ctx/src/lib.rs
chrisnas: > So there would be no need for dynamic string interning business at all
Good news   :^)
So, how are we supposed to set the string corresponding to the value of the interned attribute corresponding to the endpoint?

@ivoanjo: are there any other attributes that would make sense to "attach"

libdd-otel-thread-ctx/src/lib.rs
ivoanjo: >  it's strictly better to re-use what you can for thread context than re-implementing it yourself in unsafe C/C++ on the side...?

If you go 100% libdatadog thread context, then yes, you get all the benefits of the rust safety.

But to me this argument falls apart once you need to implement the logic to bang on the bytes directly. In that situation, you might as well bang on the bytes _always_, not "half the time is libdatadog thread context, half the time is my own code" -- I think that'll be more error prone and hard to follow. It's even harder to test ;)

> I think @ivoanjo is wary of making this part of the official public API because it's full of traps, which I can also understand.

My argument is that if you're going to do things directly, you need to really understand the spec (it's the "play in hard mode choice").. So at that point, you can look at libdatadog thread context as a reference for code, but I'm not sure having the struct definition or some other small tidbits helps very much -- you really do need to get into the spec to understand how to use it, no shortcuts anymmore.

Or, you can play in easy mode and use libdatadog thread context -- to me that's the choice here. Easy or hard, 100% libdatadog thread context or 0%.

> So here is a proposal

Ack from my side!

libdd-otel-thread-ctx/src/lib.rs
ivoanjo: [I've been keeping a list here](https://docs.google.com/document/d/1IwjjVJzEChcFPcnVV2N5Kkjg-4_Q4v4Q3ojpxntbdvY/edit?pli=1&tab=t.efaosgjya44c#bookmark=id.700gvw31vb7h).

I'd say definitely the `datadog.local_root_span_id` and the (best-effort, with limitations) `datadog.trace_endpoint`, as well as the `datadog.thread_name` and `datadog.thread_id` (in cases where thread id doesn't match what the OS can see).

Beyond that I see we in `prof-dotnet` pprofs that we have `appdomain name`, `appdomain process id` -- if you think those are useful, maybe throw them in there too? If not, don't ;) (If you do -- I suggest calling them e.g. `datadog.appdomain_name` and updating the doc I linked so we can try to keep everyone in-sync and avoid the backend needing to know 3 datadog names for the same thing)

libdd-otel-thread-ctx/src/lib.rs
yannham: > But to me this argument falls apart once you need to implement the logic to bang on the bytes directly. In that situation, you might as well bang on the bytes always, not "half the time is libdatadog thread context, half the time is my own code" -- I think that'll be more error prone and hard to follow. It's even harder to test ;)

The big difference to me is that the goal here is _to bang the bytes from the .DOTNET runtime_. Since the point is to not pay for the FFI. So this byte banging part is done in C# or whatnot. However, to install the thread context, it's very likely that you need a native shared library anyway, because of the whole TLSDESC and linker business. Hence, unless like Java you want to do custom/specific stuff in the native byte banging part, the trade-off is using ready-built libdatadog for installing the context versus creating a separate C/C++/Rust project on the side just for that, figure out the TLS dialect, linking business, etc. That is, redo mostly what's already done in `otel-thread-ctx`. Of course if you're going to bang the byte from native, then sure.

ivoanjo: > [...] because of the whole TLSDESC and linker business [...]

Right, but the tlsdesc is very easy to do from C/C++ and really hard to do from rust, so again, I'm not seeing a lot of benefits :P

The dotnet runtime isn't directly calling into rust, there's already a bunch of C++ glue code ;)

ivoanjo (approved):
👍 LGTM

PR #1920, opened 2026-04-24
Add size-benchmark crate with FFI coverage and CI integration

Eldolfin:
Experimenting with another way of measuring the final binary size with a auto-generated binary that has an array of pointers of all functions of public *-ffi functions so that they are included at link time.
The binary is optimized it with lots of LTO and other binary size optimisation tricks so that there is less noise of things that would go away for users.

# Motivation

# Additional Notes

# How to test the change?


Eldolfin:
Cool experiment, closing to reduce noise

PR #1923, opened 2026-04-27
chore(ci): adds a cron on all actions running a tests

cbeauchesne:
# What does this PR do?

Per title

# Motivation

Few months ago, dd-trace-java drops by accident two thirds of its test base. Twice. 

Following this incidents, we realized that with our growing test base, it was relatively easy to drop a part of it, as a green CI is always a trigger of "don't check anymore" (which is fine). So we implemented a safe guard on test optim that triggers an alert if ever a repo does does not report its tests to Test Optim. And this monitor caught one other occurences few weeks later ...

The caveats is that we do not have any reliable way to trigger it : once a commit is pushed, we can wait for several minutes, even hours before having all the tests reported. So we implemented it with a compromise : every day we must have the full test base. In consequences, we need to have a schedule that run the test base once a day (including WE, limitation of datadog monitors) , to avoid triggering an alert for days that does not have any commit. 

I also adapted the concurrency clause to have all runs runnings on pushes on `main`

# Additional Notes

# How to test the change?


.github/workflows/coverage.yml
```diff
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - main
+  schedule:
+    - cron: '0 6 * * *'
```
ekump, line 9: This is 1am ET / 7am CET, which guarantees no one will be around if it fails. I assume that's ok?

ekump (approved):
I'm not sure I fully understand what problem this is solving, but I see no harm in it, if it's a precautionary measure. LGTM.

.github/workflows/coverage.yml
cbeauchesne: Yes, trying to not use CPU when they are needed. And it's fine if they fail, the Green CI SLI will track this as any other run.

PR #1927, opened 2026-04-27
refactor(sampling): move the sampling logic from dd-trace-rs [APMSP-2946]

bantonsson:
# What does this PR do?

Moves the sampling logic from `dd-trace-rs` so that it can be reused.

# Motivation

Reuse all the things.

# Additional Notes

Has been tested and benchmarked with the code in `dd-trace-rs`.

# How to test the change?

Unit tests and benchmarks are here.


libdd-sampling/src/rate_limiter.rs
bwoebi, line 1: Do you know that there's already a rate limiter in libdd-common/src/rate_limiter.rs?

bantonsson: Nope I didn't. Thanks. I'll look at it.

bwoebi: It's by the way the limiter PHP uses, and is a sliding window rather than discrete windows.

libdd-sampling/src/sampling_rule_config.rs
```diff
@@ -0,0 +1,232 @@
...
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::fmt::Display;
+use std::ops::Deref;
+use std::str::FromStr;
+
+/// Configuration for a single sampling rule
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
+pub struct SamplingRuleConfig {
```
iunanua, line 12: `SamplingRuleConfig` is almost identical to [TracingSamplingRule](https://github.com/DataDog/libdatadog/blob/main/datadog-remote-config/src/config/dynamic.rs#L69)

Someday we could unify them:  `TracingSamplingRule` reveived via RC update config rules that update sampler rules...

bantonsson: yes, there is an overlap with the types now so that `dd-trace-rs` can work before we refactor the RC client

libdd-sampling/src/rate_limiter.rs
bantonsson: I had a dig through the specs for `_dd.limit_psr` and how it should be reported back to the agent, which is what this rate limiter is implementing, and I can't see how you could do that with the rate limiter in `libdd-common`.

libdd-sampling/src/glob_matcher.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    /// Returns the pattern (lowercase version)
+    pub fn pattern(&self) -> String {
+        self.pattern_lower.clone()
+    }
+
+    /// Checks if the given subject matches the glob pattern
+    /// The match is case insensitive.
+    pub fn matches(&self, subject: &str) -> bool {
+        let subject_lower = subject.to_lowercase();
```
ekump, line 54: `to_lowercase()` is going to allocate a new string on every match. Would it be more performant to use `eq_ignore_ascii_case()` when checking for a match below?

libdd-sampling/src/agent_service_sampler.rs
```diff
@@ -0,0 +1,51 @@
...
+
+#[derive(Debug, serde::Deserialize)]
+pub struct AgentRates<'a> {
+    #[serde(borrow)]
+    pub rate_by_service: Option<HashMap<&'a str, f64>>,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct ServicesSampler {
+    inner: Arc<RwLock<HashMap<String, RateSampler>>>,
```
yannham, line 21: Is this map is likely to be contended in write mode? if yes, maybe a sharded map could be considered.

libdd-sampling/src/agent_service_sampler.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub fn get(&self, service: &str) -> Option<RateSampler> {
+        self.inner.read().unwrap().get(service).cloned()
+    }
+
+    pub fn update_rates<I: IntoIterator<Item = (String, f64)>>(&self, rates: I) {
+        let new_rates: HashMap<_, _> = rates
+            .into_iter()
+            .map(|(s, r)| (s, RateSampler::new(r)))
+            .collect();
+        *self.inner.write().unwrap() = new_rates;
```
yannham: We could avoid re-allocation here by reusing the hashmap, instead of re-allocating a new one (`clear + extend`). Though if it's not a common operation, it might not be worth doing (also, it keeps the previous allocation, which can waste space).

libdd-sampling/src/agent_service_sampler.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    // used for testing purposes
+
+    #[allow(dead_code)]
```
yannham: Can't we avoid the `allow(dead_code)` with `[cfg(test)]` or something like that?

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+            let Some(new_rates) = new_rates.rate_by_service else {
+                return;
+            };
+            service_samplers.update_rates(new_rates.into_iter().map(|(k, v)| (k.to_string(), v)));
+        })
+    }
+
+    /// Creates a callback for updating sampling rules from remote configuration
+    /// # Returns
+    /// A boxed function that takes a slice of SamplingRuleConfig and updates the sampling rules
```
yannham:
```suggestion
    /// Creates a callback for updating sampling rules from remote configuration
    ///
    /// # Returns
    ///
    /// A boxed function that takes a slice of SamplingRuleConfig and updates the sampling rules
```

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+            rules_sampler.update_rules(new_rules);
+        })
+    }
+
+    /// Computes a key for service-based sampling
+    fn service_key(&self, span: &impl SpanProperties) -> String {
+        // Get service from span
+        let service = span.service().into_owned();
+        // Get env from span
+        let env = span.env();
```
yannham: Nit: comments don't add much

```suggestion
        let service = span.service().into_owned();
        let env = span.env();
```

libdd-sampling/src/glob_matcher.rs
```diff
@@ -0,0 +1,217 @@
...
+/// The glob pattern language supports `*` as a multiple character wildcard (including empty string)
+/// and `?` as a single character wildcard (not including empty string). The match is case
+/// insensitive.
+///
+/// This implementation includes an LRU cache for faster repeated matching.
+pub struct GlobMatcher {
+    /// Lowercased pattern for case-insensitive matching
+    pattern_lower: String,
+    /// LRU cache of previously matched strings to their results
+    cache: Mutex<LruCache<String, bool>>,
```
ekump: Do we need to worry about contention with the mutex? I assume this will get accessed multiple times for every span? Have you considered a `RwLock` with `peek` to reduce contention on reads?

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+        data: &impl SamplingData,
+        span: &impl SpanProperties,
+    ) -> DdSamplingResult {
+        let mut is_keep = true;
+        let mut used_agent_sampler = false;
+        let sample_rate;
+        let mut rl_effective_rate: Option<f64> = None;
+        let trace_id = data.trace_id();
+
+        // Find a matching rule
```
yannham:
```suggestion
```

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+        let matching_rule = self.find_matching_rule(span);
+
+        // Apply sampling logic
+        if let Some(rule) = &matching_rule {
+            // Get the sample rate from the rule
```
yannham: I'll will stop hereafter with deleting comments, but I feel there are many more like instances where they don't add much information over the code and increase verbosity.

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+    } else {
+        (digits - 1 - magnitude) as usize
+    };
+
+    let s = format!("{:.prec$}", rounded, prec = decimal_places);
+    // Strip trailing zeros after decimal point
+    Some(if s.contains('.') {
+        let s = s.trim_end_matches('0');
+        let s = s.trim_end_matches('.');
+        s.to_string()
```
yannham:
```suggestion
        s.trim_end_matches('0')
            .trim_end_matches('.')
            .to_string();
```

libdd-sampling/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+//! This crate provides generic sampling infrastructure including:
+//! - Trait abstractions for trace IDs, attributes, and span properties
+//! - Rate-based sampling algorithms
+//! - Rate limiting functionality
+//! - Glob pattern matching for sampling rules
+//! - Sampling-related constants
+//! - Rule-based sampling with pattern matching
+//! - Agent-provided sampling rates
+//! - Complete Datadog sampler implementation
+
```
ekump, line 15:
```suggestion
#![cfg_attr(not(test), deny(clippy::panic))]
#![cfg_attr(not(test), deny(clippy::unwrap_used))]
#![cfg_attr(not(test), deny(clippy::expect_used))]
#![cfg_attr(not(test), deny(clippy::todo))]
#![cfg_attr(not(test), deny(clippy::unimplemented))]
```

libdd-sampling/src/rate_limiter.rs
```diff
@@ -0,0 +1,358 @@
...
+                true
+            } else {
+                false
+            }
+        }
+    }
+
+    /// Update counts used to determine effective rate
+    fn update_rate_counts(&self, allowed: bool, timestamp: Instant) {
+        let mut state = self.inner.lock().unwrap();
```
ekump: [lock_or_unwrap()](https://github.com/DataDog/libdatadog/blob/main/libdd-common/src/lib.rs#L79) will come in handy when you add the panic clippy checks.

libdd-sampling/src/rate_limiter.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    /// Checks if the current request is allowed and consumes a token if it is.
+    ///
+    /// # Returns
+    /// `true` if the request is allowed, `false` otherwise
+    pub fn is_allowed(&self) -> bool {
+        let now = Instant::now();
+        let allowed = self.is_allowed_at(now);
+        self.update_rate_counts(allowed, now);
```
ekump: `is_allowed_at()` and `update_rate_counts()` both acquire `self.inner.lock()`. Do you have to worry about `is_allowed_at()` being called on another thread in between the two lock acquisitions and causing some sort of skew? And is acquiring the lock twice per call to `is_allowed()` preferable for performance?

libdd-sampling/src/constants.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared constants for the datadog_opentelemetry::sampling crate
```
ekump:
```suggestion
//! Shared constants for the libdd-sampling crate
```

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+
+    /// Returns the sampling mechanism used for the decision
+    fn get_sampling_mechanism(
+        &self,
+        rule: Option<&SamplingRule>,
+        used_agent_sampler: bool,
+    ) -> SamplingMechanism {
+        if let Some(rule) = rule {
+            match rule.provenance.as_str() {
+                // Provenance will not be set for rules until we implement remote configuration
```
ekump: is this comment still relevant?

libdd-sampling/src/sampling_rule.rs
```diff
@@ -0,0 +1,610 @@
...
+        tags: Option<HashMap<String, String>>,
+        provenance: Option<String>,
+    ) -> Self {
+        // Create glob matchers for the patterns
+        let name_matcher = name.as_deref().and_then(matcher_from_rule);
+        let service_matcher = service.as_deref().and_then(matcher_from_rule);
+        let resource_matcher = resource.as_deref().and_then(matcher_from_rule);
+
+        // Create matchers for tag values
+        let tag_map = tags.clone().unwrap_or_default();
```
ekump: I don't think this `clone()` is necessary? It compiles when I remove it.

libdd-sampling/src/rate_limiter.rs
bwoebi: But, that's what the rate() function is for on the Limiter trait?

libdd-sampling/src/rate_limiter.rs
```diff
@@ -0,0 +1,358 @@
...
+        if state.tokens_total == 0 {
+            return 1.0;
+        }
+
+        // Get rate of tokens allowed
+        state.tokens_allowed as f64 / state.tokens_total as f64
+    }
+
+    /// Helper function to calculate the effective rate based on current and optional previous rate.
+    fn _calculate_internal_effective_rate(
```
ekump: This function is used.

```suggestion
    fn calculate_internal_effective_rate(
```

libdd-sampling/src/agent_service_sampler.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    pub fn update_rates<I: IntoIterator<Item = (String, f64)>>(&self, rates: I) {
+        let new_rates: HashMap<_, _> = rates
+            .into_iter()
+            .map(|(s, r)| (s, RateSampler::new(r)))
+            .collect();
+        *self.inner.write().unwrap() = new_rates;
+    }
+
+    // used for testing purposes
```
ekump: Can we just use `#[cfg(test)]` then instead of `#[allow(dead_code)]`?

libdd-sampling/src/rules_sampler.rs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        self.inner
+            .read()
+            .unwrap()
+            .iter()
+            .find(|rule| matcher(rule))
+            .cloned()
+    }
+
+    // used for testing purposes
```
ekump: Same [comment](https://github.com/DataDog/libdatadog/pull/1927/changes#r3191080922) as in agent_service_sampler. Can we use `#[cfg(test)]`?

libdd-sampling/src/glob_matcher.rs
```diff
@@ -0,0 +1,217 @@
...
+            cache.put(subject_lower, true);
+        }
+        true
+    }
+}
+
+impl Clone for GlobMatcher {
+    fn clone(&self) -> Self {
+        // Create a new matcher with the same pattern
+        // This doesn't clone the cache since each instance maintains its own cache
```
ekump: If I'm following correctly, every matching rule lookup in `find_matching_rule` clones `SamplingRule` which in turn clones every `GlobMatcher` inside it, which then constructs an empty `LruCache` with 256 slots allocated, and then drops it after the sampling decision is made. Is a per-instance cache here actually helping performance?

libdd-sampling/src/sampling_rule.rs
```diff
@@ -0,0 +1,610 @@
...
+
+    /// Samples a trace ID using this rule's sample rate
+    pub fn sample(&self, trace_id: &impl TraceIdLike) -> bool {
+        // Delegate to the internal rate sampler's new sample method
+        self.rate_sampler.sample(trace_id)
+    }
+}
+
+/// Represents a priority for sampling rules
+#[allow(dead_code)]
```
ekump, line 232: Any reason why we don't just wire up `RuleProvenance` now?

ekump (approved):
Could you add a ticket to Jira for this?

Left some minor comments and performance concern trolling. Nothing blocking. LGTM.

libdd-sampling/src/sampling_rule.rs
ekump: Oh, wait. I think [this](https://github.com/DataDog/libdatadog/pull/1927#discussion_r3190512705) may be related. These are necessary until we wire up RC with sampling, right? I think you can ignore both of those comments then.

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+    }
+
+    /// Returns Datadog-specific sampling tags to be added as attributes
+    ///
+    /// # Parameters
+    /// * `factory` - The attribute factory to use for creating attributes
+    ///
+    /// # Returns
+    /// An optional vector of attributes to add to the sampling result
+    pub fn to_dd_sampling_tags<F>(&self, factory: &F) -> Option<Vec<F::Attribute>>
```
yannham, line 276: Nit: is it useful here to distinguish between `None` and `Some([])`?

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+                        SAMPLING_KNUTH_RATE_TAG_KEY,
+                        std::borrow::Cow::Owned(rate_str),
+                    ));
+                }
+            }
+            _ => {}
+        }
+
+        let priority = self.priority;
+        result.push(factory.create_i64(SAMPLING_PRIORITY_TAG_KEY, priority.into_i8() as i64));
```
yannham, line 323:
```suggestion
        result.push(factory.create_i64(SAMPLING_PRIORITY_TAG_KEY, self.priority.into_i8() as i64));
```

libdd-sampling/src/dd_sampling.rs
```diff
@@ -0,0 +1,403 @@
...
+}
+
+impl FromStr for SamplingPriority {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.parse::<i8>() {
+            Ok(value) => Ok(SamplingPriority::from_i8(value)),
+            Err(_) => Err(()),
+        }
```
yannham, line 94:
```suggestion
        s.parse::<i8>().map(SamplingPriority::from_i8).map_err(|_| ())
```

libdd-sampling/src/dd_sampling.rs
```diff
@@ -0,0 +1,403 @@
...
+            mechanism::AGENT_RATE_BY_SERVICE | mechanism::DEFAULT => AUTO_PAIR,
+            mechanism::MANUAL
+            | mechanism::LOCAL_USER_TRACE_SAMPLING_RULE
+            | mechanism::REMOTE_USER_TRACE_SAMPLING_RULE
+            | mechanism::REMOTE_DYNAMIC_TRACE_SAMPLING_RULE
+            | mechanism::SPAN_SAMPLING_RULE
+            | mechanism::DATA_JOBS_MONITORING => USER_PAIR,
+            mechanism::APPSEC => AUTO_PAIR,
+
+            _ => AUTO_PAIR,
```
yannham, line 135:
```suggestion
            mechanism::AGENT_RATE_BY_SERVICE 
            | mechanism::DEFAULT
            | mechanism::APPSEC => AUTO_PAIR,
            mechanism::MANUAL
            | mechanism::LOCAL_USER_TRACE_SAMPLING_RULE
            | mechanism::REMOTE_USER_TRACE_SAMPLING_RULE
            | mechanism::REMOTE_DYNAMIC_TRACE_SAMPLING_RULE
            | mechanism::SPAN_SAMPLING_RULE
            | mechanism::DATA_JOBS_MONITORING => USER_PAIR,
            _ => AUTO_PAIR,
```

libdd-sampling/src/dd_sampling.rs
```diff
@@ -0,0 +1,403 @@
...
+            pair.keep
+        } else {
+            pair.reject
+        }
+    }
+
+    /// Returns the string representation of the sampling mechanism.
+    pub fn to_cow(self) -> Cow<'static, str> {
+        match self {
+            mechanism::DEFAULT => Cow::Borrowed("-0"),
```
yannham, line 152: Out of curiosity, why are the string values negative but the i8 constants positive below? Is it because we know it's always negative, so we store a value `[-255, 0}` in a `u8` ? If yes, it might be worth mentioning in the description of `value` above, because it's not entirely obvious

libdd-sampling/src/glob_matcher.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /// Returns the pattern (lowercase version)
+    pub fn pattern(&self) -> String {
+        self.pattern_lower.clone()
+    }
+
+    /// Checks if the given subject matches the glob pattern
+    /// The match is case insensitive.
+    pub fn matches(&self, subject: &str) -> bool {
```
yannham, line 53: If I understand correctly, what you do on wildcard is that you first try to match zero character, and if that fails you backtrack to the last wildcard, consumes zero+1 character, and iterate. First, this means this is potentially quadratic (but maybe that's ok/better than something smarter but more involved like KMP for small strings?). Also, correctness is not totally obvious (for example, when you have several wildcards characters). If there's a spec or an extended description somewhere, it could be nice to include it in the docstring or link to it, if possible.

libdd-sampling/src/rate_limiter.rs
```diff
@@ -0,0 +1,358 @@
...
+            .finish()
+    }
+}
+
+impl RateLimiter {
+    /// Creates a new RateLimiter with the given rate limit.
+    ///
+    /// # Parameters
+    /// * `rate_limit` - Maximum number of spans per second:
+    ///   * rate_limit > 0: max number of requests to allow per second
```
yannham: Is it really "per second", or actually "per `time_window_ns`" ?

libdd-sampling/src/rate_limiter.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    /// Replenish tokens based on elapsed time
+    fn replenish(&self, state: &mut RateLimiterState, timestamp: Instant) {
+        let elapsed = timestamp.duration_since(state.last_update);
+
+        // Calculate new tokens to add
+        let tokens_to_add_precise: f64 =
+            (elapsed.as_nanos() as f64 / state.time_window_ns as f64) * self.rate_limit as f64;
+        let tokens_to_add: i64 = tokens_to_add_precise as i64; // Truncates fractional tokens
```
yannham:
```suggestion
        let tokens_to_add = tokens_to_add_precise as i64; // Truncates fractional tokens
```

libdd-sampling/src/rate_sampler.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+impl RateSampler {
+    // Helper method to calculate the threshold from a rate
+    fn calculate_threshold(rate: f64) -> u64 {
+        if rate >= rate::MAX_SAMPLE_RATE {
+            MAX_UINT_64BITS
+        } else {
+            (rate * (MAX_UINT_64BITS as f64)) as u64
```
yannham, line 30: Can rate be negative here? The conversion to unsigned could then be gibberish

libdd-sampling/src/rules_sampler.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// Finds the first matching rule for a span
+    pub fn find_matching_rule<F>(&self, matcher: F) -> Option<SamplingRule>
+    where
+        F: Fn(&SamplingRule) -> bool,
+    {
+        self.inner
+            .read()
+            .unwrap()
+            .iter()
+            .find(|rule| matcher(rule))
```
yannham, line 36:
```suggestion
            .find(matcher)
```

libdd-sampling/src/sampling_rule.rs
```diff
@@ -0,0 +1,610 @@
...
+            self.match_attribute_value(&status_value, matcher)
+        })
+    }
+
+    // Helper method to match attribute values considering different value types
+    fn match_attribute_value(&self, value: &impl ValueLike, matcher: &GlobMatcher) -> Option<bool> {
+        // Floating point values are handled with special rules
+        if let Some(float_val) = value.extract_float() {
+            // Check if the float has a non-zero decimal part
+            let has_decimal = float_val != (float_val as i64) as f64;
```
yannham: Note that this can be incorrect for `float_val` outside of the `i64` range. This range is huge, so it's not going to be an issue in practice probably, but just in case (a robust way to check is `float_val.fract() == 0.0`)

libdd-sampling/src/sampling_rule.rs
yannham, line 232: I don't think you need to allow `dead_code` for a `pub enum`?

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+            let new_rules = SamplingRule::from_configs(rule_configs.to_vec());
+
+            rules_sampler.update_rules(new_rules);
+        })
+    }
+
+    /// Computes a key for service-based sampling
+    fn service_key(&self, span: &impl SpanProperties) -> String {
+        // Get service from span
+        let service = span.service().into_owned();
```
hoolioh: Do we need the allocation? Cow implements display so the format statement could be:
```
format!("service:{},env:{}", span.service(), span.env())
```

libdd-sampling/src/glob_matcher.rs
```diff
@@ -0,0 +1,217 @@
...
+        let cache_size = NonZeroUsize::new(256).unwrap();
+        GlobMatcher {
+            pattern_lower: pattern.to_lowercase(),
+            cache: Mutex::new(LruCache::new(cache_size)),
+        }
+    }
+
+    /// Returns the pattern (lowercase version)
+    pub fn pattern(&self) -> String {
+        self.pattern_lower.clone()
```
hoolioh: Do we really need to clone the string? Can we just return a reference to it and let the caller decide whether to clone?

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -0,0 +1,1458 @@
...
+        let attributes = create_attributes("some-resource", "some-env");
+
+        // Both rules should match any span since they have no criteria
+        let span = TestSpan::new("", &attributes);
+        assert!(rule.matches(&span));
+        assert!(rule_with_empty_strings.matches(&span));
+    }
+
+    #[test]
+    fn test_sampling_rule_matches() {
```
hoolioh, line 719: Assertion missing?

libdd-sampling/src/sampling_rule_config.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    /// Tags that must match (key-value pairs)
+    #[serde(default)]
+    pub tags: HashMap<String, String>,
+
+    /// Where this rule comes from (customer, dynamic, default).
+    /// Not exposed in the public `datadog-opentelemetry` API — set automatically
+    /// during conversion from the public `SamplingRuleConfig` type.
+    #[serde(default = "default_provenance")]
+    pub provenance: String,
```
yannham, line 36: Is it ok that the default value of `impl Default` and the serde default don't agree?

libdd-sampling/src/types.rs
```diff
@@ -0,0 +1,173 @@
...
+/// #[derive(Clone, PartialEq, Eq)]
+/// struct MyTraceId(u128);
+///
+/// impl TraceIdLike for MyTraceId {
+///     fn to_u128(&self) -> u128 {
+///         self.0
+///     }
+/// }
+/// ```
+pub trait TraceIdLike: PartialEq + Eq {
```
yannham: `Eq` implies `PartialEq`

```suggestion
pub trait TraceIdLike: Eq {
```

libdd-sampling/src/types.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+/// A trait for extracting typed values from attribute values.
+///
+/// Provides methods for converting attribute values to common types used in sampling logic.
+pub trait ValueLike {
+    /// Extracts a float value if the value can be represented as `f64`.
+    ///
+    /// Returns `Some(f64)` for numeric types, `None` otherwise.
+    fn extract_float(&self) -> Option<f64>;
```
yannham: Ultra nit: those function could be named `as_xxx` instead of `extract_xxx`

libdd-sampling/src/v04_span.rs
```diff
@@ -0,0 +1,556 @@
...
+    /// Builds span properties by borrowing from `span` for lifetime `'a`.
+    pub fn from_span(span: &'a Span<T>) -> Self {
+        let env = span.meta.get("env").map(|v| v.borrow());
+
+        let status_code = span
+            .metrics
+            .get("http.status_code")
+            .and_then(|f| {
+                let v = *f as u64;
+                (v > 0 && v <= u32::MAX as u64).then_some(v as u32)
```
yannham, line 121:
```suggestion
                u32::try_from(v).ok()?
```

yannham (approved):

yannham:
(PS: Given the size of the PR, I went fast on tests and benches)

hoolioh (commented):
I'd add the crate to the .codecov file.

libdd-sampling/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+lru = "0.16.3"
+libdd-trace-utils = { path = "../libdd-trace-utils", version = "3.0.1", optional = true }
+
+[features]
+v04_span = ["dep:libdd-trace-utils"]
+
+[dev-dependencies]
+criterion = "0.5"
+libdd-common = { path = "../libdd-common", version = "4.0.0", features = ["bench-utils"] }
+libdd-trace-utils = { path = "../libdd-trace-utils", version = "3.0.1" }
```
hoolioh: `libdd-trace-utils` is redundant since it's in the `dependencies` section.

libdd-sampling/src/glob_matcher.rs
bantonsson: As far as I can see there is nothing in the RFC that says that it should be ASCII. I'll save this optimization for a follow up PR.

libdd-sampling/src/glob_matcher.rs
bantonsson: I'll save this for a follow up PR.

libdd-sampling/src/glob_matcher.rs
bantonsson: Will fix this.

libdd-sampling/src/datadog_sampler.rs
bantonsson: I agree. I'll clean up the comments.

libdd-sampling/src/rate_limiter.rs
bantonsson: Will fix.

libdd-sampling/src/sampling_rule_config.rs
bantonsson: I'll consolidate them.

libdd-sampling/src/sampling_rule.rs
bantonsson: Will fix.

libdd-sampling/src/glob_matcher.rs
ekump: No problem. Can we just capture all the follow ups in a Jira ticket so we don't lose track?

libdd-sampling/src/rate_limiter.rs
bantonsson: They are not the same, and the RFC mandates tokenized buckets and not a sliding window.

libdd-sampling/src/agent_service_sampler.rs
bantonsson: This is only done when we get a new rate sent back from the agent, which is infrequent.

libdd-sampling/src/agent_service_sampler.rs
bantonsson: Nope not contended in write mode.

libdd-sampling/src/glob_matcher.rs
bantonsson: I'll follow up on this in the optimization PR

libdd-sampling/src/sampling_rule.rs
bantonsson: It's only `pub(crate)`,  so it's needed (for now).

libdd-sampling/src/dd_sampling.rs
bantonsson: I'll add a comment.

libdd-sampling/src/rules_sampler.rs
bantonsson: That would require `matcher` to be `FnMut(&SamplingRule)` which it isn't.

libdd-sampling/src/datadog_sampler.rs
bantonsson: I don't think so. It should be the same thing. Do you think we should always return a `Vec` or, what is the question really?

PR #1929, opened 2026-04-27
fix(sidecar/telemetry): retry FFI telemetry batches when session conifg not yet available

cataphract:
# What does this PR do?

When the appsec helper posts telemetry via the in-process FFI path before the PHP IPC set_session_config message is processed, the telemetry client cache would be poisoned with a Config { endpoint: None } worker. All subsequent IPC enqueue_actions calls (including AddEndpoint) would reuse this bad client and never send HTTP requests to the agent, causing the Laravel8xTests "Endpoints are sent" test to time out.

Fix: when get_telemetry_client finds session_config is None, return None instead of calling get_or_create with Config::default(). The receiver task retries the batch up to 3 times with a 1.5 s delay before dropping. In the normal case set_session_config arrives within milliseconds so the first retry succeeds and no telemetry is lost.

# How to test the change?

Tested via `./gradlew test7.4-debug --tests "com.datadog.appsec.php.integration.Laravel8xTests.Endpoints are sent" -PuseHelperRust`. The error reproduced reliably on my machine after at most 3 or 4 retries.

### DIFF
--- a/datadog-sidecar/src/service/telemetry.rs
+++ b/datadog-sidecar/src/service/telemetry.rs
@@ -13,7 +13,7 @@ use crate::primary_sidecar_identifier;
 use base64::prelude::BASE64_URL_SAFE_NO_PAD;
 use base64::Engine;
 use datadog_ipc::platform::NamedShmHandle;
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, VecDeque};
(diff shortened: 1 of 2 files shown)


bwoebi (approved):
I suppose that's a good stop-gap until we unify communication over the sidecar socket.

PR #1930, opened 2026-04-28
fix(libdd-common): crashes caused by `getenv` while retrieving AAS env vars

gleocadie:
# What does this PR do?

This PR reads `/proc/self/environ` to populate AAS metadata (from env var) instead of using `getenv`.

# Motivation

`getenv` is not thread-safe. A call can crash if another thread is calling `setenv`. This will invalidate the `environ` pointers and cause a crash.

```
0x79319d662c1d __GI_getenv (stdlib/stdlib/getenv.c:84)
0x7931996ccec6 std::sys::env::unix::getenv::{{closure}}
0x7931996ccdd5 std::env::_var_os
0x7931996d1acb std::env::_var
0x7931993daf7f <libdd_common::azure_app_services::RealEnv as libdd_common::azure_app_services::QueryEnv>::get_var
0x7931993da137 core::ops::function::FnOnce::call_once
0x7931994cbffb std::sync::poison::once::Once::call_once_force::{{closure}}
0x7931996d0557 std::sys::sync::once::futex::Once::call
0x7931994ca917 libdd_profiling::exporter::profile_exporter::ProfileExporter::new
0x79319933b201 ddog_prof_Exporter_new
0x79319b15695d Datadog::UploaderBuilder::build()
0x79319b157667 ddup_upload
0x793044b34f3a __pyx_pw_7ddtrace_8internal_7datadog_9profiling_4ddup_5_ddup_7upload(_object*, _object* const*, long, _object*)
_PyErr_Occurred (/usr/src/python/Include/internal/pycore_pyerrors.h:23)
_Py_CheckFunctionResult (/usr/src/python/Objects/call.c:29)
_PyObject_VectorcallTstate (/usr/src/python/Include/internal/pycore_call.h:93)
0x63ceb6009738 PyObject_Vectorcall.cold (/usr/src/python/Objects/call.c:325)
0x63ceb6030406 _PyEval_EvalFrameDefault.cold (/usr/src/python/Python/bytecodes.c:2719)
_PyObject_VectorcallTstate (/usr/src/python/Include/internal/pycore_call.h:93)
0x63ceb600a716 method_vectorcall.cold (/usr/src/python/Objects/classobject.c:69)
0x63ceb6009324 _PyObject_VectorcallTstate (/usr/src/python/Include/internal/pycore_call.h:93)
0x793198e47284 std::thread::_State_impl<std::thread::_Invoker<std::tuple<_PeriodicThread_do_start(periodic_thread*, bool)::{lambda()#1}> > >::_M_run()
0x793198e47f60 execute_native_thread_routine
0x79319d6b2ac3 start_thread (nptl/nptl/pthread_create.c:442)
0x79319d7448d0 __clone3
```

### DIFF
--- a/libdd-common/src/azure_app_services.rs
+++ b/libdd-common/src/azure_app_services.rs
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use regex::Regex;
+#[cfg(target_os = "linux")]
+use std::collections::HashMap;
 use std::sync::LazyLock;
 
 const WEBSITE_OWNER_NAME: &str = "WEBSITE_OWNER_NAME";
@@ -25,7 +27,8 @@ enum AzureContext {
     AzureAppService,
 }
 
-/// Snapshot of the process environment as captured at exec time.
+/// Snapshot of the AAS-relevant subset of the process environment, captured at
+/// exec time.
 ///
 /// On Linux we parse `/proc/self/environ` (a kernel-managed snapshot of the
 /// initial `envp` block) instead of calling `getenv`. This avoids the
@@ -39,11 +42,36 @@ enum AzureContext {
 /// needs — every Azure App Services / Functions variable we inspect is set by
 /// the platform before the process starts.
 ///
+/// **Only the AAS variable names listed in [`AAS_VAR_NAMES`] are retained**;
+/// all other entries (which on real AAS deployments commonly include unrelated
+/// app secrets, API keys, connection strings, etc.) are dropped at parse time
+/// so we don't keep a long-lived in-memory copy of them.
+///
 /// On non-Linux platforms we fall back to `std::env::var`. AAS deployments are
 /// almost exclusively Linux containers, and Windows' `GetEnvironmentVariableW`
 /// (which `std::env::var` uses) is itself thread-safe.
 #[cfg(target_os = "linux")]
 static PROC_ENVIRON: LazyLock<HashMap<String, String>> = LazyLock::new(read_proc_self_environ);
+
+/// AAS-related variable names this module reads. Only entries with one of
+/// these names are kept in [`PROC_ENVIRON`]; everything else from
+/// `/proc/self/environ` is dropped at parse time.
+#[cfg(target_os = "linux")]
+const AAS_VAR_NAMES: &[&str] = &[
+    WEBSITE_OWNER_NAME,
+    WEBSITE_SITE_NAME,
+    WEBSITE_RESOURCE_GROUP,
+    SITE_EXTENSION_VERSION,
+    WEBSITE_OS,
+    INSTANCE_NAME,
+    INSTANCE_ID,
+    SERVICE_CONTEXT,
+    FUNCTIONS_WORKER_RUNTIME,
+    FUNCTIONS_WORKER_RUNTIME_VERSION,
+    FUNCTIONS_EXTENSION_VERSION,
+    DD_AZURE_RESOURCE_GROUP,
+    WEBSITE_SKU,
+];
 
 #[cfg(target_os = "linux")]
 fn read_proc_self_environ() -> HashMap<String, String> {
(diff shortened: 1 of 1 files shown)


libdd-common/src/azure_app_services.rs
```diff
@@ -26,10 +26,62 @@ enum AzureContext {
     AzureAppService,
 }
 
+/// Snapshot of the process environment as captured at exec time.
+///
+/// On Linux we parse `/proc/self/environ` (a kernel-managed snapshot of the
+/// initial `envp` block) instead of calling `getenv()`. This avoids the
```
KowalskiThomas:
```suggestion
/// initial `envp` block) instead of calling `getenv`. This avoids the
```

`getenv` is the function, `getenv()` is the return value of the function, so it should be "instead of calling `getenv`" 🙂

KowalskiThomas:
@codex review (asking for a Codex review because I'm not exactly a Rust expert)

KowalskiThomas (commented):
Overall this makes sense to me, but I'll wait for the Codex review to approve.

libdd-common/src/azure_app_services.rs
gleocadie: but when you call `getenv` you add the `()`:P

libdd-common/src/azure_app_services.rs
```diff
@@ -26,10 +26,62 @@ enum AzureContext {
...
+/// (which `std::env::var` uses) is itself thread-safe.
+#[cfg(target_os = "linux")]
+static PROC_ENVIRON: LazyLock<HashMap<String, String>> = LazyLock::new(read_proc_self_environ);
+
+#[cfg(target_os = "linux")]
+fn read_proc_self_environ() -> HashMap<String, String> {
+    let bytes = match std::fs::read("/proc/self/environ") {
+        Ok(b) => b,
+        Err(_) => return HashMap::new(),
+    };
```
yannham:
```suggestion
    let Ok(bytes) = std::fs::read("/proc/self/environ") else {
        return HashMap::new();
    };
```

libdd-common/src/azure_app_services.rs
```diff
@@ -26,10 +26,62 @@ enum AzureContext {
...
+    let mut map = HashMap::new();
+    for entry in bytes.split(|&b| b == 0) {
+        if entry.is_empty() {
+            continue;
+        }
+        let s = match std::str::from_utf8(entry) {
+            Ok(s) => s,
+            Err(_) => continue,
+        };
```
yannham:
```suggestion
        let Ok(s) = std::str::from_utf8(entry) else {
            continue;
        };
```

libdd-common/src/azure_app_services.rs
KowalskiThomas: I don't agree 
* "I'm calling `getenv`" means literally "`getenv()`"
* "I'm calling `getenv()`" means "I'm calling whatever `getenv` returns when I call it"

KowalskiThomas: Just like if you have a function whose parameter is a function pointer you would pass `f` as the parameter, not `f()` (unless what you want to pass is what `f` returns...)

gleocadie: Agree to disagree

KowalskiThomas (approved):
Appreciate the change 🙇 

Take the review with a grain of salt as I said I'm really not an expert here...

libdd-common/src/azure_app_services.rs
```diff
@@ -912,11 +964,12 @@ mod tests {
     }
 
     #[test]
-    fn test_get_trimmed_env_var_empty_string() {
-        env::remove_var("TEST_VAR_NONE");
-        assert_eq!(get_trimmed_env_var!("TEST_VAR_NONE"), None);
-
-        env::set_var("TEST_VAR_EMPTY_STRING", "");
-        assert_eq!(get_trimmed_env_var!("TEST_VAR_EMPTY_STRING"), None);
+    fn test_get_trimmed_env_var_missing() {
```
yannham, line 997: I think you could test `read_proc_self_environ` by comparing it to the current environment (obtained from the Rust stdlib), but this has a risk of flakiness, since for any reason if something else grew the environment in between, they might disagree. Thought maybe that has no risk of happening in the test binary? I don't know, I'll let you decide 😛

yannham (approved):
LGTM

gyuheon0h (approved):

PR #1933, opened 2026-04-28
feat(http-client): add knob to disable connection pooling

yannham:
# What does this PR do?

This PR adds a configuration knob to disable connection pooling for the HTTP client common component. This doesn't assume that the backend has specific support for connection pooling, but allows to say "if you have connection pooling, do not use it".

# Motivation

This PR has been split off #1806, which builds an agent-level HTTP client on top of the basic HTTP client. To communicate with the agent, the current behavior of existing helpers in e.g. libdd-common is to disable connection pooling by default, because the agent has a low keep-alive timeout and the most common case is to flush data on a periodic schedule. There, connection pooling is detrimental as the connection will be invalidated as soon as it is re-used, see https://github.com/DataDog/libdatadog/blob/cff72917bfe6a4f916db03294ae1ae895f58b882/libdd-common/src/http_common.rs#L118-L128.

# Additional Notes

N/A

# How to test the change?

N/A

### DIFF
--- a/libdd-http-client/src/backend/hyper_backend.rs
+++ b/libdd-http-client/src/backend/hyper_backend.rs
@@ -149,10 +149,10 @@ impl super::Backend for HyperBackend {
         transport: TransportConfig,
         allow_connection_pooling: bool,
     ) -> Result<Self, HttpClientError> {
-        let builder = http_common::client_builder();
+        let mut builder = http_common::client_builder();
 
         if !allow_connection_pooling {
-            builder = builder.pool_max_idle_per_host(0);
+            builder.pool_max_idle_per_host(0);
         }
 
         Ok(Self {
--- a/libdd-http-client/src/backend/hyper_backend.rs
+++ b/libdd-http-client/src/backend/hyper_backend.rs
@@ -145,13 +145,12 @@ fn map_hyper_error(e: hyper_util::client::legacy::Error) -> HttpClientError {
 
 impl super::Backend for HyperBackend {
     fn new(
-        _timeout: std::time::Duration,
+        client_config: &HttpClientConfig,
         transport: TransportConfig,
-        allow_connection_pooling: bool,
     ) -> Result<Self, HttpClientError> {
         let mut builder = http_common::client_builder();
 
-        if !allow_connection_pooling {
+        if !client_config.allow_connection_pooling() {
             builder.pool_max_idle_per_host(0);
         }
 
--- a/libdd-http-client/src/backend/hyper_backend.rs
+++ b/libdd-http-client/src/backend/hyper_backend.rs
@@ -147,9 +147,18 @@ impl super::Backend for HyperBackend {
     fn new(
         _timeout: std::time::Duration,
         transport: TransportConfig,
+        allow_connection_pooling: bool,
     ) -> Result<Self, HttpClientError> {
-        let client = http_common::client_builder().build(Connector::default());
-        Ok(Self { client, transport })
+        let builder = http_common::client_builder();
+
+        if !allow_connection_pooling {
+            builder = builder.pool_max_idle_per_host(0);
+        }
+
+        Ok(Self {
+            client: builder.build(Connector::default()),
+            transport,
+        })
     }
 
     async fn send(
--- a/libdd-http-client/src/backend/mod.rs
+++ b/libdd-http-client/src/backend/mod.rs
@@ -16,7 +16,11 @@ pub(crate) mod reqwest_backend;
 /// Backend`.
 pub(crate) trait Backend: Sized {
     /// Construct a new backend with the given timeout and transport.
-    fn new(timeout: Duration, transport: config::TransportConfig) -> Result<Self, HttpClientError>;
+    fn new(
+        timeout: Duration,
+        transport: config::TransportConfig,
+        allow_connection_pooling: bool,
+    ) -> Result<Self, HttpClientError>;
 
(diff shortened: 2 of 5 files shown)


libdd-http-client/src/backend/hyper_backend.rs
```diff
@@ -147,9 +147,18 @@ impl super::Backend for HyperBackend {
...
         _timeout: std::time::Duration,
         transport: TransportConfig,
+        allow_connection_pooling: bool,
     ) -> Result<Self, HttpClientError> {
-        let client = http_common::client_builder().build(Connector::default());
-        Ok(Self { client, transport })
+        let builder = http_common::client_builder();
+
+        if !allow_connection_pooling {
+            builder = builder.pool_max_idle_per_host(0);
```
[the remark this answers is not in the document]
yannham: Ah, this made me realize `--features hyper-backend` is (surprisingly I would say?) doesn't build with hyper backend. We need `--no-default-features` as well. It feels a bit like a foot gun, but I'm not sure what we can do, because I understand why we want to keep both features possible (for `--all-features` to work at least).

libdd-http-client/src/config.rs
```diff
@@ -136,6 +155,20 @@ impl HttpClientBuilder {
         self
     }
 
+    /// Allow connection pooling. Defaults to `true`.
+    ///
+    /// Note that whether pooling is actually used depends on the HTTP backend of
+    /// [libdd_http_client], though both currently available backends (reqwest and hyper) support
```
ekump: nit: I think you can refer to the crate you're in as just `[crate]`
```suggestion
    /// [crate], though both currently available backends (reqwest and hyper) support
```

libdd-http-client/src/config.rs
```diff
@@ -136,6 +155,20 @@ impl HttpClientBuilder {
...
+    /// pooling. This setting should be understood as: if set to `true`, the default behavior of the
+    /// underlying backend will be selected, which might or might not do connection pooling by
+    /// default. If set to `false`, we guarantee no connection pooling will happen.
+    ///
+    /// This setting is used by the Agent-level HTTP client.
```
ekump: Very minor nit, and I don't have a super strong opinion - I'm not sure documentation should refer to specific downstream examples? Something more generic like `Useful when talking to a server with a short keep-alive timeout, where reused pooled connections may close mid-request.` might be clearer.

libdd-http-client/src/config.rs
```diff
@@ -136,6 +155,20 @@ impl HttpClientBuilder {
...
+    pub fn allow_connection_pooling(mut self, allow: bool) -> Self {
```
ekump, line 165: We don't need to go the extreme of testing that the backend is running with or without pooling, but can we add some more test coverage to validate that we're passing the config correctly to the backend?

libdd-http-client/src/backend/mod.rs
```diff
@@ -16,7 +16,11 @@ pub(crate) mod reqwest_backend;
 /// Backend`.
 pub(crate) trait Backend: Sized {
     /// Construct a new backend with the given timeout and transport.
-    fn new(timeout: Duration, transport: config::TransportConfig) -> Result<Self, HttpClientError>;
+    fn new(
```
ekump, line 18: minor: Not a big deal with 3 params, but should we just pass the entire config?

ekump (approved):
LGTM once we have some test coverage

libdd-http-client/src/backend/hyper_backend.rs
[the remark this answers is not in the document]
ekump: > I understand why we want to keep both features possible

Ideally, we wouldn't. But I don't think Rust handles mutually exclusive features well. Outside the scope of this PR but I think what we really need is a "feature matrix" for tests. `--all-features` is purely for tests. I'm not aware of any valid production use cases.

libdd-http-client/src/config.rs
bantonsson: 👍🏼

bantonsson (approved):

libdd-http-client/src/backend/hyper_backend.rs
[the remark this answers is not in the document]
yannham: Yeah, and beside `--all-features`, I suppose feature unification can come bite you as well. As you say I think Rust just assumes features are somehow additive in its model.

libdd-http-client/src/config.rs
yannham: Oh, I didn't know that. Thanks!

libdd-http-client/src/backend/mod.rs
yannham: Yeah, I had the same dilemma internally. I went for the same approach as before, but since config is here allocated and available, I tend to think just passing config is prob nicer...

libdd-http-client/src/config.rs
yannham: On one hand I agree with the fact that this is a dependency of the agent client, so it shouldn't care about/know about it. In general I tend to leave those kind of comments because I find them helpful answer the question "why the hell did we need to do that non-obvious thing" in the future when coming back at a feature added in a dependency because of the specifics of one or a few downstream users.

That being said, in this case it's just a function. I guess using goto references answer that question, plus it's also not totally crazy as a configuration knob even forgetting the agent, so the comment is maybe not very useful.

PR #1938, opened 2026-04-30
fix(ipc): Fallback to ftruncate if fallocate gets EPERM

bwoebi:
I have no idea how fallocate is ever returning EPERM here, but apparently it is for some users. So, ... falling back it is.

### DIFF
--- a/datadog-ipc/src/platform/mem_handle.rs
+++ b/datadog-ipc/src/platform/mem_handle.rs
@@ -96,12 +96,16 @@ where
         // Use fallocate on Linux to eagerly commit the new pages: ENOSPC at resize time is
         // recoverable; a later SIGBUS mid-execution is not.
         #[cfg(target_os = "linux")]
-        nix::fcntl::fallocate(
+        match nix::fcntl::fallocate(
             fd.as_raw_fd(),
             nix::fcntl::FallocateFlags::empty(),
             0,
             new_size,
-        )?;
+        ) {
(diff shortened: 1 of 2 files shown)


datadog-ipc/src/platform/mem_handle.rs
```diff
@@ -96,12 +96,16 @@ where
         // Use fallocate on Linux to eagerly commit the new pages: ENOSPC at resize time is
         // recoverable; a later SIGBUS mid-execution is not.
```
morrisonlevi: "Try to use" now.

datadog-ipc/src/platform/mem_handle.rs
```diff
@@ -96,12 +96,16 @@ where
...
         #[cfg(target_os = "linux")]
-        nix::fcntl::fallocate(
+        match nix::fcntl::fallocate(
             fd.as_raw_fd(),
             nix::fcntl::FallocateFlags::empty(),
             0,
             new_size,
-        )?;
+        ) {
+            Err(nix::Error::EPERM | nix::Error::ENOSYS) => nix::unistd::ftruncate(fd, new_size)?,
```
morrisonlevi: Maybe `EOPNOTSUPP`? If this uses a file-based fallback then it might conceivably return that?

morrisonlevi (approved):
BTW, we think this happens because of seccomp. Patch looks good, have some very minor suggestions.

PR #1939, opened 2026-04-30
feat!: added regex-lite feature

Aaalibaba42:
# What does this PR do?

added regex-lite feature and changed code path to use libdd-common provided regex engine, which changes based on the feature.

# Motivation

Improve binary size at the cost of performance.

# Additional Notes

There is 1 place in tests in which we use regex crate directly, it's a dev dependancy, shouldn't be a problem.

# How to test the change?

cargo metadata to make sure we don't have regex with the feature, look at binary size etc.

### DIFF
--- a/datadog-ffe/Cargo.toml
+++ b/datadog-ffe/Cargo.toml
@@ -37,4 +37,3 @@ path = "benches/eval.rs"
 
 [features]
 pyo3 = ["dep:pyo3"]
-regex-lite = ["libdd-common/regex-lite"]
--- a/datadog-ffe/Cargo.toml
+++ b/datadog-ffe/Cargo.toml
@@ -18,7 +18,7 @@ chrono = { version = "0.4.38", default-features = false, features = ["now", "ser
 derive_more = { version = "2.0.0", default-features = false, features = ["from", "into"] }
 log = { version = "0.4.21", default-features = false, features = ["kv", "kv_serde"] }
 md5 = { version = "0.7.0", default-features = false }
-libdd-common = { version = "4.0.0", path = "../libdd-common", default-features = false }
+libdd-common = { version = "4.0.0", path = "../libdd-common", default-features = false, features = ["ffe"] }
 semver = "1.0"
 serde-bool = { version = "0.1.3", default-features = false }
 serde_with = { version = "3.11.0", default-features = false, features = ["base64", "hex", "macros"] }
--- a/datadog-ffe/Cargo.toml
+++ b/datadog-ffe/Cargo.toml
@@ -18,7 +18,7 @@ chrono = { version = "0.4.38", default-features = false, features = ["now", "ser
 derive_more = { version = "2.0.0", default-features = false, features = ["from", "into"] }
 log = { version = "0.4.21", default-features = false, features = ["kv", "kv_serde"] }
 md5 = { version = "0.7.0", default-features = false }
-libdd-common = { version = "4.0.0", path = "../libdd-common", default-features = false, features = ["ffe"] }
+libdd-common = { version = "4.0.0", path = "../libdd-common", default-features = false, features = ["require-regex-full"] }
 semver = "1.0"
 serde-bool = { version = "0.1.3", default-features = false }
 serde_with = { version = "3.11.0", default-features = false, features = ["base64", "hex", "macros"] }
--- a/datadog-ffe/Cargo.toml
+++ b/datadog-ffe/Cargo.toml
@@ -18,7 +18,7 @@ chrono = { version = "0.4.38", default-features = false, features = ["now", "ser
 derive_more = { version = "2.0.0", default-features = false, features = ["from", "into"] }
 log = { version = "0.4.21", default-features = false, features = ["kv", "kv_serde"] }
 md5 = { version = "0.7.0", default-features = false }
-regex = "1.10.4"
+libdd-common = { version = "4.0.0", path = "../libdd-common", default-features = false }
 semver = "1.0"
 serde-bool = { version = "0.1.3", default-features = false }
 serde_with = { version = "3.11.0", default-features = false, features = ["base64", "hex", "macros"] }
@@ -37,3 +37,4 @@ path = "benches/eval.rs"
 
 [features]
 pyo3 = ["dep:pyo3"]
+regex-lite = ["libdd-common/regex-lite"]
--- a/datadog-ffe/src/rules_based/ufc/models.rs
+++ b/datadog-ffe/src/rules_based/ufc/models.rs
@@ -3,7 +3,7 @@
 
 use std::{collections::HashMap, sync::Arc};
 
-use regex::Regex;
+use libdd_common::regex_engine::Regex;
 use serde::{Deserialize, Serialize};
 
 use crate::rules_based::{EvaluationError, FlagType, Str, Timestamp};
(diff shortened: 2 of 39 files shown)


libdd-data-pipeline/src/telemetry/mod.rs
```diff
@@ -328,6 +328,9 @@ mod tests {
     use httpmock::Method::POST;
     use httpmock::MockServer;
     use libdd_capabilities::HttpError;
+    // Use `regex::Regex` directly here because `httpmock`'s `body_matches`
+    // requires `Into<HttpMockRegex>`, which is only implemented for
+    // `regex::Regex`, not `regex_lite::Regex`.
```
yannham, line 335: I suppose the comment should be attached to the actual `use regex::Regex`?

libdd-trace-obfuscation/src/replacer.rs
```diff
@@ -333,13 +335,13 @@ mod tests {
...
             name: "http.url".to_string(),
-            re: regex::Regex::new("(token/)([^/]*)").unwrap(),
+            re: libdd_common::regex_engine::Regex::new("(token/)([^/]*)").unwrap(),
             repl: "${1}?".to_string(),
             no_expansion: false,
         };
         let rule2 = replacer::ReplaceRule {
             name: "http.url".to_string(),
-            re: regex::Regex::new("(broken/)([^/]*)").unwrap(),
+            re: libdd_common::regex_engine::Regex::new("(broken/)([^/]*)").unwrap(),
```
yannham: Nit: it's used a lot in this module, so I suppose importing `regex_engine` or even just `Regex` would make sense.

yannham (approved):

datadog-ffe/src/rules_based/ufc/models.rs
```diff
@@ -3,7 +3,7 @@
 
 use std::{collections::HashMap, sync::Arc};
 
-use regex::Regex;
+use libdd_common::regex_engine::Regex;
```
dd-oleksii, line 6: I'm going to ask to exclude FFE from this change for two reasons:
1. Our product uses user-provided regexes, so switching engine changes the behavior of _their_ app. i.e., it's a breaking change for our product.
2. Regex matching is invoked in perf-sensitive sections. With regex-lite optimizing performance away this seems like a bad idea for FFE.

dd-oleksii (changes requested):

dd-oleksii (dismissed):
see my previous review

datadog-ffe/src/rules_based/ufc/models.rs
Aaalibaba42: Yep sorry I asked again for a review too early after doing a fix that broke things so I rollbacked.

I don't think we want for the codebase as a whole to have a single crate use another codepath, especially since if we were to not touch `datadog-ffe`'s code at all and with the `regex-lite` feature enabled, we would just be bundling both `regex` and `regex-lite` crates in the final binary, which is detrimental for everyone.

What I can propose instead is to have a feature flag for `ffe` as well as `regex-lite` for `libdd-common`, and have the full regex engine chosen for everyone if `datadog-ffe` is compiled (feature flag propagation in libdd-common).

Is that acceptable to you ?

commit proposal: dce8d30f37114f449b51471e5573da91b55e83ae

datadog-ffe/Cargo.toml
```diff
@@ -37,3 +37,4 @@ path = "benches/eval.rs"
 
 [features]
 pyo3 = ["dep:pyo3"]
+regex-lite = ["libdd-common/regex-lite"]
```
dd-oleksii: major: this feature shouldn't be here given it doesn't work

datadog-ffe/Cargo.toml
```diff
@@ -18,7 +18,7 @@ chrono = { version = "0.4.38", default-features = false, features = ["now", "ser
 derive_more = { version = "2.0.0", default-features = false, features = ["from", "into"] }
 log = { version = "0.4.21", default-features = false, features = ["kv", "kv_serde"] }
 md5 = { version = "0.7.0", default-features = false }
-regex = "1.10.4"
+libdd-common = { version = "4.0.0", path = "../libdd-common", default-features = false, features = ["ffe"] }
```
dd-oleksii: minor: any reason we need to jump through the hoops to import `Regex` from `libdd-common` when we can use `regex` crate directly? Mostly worrying for added complexity and added potential failure if `libdd-common` fails to provide proper regex implementation (which my team won't see coming as we won't be requested to review changes).

`ffe` feature on `libdd-common` canceling `regex-lite` is also somewhat smelly. First, it's an unexpected interaction of features. Second, lower-level components should not know about components that build on top. If you do want to route through `libdd-common`, perhaps `regex-full` is a better feature name.

libdd-common/Cargo.toml
```diff
@@ -98,6 +99,8 @@ use_webpki_roots = ["hyper-rustls/webpki-roots"]
 # Enable this feature to enable stubbing of cgroup
 # php directly import this crate and uses functions gated by this feature for their test
 cgroup_testing = []
+# Use regex-lite instead of regex for smaller binary size / faster compile times
+regex-lite = ["dep:regex-lite"]
```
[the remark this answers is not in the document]
dd-oleksii: I believe this comment is valid?

datadog-ffe/Cargo.toml
dd-oleksii: Just noticed your [comment](https://github.com/DataDog/libdatadog/pull/1939#discussion_r3187442275). Routing through `libdd-common` makes sense, though still suggest renaming the feature flag to `regex-full`.

libdd-common/Cargo.toml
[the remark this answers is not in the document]
Aaalibaba42: It's valid but it misses the point, we want to reduce binary size of the end artifact, and dead code is discarded by the linker. Build and compile times aren't much of a concern at this point (The comment quoted has also been changed because it didn't reflect what we wanted). So yeah we still pull it when building, we still build it before compiling our artifacts, but with the way we have it setup in libdd-common, all regex code becomes dead code and gets discarded at link time.

dd-oleksii (approved):

libdd-log/src/logger.rs
```diff
@@ -4,10 +4,10 @@
 use crate::writers::{FileWriter, StdWriter};
```
VianneyRuhlmann, line 4: Are these changes linked to the regex changes ?

libdd-log/Cargo.toml
```diff
@@ -13,7 +13,7 @@ bench = false
 
 [dependencies]
 tracing = { version = "0.1", default-features = false, features = ["std"] }
-tracing-subscriber = { version = "0.3.22", default-features = false, features = ["json", "env-filter"] }
+tracing-subscriber = { version = "0.3.22", default-features = false, features = ["json"] }
```
VianneyRuhlmann, line 16: Is this linked to the regex changes ?

VianneyRuhlmann (approved):
few nits but LGTM

libdd-log/Cargo.toml
Aaalibaba42: Yes, tracing subscriber used regex automata which depended on regex

libdd-log/src/logger.rs
Aaalibaba42: Yes we needed to move out of EnvFilter to remove `regex` from transitive dependancies.

PR #1943, opened 2026-05-04
fix(crypto): gate libdd-common TLS features in remaining internal crates + add CI guard

duncanista:
## What does this PR do?

Follows up #1872 by applying the same `default-features = false` + explicit `https`/`fips` feature forwarding pattern to the four internal crates that #1872 missed, and adds a CI guard so this class of regression cannot reland silently.

### Cargo.toml changes

| Crate | Before | After |
|---|---|---|
| `libdd-trace-stats` | `libdd-capabilities-impl` ungated; no `fips` feature | `default-features = false` + `https`/`fips` features forwarding to `libdd-common` and `libdd-capabilities-impl` |
| `libdd-data-pipeline` | target-cfg `libdd-capabilities-impl` ungated; no `fips` feature | `default-features = false` + new `fips` feature mirroring the existing `https` forwarding (`libdd-common`, `libdd-capabilities-impl`, `libdd-telemetry?`, `libdd-trace-stats`, `libdd-trace-utils`, `libdd-dogstatsd-client`) |
| `libdd-dogstatsd-client` | no `fips` feature | adds `fips = ["libdd-common/fips"]` |
| `libdd-telemetry` | no `fips` feature | adds `fips = ["libdd-common/fips"]` |

### CI guard

Adds `scripts/check_crypto_providers.sh` plus `.github/workflows/check-crypto-providers.yml`. The script asserts that no `libdd-*` crate ends up with both `ring` and `aws-lc-rs` in its runtime dependency graph, in *any* of these build modes:

- default features (whatever `cargo` picks)
- `--no-default-features --features https` (when the crate exposes an `https` feature)
- `--no-default-features --features fips` (when the crate exposes a `fips` feature)

It iterates **all** `libdd-*` crates so a regression introduced via a transitive edge — even on a crate that doesn't itself expose `https`/`fips` — gets flagged. Resolution is per-crate (via `--manifest-path`) so workspace-level feature unification from other members can't mask a violation, and dev-deps are excluded so test-only graphs don't produce false positives. Locally on this branch: `crypto provider check passed for 32 crate(s)`.

## Motivation

Without these gates, downstream consumers that build with `--no-default-features --features fips` still get `ring` in the dep graph through:

```
ring → rustls → hyper-rustls → libdd-common (https feature)
                                ↑
                                activated transitively because:
                                libdd-trace-stats → libdd-capabilities-impl (default-features = true → https)
                                libdd-data-pipeline → libdd-capabilities-impl (default-features = true → https)
```

This is the same shape #1872 fixed for `libdd-trace-obfuscation`, `libdd-capabilities-impl`, and `libdd-trace-utils`; these four crates were left out. It also affects non-FIPS builds: any consumer that pulls `rustls/aws-lc-rs` directly (e.g. `datadog-lambda-extension`) ends up with **both** `ring` and `aws-lc-rs` linked into the binary, bloating release artifacts and contradicting the gating intent of #1816.

## How to test the change?

Default builds are unchanged (the `default = ["https"]` feature on each crate preserves the existing `ring` behavior).

Verify ring is absent from FIPS feature builds of the affected crates:

```
cargo tree --manifest-path libdd-trace-stats/Cargo.toml --no-default-features --features fips --edges no-dev -i ring
# Expected: warning: nothing to print.

cargo tree --manifest-path libdd-data-pipeline/Cargo.toml --no-default-features --features fips --edges no-dev -i ring
# Expected: warning: nothing to print.
```

Run the full guard locally:

```
./scripts/check_crypto_providers.sh
# Expected: crypto provider check passed for 32 crate(s)
```

Sanity-check the guard catches regressions — temporarily drop `default-features = false` from `libdd-trace-stats`'s `libdd-capabilities-impl` edge and re-run the script. It correctly fails with `libdd-trace-stats (--features fips) pulls both ring and aws-lc-rs` and `libdd-data-pipeline (--features fips) pulls both ring and aws-lc-rs`.

### DIFF
--- a/.github/workflows/check-crypto-providers.yml
+++ b/.github/workflows/check-crypto-providers.yml
@@ -11,7 +11,7 @@ env:
   CARGO_INCREMENTAL: 0
 jobs:
   check-crypto-providers:
-    name: "ring + aws-lc-rs are not both pulled into any libdd-* crate"
+    name: "ring + aws-lc-rs are not both pulled into any workspace crate"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout sources
--- a/.github/workflows/check-crypto-providers.yml
+++ b/.github/workflows/check-crypto-providers.yml
@@ -1,0 +1,27 @@
+name: 'Check crypto providers'
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - mq-working-branch-*
+env:
+  rust_version: "1.84.1"
+  CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
+jobs:
+  check-crypto-providers:
+    name: "ring + aws-lc-rs are not both pulled into any libdd-* crate"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
+      - name: Install ${{ env.rust_version }} toolchain
+        run: rustup set profile minimal && rustup install ${{ env.rust_version }} && rustup default ${{ env.rust_version }}
+      - name: Cache [rust]
+        uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # 2.8.1
+        with:
+          cache-targets: true # cache build artifacts
+          cache-bin: true # cache the ~/.cargo/bin directory
+      - name: Run crypto provider check
+        run: ./scripts/check_crypto_providers.sh
--- a/scripts/check_crypto_providers.sh
+++ b/scripts/check_crypto_providers.sh
@@ -40,16 +40,32 @@ crate_has_feature() {
 # 0 if `package` is in the runtime dep graph of `manifest`. Resolved against
 # the crate's manifest in isolation, with dev-deps excluded.
 #
-# `cargo tree -i` exits 0 with a "nothing to print" warning when the package
-# exists in the workspace but is NOT pulled by the target crate, and exits
-# non-zero only when the package id is unknown entirely. Match the package
-# heading line directly instead of relying on the exit code alone.
+# `cargo tree -i` has three observed outcomes:
+#   - exit 0, output starts with "<pkg> v..."         → pulled
+#   - exit 0, output contains "nothing to print"      → not pulled (in workspace
+#                                                       but not in this graph)
+#   - exit non-zero, "did not match any packages"     → not pulled (not in
+#                                                       workspace at all)
+# Anything else (transient registry / git failures, manifest errors, etc.)
+# is surfaced as a hard error so the guard cannot silently pass when
+# dependency resolution itself broke.
 pulls() {
     local manifest="$1" pkg="$2"
     shift 2
-    local output
-    output=$(cargo tree --manifest-path "$manifest" --edges no-dev "$@" -i "$pkg" 2>&1) || return 1
-    [[ "$output" =~ ^"$pkg"" v" ]]
+    local output rc
+    output=$(cargo tree --manifest-path "$manifest" --edges no-dev "$@" -i "$pkg" 2>&1) && rc=0 || rc=$?
+
+    if [[ "$output" =~ ^"$pkg"" v" ]]; then
+        return 0
+    fi
+    if [[ "$output" == *"nothing to print"* ]] || \
+       [[ "$output" == *"did not match any packages"* ]]; then
+        return 1
+    fi
+
+    echo "ERROR: cargo tree failed for $(basename "$(dirname "$manifest")") -i $pkg (exit $rc):" >&2
+    echo "$output" | sed 's/^/  /' >&2
+    exit 2
 }
 
 tree_for() {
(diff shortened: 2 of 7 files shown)


.github/workflows/check-crypto-providers.yml
```diff
@@ -0,0 +1,27 @@
+name: 'Check crypto providers'
+on:
+  pull_request:
+  push:
+    branches:
+      - main
```
ekump, line 6: If we're running this on PRs and MQ branches, do we need to run it on main too?

scripts/check_crypto_providers.sh
```diff
@@ -0,0 +1,127 @@
...
+        tree_for "$manifest" "aws-lc-rs" "$@"
+        return 1
+    fi
+    return 0
+}
+
+errors=0
+checked=0
+
+for manifest in libdd-*/Cargo.toml; do
```
ekump: Do you want to check more than `libdd-` prefixed crates?

duncanista: Good call — broadened in 0637cf8. The script now enumerates every workspace member via `cargo metadata --no-deps` instead of globbing `libdd-*/Cargo.toml`. That picks up the `datadog-*` / `datadog-sidecar` / `datadog-live-debugger` / `datadog-remote-config` / `datadog-ipc` crates (and any future addition) automatically, since they all depend on `libdd-common` transitively and could re-introduce the same `ring + aws-lc-rs` co-existence we're guarding against. Total crates checked went from 32 to 52.

.github/workflows/check-crypto-providers.yml
duncanista: Yep — `main` is already in the `push: branches:` list on line 6 itself. Kept it consistent with how `test.yml`, `lint.yml`, `fuzz.yml`, `coverage.yml`, etc. all trigger: `pull_request` + `push: branches: [main, mq-working-branch-*]`. Running on the post-merge `main` push gives a baseline status on the default branch and catches anything that slips through (rebase races, MQ skips, etc.) — small CI cost vs. the convention break of opting out only here.

scripts/check_crypto_providers.sh
```diff
@@ -0,0 +1,135 @@
...
+# crate_has_feature <Cargo.toml> <feature_name>
+# 0 if the [features] table declares the named feature, 1 otherwise.
+crate_has_feature() {
+    awk -v feat="$2" '
+        /^\[features\]/ { in_features = 1; next }
+        /^\[/           { in_features = 0 }
+        in_features && $1 == feat && $2 == "=" { found = 1; exit }
+        END { exit !found }
+    ' "$1"
+}
```
hoolioh:
```suggestion
# crate_has_feature <pkg_json> <feature_name>
# 0 if the crate declares the named feature, 1 otherwise.
crate_has_feature() {
    printf '%s' "$1" | jq -e --arg feat "$2" '.features | has($feat)' > /dev/null 2>&1
}

```

scripts/check_crypto_providers.sh
```diff
@@ -0,0 +1,135 @@
...
+    if crate_has_feature "$manifest" "https"; then
+        check "$manifest" "--features https" --no-default-features --features https \
+            || errors=$((errors + 1))
+    fi
+
+    if crate_has_feature "$manifest" "fips"; then
+        check "$manifest" "--features fips" --no-default-features --features fips \
+            || errors=$((errors + 1))
+    fi
+done
```
hoolioh, line 119:
```suggestion
METADATA=$(cargo metadata --no-deps --format-version 1)

mapfile -t PACKAGES < <(printf '%s' "$METADATA" | jq -c \
    '.packages[] | {manifest: .manifest_path, features: .features}')

for pkg in "${PACKAGES[@]}"; do
    manifest=$(printf '%s' "$pkg" | jq -r '.manifest')
    checked=$((checked + 1))

    check "$manifest" "default" || errors=$((errors + 1))

    if crate_has_feature "$pkg" "https"; then
        check "$manifest" "--features https" --no-default-features --features https \
            || errors=$((errors + 1))
    fi

    if crate_has_feature "$pkg" "fips"; then
        check "$manifest" "--features fips" --no-default-features --features fips \
            || errors=$((errors + 1))
    fi
done

```

hoolioh (changes requested):
The approach looks good to me, I've been thinking of using the crates reporter action but since we need to compile I think is not worth it a the moment.
One thing I'd suggest is to replace using the regexes in favour of using cargo metadata to extract the features. I left a suggestion about how to do it, LMKWYT.

scripts/check_crypto_providers.sh
duncanista: Applied as-is in 5e5fb07.

scripts/check_crypto_providers.sh
duncanista: Applied as-is in 5e5fb07. Switched the main loop to `mapfile -t PACKAGES < <(... jq -c '.packages[] | {manifest, features}')` and the per-iteration manifest extract / `crate_has_feature` calls match your suggestion exactly. Source of truth is now one structured `cargo metadata` blob instead of a metadata call + per-crate Cargo.toml awk. Locally green at 52 crates.

hoolioh (approved):

PR #1944, opened 2026-05-04
feat: use ip quantization when aggregating peer tags for trace stats

duncanpharvey:
# What does this PR do?

Quantize IP addresses for peer tags when aggregating trace stats.

# Motivation

Needed for agent side stats computation in the Serverless Compatibility Layer (used as the agent in Azure Functions) in https://github.com/DataDog/serverless-components/pull/124.

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-trace-stats/src/span_concentrator/aggregation.rs
+++ b/libdd-trace-stats/src/span_concentrator/aggregation.rs
@@ -6,6 +6,7 @@
 //! span.
 
 use hashbrown::HashMap;
+use libdd_trace_obfuscation::ip_address::quantize_peer_ip_addresses;
 use libdd_trace_protobuf::pb;
 use libdd_trace_utils::span::SpanText;
 use std::borrow::Borrow;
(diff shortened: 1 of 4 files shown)


duncanpharvey:
Being handled as part of https://github.com/DataDog/libdatadog/pull/1819

Eldolfin:
My initial plan was to merge this change with my css obfuscation change to put it behind a feature flag but in the end we decided the change is small enough to not break anything (the agent's re-obfuscating peer tags should not change them)

Eldolfin (approved):

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -961,6 +965,80 @@ mod tests {
...
+        let ipv6_keys = vec!["peer.hostname".to_string()];
+        let key = BorrowedAggregationKey::from_span(&span_ipv6, &ipv6_keys);
+        let owned = OwnedAggregationKey::from(&key);
+        assert_eq!(
+            owned.peer_tags,
+            vec![(
+                "peer.hostname".to_string(),
+                "blocked-ip-address".to_string()
+            )]
+        );
```
VianneyRuhlmann, line 1021: very nit: but this test could be a bit more meaningful if you checked that both spans are aggregated in the same bucket (the correct handling of ipv4, ipv6 and list of ips is already tested in the function unit tests)

VianneyRuhlmann (approved):
LGTM, thanks for doing it

PR #1945, opened 2026-05-05
refactor(remote-config)!: DI via RemoteConfigParsedData trait + ParserRegistry

iunanua:
# What does this PR do?

The `datadog-remote-config` crate previously owned the full deserialization pipeline for every product via a closed `RemoteConfigData` enum. Adding a new product required editing the RC crate itself, importing product types, and adding enum variants — creating compile-time coupling to `datadog-live-debugger` and `datadog-ffe` via Cargo optional features.

This PR flips the dependency: RC defines a typed `RemoteConfigContent` trait (with a `PRODUCT` const and a `parse` method), an opaque `RemoteConfigParsedData` trait-object boundary, and a `ParserRegistry` to compose products. Product crates implement `RemoteConfigContent` on their own types; a blanket `impl<T: RemoteConfigContent> RemoteConfigParsedData for T` lets the registry hold them as `Box<dyn RemoteConfigParsedData>` without knowing the concrete type.

## Follow-up changes on this branch

- **Sidecar LiveDebugger parser registration.** Pre-refactor, `datadog-sidecar` enabled the `live-debugger` Cargo feature on `datadog-remote-config`, which made the closed-enum parser handle LiveDebugger payloads. After the DI refactor that feature is gone, so `RemoteConfigManager::new()` now explicitly registers the `LiveDebuggingData` parser on top of `default_registry()` to preserve pre-refactor sidecar behavior. Added a regression test (`test_live_debugger_config_parsed`) that drives a `SERVICE_CONFIGURATION` payload through the SHM round trip and asserts the downcast to `LiveDebuggingData` succeeds.
- **`set_extra_services` on the fetchers.** Runtime-discovered extra services live on `ConfigClientState` (the per-client mutable state), not on `Target`. New API: `ConfigClientState::set_extra_services(Vec<String>)`, `SingleFetcher::set_extra_services`, `SingleChangesFetcher::set_extra_services`. `fetch_once` reads from `opaque_state` and forwards via `ClientTracer.extra_services` exactly as before. Replace-semantics: each set fully overrides the previous list. This was kept off `Target` because `Target` is `Hash + Eq + Clone` (used as a map key in `MultiTargetFetcher`) — putting growing dynamic state on it would force costly map rekeying or break the `Hash`/`Eq` contract.

## Adding a new RC product going forward

1. **In `datadog-remote-config`** (still required): add a variant to `RemoteConfigProduct` and its `Display`/`try_parse` arms in `path.rs`. The product taxonomy is still a closed enum.
2. **In the product crate**: implement `RemoteConfigContent` on the product's config type:

   ```rust
   impl RemoteConfigContent for MyProductConfig {
       const PRODUCT: RemoteConfigProduct = RemoteConfigProduct::MyProduct;
       fn parse(data: &[u8]) -> anyhow::Result<Self> { /* … */ }
   }
   ```
3. **At construction time**: chain `.with::<MyProductConfig>()` onto a `ParserRegistry` (typically `default_registry()`).

Only the parsing is fully extensible after this PR; the product enumeration is still closed. Making `RemoteConfigProduct` open (e.g. `Cow<'static, str>` or a sealed trait) would close that gap and is a reasonable follow-up.

## Motivation

Drives `datadog-remote-config`'s coupling to product crates from compile-time (Cargo features + closed enum) to runtime (registry injection), so consumers compose the products they care about without forcing the RC crate to know about every product.

## Note
Check [this](https://github.com/DataDog/libdatadog/pull/1958) other PR with a different approach


datadog-ffe/Cargo.toml
```diff
@@ -11,6 +11,8 @@ publish = false
 bench = false
 
 [dependencies]
+anyhow = { version = "1.0" }
+datadog-remote-config = { path = "../datadog-remote-config", default-features = false }
```
dd-oleksii: minor: `anyhow` looks like a public dependency of `datadog-remote-config`, so perhaps best to re-export it from there

datadog-ffe/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 
 mod flag_type;
 
+pub mod remote_config;
```
dd-oleksii: minor: doesn't need to be public
```suggestion
mod remote_config;
```

datadog-ffe/Cargo.toml
dd-oleksii: major: we're going to use `datadog-ffe` in setups without the agent / remote config in wasm. Could you guard this dependency under an optional feature (enabled by default)?

iunanua: done

iunanua: done.
One detail: `anyhow!`, `bail!`, `ensure!` macros won't work through re-export

libdd-tracer-flare/src/lib.rs
hoolioh:
```suggestion
    /// Handle the `RemoteConfigParsedData` and return the action the tracer flare
```

hoolioh:
```suggestion
    /// * `data` - RemoteConfigParsedData.
```

datadog-remote-config/src/parse.rs
```diff
@@ -3,89 +3,217 @@
...
+        .with::<DynamicConfigFile>()
+}
+
+// ── RemoteConfigValue ─────────────────────────────────────────────────────────
+
 pub struct RemoteConfigValue {
     pub source: RemoteConfigSource,
-    pub data: RemoteConfigData,
+    pub product: RemoteConfigProduct,
+    pub data: Option<Box<dyn RemoteConfigParsedData>>,
```
bwoebi: Suggestion: newtype `Box<dyn RemoteConfigParsedData>`, e.g.

`pub struct RemoteConfigParsed(Box<dyn RemoteConfigParsedData>)` and add an `impl RemoteConfigParsed { pub fn as<T: RemoteConfigParsedData>(&self): T { self.as_any().downcast_ref::<T>().expect("Invalid type for RemoteConfigParsedData") } }` or something.

datadog-remote-config/src/parse.rs
```diff
@@ -3,89 +3,231 @@
...
+impl<T: RemoteConfigContent> RemoteConfigParsedData for T {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+}
+
+/// Error returned by [`RemoteConfigContent::parse`].
+///
+/// JSON failures convert via `?`; product-specific failures with their own
+/// error types should box through [`ParseError::Custom`] rather than
```
hoolioh, line 48: I'd avoid to use anyhow so we don't force a dependency in our consumers.

datadog-ffe/Cargo.toml
iunanua: After a chat with @hoolioh, we have decided to change the `RemoteConfigContent::parse` signature and thus remove `anyhow` from the return type:
```
fn parse(data: &[u8]) -> std::result::Result<Self, ParseError>
```
wdyt?

datadog-remote-config/src/parse.rs
```diff
@@ -3,89 +3,231 @@
...
+    /// a process where a collision is a programmer error worth surfacing immediately.
+    #[must_use]
+    pub fn with<T: RemoteConfigContent>(mut self) -> Self {
+        let parser: ProductParser = Box::new(|data: &[u8]| {
+            let parsed = T::parse(data)?;
+            Ok(Box::new(parsed) as Box<dyn RemoteConfigParsedData>)
+        });
+        #[allow(clippy::expect_used)]
+        self.register(T::PRODUCT, parser)
+            .expect("ParserRegistry::with: parser already registered");
```
hoolioh: Since we're refactoring the crate it would be good to get rid of the expect/unwrap statements.

iunanua: removed! Now callers have to decide what to do with the Error

datadog-remote-config/src/parse.rs
iunanua: done! `as` is reserved so I chose `downcast`...

bwoebi (approved):
Looks good to me now.
I don't love it, but necessary.

hoolioh (approved):
LGTM. There is still work to do by reducing the API visibility and removing anyhow errors from public APIs but as discussed internally it can be done in subsequent PRs

iunanua:
> Looks good to me now. I don't love it, but necessary.

@bwoebi would you prefer [this other](https://github.com/DataDog/libdatadog/pull/1958) approach?

bwoebi:
@iunanua no - I prefer the way it _currently_ is. But I understand it's necessary to change, so it's allright.

bantonsson (approved):

datadog-remote-config/src/parse.rs
```diff
@@ -3,89 +3,247 @@
...
+/// [`ParserRegistry::with`] on the returned registry.
+pub fn default_registry() -> ParserRegistry {
+    fn build() -> std::result::Result<ParserRegistry, AlreadyRegistered> {
+        ParserRegistry::new()
+            .with::<AgentConfigFile>()?
+            .with::<AgentTaskFile>()?
+            .with::<DynamicConfigFile>()
+    }
+    #[allow(clippy::expect_used)]
+    build().expect("default_registry: internal products are distinct by construction")
```
ekump, line 182: Instead of panicking can we return an error instead?

datadog-sidecar/src/shm_remote_config.rs
```diff
@@ -576,8 +581,20 @@ pub enum RemoteConfigUpdate {
 
 impl RemoteConfigManager {
     pub fn new(invariants: ConfigInvariants) -> RemoteConfigManager {
+        #[allow(clippy::expect_used)]
+        let registry = default_registry()
+            .with::<LiveDebuggingData>()
+            .expect("RemoteConfigManager::new: LiveDebugger is distinct from default products");
```
ekump, line 587: same here, we should really be returning an error instead of panicking.

PR #1946, opened 2026-05-05
refactor(trace-utils): replace use_v05_format bool and remove infallible expect

anais-raison:
# What does this PR do?        

Replaces the `use_v05_format`: bool parameter in `collect_trace_chunks` with the existing TraceEncoding enum, and removes `#[allow(clippy::expect_used)] + .expect("infallible...")` in the v04 msgpack encoder.

# Motivation                                              

Follow-up to [#1896](https://github.com/DataDog/libdatadog/pull/1896), addressing review comments:                                                                                                                                   

- Passing a bare bool to `collect_trace_chunks` gave no context at the call site — you had to look up the function signature to understand what true/false meant. Using TraceEncoding makes the intent explicit.
- The `collect_and_process_traces` match had a catch-all arm that computed a bool from the format — replaced with exhaustive V04/V05 arms.
- The v04 encoder used `.expect("infallible...")` with a clippy allow to handle an error that can never occur. Replaced with `let _ = ...`, consistent with what was already done in the v1 encoder.

# Additional Notes

⚠️  Should be merged after [#1896](https://github.com/DataDog/libdatadog/pull/1896).

# How to test the change?

Existing tests.


libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -123,8 +123,7 @@ pub fn to_vec_with_capacity<T: TraceData, S: AsRef<[Span<T>]>>(
     capacity: u32,
 ) -> Vec<u8> {
     let mut buf = ByteBuf::with_capacity(capacity as usize);
-    #[allow(clippy::expect_used)]
-    to_writer(&mut buf, traces).expect("infallible: the error is std::convert::Infallible");
+    let _ = to_writer(&mut buf, traces);
```
Aaalibaba42: I'm not a fan of silencing errors without explanation like that. Idk what others think but I'd add a comment saying it's infaillible, but maybe it would be just noise.

libdd-trace-utils/src/tracer_payload.rs
```diff
@@ -228,16 +228,13 @@ pub fn decode_to_trace_chunks(
     data: libdd_tinybytes::Bytes,
     encoding_type: TraceEncoding,
 ) -> Result<(TraceChunks<BytesData>, usize), anyhow::Error> {
-    let (data, size) = match encoding_type {
+    let (data, size) = match &encoding_type {
```
Aaalibaba42: Why ?

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
yannham: I think the solution was already provided in the v1 PR: we can write a safe, non-panicking `unwrap_infallible`. Hasn't this been implemented there? Maybe we should put that in `libdd-common` or something. But I agree, we have a nicer way to show that the error can actually never happen here.

yannham: cf https://github.com/DataDog/libdatadog/pull/1896#discussion_r3118816954

libdd-data-pipeline/src/trace_exporter/trace_serializer.rs
```diff
@@ -74,13 +74,16 @@ impl TraceSerializer {
         &self,
         traces: Vec<Vec<Span<T>>>,
     ) -> Result<tracer_payload::TraceChunks<T>, TraceExporterError> {
+        let map_err = |e: anyhow::Error| {
```
yannham: nit: if the closure doesn't capture anything, I would use a `fn` instead (still staying within this scope)

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -158,7 +157,6 @@ pub fn to_vec_with_capacity<T: TraceData, S: AsRef<[Span<T>]>>(
 /// ```
 pub fn to_encoded_byte_len<T: TraceData, S: AsRef<[Span<T>]>>(traces: &[S]) -> u32 {
     let mut counter = super::CountLength(0);
-    #[allow(clippy::expect_used)]
-    to_writer(&mut counter, traces).expect("infallible: CountLength never fails");
+    let _ = to_writer(&mut counter, traces);
```
yannham, line 168: Same about implementing and using an `unwrap_infallible` here

libdd-trace-utils/src/tracer_payload.rs
Eldolfin: TraceEncoding should probably be Copy anyway given its size.

libdd-trace-utils/src/trace_utils.rs
```diff
@@ -592,26 +592,26 @@ pub fn enrich_span_with_azure_function_metadata(span: &mut pb::Span) {
...
+            let mut v05_traces: Vec<Vec<v05::Span>> = Vec::with_capacity(traces.len());
+            for trace in traces {
+                let trace_len = trace.len();
+                let v05_trace = trace.into_iter().try_fold(
+                    Vec::with_capacity(trace_len),
+                    |mut acc, span| -> anyhow::Result<Vec<v05::Span>> {
+                        acc.push(v05::from_v04_span(span, &mut shared_dict)?);
+                        Ok(acc)
+                    },
+                )?;
```
yannham: I think this should be a collect, instead of a `try_fold`. You can collect an iterator with item of the form `Result<T,E>` directly into a `Result<Vec<T>, E>` (it just requires type annotations). This also get rids of the capacity business (iterators are smart enough to produce similar code, since `Vec::IntoIter` is a sized iterator) and of manually pushing to the accumulator.

Don't quote me on the exact code, but something like:

```suggestion
                let v05_trace = trace.into_iter().map(
                    |span| -> anyhow::Result<Vec<v05::Span>> {
                        v05::from_v04_span(span, &mut shared_dict)
                    }).collect::<Result<Vec<_>, _>>()?;
```

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
ekump: I agree, swallowing an error is something we want to avoid. Adding a helper to `libdd-common` is sensible. We have something kind of similar for [mutex locks](https://github.com/DataDog/libdatadog/blob/0f83957630ee6eb945f70f3036f0e937a862f99a/libdd-common/src/lib.rs#L80) already. 

I would suggest tweaking @yannham's suggestion. Instead of a helper function, I'd create an extension. It's a bit more ergonomic at the call site.  

```rust
/// Extension trait that extracts the value from a `Result` whose error
/// type is uninhabited. The signature constrains callers at compile time:
/// the method is only available when the error type is `Infallible`.
///
/// No panics. The compiler proves the `Err` arm unreachable from the type.
pub trait ResultInfallibleExt<T>: sealed::Sealed {
    fn unwrap_infallible(self) -> T;
}

impl<T> ResultInfallibleExt<T> for Result<T, core::convert::Infallible> {
    #[inline(always)]
    fn unwrap_infallible(self) -> T {
        match self {
            Ok(value) => value,
            Err(never) => match never {},
        }
    }
}

mod sealed {
    pub trait Sealed {}
    impl<T> Sealed for Result<T, core::convert::Infallible> {}
}
```

And then you'd use it here like:
```rust
use libdd_common::ResultInfallibleExt;
  
let bytes = to_writer(&mut buf, traces).unwrap_infallible();
```

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
ekump: I don't think that would work here. The error type of `to_writer()` is `ValueWriteError<W::Error>`. And it looks like `W = CountLength`. `CountLength` impls std::io::write, which picks up this [error impl](https://github.com/3Hren/msgpack-rust/blob/cf880019f72e35a884659a9b29d9eb89b862670b/rmp/src/encode/mod.rs#L178), which is not infallible. 

`RmpWrite` is sealed, so I don't think there is anything we can do here to make it infallible. 

[CountLength](https://github.com/DataDog/libdatadog/blob/0f83957630ee6eb945f70f3036f0e937a862f99a/libdd-trace-utils/src/msgpack_encoder/mod.rs#L10-L27) always returns `Ok()`. I think it's safe to swallow the error, since it's not currently possible. We should add a comment making this justification clear. 

The only risk I see is what happens if `CountLength` is changed and starts returning errors. I can't think of a good way to cover this beyond adding robust fuzz testing for msgpack encoding and asserting on the msgpack length is correct. It's probably something we should do anyway, but is beyond the scope of this PR.

libdd-trace-utils/src/trace_utils.rs
```diff
@@ -1132,7 +1132,7 @@ mod tests {
     fn test_collect_trace_chunks_v05() {
```
ekump, line 1128: Should we also have a test like this for V04?

anais-raison: Sure, added it.

libdd-data-pipeline/src/trace_exporter/trace_serializer.rs
anais-raison: I put the mapping back inside the match.

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -123,8 +124,11 @@ pub fn to_vec_with_capacity<T: TraceData, S: AsRef<[Span<T>]>>(
     capacity: u32,
 ) -> Vec<u8> {
     let mut buf = ByteBuf::with_capacity(capacity as usize);
-    #[allow(clippy::expect_used)]
-    to_writer(&mut buf, traces).expect("infallible: the error is std::convert::Infallible");
+    // `ByteBuf`'s `RmpWrite::Error` is `Infallible`, so `to_writer` cannot fail. The compiler
+    // proves the `Err` arm unreachable through `unwrap_infallible`.
```
yannham: Nit: the name and the documentation of `unwrap_infallible` are enough, IMHO.

```suggestion
```

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -399,7 +400,11 @@ pub fn to_vec_with_capacity<T: TraceData, S: AsRef<[Span<T>]>>(
     metadata: &TracerMetadata,
 ) -> Vec<u8> {
     let mut buf = ByteBuf::with_capacity(capacity as usize);
-    let _ = encode_payload(&mut buf, traces, metadata); // infallible: ByteBuf write never fails
+    // `ByteBuf`'s `RmpWrite::Error` is `Infallible`, so `encode_payload` cannot fail. The
+    // compiler proves the `Err` arm unreachable through `unwrap_infallible`.
```
yannham: Same nit
```suggestion
```

yannham (approved):
Left a couple doc nits but LGMT 👍

PR #1947, opened 2026-05-05
fix(libdd-trace-obfuscation): cargo clippy fix with all lints

Eldolfin:
# What does this PR do?

Applies a restrictive workspace Clippy/Rust lint config and opts in `libdd-trace-obfuscation`, fixing the resulting crate-local lints without changing the behavior.

# Motivation

Tighten lint coverage incrementally by starting with a single crate.

# Additional Notes

# How to test the change?

CI passing is sufficient validation.


libdd-trace-obfuscation/src/ip_address.rs
```diff
@@ -34,27 +34,29 @@ static PREFIX_REGEX: LazyLock<Regex> = LazyLock::new(|| {
...
 
     let quantized_values = values
         .map(|v| {
-            if let Some(quantize_string) = quantize_ip(v) {
-                should_return_new_string = true;
-                Cow::from(quantize_string)
-            } else {
-                Cow::from(v)
-            }
+            quantize_ip(v).map_or_else(
```
Eldolfin: Not sure about this lint. I personally find it less clear

libdd-trace-obfuscation/Cargo.toml
```diff
@@ -1,9 +1,11 @@
 [package]
 name = "libdd-trace-obfuscation"
-version = "2.0.0"
+version = "2.1.0"
```
Eldolfin: ?

duncanista (approved):

libdd-trace-obfuscation/Cargo.toml
ekump: Where did this come from? The current released version is 2.0.0. @iunanua any idea?

Eldolfin: Reverted. 🫣

Eldolfin: Looking closer, adding the #[must_use] to fix the [clippy::must_use_candidate](https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate) attribute is considered a breaking change from the [semver checks perspective](https://github.com/DataDog/libdatadog/actions/runs/25521931720/job/74907982040?pr=1947#step:7:65).

libdd-trace-obfuscation/benches/benchmarks/replace_trace_tags_bench.rs
```diff
@@ -21,14 +22,14 @@ fn criterion_benchmark(c: &mut Criterion) {
...
         error: 0,
         resource: "GET /some/raclette".to_string(),
         service: "django".to_string(),
         name: "django.controller".to_string(),
         span_id: 123,
-        start: 1448466874000000000,
-        trace_id: 424242,
+        start: 1_448_466_874_000_000_000,
+        trace_id: 424_242,
         meta: HashMap::from([
```
Aaalibaba42, line 32: I'm not sure this lint is good, IDs are not meant to be read 3 by 3

libdd-trace-obfuscation/src/sql.rs
```diff
@@ -123,10 +127,9 @@ fn find_quoted_string_end(bytes: &[u8], start: usize) -> Option<usize> {
...
                     continue;
-                } else {
-                    result = Some(i + 1);
-                    break;
                 }
+                result = Some(i + 1);
+                break;
             }
             i += 1;
         }
```
Aaalibaba42, line 133: This is "nice" too but I don't think it makes things more readable tbh, it surely changes nothing in the codegen so idk.

Aaalibaba42 (approved):
Overall this is great, glancing over the thing a little, I just don't like the ubiquitous `111_222_333` formatting given we are an ID heavy codebase, we rarely would have raw numbers above 999 compared to the numbers of times they would be IDs we don't want to format this way.

libdd-trace-obfuscation/src/json/scanner.rs
```diff
@@ -5,7 +5,7 @@
 
 /// Opcode returned by [`Scanner::step`] for each input char.
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
-pub(crate) enum Op {
+pub enum Op {
```
yannham, line 8: Why is this enum made public? It's changing the public API

libdd-trace-obfuscation/src/json/scanner.rs
```diff
@@ -57,7 +57,7 @@ enum State {
 
 /// A streaming JSON scanner. Feed chars one at a time via [`Scanner::step`];
 /// the returned [`Op`] describes the structural significance of each char.
-pub(crate) struct Scanner {
+pub struct Scanner {
```
yannham, line 60: Same question

libdd-trace-obfuscation/src/redis_tokenizer.rs
```diff
@@ -21,6 +21,7 @@ pub struct RedisTokenizerScanResult<'a> {
 }
 
 impl<'a> RedisTokenizer<'a> {
+    #[must_use]
```
yannham, line 24: Not sure about the must_use

yannham (commented):
I don't feel too comfortable stamping this 1k+ line change PR because it's a mix of trivial, semantic-preserving style changes (basically `clippy --fix`, as announced in the PR title) and two non trivial refactorings in the scanner and elsewhere (splitting huge functions into smaller ones). Would that be possible to split off those two refactorings first as two independent PRs (even it's just moving code around, it'll make it easier to make sure the changes are OK), and keep the final PR to be only the clippy-fix-like easy to review changes?

libdd-trace-obfuscation/src/redis_tokenizer.rs
Eldolfin: Why ? Curious because I don't see in what case we would need to create a `RedisTokenizer` without storing it (or any other obfuscator for that matter). The corresponding lint ([clippy::must_use_candidate](https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate)) makes sense for me: when functions are completely pure they must be marked as #[must_use] because not using its result is almost certainly a bug

libdd-trace-obfuscation/src/json/scanner.rs
Eldolfin: I think these are from this lint: [redundant_pub_crate](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pub_crate). Not a game changing lint, but the `(crate)` is indeed useless because this module is already private so it won't show up as public anyway.

Eldolfin:
> I don't feel too comfortable stamping this 1k+ line change PR because it's a mix of trivial, semantic-preserving style changes (basically `clippy --fix`, as announced in the PR title) and two non trivial refactorings in the scanner and elsewhere (splitting huge functions into smaller ones). Would that be possible to split off those two refactorings first as two independent PRs first (even it's just moving code around, it'll make it easier to make sure the changes are OK), and keep the final PR to be only the clippy-fix-like easy to review changes?

@yannham I agree, this could have been split into 2 PRs, I can still relatively easily split it now because this refactor is mostly contained in a single commit (bb56e9b7c0c1554e394d0fd8b4754297c861a9b4).

On the other hand this crate is so heavily tested that I personally feel confortable merging this refactor without thinking too much about it (the full [testsuite](https://github.com/DataDog/obfuscation-parity-tester/blob/69a9bfc500fbe2e18de16e8cfbbc9dc1c7f4d719/crates/obfuscation-parity-tester/testcases/testsuite.jsonl) is not in this repo, but there is 303 sql test cases and 27 json obfuscation test cases accros different configs to give an idea).

Lmk if you still think it's important do split it knowing that. Anyway, I'll try to go in smaller steps for crates I'm not as familiar with 👍🏽 


BTW: the [too_many_lines](https://rust-lang.github.io/rust-clippy/master/index.html?search=too_many_lines#too_many_lines) lint can be configured to change the default threshold (default is 100 lines). I feel like the default is already reasonable but lmk if you think otherwise !

libdd-trace-obfuscation/src/redis_tokenizer.rs
yannham: Ok, I wasn't sure it was an actual lint or just something an LLM added. It's just that I couldn't see anyone using a `new` function and not using the result. But I guess it never hurts and kinda makes sense. Sorry for the noise

Eldolfin: Actually AFAIK this code is not called by anyone yet. It's here for completion kind of, and might be used later if we decide to do trace obfuscation on the client side

yannham:
I don't want to block this PR on splitting, and there are already a bunch of approves. However it's always good to get a second pair of eyes on the code, so even when you're confident on the change, we should strive to make reviews easy (or it kinda defeats the purpose to review in the first place). Let this be a reminder for our future selfs :slightly_smiling_face:

libdd-trace-obfuscation/benches/benchmarks/replace_trace_tags_bench.rs
Eldolfin: fixed and [lint](https://rust-lang.github.io/rust-clippy/master/index.html?search=unreadable_literal#unreadable_literal) allowed

libdd-trace-obfuscation/src/sql.rs
Eldolfin: thanks for pointing that out, I allowed the lint because I find it less readable too

libdd-trace-obfuscation/src/json/scanner.rs
yannham: It's strange that the default behavior is to make them `pub`. I would have removed the `pub` as the default fix...

Eldolfin: But these symbols are still used in other modules of the crate (json_scanner::{Op, Scanner} are used in the public `json`module)., so they need to at least be pub(crate).

Now that I say that it would be clearer if `json_scanner` was a private submodule of `json`

yannham: Ok, interesting. So clippy thinks it's clearer to have a `pub item` inside a non-pub module rather than a `pub(crate)`. I personally respectfully disagree (you now need to know the visibility of the whole module chain to understand whether this symbol is actually pub or not, while `pub(crate)` tells you immediately) but clippy has more authority than me, so I give up 😛 thanks for the explanation!

yannham (approved):
Thanks for making the PR easier to review and for improving the repo in general 🙂

libdd-trace-obfuscation/src/json/scanner.rs
Aaalibaba42: > It's strange that the default behavior is to make them pub. I would have removed the pub as the default fix...

That would have changed the sementics, like:
```rs
mod m {
  mod n {
    /*pub*/ fn internal_foo() {}
  }

  fn foo() {
    n::internal_foo();
  }
}
```

Works or not depending on the `pub`, the lint is correct in that the semantically preserving change is to just remove the "crate" qualifier. That is, equivalent to ask the question "can you name the containing module", the effective visibility is clearer in this way imho.

PR #1948, opened 2026-05-06
perf: pre-compute string messagepack encoding

paullegranddc:
# Changes

* Force pre-computation of static strings encoding in messagepack.
* Bump rmp-serde dependency to latest minor version in libdatadog

# Motivation

This makes the msgpack_encoder::v04 benchmark 10% faster.

### DIFF
--- a/libdd-trace-utils/src/msgpack_encoder/v04/span.rs
+++ b/libdd-trace-utils/src/msgpack_encoder/v04/span.rs
@@ -10,8 +10,21 @@ use rmp::encode::{
 use rmp::Marker;
 use std::borrow::Borrow;
 
+const fn msp_string_encoding_len(s: &str) -> usize {
+    let length_marker_len = if s.len() < 32 {
+        1
+    } else if s.len() < 256 {
+        2
+    } else if s.len() <= (u16::MAX as usize) {
+        3
+    } else {
+        5
+    };
+    length_marker_len + s.len()
+}
+
 // Compute the encoding of a string to messagepack in a const manner
-const fn msp_const_string_encoding(s: &str) -> ([u8; 256], usize) {
+const fn msp_const_string_encoding<const ENCODING_LEN: usize>(s: &str) -> [u8; ENCODING_LEN] {
     // copy_to_slice is not const yet
     const fn write_be(value: u64, first_n_bytes: usize, s: &mut [u8], idx: &mut usize) {
         let mut i = 0;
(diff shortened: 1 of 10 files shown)


libdd-trace-utils/src/msgpack_encoder/v04/span.rs
```diff
@@ -7,8 +7,73 @@ use rmp::encode::{
...
+    }
+    if matches!(marker, Marker::Str32) {
+        write_be(len, 4, &mut storage, &mut idx);
+    }
+    let mut i = 0;
+    while i < s.len() {
+        storage[idx + i] = s.as_bytes()[i];
+        i += 1;
+    }
+    idx += s.len();
```
yannham:
```suggestion
    while i < len {
        storage[idx + i] = s.as_bytes()[i];
        i += 1;
    }
    idx += len;
```

libdd-trace-utils/src/msgpack_encoder/v04/span.rs
```diff
@@ -7,8 +7,73 @@ use rmp::encode::{
     write_bin, write_bool, write_f64, write_i64, write_sint, write_str, write_u32, write_u64,
     write_u8, RmpWrite, ValueWriteError,
 };
+use rmp::Marker;
 use std::borrow::Borrow;
 
+// Compute the encoding of a string to messagepack in a const manner
+const fn msp_const_string_encoding(s: &str) -> ([u8; 256], usize) {
```
yannham: It's probably useless given the size involved, but as a thought experiment, I wonder if you could use const generics to produce an array of the precise size required. I guess the only thing needed is to have a const-fn or macro to get the length of a static string at compile-time, and use that as a const integer parameter.

yannham (approved):

libdd-trace-utils/src/msgpack_encoder/v04/span.rs
paullegranddc: It does work 😄

libdd-trace-utils/src/msgpack_encoder/v04/span.rs
paullegranddc: That makes the code harder because `idx` is a usize.
I could make len a usize, but I thought it'd be easier to reason with fixed size on 32 bits arch like wasm

bantonsson (approved):

libdd-trace-utils/src/msgpack_encoder/v04/span.rs
yannham: Ah, fair enough!

PR #1951, opened 2026-05-06
feat(ffe): support serial id

dd-oleksii:
# What does this PR do?

Support serial id propagation from UFC to assignment. Ported from our internal repo.


sameerank (approved):
🚢

PR #1952, opened 2026-05-06
feat: Add FFI for trace exporter

lloeki:
# What does this PR do?

Add span-building and trace-sending FFI functions to `libdd-data-pipeline-ffi`, enabling language tracers to construct spans field-by-field through the C API instead of passing pre-serialized msgpack.

New types:
- `TracerSpan`: opaque handle wrapping `Span<BytesData>`
- `TracerTraceChunks`: opaque handle wrapping `Vec<Vec<SpanBytes>>`

New functions:
- `ddog_tracer_span_new`, `_free`, `_set_meta`, `_set_metric`
- `ddog_tracer_trace_chunks_new`, `_free`, `_begin_chunk`, `_push_span`
- `ddog_trace_exporter_send_trace_chunks`

# Motivation

The Ruby tracer needs a native trace export path that bypasses Ruby-side msgpack serialization. This provides the FFI surface for it; the corresponding C extension in `dd-trace-rb` calls these functions.

Continuation of the prototype in #1661, reworked to stay close to `main`: the generic `TraceExporter<H>` and `SharedRuntime` are untouched; this is purely additive.

# Additional Notes

Spans are consumed on push, chunks are consumed on send — single-ownership enforced at the API level.

`libdd-trace-utils` is promoted from dev-dependency to regular dependency since the FFI crate now needs `SpanBytes` at build time.

Sibling PR for Ruby is at: https://github.com/DataDog/dd-trace-rb/pull/5690

# How to test the change?

`cargo test -p libdd-data-pipeline-ffi --lib` — 39 tests, all passing. The existing `trace_exporter` tests are unaffected.

### DIFF
--- a/libdd-data-pipeline-ffi/src/tracer.rs
+++ b/libdd-data-pipeline-ffi/src/tracer.rs
@@ -344,25 +344,27 @@ mod tests {
         CharSlice::from_bytes(s.as_bytes())
     }
 
-    unsafe fn make_minimal_span() -> Box<TracerSpan> {
-        let mut handle = MaybeUninit::<Box<TracerSpan>>::uninit();
-        let out = NonNull::new(handle.as_mut_ptr()).unwrap();
-        let err = ddog_tracer_span_new(
-            out,
-            cs("svc"),
-            cs("op"),
-            cs("res"),
-            cs(""),
-            1,
-            0,
-            1,
-            0,
-            0,
-            0,
-            0,
-        );
-        assert!(err.is_none());
-        handle.assume_init()
+    fn make_minimal_span() -> Box<TracerSpan> {
+        unsafe {
+            let mut handle = MaybeUninit::<Box<TracerSpan>>::uninit();
+            let out = NonNull::new(handle.as_mut_ptr()).unwrap();
+            let err = ddog_tracer_span_new(
+                out,
+                cs("svc"),
+                cs("op"),
+                cs("res"),
+                cs(""),
+                1,
+                0,
+                1,
+                0,
+                0,
+                0,
+                0,
+            );
+            assert!(err.is_none());
+            handle.assume_init()
+        }
     }
 
     #[test]
@@ -499,12 +501,14 @@ mod tests {
 
     // -- TracerTraceChunks tests --------------------------------------------
 
-    unsafe fn make_chunks(capacity: usize) -> Box<TracerTraceChunks> {
-        let mut handle = MaybeUninit::<Box<TracerTraceChunks>>::uninit();
-        let out = NonNull::new(handle.as_mut_ptr()).unwrap();
-        let err = ddog_tracer_trace_chunks_new(capacity, out);
-        assert!(err.is_none());
-        handle.assume_init()
+    fn make_chunks(capacity: usize) -> Box<TracerTraceChunks> {
+        unsafe {
+            let mut handle = MaybeUninit::<Box<TracerTraceChunks>>::uninit();
+            let out = NonNull::new(handle.as_mut_ptr()).unwrap();
+            let err = ddog_tracer_trace_chunks_new(capacity, out);
+            assert!(err.is_none());
+            handle.assume_init()
+        }
     }
 
     #[test]
--- a/libdd-data-pipeline-ffi/src/tracer.rs
+++ b/libdd-data-pipeline-ffi/src/tracer.rs
@@ -83,41 +83,32 @@ pub unsafe extern "C" fn ddog_tracer_span_new(
 ) -> Option<Box<ExporterError>> {
     catch_panic!(
         {
-            let service = match charslice_to_bytesstring(service) {
-                Ok(s) => s,
-                Err(e) => return Some(e),
-            };
-            let name = match charslice_to_bytesstring(name) {
-                Ok(s) => s,
-                Err(e) => return Some(e),
-            };
-            let resource = match charslice_to_bytesstring(resource) {
-                Ok(s) => s,
-                Err(e) => return Some(e),
-            };
-            let span_type = match charslice_to_bytesstring(span_type) {
-                Ok(s) => s,
-                Err(e) => return Some(e),
-            };
+            let inner = || -> Result<(), Box<ExporterError>> {
+                let service = charslice_to_bytesstring(service)?;
+                let name = charslice_to_bytesstring(name)?;
+                let resource = charslice_to_bytesstring(resource)?;
+                let span_type = charslice_to_bytesstring(span_type)?;
 
-            let trace_id: u128 = ((trace_id_high as u128) << 64) | (trace_id_low as u128);
+                let trace_id: u128 = ((trace_id_high as u128) << 64) | (trace_id_low as u128);
 
-            let span = SpanBytes {
-                service,
-                name,
-                resource,
-                r#type: span_type,
-                trace_id,
-                span_id,
-                parent_id,
-                start,
-                duration,
-                error,
-                ..Default::default()
+                let span = SpanBytes {
+                    service,
+                    name,
+                    resource,
+                    r#type: span_type,
+                    trace_id,
+                    span_id,
+                    parent_id,
+                    start,
+                    duration,
+                    error,
+                    ..Default::default()
+                };
+
+                out_handle.as_ptr().write(Box::new(TracerSpan(span)));
+                Ok(())
             };
-
-            out_handle.as_ptr().write(Box::new(TracerSpan(span)));
-            None
+            inner().err()
         },
         gen_error!(ErrorCode::Panic)
     )
--- a/libdd-data-pipeline-ffi/src/tracer.rs
+++ b/libdd-data-pipeline-ffi/src/tracer.rs
@@ -17,7 +17,6 @@ use libdd_common_ffi::slice::AsBytes;
 use libdd_common_ffi::CharSlice;
 use libdd_tinybytes::BytesString;
 use libdd_trace_utils::span::v04::SpanBytes;
-use std::mem::MaybeUninit;
 use std::ptr::NonNull;
 
 // ---------------------------------------------------------------------------
@@ -69,7 +68,7 @@ pub struct TracerSpan(SpanBytes);
 /// length.
 #[no_mangle]
 pub unsafe extern "C" fn ddog_tracer_span_new(
-    out_handle: &mut MaybeUninit<Box<TracerSpan>>,
+    out_handle: NonNull<Box<TracerSpan>>,
     service: CharSlice,
     name: CharSlice,
     resource: CharSlice,
@@ -84,7 +83,7 @@ pub unsafe extern "C" fn ddog_tracer_span_new(
 ) -> Option<Box<ExporterError>> {
     catch_panic!(
         {
-            let mut inner = || -> Result<(), Box<ExporterError>> {
+            let inner = || -> Result<(), Box<ExporterError>> {
                 let service = charslice_to_bytesstring(service)?;
                 let name = charslice_to_bytesstring(name)?;
                 let resource = charslice_to_bytesstring(resource)?;
@@ -106,7 +105,7 @@ pub unsafe extern "C" fn ddog_tracer_span_new(
                     ..Default::default()
                 };
 
-                out_handle.write(Box::new(TracerSpan(span)));
+                out_handle.as_ptr().write(Box::new(TracerSpan(span)));
                 Ok(())
             };
             inner().err()
@@ -210,12 +209,14 @@ pub struct TracerTraceChunks(Vec<Vec<SpanBytes>>);
 #[no_mangle]
 pub unsafe extern "C" fn ddog_tracer_trace_chunks_new(
     capacity: usize,
-    out_handle: &mut MaybeUninit<Box<TracerTraceChunks>>,
+    out_handle: NonNull<Box<TracerTraceChunks>>,
 ) -> Option<Box<ExporterError>> {
     catch_panic!(
         {
             let chunks = Vec::with_capacity(capacity);
-            out_handle.write(Box::new(TracerTraceChunks(chunks)));
+            out_handle
+                .as_ptr()
+                .write(Box::new(TracerTraceChunks(chunks)));
             None
         },
         gen_error!(ErrorCode::Panic)
@@ -337,8 +338,9 @@ mod tests {
     fn make_minimal_span() -> Box<TracerSpan> {
         unsafe {
             let mut handle = MaybeUninit::<Box<TracerSpan>>::uninit();
+            let out = NonNull::new(handle.as_mut_ptr()).unwrap();
             let err = ddog_tracer_span_new(
-                &mut handle,
+                out,
                 cs("svc"),
                 cs("op"),
                 cs("res"),
@@ -360,9 +362,10 @@ mod tests {
     fn new_sets_all_scalar_fields() {
         unsafe {
             let mut handle = MaybeUninit::<Box<TracerSpan>>::uninit();
+            let out = NonNull::new(handle.as_mut_ptr()).unwrap();
 
             let err = ddog_tracer_span_new(
-                &mut handle,
+                out,
                 cs("my-service"),
                 cs("web.request"),
                 cs("GET /users"),
@@ -473,9 +476,10 @@ mod tests {
     fn new_with_empty_strings_succeeds() {
         unsafe {
             let mut handle = MaybeUninit::<Box<TracerSpan>>::uninit();
+            let out = NonNull::new(handle.as_mut_ptr()).unwrap();
 
             let err =
-                ddog_tracer_span_new(&mut handle, cs(""), cs(""), cs(""), cs(""), 0, 0, 0, 0, 0, 0, 0);
+                ddog_tracer_span_new(out, cs(""), cs(""), cs(""), cs(""), 0, 0, 0, 0, 0, 0, 0);
             assert!(err.is_none());
 
             let span = handle.assume_init();
@@ -491,7 +495,8 @@ mod tests {
     fn make_chunks(capacity: usize) -> Box<TracerTraceChunks> {
         unsafe {
             let mut handle = MaybeUninit::<Box<TracerTraceChunks>>::uninit();
-            let err = ddog_tracer_trace_chunks_new(capacity, &mut handle);
+            let out = NonNull::new(handle.as_mut_ptr()).unwrap();
+            let err = ddog_tracer_trace_chunks_new(capacity, out);
             assert!(err.is_none());
             handle.assume_init()
         }
--- a/libdd-data-pipeline-ffi/src/tracer.rs
+++ b/libdd-data-pipeline-ffi/src/tracer.rs
@@ -222,11 +222,7 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_new(
 ) -> Option<Box<ExporterError>> {
     catch_panic!(
         {
-            let chunks = if capacity > 0 {
-                Vec::with_capacity(capacity)
-            } else {
-                Vec::new()
-            };
+            let chunks = Vec::with_capacity(capacity);
             out_handle
                 .as_ptr()
                 .write(Box::new(TracerTraceChunks(chunks)));
--- a/libdd-data-pipeline-ffi/src/tracer.rs
+++ b/libdd-data-pipeline-ffi/src/tracer.rs
@@ -28,13 +28,12 @@ use std::ptr::NonNull;
 /// Returns an error if the slice is not valid UTF-8.
 #[inline]
 fn charslice_to_bytesstring(s: CharSlice) -> Result<BytesString, Box<ExporterError>> {
-    match BytesString::from_slice(s.as_bytes()) {
-        Ok(bs) => Ok(bs),
-        Err(_) => Err(Box::new(ExporterError::new(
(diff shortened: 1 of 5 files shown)


libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,582 @@
...
+/// Returns an error if the slice is not valid UTF-8.
+#[inline]
+fn charslice_to_bytesstring(s: CharSlice) -> Result<BytesString, Box<ExporterError>> {
+    match BytesString::from_slice(s.as_bytes()) {
+        Ok(bs) => Ok(bs),
+        Err(_) => Err(Box::new(ExporterError::new(
+            ErrorCode::InvalidInput,
+            &ErrorCode::InvalidInput.to_string(),
+        ))),
+    }
```
yannham:
```suggestion
    BytesString::from_slice(s.as_bytes()).map_err(|_| {
        Box::new(ExporterError::new(
            ErrorCode::InvalidInput,
            &ErrorCode::InvalidInput.to_string(),
        ))
    })
```

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,582 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// TracerSpan
+// ---------------------------------------------------------------------------
+
+/// Opaque handle wrapping a single `Span<BytesData>`.
+pub struct TracerSpan(pub(crate) SpanBytes);
```
yannham: Why does the inner part need to be `pub(crate)`? It seems it's an FFI item only used in this module.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,582 @@
...
+/// * `error`     – Error status (0 = no error).
+///
+/// # Safety
+///
+/// `out_handle` must point to valid, writable memory for a `Box<TracerSpan>`.
+/// All `CharSlice` arguments must point to valid memory for their stated
+/// length.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_tracer_span_new(
+    out_handle: NonNull<Box<TracerSpan>>,
```
yannham, line 79: Why not take `&mut Box<TracerSpan>` here? I think cbindgen will generate the same interface in the end, but it's less unsafe/annoying code on the Rust part.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,582 @@
...
+    start: i64,
+    duration: i64,
+    error: i32,
+) -> Option<Box<ExporterError>> {
+    catch_panic!(
+        {
+            let service = match charslice_to_bytesstring(service) {
+                Ok(s) => s,
+                Err(e) => return Some(e),
+            };
```
yannham: It's a detail but I wonder if that wouldn't be nicer to define a local function that returns a `Result<(), Box<ExporterError>>` inside this one that does the actual work, where you can use `?` instead of all those tedious matches, and then just finally call `sub_function.err()`

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,582 @@
...
+    )
+}
+
+// ---------------------------------------------------------------------------
+// TracerTraceChunks
+// ---------------------------------------------------------------------------
+
+/// Opaque handle wrapping `Vec<Vec<SpanBytes>>` — a list of trace chunks,
+/// each containing a list of spans.
+pub struct TracerTraceChunks(pub(crate) Vec<Vec<SpanBytes>>);
```
yannham: Same question about `pub(crate)`

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,582 @@
...
+/// unknown.
+///
+/// # Safety
+///
+/// `out_handle` must point to valid, writable memory for a
+/// `Box<TracerTraceChunks>`.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_tracer_trace_chunks_new(
+    capacity: usize,
+    out_handle: NonNull<Box<TracerTraceChunks>>,
```
yannham, line 211: Same remark about `&mut`

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,582 @@
...
+) -> Option<Box<ExporterError>> {
+    catch_panic!(
+        {
+            let chunks = if capacity > 0 {
+                Vec::with_capacity(capacity)
+            } else {
+                Vec::new()
+            };
```
yannham: It's ok to use a zero capacity and should give the same as `Vec::new()`

```suggestion
            let chunks = Vec::with_capacity(capacity);
```

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,582 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn ddog_trace_exporter_send_trace_chunks(
+    exporter: Option<&TraceExporter>,
+    chunks: Box<TracerTraceChunks>,
+    response_out: Option<NonNull<Box<ExporterResponse>>>,
+) -> Option<Box<ExporterError>> {
+    let exporter = match exporter {
+        Some(e) => e,
+        None => return gen_error!(ErrorCode::InvalidArgument),
+    };
```
yannham:
```suggestion
    let Some(exporter) = exporter else {
        return gen_error!(ErrorCode::InvalidArgument);
    };
```

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,582 @@
...
+            assert_eq!(span.0.name.as_ref(), "");
+            assert_eq!(span.0.service.as_ref(), "");
+
+            ddog_tracer_span_free(span);
+        }
+    }
+
+    // -- TracerTraceChunks tests --------------------------------------------
+
+    unsafe fn make_chunks(capacity: usize) -> Box<TracerTraceChunks> {
```
yannham: This function shouldn't be unsafe, but rather have an `unsafe` block internally - the caller has no precondition to uphold.

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: `cbindgen` generates the same C signature (`T **out_handle`) for both approaches but `&mut Box<T>` is unsound for out-parameters.

`*out_handle = Box::new(...)` on a `&mut Box<T>` first drops the old value at that location, then writes the new one, but the out-parameter starts uninitialized: there's no valid `Box<T>` there yet. Dropping uninitialized memory is undefined behavior.

`NonNull<Box<T>>` with `out_handle.as_ptr().write(...)` uses `ptr::write`, which overwrites without dropping the old value.

From the C caller's perspective:

```/dev/null/example.c#L1-3
ddog_TracerSpan *span;  // uninitialized pointer
ddog_tracer_span_new(&span, ...);  // &span -> T**, old *span is garbage
```

With `&mut Box<T>`, Rust would presumably try to `free()` that garbage pointer before writing the new one, leading to a potential crash.

lloeki: PoC:

```rust
use std::mem::MaybeUninit;
use std::ptr::NonNull;

struct Foo(u64);

impl Drop for Foo {
    fn drop(&mut self) {
        println!("  drop Foo({})", self.0);
    }
}

/// Correct: `ptr::write` overwrites without dropping the old value.
unsafe fn init_nonnull(out: NonNull<Box<Foo>>) {
    out.as_ptr().write(Box::new(Foo(42)));
}

/// Incorrect: `*out = ...` drops the old `*out` first. If the old
/// value is uninitialized, the drop tries to free a garbage pointer.
unsafe fn init_mut_ref(out: &mut Box<Foo>) {
    *out = Box::new(Foo(99));
}

fn main() {
    println!("=== NonNull + ptr::write (correct) ===");
    unsafe {
        let mut slot = MaybeUninit::<Box<Foo>>::uninit();
        let ptr = NonNull::new(slot.as_mut_ptr()).unwrap();
        init_nonnull(ptr);
        let result = slot.assume_init();
        println!("  got Foo({})", result.0);
        drop(result);
    }

    println!();
    println!("=== &mut Box<T> assignment (UB) ===");
    unsafe {
        let mut slot = MaybeUninit::<Box<Foo>>::uninit();

        // Poison the slot with a recognizable bad pointer so the
        // crash is deterministic (otherwise uninit bytes may happen
        // to be NULL which some allocators silently ignore on free).
        std::ptr::write_bytes(slot.as_mut_ptr() as *mut u8, 0xAB, std::mem::size_of::<Box<Foo>>());

        // Reinterpret uninitialized memory as &mut Box<Foo>.
        let out: &mut Box<Foo> = &mut *slot.as_mut_ptr();

        // This line does:  drop_in_place(out); ptr::write(out, Box::new(Foo(99)));
        // The drop_in_place tries to free 0xABABABABABABABAB → crash.
        init_mut_ref(out);

        let result = slot.assume_init();
        println!("  got Foo({})", result.0);
        drop(result);
    }
}
```

```console
$ rustc poc_outparam.rs -o /tmp/poc_outparam && /tmp/poc_outparam; echo "exit: $?"
=== NonNull + ptr::write (correct) ===
  got Foo(42)
  drop Foo(42)

=== &mut Box<T> assignment (UB) ===
Segmentation fault (core dumped)
exit: 139
```

lloeki: I found a way with `MaybeUninit`:

```rust
use std::mem::MaybeUninit;
use std::ptr::NonNull;

struct Foo(u64);

impl Drop for Foo {
    fn drop(&mut self) {
        println!("  drop Foo({})", self.0);
    }
}

/// Correct: `ptr::write` overwrites without dropping the old value.
unsafe fn init_nonnull(out: NonNull<Box<Foo>>) {
    out.as_ptr().write(Box::new(Foo(42)));
}

/// Incorrect: `*out = ...` drops the old `*out` first. If the old
/// value is uninitialized, the drop tries to free a garbage pointer.
unsafe fn init_mut_ref(out: &mut Box<Foo>) {
    *out = Box::new(Foo(99));
}

/// Correct: `MaybeUninit::write` overwrites without dropping the old value,
/// and the type explicitly communicates the slot may be uninitialized.
unsafe fn init_maybeuninit(out: &mut MaybeUninit<Box<Foo>>) {
    out.write(Box::new(Foo(77)));
}

fn main() {
    println!("=== 1. NonNull + ptr::write (correct) ===");
    unsafe {
        let mut slot = MaybeUninit::<Box<Foo>>::uninit();
        std::ptr::write_bytes(slot.as_mut_ptr() as *mut u8, 0xAB, std::mem::size_of::<Box<Foo>>());

        let ptr = NonNull::new(slot.as_mut_ptr()).unwrap();
        init_nonnull(ptr);
        let result = slot.assume_init();
        println!("  got Foo({})", result.0);
        drop(result);
    }

    println!();
    println!("=== 2. &mut MaybeUninit<Box<T>> + write (correct) ===");
    unsafe {
        let mut slot = MaybeUninit::<Box<Foo>>::uninit();
        std::ptr::write_bytes(slot.as_mut_ptr() as *mut u8, 0xAB, std::mem::size_of::<Box<Foo>>());

        init_maybeuninit(&mut slot);
        let result = slot.assume_init();
        println!("  got Foo({})", result.0);
        drop(result);
    }

    println!();
    println!("=== 3. &mut Box<T> assignment (UB — will crash) ===");
    unsafe {
        let mut slot = MaybeUninit::<Box<Foo>>::uninit();
        std::ptr::write_bytes(slot.as_mut_ptr() as *mut u8, 0xAB, std::mem::size_of::<Box<Foo>>());

        let out: &mut Box<Foo> = &mut *slot.as_mut_ptr();
        // drop_in_place tries to free 0xABABABABABABABAB → crash
        init_mut_ref(out);

        let result = slot.assume_init();
        println!("  got Foo({})", result.0);
        drop(result);
    }
}
```

```console
=== 1. NonNull + ptr::write (correct) ===
  got Foo(42)
  drop Foo(42)

=== 2. &mut MaybeUninit<Box<T>> + write (correct) ===
  got Foo(77)
  drop Foo(77)

=== 3. &mut Box<T> assignment (UB — will crash) ===
Segmentation fault (core dumped)
```

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,566 @@
...
+///
+/// Subsequent [`ddog_tracer_trace_chunks_push_span`] calls will append
+/// spans to this chunk until the next `begin_chunk` call.
+///
+/// # Safety
+///
+/// `handle` must be a valid pointer to a `TracerTraceChunks`.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_tracer_trace_chunks_begin_chunk(
+    handle: Option<&mut TracerTraceChunks>,
```
hoolioh, line 250: In order to be consistent with the rest of the API I'd suggest to wrap check for a null handle and returning a proper error.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,566 @@
...
+/// unknown.
+///
+/// # Safety
+///
+/// `out_handle` must point to valid, writable memory for a
+/// `Box<TracerTraceChunks>`.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_tracer_trace_chunks_new(
+    capacity: usize,
+    out_handle: &mut MaybeUninit<Box<TracerTraceChunks>>,
```
hoolioh: I would suggest change the `out_handle` type to `NonNull<Box<TracerTraceChunks>>`. We're already using that nomenclature in the trace exporter and, since C headers are generated with cbindgen, the the type for the definitions could be inconsistent across the same module.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,566 @@
...
+/// * `duration`  – Duration in nanoseconds.
+/// * `error`     – Error status (0 = no error).
+///
+/// # Safety
+///
+/// `out_handle` must point to valid, writable memory for a `Box<TracerSpan>`.
+/// All `CharSlice` arguments must point to valid memory for their stated
+/// length.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_tracer_span_new(
```
hoolioh, line 78: I would use a `SpanFields` struct to gather all data and pass just one argument. It will more performant since the parameters won't be held in the stack on every call but, more importanly, if some type changes or is added later we won't have to change the signature of the function. It's something we're already doing in the `TraceExporter` module with `TraceExporterConfig` or `TelemetryClientConfig`.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,566 @@
...
+/// Start a new chunk (trace) inside the container.
+///
+/// Subsequent [`ddog_tracer_trace_chunks_push_span`] calls will append
+/// spans to this chunk until the next `begin_chunk` call.
+///
+/// # Safety
+///
+/// `handle` must be a valid pointer to a `TracerTraceChunks`.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_tracer_trace_chunks_begin_chunk(
```
hoolioh, line 249: Would it be helpful to pass a capacity parameter in order to prevent future reallocations?

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,566 @@
...
+/// Add or overwrite a string tag (`meta`) on the span.
+///
+/// Both `key` and `value` are copied into the span.
+///
+/// # Safety
+///
+/// `handle` must be a valid pointer to a `TracerSpan`.
+/// `key` and `value` must point to valid UTF-8 memory.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_tracer_span_set_meta(
```
hoolioh, line 138: is it usual from the tracer to set the fields one by one or in a batch? If the latter is an option I´d suggest changing the function signature to pass an array of keys, array of values and length in order to save some time at the cross boundary.
I suggested chaging the signature in order to increase the API surface but having both is more than OK if is ergonomic from the tracer point of view

hoolioh (changes requested):

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: Hmm, see https://github.com/DataDog/libdatadog/pull/1952#discussion_r3200306482 addressed by https://github.com/DataDog/libdatadog/pull/1952/commits/769297cac290c86d6047c5c8699d91602ec511f0.

Would you rather have me revert 769297cac290c86d6047c5c8699d91602ec511f0?

libdd-data-pipeline-ffi/src/tracer.rs
yannham: Ah, right, I didn't think of the fact that the pointer is initially uninitialized.  `MaybeUninit` doesn't buy us anything here, so I suppose your original approach makes the most sense.

libdd-data-pipeline-ffi/src/tracer.rs
yannham: Yes, I agree with Julio here. I thought we could get rid of unsafe and use plain, simple Rust types, but `MaybeUninit` is IMHO more complicated and less idiomatic than your first version.

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: I think so, yes, when the capacity is known beforehand, which is typical of processing a trace segment for export.

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: Ack, I'll defer to your Rust expertise! Will revert.

libdd-data-pipeline-ffi/src/tracer.rs
ivoanjo: I was kinda thinking about this when I was reviewing the code on the Ruby side... I think it's possible to stack-allocate an array and then pass it all in one go to libdatadog.

lloeki: In terms of interface, that might be OK for full span conversion that happens currently.

In terms of performance do we really gain anything? It looks like this would make it traverse the kv set twice: once via `meta_iter_cb` to build the `[key1, ...], [value1, ...], length` args, and then here again to insert each into `span.0.meta`. Sure the latter is entirely C/Rust, but the former boundary crossing remains, it's just that it's Ruby->C instead of Ruby->C->Rust which I don't think is the costly part. Or am I missing something?

lloeki: FYI the next prototypal step in the native spans project will _very_ quickly be to try backing span attribute storage with Rust objects, at which point we'll need the current "single field" variant.

I'd rather avoid interface churn on the libdatadog FFI side.

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: Agreed, will do.

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: Out of scope for this PR. Tracked for a later change.

lloeki:
Review comments addressed. If acceptable, I suggest we merge this one as a first iteration of the FFI API.

A new stacked PR based on this one is brewing, which has more contentious API changes coming from the dd-trace-rb side review.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -0,0 +1,604 @@
...
+///
+/// # Safety
+///
+/// `out_handle` must point to valid, writable memory for a `Box<TracerSpan>`.
+/// All `CharSlice` fields in `fields` must point to valid memory for their
+/// stated length.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_tracer_span_new(
+    out_handle: NonNull<Box<TracerSpan>>,
+    fields: &TracerSpanFields,
```
hoolioh, line 80: I'd wrap it with an `Option` type, that way we could benefit from nullable pointer optimization on that type and check it inside the function to prevent a panic if an invalid pointer is passed to the function. [Here](https://github.com/DataDog/libdatadog/blob/0a3304c6aaf84738786b670d706a01edc22dab81/libdd-data-pipeline-ffi/src/trace_exporter.rs#L209) is an example

hoolioh (approved):
I just left a comment requesting a change, other than that LGTM overall for a first version.

yannham (approved):

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: OK I'll address in the follow-up PR #2029

lloeki:
https://github.com/DataDog/libdatadog/commit/9d7998ba8755878ce4966152b0c86444becc5d6f had all CI OK and now post-rebase there's no CI running? 

EDIT: [GitHub Actions incident](https://www.githubstatus.com/incidents/gnftqj9htp0g)

----

20260526_14h29m03s_grim

----

20260526_14h31m14s_grim

PR #1953, opened 2026-05-06
feat(trace_buffer)!: flush based on size of chunks in bytes

paullegranddc:
# Motivation

When we buffer data, we want to put a bound on the memory we are going to use. Spans can contain a varying amount of data depending on the workload instrumented.

For this reason, it is better to trigger flush and limit the trace buffer based on the amount of data stored, not the number of spans.

This of  doesn't take into account that a lot of payloads contain duplicate string that will be interned if a user uses v0.5 or v1 encoding, but it would make this a lot harder to compute

# Changes

* Add a BufferSize trait, with a byte_size method returning the number of bytes each item in the queue allocates
* Use this method to compute the total size of each trace-chunk
* Use the total size in bytes in the queue to flush and limit the size of the buffer


yannham (commented):
Do you have an idea of the impact of the new `byte_size`? Might not be on the hot path, but it's a bit sad we have to basically deeply traverse each span entirely to compute the estimate, with potential for a bit of pointer chasing. I suppose one possibility would be to try to encode the span and measure the actual size in bytes, but I suppose this happens at a later point, so we can't really rely on encoding when we decide to flush or not?

Another solution would be to make spans record their size on the go. I'm not sure any of this is required, but just wanted to mention the topic.

paullegranddc:
> Do you have an idea of the impact of the new `byte_size`? 

Yes, it is actually quite cheap. We only do pointer chasing on the Hashmaps, as the lengths are otherwise in the fat pointer metadata.
The `trace_buffer/1_senders/no_delay` benchmark is only +30% slower (making exporting a trace chunk with a single span take 87ms)

Of course, of it's a trace chunk with 100 spans, that'd be a lot more expensive...
But I think this is still a fraction of the encoding cost as we are not doing any mem copy

@yannham

yannham:
> The trace_buffer/1_senders/no_delay benchmark is only +30% slower (making exporting a trace chunk with a single span take 87ms)

Is 30% a typo? Sounds like a lot to me :stuck_out_tongue:

yannham (approved):
So it's really 30% but on a micro benchmark. The code LGTM, so I'll let you make the perf call (or decide on any follow up measurement/bench), you have more context than me. As long as we're conscious of the trade-offs we make.

PR #1954, opened 2026-05-06
feat(trace-utils)!: search all spans to populate tracer payload fields

duncanpharvey:
# What does this PR do?

When extracting span tags to set in the payload, iterate through each trace to find a non-empty value for the tag. Search root spans on each trace first then additional spans if the root span has an empty value for the specified tag.

# Motivation

If a trace payload contains multiple trace chunks, and one of those chunks does not contain a trace that has a value for a given tag (like `version`), then the top level field set in the tracer payload could be empty even when other traces could have non-empty values for the same tag.

This situation arose with the .NET tracer sending a `command_execution` span with no version in the same payload as an `azure_functions.invoke` span with a version. As a result, the trace metric generated from this payload, `trace.azure_functions.invoke.hits`, did not have a version tag set.

https://datadoghq.atlassian.net/browse/SVLS-9006

# Additional Notes

- Mimics [searchTraceForField](https://github.com/DataDog/datadog-agent/blob/1310dea928fe3015531953576a9b5378df023e08/pkg/trace/version/version.go#L53-L66) from Go Agent.
- Renamed public struct `RootSpanTags` to `TracerPayloadTags` to better reflect the source of the populated fields.

# How to test the change?

Unit tests:
- First trace root span has no fields. Second trace root span has all fields. The second root span should populate all fields.
- Root span has no fields. Child span has all fields. The child span should populate all fields.
- Root span has all fields. Child has different values for all fields. The root span should populate all fields.
- Root span has empty values for all fields. Child span has a non-empty values. The child span should populate all fields.
- `env` value is normalized by `collect_pb_trace_chunks`
- Empty `env` values after normalization are skipped. in `collect_pb_trace_chunks`
- A span with the same span_id as root is treated as the root and skipped in the child span search. Only the root span's own meta is checked for it.


libdd-trace-utils/src/trace_utils.rs
```diff
@@ -262,13 +262,30 @@ where
...
+    pub app_version: String,
+    pub hostname: String,
+    pub runtime_id: String,
+}
+
+/// Returns the first value of `field` found in `trace`, searching the root span first then all
+/// other spans.
+fn search_trace_for_field(root: &pb::Span, trace: &[pb::Span], field: &str) -> Option<String> {
+    if let Some(v) = root.meta.get(field) {
+        return Some(v.clone());
```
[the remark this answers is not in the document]
duncanpharvey: Addressed in 3c10b32a71b0fec572eb6281f31e4f8942335fbe

duncanpharvey:
@codex review

libdd-trace-utils/src/trace_utils.rs
```diff
@@ -1278,4 +1296,107 @@ mod tests {
         assert!(!span.meta.contains_key("aas.site.kind"));
         assert!(!span.meta.contains_key("aas.site.type"));
     }
+
+    #[test]
```
Lewis-E, line 1305: So I was look at pb:Span, and it looks like there's a bunch of fuzzing used to test that code.  Do you think that a [bolero](https://github.com/camshaft/bolero) `check!()` type test might be useful here? How well formed are trace payloads, generally?

libdd-trace-utils/src/trace_utils.rs
```diff
@@ -656,23 +662,35 @@ pub fn collect_pb_trace_chunks<T: tracer_payload::TraceChunkProcessor>(
...
-                    "env" => root_span_tags.env,
-                    "version" => root_span_tags.app_version,
-                    "_dd.hostname" => root_span_tags.hostname,
-                    "runtime-id" => root_span_tags.runtime_id,
+        if is_agentless {
+            // Check each field independently so that a later trace can fill in fields missing
+            // from an earlier trace.
+            let root = &trace[root_span_index];
+            if tracer_payload_tags.env.is_empty() {
+                if let Some(v) = search_trace_for_field(root, trace, "env") {
```
Lewis-E: So this `trace` gets normalized up here right: https://github.com/DataDog/libdatadog/pull/1954/changes#diff-e35e80d6b64d2cc411fc38dad84cf1cf35bd47501f9ef20761c1d20c1b5abd26R631

But it looks like that normalization can have errors, which are logged, but don't break the loop. This is sort of why I'm wondering about fuzzing/trace payloads. Can we get incorrect children passing on incorrect tags?

duncanpharvey: Between these normalize functions, only the `env` field is normalized for the fields relevant to the tracer payload change (`env`, `version`, `_dd.hostname`, `runtime-id`).
- [normalize_trace](https://github.com/DataDog/libdatadog/blob/91fd13c8a0ca5335fe39940f8764cd825bbef7e8/libdd-trace-normalization/src/normalizer.rs#L56)
- [normalize_span](https://github.com/DataDog/libdatadog/blob/91fd13c8a0ca5335fe39940f8764cd825bbef7e8/libdd-trace-normalization/src/normalizer.rs#L19)
- [normalize_chunk](https://github.com/DataDog/libdatadog/blob/91fd13c8a0ca5335fe39940f8764cd825bbef7e8/libdd-trace-normalization/src/normalizer.rs#L78)

There is a scenario where normalization fails on a span and later spans are not normalized, resulting in potentially non-normalized `env` values being set on the tracer payload. I think the safest way to handle this scenario would be to apply normalization in `collect_pb_trace_chunks` rather than modifying how `normalize_trace` short circuits on an error when iterating over spans.

```
  if let Some(mut v) = search_trace_for_field(root, trace, "env") {
      normalize_utils::normalize_tag(&mut v);
      tracer_payload_tags.env = v;
  }
```

duncanpharvey: Applied change here: 00b025cd2e578338adf11ceca1a2c876682ecb0d

libdd-trace-utils/src/trace_utils.rs
duncanpharvey: [collect_pb_trace_chunks](https://github.com/DataDog/libdatadog/blob/963dadf3681d335ad27aa13aa4cde690613ff217/libdd-trace-utils/src/trace_utils.rs#L618), where `search_trace_for_field` is called, is certainly complex enough to warrant a bolero test in general for everything that it's doing. But for what we want to test as part of this PR in [search_trace_for_field](https://github.com/DataDog/libdatadog/blob/963dadf3681d335ad27aa13aa4cde690613ff217/libdd-trace-utils/src/trace_utils.rs#L276) - "best effort, first non-empty value wins from any span regardless of structural validity". It just means the bolero test can only assert "if a value is returned, it came from some span in the trace", which is already guaranteed from the other unit tests.

Lewis-E (approved):

VianneyRuhlmann (approved):

libdd-trace-utils/src/trace_utils.rs
```diff
@@ -1278,4 +1301,107 @@ mod tests {
...
+            &TracerHeaderTags::default(),
+            &mut tracer_payload::DefaultTraceChunkProcessor,
+            true,
+        )
+        .unwrap();
+
+        let TracerPayloadCollection::V07(payloads) = result else {
+            panic!("expected TracerPayloadCollection::V07");
+        };
+        assert_eq!(payloads[0].app_version, "1.2.3");
```
lucaspimentel, line 1338: Should we also test `env`, `_dd.hostname`, and `runtime-id`?

libdd-trace-utils/src/trace_utils.rs
```diff
@@ -656,23 +662,40 @@ pub fn collect_pb_trace_chunks<T: tracer_payload::TraceChunkProcessor>(
...
+            // from an earlier trace.
+            let root = &trace[root_span_index];
+            if tracer_payload_tags.env.is_empty() {
+                if let Some(mut v) = search_trace_for_field(root, trace, "env") {
+                    // Normalize env tag in case the span it was pulled from was skipped during
+                    // normalization
+                    libdd_trace_normalization::normalize_utils::normalize_tag(&mut v);
+                    if !v.is_empty() {
+                        tracer_payload_tags.env = v;
+                    }
```
lucaspimentel, line 676: Should we test this with an `env` that will be modified during normalization and an `env` that ends up empty after normalization?

libdd-trace-utils/src/trace_utils.rs
```diff
@@ -656,23 +662,40 @@ pub fn collect_pb_trace_chunks<T: tracer_payload::TraceChunkProcessor>(
...
-            gathered_root_span_tags = true;
-            let meta_map = &trace[root_span_index].meta;
-            parse_root_span_tags!(
-                meta_map,
-                {
-                    "env" => root_span_tags.env,
-                    "version" => root_span_tags.app_version,
-                    "_dd.hostname" => root_span_tags.hostname,
-                    "runtime-id" => root_span_tags.runtime_id,
+        if is_agentless {
```
lucaspimentel, line 665: When `is_agentless` set and what does it mean in this context? If this code is running in the agent, how can it be "agentless"? 😅

libdd-trace-utils/src/trace_utils.rs
duncanpharvey: Added these additional fields to the unit tests. 👍

libdd-trace-utils/src/trace_utils.rs
duncanpharvey: Added these unit tests: https://github.com/DataDog/libdatadog/blob/551934a0e8ad951e84acd06b74f6f67439ffbce6/libdd-trace-utils/src/trace_utils.rs#L1476-L1522

libdd-trace-utils/src/trace_utils.rs
duncanpharvey: I think this comes from the fact that these trace utils are used in the tracers in addition to the Serverless Compatibility Layer and Bottlecap. From the tracer perspective, if traces are sent directly to the backend, `is_agentless: true` means "do the normalization and other tasks the Go agent would normally do." Maybe a more appropriate name for this parameter would be something like `perform_agent_processing`?

- [collect_pb_trace_chunks call in Serverless Compatibility Layer](https://github.com/DataDog/serverless-components/blob/1a1bb927ed59b7611ddb8e1b4a1c83c0adb51f5e/crates/datadog-trace-agent/src/trace_processor.rs#L167)
- [collect_pb_trace_chunks call in Bottlecap](https://github.com/DataDog/datadog-lambda-extension/blob/1cbae3411d20f74f3a53eb72f275e1e817a8fe96/bottlecap/src/traces/trace_processor.rs#L359)

lucaspimentel: Thanks for clarifying! 

> Maybe a more appropriate name for this parameter would be something like `perform_agent_processing`

Makes sense, but probably out scope of this PR.

lucaspimentel (approved):

PR #1959, opened 2026-05-07
feat(crashtracking)!: update crashtracker config ffi to have values for collect all thread config

gyuheon0h:
# What does this PR do?

[feat(crashtracking)!: collect all threads](https://github.com/DataDog/libdatadog/pull/1878) added the functionality to collect all threads. We should expose the config vals for this

# Motivation

# Additional Notes

# How to test the change?
```
pub struct Config<'a> {
    pub additional_files: Slice<'a, CharSlice<'a>>,
    /// If true, the receiver will collect stack traces for all threads in the crashing process
    /// (not just the crashing thread) using ptrace-based remote unwinding.
    pub collect_all_threads: bool, *******THIS IS NEW*******
    pub create_alt_stack: bool,
    pub demangle_names: bool,
    /// The endpoint to send the crash report to (can be a file://).
    /// If None, the crashtracker will infer the agent host from env variables.
    pub endpoint: EndpointConfig<'a>,
    /// Maximum number of non-crashing threads to collect when `collect_all_threads` is true.
    /// If 0, uses the default (`libdd_crashtracker::default_max_threads()`).
    pub max_threads: usize, *******THIS IS NEW*******
    /// Optional filename for a unix domain socket if the receiver is used asynchonously
    pub optional_unix_socket_filename: CharSlice<'a>,
    pub resolve_frames: StacktraceCollection,
    /// The set of signals we should be registered for.
    /// If empty, use the default set.
    pub signals: Slice<'a, i32>,
    /// Timeout in milliseconds before the signal handler starts tearing things down to return.
    /// If 0, uses the default timeout as specified in
    /// `libdd_crashtracker::shared::constants::DD_CRASHTRACK_DEFAULT_TIMEOUT`. Otherwise, uses
    /// the specified timeout value.
    /// This is given as a uint32_t, but the actual timeout needs to fit inside of an i32 (max
    /// 2^31-1). This is a limitation of the various interfaces used to guarantee the timeout.
    pub timeout_ms: u32,
    pub use_alt_stack: bool,
}
```

### DIFF
--- a/examples/ffi/crashtracking.c
+++ b/examples/ffi/crashtracking.c
@@ -86,6 +86,8 @@ int main(int argc, char **argv) {
   // Get the default signals and explicitly use them.
   struct ddog_crasht_Slice_CInt signals = ddog_crasht_default_signals();
   ddog_crasht_Config config = {
+      .collect_all_threads = true,
+      .max_threads = 0, // 0 uses the default (libdd_crashtracker::default_max_threads())
       .create_alt_stack = false,
       .endpoint = {.url = slice(report_path)},
       .resolve_frames = DDOG_CRASHT_STACKTRACE_COLLECTION_ENABLED_WITH_INPROCESS_SYMBOLS,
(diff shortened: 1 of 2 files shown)


gyuheon0h:
* **#1959** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/1959?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/1959?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

taegyunkim (approved):

PR #1961, opened 2026-05-08
ci: use builder on windows platforms

hoolioh:
# What does this PR do?

  Switches the Windows release-artifact build over to the shared `builder` crate, so all platforms (Linux/glibc, Linux/musl, macOS, Windows) now go through the same path. To make that possible:

  - **`builder/src/arch/windows.rs`**: adds Windows support to the builder — copies the PDB and the `*.dll.lib` import library, uses the `_ffi`-suffixed artifact names produced by the FFI crate, and no-ops the pkg-config step (Windows has no
  `pkg-config`).
  - **`builder/src/arch/mod.rs` + `profiling.rs`** (`refactor: move add_pkg_config to arch`): hoists the per-platform `add_pkg_config` into each `arch/*.rs`, with the shared file-templating logic exposed as `arch::generate_pkg_config`. This is
  what lets Windows opt out cleanly instead of carrying a `#[cfg]` in `profiling.rs`.
  - **`builder/src/builder.rs`**: skips creating `bin/` and `pkgconfig/` directories on Windows, where they're unused.
  - **`windows/build-artifacts.ps1`**: deleted — its responsibilities now live in the `builder`.
  - **CI cleanup** (separate commits, batched in because they unblock the Windows-on-builder rollout):
    - `cargo nextest run` invocations now pass `--no-tests=pass` so the workflow doesn't fail when a partition/affected-crate set yields zero tests.
    - `.github/workflows/test.yml` no longer hand-filters `builder` / `test_spawn_from_lib` out of the package list — `--no-tests=pass` makes that filtering unnecessary, and dropping it means the `builder` crate's own tests are actually
  exercised by CI when it changes.
    - Same fix applied to `miri.yml` and `coverage.yml`.

  # Motivation

  We've been maintaining two parallel build pipelines: every other platform goes through the `builder` crate, while Windows had its own PowerShell script (`windows/build-artifacts.ps1`) that re-implemented overlapping logic (artifact naming,
  output layout, header post-processing). Any change to how artifacts are produced had to be made (and kept in sync) in two places. Routing Windows through the `builder` consolidates this and lets future build changes ship in one place.

  The CI changes are required collateral: once `builder` is the path used for the Windows release build, the workflow's hard-coded `--exclude builder` is wrong, and partition-style invocations need `--no-tests=pass` so the workflow tolerates
  empty test sets.

  # Additional Notes

  - **Companion change in `libddprof-build`**: this PR depends on a matching branch in the private `libddprof-build` repo. The last commit (`dfeffd0b4 chore: reference libddprof branch that matches the changes in the builder`) temporarily points
   `.gitlab-ci.yml`'s `DOWNSTREAM_BRANCH` at that branch. **This commit must be reverted to `main` before merge** — please flag this in review if I forget.
  - Windows artifact names move from `datadog_profiling.{dll,lib,pdb}` to `datadog_profiling_ffi.{dll,lib,pdb}` plus the new `datadog_profiling_ffi.dll.lib` import library. Downstream consumers that hard-coded the old names will need to follow.
  - No changes to public APIs or to non-Windows artifact layout.


gleocadie (approved):
overall looks good. The other PR has a comment about enabled features

PR #1964, opened 2026-05-09
chore(docs): RFC 0015 introduce no_std

pawelchcki:
# What does this PR do?

Adds RFC 0015 proposing a workspace policy: prefer `no_std + alloc`; use `std` only where it earns its keep.

# Motivation

`no_std` is, for a lot of what libdatadog does, the better Rust. `core` makes async-signal-unsafe items unreachable by construction (relevant for crashtracker and profiling samplers), embedders linking statically pay for less code, and once a crate is `no_std` every new dependency comes with a forced "what does this drag in?" review.

Many migrations are mechanical (`std::` → `core::`, see [yaml/yaml-serde#8](https://github.com/yaml/yaml-serde/pull/8)). #1770 is the in-flight reference implementation for `libdd-library-config`.

# How to test the change?

Read it. Comment if you disagree.


docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
+# RFC 0015: Use `no_std` where its beneficial for consumers
+
+## Context
+
+`no_std` is, for most of what libdatadog does, the better Rust. Not because we want to run on bare metal, but because the constraints `no_std` imposes — explicit allocation, no hidden global state, no `std::` machinery dragged in transitively — line up almost perfectly with the constraints we are *already* trying to honour as a library that ships into other people's runtimes.
+
+Concretely, four things make `no_std` attractive for this workspace:
+
+1. **Signal safety by construction.** `core` and `alloc` (with a signal-safe allocator) are made of pure functions, integer math, and stack-allocated data. None of `std`'s mutex, thread-local, environment, file-descriptor, or panic-handler machinery is reachable. Code that runs in async-signal contexts — crashtracker, profiling samplers, anything called from a signal handler — is *much* easier to keep correct when `std::` is simply not in the import graph. The compiler enforces what code review otherwise has to.
```
morrisonlevi, line 9: Not really. Just because you don't call signal-unsafe items doesn't make what you do signal safe. Yes, it's certainly easier, but it's not automatic. You still have to worry about reentrancy, atomicity, etc.

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+2. **Smaller artifacts.** Embedders linking libdatadog statically pay for everything `std` pulls in, whether they use it or not. `no_std + alloc` lets us ship the same functionality with substantially less code in the final binary, and noticeably faster compiles in the tree.
+3. **Frequently, it's a mechanical change.** A surprising amount of "make this `no_std`" work is replacing `std::` with `core::` and adding `extern crate alloc;`. yaml/yaml-serde#8 is a recent example: a near-mechanical patch turned an `std` crate into a `no_std + alloc` crate without changing its API. Many of our internal crates are in the same shape.
+
+The first concrete driver in this workspace is `libdd-library-config` (prototyped in the sibling worktree `no-std-library-config`), but the case generalises: data structures, parsers, protocol definitions, error types, and signal-handler-adjacent code all benefit. The exceptions — sockets, files, threads, processes — are real but bounded.
+
+This RFC proposes the policy.
+
+## The thesis
+
+**Prefer `no_std + alloc`. Use `std` only where it is earning its keep.**
```
morrisonlevi, line 19: Why do we want `alloc`? We also shouldn't panic, and tons of users of `alloc` panic on allocation failure. What's the motivation for including `alloc` in the recommendation?

docs/RFCs/0015-no-std.md
pawelchcki: Yeah, I framed it like it solves all problems :D

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+
+1. **Signal safety by construction.** `core` and `alloc` (with a signal-safe allocator) are made of pure functions, integer math, and stack-allocated data. None of `std`'s mutex, thread-local, environment, file-descriptor, or panic-handler machinery is reachable. Code that runs in async-signal contexts — crashtracker, profiling samplers, anything called from a signal handler — is *much* easier to keep correct when `std::` is simply not in the import graph. The compiler enforces what code review otherwise has to.
+2. **Smaller artifacts.** Embedders linking libdatadog statically pay for everything `std` pulls in, whether they use it or not. `no_std + alloc` lets us ship the same functionality with substantially less code in the final binary, and noticeably faster compiles in the tree.
+3. **Frequently, it's a mechanical change.** A surprising amount of "make this `no_std`" work is replacing `std::` with `core::` and adding `extern crate alloc;`. yaml/yaml-serde#8 is a recent example: a near-mechanical patch turned an `std` crate into a `no_std + alloc` crate without changing its API. Many of our internal crates are in the same shape.
```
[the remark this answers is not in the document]
pawelchcki: sigh

docs/RFCs/0015-no-std.md
pawelchcki: I just found `no_std` + `alloc` is super easy to use and solving all my problems, and you really can have some safe'ish alloc implementations. and 

Ideally people could target `no_std` and add alloc on top if needed. 

tl;dr; `alloc` does make `no_std` quite bearable

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+
+Concretely, four things make `no_std` attractive for this workspace:
+
+1. **Signal safety by construction.** `core` and `alloc` (with a signal-safe allocator) are made of pure functions, integer math, and stack-allocated data. None of `std`'s mutex, thread-local, environment, file-descriptor, or panic-handler machinery is reachable. Code that runs in async-signal contexts — crashtracker, profiling samplers, anything called from a signal handler — is *much* easier to keep correct when `std::` is simply not in the import graph. The compiler enforces what code review otherwise has to.
+2. **Smaller artifacts.** Embedders linking libdatadog statically pay for everything `std` pulls in, whether they use it or not. `no_std + alloc` lets us ship the same functionality with substantially less code in the final binary, and noticeably faster compiles in the tree.
```
yannham, line 10: Just as a note, with the right build options (basically LTO for both the static library and the end consumer), dead-code elimination can be pretty close to what you'd have using libdatadog as a direct dependency, or so I would expect. Even without that, the linker does perform some dead-code elimination, so you don't get all of std automatically (nor necessarily all the surface used by the static library). But it's true that good code elimination is more fragile when consuming an already built static library.

docs/RFCs/0015-no-std.md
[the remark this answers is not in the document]
yannham: I think the `yaml-serde` thing does show an obstacle though: we are bound by our dependencies. If a core dependency of some libdatadog crate doesn't support no_std, that might be a problem (as an example, there's still no activity on the yaml-serde PR from upstream). That being said, we can just not support `no_std` in that specific case, if there's no choice.

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+3. **Frequently, it's a mechanical change.** A surprising amount of "make this `no_std`" work is replacing `std::` with `core::` and adding `extern crate alloc;`. yaml/yaml-serde#8 is a recent example: a near-mechanical patch turned an `std` crate into a `no_std + alloc` crate without changing its API. Many of our internal crates are in the same shape.
+
+The first concrete driver in this workspace is `libdd-library-config` (prototyped in the sibling worktree `no-std-library-config`), but the case generalises: data structures, parsers, protocol definitions, error types, and signal-handler-adjacent code all benefit. The exceptions — sockets, files, threads, processes — are real but bounded.
```
yannham, line 13: While talking about datastructures: there's no `HashMap` implementation in `alloc`, forcing the use of `BTreeMap` or something similar, which has different (and often worse, although not always) performance characteristics.

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+
+Crates that are out of scope by nature — their reason for existing is OS interaction: `datadog-sidecar*`, `datadog-ipc*`, `libdd-shared-runtime*`, `libdd-http-client`, `libdd-data-pipeline`, `spawn_worker`, all `*-ffi` shells. These stay `std`.
+
+## Drawbacks
+
+- **Build matrix grows.** Each opted-in crate adds a `--no-default-features` build to CI. Real but bounded.
+- **no_std rust ecosystem is large, but not all crates support it** Code willing to support `no_std` might have to do extra work to align dependencies with `no_std` requirements.
+- **Cognitive overhead in opted-in crates.** Contributors have to use `core::`/`alloc::` and gate `std`-only code. We consider this a feature: it forces the same discipline we'd want at code-review time anyway.
+- **Adding a dependency becomes a small research task.** Does it support `no_std`? With which features? Mostly this is good — it discourages casual dependency growth — but it is friction.
+- **Forks accumulate maintenance debt.** This risk is real and should be weighed carefully before adopting or maintaining a fork.
```
yannham, line 98: This is the first time forks are mentioned. Does this refer to forking a dependency to make it `no_std` if upstream is unwilling/unresponsive? I think this is a bad idea in general, the maintenance work is going to be unbearable with security implications, and this potentially duplicates some dependencies for `dd-trace-rs` end-users.

docs/RFCs/0015-no-std.md
ivoanjo: For Ruby we're consuming libdatadog as a dynlib and while I wouldn't call it "set in stone", I think it's very likely this won't change for a long while, so we can't rely on LTO.

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+
+1. **Signal safety by construction.** `core` and `alloc` (with a signal-safe allocator) are made of pure functions, integer math, and stack-allocated data. None of `std`'s mutex, thread-local, environment, file-descriptor, or panic-handler machinery is reachable. Code that runs in async-signal contexts — crashtracker, profiling samplers, anything called from a signal handler — is *much* easier to keep correct when `std::` is simply not in the import graph. The compiler enforces what code review otherwise has to.
+2. **Smaller artifacts.** Embedders linking libdatadog statically pay for everything `std` pulls in, whether they use it or not. `no_std + alloc` lets us ship the same functionality with substantially less code in the final binary, and noticeably faster compiles in the tree.
+3. **Frequently, it's a mechanical change.** A surprising amount of "make this `no_std`" work is replacing `std::` with `core::` and adding `extern crate alloc;`. yaml/yaml-serde#8 is a recent example: a near-mechanical patch turned an `std` crate into a `no_std + alloc` crate without changing its API. Many of our internal crates are in the same shape.
```
Eldolfin, line 11: Clippy can help here with the [clippy::std_instead_of_core](https://rust-lang.github.io/rust-clippy/master/index.html?search=std_instead_of_core#std_instead_of_core) and [clippy::std_instead_of_alloc](https://rust-lang.github.io/rust-clippy/master/index.html?search=std_instead_of_alloc#std_instead_of_alloc) lints. It can  probably auto-fix most of it too by choosing core/alloc itself :)

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
+# RFC 0015: Use `no_std` where it's beneficial for consumers
+
+## Context
+
+`no_std` is, for most of what libdatadog does, the better Rust. Not because we want to run on bare metal, but because the constraints `no_std` imposes — explicit allocation, no hidden global state, no `std::` machinery dragged in transitively — line up almost perfectly with the constraints we are *already* trying to honour as a library that ships into other people's runtimes.
+
+Concretely, four things make `no_std` attractive for this workspace:
+
+1. **Signal safety by construction.** `core` and `alloc` (with a signal-safe allocator) are made of pure functions, integer math, and stack-allocated data. None of `std`'s mutex, thread-local, environment, file-descriptor, or panic-handler machinery is reachable. Code that runs in async-signal contexts — crashtracker, profiling samplers, anything called from a signal handler — is *much* easier to keep correct when `std::` is simply not in the import graph. The compiler enforces what code review otherwise has to.
```
danielsn, line 9: There is a push to have a panic-handler around the FFI calls to prevent panics leaking into customer applications, how does this fit with that?

docs/RFCs/0015-no-std.md
danielsn: Could we solve this by having a `std` feature?

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+
+This RFC proposes the policy.
+
+## The thesis
+
+**Prefer `no_std + alloc`. Use `std` only where it is earning its keep.**
+
+Concretely, that means:
+
+- For **new crates**, the default should be `no_std + alloc` unless the crate's reason for existing is OS interaction.
```
danielsn, line 23: Or needs to use certain data-structures like hash-map.

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+- For **existing crates**, `no_std` support is added opportunistically: whenever a crate is touched substantially, or whenever a downstream consumer asks, evaluate whether the migration is cheap. If it is — and for many of our crates it will be — do it.
```
danielsn, line 24: Could we experiment with `no_std` + a `std` feature, and see how much would need to live under that feature?

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+**Crate root.** Conditional `no_std`, unconditional `alloc`. We rely on a heap; we do not target true bare-metal.
+
+```rust
+#![cfg_attr(not(feature = "std"), no_std)]
+extern crate alloc;
+```
+
+**Imports.** Use `core::` and `alloc::` everywhere they exist. Gate genuinely `std`-only items behind `#[cfg(feature = "std")]`:
```
danielsn, line 54: I had asked about this above.  How cleanly does this end up working, or do we end up with feature tags everywhere?

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+use alloc::vec::Vec;
+use core::cell::OnceCell;
+
+#[cfg(feature = "std")]
+use std::path::Path;
+```
+
+**Dependencies.** Every dependency is declared `default-features = false`. Anything the dependency only exposes under its own `std` feature is forwarded through this crate's `std` feature. Optional dependencies that are inherently `std` (`libc`, `memfd`, `prost`, etc.) live behind `dep:` in the `std` feature list.
+
+**Errors.** `thiserror` v2 and `anyhow` (with `default-features = false`) work in `no_std` and should be preferred over hand-rolled error enums.
```
danielsn, line 67: I assume you mean in `alloc` crates? `thiserror` should be preferred over `anyhow` in general for library code.  That said, what error mechanism (thisError, Anyhow, hand-rolled error codes) is orthogonal to this RFC and should get its own.

docs/RFCs/0015-no-std.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Workspace enforcement
+
+When a crate opts in:
+
+- CI builds it with `--no-default-features` in addition to the default build. Without this, a careless `use std::` lands and silently breaks embedders.
```
danielsn, line 73: Does this cause combinatorial explosion in CI?

docs/RFCs/0015-no-std.md
danielsn: Was this LLM generated?

docs/RFCs/0015-no-std.md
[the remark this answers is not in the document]
pawelchcki: > That being said, we can just not support no_std in that specific case, if there's no choice.

We can always fork permanently. At some point there is only so much we can do if someone is not responsive for upstreaming.

docs/RFCs/0015-no-std.md
pawelchcki: There were few iterations.  and possibly earlier revisions had it better fleshed out.

Overall I wouldn't demonize forks too much - yes it adds maintenance burden. But so does writing our business code n-times. 

Also currently the situation in with afew of our low level tools is such, that we simply can't use libdatadog without no_std. And we have to resort to writing our own - everything, because very few pieces of software are written to be compatible with the signal safe context.

docs/RFCs/0015-no-std.md
pawelchcki: Crates can switch between map implementation based on feature availability. e.g. if compiling for 'std' use HashMap, BTreeMap otherwise.

docs/RFCs/0015-no-std.md
pawelchcki: This is basically what I think we'll end up with. Take a look at anyhow or this error. Many very popular crates have, graceful degradation to `no_std`.

docs/RFCs/0015-no-std.md
pawelchcki: Ack. I probably sneaked too much of my preferences for thiserror :D

docs/RFCs/0015-no-std.md
pawelchcki: Since I learned clippy has lints for core and alloc. I assume maybe we won't have to do it.

Also I expect this to be only a single additional target

docs/RFCs/0015-no-std.md
pawelchcki: OOh these would probably solve 95% of problems

pawelchcki:
the clippy checks have been merged, https://github.com/DataDog/libdatadog/pull/2196

PR #1968, opened 2026-05-11
crashtracker: verify unix socket peer uid before parsing reports

paullegranddc:
### Motivation
- The UNIX-socket receiver previously accepted the first connection without authentication and trusted client-supplied crash configuration, allowing a local attacker to inject forged reports that read receiver-readable files and exfiltrate them.
- A minimal, targeted hardening is needed to prevent cross-user local spoofing without changing the reporting protocol.

### Description
- Added a Linux-only helper `ensure_same_user` that uses `getsockopt(..., SO_PEERCRED, ...)` to obtain the peer `ucred` and compares the peer UID to the receiver effective UID via `libc::geteuid()`.
- The check is invoked for accepted connections in `async_receiver_entry_point_unix_listener` before any parsing of the incoming crashreport stream.
- Imported `AsRawFd` for access to the accepted socket file descriptor and kept the change confined to `libdd-crashtracker/src/receiver/entry_points.rs` to minimize surface area.
- The change rejects connections from other users with a clear error, preserving the existing protocol and single-accept behavior for same-user clients.

### Testing
- Ran `cargo fmt --all` and formatting completed successfully.
- Ran `cargo test -p libdd-crashtracker --no-run` which compiled the crate tests successfully (tests were prepared but not executed) and completed without build errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_69fe431b2d2c832285b44ab93428817b)

### DIFF
--- a/libdd-crashtracker/src/receiver/entry_points.rs
+++ b/libdd-crashtracker/src/receiver/entry_points.rs
@@ -7,6 +7,8 @@ use crate::CrashtrackerConfiguration;
 #[cfg(target_os = "linux")]
 use crate::StacktraceCollection;
 use anyhow::Context;
+#[cfg(target_os = "linux")]
+use std::os::fd::AsRawFd;
 use std::time::Duration;
 use tokio::{
     io::{AsyncBufReadExt, BufReader},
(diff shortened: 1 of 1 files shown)


libdd-crashtracker/src/receiver/entry_points.rs
```diff
@@ -30,10 +32,38 @@ pub async fn async_receiver_entry_point_unix_listener(
     listener: &UnixListener,
 ) -> anyhow::Result<()> {
     let (unix_stream, _) = listener.accept().await?;
+    #[cfg(target_os = "linux")]
+    ensure_same_user(&unix_stream)?;
```
pawelchcki, line 36: It would be good to add tests verifying the new enforcement

bwoebi:
This assumes that the scripts themselves don't execute set(e)uid() as part of their normal execution.
I.e. the process spawns as root, then the receiver gets setup,.. later the process changes to its actual user and any further crashtracking reports are ignored.

(e.g. with the sidecar, the sidecar spawns at startup, but crashtracker gets its dedicated socket, which would now fail to access.)

paullegranddc:
> This assumes that the scripts themselves don't execute set(e)uid() as part of their normal execution. I.e. the process spawns as root, then the receiver gets setup,.. later the process changes to its actual user and any further crashtracking reports are ignored.
> 
> (e.g. with the sidecar, the sidecar spawns at startup, but crashtracker gets its dedicated socket, which would now fail to access.)

Then maybe instead of checking uid, it could be interesting to check that the connecting process either is the parent process, or has the same parent process as the crashtracker

bwoebi:
I suppose `uid == sender uid OR sender pid == ppid` as logic would be fine.

PR #1969, opened 2026-05-11
refactor(ffe): validate at parse time that total_shards is non-zero

dd-oleksii:
# What does this PR do?

This fixes a spurious finding that shard calculation may panic on division by zero.

# Motivation

Spurious security scan finding.

### DIFF
--- a/datadog-ffe/src/rules_based/ufc/models.rs
+++ b/datadog-ffe/src/rules_based/ufc/models.rs
@@ -1,7 +1,7 @@
 // Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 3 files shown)


paullegranddc (approved):

datadog-ffe/src/rules_based/ufc/models.rs
```diff
@@ -537,7 +537,7 @@ pub(crate) struct SplitWire {
 #[allow(missing_docs)]
 pub(crate) struct ShardWire {
     pub salt: String,
-    pub total_shards: u32,
+    pub total_shards: NonZeroU32,
```
sameerank, line 540: TIL this is a type haha

sameerank (approved):

PR #1977, opened 2026-05-12
perf(sampling): optimize matching and limit cache memory [APMSP-2948]

bantonsson:
# What does this PR do?

Optimizes the glob matching and limits the memory taken up by the LRU match cache.

# Motivation

A follow up PR for bigger performance questions that came up in #1927 as well as bounding the cache memory for the LRU cache that came up in a security review.

# How to test the change?

Unit tests and benchmarks are in the PR.

### DIFF
--- a/libdd-sampling/src/bounded_byte_cache.rs
+++ b/libdd-sampling/src/bounded_byte_cache.rs
@@ -16,8 +16,12 @@ use std::num::NonZeroUsize;
 /// Default maximum entry count.
 pub const DEFAULT_MAX_ENTRIES: usize = 256;
 
-/// Default maximum tracked byte size (256 KiB).
-pub const DEFAULT_MAX_BYTES: usize = 256 * 1024;
+/// Default maximum tracked byte size (64 KiB).
+///
+/// Under normal use (short keys such as service or resource names), the entry count cap binds
+/// first. This byte budget acts as a safety valve against memory exhaustion from unexpectedly
+/// large keys due to misconfiguration or adversarial input.
+pub const DEFAULT_MAX_BYTES: usize = 256 * 256;
 
 /// LRU cache bounded by both entry count and total tracked byte size.
 ///
@@ -57,7 +61,6 @@ where
 
     /// Insert `key -> value`. Entries larger than `max_bytes` are dropped silently. Otherwise
     /// LRU entries are evicted until the new entry fits.
-    #[inline]
     pub fn put(&mut self, key: K, value: V) {
         let entry_bytes = Self::entry_size(&key);
 
@@ -91,6 +94,7 @@ where
     }
 
     #[cfg(test)]
+    #[inline]
     pub fn current_bytes(&self) -> usize {
         self.current_bytes
     }
--- a/libdd-sampling/src/bounded_byte_cache.rs
+++ b/libdd-sampling/src/bounded_byte_cache.rs
@@ -1,0 +1,215 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! LRU cache wrapper with dual limits: maximum entry count AND maximum tracked byte size.
+//!
+//! `lru::LruCache` only supports an entry-count capacity, which is unsafe for caches keyed on
+//! arbitrary user strings: a few very large keys can balloon memory. `BoundedByteCache` adds
+//! a byte budget on top, evicting least-recently-used entries until both limits are satisfied.
+
+use lru::LruCache;
+use std::borrow::Borrow;
+use std::hash::Hash;
+use std::mem::size_of;
+use std::num::NonZeroUsize;
+
+/// Default maximum entry count.
+pub const DEFAULT_MAX_ENTRIES: usize = 256;
+
+/// Default maximum tracked byte size (256 KiB).
+pub const DEFAULT_MAX_BYTES: usize = 256 * 1024;
+
+/// LRU cache bounded by both entry count and total tracked byte size.
+///
+/// Byte accounting covers `key.as_ref().len() + size_of::<V>()`. Heap-allocated value contents
+/// are not tracked; this wrapper assumes small inline values (e.g. `bool`).
+pub struct BoundedByteCache<K, V>
+where
+    K: Hash + Eq + AsRef<[u8]>,
+{
+    inner: LruCache<K, V>,
+    current_bytes: usize,
+    max_bytes: usize,
+}
+
+impl<K, V> BoundedByteCache<K, V>
+where
+    K: Hash + Eq + AsRef<[u8]>,
+{
+    /// `max_entries` of zero is treated as 1 (a cache with no slots is nonsensical).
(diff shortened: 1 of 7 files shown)


libdd-sampling/src/glob_matcher.rs
```diff
@@ -48,104 +81,160 @@ impl GlobMatcher {
...
+        // Wildcard ASCII: cache results keyed on the raw subject bytes. Repeated calls with
+        // the same input (the common case in sampling rules) become an O(1) lookup, avoiding
+        // potentially O(P·S) backtracking. Different case spellings get separate entries,
+        // which is acceptable: real callers don't mix cases for the same logical value, and
+        // lowercasing for canonicalization would allocate on every call.
+        //
+        // The lock is released around `glob_match_bytes` so a slow (worst-case backtracking)
+        // match on one thread doesn't block cache hits on other threads sharing this matcher.
+        // A concurrent inserter racing us on the same key is harmless: `put` overwrites with
+        // the same boolean result.
```
iunanua, line 120: But N concurrent inserters would call `glob_match_bytes` if there is a cache miss.
I guess that's the price we have to pay.

libdd-sampling/src/bounded_byte_cache.rs
```diff
@@ -0,0 +1,199 @@
...
+        // `push` may evict an LRU entry to honor the entry-count cap; deduct its bytes.
+        if let Some((replaced_key, _)) = self.inner.push(key, value) {
+            self.current_bytes = self
+                .current_bytes
+                .saturating_sub(Self::entry_size(&replaced_key));
+        }
+        self.current_bytes += entry_bytes;
+    }
+
+    #[allow(dead_code)]
```
iunanua: should it be `#[cfg(test)]` instead of `#[allow(dead_code)]`?

libdd-sampling/src/bounded_byte_cache.rs
```diff
@@ -0,0 +1,199 @@
...
+        self.current_bytes
+    }
+
+    pub fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    fn entry_size(key: &K) -> usize {
+        key.as_ref().len() + size_of::<V>()
+    }
```
iunanua, line 108: Claude says:

`BoundedByteCache::entry_size` is an under-estimate of real heap usage. `key.as_ref().len() + size_of::<V>()` doesn't account for:
    - `Vec<u8>` overhead per key (24 bytes on 64-bit) — for a 256-entry default cap, that's 6 KiB unaccounted-for at full occupancy.
    - `lru::LruCache`'s internal doubly-linked-list node + `HashMap` bucket overhead per entry (~48-72 bytes).

Is is worth it?

```suggestion
  /// Approximate per-entry fixed overhead, in bytes. Covers:
  /// - `Vec<u8>` header on the key (24 B on 64-bit targets)
  /// - `lru` doubly-linked-list node (prev/next pointers, ~24 B)
  /// - HashMap bucket amortized (~16 B)
  ///
  /// Deliberately rounded up so `max_bytes` is a *pessimistic upper bound* on actual heap
  /// usage, not an underestimate. Recheck if the `lru` crate is upgraded across a major
  /// version — the linked-list node layout is the volatile piece.
  const PER_ENTRY_OVERHEAD: usize = 64;

  fn entry_size(key: &K) -> usize {
      PER_ENTRY_OVERHEAD + key.as_ref().len() + size_of::<V>()
  }
```

bantonsson: I think it makes sense. I'll add it and some tests.

libdd-sampling/src/bounded_byte_cache.rs
```diff
@@ -0,0 +1,215 @@
...
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq + ?Sized,
+    {
+        self.inner.get(key)
+    }
+
+    /// Insert `key -> value`. Entries larger than `max_bytes` are dropped silently. Otherwise
+    /// LRU entries are evicted until the new entry fits.
+    #[inline]
```
yannham: This looks rather big for an inline annotation (although the compiler still gets to choose)

libdd-sampling/src/bounded_byte_cache.rs
```diff
@@ -0,0 +1,215 @@
...
+                .saturating_sub(Self::entry_size(&replaced_key));
+        }
+        self.current_bytes += entry_bytes;
+    }
+
+    #[cfg(test)]
+    pub fn current_bytes(&self) -> usize {
+        self.current_bytes
+    }
+
```
yannham, line 101:
```suggestion
#[inline]
```

libdd-sampling/src/glob_matcher.rs
```diff
@@ -33,13 +57,22 @@ impl fmt::Debug for GlobMatcher {
...
 
 impl GlobMatcher {
-    /// Creates a new GlobMatcher with the given pattern
+    /// Creates a new GlobMatcher with the given pattern.
     pub fn new(pattern: &str) -> Self {
-        // Use a cache of size 256
-        let cache_size = unsafe { NonZeroUsize::new_unchecked(256) };
+        let pattern_lower = pattern.to_lowercase();
+        let pattern_is_ascii = pattern_lower.is_ascii();
+        let pattern_has_wildcards = pattern_lower.contains('*') || pattern_lower.contains('?');
```
yannham: It's really a question of taste, and I'm loosing this battle because LLM tends to have a different taste 😛 but as old-but-not-too-old man yelling at AI, I think stuff like `pattern_lower` and `pattern_is_ascii` should be inlined below.

libdd-sampling/src/glob_matcher.rs
yannham: If we really need to worry about that in the future, there are some concurrent implementations of LRU caches (https://crates.io/crates/concurrent_lru for one)

yannham (approved):

libdd-sampling/src/bounded_byte_cache.rs
```diff
@@ -0,0 +1,215 @@
...
+use std::borrow::Borrow;
+use std::hash::Hash;
+use std::mem::size_of;
+use std::num::NonZeroUsize;
+
+/// Default maximum entry count.
+pub const DEFAULT_MAX_ENTRIES: usize = 256;
+
+/// Default maximum tracked byte size (256 KiB).
+pub const DEFAULT_MAX_BYTES: usize = 256 * 1024;
```
paullegranddc: I'm not sure these limits together make sense to me, as that would require very 1KB attributes on average, which would be very surprising 🤔

paullegranddc (approved):

libdd-sampling/src/bounded_byte_cache.rs
bantonsson: The limit is only there to guard against surprising attributes (misconfiguration or attacks). I mean 1K long non normalized SQL strings is not unheard of, but maybe we should make it lower. I'll cap it at `256 * 256`.

PR #1978, opened 2026-05-12
ci: prevent miri fail by allowing nextest to pass on no-tests condition

hoolioh:
# What does this PR do?

  **CI: nextest hardening** — `.github/workflows/{miri,test,coverage}.yml`
  - Added `--no-tests=pass` to every dynamic-scope nextest invocation
    (9 total). Prevents exit-code-4 "no tests to run" failures when the
    affected-crates list contains a test-less crate.
  - Dropped three now-redundant `--exclude builder` and one jq filter
    `select(. != "builder")` that all served the same purpose as the
    new flag.
 
# Motivation
Tests can fail if no tests are run by nextest. This will allow future utility crates to be compiled without failing.

### DIFF
--- a/.github/workflows/miri.yml
+++ b/.github/workflows/miri.yml
@@ -73,6 +73,6 @@ jobs:
         with:
           cache-targets: true # cache build artifacts
           cache-bin: true # cache the ~/.cargo/bin directory
-      - run: MIRIFLAGS="-Zmiri-disable-isolation" cargo miri nextest run ${{ needs.setup.outputs.crates }} --partition count:${{ matrix.partition }}/5
+      - run: MIRIFLAGS="-Zmiri-disable-isolation" cargo miri nextest run ${{ needs.setup.outputs.crates }} --partition count:${{ matrix.partition }}/5 --no-tests=pass
       # We need to disable isolation because
       # "unsupported operation: `clock_gettime` with `REALTIME` clocks not available when isolation is enabled"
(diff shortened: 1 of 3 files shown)


.github/workflows/miri.yml
```diff
@@ -73,6 +73,6 @@ jobs:
         with:
           cache-targets: true # cache build artifacts
           cache-bin: true # cache the ~/.cargo/bin directory
-      - run: MIRIFLAGS="-Zmiri-disable-isolation" cargo miri nextest run ${{ needs.setup.outputs.crates }} --partition count:${{ matrix.partition }}/5
+      - run: MIRIFLAGS="-Zmiri-disable-isolation" cargo miri nextest run ${{ needs.setup.outputs.crates }} --partition count:${{ matrix.partition }}/5 --no-tests=pass
```
ekump, line 76: This might just be concern trolling, but is it possible in the future we incorrectly filter out test and not know about it? I guess we'd catch it if tests start breaking on the MergeQueue?

ekump (approved):

.github/workflows/miri.yml
hoolioh: According to nextest documentation that option only applies when not a single test is executed, by default nextest fails if the suite is empty. The worst thing can happen is that a crate with no tests will be merged but I guess that will be detected over the PR review process. In any case I think we can tweak the code coverage workflow to emit a warning when a crate has no tests?
On the other hand as you said, the tests for the workspace will be executed for push events so they will be run on the merge queue.

PR #1979, opened 2026-05-12
test(ffe): use canonical fixtures

leoromanovsky:
## Motivation

Use `DataDog/ffe-system-test-data` as libdatadog's canonical FFE/OpenFeature JSON fixture source instead of keeping local copied fixtures.

## Changes and Decisions

- Add `datadog-ffe/ffe-system-test-data` as a submodule pinned to merged canonical `main` commit `4446371bc1ca52bd526356927ef42d380145b118`; remove copied `datadog-ffe/tests/data` fixtures.
- Point datadog-ffe tests, benches, the UFC parse test, and the FFI harness at canonical `ufc-config.json` and `evaluation-cases/*.json`, with paths anchored from `CARGO_MANIFEST_DIR`.
- Assert canonical fixture values only. `result.reason` stays out of scope while reason behavior is still evolving.
- Initialize submodules in FFI and benchmark CI checkouts so the fixture files exist where those jobs run.
- Add CODEOWNERS coverage for `.gitmodules` and weekly Dependabot `gitsubmodule` updates.

https://datadoghq.atlassian.net/wiki/spaces/PANA/pages/6713639211/FFE+SDK+Fixture+Contribution+Guide


datadog-ffe/src/rules_based/eval/eval_assignment.rs
```diff
@@ -230,20 +230,49 @@ mod tests {
...
+                AssignmentReason::Split => "SPLIT",
+                AssignmentReason::Static => "STATIC",
+            },
+            Err(EvaluationError::FlagDisabled) => "DISABLED",
+            Err(
+                EvaluationError::DefaultAllocationNull
+                | EvaluationError::FlagUnrecognizedOrDisabled,
+            ) => "DEFAULT",
+            Err(_) => "ERROR",
+        }
```
leoromanovsky: Ignore assertion of reason; revert all code changes. That will enable migration in python & ruby now.

sameerank (approved):
It might help to include instructions like this for folks who may not be familiar

> If this is a fresh clone, initialize submodules first:
> 
> ```bash
> git submodule update --init --recursive
> ```

perhaps in this section?

https://github.com/DataDog/libdatadog/blob/6e559d75fd184d5cc46c0caf8b78c2b31a36af5e/README.md#L27-L30

datadog-ffe-test-suite/ffe-system-test-data
hoolioh, line 1: Since `datadog-ffe` is used in several SDKs we would eventually need to publish it. Having git submodules embedded into the crate proved to be problematic for third party consumers. In order to minimize risks, I would recommend the following options:

1. Create another crate for the canonical tests which will contain the git submodule and pulls the `datadog-ffe` as a dependency, this approach is used in another pojects like `serde` (`test-suite`). That way there will be no dependency to an internal repo on the published crate. That crate can be part of the workspace, the tests will be run each time `datadog-ffe` is modified and since it will be mark with `publish = false` there will be no risk of getting into trouble with downstream projects.
2. If option 1 is not feasible because of time of any project limitation, at the very least I would consider feature gate those tests and disable them by default. That way third party consumers won't fall into building issues if the tests are triggered. However this will require to modify the CI a bit in order to include that feature into the tests workflow.

hoolioh (changes requested):

hoolioh (approved):
LGTM, thank you for doing the changes.

PR #1980, opened 2026-05-12
feat(sidecar)!: Add direct exec capabilities to the sidecar via dynamic linker

bwoebi:
Via `-Wl,-e,ddog_spawn_direct_entry_body` the sidecar can now be directly executed by invoking it via `/lib64/ld-linux-*.so.2` (we read it from PT_INTERP).

This avoids memfd/tempfile and fexecve() shenanigans by default.

### DIFF
--- a/spawn_worker/src/unix/mod.rs
+++ b/spawn_worker/src/unix/mod.rs
@@ -59,7 +59,7 @@ pub fn getpid() -> libc::pid_t {
 
 /// Return the path to the dynamic linker (PT_INTERP) of the current process.
 #[cfg(target_os = "linux")]
-pub fn read_pt_interp_self() -> Option<std::path::PathBuf> {
+pub fn read_pt_interp_self() -> Option<PathBuf> {
     // Auxiliary vector entries for the current process's executable PHDRs.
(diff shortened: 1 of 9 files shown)


spawn_worker/src/unix/mod.rs
```diff
@@ -57,6 +57,53 @@ pub fn getpid() -> libc::pid_t {
...
+        let p_type = unsafe { (*ph).p_type };
+        let p_vaddr = unsafe { (*ph).p_vaddr } as usize;
+
+        if p_type == libc::PT_PHDR {
+            // load_base = runtime_addr_of_PHDRs − link-time vaddr of PHDRs
+            load_base = phdr_addr as isize - p_vaddr as isize;
+        }
+        if p_type == 3 {
+            // PT_INTERP = 3; the interpreter string lives at this vaddr.
+            interp_vaddr = p_vaddr;
```
Leiyks:
```suggestion
        if p_type == libc::PT_INTERP {
            interp_vaddr = p_vaddr;
```
Can't we do just that ?

Leiyks (approved):
looks ok to me 👌

PR #1982, opened 2026-05-13
feat: cross-language LTO to inline C TLS shim into Rust FFI

yannham:
# What does this PR do?

Add opt-in cross-language LTO so the C TLS shim (`libdd_get_otel_thread_ctx_v1`) is inlined directly into the Rust FFI functions, eliminating a function-call indirection on every TLS access.

The mode is controlled by the `LIBDD_OTEL_THREAD_CTX_INLINE` env var. A `build-optimized.sh` wrapper script sets the env var and the target-scoped `RUSTFLAGS` needed for cross-language LTO. Without the env var, the build behaves exactly as before.

# Motivation

The OTel thread-level context spec requires TLSDESC for the TLS variable, which forces us through a C shim since stable `rustc` doesn't expose TLS dialect control. This adds a function call on every `attach`/`detach`. Cross-language LTO eliminates that call.

Benchmark results (median of 3 batch runs on x86-64):

| Benchmark | Default | Inline (LTO) | Improvement |
|---|---|---|---|
| single/in_place | ~17.1 ns | ~13.5 ns | **~20%** |
| single/swap | ~21.9 ns | ~20.1 ns | **~8%** |
| multi/swap/4 | ~7.4 ns | ~5.3 ns | **~28%** |
| multi/swap/16 | ~14.5 ns | ~8.7 ns | **~40%** |

# Additional Notes

- Requires `clang` and `lld` (the toolchain's bundled `rust-lld` is used automatically).
- A wrapper script is needed because `build.rs` cannot set rustc codegen flags (`-Clinker-plugin-lto`, `-Clinker=clang`). Those must come from `RUSTFLAGS`.
- The build script validates prerequisites (clang available, LLD version) and fails with a clear error if something is missing.
- The `#[cfg(target_os/arch)]` in build scripts are replaced with the correct `CARGO_CFG_*` env vars for cross-compilation correctness.

# How to test the change?

```bash
# Default build (no inlining, same as before)
cargo build --release -p libdd-otel-thread-ctx-ffi
cargo test --release -p libdd-otel-thread-ctx-ffi

# Optimized build (cross-language LTO)
./libdd-otel-thread-ctx-ffi/build-optimized.sh

# Verify inlining succeeded (symbol should be absent)
nm target/x86_64-unknown-linux-gnu/release/liblibdd_otel_thread_ctx_ffi.so | grep libdd_get_otel_thread_ctx
# (no output = inlined)

# Verify TLSDESC preserved
readelf -r target/x86_64-unknown-linux-gnu/release/liblibdd_otel_thread_ctx_ffi.so | grep TLSDESC
# should show R_X86_64_TLSDESC for otel_thread_ctx_v1
```


libdd-otel-thread-ctx/build.rs
```diff
@@ -1,20 +1,58 @@
...
 
 fn main() {
-    // Only compile the TLS shim on Linux.
-    #[cfg(target_os = "linux")]
-    {
-        let mut build = cc::Build::new();
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+
+    if target_os != "linux" {
```
yannham, line 20: This change is somehow unrelated, but using static `cfg(target_os = "linux")` is actually not the way to go in a build script (during cross-compilation typically, the target_os of the script and of the dynamic library could differ), so fixing it passing by.

yannham:
@codex review

libdd-otel-thread-ctx-ffi/build-optimized.sh
ivoanjo, line 1: So... I'm curious (and this is not a blocker for the PR) -- does this script _need_ to exist?

That is, do the environment variables need to be defined before `cargo build` gets invoked, or is there perhaps a path to making this something that we could have as a flag in our builder crate?

I'm thinking that if we could move stuff to rust code, it would probably be a bit more ergonomic for users + easier to maintain and validate.

libdd-otel-thread-ctx-ffi/build-optimized.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Sanity-check that the C shim was actually inlined, if `nm` is available.
+if ! command -v nm &>/dev/null; then
+		echo >&2 "WARNING: skipping sanity check that the C TLS shim was inlined (\`nm\` not found)"
+else
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+    SO="$REPO_ROOT/target/$TARGET/release/liblibdd_otel_thread_ctx_ffi.so"
+
+    if [[ -f "$SO" ]] && nm "$SO" 2>/dev/null | grep -q 'libdd_get_otel_thread_ctx'; then
```
scottgerring: nit: if `nm` fails it looks like the symbol was absent and stderr is suppressed; maybe we could do something like TODO?

libdd-otel-thread-ctx-ffi/build-optimized.sh
```diff
@@ -0,0 +1,63 @@
...
+        echo >&2 "WARNING: build succeeded but the C TLS shim (libdd_get_otel_thread_ctx_v1) was NOT inlined."
+        echo >&2 "Cross-language LTO may not be working. Check that clang and lld versions are recent enough and compatible with the Rust toolchain's LLVM."
+        exit 1
+    fi
```
ivoanjo: Wait, am I reading this correctly? 👀  Shouldn't the symbol still be there for the outside readers? What am I missing?

libdd-otel-thread-ctx-ffi/build-optimized.sh
```diff
@@ -0,0 +1,63 @@
...
+done
+
+if [[ -z "$TARGET" ]]; then
+    TARGET=$(rustc -vV | sed -n 's/host: //p')
+fi
+
+# CARGO_TARGET_<TRIPLE>_RUSTFLAGS scopes the flags to the target only, keeping
+# build scripts and proc-macros unaffected.
+TARGET_ENV=$(echo "$TARGET" | tr 'a-z-' 'A-Z_')
+export "CARGO_TARGET_${TARGET_ENV}_RUSTFLAGS=-Clinker-plugin-lto -Clinker=clang"
```
scottgerring: Should we try preserve existing `CARGO_TARGET_${TARGET_ENV}_RUSTFLAGS` and append to it? I don't have enough surrounding context on our build infra landscape but blasting over it in its entirety might be unexpected.

libdd-otel-thread-ctx-ffi/build-optimized.sh
```diff
@@ -0,0 +1,63 @@
...
+export LIBDD_OTEL_THREAD_CTX_INLINE=1
+
+cargo build --release \
+    --target "$TARGET" \
+    -p libdd-otel-thread-ctx-ffi \
+    "${EXTRA_ARGS[@]}"
+
+# Sanity-check that the C shim was actually inlined, if `nm` is available.
```
scottgerring, line 52: > If `nm` was available

Will this be the case in our CI build pipeline for the repo?

libdd-otel-thread-ctx/build.rs
```diff
@@ -1,20 +1,64 @@
...
         // - On aarch64, TLSDESC is already the only dynamic TLS model so no flag is needed.
         // - On x86-64, we use `-mtls-dialect=gnu2` (supported since GCC 4.4 and Clang 19+) to force
         //   the use of TLSDESC as mandated by the spec. If it's not supported, this build will
         //   fail.
-        #[cfg(target_arch = "x86_64")]
-        build.flag("-mtls-dialect=gnu2");
-
-        build.file("src/tls_shim.c").compile("tls_shim");
-        println!("cargo:rerun-if-changed=src/tls_shim.c");
+        if target_arch == "x86_64" {
```
scottgerring: This isn't really related to this PR, but should we throw if we see Linux with a different architecture from these two? That way in the future if some enterprising engineer adds risc-v or whatever, we'll know that we have to address this.

I figure the linker flags to specify dialect differ between architectures _anyway_, so it is unlikely this will just work out of the box.

libdd-otel-thread-ctx-ffi/build-optimized.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
```
scottgerring, line 1: I haven’t looked deeply into this and it may be a rabbit hole, but could `.cargo/config.toml` help with the required codegen flags here? My understanding is that it would likely need to be workspace/repo-wide which broadens the impact compared with this wrapper script. But maybe longer-term we want this kind of LTO setup in the release pipeline anyway?

libdd-otel-thread-ctx-ffi/build-optimized.sh
```diff
@@ -0,0 +1,63 @@
...
+# Sanity-check that the C shim was actually inlined, if `nm` is available.
+if ! command -v nm &>/dev/null; then
+		echo >&2 "WARNING: skipping sanity check that the C TLS shim was inlined (\`nm\` not found)"
+else
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+    SO="$REPO_ROOT/target/$TARGET/release/liblibdd_otel_thread_ctx_ffi.so"
+
+    if [[ -f "$SO" ]] && nm "$SO" 2>/dev/null | grep -q 'libdd_get_otel_thread_ctx'; then
+        echo >&2 "WARNING: build succeeded but the C TLS shim (libdd_get_otel_thread_ctx_v1) was NOT inlined."
```
ivoanjo: Very minor: Since we have the `exit 1` at the bottom, this is not really a warning ;)

(And I think having the `exit 1` is correct btw!)

libdd-otel-thread-ctx-ffi/README.md
```diff
@@ -0,0 +1,51 @@
...
+./build-optimized.sh
+```
+
+The script auto-detects the host triple. To cross-compile:
+
+```bash
+./build-optimized.sh --target aarch64-unknown-linux-gnu
+```
+
+Extra arguments are forwarded to `cargo build`.
```
ivoanjo, line 61: It might be useful to document a way for downstreams to validate that the optimized build did happen. E.g. I'd love to be able to add a check in https://github.com/DataDog/libdatadog-rb/blob/main/spec/gem_packaging.rb to validate that the LTO was correctly in use.

ivoanjo (commented):
I really like this optimization, left a few comments since I'm a bit... confused about some of the details :D

libdd-otel-thread-ctx-ffi/build-optimized.sh
yannham: This is not the TLS symbol itself, but the small C wrapper function that makes it possible to access the symbol, the one we do want to be gone :slightly_smiling_face: (and which is a Rust artifact of not supporting TLSDESC). The TLS symbol doesn't have the `libdd_get` suffix.

libdd-otel-thread-ctx-ffi/README.md
yannham: This is already automatically verified in the script, but I can surely reproduce the steps here in the readme, if you think that's useful.

libdd-otel-thread-ctx-ffi/build-optimized.sh
ivoanjo: You're right, I'm an idiot lololol I promise I've only been drinking tea all morning and not alcohol.

libdd-otel-thread-ctx-ffi/README.md
ivoanjo: My thinking is -- the script checks, but who's checking that the script is in use? ;)

Maybe someone redid the ruby build build entirely and forgot to even add the script -- that's the thing I was thinking of catching.

libdd-otel-thread-ctx-ffi/build-optimized.sh
yannham: Unfortunately, I think we need the script (or something similar that sets the flags from outside of the build) and can't be fully automatic. This is indeed annoying. Quoting this PR's description:

> A wrapper script is needed because build.rs cannot set rustc codegen flags (-Clinker-plugin-lto, -Clinker=clang). Those must come from RUSTFLAGS.

My understanding is that by design (for reasons I'm not sure to understand, to be honest), the `build.rs` can't set compiler flags, only linker or linking-related flags. Cf https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-flags (and found other people wanting to do the same, and couldn't).

build-common/src/lib.rs
```diff
@@ -10,3 +14,30 @@ pub fn copy_and_configure_headers() {}
...
 pub use crate::cbindgen::*;
+
+/// Locate the `gcc-ld/` shim directory shipped with the Rust toolchain.
+///
+/// This directory contains an `ld.lld` wrapper that delegates to `rust-lld`.
+/// Passing it via `-B` to the C compiler driver makes it discover rust-lld
+/// before any system-wide lld, which
+///
+/// 1. Avoids the need for a system-wide LLD install.
+/// 2. Picks a recent LLD that match the Rust toolchain's LLVM version
```
scottgerring: I thought the counter example was illustrative! 
```suggestion
/// 2. Picks a recent LLD that matches the Rust toolchain's LLVM version, as opposed to e.g. CentOS 7' LLVM7 which is too old to handle TLSDESC
```

libdd-otel-thread-ctx-ffi/build.rs
```diff
@@ -2,42 +2,85 @@
...
-    let dir = PathBuf::from(sysroot)
-        .join("lib/rustlib")
-        .join(&target)
-        .join("bin/gcc-ld");
-
-    dir.join("ld.lld").exists().then_some(dir)
+/// Returns the rust-lld `gcc-ld/` directory if found; `None` means the system
+/// `ld.lld` will be used instead. Panics with a clear message when the
+/// requirements are not met.
+fn require_lld_for_inline(target_arch: &str) -> Option<PathBuf> {
```
scottgerring: The name reads a little funny to me in terms of what the function does; maybe resolve works better?

```suggestion
fn resolve_lld_for_inline(target_arch: &str) -> Option<PathBuf> {
```

scottgerring (commented):
Nice - great to see the improvements to :D

libdd-otel-thread-ctx-ffi/build-optimized.sh
scottgerring: I was wondering this too. I put a comment further down that we _might_ be able to do this with .cargo/config.toml, but that would be project wide if it did work

yannham: > I was wondering this too. I put a comment further down that we might be able to do this with .cargo/config.toml, but that would be project wide if it did work

Good point, this is why I didn't do it in libdatadog, but if you're having a `libdatadog-mylang` facade crate anyway, that might be doable option? Will add this to the readme.

libdd-otel-thread-ctx-ffi/build-optimized.sh
yannham: Yes, indeed `.cargo/config.toml` is an alternative way. However the blast radius is much bigger: as you say, this is repo-wide, and additionally it can't be triggered "dynamically" for different build modes. Typically I fear that might be hell for the CI, and even on downstream SDKs, some aren't necessarily using latest clang versions (I remember the first otel thread context PR broke PHP because they weren't using a recent enough clang). But it's an alternative, and it should be documented 👍

libdd-otel-thread-ctx-ffi/build-optimized.sh
yannham: Fair!

libdd-otel-thread-ctx-ffi/build-optimized.sh
yannham: No, and this is the point: I don't want to fight against CentOS and Alpine with the requirements for LTO (clang, lld, LLVM 18+, ...). I think this will be on the build workflow of the final dynamic library to check that.

libdd-otel-thread-ctx-ffi/build-optimized.sh
ivoanjo: Oh, I missed that bit in the PR description, although I'd say that it very much deserves to live in a **why does this script need to exist** in the script itself :)

Having said that, I understand the limitation around `build.rs` (thanks for clarifying), yet on our own setup we often use the [builder crate](https://github.com/DataDog/libdatadog/tree/main/builder) that then invokes cargo for us; could we auto-add the flags when libdatadog is built in that manner perhaps?

yannham: I'm not familiar with the builder crate, but all it takes is to set two env variable before cargo starts (`RUST_FLAGS` and the toggle introduced here). So it sounds like we could definitely do that 👍

ivoanjo: Afaik right now the builder crate is the recommended way :tm: to build libdatadog if you're not consuming it from rust.

ivoanjo: (Although supporting it is not a blocker for this PR, to be clear, although it is a really nice to have :tm:)

libdd-otel-thread-ctx-ffi/build-optimized.sh
scottgerring: annnnd I was meant to replace that TODO with a concrete suggestion

libdd-otel-thread-ctx-ffi/build-optimized.sh
scottgerring: Might be worth smoking it out on the CI for this repo just to see what breaks? Adds weight to the "here's why we have a separate script" thing 🤷  don't feel super strongly about this

yannham: The CI is annoying but can probably get sorted? The thing I'm more worried about with this is downstream users like dd-trace-php. I can try on the CI, but I think it's insufficient. We would have to make sure it doesn't break any downstream consumer of libdatadog, all the dd-trace-xxx. Even worse, with dd-trace-rs, we impose constraints on the toolchain of end-users/customers directly by transitivity. We can discuss it in the component team but my gut feeling is that it's a a bold move to force everyone to use our specific configuration and clang.

ivoanjo: > my gut feeling is that it's a a bold move to force everyone to use our specific configuration and clang.

Yeah, I'd +1 that as well. Ideally we can make it as turn-key and easy to do and hard to miss as possible, but not go as far as making it a hard requirement.

libdd-otel-thread-ctx/build.rs
yannham: Good point! Added a panic for unsupported architectures in both `libdd-otel-thread-ctx/build.rs` (the `-mtls-dialect` flag selection) and `libdd-otel-thread-ctx-ffi/build.rs` (the inline-mode `-mllvm -enable-tlsdesc` flag). This way if someone adds e.g. risc-v they'll get a clear build error pointing them to validate the TLSDESC handling and linker flags for the new arch.

hoolioh (approved):
LGTM. One caveat, probably the possibilities are scarce but I think it'd be worth to have an integration CI job that builds inline mode and asserts the invariants for the supported platforms. Since this depends on the external toolchains there could be situations in which the machinery could not work, we would silently default to not inlining the function and cause a regression. That said, I think it could be done in a following PR

PR #1985, opened 2026-05-13
feat(data-pipeline)!: CSS Trace Filters

Eldolfin:
# What does this PR do?

Implements the CSS trace-level filtering mechanism, that is applied before stats computation.

# Motivation

It's un unmet requirement from the spec.

# Additional Notes
~Please Ctrl+F `FIXME` when reviewing, I let some questions in the code that I'd like answers for before merging~

## TODO:

- [x] unit tests
- [x] system-tests https://github.com/DataDog/system-tests/pull/6952
- [x] normalization before filtering
- [x] regex keys (currently only regex values are checked)
- [x] new spec or CSS spec RFC
- [x] telemetry ? Count dropped traces with the reason
- [x] trim spaces when parsing config
- [x] Move implementation to trace-utils
- [x] stop using v4 span but generic spans instead for php
- [x] move arcswap to the trace exporter (remove update_conf)
- [x] use test agent for snapshot tests

## Wont do
- [ ] ~advanced resource normalization (sql, redis etc...).~ No other implementation does that
- [ ] ~100% normalization edge cases (empty service field + non-normalized name + filter depending on that)~ Too specific edge case. Ignoring it for now
- [ ] ~check on meta_struct too~ Not in the spec

[APMSP-2763]

[APMSP-2763]: https://datadoghq.atlassian.net/browse/APMSP-2763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -347,6 +349,12 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
     fn check_agent_info(&self) {
         if let Some(agent_info) = agent_info::get_agent_info() {
             if self.has_agent_info_state_changed(&agent_info) {
+                // FIXME: trace_filterer should only be enabled when CSS is on. (why ?)
```
bwoebi: Really? does that matter?

bwoebi: To my understanding this feature is a pre-requisite for filtering of stats to work, but it's a standalone feature, per se?

Eldolfin: I agree but it increases the risk in case the implementation is wrong: if we drop more trace than what the agent would have dropped it would affect everyone but if we apply filters only with CSS this only affects a small subset of users.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -286,6 +291,24 @@ impl TraceExporterBuilder {
...
+        self.filter_tags_regex = filter_tags_regex;
+        self
+    }
+
+    // TODO: doc
+    pub fn set_ignore_resources(&mut self, ignore_resources: Vec<String>) -> &mut Self {
+        self.ignore_resources = ignore_resources;
+        self
+    }
+
```
VianneyRuhlmann: I don't think we want to configure trace filters programmatically. They're only passed by the agent /info.

libdd-data-pipeline/src/trace_exporter/snapshots/libdd_data_pipeline__trace_exporter__single_threaded_tests__trace_filters.snap
ekump, line 1: any particular reason why the filename is using a combo of single and double underscores?

ekump:
It looks like this PR is still in progress and should probably be moved to draft.

libdd-data-pipeline/src/trace_exporter/mod.rs
ekump: I agree with @bwoebi. 

Complex features (like the trace exporter, V1, CSS, etc) should be gradually rolled out. Does this rise to the level of complexity where we need a gradual rollout (I'm not confident it does)? If so, the correct way to do that is let the SDKs handle it. We should provide an option in the builder so that SDKs can opt-in, and coordinate with LP to figure out how they want to approach the rollout. 

> risk in case the implementation is wrong

We should be confident in our implementations. Doesn't this functionality already exist in several tracers? Are there existing system tests? Do we have sufficient unit and integration tests in libdatadog? The point of gradual rollouts is to limit the blast radius of unknown unknowns when running on customer environments. This PR doesn't really introduce any new and complicated paradigms. It's behavior should be deterministic and reproducible.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -347,6 +349,12 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
+                self.trace_filterer.update_conf(
```
ekump: Isn't this overriding the values passed in via the [builder](https://github.com/DataDog/libdatadog/pull/1985/changes#diff-bc58d8d129128b72456204f3dbe30c5236e14d95d26337783dff87ef87b8ec02R295)? As mentioned in this [comment](https://github.com/DataDog/libdatadog/pull/1985/changes#r3313567074), it's ok to have an on/off switch in the builder, but we should either get these three fields from the builder or the agent, not both.

libdd-trace-utils/src/span/trace_utils.rs
```diff
@@ -60,6 +62,50 @@ where
     }
 }
 
+// FIXME: duplicated with super::get_root_span_index
+pub fn get_root_span_index_v4<T>(trace: &[Span<T>]) -> anyhow::Result<usize>
```
VianneyRuhlmann: I don't think this is specific to v4

libdd-trace-utils/src/span/trace_utils.rs
```diff
@@ -60,6 +62,50 @@ where
     }
 }
 
+// FIXME: duplicated with super::get_root_span_index
```
VianneyRuhlmann: Yes this not an ideal solution at the time we we're hoping to be able to migrate serverless from `pb::Span` to `Span` but protobuf serialization support has been stopping us from doing so. We can improve it in the future but it's fine to duplicate it for now.

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
```diff
@@ -0,0 +1,699 @@
...
+        let mut tag_regex_filters = Vec::new();
+        let mut tag_regex_key_filters = Vec::new();
+        for filter in filters {
+            let (key, value) = match filter.split_once(":") {
+                Some((key, value)) => (key, Some(value)),
+                None => (filter.as_ref(), None),
+            };
+
+            let value = match value {
+                Some(value) => match compile_anchored(value) {
```
VianneyRuhlmann: Why do we have to force it to be anchored ?

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
```diff
@@ -0,0 +1,699 @@
...
+/// Compile a regex anchored to the full string.
+fn compile_anchored(pattern: &str) -> Result<regex_engine::Regex, regex_engine::Error> {
+    regex_engine::Regex::new(&format!("^(?:{pattern})$"))
+}
+
+/// Returns `true` when `key` contains no regex metacharacters and can be used for a direct
+/// O(1) lookup.  `.` is intentionally treated as a literal (not a wildcard) in key patterns.
+fn is_literal_key(key: &str) -> bool {
+    !key.contains([
+        '*', '+', '?', '[', ']', '(', ')', '{', '}', '^', '$', ',', '\\',
```
VianneyRuhlmann: Should this include "|" ?

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
```diff
@@ -0,0 +1,699 @@
...
+    }
+}
+
+/// Compile a regex anchored to the full string.
+fn compile_anchored(pattern: &str) -> Result<regex_engine::Regex, regex_engine::Error> {
+    regex_engine::Regex::new(&format!("^(?:{pattern})$"))
+}
+
+/// Returns `true` when `key` contains no regex metacharacters and can be used for a direct
+/// O(1) lookup.  `.` is intentionally treated as a literal (not a wildcard) in key patterns.
```
VianneyRuhlmann: Regarding "." I don't see it in the spec, is it from the agent ? Also how is this enforced when matching an actual regex in the key

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
```diff
@@ -0,0 +1,699 @@
...
+    }
+
+    pub fn filter_traces<T: libdd_trace_utils::span::TraceData>(
+        &self,
+        traces: &mut Vec<Vec<libdd_trace_utils::span::v04::Span<T>>>,
+    ) {
+        let conf = self.conf.load();
+        traces.retain(|trace| {
+            let Ok(root_span_index) = get_root_span_index_v4(trace) else {
+                // FIXME: in this case it's a distributed trace ? Maybe we should remove the debug
```
VianneyRuhlmann: In the agent is this applied to based on root span only, or on local root span ?

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
```diff
@@ -0,0 +1,699 @@
...
+
+        if !conf
+            .require_key_regex
+            .iter()
+            .all(|filter| Self::check_tag_filter_with_normalization(filter, root_span))
+        {
+            return true;
+        }
+
+        if !conf.ignore_resources.is_empty() {
```
VianneyRuhlmann: In the [spec](https://datadoghq.atlassian.net/wiki/spaces/APM/pages/5758458333/v1.2.0#Filtering-Logic) the ressources is checked first.

libdd-data-pipeline/src/trace_exporter/builder.rs
ekump: I left a related [comment](https://github.com/DataDog/libdatadog/pull/1985/changes#r3313603767) elsewhere.

I agree, we should probably only rely on `/info`.

libdd-data-pipeline/src/trace_exporter/snapshots/libdd_data_pipeline__trace_exporter__single_threaded_tests__trace_filters.snap
Eldolfin: This is generated by `insta` snapshot tests. The file name corresponds to the absolute path to the test function with `::` replaced with double underscores. It corresponds to this test: https://github.com/DataDog/libdatadog/pull/1985/changes#diff-4495bd699a7cc3c377b8a943aae2c2ab9e48fa249f1e8016c5f4d45886982604R2178

I added this kind of test for end-to-end testing of the `TraceExporter` with trace filters without using the test-agent so we can have more control for the response we're getting in `/info`. It might be overkill for this simple feature but I think it could come in handy for these kind of end-to-end tests.

libdd-data-pipeline/src/trace_exporter/mod.rs
Eldolfin: I've already written a PR for adding system-test for this feature: https://github.com/DataDog/system-tests/pull/6952. It passes locally with `dd-trace-py` using this branch.

I will change it to only apply trace filters when CSS is enabled to reduce the impact.

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
Eldolfin: This is indeed not specified anywhere from what I've seen.

This behavior is different between the two implementation I've been using as reference:
- `dd-trace-dotnet` and the agent do not anchor
- `dd-trace-php` does anchor

If I had to choose I would say that it's more intuitive for the user for the patterns to automatically be anchored but in any-case this needs to be added to the spec and in the public documentation.

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
Eldolfin: Also not specified and not enforced when matching either. This is copied from `dd-trace-php` implementation but:
- `dd-trace-dotnet` doesn't do this separation, it always treats regex filters as key regex and value regex basically (which is less efficient)
- the agent doesn't support regex key

I think treating "`.`" as a literal does make sense given how common it is in tag keys but it might make it less intuitive because right now it's a literal only if there is no other regex metacharacters in which case it's actually a wildcard.

It needs to be added to the spec anyway but I see 2 options here:
- Never treat it as a literal
- Enforce it being a literal by escaping it in the final pattern if there is other metacharacters

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
Eldolfin: Yes. Needs to be fixed in `dd-trace-php` too

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
Eldolfin: Honestly not sure. Seems like traces are filtered chunk by chunk in the agent, it doesn't seem to wait for the whole distributed trace to arrive.

Needs to be double checked.

Eldolfin:
Starting to think another RFC is needed for the CSS spec because implementation differs quite a bit between existing implementations already: dd-trace-php, dd-trace-dotnet and the agent

But trace filter is not really linked to CSS: it's a requirement but it can be enabled separately so it could maybe live in another spec ?

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
VianneyRuhlmann: The trace filters is an agent feature that we are porting to the tracer to make CSS compatible with this feature. Therefore the only implementation reference is the agent and our implementation should be bug for bug compatible with the agent.

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
VianneyRuhlmann: > the agent doesn't support regex key

Then we shouldn't support regex key and update the CSS spec to be aligned with the agent

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
bwoebi: Correct :-D

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
bwoebi: Yes, that's an optimization, you basically always use `.*` or something similar.

Otherwise any key which includes a dot will need to do a full-scan match against every single tag.
I found that for any _practical_ purpose, this is a fine optimization to do.

Not doing this optimization clearly shows negatively up in performance terms for trivial value-only regexes otherwise.

But indeed, not supporting it at all is probably better :-D

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
bwoebi: This is just the impression I got from the example in the spec:
```
  "filter_tags_regex": {
    "reject": ["version:.*-beta", "experimental_.*"]
  },
```
If we were not to anchor, the leading or trailing `.*` would be irrelevant.

bwoebi: Also, it would make it surprising that suffixes are not filtered. e.g. `db[0-9]+` would match `db1`, but also `db1-staging`

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
Eldolfin: Removed key regex so not relevant anymore

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
Eldolfin: I removed regex key to follow the agent behavior so this is not relevant anymore.

libdd-data-pipeline/src/trace_exporter/trace_filter.rs
Eldolfin: I removed automatic anchoring to follow the agent behavior

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -579,11 +589,12 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
 
         // Process stats computation and drop non-sampled (p0) chunks.
         // This must run before the OTLP path so that unsampled spans are not exported.
-        let dropped_p0_stats = stats::process_traces_for_stats(
+        let dropped_stats = stats::process_traces_for_stats(
```
VianneyRuhlmann, line 619: This is not a comment for this PR just a remark for future improvement. We should be able to submit telemetry from within the stats module rather than having to pass it to the `send_traces_with_telemetry`

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -2347,6 +2361,187 @@ mod single_threaded_tests {
...
+                    meta: HashMap::from_iter([("my_require_tag".into(), "true".into())]),
+                    ..Default::default()
+                }],
+            ])
+            .as_ref(),
+        );
+        assert!(result.is_err());
+
+        // Wait for the stats worker to be active before shutting down to avoid potential flaky
+        // tests on CI where we shutdown before the stats worker had time to start
```
VianneyRuhlmann: Why do you want to wait for it after sending ? If the trace filters depend on CSS being enabled you need to wait for stats to be enabled before calling `send`

libdd-data-pipeline/src/trace_exporter/snapshots/libdd_data_pipeline__trace_exporter__single_threaded_tests__trace_filters.snap
VianneyRuhlmann: I think the right solution would be to update the test agent to support configuring trace filters in `/info`

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -290,12 +291,17 @@ pub(crate) fn process_traces_for_stats<T: libdd_trace_utils::span::TraceData>(
...
+    trace_filterer: &TraceFilterer,
+) -> libdd_trace_utils::span::trace_utils::DroppedStats {
     let status = client_side_stats.load();
     if let StatsComputationStatus::Enabled {
         stats_concentrator, ..
     } = &**status
     {
+        // FIXME: when client_computed_top_level is true, looking twice for the root span here and
+        // just below in compute_top_level_span is inefficient
+        let dropped_by_trace_filter = trace_filterer.filter_traces(traces);
```
VianneyRuhlmann: Most root span should have `parent-id=0` so I guess it's fine to look for it twice. This should be partially addressed by v1 which promotes some parameters to be trace chunk wide.

libdd-data-pipeline/src/trace_exporter/snapshots/libdd_data_pipeline__trace_exporter__single_threaded_tests__trace_filters.snap
Eldolfin: I agree, I added that kind of test as an experiment because I tought it was simpler to not have to launch a docker container for some tests but it does add code complexity. I'll try to make a PR on the test agent and update here once it's merged

libdd-data-pipeline/src/trace_exporter/mod.rs
Eldolfin: I'll try to move it up and see if it works, weird indeed 👍🏽

Eldolfin: Just remembered why: before calling `exporter.send(traces)` the exporter never calls `check_agent_info()` which is responsible for updating the `client_side_stats.status` based on the cached `/info` values.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -574,6 +582,10 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
         mut traces: Vec<Vec<Span<T>>>,
     ) -> Result<AgentResponse, TraceExporterError> {
         let mut header_tags: TracerHeaderTags = self.metadata.borrow().into();
+        // FIXME: when client_computed_top_level is true, looking twice for the root span here is
+        // inefficient and just below in process_traces_for_stats.
+        // Also, only do it when css is on
+        self.trace_filterer.filter_traces(&mut traces);
```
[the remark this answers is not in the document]
Eldolfin: Intentional

libdd-data-pipeline/src/trace_exporter/mod.rs
Eldolfin: Changed to only apply when CSS is turned on

libdd-trace-utils/src/trace_filter.rs
```diff
@@ -0,0 +1,716 @@
...
+    fn should_drop<'a>(&self, root_span: &'a impl Span<'a>) -> bool {
+        if !self.ignore_resources.is_empty() {
+            let span_resource = Span::resource(root_span);
+            // Normalization
+            let span_resource = if span_resource.is_empty() {
+                let span_name = root_span.name();
+                debug!(
+                    ?span_name,
+                    "Trace filter fixing malformed trace. Resource is empty so using name instead"
+                );
```
bwoebi: This isn't quite _malformed_ at this stage. Trace normalization (setting resource to name and such) happens later.

libdd-trace-utils/src/trace_filter.rs
```diff
@@ -0,0 +1,716 @@
...
+    /// Applies a subset of trace normalization logic from `libdd-trace-normalization` before
+    /// checking.
+    // 1. Resource filtering: If the root span's resource name matches any pattern in
+    //    ignore_resources, reject the trace.
+    // 2. Reject filtering: If any tag on the root span matches filters in filter_tags.reject or
+    //    filter_tags_regex.reject, reject the trace.
+    // 3. Require filtering: If filter_tags.require or filter_tags_regex.require contain any
+    //    filters, all of them must match tags on the root span. If any required filter doesn't
+    //    match, reject the trace.
+    fn should_drop<'a>(&self, root_span: &'a impl Span<'a>) -> bool {
```
bwoebi: Can you please make this pub? I need this, not filter_traces (having individual root spans I want to filter).

bwoebi: filter_traces makes sense when you have an already serialized batch of data where the root span is just one span of many. But not when they're still linked lists or some native data structure given by the tracers.

libdd-data-pipeline/src/trace_exporter/mod.rs
Eldolfin: test removed

VianneyRuhlmann (approved):

libdd-trace-utils/src/trace_filter.rs
VianneyRuhlmann: Maybe the logic to get root span and drop the trace can be handled by each caller and the TraceFilter only returns the decision for a given root span. This allow the Span trait to stay minimal.

bwoebi: @VianneyRuhlmann Yep, that's what I'd propose.

Eldolfin: This makes more sense I agree. I fixed it by making `filter_traces(traces)` only work for traces with v04 span. 👍🏽

PR #1986, opened 2026-05-14
fix(shared-runtime)!: fork recovery, shutdown, runtime_handle

iunanua:
## Summary

- **Fork recovery (bug):** `after_fork_parent` / `after_fork_child` used `?` inside the worker restart loop. A single `PausableWorker::InvalidState` aborted the entire batch, silencing every other component — critical in the fork child. Now log-and-continue, mirroring `before_fork`.
- **Post-shutdown spawn (bug):** `spawn_worker` after `shutdown` silently registered a worker that would never run. Added a `shutdown: AtomicBool` flag, set in `shutdown` / `shutdown_async`, checked in `spawn_worker` (native + wasm) under the workers lock so the during-shutdown race is also covered. New `SharedRuntimeError::AlreadyShutdown` variant + FFI mapping.
- **`runtime_handle()` (API design):** The method handed callers an owned tokio `Handle`, bypassing every fork-safety guarantee the crate exists to provide (tasks spawned via the handle aren't paused before fork, restarted after fork, or shut down). Replaced with `with_runtime_context(closure)`, which scopes the entered runtime to a synchronous closure. The lone production caller in `libdd-data-pipeline` is migrated.

### DIFF
--- a/libdd-shared-runtime-ffi/src/shared_runtime.rs
+++ b/libdd-shared-runtime-ffi/src/shared_runtime.rs
@@ -8,33 +8,26 @@ use std::ptr::NonNull;
 use std::sync::Arc;
 
 /// Error codes for SharedRuntime FFI operations.
-///
-/// # ABI stability
-/// Discriminants are pinned explicitly. The numeric values are part of the
-/// C ABI: existing variants must never be renumbered or reused, and new
-/// variants must be appended with a fresh value. Inserting a variant in
-/// the middle of the list silently misclassifies errors on any caller
-/// compiled against an older header.
 #[repr(C)]
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub enum SharedRuntimeErrorCode {
     /// Invalid argument provided (e.g. null handle).
-    InvalidArgument = 0,
+    InvalidArgument,
     /// The runtime is not available or in an invalid state.
-    RuntimeUnavailable = 1,
+    RuntimeUnavailable,
     /// Failed to acquire a lock on internal state.
-    LockFailed = 2,
+    LockFailed,
     /// A worker operation failed.
-    WorkerError = 3,
+    WorkerError,
     /// Failed to create the tokio runtime.
-    RuntimeCreation = 4,
+    RuntimeCreation,
     /// Shutdown timed out.
-    ShutdownTimedOut = 5,
+    ShutdownTimedOut,
     /// An unexpected panic occurred inside the FFI call.
     #[cfg(feature = "catch_panic")]
-    Panic = 6,
+    Panic,
     /// Operation rejected because the runtime has already been shut down.
-    AlreadyShutdown = 7,
+    AlreadyShutdown,
 }
 
 /// Error returned by SharedRuntime FFI functions.
--- a/libdd-shared-runtime-ffi/src/shared_runtime.rs
+++ b/libdd-shared-runtime-ffi/src/shared_runtime.rs
@@ -8,26 +8,33 @@ use std::ptr::NonNull;
 use std::sync::Arc;
 
 /// Error codes for SharedRuntime FFI operations.
+///
+/// # ABI stability
+/// Discriminants are pinned explicitly. The numeric values are part of the
+/// C ABI: existing variants must never be renumbered or reused, and new
+/// variants must be appended with a fresh value. Inserting a variant in
+/// the middle of the list silently misclassifies errors on any caller
+/// compiled against an older header.
 #[repr(C)]
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub enum SharedRuntimeErrorCode {
     /// Invalid argument provided (e.g. null handle).
-    InvalidArgument,
+    InvalidArgument = 0,
     /// The runtime is not available or in an invalid state.
-    RuntimeUnavailable,
-    /// Operation rejected because the runtime has already been shut down.
-    AlreadyShutdown,
+    RuntimeUnavailable = 1,
     /// Failed to acquire a lock on internal state.
-    LockFailed,
+    LockFailed = 2,
     /// A worker operation failed.
-    WorkerError,
+    WorkerError = 3,
     /// Failed to create the tokio runtime.
-    RuntimeCreation,
+    RuntimeCreation = 4,
     /// Shutdown timed out.
-    ShutdownTimedOut,
+    ShutdownTimedOut = 5,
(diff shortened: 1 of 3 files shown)


libdd-shared-runtime-ffi/src/shared_runtime.rs
```diff
@@ -8,24 +8,33 @@ use std::ptr::NonNull;
 use std::sync::Arc;
 
 /// Error codes for SharedRuntime FFI operations.
+///
+/// # ABI stability
+/// Discriminants are pinned explicitly. The numeric values are part of the
+/// C ABI: existing variants must never be renumbered or reused, and new
+/// variants must be appended with a fresh value. Inserting a variant in
+/// the middle of the list silently misclassifies errors on any caller
+/// compiled against an older header.
```
paullegranddc: The changes to the error codes is not necessary, users of libdatadog through the C API have no expectation of ABI backward comaptibility

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -89,6 +106,14 @@ mod native {
             let runtime_guard = self.runtime.lock_or_panic();
             let mut workers_guard = self.workers.lock_or_panic();
 
+            // Reject post-shutdown spawns under the workers lock — this is the
+            // same lock `shutdown_async` acquires before draining, so once
+            // shutdown wins the workers lock, every subsequent spawn observes
+            // the flag and bails instead of silently registering a dead worker.
```
paullegranddc: This (and further comments introduced by this PR) are way too verbose for something fairly obvious

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -42,21 +41,39 @@ mod native {
...
+        /// [`crate::Worker`] via [`Self::spawn_worker`] instead.
         ///
         /// # Errors
-        /// Returns [`SharedRuntimeError::RuntimeUnavailable`] if the runtime has been shut down.
-        pub fn runtime_handle(&self) -> Result<tokio::runtime::Handle, SharedRuntimeError> {
-            Ok(self
+        /// Returns [`SharedRuntimeError::RuntimeUnavailable`] if the runtime
+        /// has been shut down or is in a fork window.
+        pub fn with_runtime_context<F, T>(&self, f: F) -> Result<T, SharedRuntimeError>
+        where
```
paullegranddc, line 65: I am not sure this is any safer than before 🤔 
Maybe a tiny bit nicer because it moves the wasm cfg to the runtime

paullegranddc (commented):
Overall I don't really know what motivates this PR because the bug it fixes (start early exit), isn't really likely to happen. This is because to get `PausableWorker::start` to return an error, you need to be in an invalid state.

This invalid state is caused the join handle in a previous running state to have returned an error.
This is caused by one of two things:
* The task was cancelled, which never happens
* The task panics, which could happen although I don't think we have seen this happen so far.

If one of the task has panicked, we are in a pretty bad situation so I don't know if continuing with init is the best thing to do... although, what else is there to do 🤔

libdd-shared-runtime/src/shared_runtime/mod.rs
iunanua: `runtime_handle()` handed callers an owned, long-lived capability that lets them bypass every safety guarantee. An spawned task from that handle is not in the workers list, is not paused by before_fork, etc so  could it lead to confusion?

libdd-shared-runtime/src/shared_runtime/mod.rs
iunanua: agree

libdd-shared-runtime-ffi/src/shared_runtime.rs
iunanua: error codes removed

iunanua:
> Overall I don't really know what motivates this PR because the bug it fixes (start early exit), isn't really likely to happen. This is because to get `PausableWorker::start` to return an error, you need to be in an invalid state.
> 
> This invalid state is caused the join handle in a previous running state to have returned an error. This is caused by one of two things:
> 
> * The task was cancelled, which never happens
> * The task panics, which could happen although I don't think we have seen this happen so far.
> 
> If one of the task has panicked, we are in a pretty bad situation so I don't know if continuing with init is the best thing to do... although, what else is there to do 🤔

Some claude points (I'm just a humble messenger):

1. The entry points to InvalidState are broader than "cancellation or panic":
- `shutdown(Some(timeout))` cancelling `pause()` mid-flight. shutdown wraps `shutdown_async()` in `tokio::time::timeout`. If the timeout fires while a worker is mid-pause, `pause()`'s future is dropped — and because `pause()` does `mem::replace(self, InvalidState)` before awaiting the join handle, dropping the
  future strands the worker in `InvalidState`. Same thing if `pause()` is called inside any `select!` / cancellation scope.
- `on_pause()` panicking. This is a Worker trait hook user code runs during `before_fork`. The state machine is Running → InvalidState → (on_pause) → Paused; a panic in `on_pause` leaves the worker stuck in the transient `InvalidState`. The bug then surfaces on the very next `after_fork_*`.

2. The fix is mostly about removing an asymmetry, not patching a hot path:

`before_fork` already does exactly what `after_fork_*` now does — log the failed worker and keep going:
So the pre-fix code was inconsistent: a panic on the pause path during `before_fork` was tolerated, but a panic on the restart path during `after_fork_*` aborted recovery for every other worker.

VianneyRuhlmann:
- I agree with the after_fork fix.
- The runtime_handle has been removed as it was not meant to be public.
- The shutdown mechanic seems a bit complicated for something that is very unlikely to happen since the shared runtime is supposed to be shutdown when the tracer is also shutting down. Maybe it's easier to move this to a new PR since there's a lot of conflicts. I can handle it if you want.

iunanua:
> * I agree with the after_fork fix.
> * The runtime_handle has been removed as it was not meant to be public.
> * The shutdown mechanic seems a bit complicated for something that is very unlikely to happen since the shared runtime is supposed to be shutdown when the tracer is also shutting down. Maybe it's easier to move this to a new PR since there's a lot of conflicts. I can handle it if you want.

Go ahead!
tbh, I had forgotten about this PR

Thkx @VianneyRuhlmann

VianneyRuhlmann:
Close in favor of https://github.com/DataDog/libdatadog/pull/2262

PR #1822, opened 2026-03-30
feat(telemetry): add session id support to trace export

mabdinur:
## What does this PR do?

Adds optional instrumentation session IDs for telemetry on the **trace exporter** path: FFI (`TelemetryClientConfig` + wiring), Rust (`TelemetryInstrumentationSessions` + builder), C example, and a combined telemetry test for runtime id + session headers.

## Motivation

Allow trace-exporter telemetry to send `dd-session-id` / root / parent consistently with the rest of telemetry configuration.

## Additional notes

- `TelemetryConfig` is unchanged.
- Not used by **dd-trace-php** (that path uses the telemetry worker builder).

## How to test

```bash
cargo test -p libdd-data-pipeline telemetry::tests::runtime_id_and_session_headers_test --lib
cargo test -p libdd-data-pipeline-ffi trace_exporter::tests::config_telemetry_test --lib
```

Build and run `examples/ffi/trace_exporter.c` with the updated `TelemetryClientConfig` fields.

PR #1989, opened 2026-05-14
chore: bump libdd-capabilities-impl to 1.1.0 and libdd-data-pipeline to 3.1.0

mabdinur:
## What does this PR do?

Bumps two crates that added public APIs in [#1822](https://github.com/DataDog/libdatadog/pull/1822) without incrementing their version numbers. Without a version bump these APIs cannot be published to crates.io (you cannot republish the same version), blocking downstream consumers from using them without a `[patch.crates-io]` workaround.

| Crate | Old | New | New APIs |
|---|---|---|---|
| `libdd-capabilities-impl` | `1.0.0` | `1.1.0` | `NativeCapabilities` |
| `libdd-data-pipeline` | `3.0.1` | `3.1.0` | `telemetry` feature, `TelemetryInstrumentationSessions`, `TraceExporter<C>` |

`libdd-telemetry` is already at `5.0.0` on main (unpublished), so no bump is needed there.

## Motivation

`dd-trace-rs` [#216](https://github.com/DataDog/dd-trace-rs/pull/216) implements instrumentation session ID propagation using these APIs. Until this PR lands and these crates are published, `dd-trace-rs` must carry a `[patch.crates-io]` block in every workspace that consumes it — including `DataDog/system-tests` for parametric tests.

## After merge

Once this PR is merged, publish `libdd-capabilities-impl 1.1.0` and `libdd-data-pipeline 3.1.0` to crates.io (along with the already-bumped `libdd-telemetry 5.0.0`). Then `dd-trace-rs` PR #216 can be updated to reference the published versions and drop `[patch.crates-io]` entirely.

### DIFF
--- a/libdd-capabilities-impl/Cargo.toml
+++ b/libdd-capabilities-impl/Cargo.toml
@@ -3,7 +3,7 @@
 
 [package]
 name = "libdd-capabilities-impl"
-version = "1.0.0"
+version = "1.1.0"
 description = "Native implementations of libdd-capabilities traits"
 homepage = "https://github.com/DataDog/libdatadog/tree/main/libdd-capabilities-impl"
 repository = "https://github.com/DataDog/libdatadog/tree/main/libdd-capabilities-impl"
--- a/libdd-data-pipeline/Cargo.toml
+++ b/libdd-data-pipeline/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libdd-data-pipeline"
-version= "3.0.1"
+version= "3.1.0"
 description = "Trace exporter package allowing sending data from datadog SDKs to the Trace Agent."
 homepage = "https://github.com/DataDog/libdatadog/tree/main/libdd-data-pipeline"
 repository = "https://github.com/DataDog/libdatadog/tree/main/libdd-data-pipeline"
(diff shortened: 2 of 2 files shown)


ekump (changes requested):
I don't think you want to do this. Versions get updated automatically as part of the crate publishing workflow. @iunanua, would you mind helping Munir get these crates published?

mabdinur:
> * Test FFI / FFI alpine-build via docker bake (pull_request)

@ekump Thanks 🙏 . I'll move this to draft

iunanua:
:wave: it turns out that:
- [this](https://github.com/DataDog/libdatadog/actions/runs/25876809058/job/76046228068?pr=1989) check should had failed because you are modifying Cargo.toml versions but it didn't :rage: 
- the check also was successful rather than failing when libdd-telemetry version was bumped to 5.0.0 :rage: 

so, I have discovered why the check is failing and I have a fix [here](https://github.com/DataDog/libdatadog/pull/1994)
Also I have prepared a PR to restore `libdd-telemetry` version to 4.0.0 [here](https://github.com/DataDog/libdatadog/pull/1993) After this PR is merged we can continue with the "normal" release process :registered:

This PR should be closed

mabdinur:
Awesome! Thanks @iunanua 🙇

iunanua:
> Awesome! Thanks @iunanua 🙇

@mabdinur all the crates you need have been published https://github.com/DataDog/libdatadog/tags
Let me know if there is something wrong or missing.

PR #1990, opened 2026-05-15
refactor(profiling): reduce anyhow, avoid panics

morrisonlevi:
# What does this PR do?

Mostly this introduces new Error structs instead of using anyhow for some internal operations (not public API changes, unless you count changing exact error messages to be a breaking change). 

This also avoids some panics. We do more `try_reserve*` stuff, and also remove some hopefully unreachable panics, such as:

```diff
- let labels = &mut extended_label_sets[sample.labels.to_raw_id()];
+ let labels = extended_label_sets
+     .get_mut(sample.labels.to_offset())
+     .ok_or(ProfileDataError::InvalidLabelSetId)?;
```

It also extracts a helper function `expand_label_set` to fallibly convert a `&LabelSet` to a `Vec<Label>`, add endpoint info, and reserve the spare memory for the timestamp label which gets added later. (Note that it's added later because we have many observations with the same sets of labels, but the timestamp label will differ)

# Motivation

We've also seen rare panics in some of the profiling FFI functions in crash reports. On Windows we don't have the panic message so I'm not exactly sure why. So I'm working on avoiding panics and ratcheting up the code quality so we have fewer failures all around.

Additionally, anyhow collects a call stack on failure and it's not cheap, plus it allocates for this. We don't want either.

# Additional Notes

Some changes like moving constants are in preparation for another PR, but the changes are fine independent of that. This is also why changing the internal functions from anyhow to `FfiSafeErrorMessage` structs will help me: in that PR, we'll have `ddog_prof_Profile_serialize2` which doesn't use `anyhow`, and reuses these internal helpers.

This is also blocked on #1988 merging first.

# How to test the change?

Aside from the exact error messages displayed and possibly having fewer panics, this should test and act the same!

### DIFF
--- a/libdd-profiling-ffi/src/profiles/datatypes.rs
+++ b/libdd-profiling-ffi/src/profiles/datatypes.rs
@@ -5,11 +5,13 @@ use crate::string_storage::{get_inner_string_storage, ManagedStringStorage};
 use crate::{ensure_non_null_out_parameter, ArcHandle, ProfileError, ProfileStatus};
 use anyhow::Context;
 use function_name::named;
+use libdd_common::error::FfiSafeErrorMessage;
 use libdd_common_ffi::slice::{AsBytes, ByteSlice, CharSlice, Slice};
 use libdd_common_ffi::{wrap_with_ffi_result, Error, Handle, Timespec, ToInner};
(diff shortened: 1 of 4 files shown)


morrisonlevi:
Meh not focused enough, closing for more focused PRs.

PR #1991, opened 2026-05-15
refactor(profiling): try_reserve* more in encoding

morrisonlevi:
# What does this PR do?

Uses `Vec::try_reserve_exact` in 3 places in `Profile::encode` and extracts `expand_label_set` as a helper function.

# Motivation

Sometimes we get crash reports from panics crossing FFI boundaries in libdatadog, so I'm working on reducing the number of places that panics can happen.

# Additional Notes

I want to remove `anyhow::Result` but that's for another PR.

# How to test the change?

Tests the same, it just won't panic if you OOM on very specific allocations.

### DIFF
--- a/libdd-profiling/src/internal/profile/mod.rs
+++ b/libdd-profiling/src/internal/profile/mod.rs
@@ -555,7 +555,8 @@ impl Profile {
             self.upscaling_rules.upscale_values(&mut values, labels);
 
             let mut pprof_labels: Vec<_> = Vec::new();
-            pprof_labels.try_reserve_exact(labels.len())?;
+            // + 1 for the timestamp (which hasn't ben pushed yet)
+            pprof_labels.try_reserve_exact(labels.len() + 1)?;
 
             // Try not to fail between labels.push and labels.pop, which would
             // leave the push.
--- a/libdd-profiling/src/internal/profile/mod.rs
+++ b/libdd-profiling/src/internal/profile/mod.rs
@@ -496,6 +496,22 @@ impl Profile {
         Ok(encoded_profile)
     }
 
+    /// Resolves label ids into labels and appends the endpoint label, leaving
+    /// one spare slot for a sample timestamp label.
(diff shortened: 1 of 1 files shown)


libdd-profiling/src/internal/profile/mod.rs
```diff
@@ -547,11 +554,16 @@ impl Profile {
             self.check_location_ids_are_valid(&location_ids, self.locations.len())?;
             self.upscaling_rules.upscale_values(&mut values, labels);
 
-            // Use the extra slot in the labels vector to store the timestamp without any reallocs.
+            let mut pprof_labels: Vec<_> = Vec::new();
+            pprof_labels.try_reserve_exact(labels.len())?;
```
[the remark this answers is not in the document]
morrisonlevi: Yes, this slipped in when I realized I couldn't error between push and pop, and moved the reserve up.

libdd-profiling/src/internal/profile/mod.rs
```diff
@@ -496,6 +496,22 @@ impl Profile {
         Ok(encoded_profile)
     }
 
+    /// Resolves label ids into labels and appends the endpoint label, leaving
+    /// one spare slot for a sample timestamp label.
+    fn expand_label_set(&self, label_set: &LabelSet) -> anyhow::Result<Vec<Label>> {
+        let endpoint_label = self.get_endpoint_for_label_set(label_set)?;
+        // +1 for the timestamp label
+        let mut labels =
+            Vec::with_capacity(label_set.len() + usize::from(endpoint_label.is_some()) + 1);
```
gleocadie: can `Vec::with_capacity` panick ? like when no more memory is left ? is there some equivalent to the `malloc` pattern, where returning `null` means `no more memory` ?

morrisonlevi: Yes, it can panic! The AI switched this from `try_reserve_exact` and I missed it in self-review, thanks!

gleocadie (approved):
LGTM

PR #2001, opened 2026-05-15
chore(ci): use the large runners for ubuntu as we keep running out of disk space

ekump:
# What does this PR do?

Since https://github.com/DataDog/libdatadog/pull/1928 we've been intermittently running out of disk space on our GH runners for tests. This PR bumps the size of our ubuntu runners, which includes more disk space and should solve this issue.  

# Motivation

@morrisonlevi was seeing consistent failures on the MQ when running the full test suite on this PR: https://github.com/DataDog/libdatadog/pull/1988

# Additional Notes

# How to test the change?


morrisonlevi (approved):
Approved, and thanks!

PR #2002, opened 2026-05-15
refactor(profiling): avoid panic on out-of-bounds label set id

morrisonlevi:
# What does this PR do?

Avoids a panic on bounds access when the label set id is out-of-bounds. This shouldn't happen, and is a bug. If it does happen though we shouldn't panic: we should fail gracefully instead.

# Motivation

I'm working to remove panics. For now I'm hunting places which already return a `Result` so they can be made without breaking compatibility.

# Additional Notes

Nope

# How to test the change?

Tests as normal, regression test was added.


libdd-profiling/src/internal/profile/mod.rs
```diff
@@ -1068,6 +1069,14 @@ impl Profile {
     }
 }
 
+#[cold]
+fn oob_label_set() -> io::Error {
+    io::Error::new(
+        io::ErrorKind::InvalidData,
+        "out-of-bounds label set id found during serialization",
+    )
+}
```
danielsn, line 1078: whats the advantage of naming this function rather than using a closure

libdd-profiling/src/internal/profile/mod.rs
```diff
@@ -544,7 +544,8 @@ impl Profile {
 
         let iter = std::mem::take(&mut self.observations).try_into_iter()?;
         for (sample, timestamp, mut values) in iter {
-            let labels = &mut extended_label_sets[sample.labels.to_raw_id()];
+            let off = sample.labels.to_offset();
```
danielsn, line 547: what is the difference between offset and raw id?

libdd-profiling/src/internal/profile/mod.rs
morrisonlevi: In this specific case, to give it a `#[cold]` attribute so that the engine realizes it should be cold. It should never be called (only if there are bugs), so it's as cold as you can get!

Also, stylistically, it looked bad inline (because the formatter insists on formatting it across 4 lines).

libdd-profiling/src/internal/profile/mod.rs
morrisonlevi: In these specific cases, nothing, because `LabelSetId` uses `usize` as the `RawId` type and there's no shifting. I think `to_offset` is more obvious about what we're doing, which is why I changed it.

For other types, like string ids, the `to_raw_id()` might return an `i64`, but `to_offset()` is always `usize`.

danielsn (approved):

PR #2005, opened 2026-05-18
feat!: nested tokio runtime support, async API surface, drain on shutdown, flush on drop

Aaalibaba42:
# What?

Make `TraceBuffer`, `SharedRuntime`, and `TraceExporter` safe to use from inside a
host tokio runtime, expose an async API surface alongside the existing sync facade,
add a synchronous `flush_and_wait`, drain pending spans on shutdown, and fire a
best-effort flush on `Drop`.

# Why?

- `SharedRuntime::block_on` and `SharedRuntime::shutdown` panicked with
  `Cannot start a runtime from within a runtime` when called from inside an
  existing tokio context (e.g. the Rust tracer embedded in an async application).
  The original "block_on on a scoped OS thread" workaround was reverted in favor
  of a proper async API surface — the OS-thread trick still allocated a second
  tokio runtime per call and could not be wired into a host runtime's lifecycle.
- The trace exporter and stats workers held hidden `block_on` calls deep in
  `check_agent_info`, `stop_stats_computation`, and `TraceExporterBuilder::build`,
  which propagated the nested-runtime panic to every async caller.
- On shutdown, the final partial batch of spans was silently dropped because
  nothing drained the sender's buffer after the worker loop stopped.
- Dropping a `TraceBuffer` without an explicit flush also lost any buffered spans.

# How?

## `SharedRuntime` (`libdd-shared-runtime`)

- **Borrowed mode.** Added a `RuntimeBacking::{Owned(Arc<Runtime>), Borrowed(Handle)}`
  enum and a new constructor `SharedRuntime::from_handle(Handle)` so callers
  that already own a tokio runtime can share it instead of spawning a second one.
  Owned mode is unchanged; borrowed mode gives up fork-safety in exchange for
  letting `Drop` / `shutdown` work cleanly from a host worker thread without
  `block_on`. `block_on`, sync `shutdown`, and `before_fork` return a typed
  error in borrowed mode (`SharedRuntimeError::ForkUnsupportedInBorrowedMode`,
  `SyncShutdownNotSupportedInBorrowedMode`, or
  `io::Error(ErrorKind::Unsupported)` for `block_on`) instead of panicking.
- **Non-blocking shutdown.** Added `trigger_shutdown_signal()` (snapshots the
  worker set, spawns shutdown tasks on the underlying runtime, bumps a tracked
  expected count) and `wait_shutdown_done(timeout)` (parks on a `Condvar` until
  the tracked count is reached). Pairs with sync `Drop` impls that must wait
  for worker completion without `block_on`.
- **Panic-safe internals.** `Drop`, `trigger_shutdown_signal`, `wait_shutdown_done`,
  and `is_borrowed` were converted to poison-tolerant lock acquisition — a
  Drop impl must never panic, and the public shutdown surface should not
  either. Internal counter bumps in spawned shutdown tasks degrade gracefully
  on `PoisonError` (`poison.into_inner()`) so a single task panic cannot
  deadlock `wait_shutdown_done`.
- **FFI.** `ddog_shared_runtime_before_fork` now propagates the new error
  variants through `SharedRuntimeFFIError`, mapped to a new
  `SharedRuntimeErrorCode::NotSupportedInBorrowedMode` discriminator.

## `libdd-data-pipeline` async API surface

- **Builder.** `TraceExporterBuilder::build` is now a thin wrapper around a
  new `build_async`. The sync wrapper is gated `#[cfg(not(target_arch = "wasm32"))]`
  since it relies on `SharedRuntime::block_on`.
- **Exporter.** `TraceExporter::send`, `send_trace_chunks`, and `shutdown` are
  now sync wrappers over new `send_async`, `send_trace_chunks_async`, and
  `shutdown_async` counterparts; the sync wrappers are similarly gated for
  wasm32. The internal `check_agent_info`, `stop_stats_computation`, and
  `handle_stats_enabled` are now `async` end-to-end — every internal `block_on`
  was deleted. `ArcSwap::load` was changed to `load_full` where needed so the
  resulting `Send` future survives `await` points.

## `TraceBuffer` (`libdd-data-pipeline`)

- **`flush_and_wait(timeout)`.** Triggers a flush, captures the batch
  generation, and parks on a `Condvar` until the worker has processed it.
  Short-circuits on empty batches.
- **`Receiver::drain()`.** Synchronously pulls remaining chunks without
  waiting for a flush trigger; `TraceExporterWorker::shutdown` calls it to
  export any leftover spans before tearing down.
- **`Drop for TraceBuffer<T>`.** Fires a non-blocking flush-notify so the
  worker exports pending chunks instead of losing them. Errors are
  intentionally swallowed — if the runtime is already gone there is nothing
  useful to do.

# Tests

- Regression coverage for `block_on` / `shutdown` / `drop` invoked from inside
  `#[tokio::test(flavor = "multi_thread")]`, asserting the borrowed-runtime
  path.
- New `SharedRuntime` tests: `test_from_handle_borrowed_shutdown_wait`,
  `test_borrowed_mode_unsupported_apis`, `test_wait_shutdown_done_no_workers`.
- All 11 trace_buffer tests pass (`test_drop_inside_tokio_runtime`,
  `test_drop_triggers_flush`, `test_flush_and_wait`,
  `test_shutdown_drains_pending_batch`, etc.).

# Breaking changes

- `SharedRuntime::before_fork`, `after_fork_parent`, and `after_fork_child`
  now return `Result<(), SharedRuntimeError>` instead of `()`.
- `SharedRuntime::block_on` now returns `Result<F::Output, io::Error>` (was
  infallible).
- The C ABI of `ddog_shared_runtime_before_fork` now returns an optional
  boxed `SharedRuntimeFFIError`.

# Additional notes

- Sync facade (`build`, `send`, `send_trace_chunks`, `shutdown`) is preserved
  for existing callers but now panics / returns
  `io::Error(ErrorKind::Unsupported)` when invoked on a borrowed `SharedRuntime`.
  A `sync-api` cargo feature gating these methods was prototyped and reverted
  as premature; see the explanatory comment block in
  `libdd-data-pipeline/Cargo.toml` for the rationale and a pointer for future
  consumers who do want a build that statically forbids the sync facade.

### DIFF
--- a/libdd-shared-runtime/src/shared_runtime/mod.rs
+++ b/libdd-shared-runtime/src/shared_runtime/mod.rs
@@ -209,33 +209,19 @@ mod native {
         /// If the runtime is not available (e.g. after calling before_fork), a temporary
         /// single-threaded runtime is used.
         ///
-        /// Safe to call from inside another tokio runtime: when invoked from a thread that is
-        /// already driving a tokio scheduler, the future is run on a temporary OS thread via
-        /// [`std::thread::scope`] to avoid the "Cannot start a runtime from within a runtime"
-        /// panic that `Runtime::block_on` raises in that situation.
-        ///
         /// Not available on wasm32 -- use async paths instead.
         ///
         /// # Errors
         /// Returns an error if it fails to create a fallback runtime.
-        pub fn block_on<F>(&self, f: F) -> Result<F::Output, io::Error>
-        where
-            F: std::future::Future + Send,
-            F::Output: Send,
-        {
+        pub fn block_on<F: std::future::Future>(&self, f: F) -> Result<F::Output, io::Error> {
             let runtime = match self.runtime.lock_or_panic().as_ref() {
                 None => Arc::new(Builder::new_current_thread().enable_all().build()?),
                 Some(runtime) => runtime.clone(),
             };
-            Ok(block_on_outside_runtime(runtime, f))
+            Ok(runtime.block_on(f))
         }
 
         /// Shutdown the runtime and all workers synchronously with optional timeout.
-        ///
-        /// Safe to call from inside another tokio runtime (see [`Self::block_on`] for details).
-        /// When invoked from a tokio context, both the worker shutdown and the eventual
-        /// `Runtime` drop happen on a scoped OS thread; both require blocking and would
-        /// otherwise panic.
         ///
         /// Not available on wasm32 -- use [`shutdown_async`](Self::shutdown_async) instead.
         ///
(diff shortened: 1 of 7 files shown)


yannham (changes requested):
We've discussed this offline, and the consensus seemed to be that the actual issue of the runtime-in-the-runtime problem is that some functions of our API pretend to be async but actually use `block_on` under the hood. It seems one possible solution is to actually make them async, and to offer a customary blocking API at the top-level only. Then, from Rust we could only use the purely async API and not care about the shared runtime at all.

On the other hand, the current solution secretly spawns a thread and run a second runtime on the side, _while blocking the original executor's thread waiting for the answer_. This is a bit fishy, and we never want to block an async executor thread, as this will block a whole bunch of unrelated async tasks queued on this thread.

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -36,33 +37,102 @@ mod native {
...
+    /// - [`Self::Borrowed`]: a handle to a tokio runtime owned by the caller (typically a Rust
+    ///   application's host runtime). Used when the caller is itself an async tokio program and we
+    ///   must integrate with their executor instead of spinning up our own. Borrowed mode does
+    ///   **not** support fork hooks, sync `block_on`, or sync `shutdown` — callers should use the
+    ///   async/condvar shutdown paths instead.
+    #[derive(Debug, Clone)]
+    pub(super) enum RuntimeBacking {
+        Owned(Arc<Runtime>),
+        Borrowed(Handle),
+    }
```
paullegranddc, line 55: I know you have added this for dd-trace-rs but in general we don't want to us the same runtime as the calling app and always want to create a new one

Aaalibaba42: Creating a new one quickly leads to problems whenever we want to block/shutdown etc. That's why we discussed changing to borrow the parents' runtime iiuc. It's not a problem unique to dd-trace-rs

paullegranddc: Still, in dd-trace-rs I would prefer if we used our own runtime, spawning a thread to block on said runtime.
We don't know what customer are doing with their runtime, what load they put and so on....

Aaalibaba42:
This was split into #2064 and #2061, I keep this one open for now if I can rebase it on these and adapt the changes here to what's needed by dd-trace-rs.

PR #2011, opened 2026-05-18
feat(trace-utils)!: add from_string to span text

VianneyRuhlmann:
# What does this PR do?

Make the span fields mutable by requiring From<String> on SpanText and switching SpanSlice to use Cow<str>

# Motivation

Needed to apply normalization on the span, required by trace filters.

# Additional Notes

# How to test the change?


libdd-trace-utils/src/span/mod.rs
```diff
@@ -20,13 +20,13 @@ use std::{fmt, ptr};
 /// Trait representing the requirements for a type to be used as a Span "string" type.
 /// Note: Borrow<str> is not required by the derived traits, but allows to access HashMap elements
 /// from a static str and check if the string is empty.
-pub trait SpanText: Debug + Eq + Hash + Borrow<str> + Serialize + Default {
+pub trait SpanText: Debug + Eq + Hash + Borrow<str> + Serialize + Default + From<String> {
```
bantonsson, line 23: What does this trait bound add? I can't see that it's necessary.

VianneyRuhlmann: It allows to store new strings in the span e.g. an normalized/obfuscated version of a value. Otherwise SpanText instance can only be created from static str.

yannham (commented):
Just trying to understand the blast radius: who is using the now cow-based `SpanSlice` implementation in practice? Is it only for tests, or do others have to pay the cow tax? Maybe it's unavoidable, because any SDK has do deal with the filter/normalization issue anyway?

VianneyRuhlmann:
iirc we only use it in the trace_utils decoder when decoding from a slice which is done by python and is going to be used by ruby. All sdks may need to use normalization/obfuscation but not all customers will (for now only those using trace filters). I can add an extra trait `SpanOwnedData` to opt-in into the `CowSpan` when we need to apply normalization.

yannham (approved):
I see, thanks. I don't think it's worth the effort anyway, since we're moving towards native spans, and it's not clear that `Cow` has any measurable impact (it does double the size of `&str`, but whether that makes any difference...).

libdd-trace-utils/src/span/mod.rs
bantonsson: If that's the case, then maybe there should be code in the PR that needs it, since everything compiles fine without that trait bound.

paullegranddc, line 23: I would prefer a method on the trait rather than an extra trait bound.
In general I find it less less when I read code using `fn from` (or `.into`)

VianneyRuhlmann: It's needed by @Eldolfin work on the trace filters. It seemed cleaner to do this change separately to be able to discuss it

VianneyRuhlmann: discussed irl

Eldolfin (approved):
LGTM

PR #2013, opened 2026-05-19
fix(libdd-data-pipeline): cargo clippy fix with all lints

Eldolfin:
# What does this PR do?

Applies the restrictive workspace Clippy/Rust lint config and opts in `libdd-data-pipeline`, fixing the resulting lints without changing behavior.

# Motivation

Tighten lint coverage incrementally, crate by crate. [APMSP-3056]

# Additional Notes

# How to test the change?

CI passing is sufficient validation.

[APMSP-3056]: https://datadoghq.atlassian.net/browse/APMSP-3056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libdd-data-pipeline/src/agent_info/fetcher.rs
+++ b/libdd-data-pipeline/src/agent_info/fetcher.rs
@@ -59,7 +59,7 @@ async fn fetch_info_with_state_and_container_tags<C: HttpClientCapability + Slee
     Ok(FetchInfoStatus::NewState(info))
 }
 
-/// Fetch info from the given info_endpoint and compare its state to the current state hash.
+/// Fetch info from the given `info_endpoint` and compare its state to the current state hash.
 ///
 /// If the state hash is different from the current one:
 /// - Return a `FetchInfoStatus::NewState` of the info struct
@@ -104,7 +104,7 @@ pub async fn fetch_info<C: HttpClientCapability + SleepCapability>(
 
 /// Fetch and hash the response from the agent info endpoint.
 ///
-/// Returns a tuple of (state_hash, response_body_bytes, container_tags_hash).
+/// Returns a tuple of (`state_hash`, `response_body_bytes`, `container_tags_hash`).
 /// The hash is calculated using SHA256 to match the agent's calculation method.
 async fn fetch_and_hash_response<C: HttpClientCapability + SleepCapability>(
     info_endpoint: &Endpoint,
@@ -123,7 +123,7 @@ async fn fetch_and_hash_response<C: HttpClientCapability + SleepCapability>(
     let res = tokio::select! {
         biased;
         result = client.request(req) => result?,
-        _ = sleeper.sleep(timeout) => {
+        () = sleeper.sleep(timeout) => {
             return Err(anyhow!("Request to /info timed out after {:?}", timeout));
         }
     };
@@ -133,7 +133,7 @@ async fn fetch_and_hash_response<C: HttpClientCapability + SleepCapability>(
         .headers()
         .get("Datadog-Container-Tags-Hash")
         .and_then(|v| v.to_str().ok())
-        .map(|s| s.to_string());
+        .map(std::string::ToString::to_string);
 
     let body_data = res.into_body();
     let hash = format!("{:x}", Sha256::digest(&body_data));
@@ -141,7 +141,7 @@ async fn fetch_and_hash_response<C: HttpClientCapability + SleepCapability>(
     Ok((hash, body_data, container_tags_hash))
 }
 
-/// Fetch the info endpoint and update an ArcSwap keeping it up-to-date.
+/// Fetch the info endpoint and update an `ArcSwap` keeping it up-to-date.
 ///
 /// This type implements [`libdd_shared_runtime::Worker`] and is intended to be driven by a worker
 /// runner such as [`libdd_shared_runtime::SharedRuntime`].
@@ -205,9 +205,9 @@ impl<C: HttpClientCapability + SleepCapability> AgentInfoFetcher<C> {
     /// Return a new `AgentInfoFetcher` fetching the `info_endpoint` on each `refresh_interval`
     /// and updating the stored info.
     ///
-    /// Returns a tuple of (fetcher, trigger_component) where:
-    /// - `fetcher`: The AgentInfoFetcher to be run in a background task
-    /// - `response_observer`: The ResponseObserver component for checking HTTP responses
+    /// Returns a tuple of (fetcher, `trigger_component`) where:
+    /// - `fetcher`: The `AgentInfoFetcher` to be run in a background task
+    /// - `response_observer`: The `ResponseObserver` component for checking HTTP responses
     pub fn new(info_endpoint: Endpoint, refresh_interval: Duration) -> (Self, ResponseObserver) {
         // The trigger channel stores a single message to avoid multiple triggers.
         let (trigger_tx, trigger_rx) = mpsc::channel(1);
@@ -244,7 +244,7 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Wor
         if AGENT_INFO_CACHE.load().is_none() {
             return;
         }
-        self.trigger().await
+        self.trigger().await;
     }
 
     async fn trigger(&mut self) {
@@ -261,7 +261,7 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Wor
                         }
                     }
                     // Regular periodic fetch timer
-                    _ = sleeper.sleep(self.refresh_interval) => {}
+                    () = sleeper.sleep(self.refresh_interval) => {}
                 }
             }
             None => {
@@ -275,10 +275,10 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Wor
         // Release the IoStack waker stored in trigger_rx by waking the channel and drain the
         // message to avoid a spurious fetch on restart. If the channel is not empty then it has
         // already been waked.
-        if self.trigger_rx.as_ref().is_some_and(|rx| rx.is_empty()) {
+        if self.trigger_rx.as_ref().is_some_and(tokio::sync::mpsc::Receiver::is_empty) {
             let _ = self.trigger_tx.try_send(());
             self.drain();
-        };
+        }
     }
 
     async fn run(&mut self) {
@@ -306,7 +306,7 @@ impl<C: HttpClientCapability + SleepCapability> AgentInfoFetcher<C> {
                 AGENT_INFO_CACHE.store(Some(Arc::new(*new_info)));
             }
             Ok(FetchInfoStatus::SameState) => {
-                debug!("Agent info is up-to-date")
+                debug!("Agent info is up-to-date");
             }
             Err(err) => {
                 warn!(?err, "Error while fetching /info");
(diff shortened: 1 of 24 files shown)


libdd-data-pipeline/src/agent_info/fetcher.rs
```diff
@@ -34,6 +34,10 @@ pub enum FetchInfoStatus {
 /// If either the agent state hash or container tags hash is different from the current one:
 /// - Return a `FetchInfoStatus::NewState` of the info struct
 /// - Else return `FetchInfoStatus::SameState`
+#[allow(
```
yannham, line 37: If this is a clippy bug, I wonder if we shouldn't just disable it for the whole file and call it a day... that will be less place to update when we bump to 1.87, and this will happen soon ™️

libdd-data-pipeline/src/agent_info/fetcher.rs
```diff
@@ -120,28 +136,28 @@ async fn fetch_and_hash_response<C: HttpClientCapability + SleepCapability>(
     // Runtime-agnostic timeout: race the request against a capability-driven
     // sleep instead of `tokio::time::timeout`, which requires a tokio reactor
     // (not available on wasm where we run on the JS event loop).
-    let res = tokio::select! {
+    let response = tokio::select! {
```
yannham, line 139: Is it really clippy or unrelated claude changes ? 😛

libdd-data-pipeline/src/agent_info/fetcher.rs
```diff
@@ -191,7 +207,7 @@ async fn fetch_and_hash_response<C: HttpClientCapability + SleepCapability>(
 /// `C` is the capability bundle, see [`HttpClientCapability`] and [`SleepCapability`].
 /// Leaf crates pin it to a concrete type.
 #[derive(Debug)]
-pub struct AgentInfoFetcher<C: HttpClientCapability + SleepCapability> {
+pub struct AgentInfoFetcher<C: HttpClientCapability + SleepCapability + Sync> {
```
yannham, line 210: @Aaalibaba42 is that really ok to change a sync? This doesn't look like a clippy trivial fix.

libdd-data-pipeline/src/telemetry/metrics.rs
```diff
@@ -45,33 +45,33 @@ const CHUNKS_DROPPED_STR: &str = "trace_chunks_dropped";
 #[derive(Debug)]
 struct Metric {
     name: &'static str,
-    metric_type: MetricType,
+    kind: MetricType,
```
yannham, line 48: Is this really clippy-related?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -175,25 +175,27 @@ pub(crate) struct TraceExporterWorkers {
...
 ///
 /// ## V07 Serialization
 /// The Trace exporter can serialize the traces to V07 before sending them to the agent.
 ///
 /// ## Stats computation
 /// The Trace Exporter can compute stats on traces. In this case the trace exporter will start
 /// another task to send stats when a time bucket expire. When this feature is enabled the
-/// TraceExporter drops all spans that may not be sampled by the agent.
+/// `TraceExporter` drops all spans that may not be sampled by the agent.
 #[allow(missing_docs)]
```
yannham, line 197: Is this `missing_docs` still relevant (unrelated to the PR, but looks odd)

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -311,12 +314,16 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
     /// * data: A slice containing the serialized traces. This slice should be encoded following the
-    ///   input_format passed to the TraceExporter on creating.
+    ///   `input_format` passed to the `TraceExporter` on creating.
     ///
     /// # Returns
     /// * Ok(AgentResponse): The response from the agent
     /// * Err(TraceExporterError): An error detailing what went wrong in the process
     #[cfg(not(target_arch = "wasm32"))]
+    #[allow(
+        clippy::cast_possible_wrap,
```
yannham, line 324: Which cast does it refer to? I can't find it in `send`

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -375,7 +386,7 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
                 .previous_info_state
                 .load()
                 .as_deref()
-                .map(|s| s.as_str())
+                .map(std::string::String::as_str)
```
yannham, line 389:
```suggestion
                .map(String::as_str)
```

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -505,19 +520,23 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
     ) -> Result<AgentResponse, TraceExporterError> {
         let resource_info = {
             let mut r = OtlpResourceInfo::default();
-            r.service = self.metadata.service.clone();
-            r.env = self.metadata.env.clone();
-            r.app_version = self.metadata.app_version.clone();
-            r.language = self.metadata.language.clone();
-            r.tracer_version = self.metadata.tracer_version.clone();
-            r.runtime_id = self.metadata.runtime_id.clone();
+            r.service.clone_from(&self.metadata.service);
```
yannham, line 534: TIL about `clone_from`!

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -537,6 +556,10 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
 
     /// Deserializes, processes and sends trace chunks to the agent
     #[cfg(not(target_arch = "wasm32"))]
+    #[allow(
```
yannham, line 559: This is a more general remark, but unless there are several casts, I think it's better to put this annotation down near the cast. Otherwise it's harder to convince oneself that it's correct, or it's just context that is not at hand when looking at the cast.

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -209,16 +209,15 @@ pub(crate) fn handle_stats_disabled_by_agent<
...
             client_side_stats,
         );
-        match status {
-            Ok(()) => {
-                #[cfg(feature = "stats-obfuscation")]
-                update_obfuscation_config(agent_info, client_side_stats);
-                debug!("Client-side stats enabled");
-            }
-            Err(_) => error!("Failed to start stats computation"),
+        if matches!(status, Ok(())) {
```
yannham, line 212: Macro-free alternative 🙃 

```suggestion
        if let Ok(()) = status {
```

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -265,11 +264,12 @@ pub(crate) fn handle_stats_enabled(
         let mut concentrator = stats_concentrator.lock_or_panic();
         concentrator.set_span_kinds(get_span_kinds_for_stats(agent_info));
         concentrator.set_peer_tags(agent_info.info.peer_tags.clone().unwrap_or_default());
+        drop(concentrator);
```
yannham, line 267: This one is strange, explicit drop is rather unidiomatic (but in a few cases). What lint is it? If it has to be scoped, I would put it in a block instead:

```
{
   ...concentrator stuff
}
```

libdd-data-pipeline/src/trace_exporter/trace_serializer.rs
```diff
@@ -186,9 +186,9 @@ mod tests {
 
     #[test]
     fn test_build_traces_headers() {
-        let serializer = TraceSerializer::new(TraceExporterOutputFormat::V04);
+        let _serializer = TraceSerializer::new(TraceExporterOutputFormat::V04);
```
yannham, line 189: This one is fishy. Why does `build_trace_headers` not need a serializer anymore? Do we even need to build it then? Why does it return a value?

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -322,6 +329,10 @@ impl<T: Send + BufferSize + 'static> TraceBuffer<T> {
         )
     }
 
+    #[allow(
+        clippy::significant_drop_tightening,
+        reason = "guard must be held across the condvar wait; early drop would be incorrect"
```
yannham, line 334: I don't really understand what's at play here. Which guard are we talking about? What line doesn't compile if we remove the `allow`?

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -343,12 +354,17 @@ impl<T: Send + BufferSize + 'static> TraceBuffer<T> {
...
     }
 
+    #[must_use]
     pub fn queue_metrics(&self) -> QueueMetricsFetcher<T> {
         QueueMetricsFetcher {
             waiter: self.tx.waiter.clone(),
         }
     }
 
+    #[allow(
```
yannham, line 364: This looks fishy, since this function is just a thin wrapper. Isn't rather `self.tx.wait_shutdown_done` that should have the lint instead?

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -426,21 +447,23 @@ impl<T> Sender<T> {
                 return Err(TraceBufferError::TimedOut(Duration::ZERO));
             }
             let state = self.lock_state()?;
-            let (_state, res) = self
+            let (state, res) = self
                 .waiter
                 .sender_notifier
                 .wait_timeout_while(state, timeout, cond)
                 .map_err(|_| TraceBufferError::MutexPoisoned)?;
+            drop(state);
```
yannham, line 455: Matter of taste, but I found scope to be better than explicit `drop` to visualize the lifetime of objects when applicable.

```suggestion
            let res = {
                let (_state, res) = self
                    .waiter
                    .sender_notifier
                    .wait_timeout_while(state, timeout, cond)
                    .map_err(|_| TraceBufferError::MutexPoisoned)?;
                res
            };
```

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -412,6 +429,10 @@ struct Sender<T> {
 }
 
 impl<T> Sender<T> {
+    #[allow(
```
yannham, line 432: Which guard and which wait? Again, I find the justification hard to understand

libdd-data-pipeline/src/health_metrics.rs
```diff
@@ -288,21 +283,25 @@ impl SendResult {
...
-    /// should be sent to DogStatsD
+    /// should be sent to `DogStatsD`
     ///
     /// # Returns
     ///
     /// A vector of `(HealthMetric, Option<Cow<'static, str>>)` tuples where:
     /// - The first element is the metric to emit
     /// - The second element is an optional tag value for error classification
+    #[allow(
+        clippy::cast_possible_wrap,
```
yannham, line 302: This isn't needed anymore, since you use `u64::try_from`?

yannham (commented):
Sorry, I started reviewing before it was switched back to draft, sor so I think. In general I feel there are some renaming business that is unrelated to clippy (unless there's a lint around variable names, but that sounds... surprising?) that could have been a separate (it's not useful to split it, just mentioning it for future reference)

Eldolfin:
Thanks for the review, opening the PR without draft at first was a mistake on my part but I appreciate the early comments !

From what we've discussed IRL the plan would be to be less aggressive in the enabled lints at first to reduce the noise (false positives and refactors without value):
Right now lint groups `pedantic` and `nursery` are enabled in the root `Cargo.toml` but the [clippy documentation](https://doc.rust-lang.org/stable/clippy/lints.html) explicitly discourage enabling the whole group because they still have too many false positives. I'll instead cherry-pick a list of useful lints like the doc suggests.

PR #2014, opened 2026-05-19
fix(otlp): propagate _dd.p.tid from chunk root to all spans

zacharycmontoya:
# What does this PR do?

This PR updates the conversion of DD spans to OTLP spans to ensure all spans receive the high 64 bits of the 128-bit trace ID.

DD tracers set `_dd.p.tid` (high 64 bits of 128-bit trace ID) only on the chunk root per RFC #85 — the Datadog backend reconstructs the full 128-bit ID at ingest. The OTLP mapper previously only read the tag per span, so child spans were emitted with upper 64 bits zeroed and traces fragmented in pure-OTel backends.

The approach determines the chunk-level `_dd.p.tid` once in `map_traces_to_otlp` and applies it to every span in the chunk. Per-span value still wins (forward-compat with tracers that propagate everywhere). Use `find_map` over the chunk so a malformed root tag falls back to the first parseable value in the chunk rather than poisoning the whole trace.

# Motivation

Performing end-to-end tests with dd-trace-py exporting OTLP spans surfaced this issue, which results in child spans recording different trace IDs than the local root spans. Notably, the only difference was that the high 64-bits of the 128-bit trace ID were zero'ed out.

# Additional Notes
A system-test to cover this scenario is being simultaneously added in https://github.com/DataDog/system-tests/pull/6973

# How to test the change?

Regression tests have been added to `libdd-trace-utils/src/otlp_encoder/mapper.rs`.


libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -22,15 +22,30 @@ const MAX_ATTRIBUTES_PER_SPAN: usize = 128;
...
+        // `_dd.p.tid` meta tag. Scan the chunk for the first parseable value so a malformed
+        // root tag doesn't poison the rest of the chunk; absence (legacy 64-bit traces) → 0.
+        let chunk_trace_id_high: u64 = chunk
+            .iter()
+            .find_map(|s| {
+                s.meta
+                    .get("_dd.p.tid")
+                    .and_then(|v| u64::from_str_radix(v.borrow(), 16).ok())
+            })
+            .unwrap_or(0);
```
paullegranddc: the Span<T> type does represent trace ids as 128 bit ints (the highest 64 bits are discarded at serialization to v04). This is useful for instance in python where (once we release native spans) the trace id field will hold the full trace id
I'd prefer if you checked first that these are not zero before scanning through meta, for efficiency reason

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -92,15 +107,20 @@ fn build_resource(resource_info: &OtlpResourceInfo) -> Resource {
...
     // in the trace_id field; when a tracer emits a 128-bit ID the high 64 bits are propagated as
-    // the hex string meta tag "_dd.p.tid".
+    // the hex string meta tag "_dd.p.tid". Per-span tag wins (forward-compat with tracers that
+    // propagate it everywhere), then the chunk-root value resolved by the caller, then 0.
     let trace_id_high: u128 = span
         .meta
         .get("_dd.p.tid")
         .and_then(|v| u64::from_str_radix(v.borrow(), 16).ok())
-        .unwrap_or(0) as u128;
+        .unwrap_or(chunk_trace_id_high) as u128;
```
paullegranddc: Is this still necessary given that we do have the value passed as argument and all spans in a chunk should be of the same trace?

libdd-trace-utils/src/otlp_encoder/mapper.rs
zacharycmontoya: Ah okay. Are you suggesting I check only the first span's trace ID field to see if it's 128 bit or add a trace ID lookup in the iteration (for each span) before checking the meta dictionary?

Also, once I have this value, do you recommend passing the full 128-bit trace ID to `map_span` or pass only the upper 64-bits like I have right now?

libdd-trace-utils/src/otlp_encoder/mapper.rs
zacharycmontoya: Hmm. I think you're right, we don't need this lookup anymore. Removing this would only break scenarios where spans in the same chunk had different values for the `_dd.p.tid` span tag, which seems like an invalid scenario. I can remove this.

libdd-trace-utils/src/otlp_encoder/mapper.rs
zacharycmontoya: Implemented a check for the trace_id field first in f4fd0e80a6731bcd8e2faf2f7893ee63c5434a6a

libdd-trace-utils/src/otlp_encoder/mapper.rs
zacharycmontoya: Removed in 17e1bc24e0088795e2d43de3171cd98a0682f011

mabdinur (approved):

paullegranddc (approved):

PR #2016, opened 2026-05-20
refactor(otel-thread-ctx): backport review suggestions from Polar Signals

yannham:
# What does this PR do?

Backports applicable review feedback from the upstream [polarsignals/custom-labels#15](https://github.com/polarsignals/custom-labels/pull/15) PR into our `libdd-otel-thread-ctx` crate.

This PR switches `get_tls_slot` from returning a reference with an unconstrained lifetime parameter to a closure-based `with_tls_slot` API, following the same pattern as `std::thread_local!`, for improved safety. This PR also moves the `extern "C"` FFI declaration inside the function body so the raw symbol is structurally inaccessible outside the safe wrapper.

More changes to follow in small themed PRs.

# Motivation

We upstreamed libdatadog's OTel process context and thread context to the custom-labels repo, and got good suggestions (courtesy of @umanwizard). Several suggestions apply directly to our vanilla `otel-thread-ctx` crate:

- The old `get_tls_slot<'a>() -> &'a AtomicPtr<...>` signature has an unconstrained lifetime equivalent to `'static`, but the TLS storage is not actually `'static` (it dies with the thread). Plus, it's forbidden to mix atomic and non-atomic accesses (although we don't do it, it's not prevented for future code changes). The closure-based API prevents this structurally, matching how `std::thread_local!` solves the same problem.
- The top-level `extern "C"` block relied on a doc comment ("CAUTION: do not use directly") to prevent misuse. Nesting it inside the accessor enforces this at the language level.

# Additional Notes

The closure is fully monomorphized and inlined by LLVM in release mode: `attach`, `detach`, and `update` produce byte-identical assembly before and after this change. No performance impact.

# How to test the change?

Tests pass (no change in behavior)

### DIFF
--- a/libdd-otel-thread-ctx/src/lib.rs
+++ b/libdd-otel-thread-ctx/src/lib.rs
@@ -73,23 +73,24 @@ pub mod linux {
         sync::atomic::{compiler_fence, AtomicPtr, AtomicU8, Ordering},
     };
 
-    extern "C" {
-        /// Return the address of the current thread's `otel_thread_ctx_v1` local.
-        ///
-        /// **CAUTION**: do not use this directly, always go through [get_tls_slot] to read and
-        /// write it atomically.
-        fn libdd_get_otel_thread_ctx_v1() -> *mut *mut c_void;
-    }
-
-    /// Return an atomic view of the TLS slot. The address calculation requires a call to a C shim
-    /// in order to use the TLSDESC dialect from Rust. The returned address is stable (per thread),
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
👍 Looks good from my side. I'll admit that as usual I'm not the most knowledgeable about the rust low-level details so probably worth having an additional rusty human do a quick pass as well :)

paullegranddc (approved):

scottgerring:
Late to the party but super excited to see the stuff from the polar' PR feeding back here :)

PR #2018, opened 2026-05-20
refactor(otel-thread-ctx): add const offset assertions for ThreadContextRecord

yannham:
# What does this PR do?

Follow-up to #2016.

Adds compile-time `mem::offset_of!` assertions for every field of `ThreadContextRecord`, catching accidental ABI breakage (field reordering, padding changes) at compile time rather than at runtime in a profiler.

The existing size-only assertion from inside `new()` is moved to a top-level `const _: () = { ... }` block next to the struct definition so it fires unconditionally regardless of code path.

# Motivation

`ThreadContextRecord` has a fixed ABI read by out-of-process eBPF readers. The previous single `size_of == 640` check would still pass if fields were reordered without changing the total size. Per-field offset assertions make this impossible.

# Additional Notes

N/A

# How to test the change?

N/A


ivoanjo (approved):
👍 LGTM Really like this one!

paullegranddc (approved):

PR #2021, opened 2026-05-21
chore(ci): make it possible to run checks on external pr

bantonsson:
# What does this PR do?

Makes it possible to run our checks on external PRs. Skips uploading results (that failed) of the secrets are unavailable, and allows the static checks to run with write permissions.

# Motivation

https://github.com/DataDog/libdatadog/pull/2020

# Additional Notes

# How to test the change?

### DIFF
--- a/.github/workflows/pr-metadata-docs-and-deps.yml
+++ b/.github/workflows/pr-metadata-docs-and-deps.yml
@@ -1,9 +1,27 @@
 name: Metadata, docs and deps
+# Using `pull_request_target` over `pull_request` for elevated `GITHUB_TOKEN`
+# privileges, so that fork PRs can also receive PR comments from this workflow.
+#
+# Security model:
+# - `pull_request_target` runs in the base repository context, so the default
+#   checkout is the target branch (`main`). This is trusted code.
+# - The `changed-crates` job intentionally does NOT override `ref`, so it checks
+#   out the base branch. The `.github/actions/crates-reporter` composite action
+#   (including the binary it compiles) comes from the trusted base, not from the
+#   fork. The action fetches all remote refs and diffs against the PR head as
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-metadata-docs-and-deps.yml
```diff
@@ -1,9 +1,27 @@
...
+#   out the base branch. The `.github/actions/crates-reporter` composite action
+#   (including the binary it compiles) comes from the trusted base, not from the
+#   fork. The action fetches all remote refs and diffs against the PR head as
+#   data only. This avoids the "pwn request" attack where an attacker modifies
+#   action code in a fork and runs it with the elevated token.
+# - All other jobs that do need the PR's source (cargo check, cargo deny) use
+#   `ref: ${{ github.event.pull_request.head.sha }}` and treat that code as
+#   passive data (read but not executed as part of the workflow itself). Those
+#   jobs invoke `cargo check`/`cargo deny` which compiles and executes code, but
+#   they do not load untrusted composite actions or shell scripts from the checkout.
```
paullegranddc, line 18: It has already been tagged by codex, but build scripts run arbitrary code and can exfiltrate the write token.

paullegranddc (changes requested):
Any write action by an external contirbutor seems extremly dangerous as we haven't considered this case to begin with when writing the original checks...

If we remove the write actions, I found the `persist-credentials: false` option on checkout which yeets the github token from the cache so it cannot be reused by other jobs, which seems like a good idea.

I'd much prefer it if we had a "manual" workflow of copying commits made by external contributors on a datadog employee branch **after review**. Then we can run the CI without fear, and the external contributor cannot push potential attacks.

bantonsson:
I'll close this for now. Need to go back to the drawing board.

PR #2022, opened 2026-05-21
feat(trace-utils)!: introduce VecMap datastructure

yannham:
# What does this PR do?

This PR is the first of a series that backports the change buffer implementation for dd-trace-js from experimentation to production in libdatadog.

This PR introduces a replacement for `HashMap`: `VecMap`. It's not used anywhere else; this is left for follow-up PRs, but the idea is to use it for the `meta`, `metrics`, and `meta_struct` fields of v04 spans.

# Motivation

The change-buffer module (introduced in subsequent PRs) experimentations showed that we spend quite some time expanding hashmaps and thus re-hashing during span reconstruction. Introducing a map-like data structure that is optimized for construction/insertion (which is the vast majority of operations) showed to be beneficial.

Even if not optimized for reads, we expect that given the small cardinality of meta and metrics in a span (< 20 entries in general), linear scan is actually going to perform well because of cache locality. Removal is one operation that is made quite more expensive (although still linear), but can be mitigated later with tombstones (also still linear, but doesn't require shifting). Tombstones aren't implemented for this first version.

# Additional Notes

Deduplication is performed at serialization time. We initially set up for sort + deduplication to avoid allocation, but this would require larger changes in libdatadog by requiring Span values to be `Ord`. For now, deduplication is based on a `HashSet` since `SpanData::Text` is already `Hash`, but this incurs one allocation. I think it's reasonable because the serialization path is not hot anyway, and we can later request `Ord` on values to go the in-place `sort_dedup` route instead if desired.

# How to test the change?

Run tests.

### DIFF
--- a/libdd-trace-utils/src/span/vec_map.rs
+++ b/libdd-trace-utils/src/span/vec_map.rs
@@ -24,7 +24,7 @@ use std::hash::Hash;
 /// enough so such that the size penalty of duplication is expected to be reasonable.
 ///
 /// **Important**: note that only [VecMap::get] and [VecMap::get_mut] are duplicate-aware, so to
-/// speak. [Vec::len], [Vec::iter], and others just delegates to the underlying `Vec`, and won't
+/// speak. [VecMap::len], [VecMap::iter], and others just delegates to the underlying `Vec`, and won't
 /// deduplicate.
 ///
 /// Explicit deduplication is currently being done automatically and on-the-fly during
--- a/libdd-trace-utils/src/span/vec_map.rs
+++ b/libdd-trace-utils/src/span/vec_map.rs
@@ -170,7 +170,7 @@ impl<K, V> Extend<(K, V)> for VecMap<K, V> {
 
 impl<K: Serialize + Eq + Hash, V: Serialize> Serialize for VecMap<K, V> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        let mut map = serializer.serialize_map(Some(self.0.len()))?;
+        let mut map = serializer.serialize_map(None)?;
         let mut seen = HashSet::with_capacity(self.len());
         for (k, v) in self.0.iter().rev() {
             if seen.insert(k) {
--- a/libdd-trace-utils/src/span/vec_map.rs
+++ b/libdd-trace-utils/src/span/vec_map.rs
@@ -5,7 +5,8 @@
 //! by a vector. Spans are mostly allocated and constructed, and more rarely read or mutated.
 //! [VecMap] is thus optimized for insertion (which is just `Vec::push`), without any hashing
 //! involved. Fetching and removing a value is, on the other hand, linear time in the size of the
-//! map.
+//! map. However, since meta and metrics are expected to be typically small (20ish elements or
+//! less), linear scan is usually still competitive with hashmap's `get`.
 
 use serde::ser::{Serialize, Serializer};
 use std::borrow::Borrow;
@@ -18,8 +19,8 @@ use std::hash::Hash;
 ///
 /// Duplicates are tolerated: [VecMap::insert] always appends, and [VecMap::get]/[VecMap::get_mut]
 /// return the *last* matching entry so that later writes shadow earlier ones. This optimizes for
-/// fast insert and construction (that might happen on the client's application hot path), avoiding
-/// a linear scan on each insert (or a potential costly full re-hashing with a hashmap).
+/// fast insertion and construction (that might happen on the client's application hot path),
+/// avoiding a linear scan on each insert, or a potential full re-hashing with a hashmap.
 /// Additionally, while overriding a metric or a meta definitively happens, it's assumed to be rare
 /// enough so such that the size penalty of duplication is expected to be reasonable.
 ///
@@ -170,14 +171,14 @@ impl<K, V> Extend<(K, V)> for VecMap<K, V> {
 
 impl<K: Serialize + Eq + Hash, V: Serialize> Serialize for VecMap<K, V> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        // We pre-compute the deduped map. If deduplication is done on the fly during serialization,
-        // we can't provide a length up front to the serializer, and the current one (rmp) will
-        // allocate an intermediate buffer defensively.
+        // We pre-compute the deduped map. If deduplication were done on the fly during
+        // serialization, we couldn't provide a length up front to the serializer, and the current
+        // one (rmp) will allocate an intermediate buffer defensively.
         self.0
             .iter()
             .map(|(k, v)| (k, v))
             // Since the iterator is sized, `collect()` should pre-allocate with the right capacity
-            // directly.
+            // in one shot.
             .collect::<HashMap<&K, &V>>()
             .serialize(serializer)
     }
--- a/libdd-trace-utils/src/span/vec_map.rs
+++ b/libdd-trace-utils/src/span/vec_map.rs
@@ -10,7 +10,7 @@
 
 use serde::ser::{Serialize, Serializer};
 use std::borrow::Borrow;
-use std::collections::HashMap;
+use std::collections::HashSet;
 use std::hash::Hash;
 
 /// A Vec-backed map that provides HashMap-like lookup by key.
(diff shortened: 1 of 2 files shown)


libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -0,0 +1,260 @@
...
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut map = serializer.serialize_map(None)?;
+        let mut seen = HashSet::with_capacity(self.len());
+        for (k, v) in self.0.iter().rev() {
+            if seen.insert(k) {
+                map.serialize_entry(k, v)?;
+            }
+        }
+        map.end()
+    }
```
Aaalibaba42, line 277: This will make everything reverse order, not just the duplicates. If serialized we don't care about ordering this is fine. Otherwise a map with upsert without reversing would achieve the same effect without reversing outputs for non-dedup entries but is slightly slower.

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -0,0 +1,260 @@
...
+///
+/// Duplicates are tolerated: [VecMap::insert] always appends, and [VecMap::get]/[VecMap::get_mut]
+/// return the *last* matching entry so that later writes shadow earlier ones. This optimizes for
+/// fast insert and construction (that might happen on the client's application hot path), avoiding
+/// a linear scan on each insert (or a potential costly full re-hashing with a hashmap).
+/// Additionally, while overriding a metric or a meta definitively happens, it's assumed to be rare
+/// enough so such that the size penalty of duplication is expected to be reasonable.
+///
+/// **Important**: note that only [VecMap::get] and [VecMap::get_mut] are duplicate-aware, so to
+/// speak. [Vec::len], [Vec::iter], and others just delegates to the underlying `Vec`, and won't
```
Aaalibaba42:
```suggestion
/// speak. [VecMap::len], [VecMap::iter], and others just delegates to the underlying `Vec`, and won't
```

libdd-trace-utils/src/span/vec_map.rs
Aaalibaba42: Also building this first would give us the dedup size and so we wouldn't need to have the `serializer.serialize_map(None)`.

Am I missing something ?

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -0,0 +1,260 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
```
Eldolfin, line 1: Github doesn't support thread comments so putting my thoughts here to keep it clear.

Overall this seems like it could be very beneficial to perf, have you considered:
1. Using a dependency that does it (ex: https://github.com/p-avital/vec-map-rs)
1. Storing keys and values in separate Vecs which as explained in [vec-map-rs's readme](https://github.com/p-avital/vec-map-rs#how-does-it-compare-with-linear_map) could make the key lookup even faster by packing them close together.
1. Adding micro benchmarks to compare HashMap/BTreeMap/VecMap(/vec-map-rs) operations on different number of keys, size of keys and size of values ?
1. Having a higher level hybrid struct that underneath uses a VecMap until a certain threshold in the number of keys is reached then switches to HashMap ? (might be overkill)

paullegranddc: > could make the key lookup even faster by packing them close together.

Attribute lists are usually pretty small (less than 10s of elements), and we need to chase one pointer when iterating anyway to read the string, and we would need to do two allocations instead of one so I have doubts on the perf gain.

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -170,7 +170,7 @@ impl<K, V> Extend<(K, V)> for VecMap<K, V> {
 
 impl<K: Serialize + Eq + Hash, V: Serialize> Serialize for VecMap<K, V> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        let mut map = serializer.serialize_map(Some(self.0.len()))?;
+        let mut map = serializer.serialize_map(None)?;
```
paullegranddc: This is not great as rpm_serde will allocate an intermediary buffer and copy it to the main buffer once it contains the length https://docs.rs/rmp-serde/latest/src/rmp_serde/encode.rs.html#474-494

I think we should deduplicate before serialization/rely on the agent semantic that last duplicate key wins

libdd-trace-utils/src/span/vec_map.rs
yannham: Those are all very good points:

1. There's always a balance when pulling in a new dependency. For libdatadog, the cons are size explosion, and supply chain attack, most notably. Since the code is really simple (it's mostly boilerplate to be honest), I felt like having a bespoke datastructure here was preferable. But I think both approaches are reasonable. Another thing with having our own is that we can customize it to our needs, typically the tombstone-like removal (well it could probably easily be implemented as a wrapper of vec-map-rs, so maybe not a great example). Meta and metrics are very special cases of maps, and we might be able to take advantage of that even further?
2. As @paullegranddc mentioned, I don't think it's worth the complexity or even that it would be a win. We do not care that much for get already. But to keep in mind for potential evolutions?
3. We plan to bench this in dd-trace-rs and other real usage. For now this code is dead, so it doesn't impact anything that exists already. But maybe having a bunch of micro benches doesn't hurt, and can be easily generated by an LLM. I'll look into it!
4. Without thinking too much into it, I think this might be overkill for now. Simplicity, beside being beneficial for maintenance, usually wins in terms of performance in a surprisingly vast number of use-cases: less branching, no pointer chasing and cache misses, etc. We should bench real tracers and only think about a more complex scheme if it turns out there are largish maps that show unacceptable perf degradation.

libdd-trace-utils/src/span/vec_map.rs
yannham: This is a point where I hesitated. At some point I had a `DedupedVecMap` wrapper, and only implemented serialize on it. However, it's surprisingly annoying to avoid any additional allocation when doing it in-place. You either need to clone the keys (maybe that's ok?), because all the nice iterators or `retain` helpers take a closure that holds key references that don't survive long enough.

One possibility is to clone keys to put them in the set. I think we could technically avoid that but that would mean re-implementing `retain` manually, basically... Probably `clone` is fine.

libdd-trace-utils/src/span/vec_map.rs
yannham: I think reverse order is fine. I should make that explicit, but `VecMap` doesn't provide any guarantees about the iteration order of stuff with different keys, as most maps (`HashMap` in particular, which has a random iteration order by default).

The problem of building the deduped map first is that it causes an additional intermediate allocation. Or we could just count in a first loop, but then we'd have to clear the `HashSet` and to start again hash everything twice. None of those are really a no-go, but I could avoid them this way.

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -0,0 +1,260 @@
...
+
+impl<K: Serialize + Eq + Hash, V: Serialize> Serialize for VecMap<K, V> {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let mut map = serializer.serialize_map(None)?;
+        let mut seen = HashSet::with_capacity(self.len());
+        for (k, v) in self.0.iter().rev() {
+            if seen.insert(k) {
+                map.serialize_entry(k, v)?;
+            }
+        }
```
paullegranddc: What about doing this then? It allocates a bit more for because we use a hashmap with the index instead of a set but does the same amount of work

```suggestion
        let mut seen = HashMap::with_capacity(self.len());
        for (i, (k, _)) in self.0.iter().rev().enumerate() {
            seen.entry(k.as_ref()).or_insert(i);
        }
        let mut map = serializer.serialize_map(Some(seen.len()))?;
        for (k, v) in seen.values().map(|i| self.0.get(i)) {
            map.serialize_entry(k, v)?;
        }
```

yannham: Right, moving everything to a `hashmap` is another possibility, but as you said it does allocate a bit more. If we do this I wonder if we shouldn't bite the bullet and just build a hashmap out of the vec and iterate over it. If the value are 64 bits, this get rid of fetching back stuff from the vec, and waste no more. If values can be locally large though, maybe not.

yannham: After discussion, we can just store references in the hashmap, which is the best of both world.

yannham: Implemented in 1a678f92734f3d8fac1b1d58e72a664b04289b5e

paullegranddc (approved):

PR #2026, opened 2026-05-22
feat(sidecar): forward FFE exposures to EVP proxy

leoromanovsky:
## Motivation

PHP FFE exposure delivery needs a native path with a cache that persists beyond a single PHP request/thread. The shared design doc is the cross-PR reference: https://docs.google.com/document/d/1NvMfTpZWLBlFmEFNjdnlMyeVpy5l7KD8qujGFco6w2w/edit?tab=t.0

This PR is exposure-only. Metrics were split into DataDog/libdatadog#2052 so reviewers can evaluate exposure cache and delivery separately from OTLP evaluation metrics.

## Changes

This adds caller-driven FFE exposure sidecar actions, exposure payload forwarding through the Agent EVP proxy, and a shared exposure cache that deduplicates repeated `(service, env, version, flag, subject)` assignments across PHP requests and sidecar connections.

The reusable FFE-domain pieces now live in `datadog-ffe` behind the `exposure-events` feature: exposure input types, the LRU deduplication cache, and JSON payload encoding. `datadog-sidecar` keeps only sidecar-specific work: deriving the agent EVP endpoint, building the HTTP request, applying the timeout, logging delivery failures, and integrating with sidecar lifecycle/actions.

Current PHP MVP path:

```mermaid
flowchart LR
    Eval["PHP native evaluation<br/>ddog_ffe_evaluate"]
    Batch["PHP tracer native memory<br/>request/thread-local exposure batch"]
    Shutdown["PHP RSHUTDOWN<br/>flush exposure batch"]
    Action["sidecar action<br/>record FFE exposures"]
    Domain["datadog-ffe<br/>feature: exposure-events<br/>types + cache + JSON encoder"]
    Sidecar["shared sidecar<br/>cross-request and cross-thread exposure cache"]
    Agent["Datadog Agent<br/>EVP proxy"]
    Intake["FFE exposure intake"]

    Eval -->|"doLog=true assignment"| Batch
    Batch --> Shutdown
    Shutdown --> Action
    Action --> Domain
    Domain --> Sidecar
    Sidecar --> Agent
    Agent --> Intake
```

Future Python/Ruby connection:

```mermaid
flowchart LR
    PyToday["dd-trace-py today<br/>host-language exposure writer"]
    RbToday["dd-trace-rb today<br/>host-language exposure writer"]
    PyFuture["dd-trace-py future<br/>explicit native opt-in"]
    RbFuture["dd-trace-rb future<br/>explicit native opt-in"]
    Native["libdatadog caller-driven<br/>FFE exposure action"]
    Shared["shared sidecar<br/>dedupe + EVP delivery"]
    Agent["Datadog Agent<br/>EVP proxy"]

    PyToday -. "current direct EVP path" .-> Agent
    RbToday -. "current direct EVP path" .-> Agent
    PyFuture -. "after ownership switch" .-> Native
    RbFuture -. "after ownership switch" .-> Native
    Native --> Shared
    Shared --> Agent
```

The future Python/Ruby arrows are intentionally not active behavior in this PR. They show why the reusable code lives in `datadog-ffe` rather than directly in sidecar internals, while preserving today's host-language ownership.

Why Python/Ruby do not double count today:

- Python and Ruby use libdatadog for evaluation only; the evaluator returns assignment metadata and does not enqueue exposure telemetry as a side effect.
- This PR adds a separate caller-driven sidecar action. Exposure emission happens only when an SDK explicitly records exposure candidates into that action. PHP wires this in its companion PR; Python and Ruby do not.
- Python and Ruby therefore keep exactly their current host-language EVP exposure writers. They are not also sending exposure candidates through this native sidecar path.
- The sidecar cache only deduplicates exposure candidates that enter the native sidecar path. It cannot protect direct host-language EVP writers, so future Python/Ruby migration must switch ownership to native logging and disable/bypass the host exposure writer for the same evaluations.

Reference implementation check: dd-trace-java follows the same exposure semantics and user ergonomics. Java's `DDEvaluator` is SDK-owned evaluation code; after resolving an assignment, it checks allocation `doLog`, builds an exposure event with flag, variant, allocation, targeting key, and context, and dispatches it through `FeatureFlaggingGateway`. `ExposureWriterImpl` subscribes to those exposure events, queues them, deduplicates with an LRU exposure cache, serializes service/env/version context, and posts to the Agent EVP proxy. Application code only calls the OpenFeature provider; it does not call an exposure API.

PHP mirrors that canonical shape, with PHP-specific lifecycle mechanics: the dd-trace-php evaluation bridge records `doLog=true` exposure candidates internally, request shutdown flushes the batch, and this PR's sidecar path owns cross-request deduplication and EVP delivery. For future Python/Ruby migration, the same rule applies: wire native exposure recording inside the SDK-owned evaluation path, and turn off the existing host-language exposure writer for those evaluations.

## Decisions

No telemetry is emitted automatically from shared libdatadog evaluator calls. SDKs must explicitly enqueue FFE telemetry actions. This remains required for Python/Ruby coexistence because those SDKs currently log exposures and metrics in host-language code.

The sidecar cache deduplicates only exposure candidates sent through this native sidecar path; it cannot deduplicate direct host-language EVP writers.

Future Python/Ruby migration must be an ownership switch, not an additional writer. When those SDKs opt into this native exposure path, their host-language exposure writers must be disabled or bypassed for the same evaluations to avoid double counting.

## Validation

Current head (`8be471fbc`) local validation:

```sh
cd /Users/leo.romanovsky/go/src/github.com/DataDog/libdatadog-ffe-sidecar-exposures
cargo fmt --check
cargo test -p datadog-ffe --features exposure-events telemetry::exposures
cargo test -p datadog-sidecar ffe_exposure
cargo check -p datadog-ffe
cargo check -p datadog-sidecar-ffi
```

Results: datadog-ffe exposure tests passed (4 passed), sidecar exposure tests passed (6 passed), default datadog-ffe check passed, sidecar FFI check passed, fmt check passed with only the repo stable-rustfmt warnings.

Prior downstream PHP behavior validation before the reusable-crate refactor, from DataDog/dd-trace-php#3910 using this PR at `6d23848a`:

```text
ffe-dogfooding subject=php-3910-split-1779981442
php7_exposures=1 php8_exposures=1
php7_metrics=0 php8_metrics=0
```

System-tests downstream validation:

```sh
TEST_LIBRARY=php ./run.sh FEATURE_FLAGGING_AND_EXPERIMENTATION tests/ffe/test_exposures.py -vv
```

Result: 11 passed in 77.53 seconds.

Related PRs: DataDog/dd-trace-php#3906, DataDog/dd-trace-php#3910, DataDog/libdatadog#2052, DataDog/system-tests#7031.


datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -405,6 +408,51 @@ impl SidecarInterface for ConnectionSidecarHandler {
...
+        // registration call.
+        let actions: Vec<SidecarAction> = actions
+            .into_iter()
+            .filter(|a| match a {
+                SidecarAction::FfeExposures(payload) => {
+                    if let Some(base) = trace_config.endpoint.as_ref() {
+                        if let Some(ep) = ffe_exposures_flusher::exposure_endpoint(base) {
+                            let payload = payload.clone();
+                            tokio::spawn(async move {
+                                let client = NativeCapabilities::new_client();
```
bantonsson: Not sure how often this happens and if it makes sense to reuse the `NativeCapabilities::new_client()` client for connection reuse.

datadog-sidecar/src/service/ffe_exposures_flusher.rs
bwoebi, line 1: Is there a reason this lives in sidecar, rather than behind a feature gate in ffe crate? Seems pretty reusable for other stuff too?

leoromanovsky: Good idea!

datadog-ffe/src/telemetry/exposures.rs
```diff
@@ -0,0 +1,318 @@
...
+            version: context.version.clone(),
+            flag_key: exposure.flag_key.clone(),
+            subject_id: exposure.subject_id.clone(),
+        };
+        let value = ExposureCacheValue {
+            allocation_key: exposure.allocation_key.clone(),
+            variant: exposure.variant.clone(),
+        };
+
+        let mut cache = self.cache.lock().unwrap_or_else(|e| e.into_inner());
```
sameerank: I'd look into this a little further, but Claude says this poses a risk of mutex poisoning panic across the FFI boundary, e.g. `ddog_sidecar_send_ffe_exposure_batch`.

https://github.com/DataDog/libdatadog/blob/af9b62bc4f968a4754637dc8f5abfa8a336e96a7/AGENTS.md#L68

One suggestion is to catch the unwind (`catch_unwind`) in the at the FFI boundary. The other is to replace the cache with `DashMap`

datadog-ffe/src/telemetry/exposures.rs
```diff
@@ -0,0 +1,318 @@
...
+
+//! Reusable FFE exposure payload and deduplication primitives.
+
+use super::FfeTelemetryContext;
+use lru::LruCache;
+use serde::{Deserialize, Serialize};
+use std::num::NonZeroUsize;
+use std::sync::{Arc, Mutex};
+
+const DEFAULT_CACHE_LIMIT: usize = 65_536;
```
sameerank, line 14: This number seems quite random without extra context. And if hitting the cache limit is expected, it might be worth exploring TTL eviction in addition to LRU

datadog-ffe/src/telemetry/exposures.rs
```diff
@@ -0,0 +1,318 @@
...
+}
+
+fn subject_attributes(json: &str) -> serde_json::Map<String, serde_json::Value> {
+    if json.is_empty() {
+        return serde_json::Map::new();
+    }
+
+    match serde_json::from_str::<serde_json::Value>(json) {
+        Ok(serde_json::Value::Object(attrs)) => attrs,
+        Ok(_) | Err(_) => serde_json::Map::new(),
```
sameerank: Instead of silent failing for invalid JSON, I wonder if a log would be additionally helpful for anyone passing subject attributes into the caller

datadog-ffe/src/telemetry/exposures.rs
```diff
@@ -0,0 +1,318 @@
...
+        context: ExposurePayloadContext::from(batch.context),
+        exposures,
+    };
+    serde_json::to_string(&payload).map(Some)
+}
+
+fn is_complete(exposure: &FfeExposure) -> bool {
+    !exposure.flag_key.is_empty()
+        && !exposure.allocation_key.is_empty()
+        && !exposure.variant.is_empty()
```
sameerank, line 124: Also `&& !exposure.subject_id.is_empty()`? Or document that an empty `subject_id` is intentionally allowed

sameerank (commented):
I think the main thing that I'd double check is the panic risk, and the rest are minor/nits. It would also be preferable to hear from another reviewer who knows Rust better

datadog-ffe/src/telemetry/exposures.rs
leoromanovsky: 64 kB - It's the size of all the other servers' caches. I'll add a note.

datadog-ffe/src/telemetry/exposures.rs
```diff
@@ -0,0 +1,335 @@
...
+
+impl Default for ExposureDeduplicator {
+    fn default() -> Self {
+        Self::new(DEFAULT_CACHE_LIMIT)
+    }
+}
+
+impl ExposureDeduplicator {
+    pub fn new(limit: usize) -> Self {
+        let limit = NonZeroUsize::new(limit).unwrap_or(NonZeroUsize::MIN);
```
sameerank: nit: `limit` can be set to 0, indicating no-deduping, but this would replace it with 1, which would deduplicate consecutive identical exposures.

In practice, you always invoke it with the default so this never happens

https://github.com/DataDog/libdatadog/blob/d780e494df56d75ba6125e26e2ff2c116dc00659/datadog-ffe/src/telemetry/exposures.rs#L43

But it would be an unexpected since the caller can pass 0 (with `usize`), implying that it is valid

datadog-ffe/src/telemetry/exposures.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    pub fn should_send(&self, context: &FfeTelemetryContext, exposure: &FfeExposure) -> bool {
+        let key = ExposureCacheKey {
+            service: context.service.clone(),
+            env: context.env.clone(),
+            version: context.version.clone(),
+            flag_key: exposure.flag_key.clone(),
+            subject_id: exposure.subject_id.clone(),
+        };
```
sameerank, line 65: nit: Could compute a hash from the keys and check for membership in a "seen" memo if there is any performance concern about cloning strings. But that has its own tradeoffs (e.g. extra memory for the hash set) so this is fine too

sameerank (approved):

datadog-ffe/src/telemetry/exposures.rs
leoromanovsky: Good call!

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -405,6 +411,49 @@ impl SidecarInterface for ConnectionSidecarHandler {
             trace_config.tracer_version.clone(),
         );
 
+        // FFE exposure actions are session-scoped, not application-scoped:
+        // dispatch them before the `applications.entry(queue_id)` check so they
+        // are not silently dropped when the PHP runtime hasn't yet registered the
+        // application via remote-config metadata. The PHP exposure writer can
+        // fire as soon as evaluations begin, which is often earlier than the
+        // first RC config registration call.
```
bwoebi: This comment is simply useless. There will never be actions enqueued, without the RC being registered.

bwoebi (approved):
Generally looks good to me, except for a weird comment.

PR #2029, opened 2026-05-22
feat!: Align tracer FFI error and response types with common conventions

lloeki:
# What does this PR do?

Two improvements to the tracer FFI surface in `libdd-data-pipeline-ffi`:

1. **Check for nullable pointers.**  

2. **Return `ByteSlice` from `ddog_trace_exporter_response_get_body`.**
   Replace the separate `*const u8` return + `out_len` out-parameter with
   a single `ByteSlice` return value, matching the existing slice
   conventions in `libdd-common-ffi`.

# Motivation

FUP to #1952. C consumers (dd-trace-rb) can now use the existing
`read_ddogerr_string_and_drop` / `get_error_details_and_drop` helpers
from `datadog_ruby_common.h` instead of custom error handling for tracer
construction functions. The `ByteSlice` return avoids an awkward split
between return value and out-parameter for the response body.

# Additional Notes

# How to test the change?

```
cargo test -p libdd-data-pipeline-ffi --lib
```

All 40 tests pass (13 tracer + 3 response + 24 trace_exporter).

### DIFF
--- a/libdd-data-pipeline-ffi/src/response.rs
+++ b/libdd-data-pipeline-ffi/src/response.rs
@@ -36,10 +36,7 @@ impl From<AgentResponse> for ExporterResponse {
 pub unsafe extern "C" fn ddog_trace_exporter_response_get_body<'a>(
     response: Option<&'a ExporterResponse>,
 ) -> ByteSlice<'a> {
-    match response.and_then(|r| r.body.as_deref()) {
-        Some(body) => ByteSlice::from(body),
-        None => ByteSlice::default(),
-    }
+    response.and_then(|r| Some(ByteSlice::from(r.body.as_deref()?))).unwrap_or_default()
 }
 
 /// Free `response` and all its contents. After being called response will not point to a valid
--- a/libdd-data-pipeline-ffi/src/response.rs
+++ b/libdd-data-pipeline-ffi/src/response.rs
@@ -3,8 +3,8 @@
 
 //! Define FFI compatible AgentResponse struct
 
+use libdd_common_ffi::slice::ByteSlice;
 use libdd_data_pipeline::trace_exporter::agent_response::AgentResponse;
-use std::ptr::null;
 
 /// Structure containing the agent response to a trace payload
 /// MUST be freed with `ddog_trace_exporter_response_free`
@@ -29,27 +29,17 @@ impl From<AgentResponse> for ExporterResponse {
     }
 }
 
-/// Return a read-only pointer to the response body. This pointer is only valid as long as
-/// `response` is valid.
+/// Return a borrowed view of the response body.  The returned slice is
+/// only valid as long as `response` is alive.  Returns an empty slice
+/// when `response` is null or the body is absent.
 #[no_mangle]
-pub unsafe extern "C" fn ddog_trace_exporter_response_get_body(
-    response: *const ExporterResponse,
-    out_len: Option<&mut usize>,
-) -> *const u8 {
-    let mut len: usize = 0;
-    let body = if response.is_null() {
-        null()
-    } else if let Some(body) = &(*response).body {
-        len = body.len();
-        body.as_ptr()
-    } else {
-        null()
-    };
-
-    if let Some(out_len) = out_len {
-        *out_len = len;
+pub unsafe extern "C" fn ddog_trace_exporter_response_get_body<'a>(
+    response: Option<&'a ExporterResponse>,
+) -> ByteSlice<'a> {
+    match response.and_then(|r| r.body.as_deref()) {
+        Some(body) => ByteSlice::from(body),
+        None => ByteSlice::default(),
     }
-    body
 }
 
 /// Free `response` and all its contents. After being called response will not point to a valid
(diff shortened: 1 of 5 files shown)


lloeki:
There are some CI failures but I'm opening this for review.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -77,40 +74,40 @@ pub struct TracerSpanFields<'a> {
...
-
-                out_handle.as_ptr().write(Box::new(TracerSpan(span)));
-                Ok(())
-            };
-            inner().err()
-        },
-        gen_error!(ErrorCode::Panic)
-    )
+    fields: Option<&TracerSpanFields>,
+) -> MaybeError {
```
hoolioh: I don't see the reason to split the error handling API in two, I think it could lead to confusion since the consumers would have to use two types and two set of functions when dealing with errors.
The only advantage is that when an error is present there is one less indirection to get the error string but that's in the cold path. On the other hand we're taking a performance hit in the hot path, with the previous model just a pointer (8B on 64bit platforms) is returned so it can be hold in one register as oppossed to to MaybeError which is 4 times larger. It'd be probably a few cycles per call, probably not really noticeable if the function is not heavily used but still I don't see any justification to increase the API surface.

libdd-data-pipeline-ffi/src/response.rs
```diff
@@ -29,27 +29,17 @@ impl From<AgentResponse> for ExporterResponse {
...
-        body.as_ptr()
-    } else {
-        null()
-    };
-
-    if let Some(out_len) = out_len {
-        *out_len = len;
+pub unsafe extern "C" fn ddog_trace_exporter_response_get_body<'a>(
+    response: Option<&'a ExporterResponse>,
+) -> ByteSlice<'a> {
```
hoolioh, line 38: It coudl be a good quality of life improvement. Nit: since this is a breaking change I would flag the PR as `feat!` so the person doing the release can now this commit introduced a breaking change in the existing API.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -209,17 +197,12 @@ pub struct TracerTraceChunks(Vec<Vec<SpanBytes>>);
...
-            let chunks = Vec::with_capacity(capacity);
-            out_handle
-                .as_ptr()
-                .write(Box::new(TracerTraceChunks(chunks)));
-            None
-        },
-        gen_error!(ErrorCode::Panic)
-    )
+) -> MaybeError {
+    let chunks = Vec::with_capacity(capacity);
```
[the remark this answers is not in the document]
hoolioh: Aside from the error handling split explained above, I would say this is a valid concern, this function can panic via a malicious/erroneous parameter so we should catch it.

hoolioh (changes requested):

libdd-data-pipeline-ffi/src/response.rs
```diff
@@ -29,27 +29,17 @@ impl From<AgentResponse> for ExporterResponse {
...
+    match response.and_then(|r| r.body.as_deref()) {
+        Some(body) => ByteSlice::from(body),
+        None => ByteSlice::default(),
```
yannham:
```suggestion
    response
        .and_then(|r| Ok(Body::from(r.body.as_deref()?)))
        .unwrap_or_default()
```

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -172,20 +163,17 @@ pub unsafe extern "C" fn ddog_tracer_span_set_metric(
...
-            };
-            span.0.metrics.insert(key, value);
-            None
-        } else {
-            gen_error!(ErrorCode::InvalidArgument)
-        },
-        gen_error!(ErrorCode::Panic)
-    )
+) -> MaybeError {
+    let inner = || -> anyhow::Result<()> {
```
yannham: Same, I would keep the `catch_panic`, just in case. There are a bunch of `std` functions called here

yannham (commented):
I agree with @hoolioh, and in general we should keep the `catch_panic`

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -273,18 +275,19 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_begin_chunk(
...
+    span: Option<Box<TracerSpan>>,
 ) -> Option<Box<ExporterError>> {
     catch_panic!(
-        if let Some(chunks) = handle {
-            if let Some(chunk) = chunks.0.last_mut() {
-                chunk.push(span.0);
-                None
-            } else {
-                gen_error!(ErrorCode::InvalidArgument)
+        match (handle, span) {
```
yannham: Very nit: why the change from the `if let` to `match`? I think clippy tends to suggest the former when the catch-all case is of the form `_ => ...`. By the way, I think a `let-else`, as used below, would be even more idiomatic/cleaner for all those "is this optional `None`" tests.

yannham (approved):
Left some nits, but LGTM otherwise 👍

hoolioh (approved):

PR #2030, opened 2026-05-22
feat(trace-utils)!: refactor trace normalization to use in TraceExporter

Eldolfin:
# What does this PR do?

Try to make trace normalization work on the spans we use in the TraceExporter too, the goal is to normalize spans client-side to be used by trace-filters and send them already normalized to the Agent.

# Motivation

# Additional Notes

# How to test the change?


Eldolfin:
To explore later, too complicated only to use in trace filters for now.

PR #2036, opened 2026-05-26
refactor(crashtracker): update imports to linux only

anais-raison:
# What does this PR do?

Put some import under linux os only.

# Motivation

These imports were falsely reported as unused on my IDE.

# Additional Notes

# How to test the change?

### DIFF
--- a/bin_tests/tests/crashtracker_bin_test.rs
+++ b/bin_tests/tests/crashtracker_bin_test.rs
@@ -25,9 +25,6 @@ use libdd_crashtracker::{
 };
 use serde_json::Value;
 
-#[cfg(target_os = "linux")]
-use libdd_crashtracker::default_max_threads;
-
 /// Macro to generate simple crash tracking tests using the new infrastructure.
 /// This replaces 16+ nearly identical test functions with a single declaration.
 macro_rules! crash_tracking_tests {
@@ -315,7 +312,7 @@ fn test_crash_tracking_multi_thread_collection() {
 #[cfg(target_os = "linux")]
 #[cfg_attr(miri, ignore)]
 fn test_crash_tracking_thread_limit() {
-    const THREAD_COUNT: usize = default_max_threads();
+    const THREAD_COUNT: usize = libdd_crashtracker::default_max_threads();
 
     let config = CrashTestConfig::new(
         BuildProfile::Release,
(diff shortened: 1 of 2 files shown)


Aaalibaba42 (approved):

bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -21,11 +21,13 @@ use bin_tests::{
     ArtifactsBuild, BuildProfile,
 };
 use libdd_crashtracker::{
-    default_max_threads, CrashtrackerConfiguration, Metadata, SiCodes, SigInfo, SignalNames,
-    StacktraceCollection,
+    CrashtrackerConfiguration, Metadata, SiCodes, SigInfo, SignalNames, StacktraceCollection,
 };
 use serde_json::Value;
 
+#[cfg(target_os = "linux")]
```
hoolioh: Since it's only used in one place wouldn't it be cleaner to inline the path at the call site. E.g: `libdd_crashtracker::default_max_threads()`

bin_tests/tests/crashtracker_bin_test.rs
```diff
@@ -21,11 +21,13 @@ use bin_tests::{
...
+use libdd_crashtracker::default_max_threads;
```
yannham: This one could probably be removed, since `default_max_threads` is used only once in this file, using `libdd_crashtracker::default_max_threads` instead at the callsite.

libdd-crashtracker/src/receiver/receive_report.rs
```diff
@@ -18,6 +18,9 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 use tokio::io::AsyncBufReadExt;
 
+#[cfg(target_os = "linux")]
+use crate::crash_info::Threads;
```
hoolioh: Same as above

yannham: This import is used in one function only, which is linux-gated, and has a bunch of imports already. Maybe we could move the `use` there, to avoid cluttering the top-level (and get rid of a `#[cfg(..)]`)?

hoolioh (approved):
LGTM but I left a couple of comments that will make the PR cleaner.

yannham (approved):

PR #2037, opened 2026-05-26
feat(trace-exporter): add fail-closed fallback to v04

anais-raison:
# What does this PR do?

Adds runtime V1 trace protocol negotiation with fail-closed fallback to V0.4. Opt in via `enable_v1_protocol()` or `DD_TRACE_AGENT_PROTOCOL_VERSION=1`. V1 is only used after the agent advertises `/v1.0/traces` in `/info`.

# Motivation

[APMSP-2809](https://datadoghq.atlassian.net/browse/APMSP-2809)
Makes V1 safe to enable: falls back to V0.4 against agents that don't support it.

# Additional Notes

Dynamic rollback supported (if the agent stops advertising V1, the exporter switches back to V0.4).

# How to test the change?

All test are working.
End to end tests against a real agent were done and successful as well.

[APMSP-2809]: https://datadoghq.atlassian.net/browse/APMSP-2809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -600,6 +627,7 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
             header_tags,
             &self.metadata,
             self.agent_payload_response_version.as_ref(),
+            self.effective_output_format(),
```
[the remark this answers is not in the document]
ekump: @anais-raison I believe this is a valid concern.

What is the expected behavior? I assume the common scenario is that we don't have the response from `/info` yet so we fallback to v04 until agent_info is populated. 

To solve for that scenario you probably want to capture the effective output at the start and pass that value through the call chain so we don't wind up serializing for one format but sending to the other format's endpoint. 

Is it possible that after we get a response from `/info` that says V1 is supported that the agent could later change that to false? If that's possible, the fix becomes more complicated because we have to consider the retry logic, which currently doesn't re-check for v1 support.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -498,6 +513,16 @@ impl TraceExporterBuilder {
         })
     }
 
+    /// Returns true when the `DD_TRACE_AGENT_PROTOCOL_VERSION` environment variable opts in to
+    /// the V1 protocol. Accepts `"1"` and `"1.0"`. Other values (including unset) yield false.
+    fn env_requests_v1_protocol() -> bool {
+        Self::parse_v1_protocol_env(std::env::var("DD_TRACE_AGENT_PROTOCOL_VERSION").ok())
```
ekump: We should not be checking env vars in libdatadog (normally). This should be handled by the SDK and passed via the builder.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -390,6 +397,26 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
+    /// Reconcile `v1_active` with the agent's currently-advertised endpoints. Called only when
+    /// V1 is configured and the agent info state has changed, so transitions are logged at most
+    /// once per change.
+    #[cfg(not(target_arch = "wasm32"))]
+    fn refresh_v1_active(&self, agent_info: &Arc<AgentInfo>) {
+        let supports_v1 = agent_info
+            .info
+            .endpoints
+            .as_ref()
+            .is_some_and(|e| e.iter().any(|p| p == "/v1.0/traces"));
```
ekump: `/v1.0/traces` should probably be a named const.

libdd-data-pipeline/tests/test_trace_exporter.rs
```diff
@@ -314,6 +314,17 @@ mod tracing_integration_tests {
 
             let data = get_v04_to_v1_trace_snapshot_test_payload("test_exporter_v04_v1_snapshot");
 
+            // V1 is gated by /info negotiation (fail-closed). Wait until the background
+            // fetcher has populated agent_info so the first send promotes v1_active=true
+            // and the payload is encoded as V1.
+            let start = std::time::Instant::now();
+            while libdd_data_pipeline::agent_info::get_agent_info().is_none() {
+                if start.elapsed() > std::time::Duration::from_secs(5) {
```
ekump, line 322: I think this is probably ok, but we need to be careful anytime we introduce waiting to tests. Could we try to stress test this a bit to see if it's flaky on CI? I'd suggest temporarily modifying the CI workflow to run the [tracing integration step](https://github.com/DataDog/libdatadog/blob/af9b62bc4f968a4754637dc8f5abfa8a336e96a7/.github/workflows/test.yml#L170) 1000x on the various platforms in our matrix in a loop (the loop should only apply to the integration tests, not the entire suite) to see if any flakes appear.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -390,6 +397,26 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
+        let previous = self.v1_active.swap(supports_v1, Ordering::Relaxed);
+        match (previous, supports_v1) {
```
ekump, line 420: Should we also log `(false, false)`? Or is there somewhere else we should log the scenario where the SDK is requesting V1, but the agent doesn't support it? I'm not sure if we care about that?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -390,6 +397,26 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
+            (false, true) => debug!("V1 trace protocol enabled (agent advertises /v1.0/traces)"),
+            (true, false) => {
+                warn!("V1 trace protocol no longer advertised by agent; falling back to v0.4")
```
ekump, line 423: Related to [this comment](https://github.com/DataDog/libdatadog/pull/2037/changes#r3306641082), when would this happen?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -347,6 +351,9 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
     fn check_agent_info(&self) {
         if let Some(agent_info) = agent_info::get_agent_info() {
             if self.has_agent_info_state_changed(&agent_info) {
```
ekump, line 360: I might be concern trolling here...but if a likely scenario we have to support is the agent advertising that it supports V1 to advertising it does not support v1 then it's possible we get trapped in a loop where we don't get a correct state hash from the agent. 

1) Agent says it's ok to send V1, we send V1.
2) Agent changes its mind and does not support V1
3) We check if agent state has changed based on the last successful response that happened before the agent changed its mind.
4) We send V1, get a 404 because the endpoint isn't available. We don't get a new state hash on a 404.
5) We keep sending V1, keep 404'ing, never get a new state hash.

We poll `/info` every 5 minutes independently of the state hash. But 5 minutes is a long time to be dropping traces. 

I think this would also apply to CSS, which has the same logic. Is this a valid concern?

CC @VianneyRuhlmann and @ajgajg1134

ajgajg1134: hmm, yeah I think it's reasonable to say that a 404 response from the agent at any point should probably invalidate the info state cache, especially for something like the newer v1 endpoint where a 404 would likely indicate the customer rolled back their agent to a non-v1 version or restarted it with the v1 endpoint explicitly disabled.

I will note that this is probably an area that we should have a clearer spec (and tests) for across all the languages since this issue is not unique to libdatadog. E.g. the go tracer just fetches once on startup and every 5s afterwards, it doesn't look at the header values at all nor does it have any special "invalidation" logic on a failed request like a 404

libdd-data-pipeline/tests/test_trace_exporter.rs
anais-raison: I added a temporary step to stress test it 👍

libdd-data-pipeline/src/trace_exporter/mod.rs
anais-raison: I added a hook in `send_trace_chunks_inner` that catches 404s on `/v1.0/traces`: when that happens we immediately flip `v1_active` to false so next `send` fallback to v0.4, and call `info_response_observer.manual_trigger()` to force an `/info` refetch instead of waiting for the next 5-minute poll. So the trap closes on the first request after a rollback rather than the next scheduled poll.

libdd-data-pipeline/src/trace_exporter/mod.rs
anais-raison: Logging every `(false, false)` would be noisy in the common "agent never supports V1" case. We could gate it behind an AtomicBool so we emit a single `warn!` the first time we see it but I don't think it's that necessary or useful.

libdd-data-pipeline/src/trace_exporter/mod.rs
anais-raison: There is three cases i can think about :
1. Agent rolled back to a version without V1
2. Agent restarted with V1 disabled via config
3. The new 404 fail-closed hook (on a 404 from /v1.0/traces we flip v1_active directly and trigger an /info refresh)

libdd-data-pipeline/src/trace_exporter/mod.rs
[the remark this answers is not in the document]
anais-raison: We now snapshot `effective_output_format()` once at the top of `send_trace_chunks_inner` and reuse it for both the serializer and the URL, so the payload and endpoint can't disagree mid-send.

libdd-data-pipeline/src/trace_exporter/mod.rs
ekump: I agree logging every `(false, false)` is not helpful. I was thinking logging it at least once could be helpful for debugging, but realistically the only reason that would happen is if the agent is version is rolled back, which we should be able to tell via other signals. We can leave it as is.

ekump (approved):
LGTM, just don't forget to remove the temporary code in the GH workflow

.github/workflows/test.yml
```diff
@@ -193,6 +193,42 @@ jobs:
           fi
         env:
           RUST_BACKTRACE: full
+      # TEMPORARY — added for PR #2037 to stress-test the /info wait loop in
+      # compare_v04_to_v1_trace_snapshot_test. REVERT this step before merging once
+      # we have confirmed the integration tests are not flaky on CI.
```
Aaalibaba42: According to the comment, this ought be left out before merging

libdd-data-pipeline/src/trace_exporter/mod.rs
Aaalibaba42: > We could gate it behind an AtomicBool

Logging into a `std::Once` is probably cleaner, if effectively the same.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -613,18 +650,39 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
+        // V1 support is re-detected as soon as the agent advertises it again.
+        if effective_format == TraceExporterOutputFormat::V1 {
+            if let Err(TraceExporterError::Request(ref e)) = result {
+                if e.status() == http::StatusCode::NOT_FOUND {
+                    if self.v1_active.swap(false, Ordering::Relaxed) {
+                        warn!(
+                            "V1 trace send returned 404; agent no longer advertises {V1_TRACES_ENDPOINT} — falling back to V0.4"
+                        );
+                    }
+                    self.info_response_observer.manual_trigger();
```
Aaalibaba42: If `v1_active` was already false before, why do the `manual_trigger` right away ?

anais-raison: Good catch 👍

PR #2039, opened 2026-05-26
feat(trace-exporter): add v1 span and its encoder

anais-raison:
# What does this PR do?

Adds the canonical V1 trace payload data model (`v1::Span`, `TracerPayload`, `TraceChunk`, typed `AttributeValue`, `SpanKind`, 128-bit `trace_id`, `error: bool`) and a matching msgpack encoder that serializes this model into the V1 wire format. New public APIs: `to_vec_from_payload`, `to_vec_from_payload_with_capacity`, `write_payload_to_slice`, `to_encoded_byte_len_from_payload`.

# Motivation

[APMSP-2810](https://datadoghq.atlassian.net/browse/APMSP-2810), M3 of the V1 trace payload rollout. Tracers will eventually produce V1 payloads natively rather than going through the v0.4→V1 conversion path implemented in [#1896](https://github.com/DataDog/libdatadog/pull/1896), so libdatadog needs a canonical V1 data model and a corresponding encoder.

# Additional Notes

- The V1 wire format keys (`SpanKey`, `SpanLinkKey`, `SpanEventKey`, `AnyValueKey`, plus the `trace_key` and `chunk_key` modules) live in `msgpack_encoder/v1/mod.rs` and are shared between the v0.4→V1 encoder and the v1::Span encoder.
- Output is byte-identical to the M1 v0.4→V1 encoder for equivalent inputs.

[APMSP-2810]: https://datadoghq.atlassian.net/browse/APMSP-2810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -864,3 +1187,943 @@ mod tests {
         );
     }
 }
+
+#[cfg(test)]
+mod v1_payload_tests {
+    //! Unit tests for the M3 encoder (`encode_payload_v1`).
```
ajgajg1134: Does M3 mean milestone 3? Is there a better name we can use here?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -864,3 +1187,943 @@ mod tests {
...
+                encoded.windows(2).any(|w| w == want),
+                "SpanKind {kind:?} should produce byte {expected_byte:#x}"
+            );
+        }
+    }
+
+    #[test]
+    fn string_interning_works_across_chunks() {
+        // The string "shared" appears in two chunks. The second occurrence must be a uint ID,
+        // not a fresh str. Compare against a baseline with a single occurrence to verify.
```
ajgajg1134: this comment implies this test asserts on the actual payload itself that the `shared` string should only appear exactly once and the second reference to it is via the index but this test doesn't seem to actually assert that?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -864,3 +1187,943 @@ mod tests {
...
+        );
+    }
+}
+
+#[cfg(test)]
+mod cross_validation_tests {
+    //! Cross-validates that the M1 encoder (v0.4 spans → V1 payload) and the M3 encoder
+    //! (v1::Span → V1 payload) produce **byte-identical** output for equivalent inputs.
+    //!
+    //! All tests are limited to deterministic content (at most one attribute key per map) so the
```
ajgajg1134: Is there a way we could swap in (just for testing) a consistent ordered hash map to make these tests more versatile?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -864,3 +1187,943 @@ mod tests {
...
+
+    use super::*;
+    use crate::span::v04::SpanBytes as V04Span;
+    use crate::span::v1::{
+        AttributeValue, SpanBytes as V1SpanBytes, SpanKind, TraceChunkBytes, TracerPayloadBytes,
+    };
+    use libdd_tinybytes::BytesString;
+
+    fn bs(s: &str) -> BytesString {
+        BytesString::from_slice(s.as_bytes()).unwrap_or_default()
```
ajgajg1134: nit: these are unit tests so maybe we should just use unwrap (or expect) to make it clearer if someone ever writes a test using a str value that fails here? (But also feel free to double check my thought here with someone who writes more rust :P )

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -864,3 +1187,943 @@ mod tests {
...
+    fn tid_bytes(high: u64, low: u64) -> [u8; 16] {
+        let mut out = [0u8; 16];
+        out[..8].copy_from_slice(&high.to_be_bytes());
+        out[8..].copy_from_slice(&low.to_be_bytes());
+        out
+    }
+
+    /// Asserts that encoding `v04` (with `metadata`) via M1 produces the same bytes as
+    /// encoding `v1` via M3. Includes a hex-diff message on mismatch.
+    #[track_caller]
```
ajgajg1134: TIL about "track_caller", very cool 😎

ajgajg1134 (changes requested):
Overall this looks good from a v1 perspective but definitely someone with more Rust review experience should take a peek.

One small idea (probably not necessary for this specific PR), can we implement a round-trip fuzzer here to verify that if we start with a valid payload and encode -> decode it we end up with the same thing? Or something like that would help build confidence that we have consistent implementations

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: Yes completely, forgot to update that to better name

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: Good catch ! Added a byte-scan assertion that counts literal occurrences of `b"shared"` in the encoded two-chunk payload and requires it to be exactly 1. Kept the size comparison as a complementary sanity check.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
anais-raison: It's do able but probably not worth it right now imo. We could do :
- A test-only feature flag that use BTreeMap/IndexMap instead of HashMap in the encoder's storage type.
- Or sort entries at encode time in both encoders.

One add complexity and the other one change production behavior for a temporary concern: once tracers switch to native span, and thus to full V1, the v0.4 to V1 encoder goes away and this whole cross-validation suite would go with it.

ajgajg1134 (approved):

libdd-trace-utils/src/span/v1/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+/// Replaces v0.4's split `meta` / `metrics` / `meta_struct` maps.
+#[derive(Debug, PartialEq)]
+pub enum AttributeValue<T: TraceData> {
+    String(T::Text),
+    Float(f64),
+    Int(i64),
+    Bool(bool),
+    Bytes(T::Bytes),
+    KeyValue(HashMap<T::Text, AttributeValue<T>>),
+    List(Vec<AttributeValue<T>>),
```
paullegranddc: Might be interesting in term of performance to Box these variants as they are way less likely to happen

libdd-trace-utils/src/span/v1/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub span_id: u64,
+    pub parent_id: u64,
+    pub start: i64,
+    pub duration: i64,
+    pub error: bool,
+    pub span_kind: SpanKind,
+    pub env: T::Text,
+    pub version: T::Text,
+    pub component: T::Text,
+    pub attributes: HashMap<T::Text, AttributeValue<T>>,
```
paullegranddc: Worth moving to `VecMap` added by @yannham from the start

libdd-trace-utils/src/span/v1/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+}
+
+/// The generic representation of a V1 span link.
+/// `T` is the type used to represent strings in the span link.
+#[derive(Debug, PartialEq, Default)]
+pub struct SpanLink<T: TraceData> {
+    pub trace_id: [u8; 16],
+    pub span_id: u64,
+    pub attributes: HashMap<T::Text, AttributeValue<T>>,
```
paullegranddc: Same feedback, might be worth using `VecMap`

libdd-trace-utils/src/span/v1/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub span_kind: SpanKind,
+    pub env: T::Text,
+    pub version: T::Text,
+    pub component: T::Text,
+    pub attributes: HashMap<T::Text, AttributeValue<T>>,
+    pub span_links: Vec<SpanLink<T>>,
+    pub span_events: Vec<SpanEvent<T>>,
+}
+
+impl<T: TraceData> Clone for Span<T>
```
paullegranddc: I'm seeing all of these `Clone` implementation and I'm worndering, where do we actually need to clone spans in the pipeline?

The Clone trait os not added on TraceData, and requiring an extra bound to use v1 span models seems not the right pattern

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -421,6 +477,273 @@ pub fn to_encoded_byte_len<T: TraceData, S: AsRef<[Span<T>]>>(
...
+/// Encodes one V1 chunk (a group of spans sharing a trace ID).
+fn encode_chunk_v1<W: RmpWrite, T: TraceData>(
+    writer: &mut W,
+    chunk: &crate::span::v1::TraceChunk<T>,
+    table: &mut StringTable,
+) -> Result<(), ValueWriteError<W::Error>> {
+    let has_origin = chunk
+        .origin
+        .as_ref()
+        .is_some_and(|o| !<T::Text as Borrow<str>>::borrow(o).is_empty());
```
paullegranddc: Nit:

```suggestion
    let origin = chunk
        .origin
        .as_ref()
        .filter(|o| !<T::Text as Borrow<str>>::borrow(o).is_empty());
```

And then use `origin.is_some()` in fields computation instead of filtering twice? Once here and once line 575 when getting the inner value?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -421,6 +477,273 @@ pub fn to_encoded_byte_len<T: TraceData, S: AsRef<[Span<T>]>>(
...
+    if has_dropped {
+        write_uint8(writer, chunk_key::DROPPED_TRACE)?;
+        rmp::encode::write_bool(writer, true).map_err(ValueWriteError::InvalidDataWrite)?;
+    }
+
+    write_uint8(writer, chunk_key::SPANS)?;
+    write_array_len(writer, chunk.spans.len() as u32)?;
+    for span in &chunk.spans {
+        span_v1::encode_span(writer, span, table)?;
+    }
```
paullegranddc: Maybe write this at the beginning, to make checking that fields len is correct a bit easier

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -421,6 +477,273 @@ pub fn to_encoded_byte_len<T: TraceData, S: AsRef<[Span<T>]>>(
...
+///
+/// let payload = TracerPayloadSlice {
+///     language_name: "rust".into(),
+///     ..Default::default()
+/// };
+/// let encoded = to_vec_from_payload(&payload);
+///
+/// assert!(!encoded.is_empty());
+/// ```
+pub fn to_vec_from_payload<T: TraceData>(payload: &TracerPayload<T>) -> Vec<u8> {
```
paullegranddc:
```suggestion
pub fn to_vec_from_payload_v1<T: TraceData>(payload: &TracerPayload<T>) -> Vec<u8> {
```

It's not obvious from reading the function name that `TracerPayload` is a v1 data model and that this encodes v1 data model to v1 on the wire format

libdd-trace-utils/src/span/v1/mod.rs
yannham: `Vec` and `Hashmap` are already "boxed", right? I mean they're fat pointers, but it's not like the structure is inlined. I'm not sure adding an indirection is worth saving 16 bytes...

paullegranddc: Depends, T::Text can be quite big (32 bytes for the ByteStr type) in which case boxing is useless as this is the same size as a Hasmap.

But I do think that having a very infrequent indirection to save 16 bytes per attribute (of text is a &str for instance ) can be a winning tradeoff

paullegranddc: We could use a thin vec and a thin map too

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
```diff
@@ -0,0 +1,326 @@
...
+///
+/// * `writer` - A RmpWriter compatible with rmp writing functions.
+/// * `value` - The attribute value to encode.
+/// * `table` - The streaming string intern table used for interning string values.
+///
+/// # Returns
+///
+/// * `Ok(())` - Nothing if successful.
+/// * `Err(ValueWriteError)` - An error if the writing fails.
+///
```
yannham, line 30: Very nitpick: this feels redudant with `# Errors`, which I think is sufficient (same for other functions of this file)

```suggestion
```

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
```diff
@@ -0,0 +1,326 @@
...
+/// # Errors
+///
+/// This function will return any error emitted by the writer.
+pub(super) fn encode_attributes_map<W: RmpWrite, T: TraceData>(
+    writer: &mut W,
+    map: &std::collections::HashMap<T::Text, AttributeValue<T>>,
+    table: &mut StringTable,
+) -> Result<(), ValueWriteError<W::Error>> {
+    write_array_len(writer, (map.len() as u32) * 3)?;
```
yannham: Where does this `3` come from? I would suggest adding a comment here for readers without the spec at hand.

libdd-trace-utils/src/span/v1/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+            AttributeValue::Bytes(v) => AttributeValue::Bytes(v.clone()),
+            AttributeValue::KeyValue(m) => AttributeValue::KeyValue(m.clone()),
+            AttributeValue::List(v) => AttributeValue::List(v.clone()),
+        }
+    }
+}
+
+/// The generic representation of a V1 span.
+///
+/// `T` is the type used to represent strings in the span, it can be either owned (e.g.
```
yannham: Nitpick, but this comments read as if `T` was actually `T::Text`. It's `T::Text` that is either `BytesString` or `&str`. `T` is the type used to represent...trace data? I'm not sure how to best describe it 🙃

libdd-trace-utils/src/span/v1/mod.rs
yannham: (If you do, you'll just have to be careful about deduplicating before serialization)

yannham (approved):

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
```diff
@@ -0,0 +1,326 @@
...
+    if !span.resource.borrow().is_empty() {
+        write_uint8(writer, SpanKey::Resource as u8)?;
+        table.write_interned(writer, span.resource.borrow())?;
+    }
+
+    write_uint8(writer, SpanKey::SpanId as u8)?;
+    write_u64(writer, span.span_id)?;
+
+    write_uint8(writer, SpanKey::Start as u8)?;
+    write_u64(writer, span.start as u64)?;
```
hoolioh: Shouldn't we mimic the same behaviour as in the v04 span where if the start is negative we default to wall-clock `now`?

libdd-trace-utils/src/span/v1/mod.rs
yannham: Fair enough. I guess a thinvec/thinmap is probably the ideal solution, although it does pull in new dependencies.

libdd-trace-utils/src/span/v1/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+}
+
+/// A V1 trace chunk: a group of spans sharing the same `trace_id`, plus chunk-level metadata.
+#[derive(Debug, PartialEq, Default)]
+pub struct TraceChunk<T: TraceData> {
+    pub trace_id: [u8; 16],
+    pub priority: Option<i32>,
+    pub origin: Option<T::Text>,
```
VianneyRuhlmann: Most text fields even optional ones e.g. `name` use `""` to represent the absence of value. Is there any reason to use an Option here ?

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
```diff
@@ -0,0 +1,326 @@
...
+            }
+        }
+        AttributeValue::KeyValue(map) => {
+            write_uint8(writer, AnyValueKey::KeyValueList as u8)?;
+            write_map_len(writer, map.len() as u32)?;
+            for (k, v) in map {
+                table.write_interned(writer, k.borrow())?;
+                encode_attribute_value(writer, v, table)?;
+            }
+        }
```
ekump, line 73: Are you setting the correct lengths for `List` and `KeyValue`? Shouldn't it be `arr.len() * 2` and map.len() * 3` respectively since we're including types?

libdd-trace-utils/src/span/v1/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+/// }
+/// ```
+#[derive(Debug, PartialEq, Default)]
+pub struct Span<T: TraceData> {
+    pub service: T::Text,
+    pub name: T::Text,
+    pub resource: T::Text,
+    pub r#type: T::Text,
+    /// 128-bit trace ID stored as big-endian bytes.
+    pub trace_id: [u8; 16],
```
ekump: Do you need `trace_id` on the span? It doesn't seem to be used anywhere today.

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
ekump: @yannham - Types are included in v1 encoding so the array len is larger than the logical "number of items". It's 2 for arrays and 3 for maps (key, type, val).

I'd make this a named const since the value is used in multiple places.

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
ekump: Wouldn't this completely break decoding in the agent? Are we not decoding these payloads in the test-agent?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
ekump: I'm not comfortable with constraining collections to size n=1. There are potential edge cases we aren't catching. 

I would question what the benefit of byte-identical tests are? It's not something that actually matters IRL (as evident by the n=1 constraint on collections in the first place). Wouldn't it be better for the tests to be structured so that the two payloads are decoded and verified to logically match? You could sort the decoded collections to assert they match.

anais-raison: @ekump Here is the solution I implemented => replaced the byte-equality `cross_validation_tests` with a single integration test `v04_and_v1_encoders_produce_equivalent_decoded_traces` in `tests/test_send_data.rs` that :
- Builds equivalent v0.4 and v1::Span payloads
- Encodes both via their respective encoders
- POSTs both to ddapm-test-agent's `/v1.0/traces`
- The test-agent groups them by trace_id and we assert the two decoded spans are structurally equal after recursive key normalization

This is a real round-trip check (no n=1 constraint) and doubles as the "round-trip fuzzer" you suggested in your overall comment @ajgajg1134.

libdd-trace-utils/src/span/v1/mod.rs
anais-raison: Applied `thin_vec::ThinVec` for `AttributeValue::List`, `Span.span_links`, and `Span.span_events` but kept `VecMap` for `KeyValue` for now.

libdd-trace-utils/src/span/v1/mod.rs
anais-raison: I didn't really though about it, added it at first to follow the v04 implem but it's removed now.

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
anais-raison: I don't have a strong opinion on this but I did follow the exact same format for comments that was use for v04. Any opinion anyone else ?

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
anais-raison: Added named consts 👍

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
anais-raison: True, it's done

libdd-trace-utils/src/span/v1/mod.rs
anais-raison: Not really, I just went with it, updated it 👍

libdd-trace-utils/src/span/v1/mod.rs
yannham: Sorry for the back-and-forth, but I would exclude `thin_vec` for now then. Since the larger variant wins, it's not really worth pulling a dependency since the limiting one is `VecMap` now.

Honestly whichever solution you chose is probably fine, the performance difference isn't going to be huge and is also quite workflow-dependent. This can easily be changed later if we want.

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
yannham: Ok, then let's keep it that way then :+1:

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
anais-raison: Yes I updated it

libdd-trace-utils/src/span/v1/mod.rs
anais-raison: Okay, back to `Vec` for `KeyValue` but I kept `ThinVec` for `SpanLinks` and `SpanEvents` because in those case I think the tradeoff is worth it.

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -315,6 +339,12 @@ fn encode_payload<W: RmpWrite, T: TraceData, S: AsRef<[Span<T>]>>(
 
     write_map_len(writer, map_len)?;
 
+    write_uint8(writer, trace_key::CHUNKS)?;
+    write_array_len(writer, traces.len() as u32)?;
+    for trace in traces {
+        encode_chunk(writer, trace.as_ref(), &mut table)?;
+    }
```
VianneyRuhlmann, line 346: Out of curiosity what's the logic behind writing the chunks/spans first instead of at the end of the payload ?

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -134,6 +143,21 @@ impl StringTable {
     }
 }
 
+/// Returns the span start time in UNIX nanos, falling back to the current wall-clock time when
+/// the input is negative. Matches the agent's `validateAndFixStartTime`, which substitutes
+/// `time.Now().UnixNano()` for invalid start values; without this, a negative `i64` would wrap
+/// to a near-`u64::MAX` timestamp on cast.
+pub(super) fn span_start_unix_nanos(start: i64) -> u64 {
```
VianneyRuhlmann: nit: I think the name is a bit misleading, span start is already in unix nanos. Maybe this should be `normalize_span_start` or `sanitize_span_start`

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
```diff
@@ -90,11 +94,24 @@ pub(super) fn encode_attribute_value<W: RmpWrite, T: TraceData>(
...
+    // Walk in reverse keeping first-seen (= last-written), then reverse to restore insertion
+    // order. `T::Text: Hash + Eq + Borrow<str>` per the `SpanText` trait.
+    let mut seen: HashSet<&str> = HashSet::with_capacity(map.len());
+    let mut deduped: Vec<(&T::Text, &AttributeValue<T>)> = map
+        .iter()
+        .rev()
+        .filter(|(k, _)| seen.insert(<T::Text as Borrow<str>>::borrow(k)))
+        .map(|(k, v)| (k, v))
+        .collect();
+    deduped.reverse();
```
VianneyRuhlmann: You should the `defensive_dedup` method of VecMap ? Also you don't need to reverse the insertion order since map keys are not ordered.

libdd-trace-utils/tests/test_send_data.rs
```diff
@@ -392,4 +392,286 @@ mod tracing_integration_tests {
...
+    }
+
+    /// Builds a TracerPayload that exercises the multi-key attribute paths the v0.4→V1 encoder
+    /// can't cover on its own (HashMap iteration order makes byte-by-byte cross-validation flaky
+    /// for n &gt; 1), plus the primitive `AttributeValue` variants the test-agent currently
+    /// supports.
+    ///
+    /// NOTE: `AttributeValue::List` and `AttributeValue::KeyValue` are deliberately omitted —
+    /// `ddapm-test-agent` v1.56.0 returns `400: Array of strings values are not supported yet`
+    /// for the `Array` variant. Once test-agent V1 support catches up, add them here too.
```
VianneyRuhlmann: Either link a jira ticket here or remove the comment if irrelevant

libdd-trace-utils/tests/test_send_data.rs
```diff
@@ -392,4 +392,286 @@ mod tracing_integration_tests {
...
+            response.status().is_success(),
+            "test-agent rejected V1 payload: status={} body={:?}",
+            response.status(),
+            String::from_utf8_lossy(response.body())
+        );
+    }
+
+    /// Builds a TracerPayload that exercises the multi-key attribute paths the v0.4→V1 encoder
+    /// can't cover on its own (HashMap iteration order makes byte-by-byte cross-validation flaky
+    /// for n &gt; 1), plus the primitive `AttributeValue` variants the test-agent currently
```
VianneyRuhlmann: fix `&gt;`

libdd-trace-utils/tests/test_send_data.rs
```diff
@@ -392,4 +392,286 @@ mod tracing_integration_tests {
...
+    }
+
+    /// Asserts the v0.4→V1 encoder and the v1::Span encoder produce decoded traces with the same
+    /// canonical content. Replaces the byte-equality cross-validation suite — instead of
+    /// comparing raw bytes (which forced n=1 collections due to HashMap order non-determinism),
+    /// we let the test-agent decode both payloads and compare the resulting structures after
+    /// recursive key sorting.
+    #[cfg_attr(miri, ignore)]
+    #[tokio::test]
+    async fn v04_and_v1_encoders_produce_equivalent_decoded_traces() {
```
VianneyRuhlmann: This seems quite complicated, wouldn't it be simpler to compare both v04->v1 and v1 with the same test agent snapshot ?

VianneyRuhlmann (approved):

PR #2040, opened 2026-05-26
feat(stats)!: add endpoint gating to client-side stats [APMSP-3361]

VianneyRuhlmann:
# What does this PR do?

Check for the stats endpoint to enable CSS.

Edit: The agent version check was removed due to compatibility with serverless agent.

# Motivation

According to spec client-side stats should only be enabled when agent version is higher than. 7.65 and the stats endpoint is available.

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-data-pipeline/src/trace_exporter/stats.rs
+++ b/libdd-data-pipeline/src/trace_exporter/stats.rs
@@ -79,6 +79,21 @@ pub(crate) struct StatsComputationConfig {
     pub(crate) obfuscation_enabled: bool,
 }
 
+/// Return true if the agent supports client-side stats.
+///
+/// This requires:
+/// - `client_drop_p0s` to be enabled on the agent,
+/// - the `/v0.6/stats` endpoint to be advertised by the agent.
+#[cfg(not(target_arch = "wasm32"))]
+fn is_stats_computation_supported(agent_info: &AgentInfo) -> bool {
+    agent_info.info.client_drop_p0s.is_some_and(|v| v)
+        && agent_info
+            .info
+            .endpoints
+            .as_ref()
+            .is_some_and(|endpoints| endpoints.iter().any(|e| e == STATS_ENDPOINT))
+}
+
 /// Return true if the agent's obfuscation version is supported by this tracer
 #[cfg(feature = "stats-obfuscation")]
 fn is_obfuscation_active(agent_info: &AgentInfo) -> bool {
@@ -196,7 +211,7 @@ pub(crate) fn handle_stats_disabled_by_agent<
     capabilities: C,
     client_side_stats: &StatsComputationConfig,
 ) {
-    if agent_info.info.client_drop_p0s.is_some_and(|v| v) {
+    if is_stats_computation_supported(agent_info) {
         let status = start_stats_computation(
             ctx,
             get_span_kinds_for_stats(agent_info),
@@ -254,7 +269,7 @@ pub(crate) async fn handle_stats_enabled(
     stats_concentrator: &Arc<Mutex<SpanConcentrator>>,
     client_side_stats: &StatsComputationConfig,
 ) {
-    if agent_info.info.client_drop_p0s.is_some_and(|v| v) {
+    if is_stats_computation_supported(agent_info) {
         let mut concentrator = stats_concentrator.lock_or_panic();
         concentrator.set_span_kinds(get_span_kinds_for_stats(agent_info));
         concentrator.set_peer_tags(agent_info.info.peer_tags.clone().unwrap_or_default());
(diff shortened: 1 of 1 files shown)


libdd-data-pipeline/src/agent_info/schema.rs
```diff
@@ -1,8 +1,99 @@
...
+/// Accepts strings of the form `MAJOR.MINOR.PATCH` with an optional
+/// pre-release/build suffix (e.g. `"7.65.0-rc.1+abc"`). Returns `None` if the
+/// field is absent, null, or cannot be parsed. A non-null but unparseable
+/// value is logged at `debug` so operators have a breadcrumb when CSS gating
+/// unexpectedly stays off.
+fn deserialize_agent_version<'de, D>(deserializer: D) -> Result<Option<AgentVersion>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let Some(raw) = Option::<String>::deserialize(deserializer)? else {
```
Eldolfin: Small detail but this implies agent version is always a string, because otherwise the `?` here is the only place we will return an `Err` and thus make the whole schema un-parsable.

In a more general note, I don't like how any parse failure in a field on the schema fails the deserialization of the whole struct because it makes for some really unclear errors. I don't have a solution tho :/

VianneyRuhlmann: The version is always a string https://github.com/DataDog/datadog-agent/blob/8f7d7790726510ee1ecd8e6b7d57cc9c406691bf/pkg/trace/api/info.go#L25 . I don't think we want to partially parse the version here, either it's a valid semver version and we can't rely on it to check for feature support or it's not and we shouldn't try to rely on it

ichinaski (approved):

libdd-data-pipeline/src/agent_info/schema.rs
ajgajg1134: One important note here is that this version can _also_ come from the serverless rust agent, which today is also a string, but greatly complicates the usage of version checking :(

ajgajg1134: e.g. today [the rust agent](https://github.com/DataDog/datadog-lambda-extension/blob/main/bottlecap/src/traces/trace_agent.rs#L444) doesn't return a version at all

libdd-data-pipeline/src/agent_info/fetcher.rs
```diff
@@ -626,17 +624,24 @@ mod single_threaded_tests {
...
-            .unwrap();
-        assert_eq!(version_1, "1");
+            .clone();
+        assert_eq!(
+            version_1,
+            Some(AgentVersion {
+                major: 7,
+                minor: 65,
+                patch: 0,
+                metadata: None,
```
ekump: I don't think you are testing metadata anywhere?

libdd-data-pipeline/src/agent_info/schema.rs
VianneyRuhlmann: I guess it's ok for now to only expect agent version then. They will need a way to let the tracers know which "flavor" of agent they're talking to anyway.

ekump: @ajgajg1134 Can we ask serverless to start returning a version?

ajgajg1134: we could but maybe a bad idea, their version doesn't align with the datadog agent so it would be confusing for their version string to be this weird arbitrary other version. I think in general we probably shouldn't be relying on specific version checks, unfortunately for now to protect us from older broken trace-agent versions we kinda need it. 🤔 
I think the right thing here is that we disable CSS if we can't parse (or there isn't) a version string. I checked with @lucaspimentel and it seems the right choice here is to assume that "no version = no css"

lucaspimentel: There are several serverless workloads than enable CSS by default setting env var `DD_TRACE_STATS_COMPUTATION_ENABLED=true`. Adding an un-conditional version check would break those. Can we skip the version check if there is an explicit `DD_TRACE_STATS_COMPUTATION_ENABLED=true`?

Adding this config check would also avoid an unexpected change for users where CSS is suddenly disabled (not specific to serverless):
- if a user already has CSS enabled via `DD_TRACE_STATS_COMPUTATION_ENABLED=true`
- user updates their tracer, but not their agent (< 7.65)
- if I understand correctly, this new version check will ignore `DD_TRACE_STATS_COMPUTATION_ENABLED` and disable CSS, even though it was working before the update

VianneyRuhlmann: The `DD_TRACE_STATS_COMPUTATION_ENABLED` config was always dependent on the agent advertising support for CSS. Previously support was only checked by the presence of `drop_p0s=true` in /info however according to the spec we also need the a supported agent version >=7.65 and the `/v0.6/stats` endpoint to be advertised. I don't think it's a good idea to allow unconditional CSS to be configured

lucaspimentel: (edit: I posted this comment before refreshing, didn't see the comment above)

I suppose the rust agents can fake it and return version 7.65, but that's even hackier than gating features by using version checks instead of feature detection. 😅

lucaspimentel: > The `DD_TRACE_STATS_COMPUTATION_ENABLED` config was always dependent on the agent advertising support for CSS.

Exactly. Feature detection vs version checks. The whole point of adding features to `/info` was to move away from version checks. Why are we going back? 😞 

>  I don't think it's a good idea to allow unconditional CSS to be configured

I agree. I didn't mean unconditional. I meant skip only the _new_ version check (which is arguably a breaking change for users who already had CSS working with `DD_TRACE_STATS_COMPUTATION_ENABLED=true`), not the check for `drop_p0s=true`.

VianneyRuhlmann: I'm fine with removing the agent version and relying only on `client_drop_p0s` and the endpoint being available. Also I think the [rust agent currently hardcodes](https://github.com/DataDog/datadog-lambda-extension/blob/6ded47f7443ff706e18f1f8a3d1868498d021edc/bottlecap/src/traces/trace_agent.rs#L461) `client_drop_p0s` to be disabled which means CSS won't be enabled regardless of `DD_TRACE_STATS_COMPUTATION_ENABLED`

lucaspimentel: > I think the [rust agent currently hardcodes](https://github.com/DataDog/datadog-lambda-extension/blob/6ded47f7443ff706e18f1f8a3d1868498d021edc/bottlecap/src/traces/trace_agent.rs#L461) `client_drop_p0s` to be disabled

Yes! We had to fix that bug [recently](https://github.com/DataDog/datadog-lambda-extension/pull/1155) because _that_ rust agent (there are two rust agents!) was lying to the tracers, announcing CSS support. For now, we can't enable CSS in AWS Lambda at all, but some tracers started enabling CSS by default.

The workloads that enable CSS use the other rust agent (https://github.com/DataDog/serverless-components/tree/main/crates/datadog-serverless-compat), which reports support for `client_drop_p0s` correctly.

Happy to take this thread to Slack if this is getting too long 😅

PR #2043, opened 2026-05-26
refactor(span)!: use VecMap for `meta`, `metrics` and `meta_struct` for v04 spans

yannham:
Depends on #2022 and #2049.

# What does this PR do?

This PR continues the native span performance work required to land native spans in dd-trace-js. Follow-up of #2022. Actually use `VecMap` in the `Span` data structure.

# Motivation

Performance improvement following dd-trace-js native span experiments. See #2022.

# Additional considerations

There are some deduplication design choices. We deduplicate before serializing to the agent, because while the current agent implementation would support duplicate keys in a map (with the same semantics of "last one wins"), this is not part of the spec/API, so we can't rely on it. We also deduplicate defensively in the msgpack encoder, but it should already be deduped at this stage.

For byte estimate (`byte_size()`), we make the choice of not deduplicating. This means potentially overestimating the size of a chunk (and reaching the limit sooner), in exchange of delaying the deduplication to serialization time. 

# How to test the change?

Run tests.

### DIFF
--- a/datadog-sidecar-ffi/src/span.rs
+++ b/datadog-sidecar-ffi/src/span.rs
@@ -5,6 +5,7 @@ use libdd_common_ffi::slice::{AsBytes, CharSlice};
 use libdd_tinybytes::{Bytes, BytesString};
 use libdd_trace_utils::span::v04::{
     AttributeAnyValueBytes, AttributeArrayValueBytes, SpanBytes, SpanEventBytes, SpanLinkBytes,
+    VecMap,
 };
 use std::borrow::Cow;
 use std::collections::HashMap;
@@ -51,25 +52,35 @@ fn insert_hashmap<V>(map: &mut HashMap<BytesString, V>, key: CharSlice, value: V
 }
 
 #[inline]
-fn remove_hashmap<V>(map: &mut HashMap<BytesString, V>, key: CharSlice) {
+fn insert_vec_map<V>(map: &mut VecMap<BytesString, V>, key: CharSlice, value: V) {
+    if key.is_empty() {
+        return;
+    }
     let bytes_str_key = convert_char_slice_to_bytes_string(key);
-    map.remove(&bytes_str_key);
+    map.insert(bytes_str_key, value);
 }
 
 #[inline]
-fn exists_hashmap<V>(map: &HashMap<BytesString, V>, key: CharSlice) -> bool {
+fn remove_vec_map_slow<V>(map: &mut VecMap<BytesString, V>, key: CharSlice) {
+    let bytes_str_key = convert_char_slice_to_bytes_string(key);
+    map.remove_slow(&bytes_str_key);
+}
+
+#[inline]
+fn exists_vec_map<V>(map: &VecMap<BytesString, V>, key: CharSlice) -> bool {
     let bytes_str_key = convert_char_slice_to_bytes_string(key);
     map.contains_key(&bytes_str_key)
 }
 
 /// The return value is an owned array of slices (`Box<[CharSlice<'a>]>`) that must be dropped
 /// explicitly.
-fn get_hashmap_keys<'a, V>(
-    map: &'a HashMap<BytesString, V>,
+fn get_vec_map_keys<'a, V>(
+    map: &'a VecMap<BytesString, V>,
     out_count: &mut usize,
 ) -> *mut CharSlice<'a> {
-    let mut keys: Vec<&str> = map.keys().map(|b| b.as_str()).collect();
+    let mut keys: Vec<&str> = map.iter().map(|(k, _)| k.as_str()).collect();
     keys.sort_unstable();
+    keys.dedup();
 
     let slices: Box<[CharSlice]> = keys
         .iter()
@@ -182,8 +193,8 @@ pub extern "C" fn ddog_trace_new_span_with_capacities(
     new_vector_push(
         trace,
         SpanBytes {
-            meta: HashMap::with_capacity(meta_size),
-            metrics: HashMap::with_capacity(metrics_size),
+            meta: VecMap::with_capacity(meta_size),
+            metrics: VecMap::with_capacity(metrics_size),
             ..SpanBytes::default()
         },
     )
@@ -324,7 +335,7 @@ pub extern "C" fn ddog_get_span_error(span: &mut SpanBytes) -> i32 {
 
 #[no_mangle]
 pub extern "C" fn ddog_add_span_meta(span: &mut SpanBytes, key: CharSlice, value: CharSlice) {
-    insert_hashmap(
+    insert_vec_map(
         &mut span.meta,
         key,
         BytesString::from_slice(value.as_bytes()).unwrap_or_default(),
@@ -333,7 +344,7 @@ pub extern "C" fn ddog_add_span_meta(span: &mut SpanBytes, key: CharSlice, value
 
 #[no_mangle]
 pub extern "C" fn ddog_del_span_meta(span: &mut SpanBytes, key: CharSlice) {
-    remove_hashmap(&mut span.meta, key);
+    remove_vec_map_slow(&mut span.meta, key);
 }
 
 #[no_mangle]
@@ -355,7 +366,7 @@ pub extern "C" fn ddog_get_span_meta<'a>(
 
 #[no_mangle]
 pub extern "C" fn ddog_has_span_meta(span: &mut SpanBytes, key: CharSlice) -> bool {
-    exists_hashmap(&span.meta, key)
+    exists_vec_map(&span.meta, key)
 }
 
 /// The return value is an owned array of slices (`Box<[CharSlice]>`) that must be freed explicitly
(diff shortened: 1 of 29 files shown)


libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -45,7 +45,7 @@ where
         for (k, v) in &self.meta {
             size += k.as_ref().len() + v.as_ref().len();
         }
-        for k in self.metrics.keys() {
+        for (k, _) in &self.metrics {
```
yannham: TODO: this is wrong, because keys can be duplicated.

Solution 1: don't dedup. The size will be over-estimated.
Solution 2: deduplicate here. Costly (2 times with serialization).
Solution 3: deduplicate upfront (and eg keep a `dirty` flag in the vecmap to avoid double dedup).

datadog-sidecar-ffi/src/span.rs
```diff
@@ -51,25 +52,35 @@ fn insert_hashmap<V>(map: &mut HashMap<BytesString, V>, key: CharSlice, value: V
 }
 
 #[inline]
-fn remove_hashmap<V>(map: &mut HashMap<BytesString, V>, key: CharSlice) {
+fn insert_vec_map<V>(map: &mut VecMap<BytesString, V>, key: CharSlice, value: V) {
+    if key.is_empty() {
+        return;
+    }
+    let bytes_str_key = convert_char_slice_to_bytes_string(key);
+    map.insert(bytes_str_key, value);
```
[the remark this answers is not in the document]
yannham: Indeed, the msgpack encoding doesn't go through `Serialize` and needs to handle deduplication as well.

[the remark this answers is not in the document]
yannham: Done in latest version (`defensive_dedup`).

libdd-trace-utils/src/span/mod.rs
```diff
@@ -21,7 +21,9 @@ use std::{fmt, ptr};
 /// Trait representing the requirements for a type to be used as a Span "string" type.
 /// Note: Borrow<str> is not required by the derived traits, but allows to access HashMap elements
 /// from a static str and check if the string is empty.
-pub trait SpanText: Debug + Eq + Hash + Borrow<str> + Serialize + Default + From<String> {
+pub trait SpanText:
+    Debug + Eq + Hash + Clone + Borrow<str> + Serialize + Default + From<String>
```
yannham: I think it's ok to add `Clone`, since it can always be derived as `From<String> + Borrow<str>` (although in a potentially unoptimal way) ?

libdd-trace-utils/src/span/trace_utils.rs
```diff
@@ -13,16 +13,13 @@ const TRACER_TOP_LEVEL_KEY: &str = "_dd.top_level";
 const MEASURED_KEY: &str = "_dd.measured";
 const PARTIAL_VERSION_KEY: &str = "_dd.partial_version";
 
-fn set_top_level_span<T>(span: &mut Span<T>, is_top_level: bool)
+fn set_top_level_span<T>(span: &mut Span<T>)
```
yannham, line 16: unrelated change, but `is_top_level` is always used with `false`, so... it does get rid of a `remove` call.

libdd-data-pipeline/src/trace_buffer/mod.rs
paullegranddc: Now that we have a `dirty` flag maybe this should be changed to solution 3?

libdd-trace-utils/src/span/mod.rs
paullegranddc: Why do you need it? SpanText used to be Clone until I removed it for the python span text (because cloning without allocation requires holding the GIL)
 I think we don't want people to use this API in general, even if it's derivable as it makes it harder to make performance issues

paullegranddc: Ok I think I found why...
`VecMap::dedup` does `self.data.retain(|(k, _)| seen.insert(k.clone()));` reuiring the Clone trait bound on keys, which I don't think is actually required. You could just insert a ref (I think?)

libdd-trace-utils/src/msgpack_decoder/decode/string.rs
```diff
@@ -25,21 +25,20 @@ pub fn read_nullable_string<T: DeserializableTraceData>(
...
-/// Read a hashmap of (string, string) from the slices `buf`.
+/// Read a vec of (string, string) pairs from the slices `buf`.
 ///
 /// # Errors
 /// Fails if the buffer does not contain a valid map length prefix,
 /// or if any key or value is not a valid utf8 msgpack string.
-/// Null values are skipped (key not inserted into map).
+/// Null values are skipped (key not inserted into vec).
 #[inline]
 pub fn read_str_map_to_strings<T: DeserializableTraceData>(
```
paullegranddc: Maybe this should be renamed to `read_str_map_to_vecmap` , because the name is confusing when compared to `read_str_map_to_hashmap`

libdd-data-pipeline/src/trace_buffer/mod.rs
yannham: As explained in the PR description, I initially thought we shouldn't explicitly dedup here and just hope for the best (still try to dedup before estimation as much as possible, but if not, it's ok to overestimate).

That being said, now that we have `defensive_dedup()`, if we expect spans to be almost always deduped, maybe it's also reasonable to use it here...

libdd-trace-utils/src/span/mod.rs
yannham: It's not possible, because the reference to the keys `retain` gives to your closure you is short-lived. And for a good reason: since retain (or anything else that removes duplicates while keeping the vec contiguous) mutates the vector and move stuff around, inner references are not preserved. One solution is to do it in two passes (remembering in the first which indices to keep or not, e.g in a `Vec<bool>`). I believe it's also possible to implement a custom (probably unsafe) `retain` to do it in one go (you store the reference to the key after you've shifted the first occurrence found, so you know it won't move anymore). Though with the second version we lose some stdlib niceties around panic safety, or we would have to re-implement that part ourselves as well.

yannham: I'm tempted to say let's go with the double pass for now. It's safe, it does allocate a `Vec<bool>` but we need to allocate a `HashSet` already, so it's not like it makes a big difference. Leaving a one-pass, unsafe function for later performance work.

yannham: Implemented in https://github.com/DataDog/libdatadog/pull/2069. Meanwhile, clone bound removed from this PR.

libdd-trace-utils/src/msgpack_decoder/decode/string.rs
```diff
@@ -25,21 +25,21 @@ pub fn read_nullable_string<T: DeserializableTraceData>(
...
+pub fn read_str_map_to_vecmap<T: DeserializableTraceData>(
     buf: &mut Buffer<T>,
-) -> Result<HashMap<T::Text, T::Text>, DecodeError> {
+) -> Result<VecMap<T::Text, T::Text>, DecodeError> {
     let len = decode::read_map_len(buf.as_mut_slice())
         .map_err(|_| DecodeError::InvalidFormat("Unable to get map len for str map".to_owned()))?;
 
-    #[allow(clippy::expect_used)]
-    let mut map = HashMap::with_capacity(len.try_into().expect("Unable to cast map len to usize"));
+    let mut map = VecMap::with_capacity(len.try_into().unwrap_or_default());
```
ekump, line 42: Nice, could we also do something similar [here](https://github.com/DataDog/libdatadog/pull/2043/changes#diff-a6cbc15b129c1eaa44b1620c6c436b19a5b0b658f27f7da144c1b821869c05feR83-R84)?

libdd-data-pipeline/src/trace_buffer/mod.rs
yannham: After discussion, we decided to dedup in the trace buffer before byte estimation for now.

yannham: Implemented in 970bb3713d1d0c0d1b4d2569bce3e341c55fbb85

libdd-trace-utils/src/span/v04/mod.rs
```diff
@@ -66,11 +68,16 @@ fn is_empty_str<T: Borrow<str>>(value: &T) -> bool {
 /// [`SpanValue`] trait:
 /// ```
 /// use libdd_trace_utils::span::{v04::Span, TraceData};
-/// fn foo<T: TraceData>(span: Span<T>) {
+/// use std::borrow::Borrow;
+/// fn foo<T: TraceData>(span: Span<T>)
+/// where
+///     T::Text: Borrow<str>,
+/// {
```
paullegranddc: SpanText already has a Borrow<str> bound I think?

paullegranddc (approved):

PR #2046, opened 2026-05-27
feat(native-spans)!: change buffer foundation

yannham:
# What does this PR do?

Add foundation types for dd-trace-js change buffer: buffer, opcodes and trace. Follow up of #2043.

# Motivation

See #2022. Required for the integration of native spans into dd-trace-js.

# Additional Notes

This PR just introduces the types and doesn't use them yet, hence the `allow(unused)`.

# How to test the change?

Unit tests provided.


libdd-trace-utils/src/change_buffer/buffer.rs
```diff
@@ -0,0 +1,179 @@
+use crate::change_buffer::utils::*;
+use crate::change_buffer::{ChangeBufferError, Result};
+
+/// A handle to a change buffer shared with another runtime. The memory is shared, meaning that
+/// cloning is cheap (copying the pointer and length).
+#[derive(Clone, Copy)]
```
ekump: Would it be better for this to be only `Clone` to prevent any accidental implicit copying?

Actually, does this need to even be `clone`?

libdd-trace-utils/build.rs
ekump, line 1: Is this file supposed to be here? It doesn't look like it's used for anything.

libdd-trace-utils/Cargo.toml
```diff
@@ -45,6 +45,7 @@ libdd-tinybytes = { version = "1.1.1", path = "../libdd-tinybytes", features = [
     "serialization",
 ] }
 indexmap = "2.11"
+rustc-hash = "2"
```
ekump, line 48: Is this used anywhere?

libdd-trace-utils/src/change_buffer/buffer.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub fn read<T: Copy + FromBytes>(&self, index: &mut usize) -> Result<T> {
+        let size = std::mem::size_of::<T>();
+        // Safety: the allocation of `self.ptr` is guaranteed to be valid for read and writes at
+        // construction time. We do not materialize other references during the lifetime of `slice`.
+        let slice = unsafe { self.as_slice() };
+        let bytes = slice
+            .get(*index..*index + size)
+            .ok_or(ChangeBufferError::ReadOutOfBounds {
+                offset: *index,
+                len: self.len,
```
ekump: minor: could `len` for this this error be slightly confusing? Is it reasonable for someone reading this error to assume `len` means the length of bytes attempting to be read? `self.len` is the length of the entire buffer.

libdd-trace-utils/src/change_buffer/buffer.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /// Write a raw `u32` in the buffer.
+    pub fn write_u32(&mut self, offset: usize, value: u32) -> Result<()> {
+        let len = self.len;
+        // Safety: the allocation of `self.ptr` is guaranteed to be valid for read and writes at
+        // construction time. We do not materialize other references during the lifetime of `slice`.
+        let slice = unsafe { self.as_mut_slice() };
+        let target = slice
+            .get_mut(offset..offset + 4)
```
ekump: This is doing unchecked `usize` adds, which I think is normally fine. Is there any chance that we could get a corrupted or malicious offset value upstream that would trigger an overflow here?

libdd-trace-utils/src/change_buffer/span_header.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub trace_id_hi: u64, // offset 16
+    pub parent_id: u64,   // offset 24
+    pub start: i64,       // offset 32
+    pub duration: i64,    // offset 40
+    pub error: i32,       // offset 48
+    pub name_id: u32,     // offset 52
+    pub service_id: u32,  // offset 56
+    pub resource_id: u32, // offset 60
+    pub type_id: u32,     // offset 64
+    /// Index into ChangeBufferState.spans for meta/metrics overflow data.
```
ekump, line 29: I may just be confused, but is this doc comment accurate?

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -0,0 +1,111 @@
+//! Change buffer.
+//!
+//! A change buffer is a contiguous shared memory area between libdatadog and an external runtime.
+//! In order to amortize the cost of crossing the FFI when using native spans, the runtime write
+//! events in the change buffer instead many times and only flush it by batch, where the call to
+//! libdatadog happens. Libdatadog processes the change buffer and reconstruct the corresponding
+//! spans.
```
ekump: minor grammar tweak: 

```suggestion
//! In order to amortize the cost of crossing the FFI when using native spans, the runtime writes
//! events into the change buffer instead of calling libdatadog many times, and only flushes by
//! batch — that flush is where the call to libdatadog happens. Libdatadog then processes the change
//! buffer and reconstructs the corresponding spans.
```

libdd-trace-utils/src/change_buffer/buffer.rs
yannham: I would tend to think cloning/copying should be ok, since `ChangeBuffer` is effectively just a shared pointer (and a length). It's not needed on the libdatadog side (it compiles if I remove it), but not sure about the code on libdatadog-nodejs side though.

libdd-trace-utils/build.rs
yannham: Indeed, seems to be a leftover from older iteration :+1:

libdd-trace-utils/Cargo.toml
yannham: Nope, leftover from hashmap times. Removed.

libdd-trace-utils/src/change_buffer/buffer.rs
yannham: I would expect a out of bounds error to report the faulty index and the actual length. I don't think `len` is confusing _per se_, but I agree the combination here is, because `index` is not necessarily out of bounds (some value in `index..index+size` is). I've added a field for `value_len` and renamed the `len` to `buffer_len` to avoid confusion, plus documentation of those fields.

libdd-trace-utils/src/change_buffer/buffer.rs
yannham: Added `checked_add` for public methods, and made `read_unchecked` private.

libdd-trace-utils/src/change_buffer/span_header.rs
yannham: It is accurate, but confusing because `ChangeBufferState` is introduced in the next PR of the series (#2055)... Artifact of splitting one big chunk after-the-fact :shrug:

morrisonlevi:
Just driving by but noticed this:

> This PR just introduces the types and doesn't use them yet, hence the allow(unused).

But it's also a breaking change by virtue of the `!` in the title. That's an odd combination.

libdd-trace-utils/src/change_buffer/buffer.rs
paullegranddc: I'm not sure `Clone` is safe to implement since when using the `ChangeBuffer` from rust you can do something like

```
let mut b = ChangeBuffer::new();
let b_clone = b.clone();
b.as_mut_slice(); // mutable borrow of the memory while b_clone also owns the memory
```

libdd-trace-utils/src/change_buffer/buffer.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    /// Read a value without bounds checking.
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure `*index + size_of::<T>() <= self.len` and that `index + size_of<T>::() <
+    /// usize::MAX`.
+    #[inline(always)]
+    unsafe fn read_unchecked<T: Copy + FromBytes>(&self, index: &mut usize) -> T {
+        let size = std::mem::size_of::<T>();
```
paullegranddc: Would prefer if `FromBytes` has an associated const  `BYTE_SIZE: usize` and have the impl define the const as `size_of`

libdd-trace-utils/src/change_buffer/buffer.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /// Read a value without bounds checking.
+    ///
+    /// # Safety
+    ///
+    /// Caller must ensure `*index + size_of::<T>() <= self.len` and that `index + size_of<T>::() <
+    /// usize::MAX`.
+    #[inline(always)]
+    unsafe fn read_unchecked<T: Copy + FromBytes>(&self, index: &mut usize) -> T {
```
paullegranddc: is the Copy bound needed here?

libdd-trace-utils/src/change_buffer/buffer.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    /// # Safety
+    ///
+    /// Same safety conditions as [std::slice::from_raw_parts_mut].
+    unsafe fn as_mut_slice(&mut self) -> &mut [u8] {
+        unsafe { std::slice::from_raw_parts_mut(self.ptr, self.len) }
+    }
+
+    /// Read a value of type `T` starting at offset `index`.
+    pub fn read<T: Copy + FromBytes>(&self, index: &mut usize) -> Result<T> {
```
paullegranddc: I feel like this should be expressed in term of checks + call to `read_unchecked` since the logic is so close

yannham:
> But it's also a breaking change by virtue of the ! in the title. That's an odd combination.

Ah, this was a leftover (and not even at the right place). Removed.

libdd-trace-utils/src/change_buffer/buffer.rs
```diff
@@ -0,0 +1,198 @@
...
+    pub fn write_u32(&mut self, offset: usize, value: u32) -> Result<()> {
+        let len = self.len;
+        // Safety: the allocation of `self.ptr` is guaranteed to be valid for read and writes at
+        // construction time. We do not materialize other references during the lifetime of `slice`.
+        let slice = unsafe { self.as_mut_slice() };
+        let out_of_bounds_err = || ChangeBufferError::WriteOutOfBounds {
+            offset,
+            value_len: offset + 4,
+            buffer_len: len,
+        };
```
paullegranddc, line 74: I would think that rustc is smart enough to not materialize this if we don't go through the error paths

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+fn deferred_meta_insert(vec: &mut Vec<(u32, u32)>, key_id: u32, val_id: u32) {
+    for entry in vec.iter_mut() {
+        if entry.0 == key_id {
+            entry.1 = val_id;
+            return;
+        }
+    }
+    vec.push((key_id, val_id));
+}
```
paullegranddc, line 122: This is just equivalent to vec_insert::<u32, u32> no?

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+fn deferred_metric_insert(vec: &mut Vec<(u32, f64)>, key_id: u32, val: f64) {
```
paullegranddc, line 124: same here, vec_insert::<u32, f64>

libdd-trace-utils/src/change_buffer/span_header.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// Set when the span is allocated.
+    pub active: u32, // offset 68 (1 = in use, 0 = free)
+}
+
+/// Field offsets for JS DataView access.
+pub mod offsets {
+    pub const SPAN_ID: usize = 0;
```
paullegranddc:
```suggestion
    pub const SPAN_ID: usize = mem::offset_of!(SpanHeader, span_id);
```

paullegranddc (approved):

libdd-trace-utils/src/change_buffer/mod.rs
paullegranddc: Replaced by VecMap

libdd-trace-utils/src/change_buffer/buffer.rs
yannham: So `as_mut_slice` is unsafe, but I get your point, it's easy to misuse. If there's no usage of `Clone`, let's remove the bounds then.

yannham:
> Ah, this was a leftover (and not even at the right place). Removed.

No, in fact the CI complains. I suspect the addition of public items is considered a breaking change. It's not, but let's please the CI for now.

libdd-trace-utils/src/change_buffer/buffer.rs
yannham: `read_unchecked` has been removed in the end, it's not used anymore in following PRs.

PR #2047, opened 2026-05-27
fix(send_with_retry): follow max retries of the strategy

VianneyRuhlmann:
# What does this PR do?

Fix an off-by 1 error in `send_with_retry` in the number of retries. The `max_retries` field of the RetryStrategy was effectively used as `max_attempts`.

The default strategy has not been modified and will no do up to 5 retries so 6 attempts (previously 5 attempts). However the OTLP code indicates that the intent was to have 5 total attempts so the strategy has ben updated to keep the same total of attempts.

# Motivation

There's an off by 1 error in the send_with_retry count of retries.

# Additional Notes

# How to test the change?


libdd-data-pipeline/src/otlp/exporter.rs
```diff
@@ -11,8 +11,8 @@ use libdd_trace_utils::send_with_retry::{
     send_with_retry, RetryBackoffType, RetryStrategy, SendWithRetryError,
 };
 
-/// Max total attempts for OTLP export (1 initial + up to 4 retries on transient failures).
-const OTLP_MAX_ATTEMPTS: u32 = 5;
+/// Max retries for OTLP export.
+const OTLP_MAX_RETRIES: u32 = 4;
```
ekump, line 15: minor: should we do the same for the default elsewhere so we maintain the current behavior? I guess an argument against this would be we weren't honoring what we said we were doing before and this makes it right.

ekump (approved):

libdd-data-pipeline/src/otlp/exporter.rs
VianneyRuhlmann: > we weren't honoring what we said we were doing before and this makes it right.

That was my thought here the comment shows that the intent was to have exactly 5 attempts whereas the default strategy was stating 5 max_retries and not 5 max_attempts

hoolioh (approved):

PR #2048, opened 2026-05-27
perf(profiling)!: reduce profiler arena memory footprint

taegyunkim:
# What does this PR do?

Reduces the profiler arena memory floor while preserving larger-workload behavior by adding opt-in capped geometric growth to `ChainAllocator` and using it for profiler arenas.

This PR uses the `ProfilesDictionary` Criterion benchmark added in #2088 so the GitLab benchmark job can compare the change.

Changes:
- Adds capped geometric growth to `ChainAllocator`.
- Adds `ChainAllocator::new_capped_in(initial, max, allocator)` for callers that want a smaller initial chunk but a historical/max chunk size after growth.
- Keeps `ChainAllocator::new_in(...)` using fixed-size routine chunks. Geometric growth is opt-in through `new_capped_in(...)`.
- Lowers profiler dictionary arena initial chunks from 1 MiB to 64 KiB, capped at the historical 1 MiB chunk size.
- Lowers per-profile `StringTable` initial chunks from 4 MiB to 512 KiB, capped at the historical 4 MiB chunk size.
- Keeps `ParallelStringSet` / `ParallelSliceSet` at 16 shards and does not change shard selection.

# Motivation

Python profiler memory analysis showed that common profiles keep only tens to hundreds of KiB of dictionary/string-table content, but libdatadog reserved much larger arena chunks up front. This created a high per-process memory floor, especially across forked workers.

The smaller initial chunks reduce that floor. Geometric growth avoids keeping large/high-cardinality services on tiny chunks indefinitely, so they ramp back to the previous chunk sizes after a few growth events.

The 64 KiB dictionary initial chunk size is a compromise between lowering the fixed per-shard memory floor and avoiding very small chunks. With 16 string shards, it lowers the initial string-set arena floor from 16 MiB to 1 MiB, while geometric growth reaches the historical 1 MiB chunk size after four growths.

# Shard count

This PR keeps the existing 16-shard default and shard selection logic.

I originally explored reducing `ParallelStringSet` / `ParallelSliceSet` from 16 shards to 4, but dropped that from this PR. The extra memory saved by reducing shards after the arena-size change is relatively small (`12 * 64 KiB = 768 KiB` for the string set), while 16 shards preserve better concurrent insertion headroom.

Consumer concurrency summary:
- ddprof appears effectively single-writer for dictionary insertion: the worker thread interns while processing events, while export serializes an inactive profile buffer.
- dd-trace-py can have dictionary insertion concurrency today: one native stack sampler thread can intern off-GIL while one Python/Cython collector path is active under the GIL.
- Future free-threaded/no-GIL Python could increase producer concurrency.

So this PR focuses on the main memory win: smaller initial arenas with capped growth, without reducing shard count.

# Additional Notes

Expected growth patterns:
- Dictionary arenas: `64 KiB -> 128 KiB -> 256 KiB -> 512 KiB -> 1 MiB -> ...`
- Per-profile `StringTable`: `512 KiB -> 1 MiB -> 2 MiB -> 4 MiB -> ...`

Oversized individual allocations still allocate chunks large enough for the request, even if larger than the routine growth cap.

Approximate initial arena floor after this change:
- `ParallelStringSet`: `16 * 64 KiB = 1 MiB` instead of `16 * 1 MiB = 16 MiB`.
- `FunctionSet`: `4 * 64 KiB = 256 KiB` instead of `4 * 1 MiB = 4 MiB`.
- `MappingSet`: `2 * 64 KiB = 128 KiB` instead of `2 * 1 MiB = 2 MiB`.
- Per-profile `StringTable`: `512 KiB` instead of `4 MiB`.

# How to test the change?

Latest local validation:

```sh
cargo +nightly-2026-02-08 fmt --all -- --check
cargo test -p libdd-alloc chain::tests -- --nocapture
cargo test -p libdd-profiling profiles::collections::parallel::string_set::tests::test_hash_distribution -- --nocapture
cargo test -p libdd-profiling initial_dictionary_arena_floor_is_small -- --nocapture
cargo test -p libdd-profiling dictionary_grows_past_initial_arenas_and_preserves_handles -- --nocapture
```

Previously ran:

```sh
cargo check -p libdd-alloc -p libdd-profiling
cargo check -p libdd-profiling --benches
cargo +stable clippy -p libdd-alloc -p libdd-profiling --all-targets --all-features -- -D warnings
cargo nextest run -p libdd-alloc -p libdd-profiling
cargo test --doc -p libdd-alloc -p libdd-profiling
```

[PROF-14423](https://datadoghq.atlassian.net/browse/PROF-14423)

[PROF-14423]: https://datadoghq.atlassian.net/browse/PROF-14423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libdd-alloc/src/chain.rs
+++ b/libdd-alloc/src/chain.rs
@@ -17,9 +17,8 @@ use core::ptr::NonNull;
 /// [ChainAllocator] creates a new [LinearAllocator] when the current one
 /// doesn't have enough space for the requested allocation, and then links the
 /// new [LinearAllocator] to the previous one, creating a chain. This is where
-/// its name comes from. Each successful growth doubles the target chunk size
-/// up to a cap, so small arenas retain a low initial footprint while larger
-/// workloads quickly converge to larger chunks.
+/// its name comes from. When configured with a larger maximum chunk size,
+/// each successful growth doubles the target chunk size up to that cap.
 pub struct ChainAllocator<A: Allocator + Clone> {
     top: UnsafeCell<ChainNodePtr<A>>,
     /// The size hint for the next linear allocator chunk.
@@ -92,11 +91,6 @@ impl<A: Allocator + Clone> ChainAllocator<A> {
     /// is worth it. This is somewhat arbitrarily chosen at the moment.
     const MIN_NODE_SIZE: usize = 4 * Self::CHAIN_NODE_OVERHEAD;
 
-    /// Default cap for routine geometric growth. This preserves the historical
-    /// chunk size used by profiling dictionaries while allowing smaller initial
-    /// chunks to ramp up quickly.
-    const DEFAULT_MAX_NODE_SIZE: usize = 1024 * 1024;
-
     const fn normalize_node_size(size: usize) -> usize {
         if size < Self::MIN_NODE_SIZE {
             Self::MIN_NODE_SIZE
@@ -106,26 +100,22 @@ impl<A: Allocator + Clone> ChainAllocator<A> {
     }
 
     /// Creates a new [ChainAllocator]. The `chunk_size_hint` is used as the
-    /// initial size hint for chunks of the chain. Routine growth doubles the
-    /// chunk size up to at least `Self::DEFAULT_MAX_NODE_SIZE`. Note that the
-    /// [ChainAllocator] will use some bytes at the beginning of each chunk of
-    /// the chain. The number of bytes is [Self::CHAIN_NODE_OVERHEAD]. Keep
-    /// this in mind when sizing your hint if you are trying to be precise,
-    /// such as making sure a specific object fits.
+    /// size hint for chunks of the chain. Note that the [ChainAllocator] will
+    /// use some bytes at the beginning of each chunk of the chain. The number
+    /// of bytes is [Self::CHAIN_NODE_OVERHEAD]. Keep this in mind when sizing
+    /// your hint if you are trying to be precise, such as making sure a
+    /// specific object fits.
+    ///
+    /// To use geometric growth, use [Self::new_capped_in].
     pub const fn new_in(chunk_size_hint: usize, allocator: A) -> Self {
-        let initial_node_size = Self::normalize_node_size(chunk_size_hint);
-        let max_node_size = if initial_node_size < Self::DEFAULT_MAX_NODE_SIZE {
-            Self::DEFAULT_MAX_NODE_SIZE
-        } else {
-            initial_node_size
-        };
-        Self::new_capped_in(initial_node_size, max_node_size, allocator)
+        Self::new_capped_in(chunk_size_hint, chunk_size_hint, allocator)
     }
 
     /// Creates a new [ChainAllocator] whose routine growth starts at
     /// `chunk_size_hint` and doubles until reaching `max_chunk_size_hint`.
-    /// Requests larger than the cap are still honored by allocating an
-    /// oversized chunk for that request.
+    /// If the maximum is the same as the initial size, routine chunks keep a
+    /// fixed size. Requests larger than the cap are still honored by allocating
+    /// an oversized chunk for that request.
     pub const fn new_capped_in(
         chunk_size_hint: usize,
         max_chunk_size_hint: usize,
@@ -451,7 +441,7 @@ mod tests {
 
     #[test]
     fn test_geometric_growth() {
-        let allocator = ChainAllocator::new_in(4096, Global);
+        let allocator = ChainAllocator::new_capped_in(4096, 1024 * 1024, Global);
         let layout = Layout::new::<u8>();
 
         let _ = allocator.allocate(layout).unwrap();
@@ -473,6 +463,25 @@ mod tests {
         assert!(
             third_chunk >= second_chunk * 2,
             "third chunk should grow geometrically: second={second_chunk}, third={third_chunk}"
+        );
+    }
+
+    #[test]
+    fn test_new_in_uses_fixed_size_chunks() {
+        let allocator = ChainAllocator::new_in(4096, Global);
+        let layout = Layout::new::<u8>();
+
+        let _ = allocator.allocate(layout).unwrap();
+        let first_reserved = allocator.reserved_bytes();
+        fill_to_capacity(&allocator);
+
+        let _ = allocator.allocate(layout).unwrap();
+        let second_reserved = allocator.reserved_bytes();
+        let second_chunk = second_reserved - first_reserved;
+
+        assert!(
+            second_chunk < first_reserved * 2,
+            "new_in should not enable geometric growth: first={first_reserved}, second={second_chunk}"
         );
     }
 
@@ -553,7 +562,7 @@ mod tests {
 
         let bool_layout = Layout::new::<bool>();
 
-        const GROWTH_ITERATIONS: usize = 16;
+        const GROWTH_ITERATIONS: usize = 100;
 
         // test that it fills to capacity a few times.
         for _ in 0..GROWTH_ITERATIONS {
--- a/libdd-alloc/src/chain.rs
+++ b/libdd-alloc/src/chain.rs
@@ -21,7 +21,9 @@ use core::ptr::NonNull;
 /// each successful growth doubles the target chunk size up to that cap.
 pub struct ChainAllocator<A: Allocator + Clone> {
     top: UnsafeCell<ChainNodePtr<A>>,
-    /// The size hint for the next linear allocator chunk.
+    /// The size hint for the next linear allocator chunk. This is a [Cell]
+    /// because [Allocator::allocate] takes `&self`, but growth updates the
+    /// next routine chunk size. [ChainAllocator] is not [Sync].
     node_size: Cell<usize>,
     /// The maximum size hint used for routine geometric growth. Individual
     /// oversized allocations can still request larger chunks.
@@ -148,6 +150,9 @@ impl<A: Allocator + Clone> ChainAllocator<A> {
         if Layout::from_size_align(next, align).is_ok() {
             next
         } else {
+            // `current` came from a valid layout. If the next geometric target
+            // is invalid, keep the last valid routine size instead of storing
+            // an invalid future hint.
             current
         }
     }
--- a/libdd-alloc/src/chain.rs
+++ b/libdd-alloc/src/chain.rs
@@ -4,7 +4,7 @@
 use crate::LinearAllocator;
 use crate::{AllocError, Allocator};
 use core::alloc::Layout;
-use core::cell::UnsafeCell;
+use core::cell::{Cell, UnsafeCell};
 use core::mem::size_of;
 use core::ptr::NonNull;
 
@@ -17,11 +17,16 @@ use core::ptr::NonNull;
 /// [ChainAllocator] creates a new [LinearAllocator] when the current one
 /// doesn't have enough space for the requested allocation, and then links the
 /// new [LinearAllocator] to the previous one, creating a chain. This is where
-/// its name comes from.
+/// its name comes from. Each successful growth doubles the target chunk size
+/// up to a cap, so small arenas retain a low initial footprint while larger
+/// workloads quickly converge to larger chunks.
 pub struct ChainAllocator<A: Allocator + Clone> {
     top: UnsafeCell<ChainNodePtr<A>>,
-    /// The size hint for the linear allocator's chunk.
-    node_size: usize,
+    /// The size hint for the next linear allocator chunk.
+    node_size: Cell<usize>,
+    /// The maximum size hint used for routine geometric growth. Individual
+    /// oversized allocations can still request larger chunks.
+    max_node_size: usize,
     allocator: A,
 }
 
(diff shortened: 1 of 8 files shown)


ivoanjo:
Note that historically the tension here was between fragmentation and memory use -- that's why we set the higher defaults. (See for instance https://docs.google.com/document/d/1g_H7G9s_H9yoxlpyw_B0aoUyIVmo0ZQBzQkp5EUUyX8/edit?tab=t.0 )

This not to say that we can't or shouldn't adjust these numbers, it's more to add context to why larger numbers were chosen rather than starting with smallest possible and just letting it grow.

taegyunkim:
> Note that historically the tension here was between fragmentation and memory use -- that's why we set the higher defaults. (See for instance https://docs.google.com/document/d/1g_H7G9s_H9yoxlpyw_B0aoUyIVmo0ZQBzQkp5EUUyX8/edit?tab=t.0 )
> 
> This not to say that we can't or shouldn't adjust these numbers, it's more to add context to why larger numbers were chosen rather than starting with smallest possible and just letting it grow.

@ivoanjo Thanks for the context! That makes sense, and this is why this PR uses capped geometric growth. 

A couple of differences make this less risky than the story from your report:

- These profiler dictionary/per profile string-table arenas use `ChainAllocator<VirtualAllocator>`, so on Unix they allocate via `mmap`, not glibc `malloc`. 
- The chunks are arena-owned and long-lived. We're not creating `malloc`/`free` churn interleaved with runtime allocations.
- Larger workloads converge back to the historical chunk sizes. 

So this keeps the lower memory floor for small/common profiles, while avoiding the "smallest possible and just keep growing tiny chunks" behavior. 

I agree we should validate this with real workloads, especially Ruby if we're worried about fragmentation.

ivoanjo:
Ahh that's great, thanks for the extra context. In particular, I missed the detail where these come from `mmap` directly -- in that case I indeed expect the likelihood of fragmentation is way way lower (e.g. address space fragmentation could be possible but... I've not heard of it happening very commonly so hopefully the kernel/glibc do a good job there?).

Excited to see the improvements from this one :D

taegyunkim:
@ivoanjo the DoE results look very good for Python with this change

For all three archetypes, we see reduction in heap live size, heap live samples, allocated memory, allocations without change in cpu-time. 

[Enterprise](https://benchmarks.datadoghq.com/profiling/comparison?query=service%3A%22doe%2Fpython%22%20doe.hash%3A80404cfded1445e5a3888abbeec9ac0b%20runtime%3Aebpf&agg_m=%40prof_ebpf_cpu_cores&agg_m_source=base&agg_t=sum&compare_end_A=1780602228272&compare_end_B=1780603655342&compare_query_A=service%3A%22doe%2Fpython%22%20language%3Anative%20library_git_commit_sha%3A58ac3af%20archetype%3Aenterprise%20process.environment_variable.profiling%3Atrue&compare_query_B=service%3A%22doe%2Fpython%22%20language%3Anative%20library_git_commit_sha%3A30dd566%20archetype%3Aenterprise%20process.environment_variable.profiling%3Atrue&compare_start_A=1780598628272&compare_start_B=1780600055342&compareValuesMode=absolute&extra_search_fields=%7B%22filters_query%22%3A%22%22%2C%22sample_type%22%3A%22cpu-time%22%7D&fromUser=true&my_code=enabled&profile_type=heap-live-size&profiling-flame-graph__filter=show_from%28file%3A%22_native.cpython-313-x86_64-linux-gnu.so%22%29&refresh_mode=paused&viz=flame_graph&from_ts=1775422333876&to_ts=1775425933876&live=false)

Screenshot 2026-06-05 at 9 54 57 AM

[Latency](https://benchmarks.datadoghq.com/profiling/comparison?query=service%3A%22doe%2Fpython%22%20doe.hash%3A80404cfded1445e5a3888abbeec9ac0b%20runtime%3Aebpf&agg_m=%40prof_ebpf_cpu_cores&agg_m_source=base&agg_t=sum&compare_end_A=1780602228272&compare_end_B=1780603642346&compare_query_A=service%3A%22doe%2Fpython%22%20language%3Anative%20library_git_commit_sha%3A58ac3af%20archetype%3Alatency%20process.environment_variable.profiling%3Atrue&compare_query_B=service%3A%22doe%2Fpython%22%20language%3Anative%20library_git_commit_sha%3A30dd566%20archetype%3Alatency%20process.environment_variable.profiling%3Atrue&compare_start_A=1780598628272&compare_start_B=1780600042346&compareValuesMode=absolute&extra_search_fields=%7B%22filters_query%22%3A%22%22%2C%22sample_type%22%3A%22cpu-time%22%7D&fromUser=true&my_code=enabled&profile_type=heap-live-size&profiling-flame-graph__filter=show_from%28file%3A%22_native.cpython-313-x86_64-linux-gnu.so%22%29&refresh_mode=paused&viz=flame_graph&from_ts=1775422333876&to_ts=1775425933876&live=false)

Screenshot 2026-06-05 at 9 55 15 AM

[Throughput](https://benchmarks.datadoghq.com/profiling/comparison?query=service%3A%22doe%2Fpython%22%20doe.hash%3A80404cfded1445e5a3888abbeec9ac0b%20runtime%3Aebpf&agg_m=%40prof_ebpf_cpu_cores&agg_m_source=base&agg_t=sum&compare_end_A=1780602228272&compare_end_B=1780603658499&compare_query_A=service%3A%22doe%2Fpython%22%20language%3Anative%20library_git_commit_sha%3A58ac3af%20archetype%3Athroughput%20process.environment_variable.profiling%3Atrue&compare_query_B=service%3A%22doe%2Fpython%22%20language%3Anative%20library_git_commit_sha%3A30dd566%20archetype%3Athroughput%20process.environment_variable.profiling%3Atrue&compare_start_A=1780598628272&compare_start_B=1780600058499&compareValuesMode=absolute&extra_search_fields=%7B%22filters_query%22%3A%22%22%2C%22sample_type%22%3A%22cpu-time%22%7D&fromUser=true&my_code=enabled&profile_type=heap-live-size&profiling-flame-graph__filter=show_from%28file%3A%22_native.cpython-313-x86_64-linux-gnu.so%22%29&refresh_mode=paused&viz=flame_graph&from_ts=1775422333876&to_ts=1775425933876&live=false)

Screenshot 2026-06-05 at 9 54 48 AM

libdd-profiling/src/collections/string_table/mod.rs
```diff
@@ -83,10 +83,13 @@ impl StringTable {
...
-        // string tables in memory at a given time. Talk to .NET profiling
-        // engineers before making this any bigger.
-        const SIZE_HINT: usize = 4 * 1024 * 1024;
-        let bytes = ChainAllocator::new_in(SIZE_HINT, VirtualAllocator {});
+        // string tables in memory at a given time. Larger profiles grow
+        // geometrically up to the historical 4 MiB chunk size, while common
+        // profiles fit comfortably below this initial size. Talk to .NET
+        // profiling engineers before making this any bigger.
+        const SIZE_HINT: usize = 512 * 1024;
+        const MAX_SIZE_HINT: usize = 4 * 1024 * 1024;
```
morrisonlevi: I think the historical SIZE_HINT comment belongs  to MAX_SIZE_HINT now? Meaning we should split the comments here.

libdd-alloc/src/chain.rs
```diff
@@ -17,11 +17,16 @@ use core::ptr::NonNull;
 /// [ChainAllocator] creates a new [LinearAllocator] when the current one
 /// doesn't have enough space for the requested allocation, and then links the
 /// new [LinearAllocator] to the previous one, creating a chain. This is where
-/// its name comes from.
+/// its name comes from. Each successful growth doubles the target chunk size
```
danielsn: have we experimented with other factors? e.g. 1.5x would still grow geometrically, but not as fast.

libdd-alloc/src/chain.rs
```diff
@@ -87,38 +92,89 @@ impl<A: Allocator + Clone> ChainAllocator<A> {
...
+
+    /// Creates a new [ChainAllocator]. The `chunk_size_hint` is used as the
+    /// initial size hint for chunks of the chain. Routine growth doubles the
+    /// chunk size up to at least `Self::DEFAULT_MAX_NODE_SIZE`. Note that the
     /// [ChainAllocator] will use some bytes at the beginning of each chunk of
     /// the chain. The number of bytes is [Self::CHAIN_NODE_OVERHEAD]. Keep
     /// this in mind when sizing your hint if you are trying to be precise,
     /// such as making sure a specific object fits.
     pub const fn new_in(chunk_size_hint: usize, allocator: A) -> Self {
+        let initial_node_size = Self::normalize_node_size(chunk_size_hint);
```
danielsn: why does one of these get a function and the other is inline?

libdd-profiling/src/profiles/datatypes/profiles_dictionary.rs
```diff
@@ -163,6 +191,94 @@ mod tests {
...
+    #[test]
+    fn initial_dictionary_arena_floor_is_small() {
+        const SMALL_ARENA_HINT: usize = 64 * 1024;
+        const STRING_SHARDS: usize = crate::profiles::collections::N_SHARDS;
+
+        let dict = ProfilesDictionary::try_new().unwrap();
+
+        assert!(string_arena_reserved_bytes(&dict) <= STRING_SHARDS * SMALL_ARENA_HINT);
+        assert!(function_arena_reserved_bytes(&dict) <= 4 * SMALL_ARENA_HINT);
+        assert!(mapping_arena_reserved_bytes(&dict) <= 2 * SMALL_ARENA_HINT);
```
danielsn: where do these constants come from?

libdd-profiling/src/profiles/collections/set.rs
```diff
@@ -61,7 +61,10 @@ pub struct Set<T: Hash + Eq + 'static> {
 }
 
 impl<T: Eq + Hash + 'static> Set<T> {
-    pub const SIZE_HINT: usize = 1024 * 1024;
+    // Keep the per-shard arena small; larger dictionaries grow
+    // geometrically up to the historical 1 MiB chunk size.
+    pub const SIZE_HINT: usize = 64 * 1024;
```
danielsn: why this constant?

danielsn: should this be named `INITIAL_SIZE_HINT`

libdd-profiling/src/collections/string_table/mod.rs
```diff
@@ -83,10 +83,13 @@ impl StringTable {
...
         // we're just a piece inside it. Additionally, there may be 2 or more
-        // string tables in memory at a given time. Talk to .NET profiling
-        // engineers before making this any bigger.
-        const SIZE_HINT: usize = 4 * 1024 * 1024;
-        let bytes = ChainAllocator::new_in(SIZE_HINT, VirtualAllocator {});
+        // string tables in memory at a given time. Larger profiles grow
+        // geometrically up to the historical 4 MiB chunk size, while common
+        // profiles fit comfortably below this initial size. Talk to .NET
+        // profiling engineers before making this any bigger.
+        const SIZE_HINT: usize = 512 * 1024;
```
danielsn: for the other case, we went from 64K-1M, here we go from 512K-4M.  Why

libdd-profiling/src/collections/string_table/mod.rs
taegyunkim: Done

libdd-alloc/src/chain.rs
```diff
@@ -401,8 +553,10 @@ mod tests {
 
         let bool_layout = Layout::new::<bool>();
 
+        const GROWTH_ITERATIONS: usize = 16;
```
danielsn: why the reduction?

libdd-alloc/src/chain.rs
```diff
@@ -87,38 +92,89 @@ impl<A: Allocator + Clone> ChainAllocator<A> {
...
+
+        let Some(doubled) = current.checked_mul(2) else {
+            return max;
+        };
+        let next = if doubled > max { max } else { doubled };
+
+        if Layout::from_size_align(next, align).is_ok() {
+            next
+        } else {
+            current
```
danielsn, line 156: Why is this the right fall-back when from_size_align fails?

libdd-alloc/src/chain.rs
```diff
@@ -87,38 +92,89 @@ impl<A: Allocator + Clone> ChainAllocator<A> {
...
+        };
         Self {
             top: UnsafeCell::new(ChainNodePtr::new()),
-            // max is not a const fn, do it manually.
-            node_size: if chunk_size_hint < Self::MIN_NODE_SIZE {
-                Self::MIN_NODE_SIZE
-            } else {
-                chunk_size_hint
-            },
+            node_size: Cell::new(initial_node_size),
```
danielsn, line 134: document why this need to be a cell?

libdd-alloc/src/chain.rs
taegyunkim: No, I did not benchmark other factors. Chose 2x because the goal here is to lower the memory footprint for applications smaller dictionaries, while converging back to prior chunk sizes quickly for larger workloads.

libdd-alloc/src/chain.rs
taegyunkim: This was removed in the latest refactor. The generic 1MiB default cap was removed from `ChainAllocator`, and `new_in` now delegates to `new_capped_in(chunk_size_hint, chunk_size_hint, allocator)` so it keeps fixed-size routine chunks.

libdd-profiling/src/profiles/datatypes/profiles_dictionary.rs
taegyunkim: Updated the code to query actual shard counts instead of hard-coding the numbers here.

libdd-alloc/src/chain.rs
taegyunkim: This is reverted now.

libdd-alloc/src/chain.rs
taegyunkim: Done!

libdd-alloc/src/chain.rs
taegyunkim: `current` comes from the layout we just successfully constructed and allocated, so it is known to be valid for this alignment. If the next target is not valid, keeping `current` preserves the last valid routine chunk size instead of storing an invalid future hint that would make subsequent growth fail.

libdd-profiling/src/collections/string_table/mod.rs
taegyunkim: 512K is not a number derived from a benchmark. 

The difference is that dictionary arenas are per-shard, while `StringTable` is a single arena per profile. 

With 16 string shards, 64KiB per shard is already a 1MiB initial size. Applying the same 64KiB to `StringTable` would make the whole table start at only 64KiB, which felt too small. 

512KiB still reduces the initial reservation substantially from the historically 4MiB, while being less aggressive than the per-shard dictionary setting.

PR #2049, opened 2026-05-28
feat(trace-utils)!: add dedup convenience to VecMap

yannham:
# What does this PR do?

Follow-up of #2022. This split off some work required in #2043 for deduplication before msgpack encoding as a separate PR.

# Motivation

See #2022 for overall motivation. To integrate `VecMap`, it turns out we can't currently rely on the agent to properly handle maps with deuplicate entries (the "last win" semantics is the current implementation, but not guaranteed). We need to perform deduplication before msgpack encoding. This PR backports the required machinery into `VecMap` as a separate PR.

# Additional Notes

N/A

# How to test the change?

Tests included.

### DIFF
--- a/libdd-trace-utils/src/span/vec_map.rs
+++ b/libdd-trace-utils/src/span/vec_map.rs
@@ -555,7 +555,8 @@ mod tests {
         assert!(matches!(d, DedupedVecMap::Borrowed(_)));
         assert_eq!(d.len(), 2);
 
-        let items: Vec<_> = d.iter().collect();
+        let mut items: Vec<_> = d.iter().collect();
+        items.sort_by_key(|(k, _)| **k);
         assert_eq!(items, vec![(&"a", &1), (&"b", &2)]);
     }
 
--- a/libdd-trace-utils/src/span/vec_map.rs
+++ b/libdd-trace-utils/src/span/vec_map.rs
@@ -194,14 +194,7 @@ impl<K: Eq + Hash, V> VecMap<K, V> {
         if self.deduped {
             DedupedVecMap::Borrowed(self)
         } else {
-            DedupedVecMap::Owned(
-                self.data
-                    .iter()
-                    .map(|(k, v)| (k, v))
-                    .collect::<HashMap<_, _>>()
-                    .into_iter()
-                    .collect(),
-            )
+            DedupedVecMap::Owned(self.data.iter().map(|(k, v)| (k, v)).collect())
         }
     }
 
@@ -319,7 +312,7 @@ impl<K: Serialize + Eq + Hash, V: Serialize> Serialize for VecMap<K, V> {
 
 pub enum DedupedVecMap<'a, K, V> {
     Borrowed(&'a VecMap<K, V>),
-    Owned(Vec<(&'a K, &'a V)>),
+    Owned(HashMap<&'a K, &'a V>),
 }
 
 impl<'a, K, V> DedupedVecMap<'a, K, V> {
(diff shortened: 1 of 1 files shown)


libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -168,6 +185,51 @@ impl<K, V> VecMap<K, V> {
...
+    pub fn as_deduped_map(&self) -> DedupedVecMap<'_, K, V> {
+        if self.deduped {
+            DedupedVecMap::Borrowed(self)
+        } else {
+            static WARNED: AtomicBool = AtomicBool::new(false);
+            if !WARNED.swap(true, Ordering::Relaxed) {
+                tracing::warn!(
+                    "VecMap not deduped before encoding. Performing defensive on-the-fly dedup"
+                );
+            }
```
paullegranddc: This seems like it is a copy of the log in defensive_dedup, and that calling `defensive_dedup` on a non deduplicated vec will lead to double logging

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -168,6 +185,51 @@ impl<K, V> VecMap<K, V> {
...
+    }
+
+    /// This is a convenience wrapper around [Self::as_deduped_map] used in the msgpack encoder,
+    /// where we expect the map to be deduped, but call `as_deduped_map` as a defensive measure. If
+    /// the latter had to deduplicate and allocate a new vec, we log a warning (at most once).
+    #[allow(unused)]
+    pub(crate) fn defensive_dedup(&self) -> DedupedVecMap<'_, K, V> {
+        if !self.is_deduped() {
+            static WARNED: AtomicBool = AtomicBool::new(false);
+            if !WARNED.swap(true, Ordering::Relaxed) {
```
paullegranddc, line 208: Since the condition is unlikely, wouldn't it be better to do a relaxed atomic load, and a conditional store? (of course, we might get duplicata if used concurrently but I don't think it's that likely or we will get that many if ti happens)

libdd-trace-utils/src/span/vec_map.rs
yannham: Oops, rebase artifact. Thanks for the catch.

libdd-trace-utils/src/span/vec_map.rs
yannham: Well, this is the same discussion we had for the `dedup` flag, and it seems the wisdom is it's actually faster to unconditionally store (store buffer, avoided conditional even if predicted, yadayada). I think the case where it's not true is when the variable is contended (storing needs an exclusive cache line, which might invalidate the cache of other cores, while loading doesn't). But if we don't expect this to actually be accessed concurrently, the situation should be rather similar to `deduped` (we just have to use atomics to please the compiler)...

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -168,6 +185,44 @@ impl<K, V> VecMap<K, V> {
...
+        if self.deduped {
+            DedupedVecMap::Borrowed(self)
+        } else {
+            DedupedVecMap::Owned(
+                self.data
+                    .iter()
+                    .map(|(k, v)| (k, v))
+                    .collect::<HashMap<_, _>>()
+                    .into_iter()
+                    .collect(),
```
paullegranddc: Why the 2nd collect to VecMap here?
Wouldn't it be the exact same, with one less copy for the Owned  variant to be`DedupedVecMap::Owned(HashMap<K, V>)`

yannham: Good point, we could keep the hashmap here indeed.

bengl (approved):

PR #2050, opened 2026-05-28
feat(shared-runtime)!: use weak waker in trigger [APMSP-3371]

VianneyRuhlmann:
# What does this PR do?

Wrap the trigger future waker to use a weak reference on the runtime. This allows the runtime to be dropped and properly shutdown when calling `before_fork`

# Motivation

When reading from a channel in the trigger method, the waker being held by the channel can prevent the tokio driver from being dropped when calling `before_fork` which can cause crashes in macos as the kqueue handle becomes invalid in the child process.

# Additional Notes

# How to test the change?


libdd-shared-runtime/src/weak_waker.rs
```diff
@@ -0,0 +1,181 @@
...
+        // the original waker (e.g. the async runtime scheduler) can be released.
+        self.inner.waker.take();
+    }
+}
+
+/// Wrap a future so that the waker passed to it is held only weakly.
+///
+/// See the [module-level documentation](self) for details.
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub fn wrap<F: Future>(fut: F) -> WeakWakerFuture<F> {
```
yannham: Nit: should this be `WeakWakerFuture::new`, or at least `WeakWakerFuture::wrap` instead?

libdd-shared-runtime/src/weak_waker.rs
```diff
@@ -0,0 +1,181 @@
...
+    WeakWakerFuture {
+        fut,
+        weak_waker: None,
+    }
+}
+
+pub struct WeakWakerFuture<F: Future> {
+    fut: F,
+    weak_waker: Option<WeakWaker>,
```
yannham: Naive question: isn't there a double `Option` wrapping here? If `AtomicWaker` is already morally an `Option<Waker>`, could we just use `WeakWaker` here and put "none" in the atomic waker?

yannham (approved):
This solution adds a bit of code but looks more robust than the draining channel/`on_pause` business, is still reasonable and 👍 for the very clear documentation of the `weak_waker` module

libdd-shared-runtime/src/weak_waker.rs
paullegranddc: This is an optimisation so that we never allocate a WeakWaker of the Future is never polled. (A race bewteen two futures with one already ready for instance).

The Option doesn't cost anything since `WeakWaker` is a wrapper over Arc

paullegranddc (approved):

libdd-shared-runtime/src/weak_waker.rs
VianneyRuhlmann: You're right I can remove it, it makes the code way more readable. However it means I have to allocate the WeakWaker `Arc` and the `AtomicWaker` in `new` rather than on the first call to poll, but I think it's worth it.

yannham: Ah, there's an `Arc` in between indeed. Not sure what's the optimal trade-off, I guess it depends if you expect the "no allocation" case to happen often. I personally don't have a strong opinion

PR #2051, opened 2026-05-28
feat(data-pipeline)!: add fork safety hooks and cancellation token for trace exporter FFI

lloeki:
# What does this PR do?

Add fork safety and cooperative cancellation support to the trace exporter FFI:

1. **Fork hooks**: Expose `ddog_trace_exporter_before_fork`, `_after_fork_in_parent`,
   and `_after_fork_in_child` that delegate to the underlying `SharedRuntime`. These
   allow C callers to coordinate the tokio runtime lifecycle around process forks.

2. **Cancellation token**: Introduce `ddog_trace_exporter_cancel_token_new`, `_cancel`,
   and `_drop` for managing cancellation tokens. Thread the token into
   `ddog_trace_exporter_send_trace_chunks` via `tokio::select!` so callers can
   cooperatively abort in-flight HTTP requests.

# Motivation

FUP to #1952.

Ruby application servers (Puma, Unicorn, Passenger) fork worker processes. The tokio
runtime does not survive `fork()` — its threads are not carried over. Without the fork
hooks, the trace exporter is dead in child processes.

The existing `RUBY_UBF_IO` unblock function on the dd-trace-rb side sends a signal
that cannot actually cancel an in-flight Rust HTTP request. The cancellation token
enables cooperative abort: when Ruby interrupts a thread (shutdown, `Thread#kill`),
the UBF cancels the token, which causes `tokio::select!` to abort the send and return
promptly.

# Additional Notes

- Added a `pub fn shared_runtime()` accessor to `TraceExporter` since the field was private.
- The cancellation token uses `Handle<CancellationToken>` from `libdd-common-ffi`, same
  pattern as profiling-ffi, with a `ddog_TraceExporterCancelToken` cbindgen rename to
  avoid symbol conflicts.
- The `cancel` parameter on `send_trace_chunks` is nullable; passing `NULL` preserves
  existing behavior.

# How to test the change?

```
cargo test -p libdd-data-pipeline-ffi --lib
```

46 tests pass (40 existing + 6 new for fork hooks and cancellation token).


libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -290,6 +293,53 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_push_span(
...
+#[no_mangle]
+pub extern "C" fn ddog_trace_exporter_cancel_token_new() -> Handle<TokioCancellationToken> {
+    Handle::from(TokioCancellationToken::new())
+}
+
+/// Cancel a cancellation token.
+///
+/// All clones of the same token observe the cancellation. If the token is
+/// currently passed to [`ddog_trace_exporter_send_trace_chunks`], the
+/// in-flight HTTP request will be aborted cooperatively.
```
marcotc: What does `cooperatively` mean here?

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -290,6 +293,53 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_push_span(
...
+/// [`ddog_trace_exporter_cancel_token_drop`].
+#[no_mangle]
+pub extern "C" fn ddog_trace_exporter_cancel_token_new() -> Handle<TokioCancellationToken> {
+    Handle::from(TokioCancellationToken::new())
+}
+
+/// Cancel a cancellation token.
+///
+/// All clones of the same token observe the cancellation. If the token is
+/// currently passed to [`ddog_trace_exporter_send_trace_chunks`], the
```
marcotc:
```suggestion
/// currently help by [`ddog_trace_exporter_send_trace_chunks`], the
```

Wouldn't "currently held" be more accurate than "currently passed" here, since "ddog_trace_exporter_send_trace_chunks" might have finished its work already, thus not holding the token anymore?

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -290,6 +293,53 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_push_span(
...
+/// The returned handle must be freed with
+/// [`ddog_trace_exporter_cancel_token_drop`].
+#[no_mangle]
+pub extern "C" fn ddog_trace_exporter_cancel_token_new() -> Handle<TokioCancellationToken> {
+    Handle::from(TokioCancellationToken::new())
+}
+
+/// Cancel a cancellation token.
+///
+/// All clones of the same token observe the cancellation. If the token is
```
marcotc: What if the token was not passed to `ddog_trace_exporter_send_trace_chunks`, but will soon, is a cancellation: scheduled to happen; `ddog_trace_exporter_send_trace_chunks` won't start because it's already cancelled; nothing?
It's good to document what happens when the token is not held yet.
And also, if the token is no longer held (aka `ddog_trace_exporter_send_trace_chunks` is done), which I assume means nothing happens on `ddog_trace_exporter_cancel_token_cancel`.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -290,6 +293,53 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_push_span(
...
+/// currently passed to [`ddog_trace_exporter_send_trace_chunks`], the
+/// in-flight HTTP request will be aborted cooperatively.
+///
+/// # Safety
+///
+/// * `token` must point to a valid [`Handle`] returned by
+///   [`ddog_trace_exporter_cancel_token_new`].
```
marcotc:
```suggestion
/// * `token` is the return of [`ddog_trace_exporter_cancel_token_new`].
```

I don't think we need to say that we should give this function a "valid" input.

marcotc: The same applies to `ddog_trace_exporter_cancel_token_drop`.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -290,6 +293,53 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_push_span(
...
+    mut token: *mut Handle<TokioCancellationToken>,
+) {
+    if let Ok(inner) = token.to_inner_mut() {
+        inner.cancel();
+    }
+}
+
+/// Free a cancellation token handle.
+///
+/// After this call the pointer is invalid and must not be reused.
```
marcotc: Not sure if this is useful additional information "After this call the pointer is invalid and must not be reused.".

Freed resources cannot be used and are likely invalid.

If we want to provide additional context here, the only thing I can think of is: what happens if you free it while `ddog_trace_exporter_send_trace_chunks` still owns it. But I feel like that's kind of a natural contract in Rust/native: freeing things in use == bad.
Not sure if we need an extra comment here.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -310,25 +366,144 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_push_span(
 /// * `chunks` is consumed and must not be used after this call.
 /// * If `response_out` is non-null it must point to valid writable memory for a
 ///   `Box<ExporterResponse>`.
+/// * If `cancel` is non-null it must point to a valid cancellation token handle.
```
marcotc: Similar comment to those in `ddog_trace_exporter_cancel_token_cancel`: is it even possible to pass an invalid cancellation token since the argument is typed `*mut Handle<TokioCancellationToken>`?

libdd-data-pipeline-ffi/src/tracer.rs
marcotc: There's more of these `Safety` comment blocks in down this file. All the ones in a similar "type-checked argument must be type-valid" is likely not needed.

libdd-data-pipeline-ffi/Cargo.toml
```diff
@@ -36,4 +36,6 @@ libdd-shared-runtime = { version = "1.0.0", path = "../libdd-shared-runtime" }
 libdd-common-ffi = { path = "../libdd-common-ffi", default-features = false }
 libdd-tinybytes = { path = "../libdd-tinybytes" }
 libdd-trace-utils = { path = "../libdd-trace-utils" }
+tokio = { version = "1.23", features = ["macros"] }
+tokio-util = "0.7.11"
```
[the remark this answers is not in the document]
marcotc: edit: I think I was wrong, looks like a legit failure: https://github.com/DataDog/libdatadog/actions/runs/26578731968/job/78305509370?pr=2051

original comment:
This feels like a case of "AI doesn't know how "libdd-data-pipeline-ffi" is used.
This might information be worth writing down (maybe in https://github.com/DataDog/libdatadog/blob/22e92370e5d66c0efd6ac5fe949bd8b6514fd902/libdd-data-pipeline-ffi/README.md, or other place if this is common across multiple `libdatadog/*` crates).

marcotc (approved):
Despite my other comments, the implementation looks good 👍

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -310,25 +366,144 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_push_span(
...
 #[no_mangle]
 pub unsafe extern "C" fn ddog_trace_exporter_send_trace_chunks(
     exporter: Option<&TraceExporter>,
     chunks: Box<TracerTraceChunks>,
     response_out: Option<NonNull<Box<ExporterResponse>>>,
+    mut cancel: *mut Handle<TokioCancellationToken>,
```
VianneyRuhlmann: I think it's better to not use the Handle and use an `Option<&TokioCancellationToken>` instead. This way the null pointer is treated as None without any conversion.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -310,25 +366,144 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_push_span(
...
+                    Err(io_err) => Err(io_err.into()),
                 }
-                None
+            } else {
+                exporter.send_trace_chunks(chunks.0)
+            };
+
+            match result {
+                Ok(resp) => {
+                    if let Some(out) = response_out {
```
VianneyRuhlmann: The cancel token is a nice add-on and should be added to the actual trace exporter, we try to keep logic in the *-ffi crates to a minimum.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -209,6 +209,11 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
         TraceExporterBuilder::default()
     }
 
+    /// Returns a reference to the underlying [`SharedRuntime`].
+    pub fn shared_runtime(&self) -> &SharedRuntime {
+        &self.shared_runtime
+    }
+
```
VianneyRuhlmann: The SharedRuntime reference in the TraceExporter should remain private if you need to interact with the SharedRuntime (e.g. for fork hooks) you should create a SharedRuntime with `libdd-shared-runtime-ffi` and pass it to the TraceExporter builder

libdd-data-pipeline-ffi/src/tracer.rs
VianneyRuhlmann: Also the `send_trace_chunks_async` has a bug and shouldn't be used currently (it should have been marked as experimental). I can get it fixed, but the simplest way to bypass it is to add support for the cancel token to `send_trace_chunks`

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -310,25 +366,144 @@ pub unsafe extern "C" fn ddog_tracer_trace_chunks_push_span(
...
+                    tokio::select! {
+                        res = exporter.send_trace_chunks_async(chunks.0) => res,
+                        _ = ct.cancelled() => Err(
+                            std::io::Error::new(
+                                std::io::ErrorKind::Interrupted,
+                                "send cancelled via cancellation token",
+                            ).into()
+                        ),
+                    }
+                });
```
VianneyRuhlmann: It should be called out in the function's doc that using the cancel token may result in data loss

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: Should be addressed now.

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: Revamped this completely.

libdd-data-pipeline-ffi/src/tracer.rs
lloeki: I think that's totally expected, and data loss / resubmit after fork is on the caller (and protocol) to handle (or not handle).

IIRC the current code accepts the loss.

libdd-data-pipeline/src/trace_exporter/mod.rs
lloeki: Revamped, creating a SharedRuntime.

This needs a bit of change on the feature side, see: https://github.com/DataDog/libdatadog/pull/2051/commits/d174341cb9a05de2e3277d28656bad5dc3a6e344

libdd-profiling-ffi/Cargo.toml
```diff
@@ -33,6 +33,9 @@ datadog-library-config-ffi = ["dep:libdd-library-config-ffi"]
 ddcommon-ffi = ["dep:libdd-common-ffi"]
 ddsketch-ffi = ["dep:libdd-ddsketch-ffi"]
 datadog-ffe-ffi = ["dep:datadog-ffe-ffi"]
+# Ships the generic shared-runtime fork-lifecycle FFI inside this single aggregate
+# library so there is exactly one copy of the `libdd-shared-runtime` crate at runtime.
```
VianneyRuhlmann:
```suggestion
```
I don't think we need to justify to ship it in the main artifact

VianneyRuhlmann (approved):
LGTM

libdd-profiling-ffi/Cargo.toml
lloeki: Fair enough!

PR #2055, opened 2026-05-28
feat(trace-utils)!: change buffer implementation

yannham:
# What does this PR do?

This PR backports the change buffer implementation required for dd-trace-js adoption of native spans. Follow-up of #2046.

The change buffer is a contiguous region of memory where span operations are encoded from the JS runtime, similar to a byte code. The implementation decode all those span events and reconstruct the corresponding span objects.

# Motivation

Required to land native spans dd-trace-js.

# How to test the change?

tests included.

### DIFF
--- a/libdd-trace-utils/src/change_buffer/buffer.rs
+++ b/libdd-trace-utils/src/change_buffer/buffer.rs
@@ -80,15 +80,17 @@ impl ChangeBuffer {
         Ok(())
     }
 
-    pub(crate) fn read_arg<T: Clone>(
+    /// Read a string index from the buffer, fetch the string from the string table, clone it and
+    /// return it.
+    pub(crate) fn read_string<T: Clone>(
         &self,
         string_table: &super::StringTable<T>,
         index: &mut usize,
     ) -> Result<T> {
-        let num: u32 = self.read(index)?;
+        let idx: u32 = self.read(index)?;
         string_table
-            .get(num)
-            .ok_or(ChangeBufferError::StringNotFound(num))
+            .get(idx)
+            .ok_or(ChangeBufferError::StringNotFound(idx))
     }
 
     /// Clear the op count, which is stored in the first 4 bytes of the buffer. This effectively
--- a/libdd-trace-utils/src/change_buffer/buffer.rs
+++ b/libdd-trace-utils/src/change_buffer/buffer.rs
@@ -3,11 +3,12 @@
 
 use crate::change_buffer::utils::*;
 use crate::change_buffer::{ChangeBufferError, Result};
+use std::ptr::NonNull;
 
-/// A handle to a change buffer shared with another runtime. The memory is shared, meaning that
-/// cloning is cheap (copying the pointer and length).
+/// A handle to a fixed-size change buffer shared with another runtime. The memory is shared and
+/// owned/managed by the external runtime. The size of the buffer must not change once instantiated.
 pub struct ChangeBuffer {
-    ptr: *mut u8,
+    ptr: NonNull<u8>,
     len: usize,
 }
 
@@ -19,25 +20,22 @@ impl ChangeBuffer {
     /// The underlying raw memory must not be freed until after this struct's
     /// lifetime. Having the calling code manage the memory makes it simpler to
     /// integrate with managed runtimes.
-    pub unsafe fn from_raw_parts(ptr: *const u8, len: usize) -> Self {
-        Self {
-            ptr: ptr as *mut u8,
-            len,
-        }
+    pub unsafe fn from_raw_parts(ptr: NonNull<u8>, len: usize) -> Self {
+        Self { ptr, len }
     }
 
     /// # Safety
     ///
     /// Same safety conditions as [std::slice::from_raw_parts].
     unsafe fn as_slice(&self) -> &[u8] {
-        unsafe { std::slice::from_raw_parts(self.ptr, self.len) }
+        unsafe { std::slice::from_raw_parts(self.ptr.as_ptr(), self.len) }
     }
 
     /// # Safety
     ///
     /// Same safety conditions as [std::slice::from_raw_parts_mut].
     unsafe fn as_mut_slice(&mut self) -> &mut [u8] {
-        unsafe { std::slice::from_raw_parts_mut(self.ptr, self.len) }
+        unsafe { std::slice::from_raw_parts_mut(self.ptr.as_ptr(), self.len) }
     }
 
     /// Read a value of type `T` starting at offset `index`.
@@ -82,6 +80,17 @@ impl ChangeBuffer {
         Ok(())
     }
 
+    pub(crate) fn read_arg<T: Clone>(
+        &self,
+        string_table: &super::StringTable<T>,
+        index: &mut usize,
+    ) -> Result<T> {
+        let num: u32 = self.read(index)?;
+        string_table
+            .get(num)
+            .ok_or(ChangeBufferError::StringNotFound(num))
+    }
+
     /// Clear the op count, which is stored in the first 4 bytes of the buffer. This effectively
     /// reset the buffer (semantically), but without actually zeroing the rest.
     pub fn clear_count(&mut self) -> Result<()> {
@@ -92,13 +101,14 @@ impl ChangeBuffer {
 #[cfg(test)]
 mod tests {
     use super::ChangeBuffer;
+    use super::NonNull;
     use crate::change_buffer::Result;
 
     #[test]
     fn buffer_creation_and_slices() {
         let mut buf = unsafe {
-            let buf: [u8; 256] = std::mem::zeroed();
-            ChangeBuffer::from_raw_parts(buf.as_ptr(), 256)
+            let mut buf: [u8; 256] = std::mem::zeroed();
+            ChangeBuffer::from_raw_parts(NonNull::new(buf.as_mut_ptr()).unwrap(), 256)
         };
         {
             // Safety: slice is the only reference to the buffer in its scope.
@@ -117,7 +127,9 @@ mod tests {
         let example =
             b"This is an example string, long enough to get 16 bytes out of it, without issue.";
         let mut ex_buf = example.to_vec();
-        let mut buf = unsafe { ChangeBuffer::from_raw_parts(ex_buf.as_mut_ptr(), ex_buf.len()) };
+        let mut buf = unsafe {
+            ChangeBuffer::from_raw_parts(NonNull::new(ex_buf.as_mut_ptr()).unwrap(), ex_buf.len())
+        };
         let mut index = 8;
         assert_eq!(8101238474429984353, buf.read::<u64>(&mut index)?);
         assert_eq!(7956016061199967596, buf.read::<u64>(&mut index)?);
@@ -132,7 +144,9 @@ mod tests {
     #[test]
     fn clear_count() -> Result<()> {
         let mut buffer = vec![0xFFu8; 64];
-        let mut buf = unsafe { ChangeBuffer::from_raw_parts(buffer.as_mut_ptr(), buffer.len()) };
+        let mut buf = unsafe {
+            ChangeBuffer::from_raw_parts(NonNull::new(buffer.as_mut_ptr()).unwrap(), buffer.len())
+        };
         buf.clear_count()?;
         let mut index = 0;
         assert_eq!(0, buf.read::<u32>(&mut index)?);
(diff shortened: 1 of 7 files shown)


yannham:
I'm converting back to draft in order to take a closer look at unsafe code/unbound reads and see if we can avoid them.

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +89,793 @@ pub mod span_header;
...
+pub struct ChangeBufferState<T: TraceData> {
+    change_buffer: ChangeBuffer,
+    spans: Vec<Option<Span<T>>>,
+    traces: SmallTraceMap<T::Text>,
+    /// Interned string table (O(1) lookup vs HashMap).
+    ///
+    /// Note: currently the string table never shrinks (it is never compacted). When entries are
+    /// evicted (freeing the backing strings), a small amount of memory is leaked (to hold the
+    /// `None` value).
+    string_table: Vec<Option<T::Text>>,
```
paullegranddc: Nit: But could be interesting to replace these Vec<Option<T>> with a `Vec<MaybeUninit<T>>` + bitmap to save 8 bytes per item

libdd-trace-utils/src/change_buffer/buffer.rs
danielsn: If the calling code is supposed to own the memory, good to document that here

danielsn, line 10: This is a slice, why does it need its own struct?

danielsn, line 63: Should this be `unsafe`?

danielsn, line 12: can this be resized, or is len expected to be fixed once set?

danielsn: Can this be null?  If so, if `len` required to be `0`?

libdd-trace-utils/src/change_buffer/mod.rs
danielsn, line 1: I haven't looked at this file in detail, but 880 lines of new code for this seem like a lot.

libdd-trace-utils/src/change_buffer/buffer.rs
yannham: That's a good point. I think we still want to at least have a wrapper, to implement our own methods, but could probably just wrap a slice, yeah.

yannham: I don't think so. There's no particular pre-conditions to enforce from the caller. Since raw pointers are `!Send` and `!Sync`, the change buffer can't be shared across threads, and it's safe to use from a single thread.

libdd-trace-utils/src/change_buffer/mod.rs
yannham: That's a fair comment. I think most of this code is rather verbose but not so complicated; there's a good amount of caches, so structs with many fields and a bunch of specialized functions. This was put together while working on the dd-trace-js integration and will be used only here for now. One avenue is that since our performance exploration was not "linear", it's possible that some optimizations that we did at the beginning for limited gains end up not being really required now that we've had much bigger gains in other part of the Rust-JS interop.

We should benchmark once it can be used in dd-trace-js in a more end-to-end way and cut the fluff that doesn't make much difference.

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +88,793 @@ pub mod span_header;
...
+    tracer_service: T::Text,
+    tracer_language: T::Text,
+    pid: u32,
+    /// Default meta tags automatically applied to every new span via create_span.
+    default_meta: Vec<(T::Text, T::Text)>,
+    /// Contiguous array of span headers for direct JS DataView access. JS writes numeric and
+    /// string-ID fields directly here. Rust reads them during flush_chunk.
+    pub span_headers: Vec<SpanHeader>,
+    /// Free list of recycled header indices (from finished spans).
+    header_free_list: Vec<u32>,
```
paullegranddc: discussed before but I thinl this isn't used anywhere

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +88,793 @@ pub mod span_header;
...
+    pub fn string_table_len(&self) -> usize {
+        self.string_table.len()
+    }
+
+    pub fn span_pool_len(&self) -> usize {
+        self.span_pool.len()
+    }
+
+    pub fn recycle_spans(&mut self, spans: Vec<Span<T>>) {
+        let available = 128usize.saturating_sub(self.span_pool.len());
```
paullegranddc: The 128 here should probably be a named constant, like `MAX_POOL_SPANS`

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +88,793 @@ pub mod span_header;
...
-}
 
-fn deferred_metric_insert(vec: &mut Vec<(u32, f64)>, key_id: u32, val: f64) {
-    for entry in vec.iter_mut() {
-        if entry.0 == key_id {
-            entry.1 = val;
-            return;
+    pub fn flush_chunk(
+        &mut self,
+        slot_indices: Vec<u32>,
```
paullegranddc:
```suggestion
        slot_indices: &[u32],
```

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +88,793 @@ pub mod span_header;
...
+    fn materialize_deferred_tags(&mut self, slot: u32, span: &mut Span<T>) {
+        let idx = slot as usize;
+        if idx < self.deferred_meta.len() {
+            let pairs: Vec<(u32, u32)> = self.deferred_meta[idx].drain(..).collect();
+
+            for (key_id, val_id) in pairs {
+                if let (Some(key), Some(val)) = (self.get_string(key_id), self.get_string(val_id)) {
+                    span.meta.insert(key, val);
+                }
+            }
```
paullegranddc:
```suggestion
            for (key_id, val_id) in &self.deferred_meta[idx] {
                if let (Some(key), Some(val)) = (self.get_string(*key_id), self.get_string(*val_id)) {
                    span.meta.insert(key, val);
                }
            }
            self.deferred_meta.clear();
```

Same remark for deferred_metrics

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +88,793 @@ pub mod span_header;
...
+                    span.metrics.insert(key, val);
+                }
+            }
+        }
+    }
+
+    pub fn materialize_slot(&mut self, slot: u32) {
+        let idx = slot as usize;
+        let mut meta_pairs: Vec<(T::Text, T::Text)> = Vec::new();
+        let mut metric_pairs: Vec<(T::Text, f64)> = Vec::new();
```
paullegranddc: are the intermediary vec here really necessary?

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +88,793 @@ pub mod span_header;
...
+            span.meta
+                .insert(self.str_base_service.clone(), self.tracer_service.clone());
+        }
+
+        span.meta
+            .insert(self.str_language.clone(), self.tracer_language.clone());
+        span.metrics
+            .insert(self.str_process_id.clone(), f64::from(self.pid));
+
+        if let Some(trace) = self.traces.get(&span.trace_id) {
```
paullegranddc: This looks like a lookup we do per span, but all spans in a trace chunks are process together and get the same value

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +88,793 @@ pub mod span_header;
...
+    fn copy_in_chunk_tags(&self, span: &mut Span<T>) {
+        if let Some(trace) = self.traces.get(&span.trace_id) {
+            span.meta
+                .extend(trace.meta.iter().map(|(k, v)| (k.clone(), v.clone())));
+            span.metrics
+                .extend(trace.metrics.iter().map(|(k, v)| (k.clone(), *v)));
+        }
+    }
+
+    fn process_one_span(&self, span: &mut Span<T>) {
```
paullegranddc: I'd prefer `process_span` but 🤷

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +88,793 @@ pub mod span_header;
...
+
+            if is_local_root {
+                self.copy_in_sampling_tags(&mut span);
+                span.metrics.insert(self.str_top_level.clone(), 1.0);
+                is_local_root = false;
+            }
+            if is_chunk_root {
+                self.copy_in_chunk_tags(&mut span);
+                is_chunk_root = false;
+            }
```
paullegranddc, line 300: This logic here would indicate that we have only one chunk root, so one chunk

All spans in a chunk are supposed to belong to the same trace.
But then under we count the trace_ids we've see 🤔 
So either the seen_trace_ids vec can only contain one trace id, or we are not tagging all traces

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -90,43 +88,793 @@ pub mod span_header;
...
+        // pointers to save intermediate dereferences.
+        //
+        // They are safe to use as long as we don't modify their parent vector and keep a unique
+        // borrow to `self` overall.
+        //
+        // Using mutable references instead would be better, but is not possible (or very hard)
+        // given we iterate and nest calls to different submethods taking a `&mut self`.
+        let mut cached_span_ptr: *mut Span<T> = std::ptr::null_mut();
+        let mut cached_deferred_meta: *mut VecMap<u32, u32> = std::ptr::null_mut();
+        let mut cached_deferred_metrics: *mut VecMap<u32, f64> = std::ptr::null_mut();
```
paullegranddc: do we really need this optimization now that we index into a vec?

libdd-trace-utils/src/change_buffer/mod.rs
yannham: I'm pretty sure I tried that and it didn't work. Or maybe it was lost in the rebase? :thinking: let me try again

libdd-trace-utils/src/change_buffer/mod.rs
yannham: I'm wondering as well. I tended to keep the spirit of the original implementation, but that would get rid of a bunch of unsafe...

libdd-trace-utils/src/change_buffer/mod.rs
yannham: Aren't `T::Text` pointers in most cases? Or at least value that would get the niche optimization of `Option`? It could be interesting to look at it, but I feel like this is premature optimization without looking closely at more data. I'll record the idea somewhere though.

libdd-trace-utils/src/change_buffer/mod.rs
yannham: To be sure to understand what you're saying, you're suggestion the trace should bet fetched once and for all by the caller and provided to this helper instead?

libdd-trace-utils/src/change_buffer/mod.rs
yannham: I suppose they were introduced for borrowing reasons, as many methods take `self` and thus can't be used while iterating mutably over one of the internal vec. One possibility would be to push many helper down (so `get_xxx_arg` to `ChangeBuffer`, introduce a newtype `StringTable` so that `get_string` can be moved there, etc.) better splitting the borrow over `self`. I'll see how it look.

rochdev (changes requested):
There are a few issues with this implementation across the board which would be a bit difficult to explain so I just opened a PR on top of this PR instead: https://github.com/DataDog/libdatadog/pull/2105

Feel free to reimplement based on the summary if that's easier, I thought it would be a smaller change when I started on it 😅

yannham:
@rochdev would that be ok to just merge this one first (after addressing reviews) and then proceed with your PR? I don't expect to do a lot of modifications and the code isn't used anywhere yet anyway. That'll just make for an easier review experience. If yes, please lift the request for changes.

libdd-trace-utils/src/change_buffer/mod.rs
yannham: Got rid of the slot

rochdev (approved):

libdd-trace-utils/src/change_buffer/buffer.rs
yannham: In the end I don't think re-eusing slice for two fields is worth it, plus now I use `NonNull`.

yannham: Fixed. Mentioned in the doc now.

libdd-trace-utils/src/change_buffer/mod.rs
yannham: I got rid of deferred stuff in general. I want to see it has any impact first.

paullegranddc (approved):

PR #2061, opened 2026-05-29
feat(shared-runtime)!: SharedRuntime Borrowed & Owned mode

Aaalibaba42:
# What ?

Split `SharedRuntime` into a trait with three implementations:

- `ForkSafeRuntime` *(native only)*: owns a multi-thread tokio runtime, supports fork
  hooks (`before_fork` / `after_fork_parent` / `after_fork_child`) and synchronous
  `shutdown`.
- `BasicRuntime` *(native only)*: wraps a library-built or caller-provided
  `Arc<tokio::runtime::Runtime>`, no fork hooks, no synchronous shutdown.
- `LocalRuntime` *(wasm32 only)*: single-threaded executor that spawns workers via
  `wasm_bindgen_futures::spawn_local`, no fork protocol, async-only.

`BlockingRuntime` *(native only)* is a sub-trait of `SharedRuntime` that adds
`block_on`; implemented by `ForkSafeRuntime` and `BasicRuntime`. Sync facades on the
trace exporter (`send` / `shutdown` / `build`) bound their runtime parameter on it.

`TraceExporter` is generic over the runtime — `TraceExporter<C, R: SharedRuntime>`
— so callers pick `ForkSafeRuntime`, `BasicRuntime`, or `LocalRuntime` explicitly. The
FFI pins `R = ForkSafeRuntime` for ABI stability.

# Why ?

Some callers already have a tokio runtime they want to reuse instead of letting
libdatadog create its own. The previous `SharedRuntime` only supported the
fork-safe owned case. Splitting into distinct types makes the lifecycle and
fork-safety contract explicit and lets callers pick the model that matches their
environment.

The wasm32 target previously shared a file with the native implementation behind
`#[cfg]` walls; it is now a dedicated module with a clean separation.

# How ?

- Introduce a `SharedRuntime` trait (`new`, `spawn_worker`, `shutdown_async`).
- Add a native-only `BlockingRuntime: SharedRuntime` sub-trait with `block_on`.
- Move the existing owned-runtime logic into `ForkSafeRuntime`; fork hooks and sync
  `shutdown` are inherent methods (not part of the trait).
- Add `BasicRuntime::with_worker_threads` (library-built) and
  `BasicRuntime::from_handle(Arc<Runtime>)` (caller-provided).
- Extract the wasm32 `spawn_local` path from `fork_safe.rs` into its own
  `local.rs` module as `LocalRuntime`.
- Make `TraceExporter<C, R>` and `TraceExporterBuilder<R>` generic over the runtime;
  sync entry points additionally require `R: BlockingRuntime`. `Default` for the
  builder is impl'd only for `R = ForkSafeRuntime` on native so
  `TraceExporterBuilder::default()` resolves unambiguously.
- `restart_on_fork = true` is silently ignored (with a `warn!`) on `BasicRuntime`
  and `LocalRuntime` since they do not implement a fork protocol.
- Update FFI to pin `R = ForkSafeRuntime` in its `TraceExporter` type alias.

# Additional Notes

- Breaking change for Rust callers of `libdd_shared_runtime`: `SharedRuntime` is
  now a trait; use `ForkSafeRuntime::with_worker_threads(1)` (or the trait method
  `SharedRuntime::new()` with the trait in scope) instead of `SharedRuntime::new()`.
- FFI handle type changes from `SharedRuntime` to `ForkSafeRuntime`.


libdd-shared-runtime/src/shared_runtime/borrowed.rs
```diff
@@ -0,0 +1,232 @@
...
+        // Poison-tolerant: counter state has no invariants a panicking holder could break.
+        let state = self
+            .shutdown_tracker
+            .state
+            .lock()
+            .unwrap_or_else(|poison| poison.into_inner());
+        let (_state, res) = self
+            .shutdown_tracker
+            .cv
+            .wait_timeout_while(state, timeout, |s| s.completed < s.expected)
```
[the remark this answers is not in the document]
Aaalibaba42: Introducing a `WouldDeadLock` error to address this, suggesting to the caller this ought be done on a separate thread.

Aaalibaba42:
@codex review

libdd-shared-runtime/src/shared_runtime/borrowed.rs
```diff
@@ -0,0 +1,252 @@
...
+        // current-thread, parking this thread on the condvar starves them and we deadlock
+        // until `timeout` fires; reject the call instead.
+        if self.handle.runtime_flavor() == RuntimeFlavor::CurrentThread {
+            return Err(SharedRuntimeError::WouldDeadlock);
+        }
+
+        let (_state, res) = self
+            .shutdown_tracker
+            .cv
+            .wait_timeout_while(state, timeout, |s| s.completed < s.expected)
```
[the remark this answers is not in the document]
Aaalibaba42: `RuntimeFlavor::CurrentThread` doesn't cover `worker_threads(1)` multi-thread runtimes, and I can't detect that case programmatically for tokio doesn't expose worker count or distinguish executor threads from blocking threads through the Handle API...

I've changed the documentation to better reflect this as a caller responsibility.

VianneyRuhlmann:
It looks like the description is a bit off mentioning fixes and bugs that seem to be coming from previous commits of the branch and not from the current state of the feature

libdd-shared-runtime/src/shared_runtime/borrowed.rs
```diff
@@ -0,0 +1,259 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Borrowed-mode runtime.
+//!
+//! [`BorrowedRuntime`] runs workers on a caller-owned tokio runtime (via [`Handle`]), avoiding a
+//! second tokio runtime in the process ("tokio-inside-tokio" panics).
+//!
+//! Fork hooks, synchronous `block_on`, and runtime shutdown are absent — those would touch a
+//! runtime or tasks we don't own. Workers *we* spawned can still be stopped via
```
VianneyRuhlmann: I don't think that's true, the shared runtime only cancels it's own tasks (i.e. PausableWorkers) and dropping the runtime wouldn't impact the rest of the app since the runtime is wrapped in an Arc. Still I agree that these shouldn't be implemented since the fork hooks wouldn't be able to drop the runtime + there's a risk of blocking within a async context.

Aaalibaba42:
> It looks like the description is a bit off mentioning fixes and bugs that seem to be coming from previous commits of the branch and not from the current state of the feature

I tried letting a cheap AI read the commits and make a description, I don't think it's too valuable and can even be detrimental, as it were.

Aaalibaba42:
I rewrote it cleaner and straighter to the point

libdd-shared-runtime/src/shared_runtime/borrowed.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub fn from_handle(handle: Handle) -> Self {
+        Self {
+            handle,
+            registry: WorkerRegistry::new(),
+            shutdown_tracker: Arc::new(ShutdownTracker::default()),
+        }
+    }
+
+    /// Returns a clone of the caller-supplied tokio runtime handle.
+    pub fn runtime_handle(&self) -> Handle {
```
yannham: Most `Handle` methods take `&self`. Any reason to do a preventive clone here, instead of having a getter and letting the caller decide to clone if needed?

libdd-shared-runtime/src/shared_runtime/borrowed.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Fire-and-forget pause for any still-running workers; never blocks.
+    ///
+    /// For a guaranteed drain, call [`trigger_shutdown_signal`](Self::trigger_shutdown_signal) +
+    /// [`wait_shutdown_done`](Self::wait_shutdown_done) before dropping.
+    fn drop(&mut self) {
+        let workers = {
+            let mut guard = match self.registry.workers.lock() {
+                Ok(g) => g,
+                Err(poison) => poison.into_inner(),
+            };
```
yannham: Nit

```suggestion
            let mut guard = self.registry.workers.lock().unwrap_or_else(|poison| poison.into_inner());
```

libdd-shared-runtime/src/shared_runtime/borrowed.rs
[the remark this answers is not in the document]
yannham: I wonder if this kind of requirement is practical. If the borrowed runtime is, downstream, some random client's app tokio runtime, do we really have any control over how it is configured? I'm not sure how `wait_shutdown_done` is called, but if it's called in an async context, I suppose we could design an async wait operation instead (e.g., when polled, the future checks if `s.completed < s.expected`)?

libdd-shared-runtime/src/shared_runtime/borrowed.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Does not affect host-runtime tasks. Worker errors are logged. Use either this or the
+    /// [`trigger_shutdown_signal`](Self::trigger_shutdown_signal) pair — not both.
+    pub async fn shutdown_async(&self) {
+        // Set triggered before draining so concurrent spawn_worker calls are rejected first.
+        // Poison-tolerant: this function doesn't fail.
+        {
+            let mut state = self
+                .shutdown_tracker
+                .state
+                .lock()
```
yannham: We are potentially locking a sync mutex in a tokio runtime, which might block the executor's thread. Is that ok in the current context (and if yes, why?)? I'm not sure to have the bigger picture. Are the other mutex-locking functions like `spawn_worker` and cos also expected to be run from within a runtime? I guess if yes, we can just use a tokio async mutex. If the mutex can be locked/unlocked from both within and outside the runtime, then I'm not sure...

yannham (commented):
The code LGTM overall, but I'm not entirely sure to fully grasp in which context (sync or async) is this API supposed to be consumed. In particular we use standard sync thread mechanisms (condvars, mutexes) in an async context, which could be problematic (even in a multi-thread context, a condvar will block the current executor thread, which is not ideal).

On a different front, I remember @paullegranddc said he would rather spawn our own shared runtime in a separate thread than hooking in the client's runtime. I don't have a strong opinion myself, but I wonder if this discussion had a conclusion.

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
```diff
@@ -0,0 +1,630 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
```
Aaalibaba42, line 1: This whole file's cfg gates are pretty akward, I guess I could split a `OwnedNative` and `OwnedWasm` versions in separate files and just re export the correct one in here, but it seems clunky.

Aaalibaba42:
Since the implementation was overhauled, I'll just resolve past conversations here, as they don't fit the current implementation.

libdd-shared-runtime/src/shared_runtime/basic.rs
```diff
@@ -0,0 +1,222 @@
...
+/// A [`SharedRuntime`] backed by a caller-owned `tokio::runtime::Runtime`.
+///
+/// Holds an `Arc<Runtime>` so the runtime stays alive as long as this struct does,
+/// even if the original owner drops their clone. The runtime itself is not affected
+/// by operations on this struct — only the workers tracked here are. Borrowed-mode
+/// is **not** fork-safe and does not provide a synchronous shutdown.
+#[derive(Debug)]
+pub struct BorrowedSharedRuntime {
+    runtime: Arc<tokio::runtime::Runtime>,
+    workers: Arc<Mutex<Vec<WorkerEntry>>>,
```
yannham, line 30: Should we use a blocking mutex here instead of an async one? Since we use the mutex from both sync (`spawn_worker`)and async (`shutdown_async`), it looks like an async mutex is more appropriate, as it won't block the executor thread and still offer a `lock_blocking` method.

libdd-shared-runtime/src/shared_runtime/borrowed.rs
```diff
@@ -0,0 +1,222 @@
...
+            worker: pausable_worker,
+        });
+
+        Ok(WorkerHandle {
+            worker_id,
+            workers: self.workers.clone(),
+        })
+    }
+
+    fn runtime_handle(&self) -> Result<tokio::runtime::Handle, SharedRuntimeError> {
```
yannham: I think we can do most operations on a handle by reference; why do we clone it here?

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -345,6 +138,8 @@ pub enum SharedRuntimeError {
     RuntimeCreation(io::Error),
     /// Shutdown timed out.
     ShutdownTimedOut(std::time::Duration),
+    /// The requested operation is not supported on a borrowed-handle runtime.
+    UnsupportedInBorrowedMode,
```
yannham: Maybe this question will be answered later, but which operations return this error? Since you've split the fork support from the base, I would expect that the `SharedRuntime` trait API is fully supported by the borrowed runtime.

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
yannham: 👍 this. Conditional compilation makes things hard to read really fast. If you want to keep everything in one place, you can also use local submodules `native` and `wasm` if you prefer, at least it's only one annotation per submodule (plus the one for the export).

libdd-shared-runtime/src/shared_runtime/owned.rs
```diff
@@ -0,0 +1,634 @@
...
+            })
+        }
+    }
+
+    /// Creates an `OwnedSharedRuntime` with the given runtime flavor.
+    /// For [`OwnedKind::CurrentThread`] a dedicated driver thread is also spawned.
+    #[cfg(not(target_arch = "wasm32"))]
+    pub fn with_kind(kind: OwnedKind) -> Result<Self, SharedRuntimeError> {
+        let runtime = Arc::new(build_runtime_for(&kind)?);
+        let driver = if matches!(kind, OwnedKind::CurrentThread) {
```
yannham: Pure nitpick/taste, I prefer to avoid macros when there's a core equivalent core construct, but feel free to ignore.

```suggestion
        let driver = if let OwnedKind::CurrentThread = kind {
```

yannham (approved):
Left a few remarks but otherwise looks good (only skimmed through the native implementation, as I suppose it's just the original shared runtime moved)👍

libdd-shared-runtime/src/shared_runtime/owned.rs
Aaalibaba42: I find this syntax really weird, it reads like trying to bind an lvalue to an xvalue (maybe it's my C++ days warping my view). I don't mind deriving PartialEq and `==` but `matches` seemed more colloquial.

libdd-shared-runtime/src/shared_runtime/mod.rs
Aaalibaba42: For instance if you spawn a worker with `restart_on_fork`. We could separate the API in separate functions, this would make this error variant not needed, but it would also be more uneasy imho, this error is explicit enough and would trigger immediately if the shared runtime is used incorrectly, so I don't see it as a problem but I might be wrong.

libdd-shared-runtime/src/shared_runtime/basic.rs
Aaalibaba42: As discussed with @VianneyRuhlmann, the tokio doxa trends more towards prefering `std::sync::Mutex` as long as you don't hold a Mutex accross `await` points.

- In `spawn_worker` it is locked, we push a `WorkerEntry`, and we drop -> no await
- In `shutdown_async` we lock, `mem::take` the vector, then dropped. The async happen on the moved-out vector after the guard is dropped.

libdd-shared-runtime/src/shared_runtime/borrowed.rs
Aaalibaba42: A `tokio::runtime::Handle` is already an `Arc` under the hood, and giving ownership is more ergonomic than managing references ig

libdd-shared-runtime/src/shared_runtime/owned.rs
yannham: if-let is just an `if` version of pattern matching / destructuring for me, where an arm can bind 0 or more variables :slightly_smiling_face: but it's really unimportant so do as you feel is right

yannham: (And agreed on the `PartialEq` -- we shouldn't use it just for this, match-like construct are better suited)

libdd-shared-runtime/src/shared_runtime/borrowed.rs
```diff
@@ -0,0 +1,222 @@
...
+use super::pausable_worker::{tokio_spawn_fn, PausableWorker};
+use super::{BoxedWorker, SharedRuntime, SharedRuntimeError, WorkerEntry, WorkerHandle};
+use crate::worker::Worker;
+use futures::stream::{FuturesUnordered, StreamExt};
+use libdd_common::MutexExt;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::{Arc, Mutex};
+use tracing::{debug, error};
+
+/// A [`SharedRuntime`] backed by a caller-owned `tokio::runtime::Runtime`.
```
VianneyRuhlmann: The runtime is not really owned by the caller since it's in an Arc. It's rather that the underlying tokio::Runtime is exposed to the caller

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -1,266 +1,59 @@
...
+    /// it; `false` drops it without calling shutdown. [`BorrowedSharedRuntime`] rejects `true`.
+    fn spawn_worker<T: Worker + Sync + 'static>(
+        &self,
+        worker: T,
+        restart_on_fork: bool,
+    ) -> Result<WorkerHandle, SharedRuntimeError>;
+
+    /// Returns the runtime handle, or [`SharedRuntimeError::RuntimeUnavailable`] if shut down.
+    #[cfg(not(target_arch = "wasm32"))]
+    fn runtime_handle(&self) -> Result<tokio::runtime::Handle, SharedRuntimeError>;
```
VianneyRuhlmann: I think this method should not be public. You can remove it from the trait.

libdd-shared-runtime/src/shared_runtime/mod.rs
VianneyRuhlmann: If this is just for the `restart_on_fork` I don't think this is needed, this parameter has been added as workaround for python but I'm planning to remove it as part of the telemetry refactor so you can just ignore it in the BorrowedRuntime implementation

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
```diff
@@ -0,0 +1,694 @@
...
+    };
+
+    /// Flavor of tokio runtime constructed by [`OwnedSharedRuntime`].
+    #[derive(Debug, Clone)]
+    pub enum OwnedKind {
+        /// Multi-threaded runtime with the given worker-thread count.
+        MultiThread { worker_threads: usize },
+        /// Current-thread runtime; a dedicated driver thread is spawned so tasks make progress.
+        CurrentThread,
+    }
```
VianneyRuhlmann: Maybe I'm mistaken but this CurrentThread flavor is only to be used for dd-trace-rs right ?
If it's the case I don't think the logic to run such a runtime should live here:
1) You could reuse the BorrowedSharedRuntime and spawn the thread to run the current thread runtime in dd-trace-rs
2) Some method are still using `block_on` which would crash if called within runtime context (which could happen in dd-trace-rs)
3) You don't need the fork hooks anyway since dd-trace-rs doesn't fork.

libdd-shared-runtime/src/shared_runtime/owned.rs
```diff
@@ -0,0 +1,694 @@
...
+    /// Owns a tokio runtime and manages [`PausableWorker`]s on it.
+    ///
+    /// Supports the full fork protocol ([`before_fork`](Self::before_fork) /
+    /// [`after_fork_parent`](Self::after_fork_parent) /
+    /// [`after_fork_child`](Self::after_fork_child)) and synchronous [`shutdown`](Self::shutdown).
+    ///
+    /// # Mutex lock order
+    /// `runtime` → `driver` → `workers`. Avoid holding multiple locks simultaneously.
+    #[derive(Debug)]
+    pub struct OwnedSharedRuntime {
```
VianneyRuhlmann: Regarding naming, the real reason we need this SharedRuntime to "own" the tokio runtime is to manage forks maybe this one should be named `ForkSafeSharedRuntime`

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
```diff
@@ -0,0 +1,694 @@
...
+    fn spawn_current_thread_driver(
+        rt: &Arc<Runtime>,
+    ) -> Result<CurrentThreadDriver, SharedRuntimeError> {
+        let cancel = CancellationToken::new();
+        let cancel_clone = cancel.clone();
+        let rt = rt.clone();
+        let thread = std::thread::Builder::new()
+            .name("shared-runtime-driver".into())
+            .spawn(move || {
+                rt.block_on(cancel_clone.cancelled());
```
bantonsson: Are we guaranteed that `shutdown` is called so that this thread wakes up? If we just `drop` the `OwnedSharedRuntime` it won't cancel this one, and the thread doesn't shut down.

libdd-shared-runtime/src/shared_runtime/owned.rs
```diff
@@ -0,0 +1,694 @@
...
+                .as_ref()
+                .ok_or(SharedRuntimeError::RuntimeUnavailable)?
+                .handle()
+                .clone())
+        }
+
+        /// Pauses all workers before `fork()`. Worker pause errors are logged, not propagated.
+        pub fn before_fork(&self) {
+            debug!("before_fork: pausing all workers");
+            if let Some(runtime) = self.runtime.lock_or_panic().take() {
```
bantonsson: Should we make the mutex guard explicit here? It's not obvious from the code that a lock is being held for the whole duration of the call.

libdd-shared-runtime/src/shared_runtime/borrowed.rs
Aaalibaba42: I think this comment was made before we changed strategies, I don't think it still applies, does it ?

libdd-shared-runtime/src/shared_runtime/mod.rs
Aaalibaba42: I removed it completely, I believe I remember us discussing it was not useful

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
Aaalibaba42: I'll add the cancellation logic to the Drop implementation

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
Aaalibaba42: > this CurrentThread flavor is only to be used for dd-trace-rs right ?

For now, I think so, yes. Maybe in the long run the php sidecar or node native plugins or whatever else that may have Rust in their stack will want to use it but it is not planned as of late if I'm not mistaken.

> If it's the case I don't think the logic to run such a runtime should live here

I thought that making this kind of component more flexible from the get-go was valuable in adapting them to future integrations more easiely, but I agree much of the logic here could be handled in dd-trace-rs directly.

I don't feel strongly about it, I don't mind removing things here. Just voicing a little push-back before committing to it.

libdd-shared-runtime/src/shared_runtime/borrowed.rs
yannham: It's not very important, but I think disagree. If you can do everything by ref, I feel like the general design approach of Rust would be to not clone on behalf on the caller, but let them decide/do it if they need. That gives them more control, and avoid an arc clone. Why giving ownership is more ergonomic?

Granted in this case the clone is probably irrelevant, so whatever :shrug: .

Aaalibaba42:
As it is currently, I don't like how wasm is handled. Sementically it's attached to the ForkSafeRuntime, which is straight up a lie, but changing that would likely require a new implementer of the SharedRuntime trait. I'll sleep on it I'm not sure of the best design

libdd-shared-runtime/src/shared_runtime/borrowed.rs
Aaalibaba42: I think you were right. [This method has disappeared though](https://github.com/DataDog/libdatadog/pull/2061#discussion_r3413217921), so resolving this o/

Aaalibaba42:
@codex review

libdd-shared-runtime/src/shared_runtime/basic.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        let boxed_worker: BoxedWorker = Box::new(worker);
+        debug!(?boxed_worker, "Spawning worker on BasicRuntime");
+        let mut pausable_worker = PausableWorker::new(boxed_worker);
+
+        pausable_worker.start(tokio_spawn_fn(self.runtime.handle()))?;
+
+        let worker_id = self.next_worker_id.fetch_add(1, Ordering::Relaxed);
+
+        self.workers.lock_or_panic().push(WorkerEntry {
```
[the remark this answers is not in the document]
Aaalibaba42: I'd wager the risky window is very narrow, but worth a fix anyway

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
Aaalibaba42: After IRL discussions, we ended up deciding to split the sementic barrier from Owned vs Borrowed to ForkSafe vs NotForkSafe. It makes more sense since we don't really care about the borrowing part, but we care about what the thing can *do*. This addresses also the fact that a CurrentThread mode for the kind OwnedSharedRuntime was not fork safe and so is no longer part of this API.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -251,7 +252,15 @@ impl TraceExporterBuilder {
...
-    pub fn set_shared_runtime(&mut self, shared_runtime: Arc<OwnedSharedRuntime>) -> &mut Self {
+    ///
+    /// Accepts a [`DefaultRuntime`], which resolves to [`ForkSafeRuntime`] on native and
+    /// [`libdd_shared_runtime::LocalRuntime`] on wasm32. On native, the exporter calls
+    /// [`ForkSafeRuntime::block_on`] from its sync entry points and assumes the host wires
+    /// up the fork hooks (`before_fork` / `after_fork_*`) around `fork()` — a plain
+    /// `BasicRuntime` or a raw `tokio::runtime::Runtime` are not accepted here. On wasm32
+    /// the sync entry points do not exist and [`LocalRuntime`](libdd_shared_runtime::LocalRuntime)
+    /// is used for async spawning only.
+    pub fn set_shared_runtime(&mut self, shared_runtime: Arc<DefaultRuntime>) -> &mut Self {
```
VianneyRuhlmann: I think the shared runtime should take any runtime implementing the SharedRuntime trait and choosing the type of the runtime should be up to the caller.

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
```diff
@@ -0,0 +1,426 @@
...
+}
+
+/// Owns a tokio runtime and manages [`PausableWorker`]s on it.
+///
+/// Supports the full fork protocol ([`before_fork`](Self::before_fork) /
+/// [`after_fork_parent`](Self::after_fork_parent) /
+/// [`after_fork_child`](Self::after_fork_child)) and synchronous [`shutdown`](Self::shutdown).
+///
+/// # Mutex lock order
+/// `runtime` → `workers`.
```
VianneyRuhlmann: nit: This shouldn't be in the public doc maybe add a rustdoc comment on the fields themselves.

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -643,18 +643,28 @@ pub trait Export<T>: Send + Debug {
 }
 
 #[derive(Debug)]
-pub struct DefaultExport<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> {
-    trace_exporter: TraceExporter<C>,
+pub struct DefaultExport<C, R>
+where
+    C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static,
+    R: SharedRuntime + std::fmt::Debug + Send + Sync + 'static,
```
Eldolfin, line 649:
```suggestion
    R: SharedRuntime,
```

Is there a reason to not make the SharedRuntime trait require these other trait itself ? Given that all 3 implementors already satisfy `std::fmt::Debug + Send + Sync + 'static`, I feel like this makes more sense

The trait definition could look like this:
```rs
pub trait SharedRuntime: std::fmt::Debug + Send + Sync + 'static {
```

Also applies to `HttpClientCapability`.

libdd-shared-runtime/src/shared_runtime/basic.rs
```diff
@@ -0,0 +1,291 @@
...
+    fn spawn_worker<T: Worker + Sync + 'static>(
+        &self,
+        worker: T,
+        restart_on_fork: bool,
+    ) -> Result<WorkerHandle, SharedRuntimeError> {
+        if restart_on_fork {
+            warn!(
+                "restart_on_fork is ignored on BasicRuntime: regular mode is not fork-safe; \
+                 use ForkSafeRuntime if you need fork hooks"
+            );
```
Eldolfin, line 89: Seems the kind of place a panic would be more appropriate for me: it's a developer error not a end-user configuration issue.
It should never happen for a end-user if the issue is spotted by the developer that misused libdatadog, and a panic is harder to miss than a warning.

libdd-shared-runtime/src/lib.rs
```diff
@@ -8,18 +8,36 @@
...
+//!   [`ForkSafeRuntime::block_on`] and [`ForkSafeRuntime::shutdown`].
+//! - [`BasicRuntime`] *(native only)* — the regular (non-fork-safe) variant. Its internal tokio
+//!   runtime can be library-built ([`BasicRuntime::new`] / [`BasicRuntime::with_worker_threads`])
+//!   or supplied by the caller as an `Arc<tokio::runtime::Runtime>`
+//!   ([`BasicRuntime::from_handle`]).
+//! - [`LocalRuntime`] *(wasm32 only)* — single-threaded local executor; spawns workers via
+//!   `wasm_bindgen_futures::spawn_local`. No fork protocol, no `block_on`, async-only.
+//!
+//! Components such as the trace exporter can share one runtime instead of each creating their
+//! own, reducing thread and resource overhead.
```
Eldolfin, line 12:
```suggestion
//! A shared tokio runtime for running background workers across multiple components.
//!
//! Components such as the trace exporter can share one runtime instead of each creating their
//! own, reducing thread and resource overhead.
//!
//! # Choosing a runtime
//!
//! | Runtime             | Target | Threads | Fork-safe           | `block_on` | When to use                                                                                                                                 |
//! |---------------------|--------|---------|---------------------|------------|---------------------------------------------------------------------------------------------------------------------------------------------|
//! | [`ForkSafeRuntime`] | native | multi   | yes — full protocol | yes        | Default for native code that may run in a forking process (e.g. Ruby, Python runtimes).                                                     |
//! | [`BasicRuntime`]    | native | multi   | no                  | yes        | Native code where `fork()` is not a concern; optionally share an existing `Arc<tokio::runtime::Runtime>` via [`BasicRuntime::from_handle`]. |
//! | [`LocalRuntime`]    | wasm32 | single  | n/a                 | no         | WebAssembly; spawns via `wasm_bindgen_futures::spawn_local`.                                                                                |
//!
//! ## Fork protocol ([`ForkSafeRuntime`] only)
//!
//! Call these around every `fork()` to prevent deadlocks in child processes:
//!
//! 1. [`ForkSafeRuntime::before_fork`] — pauses workers
//! 2. `fork()`
//! 3. parent: [`ForkSafeRuntime::after_fork_parent`] — resumes workers
//! 4. child: [`ForkSafeRuntime::after_fork_child`] — restarts workers on a fresh runtime

```

Easier to read as a table than a dense list imo, especially once rendered. (IA generated table please double check)

Image

Eldolfin (commented):
nits

libdd-data-pipeline/src/telemetry/mod.rs
```diff
@@ -673,7 +673,7 @@ mod tests {
     #[test]
     fn chunks_dropped_serialization_error_test() {
         let payload = Regex::new(r#""metric":"trace_chunks_dropped","points":\[\[\d+,1\.0\]\],"tags":\["src_library:libdatadog","reason:serialization_error"\],"common":true,"type":"count"#).unwrap();
-        let shared_runtime = SharedRuntime::new().expect("Failed to create runtime");
+        let shared_runtime = ForkSafeRuntime::new().expect("Failed to create runtime");
         let server = MockServer::start();
         let mut telemetry_srv = server.mock(|when, then| {
             when.method(POST).body_matches(payload);
```
VianneyRuhlmann, line 679: <p class='graphite__hidden'><i>[Re: line +712]</i></p>
Did you merged these tests on purpose ?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2061?utm_source=unchanged-line-comment">Graphite</a>.</p>

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -252,7 +310,13 @@ impl TraceExporterBuilder {
     }
 
     /// Set a shared runtime used by the exporter for background workers.
-    pub fn set_shared_runtime(&mut self, shared_runtime: Arc<SharedRuntime>) -> &mut Self {
+    ///
+    /// The runtime type is selected via the builder's `R` parameter — typically
+    /// [`libdd_shared_runtime::ForkSafeRuntime`] on native (when the host wants the fork
+    /// protocol), [`libdd_shared_runtime::BasicRuntime`] to wrap a caller-owned tokio
+    /// runtime, or [`libdd_shared_runtime::LocalRuntime`] on wasm. Sync entry points
+    /// (`build` / `send` / `shutdown`) additionally require `R: BlockingRuntime`.
```
VianneyRuhlmann: I think instructions on which runtime to choose should live in the shared runtime crate to avoid ending up with stale documentation.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -179,11 +181,20 @@ impl From<TraceExporterInputFormat> for DeserInputFormat {
...
-/// pin it to a concrete type (`NativeCapabilities` or `WasmCapabilities`).
-/// Task spawning is handled internally by `SharedRuntime`.
+/// `C` is the capabilities bundle (HTTP, sleep). Leaf crates pin it to a concrete type
+/// (`NativeCapabilities` or `WasmCapabilities`).
+///
+/// `R` is the [`SharedRuntime`] used to host background workers. There is no cross-target
+/// default — callers spell the concrete runtime they want
+/// ([`libdd_shared_runtime::ForkSafeRuntime`] for the fork protocol,
+/// [`libdd_shared_runtime::BasicRuntime`] to wrap a caller-owned tokio runtime,
+/// [`libdd_shared_runtime::LocalRuntime`] on wasm). Sync facades such as
```
VianneyRuhlmann: Same this should be documented in the runtime crate

libdd-shared-runtime/src/shared_runtime/basic.rs
VianneyRuhlmann: We want to avoid panics as they don't break the whole customer app not just the tracer.

libdd-shared-runtime/src/shared_runtime/basic.rs
```diff
@@ -0,0 +1,291 @@
...
+            .collect();
+
+        futures.collect::<()>().await;
+    }
+}
+
+impl BlockingRuntime for BasicRuntime {
+    /// Forwards to the underlying tokio runtime. Infallible — the inner `Arc<Runtime>` is
+    /// kept alive for the lifetime of this struct, so there is no fork-window fallback to
+    /// build.
```
VianneyRuhlmann: I don't think you need to explain why this one is infallible compared to the forksafe one

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -1,266 +1,88 @@
...
-                None => Arc::new(Builder::new_current_thread().enable_all().build()?),
-                Some(runtime) => runtime.clone(),
-            };
-            Ok(runtime.block_on(f))
-        }
+/// Common interface for all [`SharedRuntime`] implementations.
+///
+/// Fork hooks are inherent methods on [`ForkSafeRuntime`] only (native target). Synchronous
+/// `block_on` lives on the [`BlockingRuntime`] subtrait, which only the native runtimes
+/// implement. Not object-safe — use `impl SharedRuntime` generics rather than
```
VianneyRuhlmann: I'm not sure we need to call this out, btw it is now called dyn compatibility.

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -1,266 +1,88 @@
...
+/// This is the bound required by sync facades in higher layers (e.g. the trace exporter's
+/// `send` / `build` / `shutdown` methods) to block on async work without entering a tokio
+/// context themselves. Wasm targets do not implement this trait because there is no
+/// thread to block.
+#[cfg(not(target_arch = "wasm32"))]
+pub trait BlockingRuntime: SharedRuntime {
+    /// Drives `f` to completion on the runtime's own executor, blocking the current thread.
+    ///
+    /// Returns an [`io::Error`] if a temporary executor must be built (e.g. in the
+    /// [`ForkSafeRuntime`] fork window) and construction fails.
```
VianneyRuhlmann: All the doc seems very specific to struct implementing the trait (e.g. wasm, forksafe failures) rather than detailling what contract the trait defines.

VianneyRuhlmann (approved):
Few comments regarding the doc but LGTM.  
I think this is a real improvement for the sahred runtime semantics, thanks for doing this

libdd-shared-runtime/src/shared_runtime/basic.rs
Eldolfin: I agree panics are not good for error handling in general, but unless I misunderstood something, hitting this case should only happen when someone using the libdatadog updates and misconfigures the runtime (I assume the runtime/`restart_on_fork` are not user-configurable), so it should be made as obvious as possible that they need to change that before merging.

A warn could be missed if they don't have native logs enabled whereas a panic would immediately fail their tests & ci making it obvious where the problem is and prevent them merging code that has this mistake.

libdd-data-pipeline/src/trace_buffer/mod.rs
Aaalibaba42: I don't think `Debug` should be at the trait level given it sementically is attached to the specific struct and not a contract of the trait itself, but I agree with the rest

libdd-shared-runtime/src/shared_runtime/basic.rs
Aaalibaba42: iiuc this parameter will be gone soon:tm:, that's why it's just a warn here. We definitely don't want to panick, and a warn like that is not great long term, but for something that'll be gone at some point it's a fine and low friction work around. https://github.com/DataDog/libdatadog/pull/2061#discussion_r3413268917

libdd-data-pipeline/src/telemetry/mod.rs
Aaalibaba42: I did not merge any tests (?). I don't see what I'm supposed to notice with graphite, my changes in the file compared to main is just renaming SharedRuntime (which used to be a struct but is now a trait) to ForkSafeRuntime.

VianneyRuhlmann: Sorry nevermind, since I was reviewing diff between two commits graphite showed me the `telemetry_from_ok_response_with_p0_drops_test` test that was remove by merging main 😩

libdd-data-pipeline/src/trace_buffer/mod.rs
Aaalibaba42: It breaks stuff around wasm because JSValues etc for the bindings are !Sync and !Send..... I think I'll just rollback. Even though I agree it's better to move the constraints up, it leads to boilerplate that I don't really like :/

PR #2064, opened 2026-05-29
feat(data-pipeline): move the async boundary up

Aaalibaba42:
# Summary

Today `SharedRuntime::block_on` is buried deep inside the data pipeline (in
`check_agent_info`, `stop_stats_computation`, telemetry start-up and `send_deser`).
That forces every internal path to block on a runtime even when the caller is
already async, and makes it impossible for an async host to drive libdatadog
without nested-runtime hazards.

This PR pushes the async/sync boundary **up** to the top-level public entry
points. The internals are now async all the way down, and the synchronous
methods are thin `block_on` facades over their `_async` counterparts. This is a
prerequisite for letting async consumers (e.g. a Rust service embedding
`dd-trace-rs`) call the `_async` API directly.

The change is self-contained to `libdd-data-pipeline` and does not touch
`libdd-shared-runtime`; the sync facades keep using the existing owned-mode
`SharedRuntime::block_on`.

# What changed

`trace_exporter/mod.rs`
- `send`, `send_trace_chunks`, `shutdown` are now thin `block_on` facades over
  `send_async`, `send_trace_chunks_async` and the new `shutdown_async`.
- `check_agent_info` is now `async` (native + wasm); it awaits the stats path
  and takes an owned status snapshot via `load_full()` to avoid holding a
  `!Send` `ArcSwap` guard across `.await`.
- Removed `send_deser` — its deserialize-and-send logic already lives in
  `send_async`.

`trace_exporter/builder.rs`
- Added `build_async` (the real builder body); `build` is now a `block_on`
  facade that materializes the `SharedRuntime` up front so both share a single
  instance.
- Telemetry start-up is awaited directly (`client_tel.start().await`) instead of
  `block_on`.
- The `!Send` `Handle::enter()` guard is scoped to a block so it drops before any
  `.await`, keeping the future `Send`.

`trace_exporter/stats.rs`
- `stop_stats_computation` and `handle_stats_enabled` are now `async` and await
  the stats-worker shutdown directly instead of routing through `block_on`.

# Notes / compatibility

- Public sync signatures (`send`, `send_trace_chunks`, `shutdown`, `build`) are
  unchanged, so FFI callers are unaffected.
- The sync `build` facade is gated to native (`#[cfg(not(target_arch = "wasm32"))]`),
  matching the existing convention where `send`/`send_trace_chunks` are
  native-only and wasm consumers use the `_async` variants (`block_on` does not
  exist on wasm). `build_async` is cross-platform.

### DIFF
--- a/libdd-data-pipeline/src/trace_exporter/builder.rs
+++ b/libdd-data-pipeline/src/trace_exporter/builder.rs
@@ -296,8 +296,33 @@ impl TraceExporterBuilder {
         self
     }
 
-    #[allow(missing_docs)]
+    /// Build the [`TraceExporter`] synchronously.
+    ///
+    /// Sync facade over [`Self::build_async`]; panics inside an existing tokio context.
+    /// Not available on wasm — use [`Self::build_async`] there.
+    #[cfg(not(target_arch = "wasm32"))]
     pub fn build<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static>(
+        mut self,
+    ) -> Result<TraceExporter<C>, TraceExporterError> {
+        let shared_runtime = match self.shared_runtime.as_ref() {
+            Some(rt) => rt.clone(),
+            None => {
+                let rt = Self::new_shared_runtime()?;
+                self.shared_runtime = Some(rt.clone());
+                rt
+            }
+        };
+        shared_runtime.block_on(self.build_async::<C>())?
+    }
+
+    /// Build the [`TraceExporter`] asynchronously.
+    ///
+    /// Awaits all async setup (e.g. telemetry start-up). `C::new_client()` is invoked
+    /// inside `shared_runtime`'s tokio context via a scoped `EnterGuard`, so the caller
+    /// does not have to already be on that runtime.
+    pub async fn build_async<
+        C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static,
+    >(
         self,
     ) -> Result<TraceExporter<C>, TraceExporterError> {
         if !Self::is_inputs_outputs_formats_compatible(self.input_format, self.output_format) {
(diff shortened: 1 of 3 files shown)


hoolioh (approved):

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -318,18 +341,20 @@ impl TraceExporterBuilder {
...
+            let _guard = shared_runtime
+                .runtime_handle()
+                .map_err(|e| {
+                    TraceExporterError::Builder(BuilderErrorKind::InvalidConfiguration(
+                        e.to_string(),
+                    ))
+                })?
+                .enter();
+            C::new_client()
+        };
```
VianneyRuhlmann: Looking at the new_client impl I don't see any reference to the runtime handle. Anyway the runtime should already be accessible since the method is async. Finally capabilities shouldn't hold a runtime handle since the runtime may be recreated when we fork.

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -254,9 +249,9 @@ fn update_obfuscation_config(
 }
 
 #[cfg(not(target_arch = "wasm32"))]
-/// Handle stats computation when it's already enabled
-pub(crate) fn handle_stats_enabled(
-    ctx: &StatsContext,
+/// Reconcile stats config when computation is already enabled.
+/// Async because the disabled-by-agent branch awaits [`stop_stats_computation`].
```
VianneyRuhlmann: nit: I don't think that's needed in the doc

VianneyRuhlmann (approved):

Aaalibaba42:
#test Windows latest is too slow for all_checks so failing merge queue :c

PR #2067, opened 2026-06-01
feat(data-pipeline)!: export client-computed span stats as OTLP trace metrics

mabdinur:
## Overview

Adds native OTLP trace-metrics export to \`libdd-data-pipeline\`. The span concentrator's per-cell stats are encoded as a \`traces.span.sdk.metrics.duration\` delta histogram and POSTed to a configured \`/v1/metrics\` endpoint on a 10s cadence. This is the shared Rust implementation consumed by dd-trace-py and other Datadog tracers — tracers only supply configuration.

**Non-obvious details for reviewers**

- \`count\`/\`sum\`/\`min\`/\`max\` are **exact** (from per-cell accumulators, ns→s), not reconstructed from the DDSketch. The bucket distribution is projected from the DDSketch onto a fixed 17-bucket layout matching the OTel spanmetrics-connector defaults so payloads are comparable across tracers.
- With OTLP metrics enabled, \`check_agent_info\` still runs V1 protocol negotiation but skips stats enable/disable — the concentrator starts unconditionally and is not affected by agent info.
- \`_dd.stats_computed: "true"\` is added to every outbound OTLP trace \`ResourceSpans\` (and \`Datadog-Client-Computed-Stats: yes\` to the HTTP request) when stats are enabled. The Agent's OTLP receiver checks this attribute to skip its concentrator, preventing double-counted APM metrics. Set unconditionally regardless of OTel semantics mode. Both signals are backwards compatible with released Agents.
- \`grpc_method\` is stored in \`GroupedStats\` (value side), not in \`FixedAggregationKey\`. This keeps it out of \`Hash\`/\`PartialEq\` so gRPC spans with different methods merge into one agent stats bucket — matching pre-existing behaviour. The method is still surfaced on \`OtlpExactGroup\` for OTLP data points. No breaking changes.
- \`TraceExporterBuilder::set_runtime_id\` lets callers pass the language tracer's existing runtime ID. When set, it is shared across OTLP trace exports and OTLP trace-metrics so all signals can be correlated by the backend; when not set, a UUID is generated.
- OTLP metrics exports forward \`OTEL_EXPORTER_OTLP_METRICS_HEADERS\` and \`OTEL_EXPORTER_OTLP_METRICS_TIMEOUT\` from the builder, matching the behaviour of the OTLP trace exporter.

**Testing**: 160 nextest tests pass across \`libdd-trace-stats\` and \`libdd-data-pipeline\`. Companion: DataDog/dd-trace-py#18354.


libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -109,6 +109,14 @@ fn build_resource(resource_info: &OtlpResourceInfo) -> Resource {
             value: AnyValue::StringValue(resource_info.runtime_id.clone()),
         });
     }
+    // Tells Datadog Agent OTLP receivers to skip their concentrator; prevents double-counted APM
+    // metrics.
+    if resource_info.client_computed_stats {
+        attributes.push(KeyValue {
+            key: "_dd.stats_computed".to_string(),
+            value: AnyValue::StringValue("true".to_string()),
```
mabdinur, line 117: Required when an otlp trace payload is forwarded from a collector to a Datadog agent. In this scenario the stats computed header is lost and without this tag we will recompute stats: https://github.com/DataDog/datadog-agent/blob/7ac1bdeef284238c7aa8a716f26d1eb3fb2b7417/pkg/trace/api/otlp.go#L51

libdd-data-pipeline/src/otlp/exporter.rs
```diff
@@ -5,30 +5,31 @@
...
 use libdd_trace_utils::send_with_retry::{
     send_with_retry, RetryBackoffType, RetryStrategy, SendWithRetryError,
 };
+use std::time::Duration;
 
-/// Max retries for OTLP export.
-const OTLP_MAX_RETRIES: u32 = 4;
-/// Initial backoff between retries (milliseconds).
+/// Max total attempts for OTLP export (initial + retries on transient failures).
+pub(crate) const OTLP_MAX_ATTEMPTS: u32 = 5;
```
VianneyRuhlmann: nit: Any reason to use attempts instead of retry. I'm trying to converge to using retries as mixing the two can cause bugs.

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -369,26 +433,55 @@ impl StatsBucket {
         duration: i64,
         is_error: bool,
         is_top_level: bool,
+        grpc_method: &str,
     ) {
         self.data
             .entry_ref(&key)
-            .or_default()
+            .or_insert_with(|| GroupedStats {
+                grpc_method: grpc_method.to_owned(),
```
VianneyRuhlmann, line 441: question: Is it ok that only the first span's grpc method will be included and following spans matching this bucket may be tagged with a false method ?

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -430,32 +468,78 @@ impl TraceExporterBuilder {
...
+
+        let otlp_metrics_config = self.otlp_metrics_endpoint.map(|url| OtlpMetricsConfig {
+            endpoint_url: url,
+            headers: build_otlp_header_map(self.otlp_metrics_headers),
+            timeout: otlp_timeout,
+            protocol: OtlpProtocol::HttpJson,
+            otel_trace_semantics_enabled: self.otel_trace_semantics_enabled,
         });
 
+        let runtime_id = uuid::Uuid::new_v4().to_string();
```
VianneyRuhlmann: Aren't we supposed to reuse the runtime_id from the language ? like we do for telemetry

VianneyRuhlmann (approved):

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -365,6 +370,11 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
             self.refresh_v1_active(&agent_info);
         }
 
+        // OTLP trace metrics run the concentrator independently; skip stats enable/disable.
+        if self.otlp_stats_enabled {
+            return;
+        }
+
         self.trace_filterer.store(Arc::new(TraceFilterer::new(
```
VianneyRuhlmann, line 378: Should we still apply trace filters since spans filtered out are not supposed to be included in stats ?

ichinaski (approved):
Looks good from the Data Pipeline point of view

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -265,6 +284,7 @@ impl<'a> BorrowedAggregationKey<'a> {
                 service_source,
                 http_status_code: status_code,
                 grpc_status_code,
+                grpc_method,
```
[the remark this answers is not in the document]
mabdinur: Thanks for the flag, but `grpc_method` is intentionally on the *value* side (`GroupedStats`), not in `FixedAggregationKey`. It is excluded from `Hash`/`PartialEq`, so gRPC spans with different methods correctly merge into one agent stats bucket — matching pre-existing behaviour. The method is only surfaced on `OtlpExactGroup` for OTLP data points. No regression.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -352,6 +357,11 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
     /// Reconcile in-process stats state with the latest agent info.
     /// Async so the `Enabled` arm can await a stats-worker shutdown without `block_on`.
     async fn check_agent_info(&self) {
+        // OTLP trace metrics run the concentrator independently of the agent; never let agent
+        // info enable or disable stats in that mode.
+        if self.otlp_stats_enabled {
+            return;
```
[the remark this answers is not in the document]
mabdinur: The early return only skips stats enable/disable — `self.refresh_v1_active(&agent_info)` is called unconditionally before the `if self.otlp_stats_enabled { return; }` guard, so V1 protocol negotiation proceeds normally.

libdd-data-pipeline/src/otlp/exporter.rs
mabdinur: Fair nit — happy to rename back to `retries` for consistency. The rename to `attempts` was to make the semantics clearer (`OTLP_SHUTDOWN_MAX_ATTEMPTS = 1` means one try, no retries), but `retries` is fine too if that's the established convention.

libdd-trace-stats/src/span_concentrator/aggregation.rs
mabdinur: Yes, that's a known tradeoff. When multiple gRPC spans with different `rpc.method` values land in the same bucket (same service/resource/status), the first span's method is stored and subsequent spans do not update it. For agent `/v0.6/stats` this is irrelevant (the method is not serialized in `ClientGroupedStats`). For the OTLP data point it means the method attribute reflects the first span seen in that bucket — agreed it's imprecise. We could store `"multi"` or leave it blank when methods diverge, but for now this matches the pre-existing behaviour where method is not part of the aggregation key.

libdd-data-pipeline/src/trace_exporter/builder.rs
mabdinur: Good catch — you're right, this should reuse `self.metadata.runtime_id` so stats and traces can be correlated by the backend. Will fix.

libdd-data-pipeline/src/trace_exporter/mod.rs
mabdinur: The early return in `check_agent_info` only skips *updating* the filterer from agent info — the `trace_filterer` itself is still in place and applied when spans are processed. Stats are fed from the concentrator which operates on spans after sampling decisions are applied; tag-based filtering via `trace_filterer` runs at a different layer. That said, if you're asking whether spans excluded by the agent's require/reject tag filters should also be excluded from OTLP stats, that's a fair point — currently those filters only affect what's sent to the agent, not the concentrator. Happy to align this if that's the expected behaviour.

PR #2069, opened 2026-06-02
refactor(vecmap)!: avoid Clone bound from dedup

yannham:
# What does this PR do?

Re-implement `VecMap::dedup` to avoid the `Clone` bound on keys.

# Motivation

It turns out we don't want to add a `Clone` bounds on `SpanText`, which is the type of key we'll use in practice. See discussion thread following https://github.com/DataDog/libdatadog/pull/2043#discussion_r3318168887.

# Additional Notes

N/A

# How to test the change?

Existing tests.

### DIFF
--- a/libdd-trace-utils/src/span/vec_map.rs
+++ b/libdd-trace-utils/src/span/vec_map.rs
@@ -214,9 +214,7 @@ impl<K: Eq + Hash, V> VecMap<K, V> {
 
         self.as_deduped_map()
     }
-}
 
-impl<K: Hash + Eq + Clone, V> VecMap<K, V> {
     /// Remove entries with a duplicate key, only keeping the last one. After this, a flag is set
     /// internally, such that as long as the map isn't extended or mutably iterated, the next
     /// [Self::dedup] doesn't perform the work again.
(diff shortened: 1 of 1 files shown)


libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -168,7 +168,7 @@ impl<K, V> VecMap<K, V> {
     }
 }
 
-impl<K: Hash + Eq + Clone, V> VecMap<K, V> {
+impl<K: Hash + Eq, V> VecMap<K, V> {
```
ekump: minor: Would it be difficult to add a regression test with a non-clone key? I suspect nothing would break if someone added back `clone` to the impl?

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -177,12 +177,42 @@ impl<K: Hash + Eq + Clone, V> VecMap<K, V> {
...
         // Since we're going to shuffle elements around, it's not easy to keep references to keys in
-        // the deduping set. The simplest is to clone them.
-        let mut seen = HashSet::with_capacity(self.len());
+        // the deduping set while deleting some of them, since deletion in a vec shifts all other
+        // elements after it, invalidating references. When we finally call `retain`, we must not
+        // hold any reference to vecmap elements anymore.
+        //
+        // The following approaches are possible:
+        //
+        // - clone the keys in the hashset. Alas , we don't want a `Clone` bound on `SpanText`
```
ekump:
```suggestion
        // - clone the keys in the hashset. Alas, we don't want a `Clone` bound on `SpanText`
```

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -177,12 +177,42 @@ impl<K: Hash + Eq + Clone, V> VecMap<K, V> {
...
+        //   (which is the type of keys in practice), as some representations can be expensive to
+        //   clone, e.g. requiring to lock the GIL in Python (python-native, reference counted
+        //   strings)
+        // - a two-pass approach. In a first pass we store seen key references in a HashSet and
+        //   build a bitmap of indices to keep. Once built, we can release the set and call
+        //   `Vec::retain` without borrowing issues. It's safe but requires an additional pass over
+        //   the vecmap and an auxiliary `Vec<bool>`, in addition to the hashset.
+        // - an unsafe, one-pass approach: if we re-implement a custom `retain`, we can store key
+        //   references in an auxiliary HashSet that are guaranteed remain valid as we move
```
ekump:
```suggestion
        //   references in an auxiliary HashSet that are guaranteed to remain valid as we move
```

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -177,12 +177,42 @@ impl<K: Hash + Eq + Clone, V> VecMap<K, V> {
...
+        //   elements: we first move an element to keep at his final location, and only then inserts
```
ekump:
```suggestion
        //   elements: we first move an element to keep at their final location, and only then insert
```

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -177,12 +177,42 @@ impl<K: Hash + Eq + Clone, V> VecMap<K, V> {
...
+            let mut seen = HashSet::with_capacity(self.len());
+            self.data.iter().map(|(k, _)| seen.insert(k)).collect()
+        };
+
+        let mut i = 0;
+        self.data.retain(|_| {
+            let kept = keep[i];
+            i += 1;
+            kept
+        });
```
ekump: minor: does this read better? Any downsides?

```suggestion
        let mut keep = keep.into_iter();                                                                                                                                                                               
        self.data.retain(|_| keep.next().unwrap_or(false)); 
```

ekump (approved):
just a few minor comments. LGTM.

libdd-trace-utils/src/span/vec_map.rs
yannham: It's nicer :+1:

libdd-trace-utils/src/span/vec_map.rs
yannham: Good idea. Should be straightforward.

PR #2070, opened 2026-06-02
fix: memory ordering issue in otel thread ctx

yannham:
# What does this PR do?

Fix a memory ordering issue in the OTel thread context publication.

# Motivation

External review of this code (upstreamed to Polar Signals) spotted this error. As originally written, it's not guaranteed that a reader of the TLS slot see a fully initialized context, because the `ThreadContext::new` method is called _after_ the fence. We want that context initialization happens-before storing the pointer in the TLS slot from the reader's perspective, which requires to move context initialization before the fence.

# Additional Notes

N/A

# How to test the change?

N/A

### DIFF
--- a/libdd-otel-thread-ctx/src/lib.rs
+++ b/libdd-otel-thread-ctx/src/lib.rs
@@ -429,15 +429,13 @@ pub mod linux {
                     compiler_fence(Ordering::SeqCst);
                     current.valid.store(1, Ordering::Relaxed);
                 } else {
+                    let ctxt = ThreadContext::new(trace_id, span_id, local_root_span_id, attrs)
+                        .into_ptr()
+                        .as_ptr();
                     // No need for `AcqRel`, see [^tls-slot-ordering].
                     compiler_fence(Ordering::Release);
                     // `ThreadContext::new` already initialises `valid = 1`.
-                    let _ = Self::swap(
-                        slot,
-                        ThreadContext::new(trace_id, span_id, local_root_span_id, attrs)
-                            .into_ptr()
-                            .as_ptr(),
-                    );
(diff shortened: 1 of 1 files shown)


paullegranddc (approved):

libdd-otel-thread-ctx/src/lib.rs
```diff
@@ -429,15 +429,13 @@ pub mod linux {
...
                     // No need for `AcqRel`, see [^tls-slot-ordering].
                     compiler_fence(Ordering::Release);
                     // `ThreadContext::new` already initialises `valid = 1`.
-                    let _ = Self::swap(
-                        slot,
-                        ThreadContext::new(trace_id, span_id, local_root_span_id, attrs)
-                            .into_ptr()
-                            .as_ptr(),
-                    );
+                    let _ = Self::swap(slot, ctxt);
```
ivoanjo, line 438: If the order on the swap was seq_cst instead of relaxed I think this would have been fine :P

yannham: You're right :slightly_smiling_face: but this is more costly. A `SeqCst` swap would be compiled down to an actual atomic swap/store/read instruction (as in a concurrent program). Using `Relaxed` + compiler fence ensures the compiler doesn't do anything stupid, but generates standard non-atomic operations for the hardware to execute.

Though I do agree, I personally reason better over the ordering of specific atomic operations than combinations with fences...

ivoanjo: On a more serious note, nice catch ;D

ivoanjo (approved):
👍 LGTM

PR #2072, opened 2026-06-03
fix(ci): always run cargo-public-api in semver-level.sh

iunanua:
# What does this PR do?

Modify `semver-level.sh` to check the output of `cargo-public-api`
Upgrade `cargo-public-api` to `0.52.0`
Upgrade `cargo-semver-checks` to `0.48.0`

# Motivation

`cargo-semver-checks` misses parameter-type changes on non-generic functions (no function_parameter_type_changed lint), returning minor when the change is actually breaking. Run `cargo-public-api` unconditionally and take the higher of the two signals via max_level so removed/changed signatures are caught even when `cargo-semver-checks` already flagged the crate at minor.


scripts/semver-level.sh
```diff
@@ -107,54 +141,112 @@ compute_semver_results() {
...
+    # we run it unconditionally and combine with semver-checks via max_level.
+    # Skip only when there is no baseline (new crate) or when semver-checks
+    # already flagged major (cannot go higher).
+    #
+    # Requires cargo-public-api >= 0.52.0: earlier versions include function
+    # parameter names in signatures, so a non-breaking parameter *rename* diffs
+    # as Removed + Added and would be falsely promoted to major. From 0.52.0
+    # parameter names are omitted by default, so only signature-meaningful
+    # changes (e.g. parameter type changes) surface here.
+    # ----------------------------------------------------------------
```
hoolioh, line 164: The comment describes a different output shape than the one the code handles, which could be misleading. Could we reword it to say the signature change surfaces under Changed items and that's why we normalize + compare old vs new there?

hoolioh (approved):

scripts/semver-level.sh
iunanua: done!

PR #2074, opened 2026-06-03
feat(data-pipeline)!: add stdout log trace exporter

bengl:
# What does this PR do?

Adds a stdout "log exporter" transport to libdatadog's trace pipeline. When enabled, `TraceExporter` writes traces as newline-delimited JSON (`{"traces":[[...]]}`) to stdout in the format consumed by the Datadog Forwarder, instead of sending them to an agent over HTTP.

New pieces:
- **`libdd-trace-utils::json_log_encoder`** — Forwarder JSON encoder: lowercase hex IDs (incl. 128-bit), integer `error`, ns `start`/`duration`, empty-field omission, greedy 256 KiB line batching, oversize-span drop.
- **`TraceExporterBuilder::set_output_to_log(max_line_size)`** — selects a stdout destination that bypasses agent-info polling, client-side stats, V1 negotiation, and telemetry. Exposed over FFI as `ddog_trace_exporter_config_set_output_to_log`.
- **`libdd_capabilities::LogWriterCapability`** — the write goes through a capability so the transport also works on wasm (where the host, e.g. JavaScript, supplies the sink). The native capability writes to stdout.

# Motivation

Serverless runtimes (primarily AWS Lambda) often have no reachable Datadog agent or Lambda extension. The established fallback is to write traces to stdout, where the Datadog Forwarder tails the logs and submits them to the trace intake. Today each tracer (dd-trace-js/py/go/java) reimplements this independently; this lands a single reusable implementation in libdatadog that native-backed tracers can adopt.

# Additional Notes

- **Breaking (Rust API):** `TraceExporter<C>` (and `TraceExporterBuilder::build`/`build_async`, `trace_buffer::DefaultExport<C>`) now require `C: LogWriterCapability`. Callers using a custom capability bundle must implement `libdd_capabilities::LogWriterCapability`; `NativeCapabilities` already implements it, so callers using `NativeCapabilities` are unaffected. PR title carries the `!` marker accordingly.
- Selection policy stays with the caller — there is no env-reading helper in libdatadog (an earlier `recommended_log_output()` was removed per review; reading env from libdatadog has caused production crashes). The SDK detects Lambda/agent/extension itself and calls `set_output_to_log`.
- `meta_struct` is intentionally omitted (raw msgpack the log intake can't interpret), matching the reference exporters.
- Writes are **synchronous/blocking** — intended for single-threaded serverless runtimes where there is no shared async reactor to stall. Log output takes precedence over an OTLP endpoint if both are configured.
- No new third-party dependencies (`serde_json` already in tree) → no `Cargo.lock` / `LICENSE-3rdparty.csv` churn.

# How to test the change?

```bash
cargo nextest run -p libdd-trace-utils -p libdd-data-pipeline -p libdd-data-pipeline-ffi -E '!test(tracing_integration_tests::)'
cargo +stable clippy -p libdd-capabilities -p libdd-capabilities-impl -p libdd-trace-utils -p libdd-data-pipeline -p libdd-data-pipeline-ffi --all-targets --all-features -- -D warnings
cargo ffi-test   # FFI crate touched
```

Key tests:
- `json_log_encoder` — golden bytes, Forwarder `is_trace` contract, hex (incl. 128-bit), size-cap batching, oversize-drop, non-finite-metric → null, multi-trace flatten, span_links/span_events present-case.
- `trace_exporter::tests::test_log_mode_send_writes_forwarder_json` — full `send(msgpack)` → decode → log branch → capability bytes (via a capturing test capability).
- `trace_exporter::tests::test_log_mode_makes_no_agent_requests` — zero agent calls + `info_fetcher` not spawned in log mode.
- `log_writer` helper tests; FFI `config_output_to_log_test`.

### DIFF
--- a/libdd-data-pipeline/src/trace_exporter/builder.rs
+++ b/libdd-data-pipeline/src/trace_exporter/builder.rs
@@ -541,6 +541,8 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
         let info_endpoint = Endpoint::from_url(add_path(&agent_url, INFO_ENDPOINT));
         let (info_fetcher, info_response_observer) =
             AgentInfoFetcher::<C>::new(info_endpoint, Duration::from_secs(5 * 60));
+        // TODO(APMSP-3609): consolidate per-mode worker gating (info-fetcher, telemetry,
+        // stats concentrator) off the selected export destination in one place.
         // In log-export mode there is no agent to poll; skip spawning the worker
         // entirely so we don't make repeated failing `/info` calls (e.g. in Lambda).
         let info_fetcher_handle = if self.output_to_log {
--- a/libdd-data-pipeline/src/trace_exporter/builder.rs
+++ b/libdd-data-pipeline/src/trace_exporter/builder.rs
@@ -8,6 +8,7 @@ use crate::otlp::{OtlpMetricsConfig, OtlpResourceInfo, OtlpTraceConfig};
 use crate::telemetry::TelemetryClientBuilder;
 use crate::trace_exporter::agent_response::AgentResponsePayloadVersion;
 use crate::trace_exporter::error::BuilderErrorKind;
+use crate::trace_exporter::log_writer::DEFAULT_LOG_MAX_LINE_SIZE;
 #[cfg(all(not(target_arch = "wasm32"), feature = "telemetry"))]
 use crate::trace_exporter::TelemetryConfig;
 #[cfg(not(target_arch = "wasm32"))]
@@ -18,7 +19,7 @@ use crate::trace_exporter::{
     TracerMetadata, INFO_ENDPOINT,
 };
 use arc_swap::ArcSwap;
-use libdd_capabilities::{HttpClientCapability, MaybeSend, SleepCapability};
+use libdd_capabilities::{HttpClientCapability, LogWriterCapability, MaybeSend, SleepCapability};
 use libdd_common::{parse_uri, tag, Endpoint};
 use libdd_dogstatsd_client::new;
 use libdd_shared_runtime::SharedRuntime;
@@ -90,6 +91,11 @@ pub struct TraceExporterBuilder<R: SharedRuntime> {
     otlp_metrics_headers: Vec<(String, String)>,
     otel_trace_semantics_enabled: bool,
     runtime_id: Option<String>,
+    /// When true, traces are written as newline-delimited JSON to stdout (the
+    /// Datadog Forwarder "log exporter" path) instead of being sent to an agent.
+    output_to_log: bool,
+    /// Optional override for the maximum size of a single emitted log line.
+    log_max_line_size: Option<usize>,
 }
 
 /// Default is impl'd for `R = ForkSafeRuntime` only so that bare
@@ -149,6 +155,8 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
             otlp_metrics_headers: Vec::new(),
             otel_trace_semantics_enabled: false,
             runtime_id: None,
+            output_to_log: false,
+            log_max_line_size: None,
         }
     }
 }
@@ -425,6 +433,31 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
         self.runtime_id = Some(id.to_owned());
         self
     }
+    /// Configure the exporter to write traces as newline-delimited JSON to stdout
+    /// (the Datadog Forwarder "log exporter" path) instead of sending them to a
+    /// Datadog agent. This is the transport used in serverless environments (e.g.
+    /// AWS Lambda) when no agent is reachable.
+    ///
+    /// `max_line_size` overrides the per-line byte cap; `None` (or `Some(0)`,
+    /// which is coerced to the default) uses the default of 256 KiB, the AWS
+    /// CloudWatch Logs per-event limit. When this is set, agent-specific behavior
+    /// (agent-info polling, client-side stats, V1 negotiation) is bypassed.
+    ///
+    /// In this mode each span's `meta` is serialized to process stdout (and thus
+    /// captured by CloudWatch Logs in Lambda); `meta_struct` is excluded because
+    /// it holds raw msgpack the log intake cannot interpret. Writes are
+    /// synchronous/blocking, so this mode targets single-threaded serverless
+    /// runtimes where blocking stdout writes do not stall a shared async reactor.
+    ///
+    /// Takes precedence over an OTLP endpoint: if both this and `set_otlp_endpoint`
+    /// are configured, traces are written to the log output and not sent via OTLP.
+    pub fn set_output_to_log(&mut self, max_line_size: Option<usize>) -> &mut Self {
+        self.output_to_log = true;
+        // Treat `Some(0)` as "use the default" (a 0 cap would drop every span);
+        // keeps parity with the FFI setter's 0-sentinel.
+        self.log_max_line_size = max_line_size.filter(|&n| n != 0);
+        self
+    }
 
     /// Build the [`TraceExporter`] synchronously.
     ///
@@ -432,7 +465,9 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
     /// Requires `R: BlockingRuntime` so the builder can drive setup on the runtime. Not
     /// available on wasm — use [`Self::build_async`] there.
     #[cfg(not(target_arch = "wasm32"))]
-    pub fn build<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static>(
+    pub fn build<
+        C: HttpClientCapability + SleepCapability + LogWriterCapability + MaybeSend + Sync + 'static,
+    >(
         mut self,
     ) -> Result<TraceExporter<C, R>, TraceExporterError>
     where
@@ -459,7 +494,7 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
     /// context. If [`set_shared_runtime`](Self::set_shared_runtime) was not called, a new
     /// runtime is constructed via [`SharedRuntime::new`].
     pub async fn build_async<
-        C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static,
+        C: HttpClientCapability + SleepCapability + LogWriterCapability + MaybeSend + Sync + 'static,
     >(
         self,
     ) -> Result<TraceExporter<C, R>, TraceExporterError> {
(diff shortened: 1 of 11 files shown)


libdd-data-pipeline/src/serverless.rs
```diff
@@ -0,0 +1,202 @@
...
+fn is_aws_lambda() -> bool {
+    matches!(read_cloud_env(), Some((_, EnvironmentType::LambdaFunction)))
+}
+
+/// Returns true if any of the explicit agent endpoint environment variables is
+/// set to a non-empty value.
+fn has_explicit_agent_endpoint() -> bool {
+    AGENT_ENDPOINT_ENV_VARS
+        .iter()
+        .any(|var| matches!(env::var(var), Ok(value) if !value.is_empty()))
```
ekump: getting / setting env vars in libdatadog should be avoided, it's led to production crashes in the past. (See [here](https://github.com/DataDog/libdatadog/pull/1810/changes#diff-d69dea9df656b56ff3aba3dfbb752eed7dd26dac35af93840d084ab5313f48c7R351) and [here](https://github.com/DataDog/libdatadog/pull/1930)). It's best if the SDKs just pass us the values via the builder.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -478,8 +500,11 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
     ///
     /// Sync facade over [`Self::send_trace_chunks_async`]; panics inside an existing
     /// tokio context. Returns the agent response (or `Unchanged` for OTLP).
+    ///
+    /// The `T: serde::Serialize` bound exists to support the log-export JSON path,
+    /// which serializes spans directly to stdout.
     #[cfg(not(target_arch = "wasm32"))]
-    pub fn send_trace_chunks<T: TraceData>(
+    pub fn send_trace_chunks<T: TraceData + serde::Serialize>(
```
ekump: @iunanua (asking you since you've been working on the semver checks recently) - I'm curious why semver checks aren't failing here? Shouldn't this be considered a breaking change?

libdd-data-pipeline/src/trace_exporter/log_writer.rs
```diff
@@ -0,0 +1,130 @@
...
+            // into a single underlying write+flush per send, amortizing the global
+            // stdout lock acquisition and LineWriter newline scan across all lines.
+            out: Mutex::new(Box::new(BufWriter::new(std::io::stdout()))),
+            max_line_size,
+        }
+    }
+
+    /// Create a writer targeting a custom sink (used in tests).
+    #[cfg(test)]
+    pub(crate) fn with_sink(out: Box<dyn Write + Send>, max_line_size: usize) -> Self {
```
ekump: This is currently gated to `#[cfg(test)]`. Isn't this going to be needed as part of the public API in order for this to work with wasm? I don't think libdatadog can write to stdout when wasm is the target? 

And speaking of wasm, should we be leveraging the capabilities crate for the log writer?

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -365,19 +394,26 @@ impl TraceExporterBuilder {
...
+                        ))
+                    })?,
+            )
+        };
         // The handle is currently only tracked for shutdown on native; on wasm
         // it is dropped here (the worker keeps running on the JS event loop
         // until the page/module is torn down).
         #[cfg(target_arch = "wasm32")]
-        let _ = info_fetcher_handle;
+        let _ = &info_fetcher_handle;
```
yannham: I feel this is useless, because instead of dropping the `info_fetcher_handle`, we drop a reference to it, which doesn't do much. Was this change intentional?

libdd-data-pipeline/src/trace_exporter/log_writer.rs
```diff
@@ -0,0 +1,130 @@
...
+
+/// Writes trace chunks as newline-delimited JSON to an output sink (stdout by
+/// default), in the format understood by the Datadog Forwarder.
+///
+/// Writes are synchronous/blocking: [`write_traces`](Self::write_traces) holds a
+/// mutex and blocks the calling thread on the underlying sink (stdout) write and
+/// flush. On a multi-threaded async runtime this can stall the reactor, so log
+/// mode is intended for serverless / current-thread runtimes (e.g. AWS Lambda).
+pub(crate) struct LogTraceWriter {
+    out: Mutex<Box<dyn Write + Send>>,
```
yannham: I think `stdout()` is already using a mutex internally. Since it's `Send + Sync` and `&Stdout: Write`, I believe you can just use [Stdout](https://doc.rust-lang.org/std/io/struct.Stdout.html) here without an outer mutex, if we always send to `Stdout`.

libdd-data-pipeline/src/trace_exporter/log_writer.rs
yannham: Ah, you need a more generic form for tests... we could define `enum Sink { Stdout(Stdout), Other(Mutex<Box<dyn Write + Send>>)`. Not sure it's worth it, if the mutex is never contended and just here to please the compiler, though.

libdd-data-pipeline/src/trace_exporter/mod.rs
yannham: It is clearly a breaking change, if the parent modules of this one are public

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -571,10 +602,29 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
+            let stats = writer
+                .write_traces(&traces)
+                .map_err(TraceExporterError::Io)?;
+            debug!(
+                spans_written = stats.spans_written,
+                spans_dropped = stats.spans_dropped,
+                "Wrote traces to log exporter"
+            );
+            return Ok(AgentResponse::Unchanged);
+        }
```
yannham, line 708: nitpick: since this is a totally different codepath that returns early, might be clearer to move that logic into its own synchronous helper (rustc is pretty good at inlining local functions, so this should be free).

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -1129,6 +1179,93 @@ mod tests {
         builder.build::<NativeCapabilities>().unwrap()
     }
 
+    /// Shared in-memory sink for asserting bytes written in log-export mode.
+    #[derive(Clone, Default)]
+    struct SharedSink(std::sync::Arc<std::sync::Mutex<Vec<u8>>>);
```
yannham: I think `Arc` is imported already? At least in the parent module?

```suggestion
    struct SharedSink(Arc<std::sync::Mutex<Vec<u8>>>);
```

libdd-data-pipeline/src/serverless.rs
```diff
@@ -0,0 +1,202 @@
...
+///    `DD_TRACE_AGENT_URL`, `DATADOG_TRACE_AGENT_HOSTNAME` are set non-empty).
+/// 3. No Datadog Lambda Extension is present on disk.
+/// 4. The environment is not a GCP/Azure serverless environment (naturally excluded because
+///    condition 1 requires AWS Lambda).
+///
+/// The env/extension detection here is intentionally Rust-embedder-only and is
+/// deliberately not exposed over the FFI: host SDKs typically already have their
+/// own serverless/agent detection and may implement their own recommendation.
+pub fn recommended_log_output() -> bool {
+    recommended_log_output_inner(Path::new(DATADOG_EXTENSION_PATH))
```
yannham: Nitpick: should probably inline this function here. It's not like the outer function adds much logic/wrapping.

libdd-trace-utils/src/json_log_encoder/span.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+        // `meta_struct` is intentionally NOT emitted: it holds raw msgpack bytes
+        // that would serialize as a JSON number array the intake cannot interpret.
+        // The reference log exporters (JS/Go/Py/Java) omit it as well.
+        if !span.span_links.is_empty() {
+            len += 1;
+        }
+        if !span.span_events.is_empty() {
+            len += 1;
+        }
```
yannham:
```suggestion
        len += has_type as usize;
        len += (!span.meta.is_empty()) as usize;
        len += (!span.metrics.is_empty()) as usize;
        // `meta_struct` is intentionally NOT emitted: it holds raw msgpack bytes
        // that would serialize as a JSON number array the intake cannot interpret.
        // The reference log exporters (JS/Go/Py/Java) omit it as well.
        len += (!span.span_links.is_empty()) as usize;
        len += (!span.span_events.is_empty()) as usize;
```

yannham (approved):

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -296,12 +302,40 @@ impl TraceExporterBuilder {
...
+    /// CloudWatch Logs per-event limit. When this is set, agent-specific behavior
+    /// (agent-info polling, client-side stats, V1 negotiation) is bypassed.
+    ///
+    /// In this mode each span's `meta` is serialized to process stdout (and thus
+    /// captured by CloudWatch Logs in Lambda); `meta_struct` is excluded because
+    /// it holds raw msgpack the log intake cannot interpret. Writes are
+    /// synchronous/blocking, so this mode targets single-threaded serverless
+    /// runtimes where blocking stdout writes do not stall a shared async reactor.
+    ///
+    /// Takes precedence over an OTLP endpoint: if both this and `set_otlp_endpoint`
```
ekump, line 452: I think this is non-blocking and can be addressed in a separate PR...

I'm not a fan of the exporter containing precedence logic for configuration. I left a similar comment on @paullegranddc's agentless [PR](https://github.com/DataDog/libdatadog/pull/2081#discussion_r3358410631). I believe config logic should live in the SDKs as much as possible, and if the builder receives conflicting config settings, we should return an error. If we want to support precedence to avoid setup friction for customers that should be handled by the SDKs. I don't think there is any harm in merging it as is for now. Let's just capture it in a Jira ticket.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -365,19 +399,26 @@ impl TraceExporterBuilder {
...
-                .spawn_worker(info_fetcher, false)
-                .map_err(|e| {
-                    TraceExporterError::Builder(BuilderErrorKind::InvalidConfiguration(
-                        e.to_string(),
-                    ))
-                })?;
-        // The handle is currently only tracked for shutdown on native; on wasm
-        // it is dropped here (the worker keeps running on the JS event loop
-        // until the page/module is torn down).
+        // In log-export mode there is no agent to poll; skip spawning the worker
```
ekump, line 546: I left a [comment](https://github.com/DataDog/libdatadog/pull/2081#discussion_r3358449185) about agent info and telemetry workers being unnecessarily spawned in @paullegranddc's agentless PR. Beyond the scope of this PR, but after these two PRs land we should discuss refactoring the builder a little bit to handle this a bit more elegantly.

libdd-trace-utils/src/json_log_encoder/mod.rs
```diff
@@ -0,0 +1,497 @@
...
+    #[test]
+    fn empty_input_writes_nothing() {
+        let traces: Vec<Vec<SpanSlice>> = vec![];
+        let mut out = Vec::new();
+        let stats = encode_traces(&traces, &mut out, MAX).unwrap();
+        assert_eq!(stats, EncodeStats::default());
+        assert!(out.is_empty());
+    }
+
+    /// Mirrors the Datadog Forwarder `is_trace` detection contract (spec §6.1):
```
ekump: minor: Since these are technically public docs, if we are going to refer to specs in them by section, the spec should be public too. There is an rfc folder in the libdatadog workspace where you can place the spec as markdown and link to. Alternatively, you could just not mention the spec in the rustdoc comments and not worry about it.

ekump: Oh, hold on, this is a test. There shouldn't be rustdoc comments at all. I'd suggest just changing the `///` to `//`. I don't have any issue with the spec language in the comment if it's just a code comment and not a rustdoc comment.

ekump (approved):
A couple of minor comments, mostly calling out future work for the team. LGTM.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -365,19 +399,26 @@ impl TraceExporterBuilder {
...
+                        ))
+                    })?,
+            )
+        };
+        // The handle is only tracked for shutdown on native; on wasm the `workers`
+        // field is cfg'd out, so drop it here (the worker keeps running on the JS
+        // event loop until the page/module is torn down).
         #[cfg(target_arch = "wasm32")]
-        let _ = info_fetcher_handle;
+        drop(info_fetcher_handle);
```
yannham, line 565: Claude can't refrain from rewriting unrelated stuff with bad taste, can't he 😛 (I'm mostly joking, I don't care what syntax people use to drop value, although I do personally prefer the former)

libdd-data-pipeline/src/trace_exporter/mod.rs
yannham: >nitpick: since this is a totally different codepath that returns early, might be clearer to move that logic into its own synchronous helper (rustc is pretty good at inlining local functions, so this should be free).

Up this nitpick (although not blocking, I'll let you decide);

yannham (approved):
I like the log capability, even is slightly more verbose, since we get mocking cleanly. Beside one previous nitpick that still applies, LGTM;

libdd-data-pipeline/src/trace_exporter/mod.rs
ekump: 👍 Applied your suggestion

PR #2075, opened 2026-06-03
feat(live-debugger): agentless intake forwarding

tylfin:
## What

Adds a transparent path for tracers to send Live Debugger and Symbol Database (SymDB) data **directly to the Datadog intake** when no agent is present, matching the behavior of the agent's debugger proxy routes (`/debugger/v1/diagnostics`, `/debugger/v2/input`, `/symdb/v1/input`). This extends the existing `datadog-live-debugger` crate and its FFI rather than adding a new crate.

## Why

In agentless mode the tracer has no local agent to enrich and forward debugger/SymDB payloads to intake. This gives `libdatadog` the agent's forwarding behavior so tracers (e.g. dd-trace) can ship the same payloads transparently, choosing agent vs. agentless based on whether an API key is present.

## Changes

**Core (`datadog-live-debugger/src/sender.rs`)**
- `debugger_intake_endpoint(site, api_key)` builds an agentless endpoint targeting `debugger-intake.{site}` (mirrors profiling's `agentless()`).
- Factored the per-track path derivation into `derive_endpoint_path()`; agentless requests use `/api/v2/debugger`, agent requests use the proxy paths. `file://` endpoints (tests) are left untouched.
- `Config` gained `symdb_endpoint` plus dual-ship endpoint vectors, with `set_symdb_endpoint`, `add_additional_debugger_endpoint`, and `add_additional_symdb_endpoint`.
- `send()` and `send_symdb()` send the primary endpoint first, then fan out best-effort to any additional endpoints (mirroring the agent's `*_additional_endpoints`, where non-primary responses are discarded). Sending the primary first keeps a slow or stalled additional endpoint from delaying the intake the caller depends on.
- `send_symdb()` forwards raw bytes verbatim with `DD-EVP-ORIGIN: agent-symdb` and tags carried in the `X-Datadog-Additional-Tags` header (matching `symdb.go`).
- `PayloadSender::new` split into `new` / `new_to_endpoint`; existing callers (e.g. `datadog-sidecar`) are unaffected.

**FFI (`datadog-live-debugger-ffi/src/sender.rs`)**
- `ddog_live_debugger_endpoint_from_site_and_api_key` for the agentless endpoint (freed via the existing `ddog_endpoint_drop`).
- A config-builder surface (`..._sender_config_new` / `..._set_endpoint` / `..._set_symdb_endpoint` / `..._add_additional_endpoint` / `..._add_additional_symdb_endpoint` / `..._drop`) and `ddog_live_debugger_spawn_sender_with_config`. The existing `ddog_live_debugger_spawn_sender` is unchanged.
- `ddog_live_debugger_send_symdb_data` to enqueue a raw SymDB body (zero-copy across the boundary via `OwnedCharSlice`).
- The sender routine logs the track name via a static `&str` instead of formatting one per send, dropping an allocation on the hot path.

`DebuggerType` is intentionally left unchanged (SymDB is handled as a separate path) because `datadog-sidecar` does an exhaustive match on it.

## Testing

- `cargo check` on both crates and `datadog-sidecar`
- `cargo +nightly-2026-02-08 fmt --check`, `cargo clippy --all-targets --all-features -D warnings`
- `cargo test -p datadog-live-debugger -p datadog-live-debugger-ffi` (14 pass, 4 new)
- cbindgen header verified; `cargo ffi-test` (only pre-existing `crashtracking`/`ffe` env failures, unrelated to this change)
- No new dependencies; `Cargo.lock` untouched.

### DIFF
--- a/datadog-live-debugger-ffi/src/sender.rs
+++ b/datadog-live-debugger-ffi/src/sender.rs
@@ -4,7 +4,9 @@
 use crate::send_data::serialize_debugger_payload;
 use datadog_live_debugger::debugger_defs::DebuggerPayload;
 use datadog_live_debugger::sender;
-use datadog_live_debugger::sender::{generate_tags, Config, DebuggerType};
+use datadog_live_debugger::sender::{
+    debugger_intake_endpoint, generate_tags, Config, DebuggerType,
+};
 use libdd_common::tag::Tag;
 use libdd_common::Endpoint;
 use libdd_common_ffi::slice::AsBytes;
@@ -42,6 +44,10 @@ impl Drop for OwnedCharSlice {
 enum SendData {
     Raw(Vec<u8>, DebuggerType),
     Wrapped(OwnedCharSlice, DebuggerType),
+    SymDB {
+        data: OwnedCharSlice,
+        content_type: String,
+    },
 }
 
 async fn sender_routine(config: Arc<Config>, tags: String, mut receiver: mpsc::Receiver<SendData>) {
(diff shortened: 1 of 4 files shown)


datadog-live-debugger-ffi/src/sender.rs
```diff
@@ -56,18 +62,37 @@ async fn sender_routine(config: Arc<Config>, tags: String, mut receiver: mpsc::R
         let config = config.clone();
         let tags = tags.clone();
         tracker.spawn(async move {
-            let (debugger_type, data) = match data {
-                SendData::Raw(ref vec, r#type) => (r#type, vec.as_slice()),
-                SendData::Wrapped(ref wrapped, r#type) => (r#type, wrapped.slice.as_bytes()),
+            let (kind, len, result) = match data {
+                SendData::Raw(ref vec, r#type) => (
+                    format!("{type:?}"),
```
bwoebi: This unconditionally allocs a string for every send. Not _bad_, but unnecessary :-/

datadog-live-debugger/src/sender.rs
```diff
@@ -287,12 +418,96 @@ pub async fn send(
...
 ) -> anyhow::Result<()> {
-    let mut batch = PayloadSender::new(config, debugger_type, percent_encoded_tags)?;
+    let endpoints = config.debugger_endpoints_for(debugger_type);
+    let (primary, additional) = endpoints
+        .split_first()
+        .ok_or_else(|| anyhow::anyhow!("no endpoint configured for {debugger_type:?}"))?;
+
+    // Send the primary first so a slow or stalled additional endpoint cannot
+    // delay the intake whose result the caller actually depends on.
+    let result = send_to_endpoint(payload, primary, debugger_type, percent_encoded_tags).await;
```
bwoebi: You should probably send them all in parallel, rather than sequential await, same with send_symdb.

bwoebi (approved):
Aside from the sequential await, looks good to me.

PR #2078, opened 2026-06-04
chore(common)!: replace native-certs with platform-verifier

VianneyRuhlmann:
# What does this PR do?

Replace [rustls-native-certs](https://github.com/rustls/rustls-native-certs) with [rustls-platform-verifier](https://github.com/rustls/rustls-platform-verifier) as recommended in rustls-native-certs.

# Motivation

Potential fix to a crash in dd-trace-py on macos

# Additional Notes

# How to test the change?


ivoanjo:
Can you share a bit more on info on why we're doing this migration and why it is safe? Just to make sure I understand what changes here, as getting this wrong means customers not getting data (e.g. because some TLS connections fail due to missing certificates).

VianneyRuhlmann:
The main reason I'm trying to do it now is because it makes loading the certificate lazy on macos. This helps mitigating a bug in dd-trace-py due to the macos Security.framework api not being thread safe.
Still this is recommended by the `rustls` project to replace native-certs and is already used in [profiling](https://github.com/DataDog/libdatadog/blob/da8cbcb8b81b5b46d8d06da494157d6c74eabf0e/libdd-profiling/src/exporter/tls.rs#L50). It's still using `native-certs` under the hood for linux so behavior should stay the same.
That being said I'm not very familiar with the tls logic of ddcommon, so I'd appreciate any feedback you have.

ivoanjo:
> That being said I'm not very familiar with the tls logic of ddcommon, so I'd appreciate any feedback you have.

I'm not very familiar with it either, which is why I was kinda pointing out that getting this wrong can have a very high cost (ask me how I know.... https://github.com/DataDog/libdatadog/pull/1796 / https://github.com/DataDog/dd-trace-rb/issues/5508 ) so it's worth being very careful with changes here.

paullegranddc (commented):
Can you test that https still works with either fips and normal https on linux?

libdd-common/src/connector/mod.rs
```diff
@@ -126,42 +126,20 @@ mod https {
...
         hyper_rustls::HttpsConnector<hyper_util::client::legacy::connect::HttpConnector>,
     > {
+        use rustls_platform_verifier::BuilderVerifierExt;
+
         ensure_crypto_provider_initialized(); // One-time initialization of a crypto provider if needed
 
-        let certs = load_root_certs()?;
         let client_config = ClientConfig::builder()
-            .with_root_certificates(certs)
+            .with_platform_verifier()?
```
ekump, line 137: I think this will potentially change behavior for macOS and Windows users who use `SSL_CERT_FILE` and `SSL_CERT_DIR`? If a user is using a custom CA here it won't pick them up. This might be ok, since I don't think it's common for Windows to set this (worth checking with someone from .NET), and most macOS usage is for local dev, so the agent would be on localhost anyway and not using https? 

At a minimum, it's worth verifying and documenting.

ekump:
> Can you test that https still works with either fips and normal https on linux?

@paullegranddc @VianneyRuhlmann - I _think_ it should work ok, and also shouldn't pull in `ring` as a dep. But, it's worth testing as Paul suggests. Since it's fips we'll need a separate step that has `--no-default-features --features fips` flags set.

libdd-common/Cargo.toml
```diff
@@ -37,10 +37,7 @@ regex-lite = { version = "0.1", optional = true }
 # backend. We install the ring provider explicitly in connector/mod.rs instead.
 reqwest = { version = "0.13.2", features = ["rustls-no-provider", "hickory-dns"], default-features = false, optional = true }
 criterion = { version = "0.5.1", optional = true }
-# Pinned to <0.8.3: version 0.8.3+ pulls in openssl-probe@0.2 which probes multiple
-# certificate directories and parses individual cert files instead of loading a single
-# bundle, adding unnecessary I/O overhead in latency-sensitive environments.
-rustls-native-certs = { version = ">=0.8.1, <0.8.3", optional = true }
+rustls-platform-verifier = { version = "0.6", optional = true }
```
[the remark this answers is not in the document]
ekump: I think we want to keep this pinned below 0.8.3 to prevent pulling in `openssl-probe`. It is explicitly called out in #1816

libdd-common/src/connector/mod.rs
```diff
@@ -126,42 +126,20 @@ mod https {
...
-            }
-        }
-        // TODO(paullgdfc): log errors even if there are valid certs, instead of ignoring them
-
-        for cert in cert_result.certs {
-            //TODO: log when invalid cert is loaded
-            roots.add(cert).ok();
-        }
-        if roots.is_empty() {
-            return Err(errors::Error::NoValidCertifacteRootsFound.into());
```
ekump, line 161: Can / should we remove this error type? It doesn't look like it's used anywhere anymore.

ekump (approved):
Not an area I'm an expert in, but looks reasonable. Let's just add some tests. Otherwise LGTM.

libdd-common/src/connector/mod.rs
VianneyRuhlmann: I've checked with the dotnet guild and we don't support it (actually dotnet themselves [refused to add support](https://github.com/dotnet/runtime/issues/86243)). Since profiling already switched to platform-verifier for the reqwest client I think it's fine to move on.

libdd-common/Cargo.toml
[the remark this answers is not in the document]
VianneyRuhlmann: I did the change and I'm fine with pinning it. However this doesn't look sustainable to have to pin to a patch release especially if `platform-verifier` end up bumping the version they use.

libdd-common/src/connector/mod.rs
ekump: I agree, this isn't sustainable. I'll create a jira ticket to look into this deeper.

paullegranddc (approved):

PR #2081, opened 2026-06-04
feat(data-pipeline): add agentless export

paullegranddc:
# Motivation

Send traces to the agentless endpoint to replace the JS and python exporters

# What changes

* Additional JSON agentless encoder
* Additional configuration (agentless enablement, endpoint, API key)
* Agentless traces need top level computation, but not dropping spans


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -575,6 +644,18 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
+        if let Some(ref config) = self.agentless_config {
+            // For agentless we want to tag top level spans, but not perform
+            // stats aggregation or span drops
+            if !self.client_computed_top_level {
+                for chunk in traces.iter_mut() {
+                    libdd_trace_utils::span::trace_utils::compute_top_level_span(chunk);
+                }
+            }
+
+            return self.send_agentless_traces_inner(traces, config).await;
```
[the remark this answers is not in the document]
paullegranddc: No, not dropping traces and  aggregating stats is expected

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -429,7 +459,35 @@ impl TraceExporterBuilder {
...
         };
 
-        let otlp_config = self.otlp_endpoint.map(|url| {
+        // Resolve transport selection: OTLP takes precedence over agentless when both are set.
+        let (otlp_endpoint, agentless_endpoint, agentless_api_key) = match (
+            self.otlp_endpoint.as_ref(),
+            self.agentless_endpoint.as_ref(),
+        ) {
+            (Some(_), Some(_)) => {
+                tracing::warn!(
```
ekump: Should the builder fail if it has conflicting configuration? SDK configuration is complex. I'm of the opinion that it's better for the complexity to be concentrated in the SDKs and let them handle things like precedence.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -429,7 +459,35 @@ impl TraceExporterBuilder {
...
+                (self.otlp_endpoint, None, None)
+            }
+            _ => (
+                self.otlp_endpoint,
+                self.agentless_endpoint,
+                self.agentless_api_key,
+            ),
+        };
+
+        let agentless_config = match (agentless_endpoint, agentless_api_key) {
```
ekump, line 683: agentless config is going to happen after `AgentInfoFetcher` is spawned [here](https://github.com/DataDog/libdatadog/pull/2081/changes#diff-bc58d8d129128b72456204f3dbe30c5236e14d95d26337783dff87ef87b8ec02R396). Won't that fail every 5 minutes when it polls an agent that doesn't exist?

libdd-data-pipeline/src/trace_exporter/builder.rs
ekump: On the same topic, the build should probably error out if `set_url` and `output_format` is used and agentless is set at the same time.

libdd-data-pipeline/src/trace_exporter/builder.rs
ekump: I think the same applies to telemetry? It's built against the agent url.

ekump: I think all of this also applies to otlp mode for the exporter.

libdd-data-pipeline/src/agentless/exporter.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::trace_exporter::error::{InternalErrorKind, RequestError, TraceExporterError};
+use http::HeaderMap;
+use libdd_capabilities::{HttpClientCapability, SleepCapability};
+use libdd_common::Endpoint;
+use libdd_trace_utils::send_with_retry::{
+    send_with_retry, RetryBackoffType, RetryStrategy, SendWithRetryError,
+};
+use tracing::error;
+
+const AGENTLESS_MAX_ATTEMPTS: u32 = 3;
```
ekump: After #2047 this should be renamed `AGENTLESS_MAX_RETRIES`

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    if !span.meta_struct.is_empty() {
+        map.serialize_entry(
+            "meta_struct",
+            &ser_fn!(<T: TraceData> |ser, span: &'a Span<T>| {
+                let mut ms = ser.serialize_map(Some(span.meta_struct.len()))?;
+                for (k, v) in span.meta_struct.iter() {
+                    let key: &str = k.borrow();
+                    let bytes: &[u8] = v.borrow();
+                    // Encode as a JSON array of u8 (default serde behavior for &[u8]).
```
ekump: Is this the right thing to do? If i'm reading [js correctly](https://github.com/DataDog/dd-trace-js/blob/master/packages/dd-trace/src/encode/agentless-json.js#L73) (a big if) they just send `meta_struct` values as json, not msgpack encoded bytes.

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+            let mut meta = ser.serialize_map(None)?;
+            for (k, v) in span.meta.iter() {
+                let key: &str = k.borrow();
+                if key == "_dd.p.tid" {
+                    p_tid_seen = true;
+                }
+                let val: &str = v.borrow();
+                meta.serialize_entry(key, val)?;
+            }
+            if !p_tid_seen && upper_bits != 0 {
```
ekump, line 232: Is it intentional that `_dd.p.tid` is the only key that's deduped?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -503,6 +550,28 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
         self.send_trace_chunks_inner(trace_chunks).await
     }
 
+    /// Sends trace chunks to the Datadog agentless intake (`/v1/input`) as JSON.
+    async fn send_agentless_traces_inner<T: TraceData>(
```
ekump, line 636: Where is max payload size handled? In the SDKs? I assume the endpoint has a max size that we have to respect?

libdd-data-pipeline/examples/send-traces-agentless.rs
ekump, line 1: I believe llm obs uses the test agent as a stand-in for intake in tests. Could we do the same and add integration tests for agentless payloads?

libdd-data-pipeline/src/trace_exporter/mod.rs
paullegranddc: It has a 5MB size limit I believe. I don't think this limit should be enforced in the trace exporter though.
It should probably be at the trace buffer.

Or maybe we could split incoming list of trace chunks in multiple payloads 🤔

ekump: Ah, that's a good point. The trace buffer should handle flushing below the limit. 

Although, looking at the trace buffer in data pipeline we don't enforce limits or split up chunks. [We check after a chunk is added](https://github.com/DataDog/libdatadog/blob/a76412ccfb445db47f1153c638f616d931ed46f7/libdd-data-pipeline/src/trace_buffer/mod.rs#L475), and flush above `flush_trigger_bytes`. It's theoretically possible that if we receive a significantly large chunk we flush something that exceeds the intake limit. In a similar vain, [we only drop chunks after max_buffered_bytes is exceeded](https://github.com/DataDog/libdatadog/blob/a76412ccfb445db47f1153c638f616d931ed46f7/libdd-data-pipeline/src/trace_buffer/mod.rs#L242), which can also trigger a flush that's beyond intake's limit.

What may make this an issue for agentless is that the max size for the agent is significantly larger than intake's limit. And the agent handles splitting up the payloads for intake's limits. So the trace buffers can usually get away with "trigger a flush once we exceed a limit". 

I'm not sure this is a practical problem today that we need to deal with in this PR? I think we can tackle it separately?

bwoebi:
I have some questions:
- Why is this serializing directly to JSON, rather than converting pb::TraceChunk (v07)? This would also be the proper long-term plan to just support v1 out of the box later.
- We do have quite a bit of machinery in SendData to submit `send_with_protobuf()` - all there; including the existing retry mechanisms etc.
- `Endpoint` has an `api_key` field, which is meant to distinguish the payload it gets sent to.

This PR feels like a lot of re-inventing the wheel.

paullegranddc:
> Why is this serializing directly to JSON, rather than converting pb::TraceChunk (v07)?

Because it is not the same schema.

The agentless endpoint expects something like `{ traces: { runtime_id, <other tracer meta >, spans: [ {span_id: str (hex encoded), <other fields>} ]} }`
Whereas V07 is a `TracerPayload` which is `{ <tracer meta>, chunks: [ { <chunk meta>, spans: [ {span_id: u64 } ]  } ] }`

As to why I do the conversion at encoding, creating an `struct AgentlessPayload` deriving serde and converting v04 span to this payload is way less efficient than doing it since we need to allocate intermediary structs 

> We do have quite a bit of machinery in SendData to submit send_with_protobuf() - all there; including the existing retry mechanisms etc.
Endpoint has an api_key field, which is meant to distinguish the payload it gets sent to.

I mean, `send_agentless_traces_http` is not that long,  we do use the `send_with_retry function` so the logic is encapsulated and has the advantage of being independent from the rest of  SendData, no mixed code path between different encoding and endpoints, easy to remove if we decide to drop agentless submission.

I agree that the api key should be passed in the `Endpoint` and I missed that I could reuse `Endpoint::set_standard_headers` https://github.com/DataDog/libdatadog/blob/main/libdd-common/src/lib.rs#L350

bwoebi:
Makes sense, there is quite some overhead. The v1 protocol is much better at being transformed from what agent expects to what intake expects.
Let's leave this that way then.

I would prefer though if this weren't part of data-pipeline, but SendData as well. (i.e. just moving the code there).
Because the data-pipeline path is not accessible to the sidecar path at least (which we'll probably use too in 1~2 months).

libdd-trace-utils/src/agentless_encoder/mod.rs
paullegranddc: fixed, we now transcode the msp bytes to a json object

libdd-data-pipeline/src/trace_exporter/builder.rs
paullegranddc: Made agentless mode incompatible with OLTP and agent.

OTLP is not incompatible with the agent settings though (and maybe shoudln't be? Users could be using the datadog agent still so we still get telemetry, but decide to export traces in otlp format somewhere else)

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,437 @@
...
+//! accepted by the Datadog HTTP trace intake (`POST /v1/input`).
+//!
+//! ## Differences from the regular agent (msgpack v04) encoding
+//!
+//! - **Wire format**: JSON, wrapped as `{"traces": [ {hostname, env, ..., spans: [...] }, ... ]}`.
+//!   Per-trace metadata (hostname, env, language*, tracerVersion, runtimeID, containerID) is
+//!   inlined on each trace instead of being carried in request headers. Hostname is always emitted
+//! - **IDs**: `trace_id`, `span_id`, `parent_id` are lowercase hex strings (16 chars; 32 for
+//!   span-link trace IDs)
+//! - **128-bit trace IDs**: only the low 64 bits go into `trace_id`; the `_dd.p.tid` meta tag carie
```
VianneyRuhlmann:
```suggestion
//! - **128-bit trace IDs**: only the low 64 bits go into `trace_id`; the `_dd.p.tid` meta tag carries
```

libdd-data-pipeline/examples/send-traces-agentless.rs
```diff
@@ -0,0 +1,118 @@
...
+use rand::random;
+use std::{collections::HashMap, sync::Arc, time::UNIX_EPOCH};
+
+fn get_span(now: i64, trace_id: u128, span_id: u64) -> SpanBytes {
+    let duration = 1_000_000 * span_id as i64;
+    SpanBytes {
+        trace_id,
+        span_id,
+        parent_id: span_id.saturating_sub(1),
+        duration: 1_000_000 * span_id as i64,
```
Aaalibaba42: nit: maybe reuse duration you just created above ?

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,437 @@
...
+    let cut = max_len.saturating_sub(suffix_len);
+    // Find the previous char boundary so we don't slice in the middle of a UTF-8 sequence.
+    let mut end = cut;
+    while end > 0 && !s.is_char_boundary(end) {
+        end -= 1;
+    }
+    let mut truncated = String::with_capacity(end + suffix_len);
+    truncated.push_str(&s[..end]);
+    truncated.push_str(TRUNCATION_SUFFIX);
+    truncated
```
Aaalibaba42: nit:

```suggestion
    s.truncate(end);
    s.push_str(TRUNCATION_SUFFIX);
    s
```

Aaalibaba42 (commented):
Tiny nits

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,437 @@
...
+//!   upper 64 bits
+//! - **Span links / events**: not top-level fields. They are JSON-stringified into
+//!   `meta["_dd.span_links"]` and `meta["events"]`, each truncated to 25_000 chars. No top-level
+//!   `links` field is emitted.
+//! - **Stats / top-level flags**: the intake has no trace-agent to compute them, so the encoder
+//!   injects `meta["_dd.compute_stats"]="1"` on the first span of each chunk and
+//!   `metrics["_trace_root"]=1` where applicable.
+//! - **Non-finite metrics** (NaN/Inf) are dropped (JSON can't represent them).
+//!
+//! Left todo is span normalization (service/name/resource/type truncation + defaults)
```
VianneyRuhlmann: Is this a blocker to release ?

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,437 @@
...
+    Serializer,
+};
+use std::borrow::Borrow;
+
+/// Maximum allowed size of a `meta` value before truncation.
+const MAX_META_VALUE_LEN: usize = 25_000;
+/// Suffix appended when a `meta` value is truncated.
+const TRUNCATION_SUFFIX: &str = "...";
+
+macro_rules! ser_fn {
```
VianneyRuhlmann, line 68: Can you add some doc for the macro ?

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,435 @@
...
+        "resource",
+        if resource_str.is_empty() {
+            name_str
+        } else {
+            resource_str
+        },
+    )?;
+    map.serialize_entry("service", service_str)?;
+    map.serialize_entry("error", &span.error)?;
+    map.serialize_entry("start", &span.start)?;
```
VianneyRuhlmann: Do we need to apply the unix normalization done for v04 msgpack encoding ? (i.e. Replacing negative by 0)

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,460 @@
...
+            let mut p_tid_seen = false;
+            let mut span_links_seen = false;
+            let mut events_seen = false;
+            let mut compute_stats_seen = false;
+
+            let mut meta = ser.serialize_map(None)?;
+            for (k, v) in span.meta.iter() {
+                let key: &str = k.borrow();
+                match key {
+                    "_dd.p.tid" => p_tid_seen = true,
```
VianneyRuhlmann, line 223: We should document that span.span_link/span_events will be dropped if the field already exists in meta

libdd-trace-utils/src/agentless_encoder/mod.rs
paullegranddc: No

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,460 @@
...
+            if !span_links_seen && !span.span_links.is_empty() {
+                if let Some(s) = serialize_span_links(&span.span_links) {
+                    meta.serialize_entry("_dd.span_links", &s)?;
+                }
+            }
+            if !events_seen && !span.span_events.is_empty() {
+                if let Some(s) = serialize_span_events(&span.span_events) {
+                    meta.serialize_entry("events", &s)?;
+                }
+            }
```
VianneyRuhlmann, line 249: We should document that span.span_events/span_links are dropped if the field already exists in meta

libdd-trace-utils/src/agentless_encoder/mod.rs
paullegranddc: No, I added dedup for other attributes that we set

libdd-data-pipeline/src/agentless/config.rs
```diff
@@ -0,0 +1,30 @@
...
+pub const DEFAULT_AGENTLESS_TIMEOUT: Duration = Duration::from_secs(15);
+
+///Agentless trace exporter configuration.
+#[derive(Clone)]
+pub struct AgentlessTraceConfig {
+    /// Full URL to POST traces to (e.g.
+    /// `https://public-trace-http-intake.logs.datadoghq.com/v1/input`).
+    pub endpoint_url: String,
+    /// Datadog API key used for the `dd-api-key` header.
+    pub api_key: String,
```
VianneyRuhlmann, line 17: nit: Any reason to not use Endpoint to handle the api key ?

libdd-data-pipeline/src/agentless/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+//!   or a custom URL) using `dd-api-key` auth, instead of msgpack to the local agent's
+//!   `/v0.4/traces`. The host language resolves the URL from `DD_SITE` and supplies the API key;
+//!   the exporter reads no environment variables.
+//! - **Encoding**: JSON (see [`libdd_trace_utils::agentless_encoder`]) instead of msgpack v04. See
+//!   that module for the payload-shape differences.
+//! - **Retries**: up to 2 retries with exponential backoff starting at 1 s and no cap (the agent
+//!   path uses its own strategy).
+//! - **Mutual exclusion with OTLP**: if both an OTLP and an agentless endpoint are configured on
+//!   the builder, OTLP wins and the agentless config is silently dropped with a warning at build
+//!   time.
```
ekump:
```suggestion
//! - **Mutually exclusive transport**: agentless cannot be combined with OTLP                                                                                                                                                                                                                 
//!   ([`set_otlp_endpoint`]), a caller-supplied agent URL ([`set_url`]), or log output;                                                                                                                                                                                                       
//!   configuring any of them together causes `build`/`build_async` to return                                                                                                                                                                                                                  
//!   `BuilderErrorKind::InvalidConfiguration`.
```

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -1121,35 +1127,6 @@ mod tests {
         }
     }
 
-    #[cfg_attr(miri, ignore)]
```
ekump, line 1124: We still create an `InvalidUrl` error if you pass an empty string. We should probably continue to test that?

libdd-data-pipeline/examples/send-traces-agentless.rs
ekump: bump on this. Fine to be a follow-up PR if necessary.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -401,6 +415,36 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
...
+    /// When set, APM trace spans are sent directly to the Datadog HTTP intake in JSON format
+    /// (`POST /v1/input`) instead of through the Datadog Agent. The host language is responsible
+    /// for resolving the endpoint URL (default
+    /// `https://public-trace-http-intake.logs.{DD_SITE}` or a custom override) and the API key
+    /// from its configuration. This crate does not read environment variables.
+    ///
+    /// Agentless trace export is mutually exclusive with both OTLP trace export
+    /// ([`Self::set_otlp_endpoint`]) and a configured agent URL ([`Self::set_url`]);
+    /// combining either with this method causes [`Self::build`]/[`Self::build_async`]
+    /// to return [`BuilderErrorKind::InvalidConfiguration`].
```
ekump, line 429:
```suggestion
    /// to return [`BuilderErrorKind::InvalidConfiguration`].
   /// the output format is ignored in agentless mode; payloads are always
  /// JSON
```

libdd-trace-utils/src/agentless_encoder/mod.rs
ekump: minor: Would you mind changing this to `// TODO: ...` and adding a Jira ticket to track it?

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,460 @@
...
+        )?;
+    }
+    map.end()
+}
+
+/// Serialize span links to a JSON string suitable for `meta['_dd.span_links']`.
+///
+/// Returns `None` if serialization fails. The result is truncated to
+/// [`MAX_META_VALUE_LEN`] characters with a trailing `"..."` if it would
+/// otherwise exceed that limit.
```
ekump, line 307: It's ok to defer to another PR, but if we truncate it's going to result in malformed JSON and get rejected at intake. I think #980 also introduces the same issue.

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -0,0 +1,460 @@
...
+    if !span.meta_struct.is_empty() {
+        map.serialize_entry(
+            "meta_struct",
+            &ser_fn!(<T: TraceData> |ser, span: &'a Span<T>| {
+                let mut ms = ser.serialize_map(None)?;
+                for (k, v) in span.meta_struct.iter() {
+                    let key: &str = k.borrow();
+                    let bytes: &[u8] = v.borrow();
+
+                    // abort whole payload on malformed entry
```
ekump, line 293: Ok to defer to another PR: Should we really be aborting the entire payload if a single span's meta_struct fails to serialize? I think dd-trace-js will drop just the span.

ekump (approved):
A few potential follow-ups, but nothing blocking. LGTM.

libdd-trace-utils/src/agentless_encoder/mod.rs
paullegranddc: The issue is that there is no easy way to both:
* Not add overhead by going though an intermediary buffer
* Rollback partial invalid writes

In practice since meta struct bytes are generated by tracers from objects when passing spans to libdatadog, this is not an error path that's going to happen anyway

libdd-data-pipeline/examples/send-traces-agentless.rs
paullegranddc: There is currently no support for this in the test agent.
LLMObs spans are a different feature

ekump: Well, that's a bummer

libdd-trace-utils/src/agentless_encoder/mod.rs
bantonsson: I agree with you here, but it will probably come back and bite us...

> this is not an error path that's going to happen anyway

bantonsson (approved):

PR #2082, opened 2026-06-04
fix(remote-config): notification of multi-processing and runtime deduplication

bwoebi:
We want the session_id as runtime_id key, but still deduplicate *processes* according to their runtime_id, otherwise notifcation will not reach all targeted processes.

Additionally submit the remote_config_generation, so that processes which start up and have read the present configuration, get notified when a new configuration has intermittently been read. (race condition)

### DIFF
--- a/datadog-remote-config/src/fetch/multitarget.rs
+++ b/datadog-remote-config/src/fetch/multitarget.rs
@@ -174,6 +174,7 @@ pub trait MultiTargetHandlers<
 
 struct RuntimeInfo<N: NotifyTarget> {
     notify_target: N,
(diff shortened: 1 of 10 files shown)


morrisonlevi (approved):
I spent some time with Codex trying to understand it. I looks okay to me based on my weak understanding. I think Session ID == Runtime ID for runtimes which correctly regenerated the ID on fork, and the Root/Parent Session ID are related to tracking the parent/child relationships. Based on the idea that Session ID == Runtime ID, this looks fine. I will certainly not claim to be an expert here!

PR #2083, opened 2026-06-04
fix(sidecar): include `language:native` tag for sidecar crashtracker

gyuheon0h:
# What does this PR do?

Crashtracker crash logs can include the language tag. If included, [in the backend](https://github.com/DataDog/dd-go/blob/31428f1894713427d921abed736e6a0ce7f2fb94/trace/apps/tracer-telemetry-intake/logs_writer.go#L197), it will be used to set `Log.Ddsource`. It is also used to categorize `instrumentation-telemetry-data` crashes into `instrumentation-telemetry-data-{runtime}`. This PR sets it.

# Motivation
I noticed all crash logs from the sidecar are [missing source](https://app.datadoghq.com/logs?query=is_crash%3Atrue%20source%3Aunknown&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AwAAAZ6UJqMO6s8cewAAABhBWjZVSnFNT0FBQUZvVlVKNFZqYlNBQUEAAAAkZDE5ZTk0MjYtYjBlOS00YmQyLTk3ZTgtZjVjNmMyYmU4YWVkAAADRg&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1778010046228&to_ts=1780602046228&live=true)

# Additional Notes

# How to test the change?

### DIFF
--- a/datadog-sidecar/src/unix.rs
+++ b/datadog-sidecar/src/unix.rs
@@ -286,7 +286,7 @@ fn init_crashtracker(dependency_paths: Option<*const *const libc::c_char>) -> an
         "severity:crash".to_string(),
         format!("library_version:{}", crate::sidecar_version!()),
         "library:sidecar".to_string(),
-        "language:native".to_string(),
+        "language:php".to_string(),
     ];
 
     libdd_crashtracker::init(
--- a/datadog-sidecar/src/unix.rs
+++ b/datadog-sidecar/src/unix.rs
@@ -281,18 +281,13 @@ fn init_crashtracker(dependency_paths: Option<*const *const libc::c_char>) -> an
             config_builder = config_builder.endpoint_test_token(test_token);
         }
     }
-    let mut tags = vec![
+    let tags = vec![
         "is_crash:true".to_string(),
         "severity:crash".to_string(),
         format!("library_version:{}", crate::sidecar_version!()),
         "library:sidecar".to_string(),
         "language:native".to_string(),
     ];
-    if let Ok(env) = std::env::var("DD_ENV") {
-        if !env.is_empty() {
-            tags.push(format!("env:{env}"));
-        }
-    }
 
     libdd_crashtracker::init(
         config_builder.build()?,
--- a/datadog-sidecar/src/unix.rs
+++ b/datadog-sidecar/src/unix.rs
@@ -281,6 +281,19 @@ fn init_crashtracker(dependency_paths: Option<*const *const libc::c_char>) -> an
             config_builder = config_builder.endpoint_test_token(test_token);
         }
     }
+    let mut tags = vec![
+        "is_crash:true".to_string(),
+        "severity:crash".to_string(),
+        format!("library_version:{}", crate::sidecar_version!()),
+        "library:sidecar".to_string(),
+        "language:native".to_string(),
+    ];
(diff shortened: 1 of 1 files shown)


gyuheon0h:
* **#2083** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2083?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2083?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

datadog-sidecar/src/unix.rs
```diff
@@ -281,6 +281,19 @@ fn init_crashtracker(dependency_paths: Option<*const *const libc::c_char>) -> an
...
+    let mut tags = vec![
+        "is_crash:true".to_string(),
+        "severity:crash".to_string(),
+        format!("library_version:{}", crate::sidecar_version!()),
+        "library:sidecar".to_string(),
+        "language:native".to_string(),
+    ];
+    if let Ok(env) = std::env::var("DD_ENV") {
+        if !env.is_empty() {
+            tags.push(format!("env:{env}"));
```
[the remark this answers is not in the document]
gyuheon0h: @bwoebi is this relevant?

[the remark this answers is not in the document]
bwoebi: Yes, that's right.

For starters, DD_ENV is often set via ini (only visible to the PHP process) and not env var. And second, the sidecar _can_ manage multiple envs.

I would just omit it for the sidecar itself.

datadog-sidecar/src/unix.rs
```diff
@@ -281,6 +281,19 @@ fn init_crashtracker(dependency_paths: Option<*const *const libc::c_char>) -> an
             config_builder = config_builder.endpoint_test_token(test_token);
         }
     }
+    let mut tags = vec![
+        "is_crash:true".to_string(),
+        "severity:crash".to_string(),
+        format!("library_version:{}", crate::sidecar_version!()),
+        "library:sidecar".to_string(),
+        "language:native".to_string(),
```
bwoebi: Keep in mind that we're going to see sidecar crashes in two places still. Sometimes the sidecar will run in thread mode, and then be part of the php process, rather being its own process.

I'd probably use `language:php` here (or, in a future where the sidecar has more users, ensure that the sidecar is provided its host language via some config).

gyuheon0h: What if we started off as `native`, then on `set_session_config`, we can update the crashtracker metadata with whatever SessionConfig.language is?

bwoebi: I mean, the point is no fragmentation? It's pretty pointless if a startup issue goes invisible, because no-one ever looks for _that_ service.

gyuheon0h: But by doing so, the sidecar's crash report metadata will always reflect whatever connection it was serving? On startup, the language is set to native, when it serves a php session, it will be set to php?

gyuheon0h: 🤷 I will set to php for now, this can be extended when more users use the sidecar

bwoebi (approved):

PR #2088, opened 2026-06-05
test(profiling): add profiles dictionary benchmarks

taegyunkim:
## What does this PR do?

Adds a focused Criterion benchmark for `ProfilesDictionary` unique string insertion.

The benchmark uses `ProfilesDictionary` and covers 1, 2, 4, and 16 producer threads, so follow-up arena sizing/growth changes have a baseline in the GitLab benchmark job.

#2048 changes the default configs for the ProfilesDictionary

## Why these thread counts?

This benchmark is focused on dictionary string interning. It is not a full end-to-end profiler benchmark.

| Thread count | Why it is useful |
|---|---|
| 1 | Approximates single-writer consumers such as ddprof. ddprof appears to intern from the worker thread while export happens from a separate inactive profile buffer. |
| 2 | Approximates current dd-trace-py contention: one native stack sampler thread can intern off-GIL while one Python/Cython collector path is active under the GIL. |
| 4 / 16 | Stress cases for future higher-concurrency scenarios, including possible free-threaded/no-GIL Python work. |

In dd-trace-py, profile mutation and serialization are guarded by `profile_mtx`, but dictionary interning can happen before a sample is added to the profile. This means dictionary insertion can still be concurrent even when profile writes are serialized.

## What this does not cover

This benchmark does not model every profiler behavior. In particular, it does not cover:

- function or mapping insertion,
- duplicate-heavy/cached lookup workloads,
- profile serialization reading from the dictionary,
- full profiler end-to-end behavior.

## Why only `ProfilesDictionary`?

I originally tried an additional synthetic benchmark comparing 4 vs 16 shards. Local exploratory results suggested 16 shards helps once there is concurrent insertion:

```text
1 thread:   4 shards ~40 µs,   16 shards ~55 µs
2 threads:  4 shards ~157 µs,  16 shards ~129 µs
4 threads:  4 shards ~385 µs,  16 shards ~280 µs
16 threads: 4 shards ~2.64 ms, 16 shards ~1.61 ms
```

However, that synthetic comparison also changed total initial hash-table capacity because the capacity was applied per shard. Since the current follow-up keeps the production shard count at 16, this PR stays minimal and only adds the `ProfilesDictionary` benchmark.

If we revisit shard count later, we should add a dedicated shard-count benchmark that holds total starting capacity constant across shard counts.

## How to test the change?

- cargo +nightly-2026-02-08 fmt --all -- --check
- cargo check -p libdd-profiling --benches
- cargo +stable clippy -p libdd-profiling --benches -- -D warnings

[PROF-14423](https://datadoghq.atlassian.net/browse/PROF-14423)

[PROF-14423]: https://datadoghq.atlassian.net/browse/PROF-14423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libdd-profiling/benches/profiles_dictionary.rs
```diff
@@ -0,0 +1,162 @@
...
+        let bytes = black_box(string.as_bytes());
+        black_box(set.try_insert_common(bytes, bytes).unwrap());
+    }
+}
+
+fn insert_sharded_strings_concurrently<const N: usize>(
+    strings: &[Vec<String>],
+) -> Sharded<SliceSet<u8>, N> {
+    let set =
+        Sharded::<SliceSet<u8>, N>::try_new_with_min_capacity(HASH_TABLE_MIN_CAPACITY).unwrap();
```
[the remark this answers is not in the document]
taegyunkim: Good catch. I dropped the synthetic shard-count benchmark from this PR so it only adds the production `ProfilesDictionary` benchmark. If we revisit shard count later, we should add a dedicated shard-count benchmark that holds total starting capacity constant across shard counts.

morrisonlevi (commented):
How stable is the benchmark in your experience so far? It seems like it would be pretty variable, but at least you have a startup barrier in there.

libdd-profiling/benches/profiles_dictionary.rs
```diff
@@ -0,0 +1,84 @@
...
+
+fn make_strings(thread_count: usize) -> Vec<Vec<String>> {
+    (0..thread_count)
+        .map(|thread_id| {
+            (0..STRINGS_PER_THREAD)
+                .map(|string_id| {
+                    format!(
+                        "/opt/datadog/profiler/thread-{thread_id}/module-{string_id:04}/function-{}::{}",
+                        string_id % 97,
+                        string_id.wrapping_mul(2_654_435_761usize)
```
morrisonlevi: Is there any rhyme or reason for these two contants? Just arbitrary but different primes?

libdd-profiling/benches/profiles_dictionary.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+pub fn bench_profiles_dictionary(c: &mut Criterion) {
+    let mut group = c.benchmark_group("profiles_dictionary/unique_string_inserts");
+    group.warm_up_time(Duration::from_secs(1));
+    group.measurement_time(Duration::from_secs(5));
+    group.sample_size(10);
+
+    for thread_count in THREAD_COUNTS {
+        let strings = make_strings(thread_count);
```
morrisonlevi: This is made per-thread because `Vec<Vec<T>>` can easily be made to `&[Vec<T>]` but `&[&[T]]` changes memory layout, I think?

If so... why is it a `Vec<Vec<T>>` anyway? What's the purpose of the nesting? I doubt it's too important, just trying to understand.

libdd-profiling/benches/profiles_dictionary.rs
```diff
@@ -0,0 +1,84 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+use criterion::{black_box, criterion_group, BatchSize, BenchmarkId, Criterion, Throughput};
+use libdd_profiling::profiles::datatypes::ProfilesDictionary;
+use std::sync::Barrier;
+use std::thread;
+use std::time::Duration;
+
+const THREAD_COUNTS: [usize; 4] = [1, 2, 4, 16];
```
danielsn, line 11: do we need to ensure we have enough cores?

libdd-profiling/benches/profiles_dictionary.rs
```diff
@@ -0,0 +1,84 @@
...
+const STRINGS_PER_THREAD: usize = 1024;
+
+fn make_strings(thread_count: usize) -> Vec<Vec<String>> {
+    (0..thread_count)
+        .map(|thread_id| {
+            (0..STRINGS_PER_THREAD)
+                .map(|string_id| {
+                    format!(
+                        "/opt/datadog/profiler/thread-{thread_id}/module-{string_id:04}/function-{}::{}",
```
danielsn: Should we vary the lengths of the strings? I'd expect us to have both filenames, but also functionnames?

libdd-profiling/benches/profiles_dictionary.rs
```diff
@@ -0,0 +1,84 @@
...
+                    )
+                })
+                .collect()
+        })
+        .collect()
+}
+
+fn insert_profile_strings(dict: &ProfilesDictionary, strings: &[String]) {
+    for string in strings {
+        black_box(dict.try_insert_str2(black_box(string.as_str())).unwrap());
```
danielsn: why do we need two black boxes?

libdd-profiling/benches/profiles_dictionary.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+fn insert_dictionary_strings_concurrently(strings: &[Vec<String>]) -> ProfilesDictionary {
+    let dict = ProfilesDictionary::try_new().unwrap();
+
+    if let [strings] = strings {
```
danielsn, line 55: Is this just to prevent the barrier if there is only one thread?

libdd-profiling/benches/profiles_dictionary.rs
```diff
@@ -0,0 +1,84 @@
...
+                        string_id % 97,
+                        string_id.wrapping_mul(2_654_435_761usize)
+                    )
+                })
+                .collect()
+        })
+        .collect()
+}
+
+fn insert_profile_strings(dict: &ProfilesDictionary, strings: &[String]) {
```
danielsn, line 46: Why do we pass in the list of strings rather than generating them here

libdd-profiling/benches/profiles_dictionary.rs
taegyunkim: Yes, it is.

libdd-profiling/benches/profiles_dictionary.rs
taegyunkim: Defined constants and added comments

libdd-profiling/benches/profiles_dictionary.rs
taegyunkim: I checked the GitLab benchmark config and we don't set a CPU request/limit or any explicit core cap for the job. It runs on `runner:apm-k8s-tweaked-metal` benchmark runner, which turned out to be Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz, from https://benchmarking-service.us1.prod.dog/benchmarks/36659289.

libdd-profiling/benches/profiles_dictionary.rs
taegyunkim: Good point, dropped inner one.

libdd-profiling/benches/profiles_dictionary.rs
taegyunkim: To avoid capturing string formatting/allocation in the bench

libdd-profiling/benches/profiles_dictionary.rs
taegyunkim: Updated

libdd-profiling/benches/profiles_dictionary.rs
taegyunkim: Each inner Vec<T> is passed to one worker thread. We pre-partition the strings by thread so the measured path doesn’t need to generate or chunk the input.

taegyunkim:
> How stable is the benchmark in your experience so far? It seems like it would be pretty variable, but at least you have a startup barrier in there.

@morrisonlevi Good question, I checked this again.

I think I remember trying benchmark runs on gitlab multiple times and got stable results, I'm going to re-run them again at the latest commit for updated data.

From my local runs using workspaces, 

   | case | observed local behavior |
   |---|---|
   | 1 thread | ~278 to 280 µs, confidence interval width under 1% |
   | 2 threads | ~547 to 549 µs, confidence interval width around 1% |
   | 4 threads | ~730 to 735 µs, confidence interval width around 1 to 2% |
   | 16 threads | ~2.25 to 2.31 ms, noisier, up to ~7% CI width in one run |

at least for 1 ~2 threads cases which I believe more resembles the case of dd-trace-py and ddprof profilers, are stable enough to be used as a baseline. 

Yeah, the barrier helps avoid measuring staggered thread starts.

danielsn (approved):

PR #2091, opened 2026-06-06
feat(otlp)!: Export OTLP spans with attribute-level OTel compatibility

zacharycmontoya:
# What does this PR do?

This PR adds a configuration to the Trace Exporter to export spans with purely OpenTelemetry semantics. When enabled, this configuration updates the DD Span => OTLP Span mapping to exclude the following span tags from the output:
- `service.name`
- `operation.name`
- `resource.name`
- `span.type`
- `error.msg`
- `error.message`
- `span.kind`

This change also fixes the DD Span error message lookup to account for both the `error.msg` and `error.message` tag keys, since they are used interchangeably across Datadog SDKs.

# Motivation

We want to enable OpenTelemetry SDK users to migrate to the Datadog SDK in their language ecosystem, so this change allows them to export OTLP spans from the Datadog SDK without the shape of their data changing. This is most notable when using the OpenTelemetry Tracing API and OpenTelemetry Instrumentation Libraries to generate traces -- with this new semantic mode, the OTLP span should reach the backend without its semantics changing.

# Additional Notes

This may result in a breaking change since it affects the public function `libdd_trace_utils::otlp_encoder::map_traces_to_otlp`.

# How to test the change?

This is tested through unit tests in `libdd-trace-utils/src/otlp_encoder/mapper.rs` that validate the Datadog-specific attributes are excluded from the OTLP span.


libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -30,6 +30,7 @@ const MAX_ATTRIBUTES_PER_SPAN: usize = 128;
 pub fn map_traces_to_otlp<T: TraceData>(
     trace_chunks: Vec<Vec<Span<T>>>,
     resource_info: &OtlpResourceInfo,
+    enable_otel_trace_compatibility: bool,
```
zacharycmontoya: For maintainers, is it preferred to make a breaking change by adding this bool to the public function? Or add a separate overload so as to not break back-compat? Or smuggle the bool value into the resource info (where it doesn't really belong)?

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -296,6 +297,16 @@ impl TraceExporterBuilder {
         self
     }
 
+    /// Enable OTel trace compatibility mode: omit DD-specific per-span attributes
+    /// (`service.name`, `operation.name`, `resource.name`, `span.type`) from the OTLP payload.
+    ///
+    /// Use this when exporting to a native OTel backend that does not expect Datadog semantics,
+    /// for example when `DD_TRACE_OTEL_COMPATIBILITY_ENABLED=true`.
```
mabdinur: I am not a big fan of this name. It implies that dd specific attributes break otel support. The existence of dd attributes shouldn't break otlp ingestion

zacharycmontoya: Do you have any suggestions for a new name? The intent is that we do not introduce overriding fields such as `operation.name` and `service.name` so the data remains identical to what the OTel SDK produces. These DD attributes do not break ingestion, but they change identifying information about a span so it reports differently when DD SDK handles it, so I'm introducing a flag to remove the transformation

marcotc (approved):

libdd-data-pipeline/src/trace_exporter/builder.rs
marcotc: @mabdinur would something like `DD_TRACE_OTEL_FORMAT` make more sense? The goal is to emit as close to pure OTEL data as possible with this flag.

zacharycmontoya: Or something like `DD_TRACE_OTEL_SEMANTIC_CONVENTIONS_ENABLED`? This would simultaneously _keep_ data in the format provided by OpenTelemetry instrumentations and in the language SDKs would _enable_ us to generate spans from our integrations with OpenTelemetry semantic conventions (though I don't think we should give the user the ability to select a spec version themselves, we would just select a version and publish what version we choose)

zacharycmontoya: We discussed this offline and landed on `DD_TRACE_OTEL_SEMANTICS_ENABLED` as an agreeable name. I've updated this in 06a721229e96d8c344a355ffcd510b0c2b623223

link04 (approved):
Nothing I found with the AI could be breaking based on the expectations.

marcotc (approved):

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -30,6 +30,7 @@ const MAX_ATTRIBUTES_PER_SPAN: usize = 128;
 pub fn map_traces_to_otlp<T: TraceData>(
     trace_chunks: Vec<Vec<Span<T>>>,
     resource_info: &OtlpResourceInfo,
+    otel_trace_semantics_enabled: bool,
```
yannham, line 33: Very nitpick, but maybe this could be a proper `enum`, like `TraceSemantics { Datadog, Otel }`, to avoid boolean blindness. Though unimportant and a matter of taste, so feel free to ignore.

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -358,11 +382,31 @@ fn map_attributes<T: TraceData>(span: &Span<T>, resource_service: &str) -> (Vec<
...
-        + (if has_resource_name { 1 } else { 0 })
-        + span.meta.len()
+    let excluded_compat_tags = if otel_trace_semantics_enabled {
+        span.meta.contains_key("error.msg") as usize
+            + span.meta.contains_key("error.message") as usize
+            + span.meta.contains_key("span.kind") as usize
+    } else {
+        0
+    };
+    let total = (if has_per_span_service && !otel_trace_semantics_enabled {
```
yannham: You can use boolean to numeric conversion to avoid the if chains here, something like

```rust
let total = (has_per_span_service && !otel_trace_semantics_enable) as usize
+ ...
```

yannham (approved):
Left some nits but LGTM

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -296,6 +297,16 @@ impl TraceExporterBuilder {
         self
     }
 
+    /// Enables OTel trace semantics, which does not add DD-specific per-span attributes
+    /// (`service.name`, `operation.name`, `resource.name`, `span.type`) to the OTLP payload.
+    /// This is useful when exporting to a native OTel backend that does not expect Datadog
+    /// semantics. The host language tracer is expected to observe this behavior by setting the
+    /// `DD_TRACE_OTEL_SEMANTICS_ENABLED` environment variable to `true`.
+    pub fn enable_otel_trace_semantics(&mut self) -> &mut Self {
```
ekump, line 340: Do you need to add a corresponding function to the FFI API?

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -35,4 +35,7 @@ pub struct OtlpTraceConfig {
     /// Protocol (for future use; currently only HttpJson is supported).
     #[allow(dead_code)]
     pub(crate) protocol: OtlpProtocol,
+    /// When `true`, does not add DD-specific per-span attributes (`service.name`,
```
ekump: Aren't you also stripping `error.msg`, `error.message`, and `span.kind`?

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -296,6 +297,16 @@ impl TraceExporterBuilder {
         self
     }
 
+    /// Enables OTel trace semantics, which does not add DD-specific per-span attributes
+    /// (`service.name`, `operation.name`, `resource.name`, `span.type`) to the OTLP payload.
```
ekump: Same comment as [here](https://github.com/DataDog/libdatadog/pull/2091/changes#r3431533309).

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -139,8 +146,15 @@ fn map_span<T: TraceData>(
         .get("span.kind")
         .map(|v| tag_to_otlp_kind(v.borrow()))
         .unwrap_or_else(|| dd_type_to_otlp_kind(span.r#type.borrow()));
-    let (attributes, dropped_attributes_count) = map_attributes(span, resource_service);
-    let error_msg = span.meta.get("error.msg").map(|v| v.borrow().to_string());
+    let (attributes, dropped_attributes_count) =
+        map_attributes(span, resource_service, otel_trace_semantics_enabled);
+    // DD tracers use either "error.msg" or "error.message".
+    // A span should only carry one of these, so the order of preference is arbitrary.
```
ekump: nit and ignore if I'm wrong....

```suggestion
// A span carries at most one of these; error.message is used by all SDKs except .NET, which uses error.msg
```

libdd-trace-utils/src/otlp_encoder/mapper.rs
ekump: How about...

```rust
let promoted = if otel_trace_semantics_enabled {                                                                                                                                                                                                                                      
    0                                                                                                                                                                                                                                                                                 
} else {                                                                                                                                                                                                                                                                              
    has_per_span_service as usize                                                                                                                                                                                                                                                     
        + has_operation_name as usize                                                                                                                                                                                                                                                 
        + has_span_type as usize                                                                                                                                                                                                                                                      
        + has_resource_name as usize                                                                                                                                                                                                                                                  
};                                                                                                                                                                                                                                                                                    
let total = promoted                                                                                                                                                                                                                                                                  
    + (span.meta.len() - excluded_compat_tags)                                                                                                                                                                                                                                        
    + span.metrics.len()                                                                                                                                                                                                                                                              
    + span.meta_struct.len();
```

ekump (approved):
minor comments, LGTM.

libdd-data-pipeline/src/otlp/config.rs
zacharycmontoya: Yep, updated in d5827ffc1238f994ccd54e53e4a611a35e3ff3a7

libdd-data-pipeline/src/trace_exporter/builder.rs
zacharycmontoya: Thanks, updated in d5827ffc1238f994ccd54e53e4a611a35e3ff3a7

libdd-trace-utils/src/otlp_encoder/mapper.rs
zacharycmontoya: Thanks, updated in d5827ffc1238f994ccd54e53e4a611a35e3ff3a7

libdd-trace-utils/src/otlp_encoder/mapper.rs
zacharycmontoya: Thanks, updated in d5827ffc1238f994ccd54e53e4a611a35e3ff3a7

libdd-trace-utils/src/otlp_encoder/mapper.rs
zacharycmontoya: Discussed offline, the public function map_traces_to_otlp is being updated to accept a bool argument and since our SDKs are the only consumers, this change will be accepted and a new major being published. No other changes to public methods / FFI is needed

libdd-data-pipeline/src/trace_exporter/builder.rs
zacharycmontoya: I'll follow up and post a change to the FFI API as a separate PR, as this is only consumed in dd-trace-py for now

libdd-trace-utils/src/otlp_encoder/mapper.rs
zacharycmontoya: Thanks for the feedback. I'll keep the boolean for this argument since this won't be expanding any further

PR #2093, opened 2026-06-08
test: fix timeouts on heavily contended scenarios

hoolioh:
# What does this PR do?

Three related test-only fixes for flakes that surface on heavily loaded / slow CI runners (observed on CentOS and macOS-15). No production code changes — all edits are in #[cfg(test)] modules and dev-dependencies.

## Changes

1. retry_strategy.rs — virtual clock for backoff tests

The four test_retry_strategy_{constant,linear,exponential,jitter} tests measured wall-clock time around a single delay().await and asserted both a lower and upper bound (±100 ms tolerance). On macOS-15, test_retry_strategy_jitter measured 682 ms for a delay whose maximum is ~149 ms — pure scheduler overshoot.

Switched them to #[tokio::test(start_paused = true)]. Since the tests already use tokio::time::Instant and the sleep capability is tokio::time::sleep, tokio's paused clock auto-advances virtual time to the next timer deadline when the runtime is idle, so measured elapsed equals the exact slept duration — deterministic and independent of runner load. As a bonus these now run in ~2 ms instead of hundreds. Added tokio's test-util feature (required for start_paused).

2. send_data/mod.rs — widen timeout margins in the send-timeout tests

request_error_timeout_v04 / _v07 ran a real hyper request against an httpmock server, racing a 200 ms client timeout against a 500 ms mock delay. On a slow runner the 200 ms timeout could fire before the request reached the server, so httpmock recorded zero calls and assert_calls_async panicked with "No request has been received by the mock server."

Widened the margin so the request reliably lands while the timeout path is still exercised: client timeout 200 → 500 ms, mock delay 500 → 1000 ms. To keep wall time bounded despite the larger timeout, pinned a constant 100 ms retry backoff (instead of the default exponential ~3.1 s of sleeps) — max_retries stays 5, so requests_count == 6 and all other assertions are unchanged. (These can't use the paused clock from change #1, since they depend on genuine socket I/O.)

3. libdd-common/src/timeout.rs — drop an unreliable upper-bound assertion

A TimeoutManager test asserted elapsed < 100 ms after a 10 ms sleep. The lower bound (elapsed >= 10 ms) is the meaningful guarantee; the upper bound is the same wall-clock-vs-scheduling flake. Removed the upper-bound check, keeping the lower bound.

# Motivation

Several tests assert on real wall-clock elapsed time around .await points or sleeps. On a CPU-starved runner the timer fires on schedule but the task isn't scheduled back promptly, so measured elapsed time overshoots the assertion's upper bound — or, in the send tests, an aggressive client timeout fires before the request is even pushed onto the socket. These are timing races, not logic bugs.


libdd-common/src/timeout.rs
```diff
@@ -91,9 +91,6 @@ mod tests {
         std::thread::sleep(Duration::from_millis(10));
         let elapsed = manager.elapsed();
         assert!(elapsed >= Duration::from_millis(10));
-
-        #[cfg(not(miri))] // miri allows the clock to go arbitrarily fast
-        assert!(elapsed < Duration::from_millis(100)); // Should be reasonable
```
ekump, line 96: I think the same is being done [here](https://github.com/DataDog/libdatadog/blob/f9b10c6ed5a3d37cbbd067db4d4e8f5a96c84ea0/libdd-common/src/timeout.rs#L62). Should we remove that too?

ekump (approved):

libdd-common/src/timeout.rs
hoolioh: Good catch. I'll remove it too.

PR #2095, opened 2026-06-08
feat(otel-thread-ctx)!: add self check capability

yannham:
# What does this PR do?

This PR adds a self-test capability to libdd-otel-thread-ctx (and to the FFI), for both shared libraries and statically linked executables. The test checks that the TLS symbol is visible in the dynsym table of the library, and that the model is TLSDESC

# Motivation

OTel thread context requires a specific linking setup, which if not met, makes the context non-discoverable and thus the whole feature unusable. It's thus useful, at least for the CI and post-build sanity checks of downstream profilers to be able to verify that a build is functional.

# Additional Notes

The PR is feature-gated so that the `readelf` dependency isn't pulled if the check is not useful (typically, it can be peformed at build time but be skipped from the final tracer/profiler if desired).

# How to test the change?

The existing test has been refactored to use that code (instead of commands). I've checked that removing the TLSDESC dialect configuration option from the linker, or removing the dynamic list which doesn't export the symbol, make the test fail with the expected error message.


ivoanjo:
I'm curious -- how would this interact with https://github.com/DataDog/libdatadog/pull/2129 ?

yannham:
#2129 should be an internal and feature-preserving refactoring that doesn't change anything from the reader side, so the implementation of this sanity check doesn't depend in any way on whether #2129 is used or not (we basically just read our own elf like an external reader), and #2129 should ideally pass the sanity check once it lands as well.

libdd-otel-thread-ctx-ffi/Cargo.toml
```diff
@@ -15,12 +15,16 @@ crate-type = ["staticlib", "cdylib", "lib"]
...
+libdd-common-ffi = { path = "../libdd-common-ffi", default-features = false, optional = true }
 libdd-otel-thread-ctx = { path = "../libdd-otel-thread-ctx" }
 
 [features]
 default = ["cbindgen"]
 cbindgen = ["build_common/cbindgen", "libdd-common-ffi/cbindgen"]
+sanity-check = ["dep:libdd-common-ffi", "libdd-otel-thread-ctx/sanity-check"]
+
+[dev-dependencies]
+libdd-otel-thread-ctx = { path = "../libdd-otel-thread-ctx", features = ["sanity-check"] }
```
ivoanjo, line 27: Very minor: I'm curious, why does this still need to be a dev dependency? Is it to enable the `elf_properties` test?

If so, should we maybe move it to `libdd-otel-thread-ctx`? Since it's a trivial pass-through, we could maybe slightly simplify things.

ivoanjo (approved):
👍 LGTM this looks good! Thanks for the patience with my extremely slow review 😅

It might be worth getting a second quick pass from @scottgerring or @nsavoire since they've been spending a lot more time on the elf details than I have, but I'd say not a blocker (e.g. they can give a pass after we merge too)

libdd-otel-thread-ctx-ffi/Cargo.toml
yannham: Yes it is for the `elf_properties` test.

Unfortunately we can't really move it to `libdd-otel-thread-ctx`: it's almost a pass-through, but instead of using the current ELF file (which would be the test executable on the `libdd-otel-thread-ctx`), this has a bit of logic to apply the test on the `.so` shared library instead. Put differently, moving it to `libdd-otel-thread-ctxt` with the default behavior (self-ELF) would test the statically linking case, while this one tests the dynamically linked case, which has to live within the FFI.

ivoanjo: Ah, I see. Maybe drop that as a comment somewhere? That's a very subtle, yet important point, so I think it's important to preserve it.

libdd-otel-thread-ctx/src/sanity_check.rs
```diff
@@ -0,0 +1,142 @@
...
+    let addr = sanity_check as *const () as usize;
+    let maps =
+        std::fs::read_to_string("/proc/self/maps").context("failed to read /proc/self/maps")?;
+    for line in maps.lines() {
+        // Format: address perms offset dev inode [pathname]
+        let fields: Vec<&str> = line.split_whitespace().collect();
+        if fields.len() < 6 {
+            continue;
+        }
+        let path = fields[5];
```
scottgerring: I think this will break if the path contains whitespace; pretty unlikely but might not hurt  to join across `5..`

libdd-otel-thread-ctx-ffi/src/lib.rs
```diff
@@ -8,6 +8,20 @@
 #[cfg(target_os = "linux")]
 pub use linux::*;
 
+/// Verify that this binary was linked with the correct options such that the thread contexts are
+/// visible to an external reader (typically the eBPF profiler).
+///
+/// Returns `VoidResult::Ok` if all checks pass, or a `VoidResult::Err` with a
+/// diagnostic message on failure.
+#[cfg(all(target_os = "linux", feature = "sanity-check"))]
```
scottgerring, line 16: do we arch gate for non-amd64-or-arm64 otherwise in libdatadog, or do we just assume all linux builds must happen on one of these?

scottgerring (approved):
cool idea!
couple nits but lgtm

libdd-otel-thread-ctx-ffi/src/lib.rs
yannham: I don't know for libdatadog, but we should surely do that for libdd-otel-thread-ctx at least. I'll double check and add a compile-time error for unsupported architectures, if not already there.

yannham: Yup, guarded here: https://github.com/DataDog/libdatadog/blob/4b79b7ed87113bea01db583d54e13fb0c2a19e74/libdd-otel-thread-ctx/build.rs#L24

ivoanjo:
@yannham I'm going to go ahead and re-add this to the merge queue, hope that's fine with you ;)

PR #2100, opened 2026-06-09
fix(crashtracking): use `AtomicBool` readiness signal

gyuheon0h:
# What does this PR do?

This PR replaces potentially flaky barrier + sleep synchronization in multi-thread collection tests where workers may have not entered user-space code before ptrace captures them

Agent input here:

```
When the crash triggers, the worker might still be:

In kernel space returning from the futex syscall (barrier implementation)
In glibc's pthread barrier code which on CentOS 7's glibc 2.17 may lack proper .eh_frame unwind info
```

# Motivation

Reports of the multi thread collection tests flaking on CentOS7.

# Additional Notes

# How to test the change?

### DIFF
--- a/bin_tests/src/modes/unix/test_017_multi_thread_collection.rs
+++ b/bin_tests/src/modes/unix/test_017_multi_thread_collection.rs
@@ -13,27 +13,6 @@ use std::time::{Duration, Instant};
 
 pub struct Test;
 
-static WORKER_0_READY: AtomicBool = AtomicBool::new(false);
-static WORKER_1_READY: AtomicBool = AtomicBool::new(false);
-
-#[inline(never)]
-fn worker_fn_0() {
-    WORKER_0_READY.store(true, Ordering::Release);
-    loop {
-        std::hint::black_box(0x17_00u64);
-        std::hint::spin_loop();
-    }
-}
-
-#[inline(never)]
-fn worker_fn_1() {
-    WORKER_1_READY.store(true, Ordering::Release);
-    loop {
-        std::hint::black_box(0x17_01u64);
-        std::hint::spin_loop();
-    }
-}
-
 impl Behavior for Test {
     fn setup(
         &self,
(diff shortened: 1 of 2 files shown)


gyuheon0h:
* **#2100** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2100?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2100?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bin_tests/src/modes/unix/test_017_multi_thread_collection.rs
```diff
@@ -7,15 +7,18 @@
...
 pub struct Test;
 
-// Black box to prevent compiler optimization
+static WORKER_0_READY: AtomicBool = AtomicBool::new(false);
+static WORKER_1_READY: AtomicBool = AtomicBool::new(false);
+
 #[inline(never)]
 fn worker_fn_0() {
+    WORKER_0_READY.store(true, Ordering::Release);
     loop {
```
danielsn: we still need a comment explaining why we do this

bin_tests/src/modes/unix/test_017_multi_thread_collection.rs
```diff
@@ -46,26 +50,24 @@ impl Behavior for Test {
     }
 
     fn post(&self, _output_dir: &Path) -> anyhow::Result<()> {
-        let barrier = Arc::new(Barrier::new(3));
+        WORKER_0_READY.store(false, Ordering::Release);
```
danielsn: Can you document why both the false and true posts are `release`?

bin_tests/src/modes/unix/test_017_multi_thread_collection.rs
```diff
@@ -7,15 +7,18 @@
...
-use std::sync::{Arc, Barrier};
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 pub struct Test;
 
-// Black box to prevent compiler optimization
+static WORKER_0_READY: AtomicBool = AtomicBool::new(false);
```
danielsn: If you're rolling your own barrier here, add a comment for future maintainers

hoolioh (approved):
LGTM although, as pointed out, it'd be great to have a few comments for clarifying the fix.

bin_tests/src/modes/unix/test_017_multi_thread_collection.rs
yannham: To be very pedantic, I think we could use `Relaxed` here, because we're before thread spawning (so those stores are sequenced-before whatever threads execute).

Similarly since the boolean doesn't protect any non-atomic memory, we can also use relaxed loads as well when we read from it (stronger ordering doesn't make any promise that something will arrive "sooner"; just what other memory accesses can look like they were re-ordered with the atomic store).

All of that being said, we're in a test where performance doesn't matter, so we don't really care about using a stronger ordering either 🤷‍♂️

bin_tests/src/modes/unix/test_017_multi_thread_collection.rs
yannham: Very nit: you could move the `static` variables directly into `post`, instead of the top-level

yannham (approved):
Agreed on more documentation, otherwise LGTM

PR #2103, opened 2026-06-10
refactor(rc-fetcher)!: simplify code

paullegranddc:
# Motivation

Before adding more code for agentless fetching, I thought we could first simplify a bit the current implementation.

# Changes

* Use hashbrown + Equivalent trait impl to replace casting to dyn struct for hasmap borrowed indexations
* remove agentless endpoint building. This is a feature that needs to be enabled through an explicit setting, not a slightly different endpoint
* copy target path instead of using transmute 'static cast to reference other data
* Use let else .. to flatten data validation
* rename opaque_state to client_state to match the type na me
* extract building config_request to separate functions
* clear client_state.last_error only after the request has succeeded

### DIFF
--- a/libdd-remote-config/src/fetch/fetcher.rs
+++ b/libdd-remote-config/src/fetch/fetcher.rs
@@ -2,13 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::targets::{Root, TargetsList};
-use crate::{
-    RemoteConfigCapabilities, RemoteConfigPath, RemoteConfigPathRef, RemoteConfigPathType,
-    RemoteConfigProduct, Target,
-};
+use crate::{RemoteConfigCapabilities, RemoteConfigPath, RemoteConfigProduct, Target};
 use base64::Engine;
+use hashbrown::HashMap;
 use http::uri::PathAndQuery;
-use http::uri::Scheme;
 use http::StatusCode;
 use http_body_util::BodyExt;
 use libdd_common::{http_common, Endpoint, MutexExt};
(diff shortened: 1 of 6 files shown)


libdd-remote-config/src/fetch/fetcher.rs
```diff
@@ -491,96 +487,84 @@ impl<S: FileStorage> ConfigFetcher<S> {
...
-        opaque_state.last_configs = response.client_configs;
-        opaque_state.last_config_paths = config_paths;
+        client_state.targets_version = targets_list.signed.version as u64;
+        client_state.last_config_paths = config_paths;
         Ok(Some(configs))
     }
 }
 
-fn get_product_endpoint(subdomain: &str, endpoint: &Endpoint) -> Endpoint {
+fn get_agent_configs_endpoint(endpoint: &Endpoint) -> Endpoint {
```
bwoebi, line 560: Are you removing this to re-add it again in the next PR? :-D

bwoebi (approved):
Looks fairly mechanical and makes sense.

libdd-remote-config/src/fetch/fetcher.rs
bwoebi: > remove agentless endpoint building. This is a feature that needs to be enabled through an explicit setting, not a slightly different endpoint

I disagree here, the endpoint.api_key _is_ the setting in my opinion. The code paths taken shall depend on that.

paullegranddc: Probably, but I prefer doing all additions at once.
The logic to switch to the backend based on the absence of scheme in the url seemed weird, like what happens if a user sets the DD_TRACE_AGENT_URL to `localhost:8126` without putting http?

bwoebi: Yeah, that's indeed weird. Can't happen in PHP because tracer will enforce http, but yeah.

PR #2104, opened 2026-06-10
feat(profiling): Add setting to omit `local root span id` from serialized pprof [PROF-15045]

ivoanjo:
# What does this PR do?

This PR adds a new API to profiles: `set_omit_local_root_span_id_when_serializing`. This API can be used to omit the `local root span id` from serialized pprof files.

**Note that nothing else changes**: The `local root span id` is still being tracked and used for everything in the profile, only never actually written in the pprof. Also, this is off by default and needs to be set on every profile before serialization (gets disabled again on profile reset).

# Motivation

In <https://datadoghq.atlassian.net/wiki/spaces/PROF/pages/6821413373/ADR+2026-06-04.1+Continue+using+span-child+filter+mode+remove+local-root-span+backend+support> we discussed experimenting with having libraries not emit local root span id. This is very relevant for implementing OTel thread context, since "local root span" is not a common otel concept and we need to do extra work in all libraries to try to keep it.

Having this in libdatadog and in an off-by-default opt-in-api will allow us to easily test this.

# Additional Notes

N/A

# How to test the change?

This change includes test coverage. I plan to wire this up to dd-trace-rb and add an end-to-end test there.

### DIFF
--- a/libdd-profiling/src/internal/profile/mod.rs
+++ b/libdd-profiling/src/internal/profile/mod.rs
@@ -1504,11 +1504,12 @@ mod api_tests {
     }
 
     #[test]
-    fn omit_local_root_span_id_when_serializing() -> anyhow::Result<()> {
+    fn omit_local_root_span_id_when_serializing() {
         let sample_types = [api::SampleType::CpuSamples, api::SampleType::WallTime];
 
-        let mut profile: Profile = Profile::new(&sample_types, None);
-        profile.set_omit_local_root_span_id_when_serializing(true);
+        let mut regular_profile: Profile = Profile::new(&sample_types, None);
+        let mut omit_profile: Profile = Profile::new(&sample_types, None);
+        omit_profile.set_omit_local_root_span_id_when_serializing(true);
 
         let sample = api::Sample {
             locations: vec![],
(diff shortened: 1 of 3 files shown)


libdd-profiling/src/internal/profile/mod.rs
```diff
@@ -1486,6 +1503,43 @@ mod api_tests {
...
+            }],
+        };
+
+        profile.try_add_sample(sample, None)?;
+        profile.add_endpoint(10, Cow::from("my endpoint"))?;
+
+        let serialized_profile = roundtrip_to_pprof(profile)?;
+        let sample = serialized_profile.samples.first().expect("sample");
+
+        assert_eq!(sample.labels.len(), 1);
```
gyuheon0h: Nit: `assert_eq!(sample.labels.len(), 1)` proves the local root span id label was removed, but only indirectly, maybe a more explicit check here will make the intention more clear

gyuheon0h (approved):

libdd-profiling/src/internal/profile/mod.rs
ivoanjo: @datadog fix this

ivoanjo: (lol)

danielsn (approved):
I guess we still need to set the local-root-span as a label to get the endpoint?

ivoanjo:
> I guess we still need to set the local-root-span as a label to get the endpoint?

Yes -- this is why the change is only on the serializer; we're currently using the "local root span id" as a key for updating the endpoint later.

In the future, libraries could keep using local root span id as the key, or may use any other key (e.g. can even be a ever-incrementing counter, libdatadog doesn't particularly care) but for now I didn't want to touch those interfaces/APIs.

libdd-profiling/src/internal/profile/mod.rs
ivoanjo: This is a good point -- I pushed e35c67de5465698570c9ccda49637edd401c74b1 to address it. In particular, I redesigned the test so it tests a regular profile vs one with this flag, and asserts that they are the same minus the label.

PR #2105, opened 2026-06-11
refactor(change-buffer)!: replace slot index with span_id, fix segment isolation

rochdev:
## Summary

- **Segment isolation bug**: When service A → B → A creates two segments sharing the same `trace_id`, trace-level state (`SetTraceOrigin`, `SetTraceMetaAttr`, `SetTraceMetricsAttr`) was keyed by `trace_id`, so segment 2's writes overwrote segment 1's values before it was flushed. Fixed by adding `segment_id` (u64) to Create ops and keying `SmallSegmentMap` by segment_id instead.
- **Slot index removal**: Operations carried a `u32 slot_index` — a JS-managed array index into four parallel `Vec`s. JS had to allocate/free slot numbers; reusing a slot prematurely would silently drop a live span; `ensure_slot` had to grow all four Vecs in lockstep. Replaced with `span_id` (u64) as the natural key into `HashMap`s, removing the slot allocation contract entirely.
- **Opcode size**: Was `u64` because JS wrote via `setBigUint64`, wasting 6 bytes per op with only 17 opcodes. Reduced to `u16`.
- **SpanHeader removal**: `materialize_header` was an unused direct-write path where JS wrote span fields into a `#[repr(C)]` struct in WASM memory, bypassing the change buffer protocol. It had no `segment_id` field and would have reproduced the isolation bug. Deleted along with `span_header.rs`.
- **Rename `Trace` → `Segment`**: `SmallTraceMap`/`Trace` renamed to `SmallSegmentMap`/`Segment` since the map holds per-segment state, not per-trace state.

## Test plan

- [ ] `cargo test -p libdd-trace-utils --lib --features change-buffer -- change_buffer` passes (16 tests including 3 new segment isolation tests)
- [ ] `cargo check -p libdd-trace-utils --features change-buffer` clean

> 🤖 Generated with [Claude Code](https://claude.com/claude-code)


libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -391,89 +392,135 @@ where
...
+                    segment.origin = Some(origin);
                 }
             }
             OpCode::BatchSetMeta => {
                 let count: u32 = self.get_num_arg(index)?;
-                for _ in 0..count {
-                    let key_id: u32 = self.get_num_arg(index)?;
-                    let val_id: u32 = self.get_num_arg(index)?;
-                    deferred_meta_at(&mut self.deferred_meta, slot)?.insert(key_id, val_id);
+                if !cached_deferred_meta.is_null() {
```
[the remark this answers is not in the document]
rochdev: No longer relevant, this code was deleted.

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -253,97 +228,91 @@ where
...
 
-            let mut span = maybe_span.ok_or(ChangeBufferError::SpanNotFound(*slot as u64))?;
+        for span_id in &span_ids {
+            let segment_id = self.segment_ids.get(span_id).copied().unwrap_or(0);
 
-            self.materialize_deferred_tags(*slot, &mut span);
+            let mut span = self
+                .spans
+                .remove(span_id)
+                .ok_or(ChangeBufferError::SpanNotFound(*span_id))?;
```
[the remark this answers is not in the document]
rochdev: Fixed.

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -167,37 +165,38 @@ fn new_span_pooled<T: TraceData>(
...
+/// most recently processed span.
+///
+/// Avoids repeated HashMap lookups for consecutive ops on the same span.
+/// All pointers must either be all null (when `span_id == 0`) or point into
+/// the live entries of the owning `ChangeBufferState`. They are invalidated
+/// (set to null via `Default`) before any HashMap insertion that could
+/// trigger a rehash — i.e. before every Create op.
+struct SpanCache<T: TraceData> {
+    span_id: u64,
+    span_ptr: *mut Span<T>,
```
danielsn: Having `*mut` pointers here seems a little scary in terms of ensuring stacked borrows etc when they're used

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -208,14 +207,13 @@ where
     ) -> Self {
         ChangeBufferState {
             change_buffer,
-            spans: Vec::with_capacity(256),
-            traces: SmallTraceMap::default(),
+            spans: HashMap::with_capacity(256),
```
danielsn: why 256?

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -253,97 +251,92 @@ where
 
     pub fn flush_chunk(
         &mut self,
-        slot_indices: &[u32],
+        span_ids: Vec<u64>,
```
danielsn: Should this be a slice?

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -253,97 +251,92 @@ where
...
-                .and_then(|opt| opt.take());
+        let mut spans_vec = Vec::with_capacity(span_ids.len());
+        let mut seen_segment_ids: Vec<(u64, usize)> = Vec::new();
 
-            let mut span = maybe_span.ok_or(ChangeBufferError::SpanNotFound(*slot as u64))?;
+        for span_id in &span_ids {
+            let (mut span, segment_id) = self
+                .spans
+                .remove(span_id)
+                .ok_or(ChangeBufferError::SpanNotFound(*span_id))?;
```
danielsn, line 305: If this fails, we stop removing the other spans. Is that what we want?

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -253,97 +251,92 @@ where
...
 
-            self.materialize_deferred_tags(*slot, &mut span);
+            self.materialize_deferred_tags(*span_id, &mut span);
 
-            chunk_trace_id = Some(span.trace_id);
+            if let Some(entry) = seen_segment_ids
```
danielsn: Is this data-structure small enough that `iter_mut.find` is faster than keeping a set?

libdd-trace-utils/src/change_buffer/segment.rs
```diff
@@ -3,8 +3,14 @@
 
 use crate::span::vec_map::VecMap;
 
+/// Per-segment state for a trace chunk.
+///
+/// A segment is an independent visit by a service to a distributed trace. Multiple
```
danielsn, line 8: How do segments differ from spans?

libdd-trace-utils/src/change_buffer/operation.rs
```diff
@@ -36,10 +36,10 @@ pub enum OpCode {
     // TODO: SpanLinks, SpanEvents, StructAttr
 }
 
-impl TryFrom<u32> for OpCode {
+impl TryFrom<u16> for OpCode {
     type Error = ChangeBufferError;
 
-    fn try_from(val: u32) -> Result<Self> {
+    fn try_from(val: u16) -> Result<Self> {
```
danielsn, line 42: this could be a derive using `num-derive` or `int-enum`

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -814,3 +779,323 @@ where
...
+    fn set_trace_origin_on_second_segment_does_not_affect_first_segment() {
+        let mut w = BufWriter::new();
+
+        // Segment 1 — span_id=1: origin → "origin-A" (string id 0)
+        w.op(OP_CREATE, 1); // span_id=1 in header
+        w.u128(TRACE_ID);
+        w.u64(1); // segment_id=1
+        w.u64(0); // parent_id
+        w.op(OP_SET_TRACE_ORIGIN, 1);
+        w.u32(0); // string_id → "origin-A"
```
danielsn, line 824: Should this have a helper function?

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -814,3 +779,323 @@ where
...
+    // -----------------------------------------------------------------------
+
+    struct BufWriter {
+        data: Vec<u8>,
+        count: u64,
+    }
+
+    impl BufWriter {
+        fn new() -> Self {
+            let mut data = Vec::with_capacity(256);
```
danielsn, line 753: why 256?

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -652,146 +631,132 @@ where
...
+                    }
                 }
             }
             OpCode::BatchSetMetric => {
                 let count: u32 = self.get_num_arg(index)?;
                 for _ in 0..count {
                     let key_id: u32 = self.get_num_arg(index)?;
                     let val: f64 = self.get_num_arg(index)?;
-                    deferred_metrics_at(&mut self.deferred_metrics, slot)?.insert(key_id, val);
+                    if let Some(dm) = self.deferred_metrics.get_mut(&op.span_id) {
```
danielsn: I'd probably use `.entry` here, but YMMV

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -391,89 +400,134 @@ where
...
+                // (e.g. after materialize_span removed the deferred map entry).
+                // Skipping the loop while leaving bytes unread corrupts every
+                // subsequent op decoded from this buffer.
                 for _ in 0..count {
                     let key_id: u32 = self.get_num_arg(index)?;
                     let val_id: u32 = self.get_num_arg(index)?;
-                    deferred_meta_at(&mut self.deferred_meta, slot)?.insert(key_id, val_id);
+                    if !cache.deferred_meta.is_null() {
+                        // SAFETY: pointer into self.deferred_meta, valid per function contract.
+                        unsafe { &mut *cache.deferred_meta }.insert(key_id, val_id);
```
danielsn: Are we guaranteed to enforce stacked-borrows here?

libdd-trace-utils/src/change_buffer/mod.rs
```diff
@@ -391,89 +400,134 @@ where
...
                     let key_id: u32 = self.get_num_arg(index)?;
                     let val: f64 = self.get_num_arg(index)?;
-                    deferred_metrics_at(&mut self.deferred_metrics, slot)?.insert(key_id, val);
+                    if !cache.deferred_metrics.is_null() {
+                        // SAFETY: pointer into self.deferred_metrics, valid per function contract.
+                        unsafe { &mut *cache.deferred_metrics }.insert(key_id, val);
+                    }
                 }
             }
             OpCode::Create | OpCode::CreateSpan | OpCode::CreateSpanFull => unreachable!(),
```
danielsn: do we want to have a potential panic here?

libdd-trace-utils/src/change_buffer/segment.rs
rochdev: A segment is the part of a trace that is local to your service for a distributed trace.

libdd-trace-utils/src/change_buffer/mod.rs
yannham: This whole `seen_segment_ids` has been removed upstream in the end.

libdd-trace-utils/src/change_buffer/mod.rs
yannham: I agree. I got rid of them in the upstream PR, using an index instead. I'm not sure saving one vec indexing is worth the unsafe business.

libdd-trace-utils/src/change_buffer/mod.rs
yannham: Yes, because nothing else is touching `self.spans` in any way (creating mutable or immutable references) while we materialize a mutable reference to some span. Basically we could use `HashMap::get_mut` for the lifetime of the materialized mutable span reference instead, but of course the whole point of the pointer is to avoid having to make a hashmap fetch.

libdd-trace-utils/src/change_buffer/mod.rs
yannham: Experimental findings, I guess? I lowered to 128 to reduce memory usage and moved it to constants, so that it can be tweaked easily.

yannham (approved):
Reviewed with Roch, but since I also rebased/amended the PR, it'd better to have another pair of eyes (and a stamp).

paullegranddc (approved):
No objection on principle, but did you benchmark this on dd-trace-js vs the previous implementation?

rochdev:
> No objection on principle, but did you benchmark this on dd-trace-js vs the previous implementation?

The previous implementation was incorrect, so it doesn't matter if it was faster. Also, we shouldn't be relying on people benchmarking on their machines, they should be automated in this repo.

yannham:
> The previous implementation was incorrect, so it doesn't matter if it was faster. Also, we shouldn't be relying on people benchmarking on their machines, they should be automated in this repo.

Hey, this is precisely what I'm trying to do this week :slightly_smiling_face:

PR #2112, opened 2026-06-12
feat(remote-config)!: agentless RC fetcher

paullegranddc:
# Motivation

 In some environments (serverless, Datadog studio), tracers can't rely on a Datadog Agent to proxy Remote Config requests. This PR adds an agentless mode to libdd-remote-config so the fetcher can talk directly to the RC backend, mirroring what the Go agent does today.

 The protocol is different from the agent's /v0.7/config: the backend speaks protobuf and ships raw TUF metadata + target files that the client must verify locally. This  means embedding TUF trust roots per site and running a full Uptane-style validation on every poll.

# Changes

 - Add a new AgentlessFetcher in libdd-remote-config/src/fetch/agentless.rs:
     - TUF verification of director + config repositories using rust-tuf (DataDog fork of the library).
     - Embedded trust roots for prod, staging and gov, selected from the endpoint host. Roots can also be overridden from disk via AgentlessConfig.
     - Hash + length verification of every target against TUF-validated metadata before exposing it to the caller.
  
 - Plug agentless mode into the existing ConfigFetcher:
     - New AgentlessConfig. When set the fetcher dispatches to AgentlessFetcher instead of the agent HTTP path. Invalid configs (empty hostname, non-https endpoint, no API key) downgrade to agent mode with  log.
     - ConfigFetcher::new is now async since TUF root loading is async. The agent-only path is still immediate.
     - Propagate the server-recommended refresh interval through ConfigClientState and into the SharedFetcher run loop.
 - Update the remote_config_fetch example to switch to agentless when DD_API_KEY + DD_SITE are set.
 
 # Additional Notes
 
 Eventually this implementation could be extracted to be it's own crate and we could reuse it in other projects that need a rust client.
 For this reason, the `agentless.rs` module which contains the most of the implementation tries to encapsulate what is "RC agentless" and what is specific to the RC client in libdatadog.
This is why the caching of target files is done in a  wrapper type. This cache is currently quite complex as it permits storing files in shared memory for instance.
 
 Currently this implementation is a simplification of the agent code in two ways:
 * fetches data for only __one__ active client, which is also a simplification compared to what the agent does
 * does not control the polling loop but instead returns a refresh interval parameter which the users of the library should respect.

# How to test the change?

```
   DD_API_KEY=... DD_SITE=datadoghq.com \
     cargo run -p libdd-remote-config --example remote_config_fetch
```

Also this has been tested in dd-trace-rs 
https://github.com/DataDog/dd-trace-rs/pull/263


libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,766 @@
...
+        };
+
+        store(config_repo_mut, &root_path, &metas.roots).await?;
+        store_noversion(config_repo_mut, &timestamp_path, &metas.timestamp).await?;
+        store(config_repo_mut, &snapshot_path, &metas.snapshot).await?;
+        store(config_repo_mut, &targets_path, &metas.top_targets).await?;
+        // TODO: We do not store the delegated targets metadata
+        // This will need to be revisited in order to support proper Uptane
+        // verification of the full configuration data.
+        // store(repo, &targets_path, &metas.delegated_targets).await?;
```
raphaelgavache: I think you can safely skip them, director repo already confirms it's from datadog, for your account, integrity checks and the routing predicate. What it does not confirm is the product behind the config

libdd-remote-config/roots/prod/config_root.json
```diff
@@ -0,0 +1,63 @@
+{
+  "signed": {
+    "_type": "root",
+    "spec_version": "1.0",
+    "version": 16,
+    "expires": "2026-10-31T17:00:00Z",
```
bwoebi: Is this for real? expiration within 4 months? Are we expecting users to continuously update the tracer or have RC completely broken?
I mean, sure, you can load it from disk.., but do users even realize they have to do that?

paullegranddc: This certificate is used to bootstrap the trust chain. During the 1st call to get configuration, the backend wil give you all certificates in the chain from this one to the latest one.
Only the latest certificate in the chain has to be valid, but you can verify new certificates with expired ones.

For instance the certificates used in the agent (v1 at the very start of the chain) are marked as expired since 1970 and this is not an issue

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1165 @@
...
+        RemoteConfigPath,
+    };
+
+    pub(crate) struct TargetCache<'a, Storage: FileStorage> {
+        files: &'a Mutex<HashMap<Arc<RemoteConfigPath>, StoredTargetFile<Storage::StoredFile>>>,
+        storage: &'a Storage,
+        expire_unused_files: bool,
+    }
+
+    impl<'a, S: FileStorage> TargetCache<'a, S> {
```
bwoebi, line 1185: It's okay, though this code shares quite a bit of common logic with the agent fetcher.  E.g. would be trivial to collect/construct NewTarget in fetcher.rs and call store_batch.
I suppose that was done to move faster and avoid untangling the agent fetcher.

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1165 @@
...
+                let parsed = RemoteConfigPath::try_parse(&target.path).map_err(|e| {
+                    anyhow::format_err!("collect_handles: bad path {}: {e:?}", target.path)
+                })?;
+                let stored = files.get_mut(&parsed).ok_or_else(|| {
+                    anyhow::format_err!(
+                        "collect_handles: path {} not found in cache after fetch",
+                        target.path
+                    )
+                })?;
+                if stored.hash != target.primary_hash || stored.meta.length as u64 != target.length
```
bwoebi: What's the goal of re-verifying here? is_cached_batch() omits those entries which are having a different hash/len and fetch_target() handles verification of new ones?
(It's also the meaningful difference to the agentful fetcher path, hence I ask.)

paullegranddc: Hmm I just added the case to be a bit more defensive.
I agree it's not necessary though.

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1174 @@
...
+    /// # Errors
+    /// Returns an error if TUF root initialization fails.
+    /// This can happen for instance if the trust root certificates have expired
+    pub async fn new(cfg: AgentlessConfig, endpoint: Endpoint) -> anyhow::Result<Self> {
+        // Pick the default trust roots based on the endpoint's host. Overrides
+        // (if any) take precedence.
+        let site = endpoint
+            .url
+            .authority()
+            .map(|a| Site::from_host(a.as_str()))
```
bantonsson: Is `authority()` the right thing here? The code in `from_host` seems to expect a hostname only and not potentially `username:password@hostname:port`

paullegranddc: I passed to host for now.

But we need to figure out if people use custom endpoints (HTTP PROXY maybe which is supporter by the agent? or just a reverse proxy which could take http auth 🤔).
For datadog studio I dont think we really care though, so more of a concern in the long term

.github/workflows/lint.yml
```diff
@@ -149,6 +149,7 @@ jobs:
         run: |
           if [[ "${{ matrix.platform }}" == "windows-latest" ]]; then
             export AWS_LC_FIPS_SYS_NO_ASM=1
+            git config --global core.longpaths true
```
paullegranddc, line 152: Needed because the dependency on the rut-tuf git repo causes paths to be longer than what github supports by default and this breaks cargo fetching deps 😢

libdd-remote-config/Cargo.toml
```diff
@@ -36,24 +37,35 @@ test = ["hyper/server", "hyper-util"]
...
+    "formatting",
+], optional = true }
 tracing = { version = "0.1", default-features = false, optional = true }
 serde = "1.0"
 serde_json = { version = "1.0", features = ["raw_value"] }
 serde_with = "3"
 thiserror = "2"
 hashbrown = "0.15"
+# branch = "opw-develop"
+tuf = { git = "https://github.com/DataDog/rust-tuf/", rev = "42fc5b6ac1fe13a997b0e4d0235f0964c18da3fa", default-features = false, optional = true }
```
paullegranddc: This is temporary
To release this feature (for dd-trace-rs for instance) we want to make a github release and also probably release the fork to crates.io

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: I implemented the check

libdd-remote-config/roots/staging/director_root.json
```diff
@@ -0,0 +1,73 @@
...
+      "timestamp": {
+        "keyids": [
+          "6ca796e7b4883af3bb3d522dc0009984dcbf5ad2a6c9ea354d30acc32d8b75d1",
+          "233a529fe7c63b5b9081f6e0e2681cc227f85e04ad434d0a165a2f69b87255a6"
+        ],
+        "threshold": 2
+      }
+    },
+    "spec_version": "1.0",
+    "version": 1
```
mellon85: This is not the latest version of the director root

libdd-remote-config/src/fetch/fetcher.rs
```diff
@@ -154,10 +169,36 @@ impl<S> ConfigFetcherFilesLock<'_, S> {
...
+        let (endpoint, agentless) = match &invariants.agentless {
+            Some(agentless_cfg) => {
+                #[cfg(feature = "agentless")]
+                match (
+                    agentless::make_agentless_configs_endpoint(&invariants.endpoint),
+                    agentless_cfg.hostname.is_empty(),
+                ) {
+                    (Some(e), false) => (e, Some(agentless_cfg.clone())),
+                    (Some(_), true) => {
+                        warn!("rc_config_fetcher: agentless enabled but the hostname is empty. Downgrading to agent endpoint");
```
mellon85: shouldn't these be blocking errors instead of a downgrade? it's a misconfiguration that will make it not work as intended, and if people expect an agentless setup it will fail quietly

raphaelgavache (approved):

libdd-remote-config/roots/prod/config_root.json
mellon85: We are going to do a ceremony soon to have new version, this is the known latest

libdd-remote-config/src/fetch/fetcher.rs
```diff
@@ -201,9 +242,17 @@ impl<S> ConfigFetcherState<S> {
     }
 }
 
+#[allow(clippy::large_enum_variant)]
```
yannham, line 241: Do you not box the variant because it's cfg-gated anyway?

libdd-remote-config/src/fetch/fetcher.rs
```diff
@@ -557,9 +613,83 @@ impl<S: FileStorage> ConfigFetcher<S> {
...
+            client_id,
+            &*client_state,
+        );
+        match &mut self.mode {
+            FetcherMode::Agent => self.fetch_agent(config_req, target, client_state).await,
+            #[cfg(feature = "agentless")]
+            FetcherMode::Agentless(agentless_fetcher) => {
+                #[allow(clippy::expect_used)]
+                let client = config_req.client.expect(
+                    "RC ConfigFetcher::build_config_request should always return a `Some` client",
```
yannham, line 681: Then why is `Client` an `Option`?

libdd-remote-config/src/fetch/fetcher.rs
```diff
@@ -557,9 +613,83 @@ impl<S: FileStorage> ConfigFetcher<S> {
...
+                    client_state.opaque_backend_state = res.opaque_backend_state.clone();
+                }
+                client_state.last_error = None;
+
+                let mut config_paths: HashSet<RemoteConfigPath> = HashSet::new();
+                for target_ref in &res.targets {
+                    if let Ok(parsed) = RemoteConfigPath::try_parse(&target_ref.path) {
+                        config_paths.insert(parsed.into());
+                    }
+                }
```
yannham:
```suggestion
                let config_paths: HashSet<RemoteConfigPath> = res.targets.iter()
                    .filter_map(|target_ref| RemoteConfigPath::try_parse(&target_ref.path).ok())
                    .collect();
```

libdd-remote-config/src/fetch/shared.rs
```diff
@@ -349,6 +357,17 @@ impl SharedFetcher {
                 }
             }
 
+            if let Some(interval) = opaque_state.server_recommended_refresh_interval() {
+                // Keep the run-loop interval in sync with the server-provided value
+                // If the interval is nanoseconds is greater than u64 max, pick the max
```
yannham:
```suggestion
                // If the interval in nanoseconds is greater than u64 max, pick the max
```

libdd-remote-config/src/fetch/single.rs
```diff
@@ -27,21 +30,49 @@ pub struct ConfigOptions {
...
+    }
+
+    pub fn new_no_agentless(
+        sink: S,
+        target: Target,
+        runtime_id: String,
+        options: ConfigOptions,
+    ) -> anyhow::Result<Self> {
+        Ok(SingleFetcher {
+            fetcher: futures::executor::block_on(ConfigFetcher::new(
```
yannham: Where does this executor come from? Should we use SharedRuntime instead, and if not, why?

libdd-remote-config/src/fetch/single.rs
```diff
@@ -108,11 +148,28 @@ impl<S: FileStorage> SingleChangesFetcher<S>
...
+    ) -> anyhow::Result<Self> {
+        Ok(SingleChangesFetcher {
             changes: ChangeTracker::default(),
-            fetcher: SingleFetcher::new(sink, target, runtime_id, options),
-        }
+            fetcher: SingleFetcher::new(sink, target, runtime_id, options).await?,
+        })
+    }
+
+    pub fn new_no_agentless(
```
yannham, line 173: So... `new_with_agent` 😛 ? (well it depends if you're an intuitionistic logician or not)

libdd-remote-config/src/path.rs
```diff
@@ -81,7 +81,13 @@ impl RemoteConfigPath {
                     if parts.len() != 5 {
                         anyhow::bail!("{} is datadog and does not have exactly 5 parts", path);
                     }
-                    RemoteConfigSource::Datadog(parts[1].parse()?)
+                    let org_id: u64 = parts[1].parse()?;
+                    // The agent parses org_id as an int64; reject values it would
+                    // reject (> i64::MAX) so both clients accept/reject the same paths.
+                    if org_id > i64::MAX as u64 {
+                        anyhow::bail!("org_id {} exceeds i64::MAX in path {}", org_id, path);
```
yannham:
```suggestion
                        anyhow::bail!("org_id {org_id} exceeds i64::MAX in path {path}");
```

libdd-remote-config/src/path.rs
```diff
@@ -104,8 +110,20 @@ impl RemoteConfigPath {
                 "LIVE_DEBUGGING_SYMBOL_DB" => RemoteConfigProduct::LiveDebuggerSymbolDb,
                 product => anyhow::bail!("Unknown product {}", product),
             },
-            config_id: parts[parts.len() - 2],
-            name: parts[parts.len() - 1],
+            config_id: {
+                let config_id = parts[parts.len() - 2];
+                if config_id.is_empty() {
+                    anyhow::bail!("empty config_id in path {}", path);
```
yannham:
```suggestion
                    anyhow::bail!("empty config_id in path {path}");
```

libdd-remote-config/src/path.rs
```diff
@@ -104,8 +110,20 @@ impl RemoteConfigPath {
...
+                }
+                config_id
+            },
+            name: {
+                let name = parts[parts.len() - 1];
+                if name.is_empty() {
+                    anyhow::bail!("empty name in path {}", path);
```
yannham:
```suggestion
                    anyhow::bail!("empty name in path {path}");
```

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+        // This is fine because we compare the hash+len against TUF-validated metadata.
+        let mut read = self
+            .director_client
+            .remote_repo()
+            .fetch_target(target_path)
+            .await?;
+        let mut buf = Vec::new();
+        read.read_to_end(&mut buf).await?;
+
+        if buf.len() as u64 != target.length {
```
yannham: Very nit, but if we're sensitive code here, maybe avoid `as` and use `try_from` instead?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+            .collect::<Vec<_>>();
+        let actual_hashes = tuf::crypto::calculate_hashes_from_slice(&buf, hash_algs.as_slice())?;
+        let expected: HashMap<_, _> = target
+            .all_hashes
+            .iter()
+            .map(|(alg, val)| (alg, val))
+            .collect();
+
+        if !(actual_hashes.len() == expected.len()
+            && actual_hashes
```
yannham, line 463: Sorry if that's naive/stupid, I don't know much about the shape of those data. But is it possible for `actual_hashes` to have duplicate entries? I wonder if the test here is basically exact hashmap equality, but if you compare an associative list with potential duplicates and hashmap, then this is incorrect. Otherwise feel free to ignore/resolve

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+            .trusted_snapshot()
+            .map_or(0, |s| s.version());
+        let current_config_root_version = self.config_client.database().trusted_root().version();
+        let current_director_root_version =
+            self.director_client.database().trusted_root().version();
+
+        let all_products = c.products.iter().fold(HashSet::new(), |mut acc, p| {
+            acc.get_or_insert_with(p, String::clone);
+            acc
+        });
```
yannham: Very nit/taste: if a fold is mutating a value and returning it at each iteration, I think a for loop is more adapted

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+            .intersection(&all_products)
+            .cloned()
+            .collect::<Vec<_>>();
+
+        let now = now_unix_milli_ts();
+
+        let (has_error, error) = match c.state.as_ref() {
+            Some(state) if state.has_error => (true, state.error.clone()),
+            _ => (false, String::new()),
+        };
```
yannham, line 521: Why not use an optional for the error and `is_some` instead of `has_error` ?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+    top: &'a tuf::verify::Verified<tuf::metadata::TargetsMetadata>,
+    trusted_delegations: &'a std::collections::HashMap<
+        MetadataPath,
+        tuf::verify::Verified<tuf::metadata::TargetsMetadata>,
+    >,
+) -> Option<&'a TargetDescription> {
+    // Direct hit on the top-level targets (Datadog's are empty, but be safe).
+    if let Some(d) = top.targets().get(path) {
+        return Some(d);
+    }
```
yannham: Nit

```suggestion
    if let direct @ Some(_) = top.targets().get(path) {
        return direct;
    }
```

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+        let matches_scope = delegation
+            .paths()
+            .iter()
+            .any(|pat| target_matches_pattern(path.as_str(), pat.as_str()));
+        if !matches_scope {
+            continue;
+        }
+
+        if let Some(meta) = trusted_delegations.get(delegation.name()) {
+            if let Some(d) = meta.targets().get(path) {
```
yannham, line 992: if-let chains, when 😿

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+            candidates
+                .into_iter()
+                .filter_map(|(path, primary_hash, len)| {
+                    let parsed = RemoteConfigPath::try_parse(path).ok()?;
+                    let stored = files.get(&parsed)?;
+                    if stored.hash == primary_hash && stored.meta.length as u64 == len {
+                        Some(path)
+                    } else {
+                        None
+                    }
```
yannham:
```suggestion
                    (stored.hash == primary_hash && stored.meta.length as u64 == len).then_some(path)
```

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+            &self,
+            candidates: impl IntoIterator<Item = (&'b str, &'b str, u64)>,
+        ) -> hashbrown::HashSet<&'b str> {
+            let files = self.files.lock_or_panic();
+            candidates
+                .into_iter()
+                .filter_map(|(path, primary_hash, len)| {
+                    let parsed = RemoteConfigPath::try_parse(path).ok()?;
+                    let stored = files.get(&parsed)?;
+                    if stored.hash == primary_hash && stored.meta.length as u64 == len {
```
yannham: Same remark as before: should we use safe conversions here?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+            {
+                let parsed_path = match RemoteConfigPath::try_parse(&path) {
+                    Ok(p) => p,
+                    Err(e) => {
+                        warn!("store_batch: failed to parse remote config path {path}: {e:?}");
+                        continue;
+                    }
+                };
+                let parsed_path: Arc<RemoteConfigPath> = Arc::new(parsed_path.into());
+                let length = content.len() as i64;
```
yannham: I don't know the type of `len`, but if it's usize for example, this conversion isn't safe. I would either use `from` for infallible conversions or `try_from` otherwise

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+                let parsed = RemoteConfigPath::try_parse(&target.path).map_err(|e| {
+                    anyhow::format_err!("collect_handles: bad path {}: {e:?}", target.path)
+                })?;
+                let stored = files.get_mut(&parsed).ok_or_else(|| {
+                    anyhow::format_err!(
+                        "collect_handles: path {} not found in cache after fetch",
+                        target.path
+                    )
+                })?;
+                if stored.hash != target.primary_hash || stored.meta.length as u64 != target.length
```
yannham: Re: `as` conversions

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+    let bytes = std::fs::read(override_path)
+        .map_err(|e| format_err!("failed to read TUF root override at {override_path:?}: {e}"))?;
+    Ok(bytes)
+}
+
+/// Fake version sent to RC. We have to do this as the RC backend will not answer if the
+/// agent_version field is empty or lower than a certain version.
+///
+/// This is currently set to the last agent version released
+const FAKE_AGENT_VERSION: &str = "7.78.4";
```
mellon85: we could use a special version here to identify them more easily for debugging purposes. We have a version check for allowed agents and we can add something special. I don't know like 99.0.0

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+        match self {
+            Site::Prod => PROD_DIRECTOR_ROOT,
+            Site::Staging => STAGING_DIRECTOR_ROOT,
+            Site::Gov => GOV_DIRECTOR_ROOT,
+        }
+    }
+}
+
+/// Read a TUF root override from disk, returning the bytes
+fn load_root(override_path: &std::path::Path) -> anyhow::Result<Vec<u8>> {
```
mellon85: do we want to offer this for agentless?

libdd-remote-config/src/fetch/fetcher.rs
paullegranddc: I am not boxing it because there is only going to be a single `ConfigFetcher` per process, so the memory gain of boxing seemed negligible

libdd-remote-config/src/fetch/agentless.rs
BaptisteFoy: Can we update the RC backend instead of hardcoding this? It will cause problems with configuration routine down the line.

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+
+/// Compute the backoff delay to wait before the next `fetch_config` attempt,
+/// given the number of consecutive failures observed so far.
+fn compute_backoff(consecutive_failures: u32) -> Option<Duration> {
+    match consecutive_failures {
+        0 => None,
+        1 => Some(jitter_secs(30, 60)),
+        2 => Some(jitter_secs(60, 120)),
+        3 => Some(jitter_secs(120, 240)),
+        _ => Some(Duration::from_secs(240)),
```
BaptisteFoy: nit: the agent stops at 120 iiuc, maybe we want to align?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+        tuf::verify::Verified<tuf::metadata::TargetsMetadata>,
+    >,
+) -> Option<&'a TargetDescription> {
+    // Direct hit on the top-level targets (Datadog's are empty, but be safe).
+    if let Some(d) = top.targets().get(path) {
+        return Some(d);
+    }
+
+    // Spec-style preorder walk over the (ordered) delegation list.
+    for delegation in top.delegations().roles() {
```
BaptisteFoy, line 982: Can we add a TODO to make it compliant with Go-TUF & recursive delegations?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+
+        // Fetch from the remote __unverified__ repo.
+        // This is fine because we compare the hash+len against TUF-validated metadata.
+        let mut read = self
+            .director_client
+            .remote_repo()
+            .fetch_target(target_path)
+            .await?;
+        let mut buf = Vec::new();
+        read.read_to_end(&mut buf).await?;
```
BaptisteFoy: let's read at most `target.length`?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+    length: u64,
+    version: u64,
+    all_hashes: Vec<(&'static tuf::crypto::HashAlgorithm, tuf::crypto::HashValue)>,
+    /// Lowercase hex of the first supported hash; used for cache-hit comparisons.
+    primary_hash: String,
+}
+
+const CUSTOM_METADATA_EXPIRY_PATH: &str = "expires";
+
+impl<'a> TrustedTarget<'a> {
```
mellon85, line 269:
```suggestion
/// A single TUF target that has passed metadata verification: it comes from the
/// director's trusted `targets` database (so its signature, snapshot binding and
/// version have already been checked by rust-tuf's `update()`), it is not
/// expired, and it declares at least one hash algorithm we can verify.
///
/// This is a lightweight, already-validated *view* over a
/// [`tuf::metadata::TargetDescription`] — it borrows the target `path` and copies
/// out only the fields we need downstream. It is produced by [`trusted_targets`]
/// via [`TrustedTarget::try_create`]; construction fails (dropping the target)
/// if the target is expired or has no supported hash.
///
/// Note: holding a `TrustedTarget` means the *metadata* is trusted, not that the
/// *content* has been fetched or checked — [`AgentlessFetcher::fetch_target`]
/// still re-verifies the downloaded bytes against `length` and `all_hashes`.
impl<'a> TrustedTarget<'a> {
```

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+    pub fn next_backoff(&self) -> Option<Duration> {
+        compute_backoff(self.consecutive_failures)
+    }
+
+    /// Rebuild both TUF clients from the embedded/override roots and discard all
+    /// derived state, restarting the fetcher as if freshly constructed. Called on
+    /// `apply()` failure so a partially-advanced trusted database cannot block
+    /// subsequent polls
+    ///
+    /// TODO(rust-tuf): rebuilding from the embedded root discards any newer root
```
mellon85: Retaining the roots is a very useful securit feature of the system (to avoid a leaked key from being used as a MitM)

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+    /// The [`ClientResponse`] contains the path, and metadata info of targets
+    /// that have been sent to the client.
+    ///
+    /// This separation between content and metadata is done for 2 reasons:
+    /// 1. Remote Config does not re-send target files that we have received in a previous fetch. So
+    ///    it should already be in cache.
+    /// 2. Currently we only have a single [`remoteconfig::Client`], but targets for mutliple
+    ///    clients can be fetched at once, and we need to do a M:N mapping of target files to
+    ///    clients, with the same target that can be used by mutliple client
+    pub(crate) async fn fetch_config<Storage: FileStorage>(
```
mellon85, line 484: Are you sure this is a good idea to have it public?
Won't it risk having multiple clients calling and cause very high traffic or if you have 2 parts of one tracer calling it to drop each other's configurations? What is expected to call this method?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+
+        let active_targets = match self.apply(&response, cache, prefetched_org_uuid).await {
+            Ok(t) => t,
+            Err(e) => {
+                self.consecutive_failures = self.consecutive_failures.saturating_add(1);
+                // On any `apply()` failure the trusted databases may have been advanced
+                // in place and incrementally, leaving them inconsistent with the
+                // versions we would report next poll.
+                // Reset both clients so the next poll restarts from a clean state
+                if let Err(reset_err) = self.reset().await {
```
mellon85: This is going to reset the whole in-memory repository if it fails to apply one update? In the agent it's applied atomically using on-disk transactions. I am not saying we must use Software Transactional memory, but maybe operate on a copy and atomically swap it?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+                )
+            })??;
+        Ok(response)
+    }
+
+    /// Update the TUF-clients state to add the new data fetched from the intake,
+    /// verify it with tuf-rust and verify target files against the TUF signed information.
+    ///
+    /// After this function returns Ok, the TUF client trusted database should be in-sync
+    /// with data fetched from the backend
```
mellon85: it doesn't state that the caller *MUST* call reset if the application fail (unless the update method is changed)

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+/// Compute the backoff delay to wait before the next `fetch_config` attempt,
+/// given the number of consecutive failures observed so far.
+fn compute_backoff(consecutive_failures: u32) -> Option<Duration> {
+    match consecutive_failures {
+        0 => None,
+        1 => Some(jitter_secs(30, 60)),
+        2 => Some(jitter_secs(60, 120)),
+        3 => Some(jitter_secs(120, 240)),
+        _ => Some(Duration::from_secs(240)),
+    }
```
mellon85:
```suggestion
    consecutive_failures = min(4, consecutive_failures);
    let base = 30*consecutive_failures
    return jitter_secs(base, base+30)
```

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+
+/// Random duration in `[min_secs, max_secs]`
+fn jitter_secs(min_secs: u64, max_secs: u64) -> Duration {
+    use rand::Rng;
+    Duration::from_secs(rand::thread_rng().gen_range(min_secs..=max_secs))
+}
+
+fn now_unix_milli_ts() -> u64 {
+    u64::try_from(
+        SystemTime::now()
```
mellon85: how come not chrono?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+        .ok_or_else(|| format_err!("config client has no trusted top-level targets"))?;
+    let trusted_delegations = config_client.database().trusted_delegations();
+
+    let director_targets = director_client
+        .database()
+        .trusted_targets()
+        .ok_or_else(|| format_err!("director client has no trusted targets"))?;
+
+    for (path, dir_desc) in director_targets.targets() {
+        let cfg_desc = lookup_config_target(path, top_config_targets, trusted_delegations)
```
mellon85, line 938: isn't this kind of an hidden quadratic cost (for in a for)?

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+        /// Returns the TUF path strings whose `(primary_hash, len)` already matches the cache.
+        pub(crate) fn is_cached_batch<'b>(
+            &self,
+            candidates: impl IntoIterator<Item = (&'b str, &'b str, u64)>,
+        ) -> hashbrown::HashSet<&'b str> {
+            let files = self.files.lock_or_panic();
+            candidates
+                .into_iter()
+                .filter_map(|(path, primary_hash, len)| {
+                    let parsed = RemoteConfigPath::try_parse(path).ok()?;
```
mellon85: this parses the paths again and drops them silently if they are unparseable. Shouldn't we paths remote config path in a struct and use that through the code at this point? that should allow us to avoid doing mulitple parsing steps of the same data

libdd-remote-config/src/fetch/agentless.rs
```diff
@@ -0,0 +1,1531 @@
...
+                    }
+                })
+                .collect()
+        }
+
+        pub(crate) fn store_batch(
+            &self,
+            targets: impl IntoIterator<Item = NewTarget>,
+        ) -> anyhow::Result<()> {
+            let mut files = self.files.lock_or_panic();
```
mellon85: This smell of a future risk of TOCTOU. I think this currently works because there's a single loop doing all the calls, but apply calls it mulitple times, aquiring and dropping the lock mulitple times in once single call

libdd-remote-config/src/fetch/agentless/integration_tests.rs
```diff
@@ -0,0 +1,908 @@
...
+    responses: Arc<Mutex<VecDeque<Vec<u8>>>>,
+    requests: Arc<Mutex<Vec<remoteconfig::LatestConfigsRequest>>>,
+    /// UUID returned for GET /api/v0.1/org. Some tests want to change it mid-run
+    /// (e.g. serve a mismatching UUID) so this is a Mutex, not an atomic snapshot.
+    org_uuid: Arc<Mutex<String>>,
+    /// Count of GET /api/v0.1/org requests observed, used to assert the
+    /// concurrent prefetch (poll 1) and the lazy re-fetch (post root-rotation).
+    org_requests: Arc<Mutex<u32>>,
+}
+
```
mellon85, line 68: I think some tests are missing (tampered targer file bytes, length mismatch, unauthorized target, ..).
As it's in memory it's basically impossible to happen, but either we write them because it's cheap to do and covers us from future changes if the assumption changes; or we document it throughly (and may still be missed)

libdd-remote-config/src/fetch/agentless.rs
mellon85: I left a comment somewhere else that I'd like to have a special version for it so it can be distinguished from other agents easily, like version `100.*.*`
Or we can add a `-agentless` to the version? Can't remember what we do support

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: Changed to use try_from

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: Yeah, the loop is actually shorter, not sure what came to me when I wrote this (because yes, I not claude wrote this somehow...)

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: changed to use try_from

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: done

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: Hmmm I don't think it really happens, apart from a backend error.
That said, this is how the backend checks targets, and if a target has all the same hashes signed, even if there are duplicates I'm not sure it should be an error

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: I mean... ideally we should put the version of the rust crate here, in a new field on the top level proto message.

This is sufficient though for testing.
It looks like the major needs to be `7`, so I used `7.1000.0-agentless` which should be disctintive enough

paullegranddc: It looks like the major needs to be 7, so I used 7.1000.0-agentless which should be disctintive enough

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: Adding it to rust-tuf

libdd-remote-config/src/fetch/fetcher.rs
paullegranddc: Made passing a wrong config an error. The check is no longer here, but done when building the options passed to the client

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: Changed the code following a suggestion from Dario

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: It is `O(target_files * delegated roles)`
go-tuf does the same when validating delegations
`getTargetFileMetas` iterates over targets and for each it calls `getTargetFileMetaDelegationPath` which iterates over delegations
https://github.com/DataDog/go-tuf/blob/master/client/delegations.go#L44-L63

In practice we could optimize matching by making a prefix map (or prefix tree) of delegated roles patterns, since in practice the delegations we have matches a path like `<exact string>/<rest, either exact or  wildcard>` but I'm not sure the number of delegated roles we have justifies the complexity

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: This was written before I had to add `chrono` as a dependency. Moved to use `chrono` since we include it anyway

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: Added the comment

libdd-remote-config/src/fetch/agentless.rs
mellon85: that implementation is not optimized and is largely unmaintained, I am not surprised it's hat inefficient. This tends to be an issue when you have many configurations. I can understand the argument "not now, let's see the real world performance"; but I needed to point it out as it might become an issue.

paullegranddc: Ok, I'll leave a comment mentioning the possible perf optimization

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: Ok, now we keep root state but drop everything else

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: The RC client on libdatadog works in a polling manner. It does not control the outside loop, and just exposes a refresh interval, at which users of the library should call the `fetch` method.
So yes it could be misused and called repeatedly. But I don't think it is something we can and should protect against at this level.


This `AgentlessFetcher::fetch_config` method is currently only used by the `SingleFetcher` type, which is used in languages that maintain a single `Client` (python, nodejs) per process. 
The languages with multiple clients (php mostly) are supposed to use the `MultiFetcher`, which fetches all clients at once and then dispatches configs to them.

I haven't hooked agentless in the `MultiFetcher`, and this comment should be resolved before we do it. 
But since PHP is not a language we target right now, it can wait a bit.

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: I added this because it was in the agent config. We probably don't actually need it so I removed it

libdd-remote-config/roots/staging/director_root.json
paullegranddc: updated the roots to be in sync with https://github.com/DataDog/rc-root-signing-ceremony/tree/main/data/exported/open-source

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: Done

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: Changed the code to parse only once. I wanted the agentless.rs module to be completely decoupled from the rest of the crate to extract it into a separate latter but this lead to weird things like this.

Now we parse the paths only once

libdd-remote-config/src/fetch/agentless.rs
paullegranddc: I did this because I could not hod the lock across await points, and fetching targets from rc required calling async methods even though we read from in-memory store because of the trait tuf-rust uses

libdd-remote-config/src/fetch/agentless/integration_tests.rs
paullegranddc: If you're thinking of on-disk database tampering, it's not possible to implement with the in-memory DB as tuf-rust does not let us mutate things

paullegranddc: Ok, added the tampered payload tests, and documented the agen tests that are anot applicable

libdd-remote-config/examples/remote_config_fetch.rs
```diff
@@ -8,16 +8,74 @@ use libdd_remote_config::file_change_tracker::{Change, FilePath};
...
-use std::time::Duration;
+use std::process::Command;
 use tokio::time::sleep;
 
 const RUNTIME_ID: &str = "23e76587-5ae1-410c-a05c-137cae600a10";
 const SERVICE: &str = "testservice";
 const ENV: &str = "testenv";
 const VERSION: &str = "1.2.3";
 
+fn get_hostname() -> String {
```
mellon85, line 24: there's really no other way in rust to get the hostname besides running a command? It might not be present in some situations (pod with just the go binary for instance).

libdd-remote-config/src/fetch/agentless/integration_tests.rs
mellon85: yeah the issue is that it could be tampered by an attacker MitM

libdd-remote-config/src/fetch/fetcher.rs
```diff
@@ -105,7 +121,7 @@ impl ConfigProductCapabilities {
 }
 
 pub struct ConfigFetcherState<S, C: HttpClientCapability> {
-    target_files_by_path: Mutex<HashMap<Arc<RemoteConfigPath>, StoredTargetFile<S>>>,
+    pub(crate) target_files_by_path: Mutex<HashMap<Arc<RemoteConfigPath>, StoredTargetFile<S>>>,
```
mellon85: this exposes an internal mutex lock as a public field in the crates. As products register and have callbacks, why should this be public? (and by extension StoredTargetFile).

Can't you move the mod cache physically into fetcher.rs or make this `pub(in crate::fetch)`

libdd-remote-config/examples/remote_config_fetch.rs
paullegranddc: This is just the example. I added this at the beginning instead of adding the `hostname` crate as a dependency, since this is not code that is shipped in the library.

In general we prefer if people using the library inject these parameters, although I'd understand if you preferred that the libdd-remote-config picks the hostname by default as this is more likely to be consistent across languages (for nodejs though, we cannot as we need to compile to WASM)

libdd-remote-config/src/fetch/fetcher.rs
paullegranddc: Made it`pub(in crate::fetch)`

libdd-remote-config/examples/remote_config_fetch.rs
mellon85: then can you please add a comment about it before people copy paste this

libdd-remote-config/src/fetch/agentless.rs
mellon85: We have limit which versions can connect (we blocked some broken initial implementations for instance), that's why it works with >= 7.42.xx (or similar). I'd avoid having something specific for this crate as it will collide with the agent

libdd-remote-config/src/fetch/agentless.rs
mellon85: That makes sense given the constraint from the tuf-rust trait, but I don't think "can't hold the lock across await" fully rules out fixing this. std::sync::MutexGuard can't cross an await point, but two alternatives should work:

1. Switch the lock to tokio::sync::Mutex.
2. If we want to avoid the async-lock overhead, snapshot the state we need under one short std lock acquisition before starting the async calls if possible, so there's no window where another task can mutate things between our reads.

paullegranddc: Forgot to answer to this comment.
I ended up adding a sync method on rust-tuf to fetch targets, and holding the lock across in-between checking for presence and storing targets in the cache

mellon85 (approved):

PR #2115, opened 2026-06-12
feat(data-pipeline): OTLP HTTP/protobuf trace export

bm1549:
# What does this PR do?

Adds OTLP HTTP/protobuf as a trace-export encoding alongside HTTP/JSON, selectable via the OTel-standard `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` (`http/json` default, `http/protobuf`). The generated `prost` OTLP types are the single intermediate representation: the mapper builds them directly from native spans, protobuf is `prost::encode_to_vec`, and JSON is a serde serializer over the same types.

# Motivation

libdatadog's OTLP trace export only spoke HTTP/JSON. SDKs that honor `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` need `http/protobuf` to match the OTel default and to talk to collectors that expect protobuf.

# Additional Notes

- Prost is the single IR: no parallel hand-rolled JSON model, no string round-trip. `encode_otlp_protobuf` is `encode_to_vec`; `encode_otlp_json` is a serde serializer (`json_serializer.rs`) emitting OTLP-spec JSON (hex ids, base64 bytes, int64-as-string, lowerCamelCase, proto3 defaults omitted).
- `OtlpProtocol` is `{HttpJson, HttpProtobuf}`; `grpc` is rejected at the parse boundary (`FromStr`) rather than carried as an unsupported variant. `content_type()`/`encode()` live on the type.
- Integrates with the OTLP metrics exporter already on main: the shared low-level sender takes the content-type per request, so traces use the configured protocol while metrics stay JSON.
- Carries through `otel_trace_semantics_enabled` (OTel attribute compatibility, #2091): when set, the prost mapper omits the DD-specific span attributes and promotes the error message to the OTLP `Status`.
- Span-link W3C trace flags are carried through to OTLP `Link.flags`.
- Benchmarks for the encoder hot paths, plus allocation tuning (pre-sized Vecs, allocation-free id/timestamp/int serialization).

# How to test the change?

- `cargo test -p libdd-trace-utils -p libdd-data-pipeline -p libdd-data-pipeline-ffi`, `cargo test --doc`, clippy, fmt, and `cargo ffi-test` pass. A parity test asserts the JSON and protobuf encodings carry the same span from the one prost IR; a protobuf round-trip test asserts the encoding is lossless.
- End-to-end through dd-trace-py (DataDog/dd-trace-py#18609): emitted protobuf-only OTLP traces through a local Agent to the backend. Wire was `application/x-protobuf` (HTTP 200), spans ingested with correct service/resource and a preserved 128-bit trace id.
- Benchmarks: `cargo bench -p libdd-trace-utils --bench main -- otlp/`.

BREAKING CHANGE: removes the previously public `libdd_trace_utils::otlp_encoder::json_types` module (the hand-rolled OTLP JSON model). OTLP encoding now builds prost-generated types as the single IR. libdatadog consumers pin by version, so they pick this up on the next release.


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -546,15 +548,28 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
...
-            TraceExporterError::Internal(InternalErrorKind::InvalidWorkerState(e.to_string()))
-        })?;
+        let body = match config.protocol {
+            OtlpProtocol::HttpJson => libdd_trace_utils::otlp_encoder::encode_otlp_json(&request)
+                .map_err(|e| {
+                error!("OTLP JSON serialization error: {e}");
+                TraceExporterError::Internal(InternalErrorKind::InvalidWorkerState(e.to_string()))
+            })?,
+            OtlpProtocol::HttpProtobuf => {
+                libdd_trace_utils::otlp_encoder::encode_otlp_protobuf(&request)
```
bengl: Where's the error handling here, like the with the other protocols?

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -6,20 +6,31 @@
 use http::HeaderMap;
 use std::time::Duration;
 
-/// OTLP trace export protocol. HTTP/JSON is currently supported.
+/// OTLP trace export protocol.
```
bengl: Seems like grpc should still be excluded here?

mabdinur (approved):
Faithful protobuf encoder with safe fallbacks (no panics), correct content-type selection, and gRPC is rejected at both build and send time. No breaking API changes and good JSON/protobuf parity tests. LGTM.

mabdinur (commented):
Approval stands, these are all nits / follow-up material, nothing blocking:

1. proto_convert.rs (perf): on the http/protobuf path, mapper.rs formats every ID to a hex string and every timestamp to a string, then this code reverses it (hex_to_bytes/parse_u64) and deep-clones the whole serde tree, per span per export. Native IDs are u128/u64 already. Reasonable as a documented v1, but worth a follow-up to build the prost types straight from the native span types and skip the round-trip.

2. otlp/config.rs: OtlpProtocol is public but not #[non_exhaustive]. Nothing breaks today since it's new, but without it any future variant becomes a breaking change for downstream exhaustive matches. Worth adding now while the enum is fresh.

3. otlp_encoder/mod.rs (scope question): confirming this is traces-only, which seems like the right first step with OtlpProtocol as the encoding axis. Are metrics/logs intentionally out of scope here, handled by separate signal pipelines rather than extending this path?

4. json_types.rs (maintainability): the OTLP schema now has two hand-maintained representations (serde json_types and prost) bridged by the From impls in proto_convert, with no shared source of truth, so a field added to one side can silently diverge. The cross-encoder parity test only compares name and spanId. Could we extend it to trace_id (base64 vs raw bytes), status, and at least one attribute so drift gets caught?

5. libdd-data-pipeline-ffi/src/trace_exporter.rs: the protocol string re-parse can't fail today since the setter already restricts the values, so an error here is a silent no-op. If the accepted set ever drifts from FromStr this would quietly drop the protocol. Consider expect/log, or rely on the builder validation.

libdd-trace-utils/src/otlp_encoder/proto_convert.rs
```diff
@@ -0,0 +1,403 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Converts the hand-rolled serde OTLP request (the JSON wire model) into the generated
```
ekump: This is going to be incredibly unperformant. The mapper is taking a `Span` and turning everything into a `String` for JSON. And then this is converting everything back for protobuf, which is also doing unnecessary extra string allocations. 

The problem is the mapper is conflating mapping to OTLP and serializing the wire format. I think it would be much better to refactor the mapper to use an intermediate representation and separate concerns. Using prost for the IR probably makes the most sense? It'll be much more performant to build prost types in the mapper, then serialize them directly to JSON via Serde.

libdd-trace-utils/src/otlp_encoder/proto_convert.rs
```diff
@@ -0,0 +1,403 @@
...
+        value: Some(ProtoAnyValue::from(&k.value)),
+        // `key_ref` and `entity_refs` (on Resource) are profiling-signal-only proto fields,
+        // unused for traces. Set explicitly to their zero defaults so the converter fails to
+        // compile if the proto shape changes (rather than silently misusing
+        // `..Default::default()`).
+        key_ref: 0,
+    }
+}
+
+impl From<&j::ExportTraceServiceRequest> for ProtoReq {
```
ekump: `From` taking the request by ref, with only a borrow you can't move the owned Strings out is leading to a ton of clones in `proto_convert`. Where this is used the serde request is a single-use intermediate that quickly gets dropped after. If you change this to `From<ExportTraceServiceRequest>` you can avoid a lot of costly allocations.

libdd-data-pipeline/src/otlp/exporter.rs
```diff
@@ -41,11 +43,19 @@ pub async fn send_otlp_traces_http<C: HttpClientCapability + SleepCapability>(
         ..Endpoint::default()
     };
 
+    // `Grpc` is rejected earlier in `send_otlp_traces_inner` and never reaches this function, so it
+    // is grouped with the JSON content-type here only to keep the match exhaustive.
+    let content_type = match config.protocol {
+        crate::otlp::config::OtlpProtocol::HttpProtobuf => {
+            libdd_common::header::APPLICATION_PROTOBUF
+        }
+        crate::otlp::config::OtlpProtocol::HttpJson | crate::otlp::config::OtlpProtocol::Grpc => {
```
ekump: I wouldn't recommended silently setting the content_type to JSON here when matching on GRPC. Just return an error. The function signature is already a `Result`.

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -32,7 +43,25 @@ pub struct OtlpTraceConfig {
     pub headers: HeaderMap,
     /// Request timeout.
     pub timeout: Duration,
-    /// Protocol (for future use; currently only HttpJson is supported).
-    #[allow(dead_code)]
-    pub(crate) protocol: OtlpProtocol,
+    /// OTLP export protocol (selects body encoding and content-type).
+    pub protocol: OtlpProtocol,
```
ekump, line 78: We're matching protocol several times in this PR and it's kind of awkward because `OtlpProtocol` contains a currently unsupported variant. Why not keep the user facing `OtlpProtocol` as is, and internally have something like ` enum OtlpWireProtocol { Json, Protobuf }`? It'll make the code much cleaner and when GRPC support is added you can get rid of `OTLPWireProtocol`. Or you could do:

```rust
  impl OtlpWireProtocol {                                                                                                                                                                                                                                                               
      fn content_type(&self) -> HeaderValue { /* 2 arms */ }                                                                                                                                                                                                                            
      fn encode(&self, req: &ExportTraceServiceRequest) -> Result<Vec<u8>> { /* 2 arms */ }                                                                                                                                                                                             
  } 
```
and encapsulate the correct behavior in the type.

libdd-trace-protobuf/build.rs
```diff
@@ -36,6 +36,15 @@ fn generate_protobuf() {
 
     config.out_dir(output_path.clone());
 
+    // The vendored OpenTelemetry trace protos carry doc comments with indented example blocks
```
ekump: This means no docs for the protobuf files will be generated at all. You can just update the Cargo.toml and add `doctest = false` to `[lib]` and get rid of this.

libdd-data-pipeline-ffi/src/trace_exporter.rs
ekump: This doc needs to be updated

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -498,6 +499,37 @@ pub unsafe extern "C" fn ddog_trace_exporter_config_set_otlp_endpoint(
     )
 }
 
+/// Sets the OTLP export protocol. Accepts the OTel-standard values `http/json` (default) or
```
ekump, line 503: The doc should probably make it clear that this function is inert if an otlp endpoint isn't set.

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -565,6 +597,13 @@ pub unsafe extern "C" fn ddog_trace_exporter_new(
 
             if let Some(ref url) = config.otlp_endpoint {
                 builder.set_otlp_endpoint(url);
+                if let Some(ref proto) = config.otlp_protocol {
+                    // The FFI setter only stores "http/json"/"http/protobuf", so this parse always
+                    // succeeds here; a parse failure just leaves the builder's default protocol.
+                    if let Ok(p) = proto.parse::<libdd_data_pipeline::otlp::OtlpProtocol>() {
```
ekump: Similar to other comments, this is dangerous if there is ever code drift. This silently drops a parse failure and falls back to the default protocol.

ekump (dismissed):

bengl (changes requested):

libdd-trace-utils/src/otlp_encoder/proto_convert.rs
bengl: This to me is the fundamental problem. It re-introduces the same problems I just spent 9 months trying to fix.

libdd-data-pipeline-ffi/src/trace_exporter.rs
bm1549: Updated in 1cb1dfb86

libdd-data-pipeline-ffi/src/trace_exporter.rs
bm1549: Updated in a790182c8

libdd-data-pipeline-ffi/src/trace_exporter.rs
bm1549: Updated in a790182c8

libdd-data-pipeline/src/otlp/config.rs
bm1549: Updated in 809914eb8

libdd-data-pipeline/src/otlp/exporter.rs
bm1549: Updated in 809914eb8

libdd-trace-utils/src/otlp_encoder/proto_convert.rs
bm1549: Updated in 55541ebff

libdd-trace-utils/src/otlp_encoder/proto_convert.rs
bm1549: Good call, thanks for pushing on this. Reworked it so prost is the single IR - the mapper builds prost types directly from native spans, then protobuf is `encode_to_vec` and JSON serializes the same types via serde. No more `String` round-trip. Updated in b21be0205. @bengl should cover your concern too.

libdd-trace-protobuf/build.rs
bm1549: Good call on keeping the docs. `doctest = false` on `[lib]` doesn't actually stop `cargo test --doc` from compiling the proto comments, so instead I dropped `disable_comments`, kept the docs, and fenced the one example as a `text` block so rustdoc renders it instead of running it. Updated in 18f28c557.

libdd-data-pipeline/src/trace_exporter/mod.rs
bm1549: Added in 809914eb8 - `wire.encode()` returns a `Result` now and the error is logged + propagated, same as the others.

libdd-data-pipeline/src/otlp/config.rs
bm1549: Kept `Grpc` on the user-facing `OtlpProtocol` (`#[non_exhaustive]`, room for later), but it's excluded from the wire path - `OtlpWireProtocol` is `Json`/`Protobuf` only, and grpc errors at build when an endpoint is set. 809914eb8 / 855f4c14e.

yannham: Why add `#[non_exhaustive]` if `Grpc` is here nontheless? Do you anticipate more Otlp protocol in the future?

Also, I tend to agree with Bryan: if it's not supported now, we shouldn't lie about it nor construct it. I'm not sure of where the parsing and error reporting is done, so maybe the naive way is actually painful for some reason, but by default I would have `OtlpProtocl` and `OtlpWireProtocol` be a unique type, and handle parse errors earlier/differently.

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -6,20 +6,77 @@
...
     Grpc,
 }
 
+impl std::str::FromStr for OtlpProtocol {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "http/json" => Ok(OtlpProtocol::HttpJson),
+            "http/protobuf" => Ok(OtlpProtocol::HttpProtobuf),
+            "grpc" => Ok(OtlpProtocol::Grpc),
```
yannham: (For example, having a specialized error here instead of actually constructing the variant?

libdd-data-pipeline/src/otlp/exporter.rs
```diff
@@ -41,11 +44,10 @@ pub async fn send_otlp_traces_http<C: HttpClientCapability + SleepCapability>(
...
     };
 
+    let content_type = wire.content_type();
+
     let mut headers = config.headers.clone();
-    headers.insert(
-        http::header::CONTENT_TYPE,
-        libdd_common::header::APPLICATION_JSON,
-    );
+    headers.insert(http::header::CONTENT_TYPE, content_type);
```
yannham:
```suggestion
let mut headers = config.headers.clone();
    headers.insert(http::header::CONTENT_TYPE, wire.content_type());
```

libdd-data-pipeline/src/otlp/mod.rs
```diff
@@ -22,9 +23,12 @@
...
+pub(crate) mod config;
+pub(crate) mod exporter;
 
-pub use config::OtlpTraceConfig;
-pub use exporter::send_otlp_traces_http;
+pub use config::{OtlpProtocol, OtlpTraceConfig};
+// Internal: the resolved wire encoding. Callers select via the user-facing `OtlpProtocol`; this
+// is derived from it at the send boundary and is not part of the crate's public API.
+pub(crate) use config::OtlpWireProtocol;
+pub(crate) use exporter::send_otlp_traces_http;
```
yannham: Nit: if we export already `config` and `exporter` as `pub(crate)`, then I'm not sure it's useful to re-export specific symbols. Either we should not export `config` and `exporter`, so that only those symbols are exported, or just let people use `::config::` or `::exporter::`.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -286,6 +287,17 @@ impl TraceExporterBuilder {
         self
     }
 
+    /// Selects the OTLP export protocol. Accepts `OtlpProtocol::HttpJson` (default) or
+    /// `OtlpProtocol::HttpProtobuf`. The host language resolves this from
+    /// `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` / `OTEL_EXPORTER_OTLP_PROTOCOL`.
```
yannham: (This will make link in the docs, and the nice consequence is also if some name changes, you get a doc warning, instead of pure text)

```suggestion
    /// Selects the OTLP export protocol. Accepts [OtlpProtocol::HttpJson] (default) or
    /// [OtlpProtocol::HttpProtobuf]. The host language resolves this from
    /// `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` / `OTEL_EXPORTER_OTLP_PROTOCOL`.
```

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -429,31 +441,47 @@ impl TraceExporterBuilder {
...
-                endpoint_url: url,
-                headers,
-                timeout: self
-                    .connection_timeout
-                    .map(Duration::from_millis)
-                    .unwrap_or(DEFAULT_OTLP_TIMEOUT),
-                protocol: OtlpProtocol::HttpJson,
-            }
-        });
+            None => None,
```
yannham: Why not keep the original `map` and use a pattern match like this? The `map` is more idiomatic.

libdd-data-pipeline/src/lib.rs
```diff
@@ -13,7 +13,7 @@
 
 pub mod agent_info;
 mod health_metrics;
-pub(crate) mod otlp;
+pub mod otlp;
```
yannham: Do you need to export all symbols here? We tend to be conservative by default and only make `pub` what needs to be, to avoid breaking backward compatibility with future refactorings.

libdd-trace-utils/benches/otlp_encoding.rs
```diff
@@ -0,0 +1,133 @@
...
+use criterion::{black_box, criterion_group, BatchSize, Criterion};
+use libdd_trace_utils::msgpack_decoder;
+use libdd_trace_utils::otlp_encoder::{
+    encode_otlp_json, encode_otlp_protobuf, map_traces_to_otlp, OtlpResourceInfo,
+};
+use serde_json::{json, Value};
+
+/// A realistic OTLP-bound span: a handful of string `meta` tags and a couple of numeric
+/// `metrics`, so the per-span attribute work (the dominant cost) is exercised. The chunk root
+/// carries `_dd.p.tid` (the 128-bit trace-id high bits, resolved once per chunk).
```
yannham: Nit: avoid cluttering details

```suggestion
/// A realistic OTLP-bound span: a handful of string `meta` tags and a couple of numeric
/// `metrics`, so the per-span attribute work (the dominant cost) is exercised.
```

libdd-trace-utils/benches/otlp_encoding.rs
```diff
@@ -0,0 +1,133 @@
...
+            "parent_id": parent_id,
+            "start": 1_544_712_660_000_000_000_i64 + i as i64,
+            "duration": 1_000_000,
+            "error": 0,
+            "meta": meta,
+            "metrics": { "_sampling_priority_v1": 1, "_dd.top_level": 1 },
+            "type": "web",
+        }));
+    }
+    spans
```
yannham: I don't know why claude insist on `for` loops in Rust, but iterators are more idiomatic:

```suggestion
    let root_span_id = 100_000_000_000 + (trace_id % 1_000_000);
    (0..num_spans).map(|i| {
        let span_id = root_span_id + i as u64;
        let is_root = i == 0;
        let parent_id = if is_root { 0 } else { root_span_id };
        let mut meta = json!({
            "http.method": "GET",
            "http.url": "https://example.com/api/v1/users/12345",
            "http.status_code": "200",
            "env": "production",
            "version": "1.2.3",
            "component": "net/http",
        });
        if is_root {
            meta["_dd.p.tid"] = json!("5b8efff798038103");
        }
        
        json!({
            "service": "bench-service",
            "name": "http.request",
            "resource": "GET /api/v1/users",
            "trace_id": trace_id,
            "span_id": span_id,
            "parent_id": parent_id,
            "start": 1_544_712_660_000_000_000_i64 + i as i64,
            "duration": 1_000_000,
            "error": 0,
            "meta": meta,
            "metrics": { "_sampling_priority_v1": 1, "_dd.top_level": 1 },
            "type": "web",
        })
    }).collect()
```

libdd-trace-utils/benches/otlp_encoding.rs
```diff
@@ -0,0 +1,133 @@
...
+pub fn otlp_encoding_benches(c: &mut Criterion) {
+    let info = resource_info();
+
+    // (chunks, spans_per_chunk): a single large trace (clean per-span signal) and many small
+    // traces (the typical agent payload shape). Both ~1000 spans total.
+    for (num_chunks, num_spans) in [(1usize, 1000usize), (100usize, 10usize)] {
+        let id = format!("{num_chunks}x{num_spans}");
+        let bytes = rmp_serde::to_vec(&generate_trace_chunks(num_chunks, num_spans))
+            .expect("serialize fixture");
+        // `spans` borrows `bytes`; both live for the rest of this iteration.
```
yannham: Clutter

```suggestion
```

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
```diff
@@ -0,0 +1,504 @@
...
+    serde_json::to_vec(&OtlpJson(req))
+}
+
+/// Serialize a `&[T]` by wrapping each element with `W`.
+fn seq<'a, T, W, S>(s: S, items: &'a [T], wrap: fn(&'a T) -> W) -> Result<S::Ok, S::Error>
+where
+    W: Serialize,
+    S: Serializer,
+{
+    s.collect_seq(items.iter().map(wrap))
```
yannham, line 44: (Because I wondered myself)

```suggestion
    // Serde's implementation uses the size hint from the iterator here.
    // There's no need to go through `serialize_seq` just to give the size.
    s.collect_seq(items.iter().map(wrap))
```

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
```diff
@@ -0,0 +1,504 @@
...
+        let mut buf = DecimalBuf::default();
+        if write!(buf, "{}", self.0).is_ok() {
+            return s.serialize_str(buf.as_str());
+        }
+        s.serialize_str(&self.0.to_string())
+    }
+}
+
+#[derive(Default)]
+struct DecimalBuf {
```
yannham, line 83: What is this structure and what is the semantics?

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
```diff
@@ -0,0 +1,504 @@
...
+        }
+        m.end()
+    }
+}
+
+struct ResourceJson<'a>(&'a Resource);
+impl Serialize for ResourceJson<'_> {
+    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+        let r = self.0;
+        let mut m = s.serialize_map(None)?;
```
yannham, line 141: We do know the length here, right? It's usually better to provide or it can cause undue additional allocation in the serializer.
```suggestion
        let length = !r.attributes.is_empty() as usize + (r.dropped_attribute_count != 0) as usize
        let mut m = s.serialize_map(Some(length))?;
```

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
```diff
@@ -0,0 +1,504 @@
...
+    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+        seq(s, self.0, ScopeSpansJson)
+    }
+}
+
+struct ScopeSpansJson<'a>(&'a ScopeSpans);
+impl Serialize for ScopeSpansJson<'_> {
+    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+        let ss = self.0;
+        let mut m = s.serialize_map(None)?;
```
yannham, line 164: Same: we should pre-compute the length whenever possible. I will stop flagging them but the are a lot of occurrences in this file.

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
```diff
@@ -0,0 +1,504 @@
...
+
+struct AnyValueSeq<'a>(&'a [AnyValue]);
+impl Serialize for AnyValueSeq<'_> {
+    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+        seq(s, self.0, AnyValueJson)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
yannham, line 390: I wonder if we could have some round-trip test here (JSON -> Span -> Json or Span -> Json -> Span). Also those could be amenable to property-based testing, but that's maybe another topic for another time 🙃

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -30,343 +270,179 @@ const MAX_ATTRIBUTES_PER_SPAN: usize = 128;
...
-            key: "deployment.environment.name".to_string(),
-            value: AnyValue::StringValue(resource_info.env.clone()),
-        });
-    }
-    if !resource_info.app_version.is_empty() {
-        attributes.push(KeyValue {
-            key: "service.version".to_string(),
-            value: AnyValue::StringValue(resource_info.app_version.clone()),
-        });
+fn push_str_attr(attrs: &mut Vec<ProtoKeyValue>, k: &str, v: &str) {
```
yannham: Nit: if `push_str_attr` is only used in `build_resource`, it could be moved inside

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -30,343 +270,179 @@ const MAX_ATTRIBUTES_PER_SPAN: usize = 128;
...
-    }
-    if !resource_info.runtime_id.is_empty() {
-        attributes.push(KeyValue {
-            key: "runtime-id".to_string(),
-            value: AnyValue::StringValue(resource_info.runtime_id.clone()),
-        });
+}
+
+fn build_resource(resource_info: &OtlpResourceInfo) -> ProtoResource {
+    let mut attributes: Vec<ProtoKeyValue> = Vec::new();
```
yannham:
```suggestion
    let mut attributes = Vec::new();
```

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -30,343 +270,179 @@ const MAX_ATTRIBUTES_PER_SPAN: usize = 128;
...
-        }
-    };
-    // Set flags from sampling priority: 1 = sampled/keep, 0 = dropped.
+    let (attributes, dropped_attributes_count) = collect_span_attributes(span, resource_service);
+    let (code, message) = span_status(span);
     let flags = span
         .metrics
         .get("_sampling_priority_v1")
-        .map(|p| if *p >= 1.0 { 1u32 } else { 0u32 });
+        .map(|p| if *p >= 1.0 { 1u32 } else { 0u32 })
```
yannham:
```suggestion
        .map(|p| (*p >= 1.0) as u32)
```

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -30,343 +270,179 @@ const MAX_ATTRIBUTES_PER_SPAN: usize = 128;
...
+        span_id: span.span_id.to_be_bytes().to_vec(),
         trace_state,
+        parent_span_id,
+        flags,
         name: span.resource.borrow().to_string(),
-        kind,
-        start_time_unix_nano,
-        end_time_unix_nano,
+        kind: span_kind(span),
+        // Clamp negatives to 0 — matches the prior parse_u64 zero-fallback on negative input.
```
yannham: If `parse_u64` has been removed as part of this PR, I think it's more confusing than helpful to mention it for future readers of this code.
```suggestion
```

yannham (commented):
[EDIT: you can ignore, see my later comment below].

> Following @ekump 's comment, I also wonder why all the helpers in the mapper take spans by reference, while the main mapping functions actually has ownership. I feel like we could avoid most of the string clones when we're converting native spans to the IR, if we did this transformation destructively (which is fine, we have ownership)? Though I might be missing some details.

yannham:
Ok, I'll take back the comment on owning spans. Spans text is an abstract `SpanText` type which is generally _not_ a Rust string, so there's not much we can do here when converting to the prost IR in general.

libdd-data-pipeline/src/otlp/config.rs
ekump: @yannham `OltpWireProtocol` was [my suggestion](https://github.com/DataDog/libdatadog/pull/2115#discussion_r3438438536). My rationale was that grpc is supported in the OTEL spec, just not supported in libdatadog yet, so having it as a variant in the public enum was reasonable. `OtlpWireProtocol` as the internal enum makes the code cleaner in this scenario. I don't have any objection to dropping grpc from the public enum if people think that's the better approach.

ekump (commented):
dismissing my request changes

libdd-data-pipeline/src/otlp/config.rs
bm1549: Dropped `Grpc` from the public enum entirely. `OtlpProtocol` is now just `{HttpJson, HttpProtobuf}`, `grpc` is rejected at the parse boundary (`FromStr`), and `#[non_exhaustive]` is gone. `0b430a5cc`

libdd-data-pipeline/src/otlp/config.rs
bm1549: Heads up - this went further: since `Grpc` got dropped from the public enum, the protocol and wire-encoding sets are identical now, so I removed `OtlpWireProtocol` and folded `content_type()`/`encode()` onto `OtlpProtocol`. `0b430a5cc`

libdd-data-pipeline/src/otlp/config.rs
bm1549: Done in `0b430a5cc`

libdd-data-pipeline/src/otlp/exporter.rs
bm1549: Done in `0b430a5cc`

libdd-data-pipeline/src/trace_exporter/builder.rs
bm1549: Done in `0b430a5cc`

libdd-data-pipeline/src/trace_exporter/builder.rs
bm1549: Done in `0b430a5cc`

libdd-trace-utils/benches/otlp_encoding.rs
bm1549: Done in `0b430a5cc`

libdd-trace-utils/benches/otlp_encoding.rs
bm1549: Done in `0b430a5cc`

libdd-trace-utils/benches/otlp_encoding.rs
bm1549: Done in `0b430a5cc`

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
bm1549: Done in `0b430a5cc`

libdd-trace-utils/src/otlp_encoder/mapper.rs
bm1549: Done in `0b430a5cc`

libdd-trace-utils/src/otlp_encoder/mapper.rs
bm1549: Done in `0b430a5cc`

libdd-trace-utils/src/otlp_encoder/mapper.rs
bm1549: Done in `0b430a5cc`

libdd-trace-utils/src/otlp_encoder/mapper.rs
bm1549: Done in `0b430a5cc`

libdd-data-pipeline/src/otlp/mod.rs
bm1549: `otlp` is `pub(crate)`, so nothing inside escapes the crate - `OtlpProtocol` is the one public symbol, re-exported at the crate root because the FFI needs to name it. `config`/`exporter`/`metrics` stay crate-internal. `0b430a5cc`

libdd-data-pipeline/src/lib.rs
bm1549: Trimmed - `otlp` is `pub(crate)` again, with only `OtlpProtocol` re-exported at the crate root for the FFI. `0b430a5cc`

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
bm1549: It's a fixed-capacity stack buffer (`[u8; 24]`) that impls `fmt::Write`, so `NumStr` can format a u64/i64 to a decimal string with no heap allocation (max 20 digits, 24 bytes always fits). Added a doc comment. `0b430a5cc`

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
bm1549: I dug into this - serde_json's `serialize_map`/`serialize_seq` don't size any allocation from the length hint, they only special-case `Some(0)`, so `Some(n)` and `None` produce identical output. I benchmarked the real lever (pre-sizing the output buffer): `encoded_len()`-based sizing regressed ~3% (the extra traversal costs more than the saved reallocs), and a cheap span-count estimate was within noise. So I kept plain `to_vec` and recorded that in a comment. Happy to revisit if you feel strongly.

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
bm1549: Same as r3443141026 - serde_json ignores the map/seq length hint except for `Some(0)`, and pre-sizing the output buffer benchmarked as a wash/regression, so I left these as plain `to_vec`.

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
bm1549: Good idea - added a round-trip test. A true JSON->IR round-trip would need a deserializer mirroring the serializer (the prost types don't derive `Deserialize`), so it's a protobuf round-trip (`prost -> bytes -> prost`, asserts equal) plus the existing `json_and_protobuf_carry_same_span` parity guard. Left property-based testing as the follow-up you mentioned. `0b430a5cc`

libdd-trace-utils/src/otlp_encoder/json_serializer.rs
yannham: Ah, I was thinking about msgpack serialization, but it makes sense JSON can be done in a streaming mode... Thanks for actual checking it!

I mean in general _some_ serializers might take advantage of the size hint, but if you're only going to use this one for JSON, then it's reasonable as it is :+1:

yannham (approved):
All concerns addressed for me 👍 

In the long run we might want to look at deserializing native span directly to protobuf without going through prost and the IR at some point, but this is obviously way out of the scope of this PR and OTLP in general.

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -413,72 +251,360 @@ fn map_attributes<T: TraceData>(
...
+        trace_id: trace_id_128.to_be_bytes().to_vec(),
+        span_id: span.span_id.to_be_bytes().to_vec(),
+        trace_state,
+        parent_span_id,
+        flags,
+        name: span.resource.borrow().to_string(),
+        kind: span_kind(span),
+        // OTLP timestamps are unsigned; clamp negatives to 0 so the `as u64` cast can't wrap.
+        start_time_unix_nano: span.start.max(0) as u64,
+        end_time_unix_nano: (span.start + span.duration).max(0) as u64,
```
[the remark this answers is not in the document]
bm1549: This is quite a stretch... Yes, it can possibly overflow, but the max of an i64 is in 236 _years_. I'd opt to decline this one

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -490,25 +616,19 @@ mod tests {
             ..Default::default()
         };
         span.meta.insert(
-            libdd_tinybytes::BytesString::from_static("error.message"),
+            libdd_tinybytes::BytesString::from_static("error.msg"),
             libdd_tinybytes::BytesString::from_static("something broke"),
         );
```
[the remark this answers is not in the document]
bm1549: Although it's deprecated, `error.message` is the proper semantic
https://opentelemetry.io/docs/specs/semconv/registry/attributes/error/#deprecated-error-attributes

bengl (approved):
This is fine.

PR #2117, opened 2026-06-14
feat(datadog-ffe): server-side EVP flagevaluation payload + bincode-safe sidecar delivery

leoromanovsky:
## Motivation

PHP relies on libdatadog and the sidecar for server-side flag-evaluation delivery, so the shared sidecar path must preserve the same worker-facing EVP contract as the direct SDK implementations. This contribution gives PHP a reusable, backend-verifiable flagevaluation transport without adding a one-off EVP writer in the PHP extension, which keeps cross-SDK adoption smoother for SDK owners and APM review.

## Changes

- Keeps the sidecar action enum bincode-compatible by appending `FfeFlagEvaluationBatch` after existing variants.
- Keeps flagevaluation IPC structs bincode-safe; outbound placeholder stripping happens only in the sidecar flusher before the EVP POST.
- Shares the common Agent EVP proxy endpoint/header/request path between FFE exposures and flagevaluation delivery.
- Carries `context.evaluation` over IPC as a JSON-object string and re-expands it to an object for outbound EVP JSON.
- Preserves valid context values such as `false`, `""`, `{}`, and `[]` in user context.
- Adds the flagevaluation C ABI path without a `reason` argument.
- Supports optional real `targeting_rule.key`.
- Retries reliable enqueue using retained encoded bytes so reconnect retry does not consume the payload.
- Coalesces PHP flagevaluation batches in the sidecar by schema-visible dimensions and folds overflow into degraded buckets.
- Splits sidecar EVP POST bodies by encoded uncompressed JSON bytes under the 5 MiB EVP limit, keeping the existing 512-event chunking as a secondary guard.
- Degrades an oversized full-tier row by omitting `targeting_key` and `context`; drops and logs only if the degraded row still cannot fit.

## Decisions

- OpenFeature `reason` is not part of the native ABI, sidecar event, outbound EVP payload, or aggregation contract.
- Bincode wire compatibility is part of the PHP product path because PHP reaches EVP through worker-to-sidecar IPC before HTTP.
- FFE exposures and flagevaluation share Agent EVP proxy transport mechanics; schema-specific payload shaping remains in each flusher because exposure deduplication and flagevaluation coalescing have different contracts.
- The flagevaluation sidecar flusher owns worker-facing JSON cleanup so shared wire structs remain plain enough for positional bincode serialization.
- Degraded events omit targeting key and context while retaining visible flag, variant, allocation, error, runtime-default, and targeting-rule fields.
- Payload splitting happens after sidecar JSON cleanup because that is where the final uncompressed POST body exists; count-based chunking alone is not sufficient for the EVP 5 MiB limit.
- PHP consumes this flagevaluation endpoint through the C ABI; the companion PHP draft points its libdatadog submodule at this PR head.

```mermaid
flowchart TD
  A[sidecar receives/coalesces batches] --> B[clean and encode event JSON]
  B --> C{candidate POST <= 5 MiB?}
  C -- yes --> D[POST through Agent EVP proxy]
  C -- no --> E{single full row can degrade?}
  E -- yes --> F[omit targeting_key and context]
  F --> B
  E -- no --> G[drop and log/count]
```

### Dogfooding App

- Companion PHP draft: https://github.com/DataDog/dd-trace-php/pull/3984
- `ffe-dogfooding` `app-php7` and `app-php8-openfeature` were run with local PHP artifacts using this libdatadog sidecar path.
- Dogfooding ran with the real-backend EVP override: Agent `DD_EVP_PROXY_CONFIG_ADDITIONAL_ENDPOINTS={}`, `DD_SKIP_SSL_VALIDATION=false`, and `DD_REMOTE_CONFIGURATION_NO_TLS_VALIDATION=false`.
- Sent 12 identical evaluations per targeting key for `ffe-dogfooding-string-flag`:
  - `libdd-php7-batch-20260623T020845Z-alpha`
  - `libdd-php7-batch-20260623T020845Z-bravo`
  - `libdd-php8of-batch-20260623T020845Z-alpha`
  - `libdd-php8of-batch-20260623T020845Z-bravo`
- App-side responses returned `variant_2`.

### System Tests

- Companion draft PR: https://github.com/DataDog/system-tests/pull/7187

### Staging End-To-End

- Retriever staging query used `--datacenter us1.staging.dog --customer-auth=skip` against the `flagevaluation` track for the exact targeting keys above.
- Retriever returned one backend row per targeting key, each with `flag.key=ffe-dogfooding-string-flag`, `variant.key=variant_2`, `allocation.key=allocation-override-392dd7c149f8`, and `evaluation_count=12`:
  - `libdd-php7-batch-20260623T020845Z-alpha`: `first_evaluation=1782180525878`, `last_evaluation=1782180525985`, `timestamp=1782180525985`, `evaluation_count=12`
  - `libdd-php7-batch-20260623T020845Z-bravo`: `first_evaluation=1782180525997`, `last_evaluation=1782180526103`, `timestamp=1782180526103`, `evaluation_count=12`
  - `libdd-php8of-batch-20260623T020845Z-alpha`: `first_evaluation=1782180526245`, `last_evaluation=1782180526344`, `timestamp=1782180526344`, `evaluation_count=12`
  - `libdd-php8of-batch-20260623T020845Z-bravo`: `first_evaluation=1782180526356`, `last_evaluation=1782180526451`, `timestamp=1782180526451`, `evaluation_count=12`


datadog-sidecar-ffi/src/lib.rs
```diff
@@ -1243,6 +1264,78 @@ fn ddog_sidecar_send_ffe_exposure_batch_impl(
...
+    let flag_evaluations = try_c!(flag_evaluations
+        .iter()
+        .map(|event| ffe_flag_evaluation_from_ffi(event, &context.service))
+        .collect::<Result<Vec<_>, _>>());
+
+    if flag_evaluations.is_empty() {
+        return MaybeError::None;
+    }
+
+    try_c!(blocking::enqueue_actions_reliable(
```
bwoebi: Let's not do this.
I know, we want it to happen with high reliability, but this is actually the last-chance backpressure mechanism.
Low priority packets (like telemetry) get dropped at a lower fullness of the backpressure queue.
When regular enqueue_actions messages get dropped, this includes important messages like Traces as well. FFE is _not_ more important than a trace and doesn't get to claim privilege here. Simply using enqueue_actions makes it the same priority than traces, which is the right level. 
```suggestion
    try_c!(blocking::enqueue_actions(
```

bwoebi:
The added rust files do contain quite a bit of logic, to submit FFE events. Why not move them to the ffe-crate directly (behind the feature flags, that is)?
IIRC you were talking about reusing some of the logic in Python eventually.
Exactly these are the things which should be commonly available, non-sidecar specific.

If you ever want to enable proper agentless submission - our current approach is to have all I/O to the backend directly go through libdatadog (which then routes to agent or backend, depending on API key presence in `Endpoint`).

The receiving of the sidecar channel messages and dispatching to the coalescer and senders is obviously sidecar-specific. But the other stuff is generic.

leoromanovsky:
Updated dataflow after the review-feedback refactor:

```mermaid
flowchart TB
  FFI[datadog-sidecar-ffi<br/>build FfeFlagEvaluationBatch]
  Queue[sidecar action queue]
  Adapter[datadog-sidecar<br/>flush adapter]
  Shared[datadog-ffe<br/>EVP flag-evaluation writer]
  Agent[Datadog Agent<br/>EVP proxy]
  Worker[flageval-worker]
  Future[future Python/Ruby bindings]

  FFI --> Queue --> Adapter --> Shared --> Agent --> Worker
  Shared --> Adapter
  Future -. later .-> Shared
```

Ownership:

- `datadog-ffe`: reusable EVP flag-evaluation writer code: coalescing, degrade/drop accounting, JSON encoding, placeholder stripping, payload splitting, Agent EVP proxy endpoint derivation, POST construction, and fire-and-forget send handling.
- `datadog-sidecar`: sidecar-only adapter: queue consumption, flush scheduling, `NativeCapabilities` wiring, and exporting writer stats through sidecar self-telemetry.
- `datadog-sidecar-ffi`: C input conversion and normal sidecar queue enqueueing.

This is the EVP flag-evaluation event path, separate from the OTel `feature_flag.evaluations` metrics path. Python/Ruby bindings are not part of this PR, but the common code is now organized so they can bind to `datadog-ffe` later instead of copying sidecar-specific logic.

datadog-sidecar/src/self_telemetry.rs
```diff
@@ -282,6 +329,27 @@ impl SelfTelemetry {
                 true,
                 MetricNamespace::Tracers,
             ),
+            flagevaluation_rows_dropped: worker.register_metric_context(
+                FLAG_EVALUATION_ROWS_DROPPED_METRIC.to_string(),
+                vec![],
+                MetricType::Count,
+                true,
+                MetricNamespace::Tracers,
+            ),
```
leoromanovsky: Note from Sameeran during live review: validate this against Ruby & Python as well, show evidence.

libdd-ffe/src/telemetry/flagevaluation.rs
```diff
@@ -0,0 +1,1525 @@
...
+// ── Context pruning ──────────────────────────────────────────────────────────
+
+/// Prune evaluation context attributes to satisfy the flagevaluation bounds:
+/// - At most `MAX_CONTEXT_FIELDS` (256) entries are kept.
+/// - String values longer than `MAX_FIELD_LENGTH` (256 chars) are **skipped** (not truncated) to
+///   avoid partial-data misattribution.
+/// - Non-string values (bool, number, null) are kept regardless of their display length.
+/// - Keys are iterated in sorted order for deterministic canonical-key stability; the returned
+///   `BTreeMap` preserves that order.
+pub fn prune_context(
```
sameerank, line 219: `prune_context` appears to be referenced only by tests inside libdatadog, so `MAX_CONTEXT_FIELDS` / `MAX_FIELD_LENGTH` are not enforced by the FFI/sidecar path. The companion PHP PR https://github.com/DataDog/dd-trace-php/pull/3984 does prune upstream in `pruned_context_map`, but this FFI entrypoint generally trusts any caller-provided `evaluation_context_json`.

Could we enforce pruning here after parsing the JSON object, before re-serializing it?

https://github.com/DataDog/libdatadog/blob/a57a94aeb724a7ca50b48af631200c51b86b910e/datadog-sidecar-ffi/src/lib.rs#L1491-L1504

Or alternatively, if the intended contract is that native callers must prune before calling into this API, it would be helpful to document that near the FFI function/type so consumers don't assume libdatadog applies the bounds automatically?

datadog-sidecar/src/service/ffe_evp_proxy.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    let mut parts = base.url.clone().into_parts();
+    parts.path_and_query = Some(PathAndQuery::from_static(path));
+    let url = http::Uri::from_parts(parts).ok()?;
+    Some(Endpoint {
+        url,
+        ..base.clone()
+    })
+}
```
sameerank, line 38: Some duplication with https://github.com/DataDog/libdatadog/blob/7f642c364f3357b7d4ec2ccb9bece8c335823b62/datadog-ffe/src/telemetry/flagevaluation/sender.rs#L46-L58

datadog-sidecar-ffi/src/lib.rs
```diff
@@ -1344,6 +1437,37 @@ fn ffe_exposure_from_ffi(exposure: &FfeExposure<'_>) -> Result<SidecarFfeExposur
     })
 }
 
+fn ffe_flag_evaluation_from_ffi(
+    event: &FfeFlagEvaluation<'_>,
+    service: &str,
+) -> Result<SidecarFfeFlagEvaluationEvent, String> {
+    let evaluation = optional_json_object_string(event.evaluation_context_json)?;
```
sameerank, line 1448: I forget, is user evaluation context optional? And are we okay with `context.dd.service` attribution being conditional on that?

datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
```diff
@@ -0,0 +1,246 @@
...
+};
+use libdd_capabilities_impl::NativeCapabilities;
+use libdd_common::Endpoint;
+use std::time::Duration;
+
+const USER_AGENT: &str = concat!("ddtrace-sidecar/", env!("CARGO_PKG_VERSION"));
+const COALESCE_DELAY: Duration = Duration::from_millis(250);
+
+pub(crate) const FLAG_EVALUATION_ROWS_DROPPED_METRIC: &str = "flagevaluation.rows.dropped";
+pub(crate) const FLAG_EVALUATION_ROWS_DEGRADED_METRIC: &str = "flagevaluation.rows.degraded";
```
sameerank: The metric unit is a bit ambiguous here. `rows` and `buckets` seem to refer to the aggregated event rows, but these counters are incremented by `evaluation_count`, not by one per dropped/degraded bucket.

For example, `flagevaluation.rows.dropped = 12` sounds like 12 aggregated buckets/rows were dropped, when it could mean 1 dropped bucket that represented 12 evaluations.

Could we rename these to something like `flagevaluation.evaluations.dropped` / `flagevaluation.evaluations.degraded` (or `flagevaluation.evaluation_count.*`) so the metric name matches the unit being reported? If the intended unit is rows/buckets, then the increments should probably be by row count instead.

datadog-ffe/src/telemetry/flagevaluation.rs
```diff
@@ -0,0 +1,1525 @@
...
+    }
+}
+
+fn strip_placeholders(value: &mut serde_json::Value) {
+    match value {
+        serde_json::Value::Object(map) => {
+            for (key, child) in map.iter_mut() {
+                if key != "evaluation" {
+                    strip_placeholders(child);
+                }
```
sameerank: Nit: in `build_event_payload`, `strip_placeholders` skips recursive cleanup for any object key named `evaluation`, regardless of path.

However, the exemption is broader than the schema-specific field it is protecting. If another `evaluation` field is added elsewhere later, it would also bypass placeholder cleanup. Consider making the exemption path-aware so only `context.evaluation` is preserved.

sameerank (approved):
Great work! Nothing blocking, but consider the suggested improvements if you agree with any of them

libdd-ffe/src/telemetry/flagevaluation/sender.rs
```diff
@@ -0,0 +1,405 @@
...
+                let body_preview = truncate(resp.body().as_ref(), 256);
+                log::warn!("ffe flagevaluation sender non-2xx response {status}: {body_preview}");
+            } else {
+                log::debug!(
+                    "ffe flagevaluation sender sent flag evaluation batch, status={status}"
+                );
+            }
+        }
+        Err(e) => {
+            log::debug!("ffe flagevaluation sender request failed: {e:?}");
```
bwoebi, line 150: debug level only (also other errors in this function)?

datadog-ffe/src/telemetry/flagevaluation.rs
```diff
@@ -0,0 +1,1525 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Reusable EVP flagevaluation payload, coalescing, serialization, and sender
+//! primitives for the `flageval-worker` ingestion schema.
+//!
+//! Crate-naming note: this workspace uses `libdd-remote-config` (not
+//! `datadog-remote-config`) for the remote config crate. Downstream consumers
+//! (e.g. `dd-trace-php`) must use `libdd-remote-config` in any import paths.
+//!
```
bwoebi: What does that comment want to tell us?

libdd-ffe/src/telemetry/flagevaluation.rs
```diff
@@ -0,0 +1,1525 @@
...
+}
+
+impl<D> Default for FlagEvaluationEvpCoalescer<D> {
+    fn default() -> Self {
+        Self {
+            state: Arc::new(Mutex::new(CoalescerState::default())),
+            writer_stats: Arc::new(FlagEvaluationEvpWriterCounters::default()),
+        }
+    }
+}
```
bwoebi, line 413: Why no #[derive] here? Is this to work around `D: Default + Clone`? But the impl anyway requires it, so you don't gain much.

libdd-ffe/src/telemetry/flagevaluation.rs
```diff
@@ -0,0 +1,1525 @@
...
+            destinations: HashMap::new(),
+            flush_running: false,
+            pending_bucket_count: 0,
+            full_bucket_count: 0,
+            full_bucket_count_by_flag: HashMap::new(),
+            degraded_bucket_count: 0,
+            dropped_overflow: 0,
+        }
+    }
+}
```
bwoebi, line 384: Why no #[derive] here? Is this to work around adding `D: Default`?

datadog-ffe/src/telemetry/flagevaluation.rs
```diff
@@ -0,0 +1,1525 @@
...
+    push_payload(
+        &mut result.payloads,
+        &payload_prefix,
+        payload_suffix,
+        &mut current_events,
+    );
+    if result.payloads.len() > 1 {
+        result.payload_splits = result
+            .payload_splits
+            .max(result.payloads.len().saturating_sub(1) as u64);
```
bwoebi: I don't get this logic. Why not simply `result.payload_splits = result.payloads.len().saturating_sub(1) as u64;` and remove the manual counting in the loop.

libdd-ffe/src/telemetry/flagevaluation.rs
```diff
@@ -0,0 +1,1525 @@
...
+            encoded_event = degraded_encoded_event;
+            event_size = degraded_event_size;
+            result.degraded_oversized_rows = result
+                .degraded_oversized_rows
+                .saturating_add(degraded_event.evaluation_count);
+        }
+
+        let separator_size = usize::from(!current_events.is_empty());
+        if current_size + separator_size + event_size > payload_size_limit
+            && !current_events.is_empty()
```
bwoebi, line 665: that's rather an assertion, than an actual condition? If that's empty here _and_ the size check misses, it means that the counting logic was wrong?

datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
```diff
@@ -0,0 +1,246 @@
...
+        let destination_key = DestinationKey::new(endpoint, &batch.context);
+        if self.inner.enqueue(destination_key, batch) {
+            let coalescer = self.clone();
+            tokio::spawn(async move {
+                coalescer.flush_loop(client).await;
+            });
+        }
+    }
+
+    pub(crate) async fn flush_now(&self, client: NativeCapabilities) {
```
bwoebi, line 74: Calling flush_now() does not ensure that a concurrent flush_loop (called by chance right before the flush_now call) has finished as well.
I.e. the mere awaiting of flush_now() is not enough of a barrier to guarantee that _all_ previously sent evaluations have been submitted, leaving open a route to flakiness in tests relying on flush() being a barrier and inadvertent data loss in the final shutdown of a container.
In particular when flush_now() terminates immediately because there are no pending payloads, it's prone to data loss.

datadog-sidecar/src/service/mod.rs
```diff
@@ -100,4 +107,13 @@ pub enum SidecarAction {
...
         metrics: Vec<FfeEvaluationMetric>,
     },
+    /// Structured FFE flag evaluation batch for the EVP flagevaluation track.
+    /// The sidecar serializes and POSTs the batch to
+    /// `/evp_proxy/v2/api/v2/flagevaluation` (fire-and-forget). PHP (EMIT-07)
+    /// drives the two-tier aggregation upstream and dispatches via this action.
+    ///
+    /// Keep this appended after pre-existing variants: this enum crosses the
+    /// bincode sidecar IPC boundary, so inserting a variant before existing
+    /// variants changes their wire ordinals.
```
bwoebi, line 117: Irrelevant.

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -1083,6 +1107,11 @@ impl SidecarInterface for ConnectionSidecarHandler {
     }
 
     async fn flush(&self, _peer: PeerCredentials, options: SidecarFlushOptions) {
+        self.server
```
bwoebi: Please add a flag to SidecarFlushOptions.

Also consider sending it in parallel with traces and stats. There's no dependency and can be trivially parallelized. (unlike telemetry which may be updated by stats flushing and thus needs to be queued after.)

datadog-sidecar/src/self_telemetry.rs
leoromanovsky: Closing this scope note per follow-up: Ruby/Python validation is intentionally out of scope for this libdatadog update and was not run. Java consistency was checked for the flag-evaluation telemetry unit naming.

datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
leoromanovsky: Thank you. Addressed in https://github.com/DataDog/libdatadog/pull/2117/commits/2b6779aef46d8fe428f18e533a4a2f968c694211: `flush_now` now runs behind the same flush mutex as the background flush loop, so the explicit flush acts as a barrier for queued flag-evaluation batches.

datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
leoromanovsky: Good thinking https://github.com/DataDog/libdatadog/pull/2117/commits/3f59d64a92d3bdae42fff41ffcb8b548e093788c

libdd-ffe/src/telemetry/flagevaluation.rs
leoromanovsky: Kept the manual `Default` impl and documented why here: https://github.com/DataDog/libdatadog/pull/2117/commits/2972e30086534ea8cfc35c7bee99c3184ee9b158. The actual manual impl was introduced in https://github.com/DataDog/libdatadog/pull/2117/commits/3ea3104e2ffd971db67a32b1229af4cdea10aac6 to avoid an unnecessary `D: Default` bound.

datadog-sidecar-ffi/src/lib.rs
leoromanovsky: Yes, user evaluation context is optional, and `context.dd.service` no longer depends on it. https://github.com/DataDog/libdatadog/pull/2117/commits/789848593d4932205400424915590d0a4e673b73 keeps service attribution when `evaluation_context_json` is empty/omitted, as long as the sidecar context has a non-empty service.

libdd-ffe/src/telemetry/flagevaluation.rs
leoromanovsky: Thank you for the thoughtful comment.

Accepted; libdatadog now enforces the FFI bounds before re-serializing `evaluation_context_json`: https://github.com/DataDog/libdatadog/pull/2117/commits/789848593d4932205400424915590d0a4e673b73. That keeps upstream PHP pruning as a helpful pre-bound, but the FFI entrypoint no longer relies on every caller doing it correctly.

Flattening and cycle detection still stay upstream with SDK callers, because libdatadog receives an already-materialized JSON object over FFI; this change only defensively prunes/bounds that JSON before EVP encoding.

Overall I'm not that happy with the API - I ideally want to find a stronger way to disallow passing in junk. Do you think the current shape is good enough to accomplish our goals?

datadog-sidecar-ffi/src/lib.rs
leoromanovsky: Done in https://github.com/DataDog/libdatadog/pull/2117/commits/8dd04181d4e679e6ec9eb5c98bcd45531608a2fa: flag-evaluation enqueue now uses the standard `enqueue_actions` path rather than the reliable/no-load-shed path.

datadog-sidecar/src/service/ffe_evp_proxy.rs
leoromanovsky: Partially deduplicated in https://github.com/DataDog/libdatadog/pull/2117/commits/e16802bd06fddb3f3bc97fcdb61c11ef32115c15 by sharing the sidecar FFE EVP proxy transport between exposure and flagevaluation flushers. I did not move the `datadog-ffe` sender onto that sidecar helper because `datadog-ffe` is the reusable crate boundary and should not depend back on `datadog-sidecar`.

datadog-ffe/src/telemetry/flagevaluation.rs
leoromanovsky: Done in https://github.com/DataDog/libdatadog/pull/2117/commits/309ea95115dceb79301e343b29666497ca34eaee. Placeholder stripping is now path-aware: only root `context.evaluation` is exempt, while unrelated `evaluation` objects still recurse and clean placeholders.

libdd-ffe/src/telemetry/flagevaluation/sender.rs
leoromanovsky: Done in https://github.com/DataDog/libdatadog/pull/2117/commits/f1eb1ef35b57a69f987f36bf7e3cdf9b9b5dfb4c. Sender encode/build/timeout/request/non-2xx failure paths now log at debug; the sidecar wrapper also counts send failures without warning on best-effort fire-and-forget delivery.

datadog-ffe/src/telemetry/flagevaluation.rs
leoromanovsky: Removed the confusing crate-naming note as part of https://github.com/DataDog/libdatadog/pull/2117/commits/bdbe29c18eb2a9560e9e7b6165c6de406a98a7f8; the remaining module docs describe the reusable EVP flagevaluation behavior and bincode/EVP serialization boundary.

libdd-ffe/src/telemetry/flagevaluation.rs
leoromanovsky: Kept this manual and documented the bound rationale here: https://github.com/DataDog/libdatadog/pull/2117/commits/2972e30086534ea8cfc35c7bee99c3184ee9b158. The manual default impl was introduced in https://github.com/DataDog/libdatadog/pull/2117/commits/3ea3104e2ffd971db67a32b1229af4cdea10aac6 so `Default` does not require a destination key type to implement `Default`.

libdd-ffe/src/telemetry/flagevaluation.rs
leoromanovsky: Agreed; simplified in https://github.com/DataDog/libdatadog/pull/2117/commits/bdbe29c18eb2a9560e9e7b6165c6de406a98a7f8. Payload splits are now derived from `payloads.len().saturating_sub(1)`, and `push_payload` asserts it only receives non-empty event groups instead of using that empty case as normal control flow.

datadog-sidecar/src/service/sidecar_server.rs
leoromanovsky: Done in https://github.com/DataDog/libdatadog/pull/2117/commits/2b6779aef46d8fe428f18e533a4a2f968c694211. `SidecarFlushOptions` now has a `flag_evaluations` flag, and sidecar flush runs flag-evaluation flushing in parallel with traces/stats while telemetry remains after those flushes.

datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
```diff
@@ -0,0 +1,988 @@
...
+        _ => {}
+    }
+}
+
+/// Whether a (already-cleaned) JSON value is an empty/null placeholder that
+/// should be dropped from the POST. Numeric zeros are NOT placeholders.
+fn is_placeholder(key: &str, value: &serde_json::Value) -> bool {
+    match value {
+        serde_json::Value::Null => true,
+        serde_json::Value::Bool(b) => key == "runtime_default_used" && !b,
```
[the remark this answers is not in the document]
leoromanovsky: Addressed in https://github.com/DataDog/libdatadog/pull/2117/commits/309ea95115dceb79301e343b29666497ca34eaee. Placeholder stripping now removes null/empty placeholders on the outbound JSON path while preserving the schema-specific `context.evaluation` subtree.

datadog-ffe/src/telemetry/flagevaluation.rs
leoromanovsky: Agreed; simplified in https://github.com/DataDog/libdatadog/pull/2117/commits/bdbe29c18eb2a9560e9e7b6165c6de406a98a7f8. `payload_splits` is now derived from `result.payloads.len().saturating_sub(1)` instead of being counted manually in the loop.

datadog-sidecar/src/service/mod.rs
leoromanovsky: No code change for this one; agreed it was irrelevant to the flag-evaluation flush path.

leoromanovsky:
Had a DM conversation with Bob that after this review's comments are addressed, it would be approved; merging.

bwoebi (approved):
I realized I still have't approved it, after commenting before.

leoromanovsky:
> I realized I still have't approved it, after commenting before.

@bwoebi Thanks for following up! I was going to ping next week as we are kicking off upstream PHP integration for this.

PR #2121, opened 2026-06-15
refactor(shm)!: Extract one_way_shared_memory to IPC and prepare libdd-remote-config for python

bwoebi:
Adds pyo3 to the enums to expose them directly, fixes the acknowledgment status and adds futex_wait to shm.


datadog-ipc/src/one_way_shared_memory.rs
```diff
@@ -0,0 +1,448 @@
...
+    std::thread::sleep(timeout);
+}
+
+/// Create a writer backed by a fresh anonymous shared-memory segment, returning
+/// the writer and a clonable [`ShmHandle`] to the same segment.
+///
+/// The handle can be mapped by readers in the same process or inherited by
+/// forked children (an anonymous mapping survives `fork`), letting them build a
+/// [`OneWayShmReader`] over what this writer publishes. The segment starts at one
+/// page and grows on demand as larger buffers are written. Use
```
yannham, line 162: > The segment starts at one page and grows on demand as larger buffers are written

How does that work when the page is already shared? Is the existing data still left in place, and there's some linked list of pages in `ShmHandle`? Or is it copied over to some larger location? How does the reader notice that?

datadog-ipc/src/one_way_shared_memory.rs
```diff
@@ -0,0 +1,448 @@
...
+    ))
+}
+
+impl<T: FileBackedHandle + From<MappedMem<T>>, D> OneWayShmReader<T, D> {
+    /// Create a reader over an already-mapped segment.
+    ///
+    /// `handle` is the live mapping to read from — typically an anonymous segment
+    /// inherited across a `fork`. Passing `None` leaves the reader without a
+    /// mapping; since this constructor installs no opener, such a reader stays
+    /// inert (empty reads, `wait_for_change` sleeps) until a handle is supplied —
```
yannham, line 181: Naive question: what is the use case for allowing a reader to have no underlying handle?

datadog-ipc/src/one_way_shared_memory.rs
```diff
@@ -0,0 +1,448 @@
...
+                    reinterpret_u8_as_u64_slice(&handle.as_slice()[0..size])
+                });
+
+                // refetch, might have been resized
+                let source_data: &RawData =
+                    unsafe { reinterpret_u8_as_u64_slice(handle.as_slice()) }.into();
+                let copied_data: &RawData = new_mem.as_slice().into();
+
+                // Ensure a new write hasn't started yet
+                fence(Ordering::Acquire); // prevent loads before from being reordered with gen load after
```
yannham, line 301: I don't think an acquire fence provide that. Usually you _can_ move things past it. What you can't do is move things _before_ it.

To be fair, this kind of passive reader scheme is just [impossible to express safely in current Rust](https://github.com/m-ou-se/rfcs/blob/atomic-memcpy/text/3301-atomic-memcpy.md). The seqlock crate does use an `Acquire` fence as well, but I'm not sure what kind of guarantees it provides. On paper a `Release` fence is more adapted but it's not clear how it interacts with the `load` below since load can't be `Release`. Or maybe use a `SeqCst` fence just to be more constraining? That's what we do for the Otel process-level context, although this is more costly (in the OTel part we don't care, it's not read very often)

libdd-remote-config/src/lib.rs
```diff
@@ -39,6 +39,7 @@ pub struct Target {
 #[repr(C)]
 #[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+#[cfg_attr(feature = "pyo3", pyo3::pyclass(eq, eq_int, from_py_object))]
```
hoolioh: In this case I'd leave the the pyo3 mirroring strictly to the dd-trace-py. That's the model we use for other crates which, IMO, is more flexible and maintanable for both teams.

datadog-ipc/src/one_way_shared_memory.rs
```diff
@@ -0,0 +1,448 @@
...
+                return false;
+            }
+        }
+
+        // Raw pointer to the generation atomic inside the live mapping. It lives
+        // in the first page (before the resizable buffer), so its address is
+        // stable for the duration of the wait. `wait_for_change` and `read` are
+        // only ever called from the same (reader) thread, so no concurrent remap
+        // can invalidate this pointer.
+        let generation_ptr: *const AtomicU64 = {
```
yannham: What's the rationale behind creating a reference, then a raw pointer out of it, and then a reference again out of the raw pointer? Why not use `&data.meta.generation` directly instead? Is it to avoid holding a borrow onto `data` or something?

libdd-remote-config/src/path.rs
```diff
@@ -24,6 +25,15 @@ pub enum RemoteConfigProduct {
     AsmFeatures,
     FfeFlags,
     LiveDebugger,
+    LiveDebuggerSymbolDb,
+}
+
+#[cfg(feature = "pyo3")]
+#[pyo3::pymethods]
+impl RemoteConfigProduct {
```
hoolioh: Same as above, I'd mirror the the enum in `dd-trace-py`, implement a exhaustive match to convert the values and implement the `__str__` there.

datadog-ipc/src/one_way_shared_memory.rs
```diff
@@ -0,0 +1,448 @@
...
+        let mut mapped = self.handle.lock_or_panic();
+
+        let size = contents.len() + 1; // trailing zero byte, to keep some C code happy
+        mapped.ensure_space(std::mem::size_of::<RawMetaData>() + size);
+
+        // Safety: ShmHandle is always big enough
+        // Actually &mut mapped.as_slice_mut() as RawData seems safe, but unsized locals are
+        // unstable
+        let data = unsafe { &mut *(mapped.as_slice_mut() as *mut [u8] as *mut RawData) };
+        data.meta.generation.fetch_add(1, Ordering::AcqRel);
```
yannham: Nitpick, but I think the `Release` part here is useless, since anyone loading an odd value won't look at unprotected shared memory.

```suggestion
        data.meta.generation.fetch_add(1, Ordering::Acquire);
```

datadog-ipc/src/one_way_shared_memory.rs
```diff
@@ -0,0 +1,448 @@
...
+        data.meta.size = size;
+
+        data.as_slice_mut()[0..contents.len()].copy_from_slice(contents);
+        data.as_slice_mut()[contents.len()] = 0;
+
+        data.meta.generation.fetch_add(1, Ordering::Release);
+
+        // Wake any readers blocked in `wait_for_change` on the generation word.
+        // A wake with no waiters is a cheap no-op syscall.
+        futex_wake((&data.meta.generation as *const AtomicU64).cast::<u32>());
```
yannham, line 433: Very nit, but do we need to care about exotic big endian platforms here?

datadog-ipc/Cargo.toml
```diff
@@ -68,3 +68,6 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(polyfill_glibc_memfd)'] }
 
 [features]
 tiny-bytes = ["libdd-tinybytes"]
+# Wake `OneWayShmReader::wait_for_change` via a futex on the generation word
+# (Linux only). Without this, the wait falls back to a timed poll.
+one_way_shm_futex = []
```
yannham, line 73: Should this be a default feature on linux? you're already gating the `cfg` with `platform=linux` anyway.

yannham (approved):

datadog-ipc/src/one_way_shared_memory.rs
bwoebi: It makes no sense without opener, that's correct.

yannham: Should `new` take a `MappedMem<T>` instead of an `Option<_>`? (whether you keep the option internally so that you can `take` it later is a different question)

datadog-ipc/src/one_way_shared_memory.rs
bwoebi: We must not move the generation load _before_ the *read of the data from shm*. Which is what the acquire guarantees.
Release fences are centered around writes, not reads.

That's why the comment says "prevent loads (from shm) from being reordered...".
The read/write ordering within a same thread is guaranteed to be consistent by the rust memory model (i.e. this doesn't need to be fenced).

bwoebi: Ordering requirements when reading are always towards external writers, i.e. reads from something which might be written to by some external writer.
I.e. the compiler is not allowed to assume that source_data points to the same data than the *source* used to copy from into source_data after the fence - i.e. not allowed to alias source and target - it *must* thus create a data dependency on the target data. Which it would ordinarily be allowed to, as copied data is "obviously" identical to the source.

datadog-ipc/Cargo.toml
bwoebi: No, it shouldn't. E.g. the sidecar is not using the futex and it'll incur unnecessary syscall overhead.

libdd-remote-config/src/lib.rs
bwoebi: done, made enum inspection available via strum to achieve the same goal.

datadog-ipc/src/one_way_shared_memory.rs
bwoebi: Yes, that's what I did :-D

datadog-ipc/src/one_way_shared_memory.rs
bwoebi: No, just excluded it via target_arch.

datadog-ipc/src/one_way_shared_memory.rs
yannham: I'm not sure to understand exactly what you mean, sorry. Maybe I don't really understand the original comment.

While re-ordering is a useful mental model, it's not how the Rust memory model is defined anyway. What acquire gives you is that it synchronize the load on this thread with the corresponding store that put the value there on some other thread, creating a connection: that the store _happens-before_ the load. So all observable memory accesses (shared but potentially non-atomic) done on the writer before writing the value to the atomic are visible on the reader side after we load. Since the load happens before whatever is next done on the reader side (sequenced-before), all memory accesses (shared but potentially non-atomic) occurring after the load will also not be re-ordered before the load (or acquire fence in our case).

However, it doesn't tell anything about accesses that are sequenced **before** the atomic load/ the fence on the reader thread. The following is, I believe, a totally legal re-ordering, if the compiler or the hardware think it can't be observed by the current thread (but this is always the case for a bunch of reads):

```
read address A (non-atomic)
read address B (non-atomic)
atomic load relaxed C
acquire fence
...rest
```

to

```
atomic load relaxed C
acquire fence
read address A (non-atomic)
read address B (non-atomic)
..rest
```

Whether it can happen in this specific case because of the data dependency is another question 🤷 , but I think the acquire fence doesn't help preventing it, or at least is not guaranteed to help.

yannham: (and by the way, an acquire fence should be placed _after_ the corresponding load(s), see https://doc.rust-lang.org/std/sync/atomic/fn.fence.html#atomic---fence -- I'm not sure what is the semantics when put before)

bwoebi: Our acquire fence is after the load by `new_mem.extend_from_slice(unsafe { reinterpret_u8_as_u64_slice(&handle.as_slice()[0..size]) });`. The problem is, as you say, that I cannot make this an atomic memcpy, because rust doesn't support it currently. But it's that one which the fence is supposed to synchronize with.
I think you might have a point that the subsequent Relaxed load also needs a release fence before, requiring AcqRel.

yannham: Ah, I see, you're trying to protect this access. Indeed the seqlock-like readers like this just can't be implemented properly (at least in theory) in Rust, and the semantics of fences is a bit crazy (the whole "atomics are required" part is really unclear). I don't think there's a canonical and satisfying solution. I would say if perf is not critical here `SeqCst` might be safer (I know, I am the one usually ranting against `SeqCst` by default)

However, the seqlock crate does as you do, so I suppose it's also somehow reasonable. Acquire fences should be compiled to `dmb ishld` on arm, which provides sufficient guarantees on the hardware side in practice (all reads are completed before proceeding).

People have used `volatile_read` as well in addition to atomics, as this has lower risk of being re-ordered past atomics. Not sure if it's necessary.

bwoebi: Exactly, and the `dmb ishld` on ARM is what we need, in practice. (x86 with its total store ordering is unproblematic).

PR #2122, opened 2026-06-15
fix: Fix http PathAndQuery Uri Parsing

jkosh44-datadog:
# What does this PR do?

`http` version 1.4.1 became stricter in `PathAndQuery` input parsing to reject all inputs that do not start with `/`, including empty inputs. Previously, this library passed in an empty input in multiple places when constructing a `PathAndQuery`. Users are not able to upgrade `http` past 1.4.0 without this library breaking. This commit fixes the issue by replacing the empty inputs with `"/"`.

# Motivation

Fix this library for `http` versions >= 1.4.1.

# Additional Notes

See https://dd.slack.com/archives/C05M4JAM5BJ/p1781552060779329?thread_ts=1781534034.370109&cid=C05M4JAM5BJ

# How to test the change?

Existing unit tests should be sufficient, assuming the modified functions have coverage.

### DIFF
--- a/libdd-common/src/connector/named_pipe.rs
+++ b/libdd-common/src/connector/named_pipe.rs
@@ -19,7 +19,7 @@ pub fn named_pipe_path_to_uri(path: &Path) -> Result<hyper::Uri, hyper::http::Er
     hyper::Uri::builder()
         .scheme("windows")
         .authority(path)
-        .path_and_query("")
+        .path_and_query("/")
         .build()
 }
 
--- a/libdd-common/src/lib.rs
+++ b/libdd-common/src/lib.rs
@@ -308,7 +308,7 @@ fn encode_uri_path_in_authority(scheme: &str, path: &str) -> anyhow::Result<http
     let path = hex::encode(path);
 
     parts.authority = uri::Authority::from_str(path.as_str()).ok();
-    parts.path_and_query = Some(uri::PathAndQuery::from_static(""));
+    parts.path_and_query = Some(uri::PathAndQuery::from_static("/"));
     Ok(http::Uri::from_parts(parts)?)
 }
 
@@ -511,5 +511,24 @@ impl Endpoint {
         };
 
         Ok((builder, request_url))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::parse_uri;
+
+    /// A scheme prefix with an empty path produces an empty (and therefore
+    /// dropped) authority. parsing must reject these as malformed rather
+    /// than accept them.
+    #[test]
+    fn empty_authority_uris_are_rejected() {
+        for input in ["unix://", "windows:", "file://"] {
+            let result = parse_uri(input);
+            assert!(
+                result.is_err(),
+                "expected {input:?} to be rejected, got {result:?}"
+            );
+        }
     }
 }
--- a/libdd-common/src/connector/uds.rs
+++ b/libdd-common/src/connector/uds.rs
@@ -12,7 +12,7 @@ pub fn socket_path_to_uri(path: &Path) -> Result<hyper::Uri, hyper::http::Error>
     hyper::Uri::builder()
         .scheme("unix")
         .authority(path)
-        .path_and_query("")
+        .path_and_query("/")
         .build()
 }
 


libdd-common/src/lib.rs
```diff
@@ -308,7 +308,7 @@ fn encode_uri_path_in_authority(scheme: &str, path: &str) -> anyhow::Result<http
     let path = hex::encode(path);
 
     parts.authority = uri::Authority::from_str(path.as_str()).ok();
-    parts.path_and_query = Some(uri::PathAndQuery::from_static(""));
+    parts.path_and_query = Some(uri::PathAndQuery::from_static("/"));
```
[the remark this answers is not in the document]
jkosh44-datadog: This is not accurate, I've added a unit test to confirm.

libdd-common/src/connector/named_pipe.rs
```diff
@@ -19,7 +19,7 @@ pub fn named_pipe_path_to_uri(path: &Path) -> Result<hyper::Uri, hyper::http::Er
     hyper::Uri::builder()
         .scheme("windows")
         .authority(path)
-        .path_and_query("")
+        .path_and_query("/")
```
danielsn, line 22: is this correct for windows?

danielsn (commented):
Otherwise LGTM

libdd-common/src/connector/named_pipe.rs
jkosh44-datadog: I think so. If you look at the code, it's explicitly expecting the first byte to be `b'/'`, and I don't see any `cfg` flags based on OS.

https://docs.rs/http/latest/src/http/uri/path.rs.html#427-429

Here's the full path of `path_and_query()`:

- `path_and_query()` calls into `PathAndQuery::try_into`: https://docs.rs/http/latest/src/http/uri/builder.rs.html#93-103
- `TryFrom<&str>` calls into `TryFrom<&[u8]>`: https://docs.rs/http/latest/src/http/uri/builder.rs.html#93-103
- `TryFrom<&[u8]>` calls into `PathAndQuery::from_shared`: https://docs.rs/http/latest/src/http/uri/path.rs.html#223-229
- `from_shared` calls `scan_path_and_query`: https://docs.rs/http/latest/src/http/uri/path.rs.html#21-39
- `scan_path_and_query` returns an error if the bytes are empty: https://docs.rs/http/latest/src/http/uri/path.rs.html#415-417
- `scan_path_and_query` returns an error if the first byte isn't one of `/`, `?`, `#`: https://docs.rs/http/latest/src/http/uri/path.rs.html#427-429

I see that there's two tests at the bottom of this file that call this function, are they run on Windows? If so, is that sufficient to gain confidence that it's correct on Windows? If not, is there any additional test we can add?

I'm not super familiar with this code base or with windows so it's hard for me to say with a high level of certainty if it's correct or not.

morrisonlevi (commented):
Have you manually tested unix domain sockets at all? I think that this probably is fine but it's one of those things I think is worth testing end-to-end. If you want help, I can show you how to do this for dd-trace-php.

jkosh44-datadog:
> Have you manually tested unix domain sockets at all? I think that this probably is fine but it's one of those things I think is worth testing end-to-end. If you want help, I can show you how to do this for dd-trace-php.

@morrisonlevi I have not manually tested it, and yes I would appreciate if you could show me how. I have not done any manual testing and have been leaning on CI fully for validation.

jkosh44-datadog:
> > Have you manually tested unix domain sockets at all? I think that this probably is fine but it's one of those things I think is worth testing end-to-end. If you want help, I can show you how to do this for dd-trace-php.
> 
> @morrisonlevi I have not manually tested it, and yes I would appreciate if you could show me how. I have not done any manual testing and have been leaning on CI fully for validation.

@morrisonlevi bump, how can I test the unix domain sockets end-to-end?

paullegranddc (approved):
For the data-pipeline crates, this looks correct to me. 
The authority get's decoded all the same for unix socket/windows piped and the  `path_and_query` part of the URL is replaced to talk to specific endpoints

hoolioh (approved):

paullegranddc:
> Have you manually tested unix domain sockets at all? 

We do have integration tests using UDS in the trace exporter https://github.com/DataDog/libdatadog/blob/main/libdd-data-pipeline/tests/test_trace_exporter.rs#L345C14-L345C31

PR #2124, opened 2026-06-16
test(trace-normalization): add microbenchmarks for tag/metric/truncate normalization

yannham:
# What does this PR do?

Extends the existing criterion bench (`libdd-trace-normalization/benches/normalization_utils.rs`) to cover the normalization functions that were previously unmeasured, all of which run on every ingested span:

- `normalize_tag`: the heaviest function (per-codepoint UTF-8 scan + char-class state machine). Benched on ASCII fast-path, mixed/illegal-char, unicode (codepoint slow path), and over-length (> `MAX_TAG_LEN`) inputs.
- `normalize_metric_name`: similar complexity with one-byte lookahead. Clean, separator-collapsing, and over-length cases.
- `truncate_utf8`: over-length ASCII plus a multi-byte (3-byte) input where the limit lands mid-codepoint and forces the boundary walk-back.
- `normalize_span_start_duration`: `clean` vs `needs-clock` cases to quantify the `SystemTime` read on the pre-year-2000 path.

A `bench-internals` cargo feature is added (mirroring `libdd-sampling`) to expose the otherwise-private benched functions, without changing the shipped public API. The `[[bench]]` now requires this feature.

# Motivation

Normalization runs on every span, but the existing bench skipped the expensive per-char UTF-8 state machines. These are the functions most likely to show up as a per-span tax, so they are worth tracking.

# Additional Notes

- Follows the existing bench idioms (same group naming, `iter_batched_ref` over 1000 owned copies, `Throughput(Elements)`, `black_box` on inputs).
- Net change is small (~190 LoC).
- Default (shipped) build is unaffected: the visibility split is gated behind `bench-internals`.

# How to test the change?

```bash
cargo check -p libdd-trace-normalization
cargo bench -p libdd-trace-normalization --features bench-internals --no-run
cargo bench -p libdd-trace-normalization --features bench-internals -- --warm-up-time 1 --measurement-time 1 --sample-size 10
```

🤖 (Partly) Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/benchmark/run_benchmarks_ci.sh
+++ b/benchmark/run_benchmarks_ci.sh
@@ -22,7 +22,7 @@ pushd "${PROJECT_DIR}" > /dev/null
 
 # Run benchmarks
 message "Running benchmarks"
-cargo bench --workspace --features libdd-crashtracker/benchmarking,libdd-sampling/v04_span,libdd-sampling/bench-internals -- --warm-up-time 1 --measurement-time 5 --sample-size=200
+cargo bench --workspace --features libdd-crashtracker/benchmarking,libdd-sampling/v04_span,libdd-sampling/bench-internals,libdd-trace-normalization/bench-internals -- --warm-up-time 1 --measurement-time 5 --sample-size=200
 message "Finished running benchmarks"
 
 # Copy the benchmark results to the output directory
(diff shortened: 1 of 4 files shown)


VianneyRuhlmann (commented):
Do we really need these benches ? The existing bench on public methods already call these functions e.g. `normalize_name` and `normalize_service`. Also the normalization benches are already quite flaky so we should check these new benches are stable.

yannham:
After more consideration, I agree there's too much overlap. There _are_ some unexercised code paths in the previous benches, but I suppose they are not interesting (before year 2000 code path, > 100 bytes tags). Closing for now.

PR #2126, opened 2026-06-16
test(trace-utils): add VecMap microbenchmarks

yannham:
# What does this PR do?

Adds criterion microbenchmarks for `VecMap` (`libdd-trace-utils/src/span/vec_map.rs`), the linear-scan ordered map backing each span's `meta` and `metrics`. The benches isolate the datastructure's core operations across representative map sizes (`4, 8, 16, 32, 64`), using `BytesString` keys to match real span usage:

- `insert` (construction path), `get` (hit and miss), `get_mut`, `contains_key`
- `iter` (encode path)
- `dedup` (runs once per span on decode), with and without duplicate keys
- `as_deduped_map` (encode path), both the cheap borrow case and the allocating fallback

A new opt-in `bench-internals` feature (mirroring `libdd-sampling`) gates the bench via `required-features` so it is not built by default. The change is additive to `Cargo.toml`.

# Motivation

`VecMap` is a swappable datastructure (could be a hashmap/btree) that runs on the tracer hot path, but existing trace-utils benches only measure end-to-end msgpack encode/decode and never isolate it. Since it is linear-scan, the size-vs-operation curve is exactly what should be measured so the backing-structure choice can be revisited with data.

# Additional Notes

`VecMap` and the `span` module are already `pub`, so `bench-internals` does not gate any source visibility; it is purely an opt-in switch keeping the bench out of the default build, consistent with `libdd-sampling`. Data is fully deterministic (no randomness).

# How to test the change?

```
cargo bench -p libdd-trace-utils --features bench-internals --bench vec_map_bench --no-run
cargo bench -p libdd-trace-utils --features bench-internals --bench vec_map_bench
```

### DIFF
--- a/libdd-trace-utils/Cargo.toml
+++ b/libdd-trace-utils/Cargo.toml
@@ -17,6 +17,12 @@ name = "main"
 harness = false
 path = "benches/main.rs"
 
+[[bench]]
+name = "vec_map_bench"
+harness = false
+path = "benches/vec_map_bench.rs"
+required-features = ["bench-internals"]
+
 [dependencies]
 anyhow = "1.0"
 base64 = "0.22"
@@ -88,6 +94,10 @@ test-utils = [
     "urlencoding",
 ]
 change-buffer = []
+# Opt-in switch for crate-internal microbenchmarks (e.g. `vec_map_bench`). Mirrors the
+# `bench-internals` feature in `libdd-sampling`; the bench targets it via `required-features` so
+# they are not built by default. Not intended for downstream consumers.
+bench-internals = []
 compression = ["zstd", "flate2"]
 # FIPS mode uses the FIPS-compliant cryptographic provider (Unix only)
 fips = ["libdd-common/fips", "libdd-capabilities-impl/fips"]
--- a/libdd-trace-utils/Cargo.toml
+++ b/libdd-trace-utils/Cargo.toml
@@ -94,9 +94,7 @@ test-utils = [
     "urlencoding",
 ]
 change-buffer = []
-# Opt-in switch for crate-internal microbenchmarks (e.g. `vec_map_bench`). Mirrors the
-# `bench-internals` feature in `libdd-sampling`; the bench targets it via `required-features` so
-# they are not built by default. Not intended for downstream consumers.
+# Opt-in switch for crate-internal microbenchmarks (e.g. `vec_map_bench`).
 bench-internals = []
 compression = ["zstd", "flate2"]
 # FIPS mode uses the FIPS-compliant cryptographic provider (Unix only)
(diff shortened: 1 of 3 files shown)


libdd-trace-utils/Cargo.toml
```diff
@@ -88,6 +94,10 @@ test-utils = [
     "urlencoding",
 ]
 change-buffer = []
+# Opt-in switch for crate-internal microbenchmarks (e.g. `vec_map_bench`). Mirrors the
+# `bench-internals` feature in `libdd-sampling`; the bench targets it via `required-features` so
+# they are not built by default. Not intended for downstream consumers.
+bench-internals = []
```
VianneyRuhlmann: A bit verbose, `Opt-in switch for crate-internal microbenchmarks`is sufficient

VianneyRuhlmann (approved):

libdd-trace-utils/benches/vec_map_bench.rs
Eldolfin, line 1: Should we run the same bench cases on HashMaps to prove that `VecMap`s are more performance ?

libdd-trace-utils/Cargo.toml
yannham: Ah, I caught one instance but not this one :upside_down_face:

libdd-trace-utils/benches/vec_map_bench.rs
yannham: Good question! I thought about it, but I'm not sure criterion is the right tool for that, at least used naively. In particular it would report improvements or degradation for the stdlib hashmap for future PR, which is not very relevant for libdatadog devs.

Maybe what we could do is to make a criterion benchmark when we measure the difference or the ratio?

libdd-trace-utils/Cargo.toml
```diff
@@ -88,6 +94,10 @@ test-utils = [
     "urlencoding",
 ]
 change-buffer = []
+# Opt-in switch for crate-internal microbenchmarks (e.g. `vec_map_bench`). Mirrors the
+# `bench-internals` feature in `libdd-sampling`; the bench targets it via `required-features` so
+# they are not built by default. Not intended for downstream consumers.
```
yannham:
```suggestion
# Opt-in switch for crate-internal microbenchmarks (e.g. `vec_map_bench`).
```

libdd-trace-utils/benches/vec_map_bench.rs
yannham: (in particular my current goal for micro and macro benches is to be able to measure the evolution of perf: did a change made things better or worse, and how? I feel like comparing VecMap to HashMap is a bit different, and could be done as a one-off thing on the side)

Eldolfin: I agree, it doesn't even need to be committed even 👍🏽

yannham: But you're right, we should do that. That could also gives us hint as to the range where `VecMap` should be preferred against `HashMap` or not. Is it like 1 to 50 elements and max duplicate ratio 10%, or is it actually fast until 250 elements and can tolerate 25% duplicates? Those are interesting question to answer.

PR #2127, opened 2026-06-16
test(trace-utils): add V05 msgpack decode microbenchmark

yannham:
# What does this PR do?

Adds microbenchmarks for the V05 msgpack decode path (`msgpack_decoder::v05::from_bytes`) in `libdd-trace-utils`.

# Motivation

Only the V04 decode path was benched. The V05 wire format encodes a shared string dictionary up front and every span references its string fields (service, name, resource, type, and every meta/metrics key/value) by index into that dictionary. Decode therefore performs a dictionary lookup-and-clone per string slot, which is per-span work that has no benchmark coverage.

# Additional Notes

- The bench varies two axes that drive the dedup cost: span count (200 / 2000 / 10000 spans across 20 traces) and sharing ratio. Span shapes are realistic (8 meta tags, 3 metrics) and all data is deterministic.
- `Cargo.toml` is intentionally untouched to avoid conflicting with the sibling vecmap PR.

# How to test the change?

```
cargo bench -p libdd-trace-utils --bench main -- "msgpack_decoder::v05" \
  --warm-up-time 1 --measurement-time 1 --sample-size 10
```

🤖 Generated with [Claude Code](https://claude.com/claude-code), reviewed by a human :man_shrugging:

### DIFF
--- a/libdd-trace-utils/benches/deserialization_v05.rs
+++ b/libdd-trace-utils/benches/deserialization_v05.rs
@@ -183,7 +183,7 @@ fn bench_v05_matrix<M: criterion::measurement::Measurement>(
             let data = build_v05_payload(NUM_TRACES, spans_per_trace, unique_per_span);
             let data_as_bytes = libdd_tinybytes::Bytes::copy_from_slice(&data);
 
-            group.throughput(Throughput::Bytes(data.len() as u64));
+            group.throughput(Throughput::Elements((spans_per_trace * NUM_TRACES) as u64));
             group.bench_with_input(
                 BenchmarkId::new(sharing_label, spans_per_trace * NUM_TRACES),
                 &data_as_bytes,
--- a/libdd-trace-utils/benches/deserialization_v05.rs
+++ b/libdd-trace-utils/benches/deserialization_v05.rs
@@ -59,7 +59,7 @@ const METRICS_PER_SPAN: usize = 3;
 ///
 /// Data is fully deterministic.
 fn build_v05_payload(num_traces: usize, spans_per_trace: usize, unique_per_span: bool) -> Vec<u8> {
-    let mut dict: Vec<String> = Vec::new();
+    let mut dict = Vec::new();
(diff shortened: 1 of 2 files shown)


libdd-trace-utils/benches/deserialization_v05.rs
```diff
@@ -0,0 +1,235 @@
...
+use std::alloc::System;
+use std::collections::HashMap;
+
+use criterion::{black_box, criterion_group, BenchmarkId, Criterion, Throughput};
+use libdd_common::bench_utils::{memory_allocated_measurement, AllocatedBytesMeasurement};
+use libdd_trace_utils::msgpack_decoder;
+
+/// A V05 span is a fixed 12-element tuple. String fields (service, name, resource, type, and the
+/// meta/metrics keys/values) are `u32` indices into the shared dictionary.
+type V05Span = (
```
VianneyRuhlmann: Any reason to not reuse the span model from `libdd-trace-utils/src/span/v05/mod.rs` ?

libdd-trace-utils/benches/deserialization_v05.rs
```diff
@@ -0,0 +1,235 @@
...
+/// Builds a representative V05 payload.
+///
+/// `unique_per_span` controls the string-sharing ratio:
+///   * `false` (high sharing): all spans draw their strings from a single small shared dictionary,
+///     mirroring real tracer traffic where service names and tag keys repeat heavily.
+///   * `true` (low sharing): each span contributes its own unique strings, producing a large
+///     dictionary with little reuse.
+///
+/// Data is fully deterministic.
+fn build_v05_payload(num_traces: usize, spans_per_trace: usize, unique_per_span: bool) -> Vec<u8> {
```
VianneyRuhlmann, line 51: Might be worth using the `from_v04_span` logic to generate the v05 payload instead of reimplementing it

VianneyRuhlmann (approved):

PR #2128, opened 2026-06-16
chore(nix): add development shell and CI

lloeki:
# What does this PR do?

Adds a Nix flake development shell to the repository, with a CI workflow that exercises it.

- `flake.nix` / `flake.lock`: a devshell providing the Rust toolchain (rustc, cargo, rustfmt, clippy), cbindgen, cmake and autotools. The Rust toolchain is pinned via the `oxalica/rust-overlay` input reading the existing `rust-toolchain.toml`, so the devshell tracks the same channel as CI and rustup.
- `.github/workflows/nix.yml`: enters the devshell and builds the workspace on Linux (x86_64, aarch64) and macOS (arm64), mirroring the Nix CI already present in dd-trace-rb and libdatadog-rb.
- `CODEOWNERS`: co-owns the Nix files and workflow with the Nix guild.

# Motivation

Contributors using Nix previously kept these files locally and unshared, with the Rust version floating with the nixpkgs channel (currently 1.91.1) rather than matching the toolchain libdatadog is built with. Committing a pinned devshell gives a reproducible, low-setup environment, and the CI guard keeps it from bit-rotting.

# Additional Notes

The pin derives from `rust-toolchain.toml` (single source of truth), so a future MSRV bump updates the devshell automatically.

# How to test the change?

`nix develop --command cargo build --workspace --exclude builder`, and confirm `nix develop --command rustc --version` reports the channel from `rust-toolchain.toml`. CI runs the same build across the three platforms.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -111,3 +111,4 @@ libdd-data-pipeline/tests/test_trace_exporter_otlp_export.rs @DataDog/apm-sdk-ca
 libdd-trace-utils/src/otlp_encoder/                     @DataDog/apm-sdk-capabilities-rust
 datadog-sidecar/src/service/ffe_exposures_flusher.rs    @DataDog/libdatadog-php @DataDog/libdatadog-apm @DataDog/feature-flagging-and-experimentation-sdk
 datadog-sidecar/src/service/ffe_metrics_flusher.rs      @DataDog/libdatadog-php @DataDog/libdatadog-apm @DataDog/feature-flagging-and-experimentation-sdk
+.github/workflows/nix.yml                               @DataDog/nix-guild @DataDog/apm-common-components-core
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -97,6 +97,10 @@ tools/                               @DataDog/apm-common-components-core
 windows/                             @DataDog/libdatadog-core
 fuzz/                                @DataDog/chaos-platform
 
+# Nix
+*.nix                                @DataDog/nix-guild @DataDog/apm-common-components-core
+flake.*                              @DataDog/nix-guild @DataDog/apm-common-components-core
+
 # Specific overrides (must come after their general patterns above)
 bin_tests/tests/test_the_tests.rs    @DataDog/libdatadog-core
 bin_tests/src/bin/test_the_tests.rs  @DataDog/libdatadog-core
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -15,6 +15,11 @@ on: # yamllint disable-line rule:truthy
       - "rust-toolchain.toml"
       - "nightly-toolchain.toml"
       - ".github/workflows/nix.yml"
+  # Nightly safeguard: catches devshell drift from changes that don't match the
+  # paths above (e.g. a dependency or build-script change that needs new native
+  # tooling). Scheduled runs only fire from the default branch, against its HEAD.
+  schedule:
+    - cron: "0 4 * * *" # daily at 04:00 UTC
 
 # Default permissions for all jobs
 permissions: {}
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -42,7 +42,6 @@ jobs:
 
     permissions:
       contents: read
-      id-token: write
 
     steps:
       - name: Check CPU arch
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -6,13 +6,14 @@ on: # yamllint disable-line rule:truthy
       - main
       - mq-working-branch-*
   # Also run on PRs that touch the devshell or what it reads. Paths mirror the
-  # Nix CODEOWNERS entries, plus rust-toolchain.toml (read by the flake) and
+  # Nix CODEOWNERS entries, plus the toolchain files (read by the flake) and
   # this workflow itself.
   pull_request:
     paths:
       - "*.nix"
       - "flake.*"
       - "rust-toolchain.toml"
+      - "nightly-toolchain.toml"
       - ".github/workflows/nix.yml"
 
 # Default permissions for all jobs
@@ -71,6 +72,8 @@ jobs:
           nix develop --command rustc --version
           nix develop --command cargo --version
           nix develop --command cbindgen --version
+      - name: Check nightly formatter toolchain
+        run: nix develop .#nightly --command cargo fmt --version
       - name: Build workspace
         run: nix develop --command cargo build --workspace --exclude builder
 
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -51,6 +51,17 @@ jobs:
       - name: Check CPU arch
         run: |
           test "$(uname -m)" = "${{ matrix.platform.cpu }}"
+      - name: Free Disk Space (Linux only)
+        if: runner.os == 'Linux'
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # 1.3.1
+        with:
+          tool-cache: true
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: false
+          docker-images: false
+          swap-storage: false
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,7 +1,6 @@
 name: Test Nix
 
 on: # yamllint disable-line rule:truthy
-  pull_request:
   push:
     branches:
       - main
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,61 @@
+name: Test Nix
+
+on: # yamllint disable-line rule:truthy
+  pull_request:
+  push:
+    branches:
+      - main
+      - mq-working-branch-*
+
+# Default permissions for all jobs
+permissions: {}
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - os: darwin
+            cpu: arm64
+            base: macos-15 # always arm64-darwin
+          - os: linux
+            cpu: x86_64
+            base: ubuntu-24.04 # always x86_64-linux-gnu
+          - os: linux
+            cpu: aarch64
+            base: ubuntu-24.04-arm # always aarch64-linux-gnu
+
+    name: Test Nix (${{ matrix.platform.cpu }}-${{ matrix.platform.os }})
+    runs-on: ${{ matrix.platform.base }}
+
+    permissions:
+      contents: read
+      id-token: write
+
+    steps:
+      - name: Check CPU arch
+        run: |
+          test "$(uname -m)" = "${{ matrix.platform.cpu }}"
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: cachix/install-nix-action@ab739621df7a23f52766f9ccc97f38da6b7af14f # v31.10.5
+      - name: Print toolchain versions
+        run: |
+          nix develop --command rustc --version
+          nix develop --command cargo --version
+          nix develop --command cbindgen --version
+      - name: Build workspace
+        run: nix develop --command cargo build --workspace --exclude builder
+
+  complete:
+    name: Nix (complete)
+    runs-on: ubuntu-24.04
+    needs:
+      - test
+    steps:
+      - run: echo "DONE!"
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -18,6 +18,10 @@ on: # yamllint disable-line rule:truthy
 # Default permissions for all jobs
 permissions: {}
 
+concurrency:
+  group: ci-${{ github.ref == 'refs/heads/main' && github.run_id || github.ref }}-nix
+  cancel-in-progress: true
+
 env:
   CARGO_TERM_COLOR: always
 
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -5,6 +5,15 @@ on: # yamllint disable-line rule:truthy
     branches:
       - main
       - mq-working-branch-*
+  # Also run on PRs that touch the devshell or what it reads. Paths mirror the
+  # Nix CODEOWNERS entries, plus rust-toolchain.toml (read by the flake) and
+  # this workflow itself.
+  pull_request:
+    paths:
+      - "*.nix"
+      - "flake.*"
+      - "rust-toolchain.toml"
+      - ".github/workflows/nix.yml"
 
 # Default permissions for all jobs
 permissions: {}
--- a/flake.nix
+++ b/flake.nix
@@ -40,7 +40,7 @@
           # hardening in the shell so those builds succeed.
           hardeningDisable = [ "fortify" "fortify3" ];
 
-          buildInputs = [
+          nativeBuildInputs = [
             rust            # rustc + cargo + rustfmt + clippy, pinned via toolchain file
             pkgs.rust-cbindgen
             pkgs.cmake
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,46 @@
+{
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/release-25.11";
+
+    # cross-platform convenience
+    flake-utils.url = "github:numtide/flake-utils";
+
+    # backwards compatibility with nix-build and nix-shell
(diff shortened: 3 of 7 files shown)


lloeki:
@codex review

.github/workflows/nix.yml
```diff
@@ -0,0 +1,61 @@
...
+        platform:
+          - os: darwin
+            cpu: arm64
+            base: macos-15 # always arm64-darwin
+          - os: linux
+            cpu: x86_64
+            base: ubuntu-24.04 # always x86_64-linux-gnu
+          - os: linux
+            cpu: aarch64
+            base: ubuntu-24.04-arm # always aarch64-linux-gnu
```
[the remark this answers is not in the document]
lloeki: Seems like it's working ok?

flake.nix
```diff
@@ -0,0 +1,53 @@
...
+          # The stdenv cc-wrapper injects -D_FORTIFY_SOURCE, which glibc rejects
+          # when compiling without optimization. Some build scripts (e.g.
+          # spawn_worker's trampoline.c) compile C at -O0 with -Werror, so the
+          # resulting fortify #warning becomes a hard error. Disable fortify
+          # hardening in the shell so those builds succeed.
+          hardeningDisable = [ "fortify" "fortify3" ];
+
+          buildInputs = [
+            rust            # rustc + cargo + rustfmt + clippy, pinned via toolchain file
+            pkgs.rust-cbindgen
```
lloeki, line 45: Should `cbindgen` be pinned as well?

.github/workflows/nix.yml
```diff
@@ -0,0 +1,61 @@
+name: Test Nix
+
+on: # yamllint disable-line rule:truthy
+  pull_request:
```
hoolioh: I'd remove triggering it on every push to a pull request and leave it on main and mq. The only downside is that errors will suface later in the pipeline bu since it's a safe guard for the devshell I won't expect that it will be modified regularly.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,61 @@
...
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - os: darwin
+            cpu: arm64
+            base: macos-15 # always arm64-darwin
```
hoolioh, line 42: Is macos strictly needed? I ask because macos runners are scarce and ocuppying one for this might hurt us a bit in terms of merge times

.github/workflows/nix.yml
lloeki: Fair enough. I would also add conditionally running whenever nix files (or direct dependents such as `rust-toolchain.toml`) are updated.

.github/workflows/nix.yml
lloeki: Yes, it guarantees that it consistently works across OSes for any dev that uses Nix on a macOS machine

Should be mostly mitigated by conditional running mentioned above.

.github/CODEOWNERS
```diff
@@ -95,6 +95,10 @@ tools/                               @DataDog/apm-common-components-core
 windows/                             @DataDog/libdatadog-core
 fuzz/                                @DataDog/chaos-platform
 
+# Nix
+*.nix                                @DataDog/nix-guild @DataDog/apm-common-components-core
```
yannham, line 101: TIL there's a Nix guild 😄

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+
+          # The stdenv cc-wrapper injects -D_FORTIFY_SOURCE, which glibc rejects
+          # when compiling without optimization. Some build scripts (e.g.
+          # spawn_worker's trampoline.c) compile C at -O0 with -Werror, so the
+          # resulting fortify #warning becomes a hard error. Disable fortify
+          # hardening in the shell so those builds succeed.
+          hardeningDisable = [ "fortify" "fortify3" ];
+
+          buildInputs = [
+            rust            # rustc + cargo + rustfmt + clippy, pinned via toolchain file
```
yannham, line 44: One slightly annoying thing here is that we need to `cargo fmt` with nightly. I wonder if we should provide a `nightly` toolchain pinned as well, or even maybe propose a "libdd-fmt" alias that uses the right nightly toolchain under the hood. It's a detail though, and could be worked out in a follow up PR.

flake.nix
yannham: I'm not sure, but it's probably not critical for a first PR, since the headers aren't checked in libdatadog manually but are generated automatically when building the FFI (thus using a pinned version). Having `cbindgen` is nice for local development if you are curious about the generated C interface but it's indicative I believe (plus the fact that I expect cbindgen output to be stable?).

yannham:
Do you know what are those Test Nix / test report jobs, or where they are coming from? They have no logs and a Windows one doesn't make sense

image

flake.nix
lloeki: Coincidentally I just added a nightly toolchain as `.#nightly`.

Maybe we can even work out something that would have a closure over the nightly toolchain as an internal dependency for a nightly-based `cargo fmt` - and only that - to be exposed. That's exactly the kind of thing Nix can solve.

Indeed probably for a later PR.

lloeki:
> Do you know what are those Test Nix / test report jobs, or where they are coming from? They have no logs and a Windows one doesn't make sense
> 
> image

Interesting. I'm not sure, let me look that up.

.github/workflows/nix.yml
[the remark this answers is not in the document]
ekump: You're probably ok, since you're compiling a single build. The test workflow is doing multiple, and also doing extra stuff like compiling test fixtures for crashtracker tests, generating code coverage reports, etc. 

You could temporarily output `df` here to see how much headroom you have and potentially use the free-disk-space action like we do [here](https://github.com/DataDog/libdatadog/blob/952c2ef75cdf7b2895d7152100ea61c12ccf4439/.github/workflows/test.yml#L91) (You probably don't have to be as aggressive as the test workflow is).

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+          name = "libdatadog-devshell";
+
+          # The stdenv cc-wrapper injects -D_FORTIFY_SOURCE, which glibc rejects
+          # when compiling without optimization. Some build scripts (e.g.
+          # spawn_worker's trampoline.c) compile C at -O0 with -Werror, so the
+          # resulting fortify #warning becomes a hard error. Disable fortify
+          # hardening in the shell so those builds succeed.
+          hardeningDisable = [ "fortify" "fortify3" ];
+
+          buildInputs = [
```
ekump: nit: Should this be `nativeBuildInputs` since they are build-time host tools?

flake.nix
```diff
@@ -0,0 +1,59 @@
+{
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/release-25.11";
+
+    # cross-platform convenience
+    flake-utils.url = "github:numtide/flake-utils";
+
+    # backwards compatibility with nix-build and nix-shell
+    flake-compat.url = "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz";
```
ekump, line 12: nit: Is this necessary? `shell.nix` or `default.nix` aren't included in this repo (maybe they will be in the future?).

.github/workflows/nix.yml
ekump: You could also use a concurrency group ([example from our test workflow](https://github.com/DataDog/libdatadog/blob/952c2ef75cdf7b2895d7152100ea61c12ccf4439/.github/workflows/test.yml#L268-L270))and cancel in progress if you don't need commit level granularity.

Not necessarily a hill I want to die on, but I think using a concurrency group is a good compromise between getting signal that something broke and not potentially backing up the MergeQueue waiting for a macOS runner for a workflow that doesn't indicate an actual code issue.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,69 @@
...
+          - os: linux
+            cpu: aarch64
+            base: ubuntu-24.04-arm # always aarch64-linux-gnu
+
+    name: Test Nix (${{ matrix.platform.cpu }}-${{ matrix.platform.os }})
+    runs-on: ${{ matrix.platform.base }}
+
+    permissions:
+      contents: read
+      id-token: write
```
ekump: Is this necessary? could you get away with `contents: read`?

.github/workflows/nix.yml
```diff
@@ -0,0 +1,69 @@
+name: Test Nix
```
ekump, line 1: Can we add documentation (maybe [here](https://github.com/DataDog/libdatadog/tree/952c2ef75cdf7b2895d7152100ea61c12ccf4439#running-tests) and / or inline in this file) with instructions on what to do if there is a failure? I suspect it'll only come up on MSRV bumps.

ekump (approved):
Left a few nit comments, but I'm not a nix guy so not sure how relevant they are.

LGTM after we add a little bit of documentation and add a concurrency group for the workflow.

flake.nix
lloeki: To clarify usage as of today:
- `nix develop .#nightly` to enter a shell with the nightly toolchain
- `nix develop .#nightly -c cargo fmt` to run its nightly `cargo` right away

flake.nix
lloeki: True-ish: it's only really relevant for cross-compilation so it doesn't really matter for devshells as they are always running on the host platform, although there's some subtlety like xdg paths and such.

I'll amend for the sake of good practice.

flake.nix
lloeki: Ha, I forgot to add the shims, will add.

.github/workflows/nix.yml
lloeki: I think it would be fine without it indeed.

.github/workflows/nix.yml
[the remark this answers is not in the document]
lloeki: The free disk space action makes sense, especially since everything that matters would later come from Nix.

.github/workflows/nix.yml
lloeki: I'm not too sure what the expectation of the doc would be: some basic usage examples to onboard the Nix-unaware / a LLM agent maybe?

I did not want to be too invasive as a first PR but seems like you folks are quite on board with this so I feel like the doc you pointed to would be a good place. Will add.

.github/CODEOWNERS
lloeki: It's a small but growing group. Join us at `#nix`!

lloeki (commented):
Thanks for the astute reviews! Will address.

lloeki:
Ok so I looked up the `test report` ones:
- they're not emitted by `nix.yml`
- they apparently come from `mikepenz/action-junit-report` in `test.yml` (from `check_name` being `[${{ matrix.platform }}:${{ steps.rust-version.outputs.version }}] test report` + the matrix including windows)
- why GitHub decided they should appear under `Test Nix`, that I couldn't figure out

.github/workflows/nix.yml
```diff
@@ -0,0 +1,69 @@
...
+
+on: # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - main
+      - mq-working-branch-*
```
hoolioh, line 7:
```suggestion
 # Nightly safeguard: catches devshell drift from changes that don't match the
  # paths above (e.g. a dependency or build-script change that needs new native
  # tooling). Scheduled runs only fire from the default branch, against its HEAD.
  schedule:
    - cron: "0 4 * * *" # daily at 04:00 UTC
```

hoolioh: Instead of triggering it in every PR I think it would be better to have a nightly schedule. That way we can save some capacity for the actual pipelines.

yannham: But wouldn't you want to check tings on PRs that change Nix files at least?

lloeki: Definitely.

As is it's not going to trigger in every PR anyway, only those that change nix files or that have an impact on nix (the toolchain files).

The schedule can be nice though as a sanity check for when those don't change often.

flake.nix
```diff
@@ -0,0 +1,62 @@
+# Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+# SPDX-License-Identifier: Apache-2.0
+
+{
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/release-25.11";
```
Eldolfin: the `26.05` channel is the latest stable, why not use it instead ?
And what about using `unstable` instead ?

Eldolfin:
Hey I wanted to suggest many small changes (mostly styling/preferences), I created a [stacked PR](https://github.com/DataDog/libdatadog/pull/2136/changes) instead because it got too large

flake.nix
yannham: Unstable is nice to have the very latest version, but I think you won't hit the binary cache, which might mean a lot of full local rebuilds from source. The latest stable makes sense though :+1:

Eldolfin: Does `unstable` really makes a difference with binary cache ? I never hit that issue personally.

yannham: Unstable is actually cached, but it moves fast. I vaguely remember situations such as user A updates the flake.lock, 1 month passes, new user B checks the project out from scratch and inherits the same flake.lock, but stuff isn't in the unstable cache anymore. If you are the one updating you shouldn't have issues ofc, because when you check out the package, they are the most recent available and thus in cache. I think. I am not sure. Both latest stable and unstable are probably reasonable.

Another reason could be if you run NixOS, you usually target a specific release, and you probably don't want to duplicate many versions of core packages (the stable and the unstable one).

lloeki: - pro for unstable: you don't have to think about the source, just flake update (ironically I fell into this very trap since copied a flake that had an outdated stable source!)
- con against unstable: what Yann said, plus it's bleeding edge rolling release (although that's usually more of an issue with NixOS modules than nixpkgs packages)

 > 1 month passes, new user B checks the project out from scratch and inherits the same flake.lock, but stuff isn't in the unstable cache anymore.

Often not a problem but that's the idea yes.

> If you are the one updating you shouldn't have issues ofc, because when you check out the package, they are the most recent available and thus in cache.

IIRC only partly true, if it's a fresh commit you can have a cache miss as the commit landed but cache may not be filled yet. This is compounded by the fact that unstable moves quickly so whenever you update you have a higher chance of getting a miss.

> if you run NixOS, you usually target a specific release, and you probably don't want to duplicate many versions of core packages (the stable and the unstable one).

Even if you're on stable on both sides, the closure is entirely distinct so you can absolutely end up with duplication. There's just no guarantee (or rather, the guarantee is that the package you get is the exact build that was closed over), more like statistical happenstance.

lloeki: Picked 26.05

.github/workflows/nix.yml
lloeki: Added schedule

lloeki:
@Eldolfin thanks for the suggestions over there. I take it you made a separate PR because of the formatting.

I've swallowed a few of the tiny interesting cleanup bits in this PR.

To expedite this PR, I will leave the formatting and rust tooling decisions to be addressed on a subsequent PR, which you'll be able to resolve without my involvement.

.github/workflows/nix.yml
hoolioh: But as I undestand the workflow, the job will be executed whenever any of the files involved is modified as in:
```
pull_request:
    paths:
      - "*.nix"
      - "flake.*"
      - "rust-toolchain.toml"
      - ".github/workflows/nix.yml"  
```
Then it will be executed nightly as a safe guard so I think there's no need to trigger it in every PR on the mq/main.

.github/workflows/nix.yml
ekump: I may have been misunderstanding the test. If it can only fail if you change nix related files then there isn't much to say. If it can fail due to other changes (like if we bump our MSRV or otherwise modify our toolchain) then we'd just need instructions on how to fix.

.github/workflows/nix.yml
lloeki: Yes for pull requests, but not for pushes to `main` (e.g merge commit) or merge queue (so it will not fire and maybe fail a possible required check constraint; I'm not sure of MQ behaviour here)

If that's acceptable then it can be removed.

.github/workflows/nix.yml
lloeki: > If it can fail due to other changes (like if we bump our MSRV or otherwise modify our toolchain) then we'd just need instructions on how to fix.

Such instructions have been added to the doc (section `Debugging CI failures`), for the most predictable failure case, which would be whenever the toolchain is bumped then the flake _may_ need an update to pick up a presumably not-yet-known version at previous lock time.

flake.nix
yannham: Mostly agreed with picking a release :+1: (at worse people can override flake.nix and flake.lock locally if they really want to).

> Even if you're on stable on both sides, the closure is entirely distinct so you can absolutely end up with duplication. There's just no guarantee (or rather, the guarantee is that the package you get is the exact build that was closed over), more like statistical happenstance.

I mean, if you're pulling the same package as a dependency of the shell from the same release, it will definitely be the same derivation and come from cache? If you override it or customize it in any non trivial way, that's a different question.

PR #2130, opened 2026-06-17
chore(stats)!: submit p0 telemetry in stats

VianneyRuhlmann:
# What does this PR do?

Allow sending css drop telemetry from the stats module to avoid returning the counts just for them to be included in the send payload telemetry struct.

# Motivation

Simplify telemetry flow for css

# Additional Notes

Remove `dropped_by_trace_filter` from DroppedStats struct as it's no longer used

# How to test the change?

### DIFF
--- a/libdd-data-pipeline/src/telemetry/mod.rs
+++ b/libdd-data-pipeline/src/telemetry/mod.rs
@@ -192,7 +192,6 @@ impl SendPayloadTelemetry {
     /// * `value` - The result of sending traces with retry
     /// * `bytes_sent` - The number of bytes in the payload
     /// * `chunks` - The number of trace chunks in the payload
-    /// * `chunks_dropped_p0` - The number of P0 trace chunks dropped due to sampling
     pub fn from_retry_result(value: &SendWithRetryResult, bytes_sent: u64, chunks: u64) -> Self {
         let mut telemetry = Self::default();
         match value {
--- a/libdd-data-pipeline/src/telemetry/mod.rs
+++ b/libdd-data-pipeline/src/telemetry/mod.rs
@@ -322,6 +322,16 @@ impl TelemetryClient {
         Ok(())
     }
 
+    /// Send dropped P0 trace counts to telemetry.
+    pub fn send_p0_drops(&self, dropped_p0_traces: u64) -> Result<(), TelemetryError> {
+        if dropped_p0_traces > 0 {
+            let key = self.metrics.get(metrics::MetricKind::ChunksDroppedP0);
+            self.worker
+                .add_point(dropped_p0_traces as f64, key, vec![])?;
+        }
+        Ok(())
+    }
+
     /// Starts the client
     pub async fn start(&self) {
         _ = self
--- a/libdd-data-pipeline/src/telemetry/mod.rs
+++ b/libdd-data-pipeline/src/telemetry/mod.rs
@@ -13,7 +13,6 @@ use libdd_telemetry::worker::{
 };
 use libdd_trace_utils::{
     send_with_retry::{SendWithRetryError, SendWithRetryResult},
-    span::trace_utils::DroppedStats,
     trace_utils::SendDataResult,
 };
 use std::{collections::HashMap, time::Duration};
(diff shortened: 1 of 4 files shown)


libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -300,13 +300,19 @@ fn add_spans_to_stats<T: libdd_trace_utils::span::TraceData>(
 
 /// Process traces for stats computation and update header tags accordingly.
 /// Returns the number of P0 traces and spans that were dropped.
```
hoolioh:
```suggestion
```

libdd-data-pipeline/src/telemetry/mod.rs
```diff
@@ -195,18 +192,9 @@ impl SendPayloadTelemetry {
...
-        bytes_sent: u64,
-        chunks: u64,
-        dropped_stats: DroppedStats,
-    ) -> Self {
-        let mut telemetry = Self {
-            chunks_dropped_p0: dropped_stats.dropped_p0_traces as u64,
-            chunks_dropped_by_trace_filter: dropped_stats.dropped_by_trace_filter as u64,
-            ..Default::default()
-        };
+    /// * `chunks_dropped_p0` - The number of P0 trace chunks dropped due to sampling
```
hoolioh: I guess this is argument is not needed/used?

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -322,22 +328,24 @@ pub(crate) fn process_traces_for_stats<T: libdd_trace_utils::span::TraceData>(
...
-        libdd_trace_utils::span::trace_utils::DroppedStats {
-            dropped_p0_traces: 0,
-            dropped_p0_spans: 0,
-            dropped_by_trace_filter: 0,
+        header_tags.dropped_p0_traces = dropped_p0_stats.dropped_p0_traces;
+        header_tags.dropped_p0_spans = dropped_p0_stats.dropped_p0_spans;
+
+        // Send dropped P0 stats directly to telemetry if available
+        #[cfg(all(not(target_arch = "wasm32"), feature = "telemetry"))]
+        if let Some(telemetry_client) = telemetry {
```
hoolioh, line 341: Would it be a good idea to add a test asserting that stats are sent?

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -300,13 +300,19 @@ fn add_spans_to_stats<T: libdd_trace_utils::span::TraceData>(
...
+/// If a telemetry client is provided and stats are enabled, dropped P0 counts
+/// will be sent to telemetry.
 pub(crate) fn process_traces_for_stats<T: libdd_trace_utils::span::TraceData>(
     traces: &mut Vec<Vec<libdd_trace_utils::span::v04::Span<T>>>,
     header_tags: &mut libdd_trace_utils::trace_utils::TracerHeaderTags,
     client_side_stats: &ArcSwap<StatsComputationStatus>,
     client_computed_top_level: bool,
     trace_filterer: &TraceFilterer,
-) -> libdd_trace_utils::span::trace_utils::DroppedStats {
+    #[cfg(all(not(target_arch = "wasm32"), feature = "telemetry"))] telemetry: Option<
```
Eldolfin, line 311: I think trying to build for wasm with the telemetry feature could be a compile error so that we don't need to handle it everywhere there is telemetry ?

Right now trying to build for wasm with telemetry fails with tons of places where we didn't properly gate code parts like you did here.

Something like this in `lib.rs`
```rs
#[cfg(target = "wasm32", feature = "telemetry")]
compile_error!("Feature 'telemetry' cannot be enabled on wasm32 targets");
```

It still spits out all the compile errors but at least the first one clearly states the problem.

Eldolfin (approved):

libdd-data-pipeline/src/trace_exporter/stats.rs
VianneyRuhlmann: I thought about but a bit out of scope I'll do it in follow up

libdd-data-pipeline/src/trace_exporter/stats.rs
VianneyRuhlmann: I think it would be a bit heavy. There's a test with http mock in the telemetry component, I don't think it should be up to the caller to check that telemetry correctly sends the telemetry.

PR #2131, opened 2026-06-17
ci: add manual job to trigger dd-trace-py downstream pipeline

faydef:
## What

Adds a manual GitLab CI job, `trigger_dd_trace_py`, that triggers the downstream `DataDog/apm-reliability/dd-trace-py` pipeline on demand from the GitLab pipeline UI.

- New `.gitlab/dd-trace-py.yml` defining the `trigger_dd_trace_py` job (`when: manual`, `allow_failure: true`, `strategy: depend`).
- Configurable target branch via the `DD_TRACE_PY_BRANCH` pipeline variable (defaults to `fayssal/auto-update-libdatadog`), exposed as a UI element.
- Passes the current libdatadog commit metadata plus `UPDATE_LIBDATADOG_VERSION="true"` downstream.

## Why

The aim is to have dd-trace-py **auto-update the libdatadog version it uses** (driven by `UPDATE_LIBDATADOG_VERSION`) so we can run benchmarks against the in-flight libdatadog commit.

> [!WARNING]
> This is **experimental** — the wiring and variables may change as we iterate.

## Related

Matching dd-trace-py PR: https://github.com/DataDog/dd-trace-py/pull/18660

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,10 +4,14 @@ variables:
   DOWNSTREAM_BRANCH:
     value: "main"
     description: "downstream jobs are triggered on this branch"
+  DD_TRACE_PY_BRANCH:
+    value: "fayssal/auto-update-libdatadog"
+    description: "the dd-trace-py branch the manual trigger_dd_trace_py job runs on"
 
 include:
   - local: .gitlab/benchmarks.yml
   - local: .gitlab/fuzz.yml
+  - local: .gitlab/dd-trace-py.yml
 
 trigger_internal_build:
   variables:
--- a/.gitlab/dd-trace-py.yml
+++ b/.gitlab/dd-trace-py.yml
@@ -8,6 +8,7 @@ trigger_dd_trace_py:
     LIBDATADOG_COMMIT_SHA: $CI_COMMIT_SHA
     LIBDATADOG_SHORT_COMMIT_SHA: ${CI_COMMIT_SHORT_SHA}
     LIBDATADOG_COMMIT_TAG: $CI_COMMIT_TAG
+    UPDATE_LIBDATADOG_VERSION: "true"
   rules:
     - when: manual
       allow_failure: true
--- a/.gitlab/dd-trace-py.yml
+++ b/.gitlab/dd-trace-py.yml
@@ -1,0 +1,17 @@
+# Manual trigger for the dd-trace-py downstream pipeline.
+# This lets a maintainer test the current libdatadog commit against dd-trace-py
+# on demand from the GitLab pipeline UI.
+
+trigger_dd_trace_py:
+  variables:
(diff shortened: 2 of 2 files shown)


.gitlab/dd-trace-py.yml
```diff
@@ -0,0 +1,23 @@
...
+trigger_dd_trace_py:
+  variables:
+    LIBDATADOG_COMMIT_BRANCH: $CI_COMMIT_BRANCH
+    LIBDATADOG_COMMIT_SHA: $CI_COMMIT_SHA
+    LIBDATADOG_SHORT_COMMIT_SHA: ${CI_COMMIT_SHORT_SHA}
+    LIBDATADOG_COMMIT_TAG: $CI_COMMIT_TAG
+    UPDATE_LIBDATADOG_VERSION: "true"
+  rules:
+    # Run automatically on every push to the default branch.
+    - if: '$CI_COMMIT_BRANCH == "main"'
```
yannham, line 14: This doesn't actually run on every commit to the main branch, because the job on dd-trace-py is configured to skip every X commits, correct? If it is, maybe we should mention it here, to not scare people into thinking we'll be running the downstream pipeline and benchmarks of multiple tracers in the future on every commit to main 😰

.gitlab/dd-trace-py.yml
```diff
@@ -0,0 +1,23 @@
...
+      when: on_success
+      allow_failure: true
+    # Otherwise allow a maintainer to trigger it manually from the pipeline UI.
```
yannham, line 17: So this correspond to adding a label to the PR? Or what does "pipeline UI" refers to here?

.gitlab-ci.yml
```diff
@@ -4,10 +4,14 @@ variables:
   DOWNSTREAM_BRANCH:
     value: "main"
     description: "downstream jobs are triggered on this branch"
+  DD_TRACE_PY_BRANCH:
```
yannham, line 7: Is this branch fixed, or can we change the value of this var somehow? I suppose we want to use the same branch for continuous updating but that you'll want a temporary branch for specific comparisons.

yannham: It's actually written in the description. Sorry for the noise.

.gitlab/dd-trace-py.yml
faydef: The UI here refers to Gitlab, the mechanism to trigger from a label will be added by Augusto later today!

.gitlab/dd-trace-py.yml
faydef: As is it would run on every commit but be no-op (on dd-trace-py's side) if less than x (default=5 but maybe I was being a bit too greedy here 😅 )

yannham:
This whole line of work is not dead, but we are several people working on a common "auto-update" job now for different reasons. I think we can close this for now and re-open once we agreed on the downstream dd-trace-py job, as the trigger might look different.

PR #2133, opened 2026-06-18
ci: allow dd-trace-*/libdatadog to write pr comments on libdatadog

igoragoli:
# What does this PR do?

Allows dd-trace-* and libdatadog to write PR comments on libdatadog.

# Motivation

libdatadog will run benchmarks by triggering downstream dd-trace-* pipelines. Those pipelines need to report results back to libdatadog. One way to report results back is through PR comments.

# Additional Notes

This policy can be tightened to make it absolutely secure, but the risks it incurs (an attacker might create PR comments on branches on dd-trace-* repos' CIs) are too low compared to the complexity of setting up and testing a more fleshed-out policy. 

I'd be happy to update the policy if these risks are something we should address :)

# How to test the change?

Should be verified ad-hoc with prototype PR comments, due to how dd-octo-sts works. Changes need to be on `main` to be effective.

### DIFF
--- a/.github/chainguard/gitlab.write.benchmarking-pr-comments.sts.yaml
+++ b/.github/chainguard/gitlab.write.benchmarking-pr-comments.sts.yaml
@@ -1,0 +1,6 @@
+issuer: https://gitlab.ddbuild.io
+
+subject_pattern: "project_path:DataDog/apm-reliability/(dd-trace-.*|libdatadog):.*"
+
+permissions:
+  pull_requests: write


.github/chainguard/gitlab.write.benchmarking-pr-comments.sts.yaml
```diff
@@ -0,0 +1,6 @@
+issuer: https://gitlab.ddbuild.io
+
+subject_pattern: "project_path:DataDog/apm-reliability/(dd-trace-.*|libdatadog):.*"
+
+permissions:
+  pull_requests: write
```
ekump, line 6:
```suggestion
  subject_pattern: "project_path:DataDog/apm-reliability/dd-trace-[a-z0-9-]+:ref_type:branch:ref:(main|master)"                                                                                                                                                                         
                                                                                                                                                                                                                                                                                        
  claim_pattern:                                                                                                                                                                                                                                                                        
    project_path: "DataDog/apm-reliability/dd-trace-[a-z0-9-]+"                                                                                                                                                                                                                         
    ref_type: "branch"                                                                                                                                                                                                                                                                  
    ref_protected: "true"                                                                                                                                                                                              
    pipeline_source: "pipeline"                                                                                                                                                                                                                     
    ci_config_ref_uri: "gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-[a-z0-9-]+//.gitlab-ci.yml@refs/heads/(main|master)"                                                                                                                                                         
                                                                                                                                                                                                                                                                                        
  permissions:                                                                                                                                                                                                                                                                          
    pull_requests: write                                                                                                                                                                                                                                                                                  
```

The policy as-is is too lax. I don't think libdatadog should be included as a subject as we already have benchmark jobs triggered internally and use a different existing policy. They shouldn't share a policy with this workflow.

I don't know how these benchmarks are going to work, but my suggestion assumes they are going to run on the default branch of the SDKs.

yannham: The current plan is to have those benchmarks run on some branch of the SDK/tracers (but not the default one, since the idea is to auto update the SDK to some version of libdatadog corresponding to a PR, so this will be on a new, dedicated branch). At some point this SDK benchmark job needs to write back a comment here to report the result, which I think is the goal of this policy change. Hopes that clarifies. But also maybe there's a better way to orchestrate all of this, so open to suggestions here!

ekump: Got it. If it's a dedicated branch then my suggestion would still mostly work. You'd just need to update it to match on that branch name instead of main/master.

yannham:
(rationale for the keep-open: this is blocked on other work on the auto update job, but that work is still ongoing/active)

PR #2136, opened 2026-06-18
suggest: large suggestion to flake.nix

Eldolfin:
A large suggestions for https://github.com/DataDog/libdatadog/pull/2128

- format with alejandra for better readability
- add missing cargo-nextest, rust-src component (for lsp), alejandra
- use specialized pkg.mkShell instead of mkDerivation
- ~remove hardeningDisable, not sure it's needed~

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -40,7 +40,7 @@
           # hardening in the shell so those builds succeed.
           hardeningDisable = [ "fortify" "fortify3" ];
 
-          buildInputs = [
+          nativeBuildInputs = [
             rust            # rustc + cargo + rustfmt + clippy, pinned via toolchain file
             pkgs.rust-cbindgen
             pkgs.cmake
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,46 @@
+{
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/release-25.11";
+
+    # cross-platform convenience
+    flake-utils.url = "github:numtide/flake-utils";
+
+    # backwards compatibility with nix-build and nix-shell
+    flake-compat.url = "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz";
+
+    # pinned, exact upstream Rust toolchains
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = { self, nixpkgs, flake-utils, flake-compat, rust-overlay }:
+    # resolve for all platforms in turn
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        # packages for this system platform, with the rust-overlay applied
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [ (import rust-overlay) ];
+        };
+
+        # pinned Rust toolchain; single source of truth is ./rust-toolchain.toml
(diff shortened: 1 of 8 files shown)


flake.nix
```diff
@@ -1,62 +1,61 @@
...
+            packages = [
+              rust-toolchain
+              pkgs.rust-cbindgen
+              pkgs.cargo-nextest
+              pkgs.cmake
+              pkgs.autoconf
+              pkgs.automake
+              pkgs.libtool
+              pkgs.alejandra
+            ];
```
yannham, line 53: Nit

```suggestion
            packages = with pkgs; [
              rust-toolchain
              rust-cbindgen
              cargo-nextest
              cmake
              autoconf
              automake
              libtool
              alejandra
            ];
```

flake.nix
```diff
@@ -1,62 +1,61 @@
...
-            pkgs.rust-cbindgen
-            pkgs.cmake
-            pkgs.autoconf
-            pkgs.automake
-            pkgs.libtool
-          ];
+          overlays = [(import rust-overlay)];
         };
+        mkShell = rust-toolchain:
+          pkgs.mkShell {
```
yannham, line 32: I totally missed we didn't use `mkShell` in the previous PR, but just `mkDerivation`. This is the way to go indeed 👍

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
 channel = "1.87.0"
-components = ["rustfmt", "clippy"]
+components = ["clippy", "rust-src"]
```
yannham, line 3: Why this change? Doesn't this impact non Nix-users in some way?

Eldolfin: Because we never use the stable rustfmt anyway, also we need `rust-src` for `rust-analyzer` to allow `go-to-definition` to std functions.

I feel like the only impact it could've had is CI failing because it could not find `rustfmt` but as said above we don't use the stable `rustfmt`

flake.nix
Eldolfin: I just changed it a bit to separate rust-toolchain in a separate list in case it ever becomes a package in nixpkgs so that it doesn't get overriden.

flake.nix
lloeki: yeah I use `mkDerivation` often to do things like use clang instead of gcc on Linux or select the llvm version.

flake.nix
```diff
@@ -41,16 +41,17 @@
...
-            ];
+            packages =
+              [rust-toolchain]
+              ++ (with pkgs; [
+                rust-cbindgen
+                cargo-nextest
+                cmake
+                autoconf
+                automake
+                libtool
```
lloeki: TBH I don't see how that's better than the previous code.

This reads like DRY for DRY sake.

Eldolfin: I agree, I also prefer the explicit `pkgs.`, reverting this change

yannham: `with` has it downsides but in this is one case that is simple and clear, and it's rather idiomatic in the Nix community. Though I don't really care and don't have a strong opinion either way :shrug: 

> This reads like DRY for DRY sake.

Isn't this a good thing, as long as it's still readable, clear, and simple? For example we avoid `std::collections::HashMap` everywhere in a Rust file,  because it's clutter, we do `use std::collections::HashMap` once and use `HashMap` elsewhere. I'm not sure how this is very different.

Eldolfin:
Most useful changes were applied in the original PR. closing this one

PR #2145, opened 2026-06-22
feat(trace-utils)!: add encoder v1 to v04 + refactor

anais-raison:
# What does this PR do?

Adds a V1 → v0.4 msgpack downgrade encoder so a tracer that emits V1 spans natively can still send traces to an agent that only advertises `/v0.4/traces`.

Also renames the existing encoder modules/functions so the convention is uniform: **parent module = output wire format, file/function suffix = input span type**. The four combinations now coexist:

- `v04/span_v04.rs` — v0.4 → v0.4 (native)
- `v04/span_v1.rs`  — v1   → v0.4 (downgrade, new)
- `v1/span_v04.rs`  — v0.4 → V1   (upgrade)
- `v1/span_v1.rs`   — v1   → V1   (native)

# Motivation

[APMSP-2811](https://datadoghq.atlassian.net/browse/APMSP-2811)

[APMSP-2811]: https://datadoghq.atlassian.net/browse/APMSP-2811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1045 @@
...
+    match kind {
+        SpanKind::Internal => None,
+        SpanKind::Server => Some("server"),
+        SpanKind::Client => Some("client"),
+        SpanKind::Producer => Some("producer"),
+        SpanKind::Consumer => Some("consumer"),
+    }
+}
+
+/// Splits a 128-bit big-endian trace_id into `(low_64, high_64)`. The low half maps to v0.4's
```
yannham: Very nit, but if not precised I would expect the u64 to be using the platform endianness.

```suggestion
/// Splits a 128-bit big-endian trace_id into big-endian `(low_64, high_64)`. The low half maps to v0.4's
```

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1045 @@
...
+    counts.meta += (trace_id_high != 0) as u32;
+    counts.meta += !chunk.origin.borrow().is_empty() as u32;
+    counts.meta += chunk.sampling_mechanism.is_some() as u32;
+    counts.metrics += chunk.priority.is_some() as u32;
+    for (_, v) in span_attrs_dd.iter() {
+        counts.count_attr(v);
+    }
+    // Chunk attributes are propagated to every span. Duplicate keys between chunk and span are
+    // tolerated here — msgpack map decode keeps the last value (the span's), which matches the
+    // "span overrides chunk" precedence.
```
yannham: Is this true? I remember asking a similar question (can we have duplicate entries in e.g. metrics or meta maps of one span encoded through msgpack) and I think the answer from @ajgajg1134 was that this was the _current_ behavior of the agent indeed, but is not formalized in the spec so it could change in the future. Maybe this is not exactly the same case, but just to be sure

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1045 @@
...
+        }
+        for (k, v) in chunk_attrs_dd.iter() {
+            match v {
+                AttributeValue::String(s) => {
+                    write_str(writer, k.borrow())?;
+                    write_str(writer, s.borrow())?;
+                }
+                AttributeValue::Bool(b) => {
+                    write_str(writer, k.borrow())?;
+                    write_str(writer, if *b { "true" } else { "false" })?;
```
yannham: Nit: although it's small, we use `if *b { "true" } else { "false" }` several times. Should we have a small helper for that? Like `write_bool`?

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1045 @@
...
+    v: &AttributeValue<T>,
+) -> Result<(), ValueWriteError<W::Error>> {
+    match v {
+        AttributeValue::Bytes(b) => {
+            write_str(writer, k.borrow())?;
+            write_bin(writer, b.borrow())?;
+        }
+        AttributeValue::KeyValue(_) | AttributeValue::List(_) => {
+            write_str(writer, k.borrow())?;
+            write_bin(writer, &[])?;
```
yannham: One way to not forget the TODO 🙂 

```suggestion
            write_bin(writer, todo!())?;
```

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1045 @@
...
+        }
+        _ => {}
+    }
+    Ok(())
+}
+
+/// Encodes [`v1::SpanLink`](crate::span::v1::SpanLink)s into the v0.4 msgpack wire format
+/// (downgrade: v1 input → v0.4 output). The 128-bit `trace_id` is split into
+/// `(trace_id, trace_id_high)` u64s. Typed link attributes are downgraded to strings (`String`
+/// passes through, `Bool` becomes `"true"` / `"false"`); non-string-coercible variants are
```
yannham:
```suggestion
/// `(trace_id, trace_id_high)` u64s. Typed link attributes are downgraded to strings; non-string-coercible variants are
```

yannham (commented):
LGTM overall 👍 I think the naming convention brings more clarity indeed.

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
anais-raison: Good catch ! I refactored the encoder to merge span and chunk attributes upfront (span overrides chunk) and now each key appears exactly once on the wire. No more reliance on msgpack last-write-wins.

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -168,3 +226,72 @@ pub fn to_encoded_byte_len<T: TraceData, S: AsRef<[Span<T>]>>(traces: &[S]) -> u
     let _ = to_writer(&mut counter, traces);
     counter.0
 }
+
+/// Encodes a [`TracerPayload`] in the v0.4 wire format (downgrade path used when the agent
+/// does not advertise `/v1.0/traces`). The output is a msgpack array of traces, where each
+/// trace is itself a msgpack array of v0.4-shaped spans — matching the existing v0.4 wire
+/// format produced by [`to_vec`].
+fn encode_payload_v1<W: RmpWrite, T: TraceData>(
```
ajgajg1134: This function is named `encode_payload_v1` but the comment says it outputs in v0.4 format? I think the names here are a bit confusing as it seems all of the `_v1` functions actually emit v04? 🤔

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1184 @@
...
+//! must fall back to v0.4. The mapping is:
+//!
+//! | v1::Span field / attribute            | v0.4 field                                  |
+//! |---------------------------------------|---------------------------------------------|
+//! | `env` / `version` / `component`       | `meta["env"]` / `meta["version"]` / ...     |
+//! | `span_kind`                           | `meta["span.kind"]` (lowercase string)      |
+//! | `AttributeValue::String` / `Bool`     | `meta[k]` (`"true"` / `"false"` for bool)   |
+//! | `AttributeValue::Float` / `Int`       | `metrics[k]` (Int cast to `f64`)            |
+//! | `AttributeValue::Bytes` / `KeyValue`  | `meta_struct[k]` (re-encoded as msgpack)    |
+//! | `AttributeValue::List`                | `meta_struct[k]` (re-encoded as msgpack)    |
```
ajgajg1134: I think that for downgrading lists and keyvalue types we have to explode the keys out to the way the UI expects them (This is what intake will also need to do). e.g. for a list `key.0, key.1,...` and for KeyValue it ends up being separate flat keys `key.keyB.keyC:value`. Let me know if you need some more examples here or if the v1 docs need to be updated!

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
```diff
@@ -315,7 +316,7 @@ where
 ///   trace_key::CHUNKS       (11) → Array[Chunk, ...]
 /// }
 /// ```
-fn encode_payload<W: RmpWrite, T: TraceData, S: AsRef<[Span<T>]>>(
+fn encode_payload_v04<W: RmpWrite, T: TraceData, S: AsRef<[Span<T>]>>(
```
ajgajg1134: I think maybe the names are flipped, this function is named _v04 but the comment says v1

ajgajg1134 (changes requested):
Some of the function names are confusing me a bit / don't match the comments?
Also I think some of the v1 attribute types need to be "un-done" differently, specifically key values and lists shouldn't just get put in meta_struct and should be flattened into individual meta/metrics entries.

Once you have that changed, we should validate this with some integration tests against the real agent / intake to make sure the payloads match the agent decode and appear correctly in the UI

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -2,11 +2,69 @@
...
+        storage[0] = 0xda;
+        2
+    } else {
+        storage[0] = 0xdb;
+        4
+    };
+    copy_to_slice(storage.split_at_mut(1).1, &len.to_be_bytes(), len_bytes);
+    copy_to_slice(storage.split_at_mut(1 + len_bytes).1, s.as_bytes(), s.len());
+    storage
+}
```
Aaalibaba42: nit: I liked the previous order of the definitions better (like macro after the functions the macro uses)

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
Aaalibaba42: In the `v04` module, functions suffixed `_v1` being the input span type seem "reasonable" to my mind. It might be better to clarify this somewhere other than just [here](https://github.com/DataDog/libdatadog/pull/2145/changes#diff-2063d0efe97d76e5eceb4cf1f91f2a8dfe30998053a9d47c0af8f7471e619b90R4-R7)

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1184 @@
...
+
+    if !span.span_links.is_empty() {
+        encode_span_links(writer, &span.span_links)?;
+    }
+    if !span.span_events.is_empty() {
+        encode_span_events(writer, &span.span_events)?;
+    }
+
+    Ok(())
+}
```
Aaalibaba42, line 445: Maybe this needs to be iterative like this, at a glance there are variations in some segments, but it "feels" a bit dirty, idk if macro magic can help but it's clearly just a styling nit

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1184 @@
...
+            write_const_msg_pack_str!(writer, "int_value")?;
+            write_sint(writer, *i)?;
+        }
+        AttributeValue::Float(f) => {
+            write_map_len(writer, 2)?;
+            write_const_msg_pack_str!(writer, "type")?;
+            write_u8(writer, 3)?;
+            write_const_msg_pack_str!(writer, "double_value")?;
+            write_f64(writer, *f)?;
+        }
```
Aaalibaba42, line 604: Same with these, maybe repetition can be captured in a macro but idk if it's worth it, at least here it's spelled out clearly. Feel free to disregard

libdd-trace-utils/src/msgpack_encoder/v1/mod.rs
Aaalibaba42: Same as https://github.com/DataDog/libdatadog/pull/2145/changes#r3498000493

libdd-trace-utils/src/msgpack_encoder/v1/span_v1.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
+// Naming convention: parent module (`v1`) = output wire format, file suffix (`_v1`) = input span
+// type. This file encodes a [`crate::span::v1::Span`] into the V1 msgpack wire format (native).
+
```
Aaalibaba42: This comment is helpful but it is repeated all over the place. Maybe having this paragraph just once in a parent module would be better instead of having it in many files.

Aaalibaba42 (commented):
Mostly nits and/or styling, most of these can be disregarded

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -2,11 +2,69 @@
...
+    ($writer:expr, $str:expr) => {{
+        use rmp::encode::ValueWriteError;
+        const STRING_ENCODING_LEN: usize = super::msp_string_encoding_len($str);
+        const STRING_ENCODING: [u8; STRING_ENCODING_LEN] = super::msp_const_string_encoding($str);
+
+        $writer
+            .write_bytes(&STRING_ENCODING)
+            .map_err(ValueWriteError::InvalidDataWrite)
+    }};
+}
```
VianneyRuhlmann: Out of curiosity why does this have to be a macro ?

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -2,11 +2,69 @@
...
+
+mod span_v04;
+mod span_v1;
+
+const fn msp_string_encoding_len(s: &str) -> usize {
```
VianneyRuhlmann: I think we should avoid having several abbreviation for msgpack and `msgpack` seems to be the most widely used in the codebase.

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1184 @@
...
+        AttributeValue::Bytes(b) => {
+            write_str(writer, k.borrow())?;
+            write_bin(writer, b.borrow())?;
+        }
+        AttributeValue::KeyValue(_) | AttributeValue::List(_) => {
+            write_str(writer, k.borrow())?;
+            let mut buf = ByteBuf::with_capacity(0);
+            encode_attribute_value(&mut buf, v)
+                .map_err(super::super::flatten_value_write_infallible)
+                .unwrap_infallible();
```
VianneyRuhlmann: Why do you need to `unwrap_infallible` here can't you propagate the Result up and only unwrap in the top level function ?

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
anais-raison: This was already existing, I just moved it from `span_v04.rs` to `mod.rs` to be able to use it in `span_v1.rs`.

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
anais-raison: Sound good, updated it !

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1184 @@
...
+        }
+        for &(k, v) in &merged_attrs {
+            match v {
+                AttributeValue::String(s) => {
+                    write_str(writer, k.borrow())?;
+                    write_str(writer, s.borrow())?;
+                }
+                AttributeValue::Bool(b) => {
+                    write_str(writer, k.borrow())?;
+                    write_bool_as_str(writer, *b)?;
```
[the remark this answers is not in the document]
anais-raison: Fixed the duplicate-key issue: attributes sharing a name with a dedicated field (`env`, `version`, `component`, `span.kind`, `_dd.p.tid`, `_dd.origin`, `_dd.p.dm`, `_sampling_priority_v1`) are now filtered out before flattening into `meta`/`metrics`. The dedicated field always wins and the colliding attribute is dropped. Also added a regression test covering `span.env = "prod"` vs `attributes["env"] = "staging"`.

@ajgajg1134 does that seem good to you?

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
anais-raison: Done !

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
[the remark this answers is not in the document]
ajgajg1134: Yeah I think as long as we have at least one for each (when set) we should be good

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -168,3 +226,72 @@ pub fn to_encoded_byte_len<T: TraceData, S: AsRef<[Span<T>]>>(traces: &[S]) -> u
...
+/// trace is itself a msgpack array of v0.4-shaped spans — matching the existing v0.4 wire
+/// format produced by [`to_vec`].
+fn encode_payload_v1<W: RmpWrite, T: TraceData>(
+    writer: &mut W,
+    payload: &TracerPayload<T>,
+) -> Result<(), ValueWriteError<W::Error>> {
+    use span_v1::{encode_span, ChunkContext};
+
+    write_array_len(writer, payload.chunks.len() as u32)?;
+    for chunk in &payload.chunks {
```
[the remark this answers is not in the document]
ajgajg1134: hmm I'm a bit confused by this comment. As the trace exporter we shouldn't be dropping trace chunks when `dropped_trace` is true, for v1 _and_ for v4

ajgajg1134 (approved):
The changes look good! The new names are much clearer to me. I do think we need a better integration / system testing story here though, the unit tests do a good job asserting what we think is right here but alignment with the actual trace-agent is harder to be sure of here. Unless that testing step would happen as part of testing actual integration with something like the rust tracer?

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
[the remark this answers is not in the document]
anais-raison: We're not dropping the chunk or its spans. We still send every span normally. The only change is when `dropped_trace`: true, we force `metrics["_sampling_priority_v1"] = -1` (USER_REJECT) as a fallback signal, since v0.4 has no wire equivalent of `dropped_trace`. Does that work for you?

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
anais-raison: The `unwrap_infallible` lefts are both top-level functions now (`to_vec_with_capacity_from_v04`/`to_vec_with_capacity_from_v1`).

anais-raison:
> I do think we need a better integration / system testing story here though, the unit tests do a good job asserting what we think is right here but alignment with the actual trace-agent is harder to be sure of here. Unless that testing step would happen as part of testing actual integration with something like the rust tracer?

@ajgajg1134 

The new encoder isn't wired into the exporter's send path, the current fallback rebuilds v0.4 chunks directly from the original v04 input spans, never through this encoder **_yet_**. So there's no real end-to-end path to test against right now. But I'm planning to wire it in the PR that makes V1 the canonical internal representation for the exporter and to re-test it fully there (and later with system test when integrating in the tracers).

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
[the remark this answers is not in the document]
ajgajg1134: hmmm yeah I suppose that's okay, although I _believe_ `dropped_trace` should only be set by the trace-agent as a way to indicate that we aren't going to send that trace to the backend

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -168,3 +227,90 @@ pub fn to_encoded_byte_len<T: TraceData, S: AsRef<[Span<T>]>>(traces: &[S]) -> u
     let _ = to_writer(&mut counter, traces);
     counter.0
 }
+
+/// Encodes a [`TracerPayload`] in the v0.4 wire format (downgrade path used when the agent
+/// does not advertise `/v1.0/traces`). The output is a msgpack array of traces, where each
+/// trace is itself a msgpack array of v0.4-shaped spans — matching the existing v0.4 wire
+/// format produced by [`to_vec`]. Payload-level `env`/`app_version`/`attributes` are propagated
```
Aaalibaba42: I believe this function was renamed

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -808,6 +808,7 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
                 app_version: self.app_version,
                 runtime_id,
                 service: self.service,
+                container_id: String::new(),
```
Aaalibaba42: What's this exactly ? I don't see anywhere it's changed from an empty string or even used. If this will be utilized by subsequent PRs, maybe not include it in this one ? It seems a little confusing

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1304 @@
...
+#[derive(Default)]
+struct BucketCounts {
+    meta: u32,
+    metrics: u32,
+    meta_struct: u32,
+}
+
+/// Recursively flattens a `List`/`KeyValue` attribute into dotted-key leaf entries for the v0.4
+/// `meta` (string-valued) and `metrics` (numeric) maps — matching how intake/the UI expect nested
+/// V1 attributes to be exploded: list elements become `key.0`, `key.1`, ... and `KeyValue`
```
Aaalibaba42, line 168: Isn't there a collision risk between a key that would literally be named `a.0` ? Like `a = List["a", "b"]` and `a.0 = "MyString"` colliding ? I'm not sure if it's something that could really happen or if that's being dipterophillic

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -2,11 +2,70 @@
...
+        storage[0] = 0xd9;
+        1
+    } else if len <= (u16::MAX as u64) {
+        storage[0] = 0xda;
+        2
+    } else {
+        storage[0] = 0xdb;
+        4
+    };
+    copy_to_slice(storage.split_at_mut(1).1, &len.to_be_bytes(), len_bytes);
```
Aaalibaba42: I don't think that's right, `to_be_bytes` on u64 will give you a bigger array than just 1 byte for a size between 32 and 256, the length will be wrong

Aaalibaba42: This should be tested btw, either it is and I'm wrong, or it isn't and it'd be good to have some.

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1304 @@
...
+///
+/// This function will return any error emitted by the writer.
+pub(super) fn encode_span<W: RmpWrite, T: TraceData>(
+    writer: &mut W,
+    span: &Span<T>,
+    chunk: &ChunkContext<'_, T>,
+) -> Result<(), ValueWriteError<W::Error>> {
+    let span_attrs_dd = span.attributes.defensive_dedup();
+    let chunk_attrs_dd = chunk.attributes.defensive_dedup();
+    let payload_attrs_dd = chunk.payload_attributes.defensive_dedup();
```
Aaalibaba42: Defensive dedup could lead to allocations, it would be better to not compute everything for every span. From my understanding, span_attrs_dd is the only one that needs to be recomputed on every span, isn't it ? You can probably have some kind of ChunkContext struct for the other 2

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1304 @@
...
+    let payload_only: Vec<(&T::Text, &AttributeValue<T>)> = payload_attrs_dd
+        .iter()
+        .filter(|(k, _)| !seen_keys.contains(k))
+        .collect();
+    let merged_attrs: Vec<(&T::Text, &AttributeValue<T>)> = span_attrs_dd
+        .iter()
+        .chain(chunk_only)
+        .chain(payload_only)
+        .filter(|(k, _)| !PROMOTED_ATTR_KEYS.contains(&(*k).borrow()))
+        .collect();
```
Aaalibaba42: Same here, we're in the hotpath and this is allocating a lot with both HashSets etc. I believ this can be done by iterating over the underlying DedupedVecMap. It's probably even faster as long as the map isn't too huge

Aaalibaba42: That's a draft don't merge it as is, but something like this instead of materializing the Vecs:

```suggestion
let attrs_iter = || {
    let is_promoted = |k: &T::Text| PROMOTED_ATTR_KEYS.contains(&k.borrow());
    span_attrs_dd.iter()
        .filter(move |(k, _)| !is_promoted(k))
        .chain(chunk_attrs_dd.iter()
            .filter(move |(k, _)| !span_has(k) && !is_promoted(k)))
        .chain(payload_attrs_dd.iter()
            .filter(move |(k, _)| !span_has(k) && !chunk_has(k) && !is_promoted(k)))
};
```

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1304 @@
...
+        match v {
+            AttributeValue::Bytes(b) => bytes_attrs.push((k, b)),
+            _ => flatten_attr_into(
+                k.borrow().to_owned(),
+                v,
+                &mut meta_leaves,
+                &mut metrics_leaves,
+            ),
+        }
+    }
```
Aaalibaba42, line 300: That's a little more tricky, I don't immediately have something that comes to mind, but surely you don't need that many intermediary allocations of the Vecs and the Strings

libdd-trace-utils/src/msgpack_encoder/v04/span_v1.rs
```diff
@@ -0,0 +1,1304 @@
...
+            write_const_msgpack_str!(writer, "_dd.p.tid")?;
+            write_str(writer, &format!("{trace_id_high:016x}"))?;
+        }
+        if !chunk.origin.borrow().is_empty() {
+            write_const_msgpack_str!(writer, "_dd.origin")?;
+            write_str(writer, chunk.origin.borrow())?;
+        }
+        if let Some(mechanism) = chunk.sampling_mechanism {
+            write_const_msgpack_str!(writer, "_dd.p.dm")?;
+            write_str(writer, &format!("-{mechanism}"))?;
```
Aaalibaba42: `format!` allocates a string, these are small enough to be inlined programatically imho

Aaalibaba42 (commented):
For many of the "perf" comments, I didn't look into everything, but I've only really looked at the v04/span_v1.rs encoder, I wager there is more in other parts of the code, maybe try and have an agent scoure for similar issues. We really ought to be mindful of allocations in hotpath, and std structures like HashMaps, HashSets, Vecs, Strings, etc are heap allocated.

libdd-data-pipeline/src/trace_exporter/builder.rs
anais-raison: I had simply forgot to add this field in the encoders for V1, that's all.

Aaalibaba42: But is it expected that it is *always* an empty String ? Or am I missing some place it gets changed to something else ?

Aaalibaba42 (approved):
I reopened a few things you seem to have closed eagerly without saying why nor fixing the code, but overall it looks good enough after those are either fixed or justified

libdd-data-pipeline/src/trace_exporter/builder.rs
anais-raison: You're right my bad

PR #2149, opened 2026-06-22
fix(tools): collapse intra-file duplicate typedefs in base header

lloeki:
## What

Enhance the `dedup_headers` dev tool so the generated/bundled
`include/datadog/common.h` no longer contains duplicate typedefs.

## Why

`dedup_headers` only removed definitions from **child** headers that were
byte-identical to ones already present in the **base** header (`common.h`).
It never deduplicated definitions **within** the base header. When cbindgen
emits the same profiling type from two crate boundaries (e.g. via
`after_includes` forward declarations in `libdd-profiling-ffi/cbindgen.toml`
plus the regular body definition), the merged `common.h` ends up with
duplicate typedefs.

These are fatal for consumers compiling with `-Werror -Wtypedef-redefinition`
(C11). Two distinct classes were observed in the v36.0.0 artifacts:

1. **Forward + full-struct collision** — a `typedef struct X X;` forward
   declaration coexisting with the full `typedef struct X { ... } X;`:
   `ddog_prof_EncodedProfile`, `ddog_prof_StringId`, `OpaqueStringId`.
2. **Exact-duplicate pointer typedefs** emitted twice, identical except that
   one carries a doc comment (so the existing exact-string dedup keeps both):
   `ddog_prof_StringId2`, `ddog_prof_MappingId2`, `ddog_prof_FunctionId2`.

## How

Add a final pass (`dedup_base_typedefs`) over the assembled base header that:

- drops a bare forward `typedef struct/union/enum X X;` when a full-body
  definition of the same name `X` exists elsewhere in the file (keeping the
  body, regardless of ordering), and
- drops later duplicates of an identical typedef statement, comparing the
  statement text with any leading doc comment stripped.

Opaque forward declarations (no body) and genuine aliases
(`typedef struct A B;` with `A != B`) are preserved.

This makes `common.h` clean by construction and obsoletes downstream
post-processing workarounds (e.g. the one in `libdatadog-rb`).

## Validation

Headers were generated via the FFI crates' cbindgen build scripts and run
through `dedup_headers` exactly as `builder` invokes it.

Before — each of the six types appears twice; clang fails:

```
$ clang -std=gnu99 -Werror -Wtypedef-redefinition -I<out>/include -fsyntax-only t.c
common.h:541: error: redefinition of typedef 'ddog_prof_EncodedProfile' ...
common.h:909: error: redefinition of typedef 'ddog_prof_StringId2' ...
common.h:936: error: redefinition of typedef 'ddog_prof_MappingId2' ...
common.h:965: error: redefinition of typedef 'ddog_prof_FunctionId2' ...
common.h:1166: error: redefinition of typedef 'OpaqueStringId' ...
common.h:1516: error: redefinition of typedef 'ddog_prof_StringId' ...
6 errors generated.
```

After — each type appears exactly once; clang passes (exit 0):

```
ddog_prof_EncodedProfile : 1
ddog_prof_StringId       : 1
OpaqueStringId           : 1
ddog_prof_StringId2      : 1
ddog_prof_MappingId2     : 1
ddog_prof_FunctionId2    : 1
```

- `cargo test -p tools --lib` — 20 passed (5 new tests for the dedup pass)
- `cargo clippy -p tools --all-targets --all-features -- -D warnings` — clean
- `cargo fmt -p tools -- --check` — clean
> [!NOTE]
> **One of three coordinated changes for the libdatadog v36 duplicate-typedef header issue** (increasing order of permanence):
> - [dd-trace-rb#5928](https://github.com/DataDog/dd-trace-rb/pull/5928) — immediate CI mitigation: bump dd-trace-rb to v36 plus a temporary `-Wno-error=typedef-redefinition` stopgap.
> - [libdatadog-rb#62](https://github.com/DataDog/libdatadog-rb/pull/62) — gem-level fix: strip the duplicate typedefs during vendoring and ship `36.0.0.1.1`, without waiting for a libdatadog release.
> - **This PR** ([libdatadog#2149](https://github.com/DataDog/libdatadog/pull/2149)) — upstream fix in `dedup_headers`: makes `common.h` clean by construction and obsoletes the libdatadog-rb post-processing once released.


tools/src/lib.rs
```diff
@@ -111,6 +111,117 @@ pub mod headers {
...
+    /// with trailing whitespace removed.
+    ///
+    /// This lets two definitions that only differ by their doc comment be
+    /// compared by their statement alone.
+    fn statement_text(def: &str) -> &str {
+        let s = def.trim_start();
+        let s = if s.starts_with("/*") {
+            match s.find("*/") {
+                Some(end) => s[end + 2..].trim_start(),
+                None => s,
```
yannham:
```suggestion
            s.find("*/").map(|end| s[end + 2..].trim_start()).unwrap_or(s);
```

tools/src/lib.rs
```diff
@@ -111,6 +111,117 @@ pub mod headers {
...
+            }
+        } else {
+            s
+        };
+        s.trim_end()
+    }
+
+    fn is_ident(s: &str) -> bool {
+        !s.is_empty() && s.bytes().all(|b| b.is_ascii_alphanumeric() || b == b'_')
```
yannham: For a token to be a valid C identifier, the first character must NOT be a digit, which isn't included in this test. I don't remember in C but in Rust for example this could match `0usize` which isn't an identifier.

tools/src/lib.rs
```diff
@@ -111,6 +111,117 @@ pub mod headers {
...
+    ///    byte-identical). The later occurrence is dropped.
+    fn dedup_base_typedefs(content: &str) -> String {
+        let defs = collect_definitions(content);
+
+        let bodied_names: HashSet<&str> = defs
+            .iter()
+            .filter_map(|d| bodied_typedef_name(statement_text(d.str)))
+            .collect();
+
+        let mut seen: HashSet<&str> = HashSet::new();
```
yannham: It's not entirely clear if what `seen` includes going forward. I would suggest to give it a more explicit name (I suppose it's the `typedef`s that are not bodied?)

tools/src/lib.rs
hoolioh: I think with the current layout this could not happen because the identifiers are generated with cbindgen so it's likely that types such as 0usize will never get into the C headers since cbindgen will translate them. In any case, since it's not a really huge change, it might be worth to add a check for this situation in the case we change how the headers are generated.

hoolioh (approved):
Pending on Yann's suggestions, it looks good to me.

tools/src/lib.rs
yannham: Yes I agree it shouldn't happen in practice, but isn't this just simple and cheap to add the correct condition `s[0] is not between '0' and '9'` here ?

PR #2151, opened 2026-06-22
fix(trace-stats)!: add grpc_method to aggregation key

mabdinur:
# What does this PR do?

Removes \`grpc_method\` / \`rpc.method\` entirely from the OTLP trace metrics pipeline.

Previously \`grpc_method\` was a sidecar on \`GroupedStats\` — present in the OTLP metric payload as an explicit attribute, but not part of the aggregation key. This PR removes it from both:

- **\`FixedAggregationKey\`** — it is no longer an aggregation dimension.
- **\`OtlpExactGroup\`** — it is no longer carried in the per-cell sidecar.
- **\`build_attributes()\`** — \`rpc.method\` is no longer emitted on OTLP histogram data points.

## Rationale

For gRPC spans, \`span.resource\` already carries the full method path (e.g. \`/package.Service/Method\`). Emitting \`rpc.method\` as a separate OTLP attribute is redundant — \`span.resource\` (mapped to \`span.name\` on the data point) is the authoritative carrier of gRPC method identity and is already an aggregation dimension.

Adding \`rpc.method\` as a separate dimension would also increase cardinality for gRPC services without providing any information not already present on the metric.

Note: \`rpc.response.status_code\` (from \`grpc.status.code\`) is **kept** as a first-class aggregation dimension — it is not redundant with any existing field.

# Motivation

Correctness and cleanliness for the OTLP trace metrics feature landed in #2067.

# Risk

**Breaking changes (major semver):**

- \`FixedAggregationKey<T>\` loses its \`grpc_method: T\` field. Callers constructing the struct by name must remove the field.
- \`OtlpExactGroup\` loses its \`grpc_method\` field.
- \`build_attributes()\` loses the \`grpc_method\` parameter.

**Blast radius is limited** because:

1. The OTLP trace metrics feature (#2067) landed on \`main\` after the v36 release (June 19). No released version is affected — this only exists in unreleased code.
2. The OTLP metrics path is opt-in: the sidecar only activates it when \`set_otlp_metrics_endpoint\` is configured. No SDK enables this today.
3. Agent \`/v0.6/stats\` protobuf wire format is unchanged — \`ClientGroupedStats\` has no \`grpc_method\` field.

**Urgency: low.** No live production systems are affected. This must land before the next release that ships the OTLP metrics feature.

# How to test the change?

```
cargo nextest run -p libdd-trace-stats -p datadog-ipc -p libdd-data-pipeline
```


libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -153,7 +156,7 @@ fn get_grpc_status_code<'a>(span: &'a impl StatSpan<'a>) -> Option<u8> {
     None
 }
 
-pub(super) fn get_grpc_method<'a>(span: &'a impl StatSpan<'a>) -> &'a str {
+fn get_grpc_method<'a>(span: &'a impl StatSpan<'a>) -> &'a str {
```
mabdinur: This method is currently unreleased (so this change should be safe)

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -344,9 +350,6 @@ pub(super) struct GroupedStats {
     error_duration: u64,
     error_min: u64,
     error_max: u64,
-    // gRPC method for OTLP export only; not part of the aggregation key so agent stats are
-    // unaffected.
-    pub(super) grpc_method: String,
```
mabdinur, line 349: This change is also unreleased

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -834,13 +831,14 @@ mod tests {
                 }
                 .into_key(),
             ),
-            // grpc.method.name is carried in GroupedStats (for OTLP), not in the aggregation key.
+            // grpc.method.name is part of the aggregation key.
```
VianneyRuhlmann: I don't think this comment is needed anymore

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -275,6 +279,7 @@ impl<'a> BorrowedAggregationKey<'a> {
                 http_method,
                 http_endpoint,
                 service_source,
+                grpc_method,
```
VianneyRuhlmann: This means that we will aggregate based on grpc_method even when not using OTLP (increasing cardianlity). A given concentrator is either always used for OTLP or not at all right ? Then we could add a config option to the concentrator to ignore this field when not in otlp

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -49,6 +49,8 @@ pub struct FixedAggregationKey<T> {
     pub grpc_status_code: Option<u8>,
     pub is_synthetics_request: bool,
     pub is_trace_root: bool,
+    // Appended last to preserve bincode field-order compatibility with existing IPC messages.
```
VianneyRuhlmann: You can remove this comment

VianneyRuhlmann (approved):
Some concerns regarding cardinality for non-otlp exporter otherwise LGTM

libdd-trace-stats/src/span_concentrator/aggregation.rs
mabdinur: Done.

libdd-trace-stats/src/span_concentrator/aggregation.rs
mabdinur: Done.

libdd-trace-stats/src/span_concentrator/aggregation.rs
mabdinur: I updated this PR  to *remove* `rpc.method` from the OTLP metric payload. We no longer need this dimension in the aggregation key. Span.resouce should convey the rpc method for grpc spans. With this change cardinality should remain unchanged. I update the PR description and title

PR #2152, opened 2026-06-23
refactor(libdd-telemetry)!: avoid leaking libdd-common types in the public API

hoolioh:
# What does this PR do?

Break down endpoint settings into smaller functions which accept primitive types in order not to leak libdd-common types through the public API.

# Motivation

Most of downstream projects don't use the Endpoint abstraction because it's thought to handle connections with intake/agent so forcing them to import libdd-common just to configure the connection doesn't seem right.

### DIFF
--- a/libdd-telemetry-ffi/src/builder.rs
+++ b/libdd-telemetry-ffi/src/builder.rs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use ffi::slice::AsBytes;
-use libdd_common::Endpoint;
 use libdd_common_ffi as ffi;
 use libdd_telemetry::{
     data,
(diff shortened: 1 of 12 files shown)


libdd-telemetry-ffi/src/builder.rs
```diff
@@ -155,14 +154,60 @@ pub unsafe extern "C" fn ddog_telemetry_builder_run_metric_logs(
...
+    use_system_resolver: bool,
+) -> ffi::MaybeError {
+    let url = try_c!(url.try_to_utf8());
+    let api_key = api_key.to_utf8_lossy();
+    let test_token = test_token.to_utf8_lossy();
+    let config = &mut telemetry_builder.config;
+    // Set the api key before the url so the telemetry path is resolved correctly.
+    if !api_key.is_empty() {
+        try_c!(config.set_endpoint_api_key(Some(api_key.as_ref())));
+    }
```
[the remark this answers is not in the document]
yannham: I suppose the solution is to set to `None` if the api key is empty, instead of not setting anything?

libdd-telemetry-ffi/src/builder.rs
```diff
@@ -155,14 +154,60 @@ pub unsafe extern "C" fn ddog_telemetry_builder_run_metric_logs(
...
+/// * `api_key` / `test_token`: ignored when empty.
+/// * `timeout_ms`: pass 0 to keep the existing/default timeout.
 pub unsafe extern "C" fn ddog_telemetry_builder_with_endpoint_config_endpoint(
     telemetry_builder: &mut TelemetryWorkerBuilder,
-    endpoint: &Endpoint,
+    url: ffi::CharSlice,
+    api_key: ffi::CharSlice,
+    timeout_ms: u64,
+    test_token: ffi::CharSlice,
+    use_system_resolver: bool,
```
yannham, line 217: The endpoint seems to carry quite a bit of different fields, that now have to be set manually one by one and inflate function arguments (there's a risk to forget about one field when setting the endpoint now instead of `cfg.set_endpoint` previously). I'm just thinking out loud, but would it make sense to re-introduce an `Endpoint` (or a different name) but stand-alone in `libdd-telemetry` that packs `url`, `api_key`, `timeout_ms`, together ? At least if we always want to set them all at once.

libdd-telemetry/examples/tm-send-sketch.rs
```diff
@@ -111,12 +109,10 @@ async fn async_main() {
     let mut config = Config::from_env();
     config.direct_submission_enabled = true;
     config.debug_enabled = true;
+    let api_key = std::env::var("DD_API_KEY").unwrap();
```
yannham, line 112: I know this existed before this PR, but we should avoid reading env vars in libdatadog. Not sure this one can really be avoided in a cross-platform way though...

yannham (commented):
I initially thought you wanted to get rid of the libdd-common dependency for libdd-telemetry. But if I understand correctly, there's still a `Endpoint` lurking there internally, is that correct? Then I'm not sure I understand the motivation. Is it just to avoid having to add explicitly `libdd-common` for end-user crates just to import `Endpoint`? Then why not just re-export it from libdd-telemetry, if it's there anyway?

libdd-telemetry-ffi/src/builder.rs
hoolioh: I thought about that but I ended up adding more parameters here. However you have a point here, maybe taking the same approach as in [TelemetryClientConfig](https://github.com/DataDog/libdatadog/blob/main/libdd-data-pipeline-ffi/src/trace_exporter.rs#L38) would be better? WDYT?

libdd-telemetry/examples/tm-send-sketch.rs
hoolioh: I agree, we shouldn't be changing behaviour in a library based on environment vars, it can cause problems for our consumers. Since the scope for the PR is to solve problems with the releases we can fix the env var issue in a subsequent PR?

hoolioh:
> I initially thought you wanted to get rid of the libdd-common dependency for libdd-telemetry. But if I understand correctly, there's still a `Endpoint` lurking there internally, is that correct? Then I'm not sure I understand the motivation. Is it just to avoid having to add explicitly `libdd-common` for end-user crates just to import `Endpoint`? Then why not just re-export it from libdd-telemetry, if it's there anyway?

The aim is to reduce the ways a crate can consume a type. If the consumer only uses telemetry it would be more than ok re-exporting the symbols through telemetry so there is only one source for them. Hoewever in our current situation we would be coupling telemetry to common and that will cascade into releasing major versions for multiple crates across the workspace.
Anyway, since the topic is dense and the examples are not straightforward, we can chat about this more in depth, having another angle would help us to figure out to minimize the problem we're having with releases.

yannham (approved):

PR #2155, opened 2026-06-23
feat(crashtracking): collect all native stacks for unhandled exception and also crashing thread

gyuheon0h:
# What does this PR do?

Previously, when a runtime reported an unhandled exception, only the runtime-provided stack trace was captured; no native stacks were collected for any thread because `PR_SET_PTRACER` was not set for unhandled exceptions. Additionally, when all-thread collection was enabled for signal-based crashes, the crashing thread was excluded from the threads array.

This change:
1. Enables native stack collection for all threads (including the crashing thread) when reporting unhandled exceptions with `collect_all_threads` enabled. The receiver now has the necessary ptrace permissions to unwind every thread in the process.
2. Includes the crashing thread in the `error.threads` array (marked with `crashed: true`) for both signal-based crashes and unhandled exceptions. Its native stack is collected by the receiver using ptrace, same as every other thread.

The runtime-provided stack remains in `error.stack` as the canonical crash context, while `error.threads` provides a uniform native view of all threads at the time of the crash.

This gives us two immediate benefits:
1. Unhandled exception crash reports now include native stacks for all threads, providing visibility into what the process was doing at the OS level; not just the managed runtime view.
2. The crashing thread's native stack is now collected in the receiver alongside all other threads, establishing a path toward consolidating all stack collection in the receiver process in the future.

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-crashtracker/src/receiver/receive_report.rs
+++ b/libdd-crashtracker/src/receiver/receive_report.rs
@@ -570,6 +570,7 @@ fn collect_and_add_thread_contexts(
 
     let incomplete = stream_thread_contexts(
         parent_pid,
+        crashing_tid,
         config.max_threads(),
         budget,
         config.resolve_frames(),
--- a/libdd-crashtracker/src/receiver/receive_report.rs
+++ b/libdd-crashtracker/src/receiver/receive_report.rs
@@ -566,18 +566,12 @@ fn collect_and_add_thread_contexts(
     let crashing_tid = crashing_tid.unwrap_or(0) as i32;
     let parent_pid = parent_pid as i32;
 
-    // Use the remaining receiver budget
-    // When this expires, stream_thread_contexts stops and returns the threads
-    // collected so far, which are still emitted in the report.
-    let context_timeout = budget;
-
     let mut collected_threads = Vec::new();
 
     let incomplete = stream_thread_contexts(
         parent_pid,
-        crashing_tid,
         config.max_threads(),
-        context_timeout,
+        budget,
         config.resolve_frames(),
         |tid, captured_context| {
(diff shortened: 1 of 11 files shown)


gyuheon0h:
* **#2155** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2155?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2155?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libdd-crashtracker/src/receiver/receive_report.rs
```diff
@@ -566,18 +566,13 @@ fn collect_and_add_thread_contexts(
     let crashing_tid = crashing_tid.unwrap_or(0) as i32;
     let parent_pid = parent_pid as i32;
 
-    // Use the remaining receiver budget
-    // When this expires, stream_thread_contexts stops and returns the threads
-    // collected so far, which are still emitted in the report.
-    let context_timeout = budget;
-
```
gleocadie, line 573: Not in this PR but general question: do we have test(s) to validate that crashtracker adjust the crash report construction based the budget ? like a truncated report is sent with a comment/message "..truncated...incomplete.."?

gleocadie (approved):
one question, other than that LGTM

libdd-crashtracker/src/receiver/receive_report.rs
gyuheon0h: No we do not. We need to have this though. I'll add a task to do this.

PR #2156, opened 2026-06-24
feat: replace v04 usage by v1 in sidecar

anais-raison:
# What does this PR do?

Wires the sidecar so SDKs (PHP first) can send V1 msgpack payloads. Adds the FFI entry points, tarpc methods, blocking/sender wrappers, and server handlers — symmetric to the existing `send_trace_v04_*` chain.

When the sidecar receives V1 bytes, it decodes them through `msgpack_decoder::v1` (added in #2174), can inspect the payload, and re-encodes as V1 on the way to the agent's `/v1.0/traces` endpoint.

# Motivation

[APMSP-2813](https://datadoghq.atlassian.net/browse/APMSP-2813) : enable the sidecar V1-native path. PHP keeps computing its stats SDK-side via the SHM concentrator, so this PR only touches the trace transport pipeline.

[APMSP-2813]: https://datadoghq.atlassian.net/browse/APMSP-2813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -259,6 +249,30 @@ impl SidecarServer {
...
+    ) {
+        self.send_trace(headers, data, target, retry_interval, TraceEncoding::V04)
+    }
+
+    /// Entry point for the V1 trace path. Input bytes are a V1 msgpack `TracerPayload` from the
+    /// SDK; the [`TraceEncoding::V1`] tag drives [`decode_to_trace_chunks`] to the V1 decoder,
+    /// and [`SendData`] then re-encodes the same shape as V1 on the wire to the agent.
+    fn send_trace_v1(
+        &self,
+        headers: &SerializedTracerHeaderTags,
```
bwoebi: Do we still have SerializedTracerHeaderTags with v1?
Well, yes, we do, but only for the booleans, the string values in TracerHeaderTags are part of the Payload now.
As is, I believe this is going to send `datadog-meta-lang: <empty string>` and the other headers too in `<TracerHeaderTags as HeaderMap>::from`?

Maybe we need to split TracerHeaderTags in two parts, one for the generic headers like stats-computation (which are just bools/ints, and can be transferred without dedicated SerializedTracerHeaderTags), and one containing the strings and embedding the new struct.

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -267,15 +267,45 @@ impl SidecarServer {
...
+        generic_tags: TracerGenericTags,
+        data: tinybytes::Bytes,
+        target: &Endpoint,
+        retry_interval: u64,
+    ) {
+        let headers = TracerHeaderTags {
+            generic: generic_tags,
+            ..Default::default()
+        };
+        self.send_trace(headers, data, target, retry_interval, TraceEncoding::V1)
```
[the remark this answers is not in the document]
Leiyks: This need to be fixed indeed

libdd-trace-utils/src/tracer_payload.rs
```diff
@@ -4,6 +4,7 @@
 use crate::span::{v04, v05, v1, BytesData, SharedDictBytes, TraceData};
 use crate::trace_utils::convert_trace_chunks_v04_to_v05;
 use crate::{msgpack_decoder, trace_utils::cmp_send_data_payloads};
+use anyhow::Ok;
```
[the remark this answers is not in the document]
Leiyks: Yeah can be removed

datadog-sidecar/src/service/sidecar_server.rs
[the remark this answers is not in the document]
Leiyks: Probably worth adding a test to ensure the whole send_trace_v1_* chain is working properly

Leiyks (commented):
Except the V1 payloads being posted on the V04 endpoint and few nits, the rest looks good 👌

datadog-sidecar/src/service/sidecar_server.rs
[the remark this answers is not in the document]
anais-raison: Done 👍

datadog-sidecar-ffi/src/lib.rs
```diff
@@ -1158,6 +1158,70 @@ pub unsafe extern "C" fn ddog_sidecar_send_trace_v04_bytes(
...
+#[no_mangle]
+#[allow(clippy::missing_safety_doc)]
+pub unsafe extern "C" fn ddog_sidecar_send_trace_v1_shm(
+    transport: &mut Box<SidecarTransport>,
+    instance_id: &InstanceId,
+    shm_handle: Box<ShmHandle>,
+    len: usize,
+    tracer_header_tags: &TracerHeaderTags,
+) -> MaybeError {
+    std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
yannham: nit: import `std::panic` or `std::panic::{catch_unwind, AssertUnwindSafe)` if it's used several times.

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -270,15 +270,68 @@ impl SidecarServer {
...
+            Ok(headers) => headers,
+            Err(e) => {
+                error!("Failed to convert SerializedTracerHeaderTags into TracerHeaderTags with error {:?}", e);
+                return;
+            }
+        };
+        // Agentful sessions have their endpoint normalized to `/v0.4/traces` by
+        // `tracer::Config::set_endpoint`, since it doesn't know which encoding will be used.
+        // Redirect to the agent's `/v1.0/traces` route for the V1 payload we're about to send.
+        let target = if target.api_key.is_none() {
```
yannham: This is not this PR's fault, but this code does feel a bit out of place. I wonder if in the future we could make `tracer::Config` more encoding-agnostic to avoid this kind of ad-hoc URL rewriting.

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -961,6 +1014,61 @@ impl SidecarInterface for ConnectionSidecarHandler {
         }
     }
 
+    async fn send_trace_v1_shm(
```
yannham, line 1053: I guess this PR is just following the existing style, since it's already the case for `send_trace_v04_xxx`, but I'm not sure those functions should be async. They spawn a task on the side but don't await anything themselves. Reading the async signature I would expect that we actually await the send until completion.

yannham (approved):

datadog-sidecar-ffi/src/lib.rs
bwoebi: Please don't catch_unwind in FFI, we compile the sidecar with panic=abort.

Using catch_unwind messes with pthread_cancel(). (I see some FFE functions also do that, that's a mistake)

datadog-sidecar/src/service/sidecar_server.rs
bwoebi: The async comes from the sidecar_interface contract, which expects futures to be returned.
Allows same handling for all endpoints.

datadog-sidecar-ffi/src/lib.rs
bwoebi: Oh no, I see someone added
```
- Panics across FFI boundaries are undefined behavior. FFI entry points must catch unwinds (e.g. `std::panic::catch_unwind`) and convert them into error returns rather than letting them propagate into the caller's runtime.
```
to AGENTS.md. Why.

datadog-sidecar/src/service/sidecar_server.rs
bwoebi: Can we instead just add tracer::Config::endpoint_v1 or such? I.e. compute both endpoints in tracer::Config::set_endpoint?

datadog-sidecar/src/service/sidecar_server.rs
yannham: Does it make sense for the interface to be async then, if the semantics is to fire and forget for any implementation?

bwoebi: It's not for all, just many. There are a handful which return values.

datadog-sidecar/src/service/sidecar_server.rs
anais-raison: Done 👍

datadog-sidecar-ffi/src/lib.rs
anais-raison: Removed the catch unwind so no need

anais-raison: Remove the `catch_unwind` from the functions I added for v1 but not the FFE functions ones cause it's kinda out of scope of this PR.

bwoebi (approved):
Looks good now, thanks!

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -277,26 +277,28 @@ impl SidecarServer {
...
-            Ok(headers) => headers,
-            Err(e) => {
-                error!("Failed to convert SerializedTracerHeaderTags into TracerHeaderTags with error {:?}", e);
-                return;
-            }
+        let headers = TracerHeaderTags {
+            lang_interpreter,
+            lang_vendor,
+            generic,
+            ..Default::default()
```
bwoebi: You need to fetch the remaining data from the v1 payload actually. The agent is not clever enough to read that from the payload too :-/

PR #2161, opened 2026-06-24
feat(data-pipeline): ensure large span strings get truncated before encoding

brettlangdon:
# What does this PR do?

Add truncation of Span strings that are larger than 25k characters long before encoding.

# Motivation

In dd-trace-py, this truncation currently happens when we encode Span chunks to hand off to the trace exporter, but as we remove the Python based encoders and give Spans directly to the trace exporter we need to maintain this behavior.

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-data-pipeline/src/trace_exporter/mod.rs
+++ b/libdd-data-pipeline/src/trace_exporter/mod.rs
@@ -350,9 +350,7 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static, R: 
             None,
         );
 
-        let res = self
-            .send_trace_chunks_inner(traces, false)
-            .await?;
+        let res = self.send_trace_chunks_inner(traces, false).await?;
         if matches!(&res, AgentResponse::Changed { body } if body.is_empty()) {
             return Err(TraceExporterError::Agent(
                 error::AgentErrorKind::EmptyResponse,
@@ -574,8 +572,7 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static, R: 
         trace_chunks: Vec<Vec<Span<T>>>,
     ) -> Result<AgentResponse, TraceExporterError> {
         self.check_agent_info().await;
-        self.send_trace_chunks_inner(trace_chunks, true)
-            .await
+        self.send_trace_chunks_inner(trace_chunks, true).await
     }
 
     /// Sends trace chunks via OTLP HTTP (JSON or protobuf) when OTLP config is enabled.
--- a/libdd-data-pipeline/src/trace_exporter/mod.rs
+++ b/libdd-data-pipeline/src/trace_exporter/mod.rs
@@ -643,6 +643,10 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Tra
     ) -> Result<AgentResponse, TraceExporterError> {
         let mut header_tags: TracerHeaderTags = self.metadata.borrow().into();
 
+        // Truncate over-long string fields before any downstream processing so that stats,
+        // serialization, and the OTLP path all operate on the same normalized payload.
+        libdd_trace_utils::span::trace_utils::truncate_span_strings(&mut traces);
+
         // Process stats computation and drop non-sampled (p0) chunks.
         // This must run before the OTLP path so that unsampled spans are not exported.
         stats::process_traces_for_stats(
--- a/libdd-data-pipeline/src/trace_exporter/mod.rs
+++ b/libdd-data-pipeline/src/trace_exporter/mod.rs
@@ -329,9 +329,14 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static, R: 
 
         let format: DeserInputFormat = self.input_format.into();
 
+        // Decode into owned SpanBytes (BytesString) rather than borrowed SpanSlice (&str) so
+        // that the truncate_span_strings call inside send_trace_chunks_inner can actually
+        // shorten over-long fields.  copy_from_slice performs one allocation for the whole
+        // payload; all decoded string fields are zero-copy subslices of that buffer.
+        let owned = libdd_tinybytes::Bytes::copy_from_slice(data);
         let (traces, _) = match format {
-            DeserInputFormat::V04 => msgpack_decoder::v04::from_slice(data),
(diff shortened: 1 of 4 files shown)


libdd-trace-utils/src/span/mod.rs
```diff
@@ -24,18 +24,69 @@ use std::{fmt, ptr};
...
+    /// first `result_chars - suffix.chars().count()` code points followed by `suffix`; otherwise
+    /// return `self` unchanged.
+    ///
+    /// Implementations that cannot allocate (e.g. `&str`) return `self` unmodified.
+    fn maybe_truncate(self, max_chars: usize, result_chars: usize, suffix: &str) -> Self {
+        // Default: no allocation possible, so return unchanged.
+        // Implementations that own their storage (e.g. `BytesString`) should override this.
+        let _ = (max_chars, result_chars, suffix);
+        self
+    }
```
paullegranddc, line 38: I think it'd be better to do truncation at serialization, where we are writing to an actual buffer if we cannot guarantee it will happen for every text implementor

paullegranddc: Actually dumb idea, we'd have to add this for every serializer

brettlangdon: adding for every serializer vs needing to ensure we are calling truncation logic for all attributes/key/values/etc ?

it would be just as easy to miss truncation on something like metrics keys for example just as easy as skipping an entire serializer

libdd-trace-utils/src/span/trace_utils.rs
```diff
@@ -5,9 +5,120 @@
...
+use super::{
+    v04::{AttributeAnyValue, AttributeArrayValue, Span},
+    SpanText, TraceData,
+};
 use std::collections::{HashMap, HashSet};
 
+/// Fields whose Unicode code-point count exceeds this threshold are truncated.
+pub const MAX_SPAN_STRING_LEN: usize = 25_000;
+/// Length (in Unicode code points) to which over-long fields are truncated, including the suffix.
+pub const TRUNCATED_SPAN_STRING_LEN: usize = 2_500;
```
paullegranddc, line 17: Why is this way shorter than MAX_SPAN_STRING_LEN, we cut if it's more tham 25000 codepoints but keep only the 1st 2500.
Or is this missing a zero maybe?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -329,9 +329,14 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static, R:
 
         let format: DeserInputFormat = self.input_format.into();
 
+        // Decode into owned SpanBytes (BytesString) rather than borrowed SpanSlice (&str) so
+        // that the truncate_span_strings call inside send_trace_chunks_inner can actually
+        // shorten over-long fields.  copy_from_slice performs one allocation for the whole
+        // payload; all decoded string fields are zero-copy subslices of that buffer.
+        let owned = libdd_tinybytes::Bytes::copy_from_slice(data);
```
bantonsson: I think that this path shouldn't copy, and shouldn't truncate. It's up to the tracer to do that when encoding to msgpack. This change will have performance implications.

bantonsson (changes requested):

libdd-data-pipeline/src/trace_exporter/mod.rs
VianneyRuhlmann: > It's up to the tracer to do that when encoding to msgpack

Do you mean when encoding before sending to the trace exporter ? What about native spans ?
I think this kind of normalization step make sense in the trace exporter i.e. just before sending

> This change will have performance implications.

Agreed maybe we should bring back #2011 (it was reverted since trace filters found an alternative)

brettlangdon: I am looking to move Python to giving native spans directly to the trace exporter, seems silly that all SDKs which want to do that now also need to duplicate truncation logic?

Seems an implementation detail of the encoder not the SDK

e.g. if someone writes a long meta value, should it get truncated on write such that if they read it back they get the truncated version or the original full version?

or is truncation a guard against encoding/transferring/ingesting large strings that would get dropped or truncated on the backend anyway?

bantonsson: > I am looking to move Python to giving native spans directly to the trace exporter

Will the native spans come as msgpack, or where do the native spans come into the whole machinery? I'm just saying that if you get msgpack encoded spans from the tracer, then they should have been truncated.

brettlangdon: not msgpack, going to construct libdd v0.4 spans directly and give to trace exporter (or trace buffer in the future) as a Vec<Vec<Span>>

bantonsson: And those `v0.4` spans can be mutated I guess?

brettlangdon: chatted off github, the feedback is mostly why are we copying the msgpack received, in this case it should be assumed the SDK passing in msgpack payload to the trace exporter already did the truncation

we should/could do the truncation in the case of an SDK directly passing libdatadog Spans, but we should avoid for the msgpack payload approach

bantonsson (approved):

libdd-trace-utils/src/span/trace_utils.rs
brettlangdon: This is what we do in Python, I don't remember the original reason

original PR

https://github.com/DataDog/dd-trace-py/pull/13270

paullegranddc: The PR mentions the agent truncation policy... But it doesn't truncate the same 😅

https://github.com/DataDog/datadog-agent/blob/984dffaa80844e042a5bb24c59b351dd0def00d0/pkg/trace/agent/truncator.go#L89-L92

It truncates at 25K __bytes__ (at the nearest utf8 boundary actually), and not unicode codepoints, which can be significantly smaller for non latin characters and adds "..." after the 25K th character

brettlangdon: yeah, I don't mind diverging a bit from what dd-trace-py does, as long as it aligns with what the trace agent does 👍🏻

PR #2162, opened 2026-06-25
feat(library-config)!: caller-supplied threadlocal schema and extra process-context attributes

szegedi:
## Summary

`TracerMetadata::to_otel_process_ctx` used to hardcode `threadlocal.schema_version = "tlsdesc_v1_dev"` and offered no way to publish additional `threadlocal.*` attributes. A language-specific writer (e.g. a Node.js writer with its own V8 layout constants under a different schema) couldn't fully drive the process context through this path.

This PR introduces a `ThreadLocalMetadata` substruct that gathers all thread-local context configuration in one place, and replaces the existing `threadlocal_attribute_keys` field on `TracerMetadata` with a single `Option<ThreadLocalMetadata> threadlocal_metadata`:

```rust
#[cfg(feature = "otel-thread-ctx")]
pub struct ThreadLocalMetadata {
    pub attribute_keys: Vec<String>,
    pub schema_version: Option<String>,
    pub extra_attributes: Vec<(String, any_value::Value)>,
}
```

- **`attribute_keys`** — same role as the previous `threadlocal_attribute_keys`. The first key is still implicitly `datadog.local_root_span_id`.
- **`schema_version`** — value of `threadlocal.schema_version` published in the process context. Falls back to `"tlsdesc_v1_dev"` when `None`, preserving the behavior current callers rely on.
- **`extra_attributes`** — additional KeyValues to publish in the process context, e.g. the V8 `threadlocal.wrapped_object_offset` / `threadlocal.tagged_size` layout constants the Node.js reader needs. Values are typed directly as `any_value::Value`, so callers can use any of the OTel-supported variants (string, int, bool, double, bytes, array, kvlist).

`Option<ThreadLocalMetadata>` is the single switch for whether the `threadlocal.*` section appears at all. Illegal states like "schema_version set but no key map" are unrepresentable by construction.

## Backward compatibility

Adding fields to `ThreadLocalMetadata` (rather than directly to `TracerMetadata`) means future extensions don't break exhaustive struct literals at call sites. Existing `threadlocal_attribute_keys: Some(vec![...])` callers migrate to `threadlocal_metadata: Some(ThreadLocalMetadata { attribute_keys: vec![...], ..Default::default() })` — that's a breaking change, hence the `!` in the PR title, but in practice the only in-tree consumer is `libdatadog-nodejs`.

Also dropped `Eq, Hash` from `TracerMetadata`'s derive: `any_value::Value` contains `f64` (no `Eq`/`Hash`) and arrays (no `Hash`). Nothing in-tree puts `TracerMetadata` in a hash-keyed collection; `PartialEq` is retained.

## Test plan

- [x] `cargo test -p libdd-library-config --features otel-thread-ctx`
- [ ] Wire up `ThreadLocalMetadata` in `libdatadog-nodejs`'s `process_discovery` crate
- [ ] Have `dd-trace-js` pass schema `"nodejs_v1_dev"` + V8 layout constants through, dropping its own ad-hoc handling

Jira: [PROF-15221]

[PROF-15221]: https://datadoghq.atlassian.net/browse/PROF-15221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libdd-library-config/src/tracer_metadata.rs
+++ b/libdd-library-config/src/tracer_metadata.rs
@@ -178,15 +178,13 @@ impl TracerMetadata {
                 key_ref: 0,
             });
 
-            attributes.extend(
-                extra_attributes.iter().map(|(k,v)| {
-                    KeyValue {
-                        key: k.clone(),
-                        value: Some(AnyValue {
-                            value: Some(v.clone()),
-                        }),
-                        key_ref: 0,
-                    }));
+            attributes.extend(extra_attributes.iter().map(|(k, v)| KeyValue {
+                key: k.clone(),
+                value: Some(AnyValue {
+                    value: Some(v.clone()),
+                }),
+                key_ref: 0,
+            }));
         }
 
         ProcessContext {
--- a/libdd-library-config/src/tracer_metadata.rs
+++ b/libdd-library-config/src/tracer_metadata.rs
@@ -71,8 +71,7 @@ pub struct TracerMetadata {
     pub threadlocal_schema_version: Option<String>,
 
     /// Extra OTel process-context attributes the threadlocal writer wants to publish alongside the
-    /// key map (e.g. language-runtime layout constants that the reader needs to walk from a
-    /// discovery TLS symbol into the record). Each entry is emitted verbatim as a KeyValue.
+    /// key map (e.g. language-runtime layout constants that the reader needs to know about). Each entry is emitted verbatim as a KeyValue.
     ///
     /// Only emitted when `threadlocal_attribute_keys` is `Some`. Ignored for (de)serialization.
     #[cfg(feature = "otel-thread-ctx")]
--- a/libdd-library-config/src/tracer_metadata.rs
+++ b/libdd-library-config/src/tracer_metadata.rs
@@ -61,9 +61,8 @@ pub struct TracerMetadata {
     #[serde(skip)]
     pub threadlocal_attribute_keys: Option<Vec<String>>,
 
-    /// Value of the `threadlocal.schema_version` attribute emitted in the OTel process context.
-    /// Identifies the on-the-wire record schema the writer publishes (e.g. `"tlsdesc_v1_dev"` for
-    /// libdatadog's own TLSDESC-based writer, `"nodejs_v1_dev"` for a Node.js writer). Defaults to
+    /// Identifies the record schema the writer publishes (e.g. `"tlsdesc_v1_dev"` for
+    /// libdatadog's own writer, `"nodejs_v1_dev"` for a Node.js writer). Defaults to
     /// `"tlsdesc_v1_dev"` when `None`.
     ///
     /// Only emitted when `threadlocal_attribute_keys` is `Some`. Ignored for (de)serialization.
--- a/libdd-library-config/src/tracer_metadata.rs
+++ b/libdd-library-config/src/tracer_metadata.rs
@@ -370,7 +370,7 @@ mod tests {
         }
         .to_otel_process_ctx();
 
-        let schema_version = find_attr(&ctx, "threadlocal.schema_version")
+        let schema_version = find_extra_attr(&ctx, "threadlocal.schema_version")
             .expect("threadlocal.schema_version should be present");
         assert_eq!(
             schema_version.value,
@@ -405,15 +405,15 @@ mod tests {
         .to_otel_process_ctx();
 
         assert_eq!(
-            find_attr(&ctx, "threadlocal.wrapped_object_offset").and_then(|v| v.value.clone()),
+            find_extra_attr(&ctx, "threadlocal.wrapped_object_offset").and_then(|v| v.value.clone()),
             Some(any_value::Value::IntValue(24))
         );
         assert_eq!(
-            find_attr(&ctx, "threadlocal.tagged_size").and_then(|v| v.value.clone()),
+            find_extra_attr(&ctx, "threadlocal.tagged_size").and_then(|v| v.value.clone()),
             Some(any_value::Value::IntValue(8))
         );
         assert_eq!(
-            find_attr(&ctx, "threadlocal.runtime.name").and_then(|v| v.value.clone()),
+            find_extra_attr(&ctx, "threadlocal.runtime.name").and_then(|v| v.value.clone()),
             Some(any_value::Value::StringValue("nodejs".to_owned()))
         );
     }
--- a/libdd-library-config/src/tracer_metadata.rs
+++ b/libdd-library-config/src/tracer_metadata.rs
@@ -3,20 +3,8 @@
 use libdd_trace_protobuf::opentelemetry::proto as otel_proto;
 use std::default::Default;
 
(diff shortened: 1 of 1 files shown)


libdd-library-config/src/tracer_metadata.rs
```diff
@@ -3,6 +3,18 @@
 use libdd_trace_protobuf::opentelemetry::proto as otel_proto;
 use std::default::Default;
 
+/// Value of an additional OTel process-context attribute. Mirrors the small subset of
+/// `opentelemetry::proto::common::v1::AnyValue` variants we support for caller-supplied threadlocal
+/// extras — string and 64-bit integer, since the only consumers so far are textual schema
+/// identifiers and small numeric layout constants (e.g. struct offsets, pointer widths).
```
yannham: Are we confident this will always be the case? I feel a simple approach would just to let caller set any value supported by the process context spec (and that would get rid of this additional type). Put differently, would that be a problem / redflag if libdatadog consumers passed something else than a string or an int?

It's ok to start like this and evolve in the future (although it's a breaking change), we don't have to envision all possible use cases right now. I'm asking mostly because it adds boilerplate, so wondering if it's really necessary.

libdd-library-config/src/tracer_metadata.rs
```diff
@@ -48,6 +60,25 @@ pub struct TracerMetadata {
     #[cfg(feature = "otel-thread-ctx")]
     #[serde(skip)]
     pub threadlocal_attribute_keys: Option<Vec<String>>,
+
+    /// Value of the `threadlocal.schema_version` attribute emitted in the OTel process context.
+    /// Identifies the on-the-wire record schema the writer publishes (e.g. `"tlsdesc_v1_dev"` for
+    /// libdatadog's own TLSDESC-based writer, `"nodejs_v1_dev"` for a Node.js writer). Defaults to
+    /// `"tlsdesc_v1_dev"` when `None`.
```
yannham: Nit: fluff

```suggestion
    /// Identifies the record schema the writer publishes (e.g. `"tlsdesc_v1_dev"` for
    /// libdatadog's own writer, `"nodejs_v1_dev"` for a Node.js writer). Defaults to
    /// `"tlsdesc_v1_dev"` when `None`.
```

libdd-library-config/src/tracer_metadata.rs
```diff
@@ -48,6 +60,25 @@ pub struct TracerMetadata {
...
+    ///
+    /// Only emitted when `threadlocal_attribute_keys` is `Some`. Ignored for (de)serialization.
+    #[cfg(feature = "otel-thread-ctx")]
+    #[serde(skip)]
+    pub threadlocal_schema_version: Option<String>,
+
+    /// Extra OTel process-context attributes the threadlocal writer wants to publish alongside the
+    /// key map (e.g. language-runtime layout constants that the reader needs to walk from a
+    /// discovery TLS symbol into the record). Each entry is emitted verbatim as a KeyValue.
```
yannham:
```suggestion
    /// key map (e.g. language-runtime layout constants that the reader needs to know about). Each entry is emitted verbatim as a KeyValue.
```

yannham (commented):
The motivation LGTM. I wonder though if we shouldn't gather all thread-locals related stuff in a substruct, something like:

```
#[cfg(feature = "otel-thread-ctx")]
ThreadLocalMetadata {
        threadlocal_attribute_keys: Vec<String>,
        threadlocal_schema_version: Option<String>,
        threadlocal_extra_attributes: Vec<(String, ProcessContextAttrValue)>,
}
```

And then have in `TracerMetadata` a single field

```
    #[cfg(feature = "otel-thread-ctx")]
    #[serde(skip)]
    pub threadlocal_metadata: Option<ThreadLocalMetadata>,
```

The idea being that this avoid illegal states that are currently possible, like `threadlocal_extra_attributes` or `threadlocal_schema_version` being defined/non-empty but with `attribute_keys` being empty, which results in silently not publishing a `ProcessContext`.

With the `ThreadLocalMetadata`, the single point of activation is whether `threadlocal_metadata` is `Some` or not. So it's not possible to have it halfway defined.

This would be breaking but is ok - in practice the FFI is the only one currently using this code, so it's all within libdatadog, to the best of my knowledge.

What do you think?

yannham:
(Another solution would be to remove the `Option` around `attribute_keys` and just publish a context unconditionally)

libdd-library-config/src/tracer_metadata.rs
szegedi: Fair enough, we can either go with `Vec<(String, any_value::Value)>` or even `Vec<KeyValue>` in which case the caller needs to construct `KeyValue`. For that reason, I'd lean towards `Vec(String, any_value::Value)` although admittedly I don't have enough idiomatic Rust experience to decide. Which one would you suggest?

szegedi:
@yannham I like the suggestion of adding a struct a lot, I'm all for eliminating illegal states. I might remove the `threadlocal_` prefix from its fields though, although that's admittedly a bikeshedding:
```
#[cfg(feature = "otel-thread-ctx")]
ThreadLocalMetadata {
        attribute_keys: Vec<String>,
        schema_version: Option<String>,
        extra_attributes: Vec<(String, ProcessContextAttrValue)>,
}

**Comment by szegedi** [comment]
I updated the PR with the suggestions. 

For one, switched to `any_value::Value` for extra attributes. It does have one weird side effect, though: I had to drop `Eq` and `Hash` from `TracerMetadata`'s derive (required since `any_value::Value` can also be f64 and arrays, and those don't have `Eq` and `Hash` either). No internal consumer puts TracerMetadata in a hash-keyed collection so I think this is fine? It retains `PartialEq` so existing equality test still works.

For another, I introduced the `ThreadlocalMetadata` struct; it's indeed much cleaner like this. 

Apologies for force pushes after review; my clanker did that by amending into the original commit and force-pushing. I restored the original commit from reflog and then applied the changes on top again as separate commits, so even though it says a bunch of "force pushed" in the PR history, the first commit is again the exact same SHA that you originally reviewed.

**Review by ivoanjo — APPROVED** [review]
👍 LGTM. I'll delegate to @yannham the "make it a really nice rust API" details, yet I think overall this looks good from my side.

File: libdd-library-config/src/tracer_metadata.rs [hunk 4]
```diff
@@ -144,14 +169,24 @@ impl TracerMetadata {
...
+
+            for (k, v) in extra_attributes {
+                attributes.push(KeyValue {
+                    key: k.clone(),
+                    value: Some(AnyValue {
+                        value: Some(v.clone()),
+                    }),
+                    key_ref: 0,
+                });
+            }
```
yannham: Very nit: more idiomatic

```suggestion
            attributes.extend(
                extra_attributes.iter().map(|(k,v)| {
                    KeyValue {
                        key: k.clone(),
                        value: Some(AnyValue {
                            value: Some(v.clone()),
                        }),
                        key_ref: 0,
                    }));
```

yannham (approved):

PR #2163, opened 2026-06-25
feat(machine id): Add helpers in ddcommon to fetch the machine UUID l…

paullegranddc:
…ike the agent

# Motivation

We are replicating the agent RC client. One of the thing the agent uses is a stable identity for the agent host, decorrelated from the hostname (as it is not available in som platforms).

It fetches the UUID using a library (psutils), for which there exists no strict equivalent in rust.

This PR adds code that should be feature equal to the agent

# Changes

* Add new machine_id module in libdd-common
* Fetch machine id on different platform

# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-common/src/machine_id/linux.rs
+++ b/libdd-common/src/machine_id/linux.rs
@@ -1,0 +1,121 @@
+// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+use std::path::Path;
+
+fn read_trimmed(path: &Path) -> Option<String> {
+    let s = std::fs::read_to_string(path).ok()?;
+    let s = s.trim().to_owned();
+    if s.is_empty() {
+        None
+    } else {
+        Some(s)
+    }
+}
+
+pub fn get_machine_id_impl_paths(dmi_path: &Path, etc_path: &Path, boot_path: &Path) -> String {
+    if let Some(id) = read_trimmed(dmi_path) {
+        return id;
+    }
+    // agent compatibility:
+    // gopsutil only accepts /etc/machine-id when it's exactly 32 chars (bare hex)
+    if let Some(id) = read_trimmed(etc_path) {
+        if id.len() == 32 {
+            return id;
+        }
+    }
+    read_trimmed(boot_path).unwrap_or_default()
+}
+
+pub fn get_machine_id_impl() -> String {
+    get_machine_id_impl_paths(
+        Path::new("/sys/class/dmi/id/product_uuid"),
+        Path::new("/etc/machine-id"),
+        Path::new("/proc/sys/kernel/random/boot_id"),
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn write(path: &Path, content: &[u8]) {
+        std::fs::write(path, content).unwrap();
+    }
+
+    fn tmp_paths(
+        dir: &tempfile::TempDir,
+    ) -> (std::path::PathBuf, std::path::PathBuf, std::path::PathBuf) {
+        (
+            dir.path().join("product_uuid"),
+            dir.path().join("machine_id"),
+            dir.path().join("boot_id"),
+        )
+    }
+
+    #[test]
+    fn level1_dmi_wins_when_present() {
+        let dir = tempfile::tempdir().unwrap();
+        let (dmi, etc, boot) = tmp_paths(&dir);
+        write(&dmi, b"B08FA8A2-B01A-4D2B-BD95-FEC7E30C5AEC\n");
+        write(&etc, b"aabbccddaabbccddaabbccddaabbccdd\n");
+        write(&boot, b"cccccccccccccccccccccccccccccccc\n");
+        assert_eq!(
+            get_machine_id_impl_paths(&dmi, &etc, &boot),
+            "B08FA8A2-B01A-4D2B-BD95-FEC7E30C5AEC"
+        );
+    }
+
+    #[test]
+    fn level2_etc_used_when_dmi_absent() {
+        let dir = tempfile::tempdir().unwrap();
+        let (dmi, etc, boot) = tmp_paths(&dir);
+        write(&etc, b"aabbccddaabbccddaabbccddaabbccdd\n");
+        write(&boot, b"cccccccccccccccccccccccccccccccc\n");
+        assert_eq!(
+            get_machine_id_impl_paths(&dmi, &etc, &boot),
+            "aabbccddaabbccddaabbccddaabbccdd"
+        );
+    }
+
+    #[test]
+    fn level2_skipped_when_etc_not_32_chars() {
+        let dir = tempfile::tempdir().unwrap();
+        let (dmi, etc, boot) = tmp_paths(&dir);
+        write(&etc, b"aabbccdd-aabb-ccdd-aabb-ccddaabbccdd\n");
+        write(&boot, b"dddddddddddddddddddddddddddddddd\n");
+        assert_eq!(
+            get_machine_id_impl_paths(&dmi, &etc, &boot),
+            "dddddddddddddddddddddddddddddddd"
+        );
+    }
+
+    #[test]
+    fn level3_boot_id_as_last_resort() {
+        let dir = tempfile::tempdir().unwrap();
+        let (dmi, etc, boot) = tmp_paths(&dir);
+        write(&boot, b"cccccccccccccccccccccccccccccccc\n");
+        assert_eq!(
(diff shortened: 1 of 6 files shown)


libdd-common/src/machine_id/windows.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+fn is_wow64() -> bool {
+    let mut result: i32 = 0;
+    let ok = unsafe { IsWow64Process(GetCurrentProcess(), &mut result) };
+    ok != 0 && result != 0
+}
+
+pub fn get_machine_id_impl() -> String {
+    let access = if cfg!(target_pointer_width = "32") && is_wow64() {
```
ekump: nit: Do you need `is_wow64()`? The agent always does `KEY_READ_ | KEY_WOW64_64KEY` Won't `KEY_WOW64_64KEY` be a no-op on 64-bit?

libdd-common/src/machine_id/windows.rs
```diff
@@ -0,0 +1,85 @@
+// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
```
ekump, line 3:
```suggestion
//! Windows host machine id: reads `MachineGuid` from                                                                                                                                                                                                                              //! `HKLM\SOFTWARE\Microsoft\Cryptography` via the raw Win32 registry API.                                                                                                                                                                                                         
//!                                                                                                                                                                                                                                                                                
//! `MachineGuid` lives in the 64-bit registry view, so a 32-bit process under                                                                                                                                                                                                     
//! WOW64 must pass `KEY_WOW64_64KEY` or it gets redirected to the (empty)                                                                                                                                                                                                         
//! `WOW6432Node` copy. The value is read with the standard two-call Win32                                                                                                                                                                                                         
//! pattern: query the byte size, then read into a right-sized buffer.
```

libdd-common/src/machine_id/windows.rs
ekump:
```suggestion
    // MachineGuid is in the 64-bit view; a 32-bit process under WOW64 is                                                                                                                                                                                                          
    // redirected to WOW6432Node by default, so force the 64-bit view there.
    let access = if cfg!(target_pointer_width = "32") && is_wow64() {
```

libdd-common/src/machine_id/windows.rs
```diff
@@ -0,0 +1,85 @@
...
+            &mut data_len,
+        )
+    };
+    if status != ERROR_SUCCESS || data_type != REG_SZ {
+        // SAFETY: hkey is a valid open handle.
+        unsafe { RegCloseKey(hkey) };
+        return String::new();
+    }
+
+    let mut buf: Vec<u16> = vec![0u16; (data_len as usize).div_ceil(2)];
```
ekump, line 60:
```suggestion
    // data_len is bytes; REG_SZ holds UTF-16, so the u16 count is bytes/2 (round up).
    let mut buf: Vec<u16> = vec![0u16; (data_len as usize).div_ceil(2)];
```

libdd-common/src/machine_id/windows.rs
```diff
@@ -0,0 +1,85 @@
...
+        unsafe { RegOpenKeyExW(HKEY_LOCAL_MACHINE, subkey.as_ptr(), 0, access, &mut hkey) };
+    if status != ERROR_SUCCESS {
+        return String::new();
+    }
+
+    let value_wide = to_wide_null("MachineGuid");
+
+    let mut data_type: u32 = 0;
+    let mut data_len: u32 = 0;
+    // SAFETY: null data pointer is valid for a size-query call.
```
ekump, line 41:
```suggestion
    // SAFETY: null data pointer is valid for a size-query call.
    // First call (null data pointer) returns the value's size in bytes.
```

libdd-common/src/machine_id/windows.rs
```diff
@@ -0,0 +1,85 @@
...
+    };
+    if status != ERROR_SUCCESS || data_type != REG_SZ {
+        // SAFETY: hkey is a valid open handle.
+        unsafe { RegCloseKey(hkey) };
+        return String::new();
+    }
+
+    let mut buf: Vec<u16> = vec![0u16; (data_len as usize).div_ceil(2)];
+    let mut actual_len = data_len;
+    // SAFETY: buf has the capacity returned by the size-query call above.
```
ekump, line 62:
```suggestion
    // SAFETY: buf has the capacity returned by the size-query call above.
    // Second call reads the value into the size-query'd buffer.
```

libdd-common/src/machine_id/windows.rs
```diff
@@ -0,0 +1,85 @@
...
+        )
+    };
+    // SAFETY: hkey is a valid open handle.
+    unsafe { RegCloseKey(hkey) };
+
+    if status != ERROR_SUCCESS {
+        return String::new();
+    }
+
+    while buf.last() == Some(&0u16) {
```
ekump, line 82:
```suggestion
     // REG_SZ data includes a trailing NUL (possibly padded with more); drop them.
    while buf.last() == Some(&0u16) {
```

libdd-common/src/machine_id/macos.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
```
ekump, line 3:
```suggestion
//! macOS host machine id via `gethostuuid(3)`, returning `IOPlatformUUID`.                                                                                                                                                                                                        
```

libdd-common/src/machine_id/macos.rs
```diff
@@ -0,0 +1,39 @@
...
+/// Returns `IOPlatformUUID` via `gethostuuid(3)`, which avoids a fork+exec of `ioreg`.
+pub fn get_machine_id_impl() -> String {
+    let mut uuid = [0u8; 16];
+    let wait = libc::timespec {
```
ekump, line 10:
```suggestion
    // Zero timeout: the host UUID is static, so there's nothing to wait for.
    let wait = libc::timespec {
```

libdd-common/src/machine_id/macos.rs
```diff
@@ -0,0 +1,39 @@
...
+        tv_sec: 0,
+        tv_nsec: 0,
+    };
+    let rc = unsafe { libc::gethostuuid(uuid.as_mut_ptr(), &wait) };
+    if rc != 0 {
+        return String::new();
+    }
+    format!(
```
ekump, line 19:
```suggestion
    // Assemble the 16 raw bytes into the canonical 8-4-4-4-12 hyphenated UUID.
    format!(
```

libdd-common/src/machine_id/linux.rs
```diff
@@ -0,0 +1,121 @@
+// Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
```
ekump, line 3:
```suggestion
//! Linux host machine id, mirroring gopsutil's fallback order:                                                                                                                                                                                                                    
//! `/sys/class/dmi/id/product_uuid` (root-only, usually empty otherwise) →                                                                                                                                                                                                        
//! `/etc/machine-id` → `/proc/sys/kernel/random/boot_id`.
```

libdd-common/src/machine_id/linux.rs
```diff
@@ -0,0 +1,121 @@
...
+    let s = std::fs::read_to_string(path).ok()?;
+    let s = s.trim().to_owned();
+    if s.is_empty() {
+        None
+    } else {
+        Some(s)
+    }
+}
+
+pub fn get_machine_id_impl_paths(dmi_path: &Path, etc_path: &Path, boot_path: &Path) -> String {
```
ekump, line 24:
```suggestion
/// Resolves the id from the three candidate paths in priority order. Paths are                                                                                                                                                                                                    
/// parameters so the fallback logic is unit-testable without real `/sys`/`/etc`.
pub fn get_machine_id_impl_paths(dmi_path: &Path, etc_path: &Path, boot_path: &Path) -> String {
```

ekump (approved):
LGTM FWIW. I found the code confusing, so suggested some more comments.

ekump (approved):
LGTM FWIW. I found the code confusing, so suggested some more comments.

libdd-common/src/machine_id/windows.rs
paullegranddc: Yeah, probably not necessary. I got confused by microsoft doc and claude got defensive

PR #2167, opened 2026-06-26
fix(otel-thread-ctx): put the threadlocal attributes at the right place in the context

yannham:
# What does this PR do?

This PR stores the `threadlocal*` attributes in the `attributes` field of the process context, instead of the `resource` field.
Additionally, fix an unrelated issue (remove undue `#[cfg(..)]` gate) that was introduced in the [otel thread feature-gating](https://github.com/DataDog/libdatadog/pull/1843), preventing normal tracer metadata from being published in the context when the `otel-thread-ctx` feature is active.

# Motivation

After experimenting more with OTel thread context sharing, I found out the way we set up the process context is incorrect. The `threadlocal*` attributes should go in the `attributes`/`extra_attributes` part of the process context, not the `resource` part. Tested that it now works with the context reader locally.

# Additional Notes

N/A

# How to test the change?

I'm using this version in a prototype integration in dd-trace-rs, using the [context reader](https://github.com/scottgerring/ctx-sharing-demo/tree/main/context-reader). This is how I uncovered the issue initially. Now I'm seeing context properly read with this fix.

### DIFF
--- a/libdd-library-config/src/tracer_metadata.rs
+++ b/libdd-library-config/src/tracer_metadata.rs
@@ -115,8 +115,7 @@ impl TracerMetadata {
             threadlocal_attribute_keys,
         } = self;
 
-        #[cfg_attr(not(feature = "otel-thread-ctx"), allow(unused_mut))]
-        let mut attributes = vec![
+        let attributes = vec![
             key_value_opt("service.name", service_name),
             key_value_opt("service.instance.id", runtime_id),
             key_value_opt("service.version", service_version),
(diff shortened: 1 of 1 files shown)


libdd-library-config/src/tracer_metadata.rs
```diff
@@ -115,8 +115,7 @@ impl TracerMetadata {
             threadlocal_attribute_keys,
         } = self;
 
-        #[cfg_attr(not(feature = "otel-thread-ctx"), allow(unused_mut))]
-        let mut attributes = vec![
+        let attributes = vec![
```
ivoanjo: Minor: Maybe call these `resources` to make it even more clear that this is different form the other attributes?

libdd-library-config/src/tracer_metadata.rs
```diff
@@ -160,7 +161,7 @@ impl TracerMetadata {
                 dropped_attributes_count: 0,
                 entity_refs: vec![],
             }),
-            extra_attributes: vec![key_value_opt("datadog.process_tags", process_tags)],
+            extra_attributes,
         }
```
ivoanjo, line 167: Minor: Maybe add a test for this? (The resources/attributes are on the right place?)

ivoanjo (approved):
👍 LGTM. CI is unhappy, but other than that + some very small suggestions, this looks fine! I completely missed that we were putting these attributes in the wrong place too 😅

libdd-library-config/src/tracer_metadata.rs
yannham: Yup, I bet `attributes` vs `extra_attributes` is probably what caused the confusion to begin with...

libdd-library-config/src/tracer_metadata.rs
yannham: We actually have a test for it, but of course it's checking the previous, wrong place.. Updating it

morrisonlevi (approved):

morrisonlevi:
The merge queue timed out. I am not skilled in investigating MQ failures, and I was not able to figure out why. I merged in main in hopes that it will work this time.

PR #2168, opened 2026-06-26
feat(otel-thread-ctx)!: add shared thread context

yannham:
# What does this PR do?

This PR adds a different shared ownership model for the OTel thread context record, in addition the the current API that only supports owned context records. Doing so, we split the interface into corresponding `owned` and `shared` submodules.

Since the two modes are incompatible (trying to attach/detach in one mode and doing the other operation in the other mode is basically UB), the two modes are feature-gated in a way such that they can't co-exist (if both features are set, `owned` wins).

The FFI is left unchanged (from the point of view of an external consumer at least).

# Motivation

We want to integrate the work in dd-trace-rs. The current API assumes that the context is a heap-allocated, uniquely owned piece (which is indeed required for the update API). However, given https://github.com/open-telemetry/opentelemetry-rust/pull/3585, this won't work for dd-trace-rs, where the context will be an immutable  shared `Arc`.

There are two possible solutions:

1. expose lower-level, unsafe internals and let dd-trace-rs play with pointers conversions and handle the refcounting part. More flexible in some sense (not `Arc`-specific), but more dangerous.
2. since we'll likely never need anything else beyond the current box-owned and the new arc-shared in practice, we can make a second safe Arc-based API. This is the option I've picked.

# Additional Notes

For reviewers: I've split the existing codebase, but almost all of the new code should be located in `shared.rs`. `owned.rs` is just part of the previous API but renamed, with just a few functions factorized in `linux.rs` as they are now used from both modules. Same for the FFI: beside moving `ThreadContextHandle` directly there, no major changes here.

# How to test the change?

tests included.


libdd-otel-thread-ctx/src/linux.rs
```diff
@@ -0,0 +1,378 @@
...
+pub use owned::*;
+
+// `owned-context` deliberately takes precedence when both modes are requested, so that enabling
+// all features of the workspace (which the FFI crate does, and which requires the owned mode)
+// still builds. `build.rs` warns about this. The two modes can't coexist: see the documentation of
+// the context types about mixing them.
+#[cfg(all(feature = "shared-context", not(feature = "owned-context")))]
+mod shared;
+#[cfg(all(feature = "shared-context", not(feature = "owned-context")))]
+pub use shared::*;
```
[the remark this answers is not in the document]
yannham: This is on purpose. I can't prevent separate packages from being used together. It's not ideal to have "falsely additive" features like that, but is the only way I know in Rust unfortunately have this kind of static exclusion policy guaranteeing that one can never use both APIs at the same time.

libdd-otel-thread-ctx/src/linux/shared.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Returns the previously attached shared context, if any.
+    pub fn attach(self) -> Option<SharedThreadContext> {
+        // Move the strong reference into the TLS slot; it stays alive until detached.
+        // `ThreadContext` is `repr(transparent)` over `ThreadContextRecord`, so the data pointer is
+        // also the record pointer readers expect.
+        //
+        // Though our `SharedThreadContext` might actually come from a different thread now, it's
+        // wrapped in an `Arc` that handles drop safety by synchronizing on the reference count. The
+        // record is already `valid = 1` and, being shared, immutable.
+        let record_ptr = Arc::into_raw(self.0) as *mut ThreadContextRecord;
```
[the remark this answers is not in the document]
yannham: It's a legitimate concern, but unrelated to this PR IMHO. Was already the case for the previous owned API. I guess we can use a Rust thread-local dummy variable just for cleanup. Will consider this in a follow-up, if that sounds reasonable to reviewers

ivoanjo (commented):
So I'm ultra-busy this week so I'm not sure I'll be able to review this, but overall the problem/solution seem reasonable so no concerns.

libdd-otel-thread-ctx/src/linux/owned.rs
```diff
@@ -0,0 +1,688 @@
...
+                span_id,
+                trace_flags,
+                local_root_span_id,
+                attrs,
+            );
+
+            if is_current {
+                None
+            } else {
+                let prev = ThreadContextRecord::attach_raw(target.as_ptr());
```
ekump: Do you need to call `attach_raw` here? It's going to resolve the TLS address again. Could you just do `let prev = slot.swap(target.as_ptr(), Ordering::Relaxed);`?

libdd-otel-thread-ctx/src/linux/shared.rs
```diff
@@ -0,0 +1,219 @@
...
+        // Safety: a non-null value in the slot came from a prior `attach` of a shared context (see
+        // the type-level note on not mixing owned and shared attaches).
+        NonNull::new(ThreadContextRecord::detach_raw())
+            .map(|ptr| unsafe { Self::from_record_ptr(ptr) })
+    }
+}
+
+#[cfg(test)]
+// The tests are set to be ignored by Miri, since the inline-asm TLSDESC access isn't supported.
+mod tests {
```
ekump, line 107: Should you have a multi-threaded test to verify the per-thread-slot isolation? From what I can tell, that's not being verified at the moment.

libdd-otel-thread-ctx/src/linux/owned.rs
```diff
@@ -0,0 +1,688 @@
...
+}
+
+/// An owned (and non-moving) thread context record allocation.
+///
+/// We don't use `Box` under the hood because it precludes aliasing, while we share the context to
+/// readers through thread-level context and through the FFI. But it is a boxed
+/// `ThreadContextRecord` for all intent and purpose.
+///
+/// Since an owned context can be modified in place, it is not thread safe: it is neither `Send` nor
+/// `!Sync`. Readers rely on the fact that there can't be any concurrent writer while they interrupt
```
ekump:
```suggestion
/// Since an owned context can be modified in place, it is not thread safe: it is `!Send` and
/// `!Sync`. Readers rely on the fact that there can't be any concurrent writer while they interrupt
```

libdd-otel-thread-ctx/README.md
```diff
@@ -13,6 +13,22 @@ data.
...
+
+The crate provides exactly one of two context ownership flavours, selected by
+two mutually exclusive features:
+
+- `owned-context` (default): `OwnedThreadContext`, a record owned by a single
+  thread, which can be updated in place without reallocating. This is the one
+  used by the FFI.
+- `shared-context`: `SharedThreadContext`, an `Arc`-backed immutable record
+  that can be cloned and attached on several threads. To be consumed by another
+  Rust crate, Currently dd-trace-rs.
```
ekump:
```suggestion
  Rust crate, currently dd-trace-rs.
```

libdd-otel-thread-ctx/README.md
```diff
@@ -13,6 +13,22 @@ data.
...
+
+They are exclusive because the thread-local slot is untyped: interleaving owned
+and shared contexts would misinterpret the pointer cause UB. If both features
```
ekump:
```suggestion
and shared contexts would misinterpret the pointer and cause UB. If both features
```

ekump (approved):
I'd recommend getting another review from someone who knows this crate better, but looks reasonable to me.

libdd-otel-thread-ctx/src/linux/owned.rs
yannham: Good catch, thanks! It has slipped through the refactoring. Should definitely re-use `slot` to avoid resolve the TLS address a second time

libdd-otel-thread-ctx/src/lib.rs
```diff
@@ -8,6 +8,24 @@
...
 //! storage symbol and its accessor using inline assembly (`global_asm!` / `asm!`).
 //!
+//! ## Ownership modes
+//!
+//! The crate provides exactly one of two context ownership flavours, selected by two mutually
+//! exclusive features: `owned-context` (the default) gives `OwnedThreadContext`, a record owned by
+//! a single thread that can be updated in place, while `shared-context` gives
+//! `SharedThreadContext`, an `Arc`-backed immutable record that can be cloned and attached on
+//! several threads. They can't coexist because the TLS slot is untyped. If both features are
+//! enabled, `owned-context` wins and the build script emits a warning.
```
scottgerring: why not an error?

libdd-otel-thread-ctx/src/lib.rs
```diff
@@ -18,20 +36,24 @@
...
 //!
 //! let trace_id = [0u8; 16];
 //! let span_id = [1u8; 8];
 //! let local_root_span_id = [2u8; 8];
 //!
 //! // First call allocates a record and attaches it.
-//! ThreadContext::new(trace_id, span_id, 1, local_root_span_id, &[(0, "first")]).attach();
-//! ThreadContext::update(trace_id, span_id, 1, local_root_span_id, &[(0, "second")]);
-//! ThreadContext::detach();
+//! OwnedThreadContext::new(trace_id, span_id, 1, local_root_span_id, &[(0, "first")]).attach();
```
scottgerring: I'm kind of confused why we need `new` when owned creates a new record if the record does not exist

libdd-otel-thread-ctx/build.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use std::env;
+
+/// Warn when both ownership modes are requested at once.
+///
+/// `owned-context` and `shared-context` are mutually exclusive, and `owned-context` silently wins
+/// (see `src/linux/mod.rs`). We report that through a build script rather than a `#[deprecated]`
+/// trick or a lint, because a build-script warning doesn't turn into an error under
+/// `clippy -- -D warnings`, which CI runs with `--all-features`.
+fn main() {
```
scottgerring: is a `build.rs` the best way of doing this - could we just have compile time stuff in the top of mod.rs and save adding this?

libdd-otel-thread-ctx/src/linux/shared.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+impl SharedThreadContext {
+    /// Reconstruct a [`SharedThreadContext`] from the record pointer stored in the TLS slot,
+    /// reclaiming the strong reference that [`Self::attach`] moved into the slot.
+    ///
+    /// # Safety
+    ///
+    /// `ptr` must originate from a prior [`Self::attach`] (i.e. it is the record pointer of an
+    /// `Arc<ThreadContext>` whose strong reference was moved into the slot), and that reference
```
scottgerring: Where is the slot stored? Where do we see the drop?

libdd-otel-thread-ctx/src/linux/shared.rs
```diff
@@ -0,0 +1,294 @@
...
+        // Safety: a non-null value in the slot came from a prior `attach` of a shared context (see
+        // the type-level note on not mixing owned and shared attaches).
+        NonNull::new(ThreadContextRecord::detach_raw())
+            .map(|ptr| unsafe { Self::from_record_ptr(ptr) })
+    }
+}
+
+#[cfg(test)]
+// The tests are set to be ignored by Miri, since the inline-asm TLSDESC access isn't supported.
+mod tests {
```
scottgerring, line 107: Should you also add a test covering a single Arc<ThreadContext> being attached to multiple threads verifying they all observe the same record?

libdd-otel-thread-ctx/src/linux/owned.rs
```diff
@@ -0,0 +1,689 @@
...
+        assert_eq!(record.valid.load(Ordering::Relaxed), 1);
+        // 1 (key) + 1 (len) + 16 (root_span_id hex chars) = 18
+        assert_eq!(record.attrs_data_size, 18);
+
+        let _ = OwnedThreadContext::detach();
+    }
+
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn attribute_encoding_basic() {
```
scottgerring: Tests in here like this largely test encoding logic through the lens of the owned wrapper, but the encoding logic is feature agnostic. I don't know that it matters heaps, but the upshot of that is that if you compile for shared mode, tests testing this behaviour aren't included.

scottgerring (approved):
couple of extremely nitty nits, but this looks solid!

libdd-otel-thread-ctx/src/linux/shared.rs
yannham: Good idea :+1: added in 2c3e9decb with some additional refcount testing

libdd-otel-thread-ctx/src/linux/owned.rs
yannham: Ah right, I moved all the previous tests blindly from the parent module to `owned`. Agreed: it's cleaner if they are at the right location, and doing so they will be run no matter which feature is selected :+1: done in 6d6380ad1

PR #2169, opened 2026-06-26
fix(shared-runtime): guard shutdown() against Tokio TLS destruction

rachelyangdog:
During CPython interpreter finalization, thread-local storage is destroyed before atexit handlers fire. SharedRuntime::shutdown() calls runtime.block_on() which internally calls context::enter() to set up Tokio's CONTEXT thread-local. If that TLS slot is already destroyed, context::enter() panics with "The Tokio context thread-local variable has been destroyed", which PyO3 converts to a pyo3_runtime.PanicException. This causes a crash on every uWSGI worker shutdown when using ddtrace >=4.9.x.

Fix: check Handle::try_current().is_thread_local_destroyed() before calling block_on(). If TLS is gone, return Ok(()) early — the OS will clean up remaining Tokio threads on process exit. This eliminates both the panic and the subsequent 60s hang/SIGKILL.

Reproducer: uWSGI app with lazy-apps=true, ddtrace imported via uwsgi import=, 4 workers. SIGTERM triggers the panic on every worker.

# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


libdd-shared-runtime/src/shared_runtime/fork_safe.rs
```diff
@@ -131,6 +131,17 @@ impl ForkSafeRuntime {
     /// exceeded.
     pub fn shutdown(&self, timeout: Option<std::time::Duration>) -> Result<(), SharedRuntimeError> {
         debug!(?timeout, "Shutting down ForkSafeRuntime");
+        // block_on calls context::enter() which accesses Tokio's CONTEXT thread-local.
+        // During CPython interpreter finalization, TLS is destroyed before atexit handlers
+        // fire, causing a panic. Detect this via try_current() and bail out early —
+        // the OS will clean up remaining threads on process exit.
+        if matches!(
+            tokio::runtime::Handle::try_current(),
+            Err(ref e) if e.is_thread_local_destroyed()
```
danielsn, line 140: I worry this catches the symptom, but doesn't actually solve the problem, and we're going to play whack-a-mole with global state being destroyed in a non-deterministic order.

rachelyangdog: I originally created this PR to solve this ticket: https://datadoghq.atlassian.net/browse/APMS-19910. I'm not sure what the root fix would be. Do you have any alternatives in mind?

VianneyRuhlmann: I don't understand why this would only happen on uwsgi. If TLS is destroyed by CPython we should have the same panic on every python app calling shutdown

VianneyRuhlmann: Also this solution means we are skipping the flush on shutdown and losing data. I think we should be able to start a new current thread runtime to block on it.

VianneyRuhlmann: Ok I think we're having the same issue as https://github.com/unbit/uwsgi/issues/859 which explains why it's specific to uwsgi. Maybe we could patch uwsgi to call the python handlers before cleaning up TLS.

wantsui: Interesting, I'm on Python ERE this week for SDK so I'll tackle it from the dd-trace-py angle for this 🤔

wantsui: ~Without fixing it in uwsgi specifically (bc this can happen in other setups), I have a draft attempt here: https://github.com/DataDog/dd-trace-py/pull/18853 .~

~If folks think we should wrap uwsgi to fix this I can change approach too.~

~Edit: Actually Claude wasn't able to fully fix this yet, so the approach will change.~

PR: https://github.com/DataDog/dd-trace-py/pull/18853  will just focus on uwsgi

VianneyRuhlmann:
Adressed in https://github.com/DataDog/dd-trace-py/pull/19086

PR #2171, opened 2026-06-26
feat(data-pipeline)!: OTLP gRPC trace export

bm1549:
# What does this PR do?

Adds OTLP gRPC as a trace-export transport, selectable with `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc`. `OtlpProtocol` gains a `Grpc` variant, and `TraceExporter` dispatches through `OtlpExportMode` (`Http` or `Grpc`). The gRPC path limits response bodies to 4 MiB and reports partial-success responses. Remote-status diagnostics retain the original gRPC code. OTLP resource metadata is built once when the exporter is constructed, including the client-computed-stats marker. gRPC is native-only; wasm32 rejects it during build. The FFI accepts `grpc`, so SDKs using libdatadog can select the transport.

The transport validates its dial authority before creating the exporter. Portless `http://` endpoints use port 80. Valid explicit DNS and IPv6 ports are preserved; empty hosts, malformed ports, and userinfo are rejected. Retryable gRPC errors use bounded exponential backoff with jitter. `RESOURCE_EXHAUSTED` is retried only when the server supplies `google.rpc.RetryInfo`. A positive server delay is used as the backoff base when it fits within the 30-second retry budget; longer delays stop retries instead of retrying before the server requested. Zero or missing delays use the existing 100 ms base. Protobuf requests are shared across attempts instead of deep-cloned.

The fork-safe gRPC transport primitive landed in #2273. This PR integrates it with the exporter and FFI and includes the transport hardening found during review. Completed responses now win over subsequent connection teardown, and transient HTTP/2 stream resets are classified as retryable transport errors. Remote gRPC statuses keep their code and status classification instead of being reported as local I/O failures. The gRPC path sends the standard libdatadog user-agent and entity headers and emits the native trace-export telemetry now used by the other transports.

# Motivation

libdatadog's OTLP trace export was HTTP-only. Java and .NET implement gRPC OTLP separately; PHP, Python, Ruby, and Rust use libdatadog for trace export. Adding gRPC here lets their integrations honor `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc` instead of rejecting it or falling back to another transport.

# Additional Notes

- The transport from #2273 is a custom tonic `GrpcService` over hyper h2c. It creates a connection per request and has no persistent worker, which keeps it safe across forks.
- `OtlpProtocol::from_str("grpc")` now succeeds. Adding a variant to the exhaustive public enum is a breaking change, as noted in the commit footer.
- Transient IO and the retryable gRPC codes (`Cancelled`, `Aborted`, `OutOfRange`, `DataLoss`, `Unavailable`, and `DeadlineExceeded`) get `OTLP_MAX_RETRIES + 1` total attempts. `RetryInfo` controls recoverable throttling delays; other retries use the existing 100 ms exponential base, with up to 100 ms of jitter.
- Remote gRPC statuses map to compatible HTTP statuses for the existing `RequestError` and FFI error classification, while retaining the gRPC code in the message.
- Both OTLP transports preserve `.set_client_computed_stats()` in the resource attributes and request metadata.
- The current transport is plaintext h2c. TLS support and an FFI setter for custom OTLP headers remain follow-ups; the FFI header limitation also applies to HTTP OTLP.

# How to test the change?

- `cargo check -p libdd-data-pipeline`
- `cargo +stable clippy -p libdd-data-pipeline --all-targets --no-deps -- -D warnings`
- `cargo +nightly clippy -p libdd-data-pipeline --test test_trace_exporter_otlp_grpc -- -D warnings`
- `cargo +1.87.0 check -p libdd-data-pipeline --test test_trace_exporter_otlp_grpc`
- `cargo nextest run -p libdd-data-pipeline -E '!test(tracing_integration_tests::)'` (204 passed; 7 Docker-backed integration tests skipped locally)
- `cargo test -p libdd-data-pipeline grpc --lib` (34 passed)
- `cargo test -p libdd-data-pipeline --test test_trace_exporter_otlp_grpc` (2 passed)
- `cargo test -p libdd-data-pipeline --doc` (2 passed)
- `cargo check -p libdd-data-pipeline --target wasm32-unknown-unknown --no-default-features`
- `cargo ffi-test --skip-build --filter trace_exporter` (1 passed; the full FFI build succeeded)
- `cargo +stable clippy -p libdd-data-pipeline --all-targets -- -D warnings` currently stops on a pre-existing `collapsible_match` warning in `libdd-trace-utils/src/tracer_payload.rs`, which this PR does not modify.


libdd-data-pipeline/benches/otlp_grpc_export.rs
```diff
@@ -0,0 +1,115 @@
...
+/// 4-byte big-endian length prefix, then the message body.
+fn grpc_frame(body: &[u8]) -> Vec<u8> {
+    let mut framed = Vec::with_capacity(5 + body.len());
+    framed.push(0u8); // compression flag: 0 = uncompressed
+    framed.extend_from_slice(&(body.len() as u32).to_be_bytes());
+    framed.extend_from_slice(body);
+    framed
+}
+
+pub fn grpc_export_benches(c: &mut Criterion) {
```
yannham: What are we benchmarking in this function? I'm asking because we had slow benchmark jobs recently (and I've been one of the usual offender :whistling:). I would say we should only add benchmark if:

- we bench internal code, and not external libraries (it can be useful to bench end-to-end since we do rely a lot on external libraries, but I don't think this is the right place for those benches)
- we bench performance-sensitive code / code on the hot path
- we are interested in the _evolution_ of the performance of this code. Like, we want to avoid having regression here, or we want to optimize it in the future and thus would like to have a baseline.

Do you think those applies here?

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -6,19 +6,26 @@
...
-/// [`FromStr`](std::str::FromStr) rather than represented here, so an unsupported protocol can
-/// never be constructed and silently mishandled downstream.
+/// All three OTel-standard protocol strings parse successfully; the selection
+/// controls which send path the exporter uses:
+/// - `http/json` and `http/protobuf` → OTLP over HTTP/1.1 via
+///   [`HttpClientCapability`](libdd_capabilities::HttpClientCapability).
+/// - `grpc` → OTLP over HTTP/2 via a tonic [`Channel`](tonic::transport::Channel).
+///
+/// Plaintext gRPC (`http://` scheme, port 4317) is supported. TLS gRPC
+/// (`https://` scheme) is not yet implemented — use a TLS-terminating sidecar.
```
yannham: Nit: I find this comment confused and not very useful. I think the first line and each variant's description is enough.

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -6,19 +6,26 @@
...
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 pub enum OtlpProtocol {
     /// HTTP with a JSON body (`Content-Type: application/json`). The default.
     #[default]
     HttpJson,
     /// HTTP with a protobuf body (`Content-Type: application/x-protobuf`).
     HttpProtobuf,
+    /// gRPC over HTTP/2. Protobuf-encoded body with 5-byte gRPC framing.
+    /// Default port is 4317. Only plaintext (`http://`) is supported.
```
yannham: (Unless any of what comes after that in non-standard, I don't think it's relevant here)

```suggestion
    /// gRPC over HTTP/2
```

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -40,24 +43,30 @@ impl std::str::FromStr for OtlpProtocol {
 impl OtlpProtocol {
     /// The HTTP `Content-Type` for this protocol's body encoding. Crate-internal: the public type
     /// is only constructed/selected by callers; encoding is the exporter's job.
+    /// Only called on the HTTP path; the gRPC path uses tonic's ProstCodec.
     pub(crate) fn content_type(&self) -> http::HeaderValue {
+        #[allow(clippy::unreachable)]
         match self {
             OtlpProtocol::HttpJson => libdd_common::header::APPLICATION_JSON,
             OtlpProtocol::HttpProtobuf => libdd_common::header::APPLICATION_PROTOBUF,
+            OtlpProtocol::Grpc => unreachable!("gRPC path does not call content_type()"),
```
yannham: I would avoid unreachable here, because it makes a hypothesis on every possible caller, past and future. Basically we don't have enough information to prove it's unreachable; if someone somedays adds code that call `protocol.content_type()` without filtering `Grpc` out, this will panic.

Suggestion: either return some default value that vaguely makes sense, or if there is not, make this function return an optional. The caller can then take the responsibility of `unwrap()`, but at least they have to make that choice explicitly.

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -40,24 +43,30 @@ impl std::str::FromStr for OtlpProtocol {
...
         &self,
         req: &libdd_trace_utils::otlp_encoder::ProtoExportTraceServiceRequest,
     ) -> Result<Vec<u8>, serde_json::Error> {
+        #[allow(clippy::unreachable)]
         match self {
             OtlpProtocol::HttpJson => libdd_trace_utils::otlp_encoder::encode_otlp_json(req),
             OtlpProtocol::HttpProtobuf => {
                 Ok(libdd_trace_utils::otlp_encoder::encode_otlp_protobuf(req))
             }
+            OtlpProtocol::Grpc => unreachable!("gRPC path does not call encode()"),
```
yannham: Same remark.

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -99,10 +108,15 @@ mod tests {
     }
 
     #[test]
-    fn grpc_is_rejected_at_parse() {
-        // gRPC is unsupported, so it must not parse into a protocol: an unsupported value can
-        // never be constructed.
-        assert!(OtlpProtocol::from_str("grpc").is_err());
+    fn grpc_parses_successfully() {
```
yannham: Nit: this can be moved to the `protocol_from_str` test above, now that gRPC is not special anymore

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -99,10 +108,15 @@ mod tests {
...
+        // gRPC is now a supported protocol — it must parse without error.
+        assert_eq!(OtlpProtocol::from_str("grpc").unwrap(), OtlpProtocol::Grpc);
+    }
+
+    #[test]
+    fn grpc_config_is_send_sync() {
+        fn assert_send_sync<T: Send + Sync>() {}
+        assert_send_sync::<OtlpGrpcTraceConfig>();
     }
```
yannham: Either we don't really need `OtlpGrpcTraceConfig` to be `Send + Sync`, or if we really do, we will get a compile time error anyway, I believe. I don't think this test is very relevant at runtime.

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -118,6 +132,21 @@ mod tests {
...
+/// Parsed OTLP gRPC trace exporter configuration.
+///
+/// The endpoint URL is validated at build time and retained by the transport worker, which
+/// constructs the tonic [`Channel`](tonic::transport::Channel) on the exporter's shared runtime.
+/// Only the per-request settings below are retained in this config.
+#[derive(Clone, Debug)]
+pub struct OtlpGrpcTraceConfig {
+    /// Custom key-value pairs forwarded as gRPC request metadata.
+    pub headers: Vec<(String, String)>,
+    /// Per-request timeout (applied via [`tokio::time::timeout`]).
```
yannham:
```suggestion
    /// Per-request timeout.
```

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -118,6 +132,21 @@ mod tests {
     }
 }
 
+/// Parsed OTLP gRPC trace exporter configuration.
+///
+/// The endpoint URL is validated at build time and retained by the transport worker, which
+/// constructs the tonic [`Channel`](tonic::transport::Channel) on the exporter's shared runtime.
+/// Only the per-request settings below are retained in this config.
```
yannham:
```suggestion
/// Pre-request OTLP gRPC trace exporter configuration.
```

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+
+//! OTLP gRPC trace exporter.
+//!
+//! Sends an [`ExportTraceServiceRequest`] over a tonic gRPC channel using
+//! plaintext HTTP/2 (`http://` scheme). TLS (`https://`) is not yet supported.
+//!
+//! # gRPC framing
+//! The inner [`ProstCodecImpl`] + tonic's [`Grpc`](tonic::client::Grpc) handle
+//! the 5-byte frame prefix, protobuf encoding, and gRPC trailer parsing
+//! automatically, using [`prost`] for message encoding/decoding.
```
yannham:
```suggestion
```

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+        // `copy_to_bytes` drains the whole buffer correctly even if `DecodeBuf`'s
+        // backing store is ever non-contiguous (`chunk()` only returns the first
+        // contiguous segment, so the old chunk()+advance() form could truncate).
+        let buf = src.copy_to_bytes(src.remaining());
+        match T::decode(buf) {
+            Ok(msg) => Ok(Some(msg)),
+            Err(e) => Err(Status::internal(format!(
+                "Failed to decode protobuf message: {e}"
+            ))),
+        }
```
yannham:
```suggestion
        T::decode(buf).map(Some).map_err(|e| Status::internal(format!(
                "Failed to decode protobuf message: {e}"
        )))
```

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+
+/// gRPC path for the OTLP trace export RPC.
+const GRPC_EXPORT_PATH: &str = "/opentelemetry.proto.collector.trace.v1.TraceService/Export";
+
+/// Metadata header signalling that the client already computed APM stats, so the
+/// agent does not recompute them. Attached per-request when stats are enabled.
+const CLIENT_COMPUTED_STATS_HEADER: &str = "datadog-client-computed-stats";
+
+// ---------------------------------------------------------------------------
+// Prost codec implementation — tonic 0.14 removed ProstCodec from its public
```
yannham: Very nit: the code part could go into a submodule (even in the same file), as it quite separate/well-scoped boilerplate.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+/// A connected gRPC transport for OTLP trace export.
+///
+/// Holds the per-export config, normalized request origin, and shared channel state populated by
+/// [`OtlpGrpcChannelWorker`]. The worker is registered with the exporter's shared runtime so the
+/// tonic channel is always created on that runtime and rebuilt after fork lifecycle restarts.
+#[derive(Clone, Debug)]
+pub(crate) struct OtlpGrpcTransport {
+    pub(crate) config: OtlpGrpcTraceConfig,
+    origin: http::Uri,
+    channel_rx: watch::Receiver<Option<Channel>>,
```
yannham: What is this `Receiver` used for? Might deserve a small description comment.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+
+impl Drop for PublishedChannelGuard {
+    fn drop(&mut self) {
+        // `before_fork` cancels and joins the worker future before dropping the runtime. Clearing
+        // here prevents a send immediately after `after_fork_*` from observing the old channel.
+        self.0.send_replace(None);
+    }
+}
+
+impl OtlpGrpcChannelWorker {
```
yannham: I don't entirely understand the role of the worker. I initially thought its role would be send the gRPC stuff through the channel itself, in a concurrent way, so that the main async `send_otlp_trace_grpc` wouldn't have to wait for stuff to be send. But it's only managing a channel.

[ADDENDUM]: after asking Claude, it seems the issue is the following: the mere act of initializing a channel creates a background task on the current tokio runtime. If we stored it directly in `OtlpGrpcTransport`, since libdatadog drops the runtime upon fork, the background task will silently be gone and will lead to errors once we try to send trough it again after a fork.  This background tonic task is the one that needs to be recreated after fork. The worker is a simple wrapper that mostly does nothing but registering to fork events, so that it can properly close the channel before a fork and recreate it afterwards. I think this is not obvious to deduce from the code and the comments (which mention technical details like "a tower Buffer service" but to be honest I had no idea what this was).

I would suggest to add a concise, high-level comment on the worker or wherever it makes sense, that clearly states the problem and the solution.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+        std::future::pending::<()>().await;
+    }
+}
+
+#[async_trait]
+impl Worker for OtlpGrpcChannelWorker {
+    async fn run(&mut self) {}
+
+    async fn initial_trigger(&mut self) {
+        self.publish_until_restarted().await;
```
yannham: @VianneyRuhlmann is that ok to "loop" (in async terms) like this on `initial_trigger`? In general if this worker is only here to handle the tonic channel lifecycle, I would expect `initial_trigger()` to do some work, `shutdown` and `reset` as well, but `trigger` to just do nothing

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+    let scheme = uri.scheme().ok_or_else(|| {
+        TraceExporterError::Builder(BuilderErrorKind::InvalidUri(
+            "gRPC endpoint must include a URI scheme".to_string(),
+        ))
+    })?;
+    if scheme == &http::uri::Scheme::HTTPS {
+        return Err(TraceExporterError::Builder(
+            BuilderErrorKind::InvalidConfiguration(
+                "gRPC TLS (https://) is not yet supported; use http:// and a \
+                 TLS-terminating sidecar if encryption is required"
```
yannham: What's a TLS-terminating sidecar?

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+            endpoint: channel_endpoint,
+            channel_tx,
+        },
+    ))
+}
+
+/// Send an OTLP trace export request over gRPC.
+///
+/// Uses the `transport.channel` tonic channel with [`ProstCodecImpl`] for
+/// encoding/decoding. Custom metadata headers, the test session token, and
```
yannham:
```suggestion
/// Custom metadata headers, the test session token, and
```

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+    #[test]
+    fn grpc_origin_preserves_and_normalizes_path_prefix() {
+        let (transport, _) =
+            build_grpc_transport("http://localhost:4317/otel/", test_config()).unwrap();
+        assert_eq!(transport.origin.path(), "/otel");
+    }
+
+    #[test]
+    fn grpc_status_transient_map_to_io_with_correct_kind() {
+        // Unavailable → ConnectionRefused; DeadlineExceeded → TimedOut (not refused).
```
yannham: Nit: it's paraphrasing the code

```suggestion
```

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,514 @@
...
+            req.metadata().get(CLIENT_COMPUTED_STATS_HEADER).unwrap(),
+            "yes"
+        );
+    }
+
+    #[test]
+    fn grpc_transport_is_clone() {
+        let (transport, _) = build_grpc_transport("http://localhost:4317", test_config()).unwrap();
+        let _clone = transport.clone();
+    }
```
yannham: I don't think this test is very useful.

```suggestion
```

libdd-data-pipeline/src/otlp/mod.rs
```diff
@@ -29,9 +29,15 @@
 
 pub mod config;
 pub mod exporter;
+// gRPC OTLP export depends on tonic, which does not build for wasm32.
+#[cfg(not(target_arch = "wasm32"))]
```
yannham: Not for this PR, but interestingly, this might mean we'll have to move gRPC to a capability (if NodeJS wants to use this at some point)

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -839,11 +874,12 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
                 info_fetcher: info_fetcher_handle,
                 #[cfg(feature = "telemetry")]
                 telemetry: telemetry_handle,
+                otlp_grpc: otlp_grpc_worker_handle,
```
yannham: I think this has to be `cfg`-gated, or this won't compile under WASM.

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -1219,4 +1255,50 @@ mod tests {
...
+            .set_url("http://localhost:8126")
+            .set_otlp_protocol(crate::otlp::OtlpProtocol::Grpc);
+        let result = builder.build::<NativeCapabilities>();
+        assert!(
+            result.is_ok(),
+            "gRPC protocol without endpoint must use agent path, not fail: {result:?}"
+        );
+    }
+
+    // build() spins up the shared runtime, whose worker hits a syscall Miri can't execute.
```
yannham: (Nit: other similarly Miri-ignored tests don't have this comment, and in general IMHO we should rather justify to NOT ignore a test for Miri, rather than the opposite)

```suggestion
```

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -61,6 +64,19 @@ use tokio::task::JoinSet;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, warn};
 
+/// Selects the OTLP export transport for a [`TraceExporter`] instance.
+///
+/// Only one variant is active per instance; mutual exclusivity is enforced at
+/// build time.
```
yannham:
```suggestion
```

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -662,23 +684,26 @@ impl<
...
-            r.service = self.metadata.service.clone();
-            r.env = self.metadata.env.clone();
-            r.app_version = self.metadata.app_version.clone();
-            r.language = self.metadata.language.clone();
-            r.tracer_version = self.metadata.tracer_version.clone();
-            r.runtime_id = self.metadata.runtime_id.clone();
-            r.client_computed_stats = self.otlp_stats_enabled;
-            r
-        };
+        let resource_info = self.build_otlp_resource_info();
```
yannham: Not really related to this PR, but should we cache the resource info in the struct instead of creating it and cloning strings on every request? Unless the metadata can change? Or maybe `map_trace_to_otlp` should take `LanguageMetadata` instead. Or make `ResourceInfo` borrowed. But it does sounds wasteful to clone 7ish strings on each send (that are probably not changing often or at all), just to pass a reference to them to `map_traces_to_otlp` after.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -820,12 +876,21 @@ impl<
...
+        match &self.otlp {
+            Some(OtlpExportMode::Http(config)) => {
+                return self.send_otlp_traces_inner(traces, config).await;
+            }
+            #[cfg(not(target_arch = "wasm32"))]
+            Some(OtlpExportMode::Grpc(transport)) => {
+                return self.send_otlp_grpc_inner(traces, transport).await;
+            }
+            None => {}
         }
```
yannham:
```suggestion
        if let Some(otlp) = &self.otlp {
            libdd_trace_utils::span::trace_utils::drop_chunks(&mut traces);
            if traces.is_empty() {
                return Ok(AgentResponse::Unchanged);
            }

            match otlp {
                OtlpExportMode::Http(config) => {
                    return self.send_otlp_traces_inner(traces, config).await;
                }
                #[cfg(not(target_arch = "wasm32"))]
                OtlpExportMode::Grpc(transport) => {
                    return self.send_otlp_grpc_inner(traces, transport).await;
                }
            }
        }
```

libdd-data-pipeline/Cargo.toml
```diff
@@ -46,10 +46,17 @@ libdd-tinybytes = { version = "1.1.1", path = "../libdd-tinybytes", features = [
     "bytes_string",
     "serialization",
 ] }
-
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { version = "1.23", features = ["time", "test-util"], default-features = false }
 libdd-capabilities-impl = { version = "3.0.0", path = "../libdd-capabilities-impl", default-features = false }
+# tonic's transport (hyper/tokio/socket2) does not build for wasm32, so the gRPC
+# OTLP transport — and prost, which only the gRPC exporter uses — are gated off
+# wasm targets along with the code that uses them.
```
yannham:
```suggestion
# tonic's transport does not build for wasm32. Prost is only used by the gRPC
# exporter, so it's gated as well.
```

libdd-data-pipeline/tests/test_trace_exporter_otlp_grpc.rs
```diff
@@ -0,0 +1,242 @@
...
+//!
+//! Starts an in-process HTTP/2 gRPC server using the `h2` crate, configures
+//! a [`TraceExporter`] with `OtlpProtocol::Grpc`, sends a trace, and asserts
+//! that the server received a valid [`ExportTraceServiceRequest`] containing
+//! the expected span.
+
+#[cfg(test)]
+mod grpc_export_tests {
+    use bytes::Bytes;
+    use h2::server;
```
yannham, line 6: Any reason for using a `h2` server instead of the usual http mock server we've been using elsewhere in tests?

yannham (commented):
First, apologies for waiting so long before giving you a first review. We'll avoid that kind of delay in the future.

As far as the PR goes, I have left a bunch of comments but there's no big technical blocker on my side. However I can't approve right away because the size impact is quite important. We don't have binary size quality gates (yet), but downstream consumers of libdatadog (tracers) do, which can lead to an unfortunate situation where this would get merged but tracers would be unable to update to future releases.

A first naive question: could/should this be feature-gated? I assume you'd want all tracers to support otlp out of the box, so the answers might be no, bu asking just in case.

A good test for the impact of the binary size increase: would you mind trying to create a PR against dd-trace-py pointing to this PR's libdatadog ref, and see if it triggers the size gates for e.g. serverless?

yannham:
So, we've had a discussion with @VianneyRuhlmann about the whole worker situation. We believe we can side-step the "task in the background that needs to be restared" entirely by plugging in our own `GrpcService` implementation. It seems tonic provides a complicated default implementation through `Channel` with this background tower `Service` that reads through a channel and whatnot. But it's customizable and we can provide our own implementation of `GrpcService` instead, which seems to be rather simple (basically send some request). Doing so there's no background task anymore, we can get rid of the `wait` channel, and of the worker as well. It might even be possible to remove the "transport" feature entirely and save some binary size, though this has to be seen.

libdd-data-pipeline/src/trace_exporter/mod.rs
bm1549: done in 812e43b93

libdd-data-pipeline/src/otlp/config.rs
bm1549: done in 812e43b93

libdd-data-pipeline/src/otlp/config.rs
bm1549: done in 812e43b93

libdd-data-pipeline/src/otlp/config.rs
bm1549: done in 812e43b93

libdd-data-pipeline/src/otlp/config.rs
bm1549: done in 812e43b93

libdd-data-pipeline/src/otlp/config.rs
bm1549: done in 812e43b93

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Moot now - we dropped the worker entirely in #2273 and the transport dials a fresh connection per send (no persistent channel, nothing to restart across fork). @VianneyRuhlmann fyi, that side-steps the whole initial_trigger question.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Worker's gone - #2273 replaced the tonic Channel + worker with a per-request GrpcService that holds no background task, so there's nothing left to document here.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Gone in #2273 - no more watch channel now that the transport is per-request.

libdd-data-pipeline/src/trace_exporter/builder.rs
bm1549: That field went away with the worker in #2273. gRPC is gated off wasm at the module and the builder rejects it at build time, so there's nothing to cfg-gate here.

libdd-data-pipeline/src/trace_exporter/mod.rs
bm1549: OtlpExportMode reworked in the rebuild; it's a one-line doc now.

libdd-data-pipeline/src/trace_exporter/builder.rs
bm1549: Those builder tests were rewritten in the rebuild; comment's gone.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Transport was rewritten in #2273 - that test no longer exists.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: That code was rewritten in #2273; comment's gone.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Rewritten in #2273, n/a now.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Module doc was trimmed in #2273.

libdd-data-pipeline/Cargo.toml
bm1549: Deps section was reworked for the new transport in #2273; that comment's shorter now.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Done in #2273 - it's in its own mod prost_codec now.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Done in #2273 - it's T::decode(buf).map(Some).map_err(...) now.

libdd-data-pipeline/src/otlp/config.rs
bm1549: Applied in #2273 as `/// Per-request OTLP gRPC trace exporter configuration.` - took it as Per-request, assuming Pre-request was a typo.

libdd-data-pipeline/src/otlp/config.rs
bm1549: Done in #2273.

libdd-data-pipeline/src/otlp/config.rs
bm1549: Agreed - dropped it; the redesigned config in #2273 has no Send+Sync test.

libdd-data-pipeline/tests/test_trace_exporter_otlp_grpc.rs
bm1549: httpmock here is HTTP/1.1-only (no http2 feature) and has no trailer support, so it can't serve gRPC - the unary status rides in HTTP/2 trailers. That's why the test hand-rolls an h2 server.

libdd-data-pipeline/src/trace_exporter/mod.rs
bm1549: Did it as a nice bonus in 812e43b93 but let me know if you'd prefer I split this out into another PR

libdd-data-pipeline/src/otlp/mod.rs
bm1549: In the future yes, but today it's probably fine since Node.js can use the http/protobuf exporter, which should satisfy most users

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: A local proxy (Envoy/nginx, or the Agent) that accepts TLS and forwards plaintext h2c to the endpoint. Is there an alternative suggestion you'd give for the wording here?

yannham: I don't know if I have a better idea. I might have been slightly confused by "sidecar" here because it's a specific thing in dd-trace-php/libdatadog. Maybe substituting sidecar for proxy if you think that doesn't change the meaning too much?

bm1549: Proxy works fine - changed in the split PR
https://github.com/DataDog/libdatadog/pull/2273

libdd-data-pipeline/tests/test_trace_exporter_otlp_grpc.rs
yannham: In fact `httpmock` does have an "http2" features, but it's just not advertised for some reason. It enables the corresponding features on hyper. I think in the transport PR you do some tests that might be hard to reproduce with the simplest API of httpmock, but it might be worth looking at.

libdd-data-pipeline/benches/otlp_grpc_export.rs
bm1549: removed

libdd-data-pipeline/tests/test_trace_exporter_otlp_grpc.rs
bm1549: Just to follow up here - httpmock has an undocumented http2 feature, but it can’t return HTTP/2 trailers. Since this test needs a trailing `grpc-status: 0`, it still needs the small h2 server.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -383,10 +396,29 @@ fn grpc_status_to_error(status: Status) -> TraceExporterError {
...
+            ))
+        }
+        code => {
+            let http_status = match code {
+                Code::InvalidArgument => http::StatusCode::BAD_REQUEST,
+                Code::Unauthenticated => http::StatusCode::UNAUTHORIZED,
+                Code::PermissionDenied => http::StatusCode::FORBIDDEN,
+                Code::NotFound => http::StatusCode::NOT_FOUND,
+                Code::AlreadyExists => http::StatusCode::CONFLICT,
+                Code::ResourceExhausted => http::StatusCode::TOO_MANY_REQUESTS,
```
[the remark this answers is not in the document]
bm1549: done in 061c907

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -656,6 +663,31 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
...
 
+        let otlp_timeout = self
+            .connection_timeout
+            .map(Duration::from_millis)
+            .unwrap_or(DEFAULT_OTLP_TIMEOUT);
+        #[cfg(not(target_arch = "wasm32"))]
+        let grpc_transport = match self.otlp_endpoint.as_deref() {
+            Some(url) if self.otlp_protocol == OtlpProtocol::Grpc => Some(build_grpc_transport(
+                url,
+                OtlpGrpcTraceConfig {
```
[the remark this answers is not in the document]
bm1549: done in 061c907

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -1077,23 +1143,6 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
...
-    /// - OTLP and agentless cannot both be configured.
-    /// - Agentless cannot be combined with a caller-supplied agent URL.
-    /// - Log output cannot be combined with OTLP or agentless trace export.
-    /// - [`Self::set_agentless_timeout`] requires [`Self::set_agentless_endpoint`].
-    /// - [`Self::set_agentless_stats_endpoint`] requires agentless trace export
-    ///   ([`Self::set_agentless_endpoint`]) and is incompatible with OTLP stats
-    ///   ([`Self::set_otlp_metrics_endpoint`]).
-    ///
-    /// OTLP and an agent URL may coexist: the agent URL is still useful for auxiliary
-    /// agent endpoints (info, stats) even when trace payloads are routed to OTLP.
```
yannham, line 1103: Any reason for removing all of this? Did of it become obsolete?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -68,6 +73,20 @@ const INFO_ENDPOINT: &str = "/info";
...
 const V1_TRACES_ENDPOINT: &str = "/v1.0/traces";
+#[cfg(not(target_arch = "wasm32"))]
+const OTLP_GRPC_MAX_RETRY_DELAY: Duration = Duration::from_secs(30);
+
+#[cfg(not(target_arch = "wasm32"))]
+fn grpc_retry_delay(attempt: u32, retry_after: Option<Duration>) -> Duration {
+    let initial_delay = retry_after.unwrap_or_else(|| Duration::from_millis(OTLP_RETRY_DELAY_MS));
+    let multiplier = 2u32
+        .checked_pow(attempt.saturating_sub(1))
+        .unwrap_or(u32::MAX);
```
yannham:
```suggestion
        .saturating_pow(attempt.saturating_sub(1));
```

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -68,6 +73,20 @@ const INFO_ENDPOINT: &str = "/info";
...
+    initial_delay
+        .checked_mul(multiplier)
+        .unwrap_or(Duration::MAX)
```
yannham:
```suggestion
        .saturating_mul(multiplier)
```

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -53,15 +95,16 @@ pub(crate) mod prost_codec {
         // Shared with the `Encoder` impl below. tonic's `EncodeBuf`/`DecodeBuf` constructors are
         // private to the crate, so tests exercise this generic-over-`BufMut`/`Buf` core directly
         // instead of going through the `Encoder`/`Decoder` traits (see `codec_tests`).
-        fn encode_into(item: T, dst: &mut impl bytes::BufMut) -> Result<(), Status> {
-            item.encode(dst)
+        fn encode_into(item: Arc<T>, dst: &mut impl bytes::BufMut) -> Result<(), Status> {
```
yannham: Out of curiosity, why do we need to switch to `Arc` here? Also, couldn't the encode just take a reference, leaving the choice to the caller whether to use `Rc`, `Arc`, `&` or whatnot

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -314,23 +373,37 @@ pub(crate) async fn send_otlp_traces_grpc(
...
     .await
-    .map_err(|_| TraceExporterError::Io(std::io::Error::from(std::io::ErrorKind::TimedOut)))?
+    .map_err(|_| GrpcExportError::Retryable {
+        error: TraceExporterError::Io(std::io::Error::from(std::io::ErrorKind::TimedOut)),
+        retry_after: None,
+    })?
 }
 
-// Insert the pre-validated custom headers, the optional test-session token, and (when enabled) the
-// client-computed-stats marker into the request metadata.
```
yannham, line 331: Same question: is this removal intentional?

yannham (approved):

libdd-data-pipeline/src/trace_exporter/mod.rs
bm1549: done in e675f64

libdd-data-pipeline/src/trace_exporter/mod.rs
bm1549: done in e675f64

libdd-data-pipeline/src/trace_exporter/builder.rs
bm1549: Yeah. it said there were exactly three transports and described OTLP as only http/json. It also duplicated the checks immediately below, so I kept the one-line function doc and updated the public setter docs instead

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Yep, intentional. `attach_metadata` already says what it does. The comment was just paraphrasing the body so I left it out

libdd-data-pipeline/src/otlp/grpc_exporter.rs
bm1549: Updated in e675f643 - good call! 😄

mabdinur (approved):
Reviewed the current head, including the per-request gRPC transport lifecycle, endpoint validation, retry/status handling, OTLP metadata, FFI selection, and target gating. Focused compile, format, clippy, tests, docs, and FFI validation pass locally. I found no blocking correctness issues.

PR #2172, opened 2026-06-29
feat(telemetry)!: make telemetry worker wasm-compatible for the TraceExporter

Aaalibaba42:
# What?

Port the telemetry stack (`libdd-telemetry`, `libdd-data-pipeline`) to compile and run on `wasm32-unknown-unknown`. `TelemetryWorker`, `TelemetryWorkerHandle`, and `TelemetryClient` become generic over a capability bundle `C: HttpClientCapability + SleepCapability`; native call sites pin to `NativeCapabilities`.

# Why?

Telemetry was previously hard-gated behind `#[cfg(not(target_arch = "wasm32"))]` and was a no-op on that target for the `TraceExporter`.

# How?

- Every `tokio::time` call (sleep, timeout) is replaced by a `tokio::select!` race against `<C as SleepCapability>::new().sleep(…)`, which resolves to `setTimeout` on wasm.
- `std::time` replaced by `web-time` (`Performance.now()` / `Date.now()` on wasm, re-exports `std::time` on native).
- Shutdown signaling adds a `tokio::sync::Notify`-based async path alongside the existing native `Condvar` sync path.
- The old `MockClient`/file-endpoint logic in `http_client.rs` is deleted and ported to `NativeCapabilities` in `libdd-capabilities-impl`.

# Additional Notes

- `TelemetryClientBuilder::build` now returns `Result` (was panicking on missing fields) — Rust API break.
- `TelemetryClient::start` is now sync via `try_send_msg` — Rust API break.
- FFI and sidecar crates are unaffected at the ABI level; they use a local `type` alias pinned to `NativeCapabilities`.


libdd-telemetry/src/worker/mod.rs
```diff
@@ -925,51 +938,176 @@ impl TelemetryWorker {
...
 }
 
+/// Cross-platform shutdown signaling.
+///
+/// Native callers can sync-block via the `Mutex<bool>`/`Condvar` pair (see
+/// [`TelemetryWorkerHandle::wait_for_shutdown`]); any caller can await
+/// asynchronously via the `Notify` (see
+/// [`TelemetryWorkerHandle::wait_for_shutdown_async`]) — `Notify` is pure
+/// atomics and works on wasm too. `shutdown_finished` triggers both so either
+/// path observes the same event.
```
VianneyRuhlmann: I wonder wether this is necessary, `wait_for_shutdow` is only used when the worker is spawned on it's own thread not as a SharedRuntime worker. Since `spawn` is not available on wasm we can platform-gate shutdown

libdd-telemetry/src/worker/mod.rs
```diff
@@ -290,25 +305,27 @@ mod serialize {
...
+                return TelemetryActions::Lifecycle(deadline_action);
             };
 
-            // Otherwise run it in a timeout against the mailbox
-            match tokio::time::timeout_at(deadline.into(), self.mailbox.recv()).await {
-                Ok(mailbox_action) => mailbox_action,
-                Err(_) => Some(TelemetryActions::Lifecycle(*deadline_action)),
+            // Runtime-agnostic timeout: race the mailbox against a capability-driven sleep
+            // instead of `tokio::time::timeout_at`, which requires a tokio reactor (not
+            // available on wasm where we run on the JS event loop).
```
VianneyRuhlmann: You don't need to re-explain the capabilities every time the old comment is fine

libdd-telemetry/src/worker/mod.rs
```diff
@@ -290,25 +305,27 @@ mod serialize {
...
+            let sleeper = <C as SleepCapability>::new();
+            tokio::select! {
+                biased;
+                mailbox_action = self.mailbox.recv() => mailbox_action,
+                _ = sleeper.sleep(remaining) => Some(TelemetryActions::Lifecycle(deadline_action)),
```
VianneyRuhlmann, line 325: Not specifically for this PR but should we add a helper for timeout to the sleeper capability ?

libdd-telemetry/src/worker/mod.rs
```diff
@@ -925,51 +938,176 @@ impl TelemetryWorker {
...
-    // since the runtime is not guaranteed to exist for the lifetime of the worker.
+    // Used to spawn cancellation tasks on native. Should be None when running as a
+    // SharedRuntime worker, since the runtime is not guaranteed to exist for the
+    // lifetime of the worker. On wasm we spawn via `wasm_bindgen_futures::spawn_local`
+    // directly, so no handle is needed.
+    #[cfg(not(target_arch = "wasm32"))]
     runtime: Option<runtime::Handle>,
-
     contexts: MetricContexts,
+    _phantom: PhantomData<fn() -> C>,
```
Eldolfin, line 980: Why is this needed ? Why make the whole struct generic instead of only the functions that need it ? ( `cancel_requests_with_deadline`, and its callers `wait_for_shutdown_deadline` `wait_for_shutdown_deadline_async`)

I tried refactoring it as such and it compiles, I can send the diff instead of making suggestions because there is too many places it's being called.

libdd-telemetry/examples/tm-ping.rs
```diff
@@ -46,13 +51,26 @@ fn build_request<'a>(
...
+        .body(Bytes::from(serde_json::to_vec(telemetry)?))?;
 
-    let resp = client.request(req).await?;
+    let resp = select! {
+        biased;
+        result = client.request(req) => result?,
+        _ = sleeper.sleep(timeout) => {
+            return Err(anyhow::anyhow!("Telemetry request timed out"));
+        }
+    };
```
Eldolfin, line 73: Not sure how common this pattern will be but we could have a helper function `client.request_with_timeout(req, timeout)`

libdd-telemetry/src/worker/mod.rs
```diff
@@ -1021,24 +1159,42 @@ impl TelemetryWorkerHandle {
...
-        let Some(runtime) = &self.runtime else {
-            tracing::error!("Cannot schedule cancellation deadline: no runtime handle available");
-            return;
-        };
+    /// Schedule a deferred `CancellationToken::cancel()` to fire after `deadline`.
+    ///
+    /// Capability-driven sleep so the same code path works on native (tokio
+    /// reactor) and wasm (JS `setTimeout` via [`SleepCapability`]). The spawn
+    /// mechanism is platform-specific but the public API and observable
+    /// behavior are identical.
```
VianneyRuhlmann: No need to re-explicit the wasm specific behavior in all methods (especially public ones) this is an implementation detail

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -592,30 +588,34 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
...
             match telemetry {
-                Some(Ok((client_tel, worker))) => {
+                Some((client_tel, worker)) => {
                     let handle = shared_runtime.spawn_worker(worker, false).map_err(|e| {
                         TraceExporterError::Builder(BuilderErrorKind::InvalidConfiguration(
                             e.to_string(),
                         ))
                     })?;
-                    client_tel.start().await;
+                    let _ = client_tel.start();
```
Eldolfin: Shouldn't the error be logged here at least ?

libdd-crashtracker/src/crash_info/telemetry.rs
```diff
@@ -342,9 +342,13 @@ impl TelemetryCrashUploader {
         self.send_telemetry_payload(&payload).await
     }
 
-    /// Helper to perform actual HTTP (or file) submission via configured telemetry client
+    /// Helper to perform actual HTTP submission via the native HTTP capability.
     async fn send_telemetry_payload(&self, payload: &data::Telemetry<'_>) -> anyhow::Result<()> {
-        let client = libdd_telemetry::worker::http_client::from_config(&self.cfg);
+        use libdd_capabilities::{HttpClientCapability, SleepCapability};
+        use libdd_capabilities_impl::NativeCapabilities;
```
Eldolfin: nit: can be put at the top

libdd-telemetry/src/worker/mod.rs
```diff
@@ -1260,32 +1426,46 @@ impl TelemetryWorkerBuilder {
...
 
-    /// Spawns a telemetry worker task in the current tokio runtime
-    /// The worker will capture a reference to the runtime and use it to run its tasks
-    pub fn spawn(self) -> (TelemetryWorkerHandle, JoinHandle<()>) {
+    /// Spawns a telemetry worker task in the current tokio runtime.
+    ///
+    /// Native-only convenience entry point. Wasm callers should construct via
+    /// [`Self::build_worker`] and hand the worker to a
+    /// [`SharedRuntime`](libdd_shared_runtime::SharedRuntime) like
+    /// `LocalRuntime`.
```
VianneyRuhlmann: That seems a bit heavy, especially since the method is not available on wasm. It should be enough to document the build_worker method

libdd-telemetry/src/worker/mod.rs
```diff
@@ -1260,32 +1426,46 @@ impl TelemetryWorkerBuilder {
...
 
         (worker_handle, join_handle)
     }
 
-    /// Spawns a telemetry worker in a new thread and returns a handle to interact with it
-    pub fn run(self) -> anyhow::Result<TelemetryWorkerHandle> {
+    /// Spawns a telemetry worker in a new thread and returns a handle to interact with it.
+    ///
+    /// Native-only convenience entry point; spawning a fresh tokio runtime is not
+    /// meaningful on wasm.
```
VianneyRuhlmann: Same no need to clutter documentation

libdd-data-pipeline/src/telemetry/mod.rs
```diff
@@ -117,13 +118,28 @@ impl TelemetryClientBuilder {
     }
 
     /// Builds the telemetry client.
-    pub fn build(self) -> (TelemetryClient, TelemetryWorker) {
-        #[allow(clippy::unwrap_used)]
+    ///
+    /// `C` is the capability bundle (`NativeCapabilities` on native, `WasmCapabilities` on wasm).
+    pub fn build<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static>(
+        self,
+    ) -> Result<(TelemetryClient<C>, TelemetryWorker<C>), TelemetryError> {
```
VianneyRuhlmann, line 128: Although I agree that this is a good change, I think that changing a public function signature when it's not directly linked to your PR will make it both harder to review and force your PR to be marked as breaking. It's fine to keep it for this one but I think we should avoid this kind of unrelated changes.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -142,7 +141,10 @@ fn add_path(url: &Uri, path: &str) -> Uri {
 pub use libdd_trace_utils::tracer_metadata::TracerMetadata;
 
 /// Handles for the background workers owned by a [`TraceExporter`].
-#[cfg(not(target_arch = "wasm32"))]
+///
+/// Stored on both native and wasm so the JS host can drive symmetric shutdown
+/// via [`TraceExporter::shutdown_async`]. On native this drives the tokio
+/// workers; on wasm it drives the `LocalRuntime` worker entries.
```
VianneyRuhlmann: I don't think we need a wasm specific comment since the interface for workers is generic

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -269,53 +270,51 @@ impl<
...
+            _ = self.shutdown_workers() => Ok(()),
+            _ = sleeper.sleep(timeout) => Err(TraceExporterError::Shutdown(
+                ShutdownError::TimedOut(timeout),
+            )),
         }
     }
 
     async fn shutdown_workers(self) {
+        // Sequential rather than the previous `JoinSet`-driven concurrent stop:
+        // there are at most ~3 workers and each stop is fast, so the parallelism
```
VianneyRuhlmann: That's not true some stop require flushing to the agent which can be long especially if we have to retry. This is critical as this is delaying the tracer shutdown an thus the time to restart the customer app

libdd-telemetry/Cargo.toml
```diff
@@ -32,13 +31,24 @@ tracing = { version = "0.1", default-features = false }
...
 hashbrown = "0.15"
 bytes = "1.4"
+libdd-capabilities = { version = "2.0.0", path = "../libdd-capabilities" }
 libdd-common = { version = "5.0.0", path = "../libdd-common", default-features = false }
 libdd-shared-runtime = { version = "1.0.0", path = "../libdd-shared-runtime", default-features = false }
 libdd-ddsketch = { version = "1.0.1", path = "../libdd-ddsketch" }
+# Cross-platform `Instant`/`SystemTime`. On native it re-exports `std::time`; on
+# wasm32-unknown-unknown it backs by `Performance.now()` and `Date.now()` via
+# `js-sys`, which avoids the `time not implemented on this platform` panic that
+# `std::time::Instant::now()` raises on that target.
```
VianneyRuhlmann: No need to clutter the cargo.toml. Documenting web-time usage in rust code is enough imo, or maximum a single line comment

libdd-telemetry/src/worker/mod.rs
```diff
@@ -808,52 +825,48 @@ impl TelemetryWorker {
...
 
         debug!(
             worker.runtime_id = %self.runtime_id,
             http.timeout_ms = timeout_ms,
             "Sending HTTP request"
         );
 
+        // Runtime-agnostic timeout: race the request against a capability-driven sleep,
+        // mirroring `agent_info::fetcher::fetch_and_hash_response`. Same pattern, same
+        // reason: `tokio::time::sleep` needs a tokio reactor that wasm doesn't provide.
```
VianneyRuhlmann: It's fine to not re-explain the purpose of capability at every call-site

libdd-telemetry/src/worker/mod.rs
```diff
@@ -1197,29 +1363,29 @@ impl TelemetryWorkerBuilder {
...
-    /// the worker task cancellation deadlines. Pass `None` when the worker will be run via a
-    /// [`SharedRuntime`](libdd_shared_runtime::SharedRuntime).
-    pub fn build_worker(
+    /// On native, the optional `tokio_runtime` handle is stored in the worker handle and
+    /// used to drive cancellation deadlines from
+    /// [`TelemetryWorkerHandle::cancel_requests_with_deadline`]. Pass `None` when the worker
+    /// will be run via a [`SharedRuntime`](libdd_shared_runtime::SharedRuntime).
+    ///
+    /// On wasm, cancellation deadlines are driven by `wasm_bindgen_futures::spawn_local` and
+    /// no runtime handle is required.
```
VianneyRuhlmann: Wasm always uses the SharedRuntime right ? There should be any need to run on spawn_local

libdd-telemetry/src/worker/scheduler.rs
```diff
@@ -1,7 +1,10 @@
 // Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
-use std::time::{Duration, Instant};
+use std::time::Duration;
+// web_time::Instant is std::time::Instant on native and a Performance.now()
+// backed shim on wasm32-unknown-unknown (std::time::Instant panics there).
```
VianneyRuhlmann: nit: I don't think this needs to be called out at every import

VianneyRuhlmann (changes requested):
The change in shutdown_worker really needs to be discussed further as it can negatively impact shutdown time. Also a lot of comment are very verbose and at wasm specific explanation on public methods doc that shouldn't be cluttered with implementation details.

Aaalibaba42:
> The change in shutdown_worker really needs to be discussed further as it can negatively impact shutdown time

We can always have the previous native implementation as is in a `#[cfg(not(target_arch="wasm32"))]` to make sure we have no impact on native

libdd-telemetry/src/worker/mod.rs
Aaalibaba42: Mmh yeah it would compile but the point of the capabilities like this is to do static dispatch for a specific platform, a caller creates the TelemetryWorker & Handle pair for either native or wasm and then everything works the same. Otherwise nothing prevents a caller from creating Native/Wasm versions of the the structure via `build_worker` and then call the method with the other implementation. 

`build_worker<C>` returning `(TelemetryWorkerHandle<C>, TelemetryWorker<C>)` is the "forced" coupling that lead this. I'm not sure you can achieve the same invariants without that.

libdd-telemetry/examples/tm-ping.rs
Aaalibaba42: I'm a bit reluctant to change the shape of the client Trait directly, but I wouldn't mind a helper function in `http.rs`, but then it wouldn't be "pinned" on a capability but would just be generic over it, so caller would have to specify the capability again. I'm not sure what's the best here

Eldolfin: I'm thinking either a method with a default implementation on the trait (so that if there is a simpler/more efficient way to do it in some implementations it can be overridden), or maybe an extension trait to keep it the main trait simple.

Aaalibaba42: Like Vianney's suggestion for the sleep utility, I wouldn't mind it too much, but I think it should be done in a subsequent PR

libdd-telemetry/src/worker/mod.rs
Aaalibaba42: Yep, I'm fine with this for a subsequent PR

https://github.com/DataDog/libdatadog/pull/2172#discussion_r3512894394

libdd-telemetry/src/worker/mod.rs
Aaalibaba42: Yep, good catch, this whole codepath is effectively moot. I'll gate it out.

libdd-data-pipeline/src/telemetry/mod.rs
Aaalibaba42: Yeah you are right, I seem to recall claude unwrapping stuff somewhere and this improvement seemed like a good idea anyway, I didn't think about api breakage stuff

Aaalibaba42: I think I'd rather keep it here for now, but I'll keep in mind to be more cautious of changing public APIs

libdd-data-pipeline/src/trace_exporter/mod.rs
Aaalibaba42: I restored the old way for native, kept the new way for wasm.

libdd-capabilities-impl/src/http.rs
```diff
@@ -26,6 +28,32 @@ mod native {
         }
     }
 
+    /// Write `body` as a newline-terminated record to the file referenced by `uri` (which must
+    /// have a `file://` scheme), then return a synthetic 202 response. The same on-disk format
+    /// the pre-capability telemetry worker used when configured with a `file://` endpoint, so
+    /// downstream tests that diff against the recorded payload bytes keep working.
```
VianneyRuhlmann: nit: too specific

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -317,53 +314,66 @@ impl<
...
+            _ = sleeper.sleep(timeout) => Err(TraceExporterError::Shutdown(
+                ShutdownError::TimedOut(timeout),
+            )),
         }
     }
 
+    #[cfg(not(target_arch = "wasm32"))]
     async fn shutdown_workers(self) {
-        #[cfg(not(target_arch = "wasm32"))]
+        let mut join_set = JoinSet::new();
```
VianneyRuhlmann: You could switch to [FuturesUnordered](https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html) to also use it for wasm

libdd-telemetry/src/worker/mod.rs
VianneyRuhlmann: nit: Could it be PhantomData\<C> ?

libdd-capabilities-impl/src/http.rs
```diff
@@ -39,8 +67,15 @@ mod native {
...
         ) -> impl std::future::Future<Output = Result<http::Response<bytes::Bytes>, HttpError>> + MaybeSend
         {
-            let client = self.client.get_or_init(new_default_client).clone();
+            let client_lock = self.client.clone();
             async move {
+                // file:// URIs short-circuit to the on-disk recorder used by tests.
+                if req.uri().scheme_str() == Some("file") {
+                    let (parts, body) = req.into_parts();
+                    return write_to_file_endpoint(&parts.uri, body);
+                }
```
VianneyRuhlmann, line 74: Not blocking since there not using capabilities, but it seems that crashtracker and the reqwest client also bind custom behavior on the `file://` scheme so I wonder if patching the telemetry custom behavior in capabilities is the right solution

VianneyRuhlmann (approved):

libdd-telemetry/src/worker/mod.rs
Aaalibaba42: Then it would inherit constraints of C itself, whilst a function returning a C is freeer

libdd-capabilities-impl/src/http.rs
Aaalibaba42: There is a test in the gitlab pipeline around crashtracking that seems to consistently timeout after thousands of seconds, it's likely a real bug

Aaalibaba42: Not times out, it's OOM killed, it's really weird, it's only aarch64 linux, nothing else...

libdd-capabilities-impl/src/http.rs
```diff
@@ -39,8 +65,15 @@ mod native {
             req: http::Request<bytes::Bytes>,
         ) -> impl std::future::Future<Output = Result<http::Response<bytes::Bytes>, HttpError>> + MaybeSend
         {
-            let client = self.client.get_or_init(new_default_client).clone();
+            let client_lock = self.client.clone();
```
yannham, line 68: I think this should not be cloned, but just shared as a reference instead. This probably duplicate the `OnceLock` and thus always create a client from scratch

libdd-crashtracker/src/crash_info/telemetry.rs
```diff
@@ -355,9 +357,9 @@ impl TelemetryCrashUploader {
         self.send_telemetry_payload(&payload).await
     }
 
-    /// Helper to perform actual HTTP (or file) submission via configured telemetry client
+    /// Helper to perform actual HTTP submission via the native HTTP capability.
     async fn send_telemetry_payload(&self, payload: &data::Telemetry<'_>) -> anyhow::Result<()> {
-        let client = libdd_telemetry::worker::http_client::from_config(&self.cfg);
+        let client = NativeCapabilities::new_client();
```
yannham, line 362: Do we properly propagate configuration from `self.cfg` to the HTTP client now? Not sure what can be there.

libdd-crashtracker/src/crash_info/telemetry.rs
```diff
@@ -372,19 +374,20 @@ impl TelemetryCrashUploader {
...
+                &payload,
+            )?))?;
+
+        let timeout = core::time::Duration::from_millis({
+            if let Some(endp) = self.cfg.endpoint() {
+                endp.timeout_ms
+            } else {
+                Endpoint::DEFAULT_TIMEOUT
+            }
+        });
```
yannham, line 387:
```suggestion
        let timeout = core::time::Duration::from_millis(
            self.cfg
                .endpoint()
                .map(|endp| endp.timeout_ms)
                .unwrap_or(Endpoint::DEFAULT_TIMEOUT));
```

libdd-data-pipeline/src/telemetry/mod.rs
```diff
@@ -138,23 +154,40 @@ impl TelemetryClientBuilder {
             builder.runtime_id = Some(id);
         }
 
-        let (worker_handle, worker) = builder.build_worker(None);
+        // No cancellation runtime handle: telemetry workers driven by SharedRuntime
+        // handle shutdown via WorkerHandle::stop, not the per-handle deadline path.
+        #[cfg(not(target_arch = "wasm32"))]
+        let (worker_handle, worker) = builder.build_worker::<C>(None);
```
yannham, line 160: Isn't the point of capabilities to offer a platform-agnostic implementation? Maybe it's unrelated to this PR, but out of curiosity, why do we need a different argument here for `build_worker`?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -276,7 +273,6 @@ pub struct TraceExporter<
     /// When true, span stats are computed and exported as OTLP metrics. The concentrator is
     /// started at build time, so agent-driven stats (de)activation in `check_agent_info` is
     /// skipped.
-    #[cfg_attr(target_arch = "wasm32", allow(dead_code))]
```
yannham, line 279: Is it ok to remove? It seemed to be WASM-specific.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -317,53 +313,49 @@ impl<
...
-            }
+        let mut futures: FuturesUnordered<_> = handles.into_iter().map(|h| h.stop()).collect();
 
-            while let Some(result) = join_set.join_next().await {
-                if let Ok(Err(e)) = result {
-                    error!("Worker failed to shutdown: {:?}", e);
-                }
+        while let Some(result) = futures.next().await {
+            if let Err(e) = result {
+                error!("Worker failed to shutdown: {:?}", e);
```
yannham, line 356: shouldn't we try to shutdown the other workers still? Though it was the previous behavior, though 🤷‍♂️

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -54,11 +52,12 @@ pub(crate) struct StatsContext<'a, R: SharedRuntime> {
     pub dogstatsd: Option<std::sync::Arc<libdd_dogstatsd_client::Client>>,
     /// Optional telemetry handle forwarded to the [`StatsExporter`].
     #[cfg(feature = "telemetry")]
-    pub telemetry: Option<libdd_telemetry::worker::TelemetryWorkerHandle>,
+    pub telemetry: Option<libdd_telemetry::worker::TelemetryWorkerHandle<C>>,
+    #[cfg(not(feature = "telemetry"))]
+    pub(crate) _phantom: std::marker::PhantomData<fn() -> C>,
```
yannham, line 57: Why is this needed? I think it should deserve a comment at least

libdd-data-pipeline/Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde_json = "1.0.127"
 bytes = "1.11.1"
 sha2 = "0.10"
 either = "1.13.0"
+futures = { version = "0.3", default-features = false, features = ["alloc"] }
```
yannham, line 27: Do we use this for `FuturesUnordered`? Was the previous join not enough/wrong for some reason?

libdd-capabilities-impl/src/http.rs
Aaalibaba42: It's an Arc, so it shouldn't

libdd-data-pipeline/Cargo.toml
Aaalibaba42: It was not usable with wasm because it spawned threads

libdd-data-pipeline/src/telemetry/mod.rs
Aaalibaba42: It takes a tokio runtime as argument... it would have been very heavy to refactor

libdd-crashtracker/src/crash_info/telemetry.rs
Aaalibaba42: I don't think I'd make another PR for styling like that, but it definitely would have been good

libdd-data-pipeline/src/trace_exporter/stats.rs
Aaalibaba42: Because if telemetry is disabled no other thing uses the C associated type iirc, and the `fn() -> C` instead of just `C` to not be bound by things about C (returning a C relaxes more things than containing a C)

PR #2173, opened 2026-06-29
feat(css)!: enabled for wasm

Aaalibaba42:
# What?

Client-side stats span concentrator, stats exporter, and all surrounding state management is now compiled and active on `wasm32` targets.

# Why?

Previously, most of the stats code was gated behind `#[cfg(not(target_arch = "wasm32"))]`, so wasm builds silently skipped stats computation. As we expand wasm support, CSS should work the same as on native.

# How?

The root blocker was `std::time::SystemTime`, which panics on `wasm32-unknown-unknown`. Replacing it with `web_time::SystemTime` (which delegates to `Date.now()` in a browser/WASI context) unblocks compilation. Once that was fixed, the `cfg` guards on stats functions, structs, and module re-exports were no longer needed and were removed.

## Additional Notes

`Duration` is left as `std::time::Duration` — it's purely arithmetic and has no platform dependency. Only `SystemTime`/`UNIX_EPOCH` needed to be swapped.

### DIFF
--- a/libdd-data-pipeline/Cargo.toml
+++ b/libdd-data-pipeline/Cargo.toml
@@ -46,6 +46,8 @@ libdd-tinybytes = { version = "1.1.1", path = "../libdd-tinybytes", features = [
     "bytes_string",
     "serialization",
 ] }
+# Cross-platform `SystemTime` for client-side stats; `std::time` panics on wasm32.
+web-time = "1"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { version = "1.23", features = ["time", "test-util"], default-features = false }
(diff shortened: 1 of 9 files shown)


libdd-data-pipeline/Cargo.toml
```diff
@@ -46,6 +46,8 @@ libdd-tinybytes = { version = "1.1.1", path = "../libdd-tinybytes", features = [
     "bytes_string",
     "serialization",
 ] }
+# Cross-platform `SystemTime` for client-side stats; `std::time` panics on wasm32.
```
VianneyRuhlmann: Although it could be handy to be able to trace every deps to it's usage, I think this kind of comment can become stale very fast, I'd rather have explanation live in code directly or be very generic in toml: `wasm-compatible std::time drop-in replacement`.

VianneyRuhlmann (approved):

PR #2174, opened 2026-06-29
feat(trace-utils)!: add v1 decoder

anais-raison:
# What does this PR do?

Adds a V1 msgpack decoder in `libdd-trace-utils` and wires the V1 variant through the trace pipeline so any V1 payload can be decoded, inspected, and re-encoded.

# Motivation

[APMSP-2813](https://datadoghq.atlassian.net/browse/APMSP-2813) : Prerequisite for the sidecar V1-native path. Splitting out the `libdd-trace-utils` / `libdd-data-pipeline` plumbing so it can land before the sidecar wiring.

# Additional Notes

- Mirrors the existing v04 decoder structure.
- `trace_serializer.rs` now dispatches on `(TraceChunks, OutputFormat)`; the existing v0.4 → V1 cross-encode path is preserved.
- `trace_utils::collect_trace_chunks` renamed to `convert_trace_chunks_v04_to_v05`.

[APMSP-2813]: https://datadoghq.atlassian.net/browse/APMSP-2813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

⚠️ You can focus on those folders/files only as I merged the others from the PR #2145 to get all the refactors: 
- libdd-data-pipeline/src/trace_exporter/trace_serializer.rs
- libdd-trace-utils/src/msgpack_decoder/
- libdd-trace-utils/src/span/mod.rs
- libdd-trace-utils/src/span/vec_map.rs
- libdd-trace-utils/src/trace_utils.rs
- libdd-trace-utils/src/tracer_payload.rs
- libdd-trace-utils/tests/snapshots/compare_v1_full_payload_snapshot_test.json
- libdd-trace-utils/tests/test_send_data.rs


libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,658 @@
...
+            trace_key::TRACER_VERSION => payload.tracer_version = read_interned_string(buf, table)?,
+            trace_key::RUNTIME_ID => payload.runtime_id = read_interned_string(buf, table)?,
+            trace_key::ENV_REF => payload.env = read_interned_string(buf, table)?,
+            trace_key::HOSTNAME_REF => payload.hostname = read_interned_string(buf, table)?,
+            trace_key::APP_VERSION_REF => payload.app_version = read_interned_string(buf, table)?,
+            trace_key::ATTRIBUTES => {
+                payload.attributes = span::read_attributes_map(buf, table)?;
+            }
+            unknown => {
+                return Err(DecodeError::InvalidFormat(format!(
```
ajgajg1134: Don't error out on an unknown payload key, we want to make sure we're forward compatible in case we ever add new keys to this format. If it's an unknown key we should silently drop it

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,658 @@
...
+use crate::msgpack_decoder::decode::error::DecodeError;
+use crate::span::v1::{TraceChunk, TracerPayload, TracerPayloadBytes, TracerPayloadSlice};
+use crate::span::DeserializableTraceData;
+use rmp::decode;
+use std::borrow::Borrow;
+
+// Integer keys used by the V1 wire format. Kept in sync with the encoder side
+// (`msgpack_encoder::v1::{trace_key, chunk_key, SpanKey, SpanLinkKey, SpanEventKey, AnyValueKey}`).
+
+pub(super) mod trace_key {
```
ajgajg1134, line 16: Looks like the ContainerID key is missing here

ajgajg1134 (changes requested):
Another thing we can do here to build confidence in the decoder is take the working dd-trace-go implementation and generate a bunch of various traces and put them just in to flat files, then decode them with this implementation and the trace-agent implementation to make sure they align? (Maybe there's an easier way to check this though)

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
ekump: The encoder (v1 and v04->V1) is also not setting ContainerID in-body.

libdd-trace-utils/src/send_data/mod.rs
```diff
@@ -407,6 +407,24 @@ impl SendData {
...
 
+                futures.push(self.send_payload(
+                    capabilities,
+                    chunks,
+                    payload,
+                    headers,
+                    endpoint.as_ref(),
+                ));
+            }
+            TracerPayloadCollection::V1(payload) => {
```
ekump, line 418: Relevant to this [thread](https://github.com/DataDog/libdatadog/pull/2174/changes#r3494017357), if we're going to set ContainerId in-body, you probably want to strip it out of the headers? It's one of the default ones. Should there be a single set of default headers for all protocols? CC @ajgajg1134

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,658 @@
...
+            chunk_key::TRACE_ID => {
+                let len = decode::read_bin_len(buf.as_mut_slice()).map_err(|_| {
+                    DecodeError::InvalidFormat("V1 chunk trace_id bin len read failure".to_owned())
+                })?;
+                if len != 16 {
+                    return Err(DecodeError::InvalidFormat(format!(
+                        "V1 chunk trace_id must be 16 bytes, got {len}"
+                    )));
+                }
+                let bytes = buf.try_slice_and_advance(16).ok_or_else(|| {
```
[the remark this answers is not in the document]
ekump: 👍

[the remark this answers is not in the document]
ekump: @anais-raison I don't think this applies only to trace-id, it's anything that calls `try_slice_and_advance` (including the v04 decoder). Could you just add a guard in that function?

```Rust      
if bytes > buf.len() {                                                                                                                                                                                     
   return None;                                                                                                                                                                                      
}
```

libdd-trace-utils/src/send_data/mod.rs
ajgajg1134: I think we should actually keep all the normal standard headers here, they're useful in cases where the payload fails to decode for any variety of reasons so that we can tag the metrics with stuff from the HTTP headers which are more likely to make it across successfully

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+                    self.seen.len()
+                ))
+            })
+    }
+
+    /// Records a freshly-read inline string and returns it (cloned for reuse).
+    fn record(&mut self, s: T::Text) -> T::Text {
+        self.seen.push(s.clone());
+        s
+    }
```
yannham: If we clone it, couldn't we take it by reference instead and not return it?

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+/// - `str`/`fixstr` → read and intern, return the value
+/// - any unsigned int marker → resolve the table reference
+pub(super) fn read_interned_string<T: DeserializableTraceData>(
+    buf: &mut Buffer<T>,
+    table: &mut StringTable<T>,
+) -> Result<T::Text, DecodeError>
+where
+    T::Text: Clone,
+{
+    let slice: &[u8] = buf.as_mut_slice();
```
yannham:
```suggestion
    let slice: &[u8] = buf.as_slice();
```

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+    } else if is_uint {
+        let id: u64 = decode::read_int(buf.as_mut_slice()).map_err(|_| {
+            DecodeError::InvalidFormat("V1 interned string reference uint read failure".to_owned())
+        })?;
+        table.resolve(id)
+    } else {
+        Err(DecodeError::InvalidFormat(format!(
+            "Unexpected msgpack marker 0x{marker_byte:02x} for V1 interned string"
+        )))
+    }
```
yannham: Why not have a proper single match expression instead of storing `matches!` results in boolean and then having an if-then-else?

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+
+    for _ in 0..map_len {
+        let key = decode::read_int::<u8, _>(buf.as_mut_slice())
+            .map_err(|_| DecodeError::InvalidFormat("V1 chunk key (u8) read failure".to_owned()))?;
+        match key {
+            chunk_key::TRACE_ID => {
+                let len = decode::read_bin_len(buf.as_mut_slice()).map_err(|_| {
+                    DecodeError::InvalidFormat("V1 chunk trace_id bin len read failure".to_owned())
+                })?;
+                if len != 16 {
```
yannham: should `16` be in a constant somewhere? I think it's repeated below

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
```diff
@@ -0,0 +1,339 @@
...
+            span_key::ENV => span.env = read_interned_string(buf, table)?,
+            span_key::VERSION => span.version = read_interned_string(buf, table)?,
+            span_key::COMPONENT => span.component = read_interned_string(buf, table)?,
+            span_key::KIND => {
+                let kind: u32 = decode::read_int(buf.as_mut_slice()).map_err(|_| {
+                    DecodeError::InvalidFormat("V1 span_kind u32 read failure".to_owned())
+                })?;
+                // OTEL spec: unset / unknown → Internal.
+                span.span_kind = match kind {
+                    2 => SpanKind::Server,
```
yannham: Very nit, but should the values on the left be constants as well? Like `OTEL_SPAN_KIND_SERVER` or something

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
```diff
@@ -0,0 +1,339 @@
...
+        DecodeError::InvalidFormat("V1 attributes flat array len read failure".to_owned())
+    })?;
+
+    if flat_len % FLAT_ATTR_STRIDE != 0 {
+        return Err(DecodeError::InvalidFormat(format!(
+            "V1 attributes flat array length {flat_len} is not a multiple of {FLAT_ATTR_STRIDE}"
+        )));
+    }
+
+    let entries = (flat_len / FLAT_ATTR_STRIDE) as usize;
```
yannham, line 131: I don't have my LSP at hand, but is this conversion guaranteed to never over/underflow (eg on 32 bits)?

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
```diff
@@ -0,0 +1,339 @@
...
+    let mut map = VecMap::with_capacity(entries);
+
+    for _ in 0..entries {
+        let key = read_interned_string(buf, table)?;
+        let value = read_typed_attribute_value(buf, table)?;
+        map.insert(key, value);
+    }
+
+    // V1 attributes are a flat array of triplets, not a real msgpack map, so duplicate keys are
+    // possible; leave `deduped` false so re-encoding applies the defensive last-write-wins dedup.
```
yannham, line 141: I wonder if we could somehow avoid that in the future (that is not for this PR necessarily). When we read msgpack-encoded data here, where do they come from? If they come from tracers, I guess we could require them to send deduped maps (after all the agent doesn't guarantee to support maps with duplicates either, so it's not unreasonable). So that we don't have to do an additional pass if in practice everything we receive here is deduped.

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
```diff
@@ -0,0 +1,339 @@
...
+                decode::read_array_len(buf.as_mut_slice()).map_err(|_| {
+                    DecodeError::InvalidFormat("V1 attribute Array len read failure".to_owned())
+                })?;
+            // Array is a flat sequence of [type, value] pairs.
+            if array_len_with_stride % TYPED_VALUE_STRIDE != 0 {
+                return Err(DecodeError::InvalidFormat(format!(
+                    "V1 attribute Array length {array_len_with_stride} is not a multiple of {TYPED_VALUE_STRIDE}"
+                )));
+            }
+            let n = (array_len_with_stride / TYPED_VALUE_STRIDE) as usize;
```
yannham, line 192: (same question about conversion)

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
```diff
@@ -0,0 +1,339 @@
...
+                let slice: &[u8] = bytes.borrow();
+                if slice.len() != 16 {
+                    return Err(DecodeError::InvalidFormat(format!(
+                        "V1 span_link trace_id expected 16 bytes, got {}",
+                        slice.len()
+                    )));
+                }
+                let mut arr = [0u8; 16];
+                arr.copy_from_slice(slice);
+                link.trace_id = arr;
```
yannham: Nit, but can't we copy directly in `trace_id`, instead of using an intermediate `arr`?

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
```diff
@@ -0,0 +1,339 @@
...
+                link.attributes = read_attributes_map(buf, table)?;
+            }
+            span_link_key::TRACE_STATE => {
+                link.tracestate = read_interned_string(buf, table)?;
+            }
+            span_link_key::FLAGS => {
+                let v: u64 = decode::read_int(buf.as_mut_slice()).map_err(|_| {
+                    DecodeError::InvalidFormat("V1 span_link flags read failure".to_owned())
+                })?;
+                link.flags = v as u32;
```
yannham: This could overflow for `v > u32::MAX`. I would suggest using `u32::try_from(v).unwrap_or(whatever value you think makes sense as a default, or even maybe throw an error?)`.

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -2,11 +2,70 @@
...
+const fn msgpack_string_encoding_len(s: &str) -> usize {
+    let length_marker_len = if s.len() < 32 {
+        1
+    } else if s.len() < 256 {
+        2
+    } else if s.len() <= (u16::MAX as usize) {
+        3
+    } else {
+        5
+    };
```
yannham: Matter of taste so feel free to ignore, but I sometime forget about range patterns and I find them cool 😎 

```suggestion
    let u16_max = u16::MAX as usize;
    let length_marker_len = match s.len() {
        0..32 => 1,
        32..256 => 2,
        256..u16_max => 3,
        _ => 5,
    }
```

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
```diff
@@ -2,11 +2,70 @@
...
+    length_marker_len + s.len()
+}
+
+// Compute the encoding of a string to msgpack in a const manner
+const fn msgpack_const_string_encoding<const ENCODING_LEN: usize>(s: &str) -> [u8; ENCODING_LEN] {
```
yannham, line 22: I'm sure I've seen this function and the following macro somewhere else in Paul's PR some time ago. Is this a duplicate? Should we extract it in a reusable place?

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -61,10 +61,9 @@ impl<K, V> Default for VecMap<K, V> {
     }
 }
 
-// This implementation allocates, which isn't expected for equality testing (not allocating would be
-// rather tedious though). It's fine for tests (where `PartialEq` is currently needed), so we
-// cfg-gate it to avoid unintended usage in prod.
-#[cfg(any(test, feature = "test-utils"))]
+// This implementation allocates, which isn't ideal for equality testing (not allocating would be
+// rather tedious though), but is needed both by production code (e.g. comparing V1 tracer metadata
+// before coalescing) and by tests.
```
yannham: I'm not sure where you need to equate vecmaps, but if your keys are `Ord`, maybe a saner implementation of `Eq`/`PartialEq` would be to sort (with a stable sort) first and then compare the underlying vec element by element.

Also, do you need the `Eq` instance or would a `slow_compare` method work as well? I think I'd like to avoid making this instance the default for everyone; I think it can be surprising for an `Eq` instance to allocate, but people wouldn't know anymore, they would just use it.

Yet another possibility would be to have a thin wrapper `pub VecMapSlowCompare(VecMap)` (but with a better name obviously 😅 ) and implement `Eq` on this instead. The only issue is that this doesn't support nested `VecMap`.

If `compare_deduped`/`slow_compare` or whatever you call a single method is sufficient here instead of making the `Eq` instance public, then I would suggest to go that way for this PR. If not (typically if there are VecMaps within VecMaps), then we can think about it for follow-up work.

libdd-trace-utils/src/tracer_payload.rs
```diff
@@ -80,17 +85,41 @@ impl TracerPayloadCollection {
...
     /// ```
-    pub fn append(&mut self, other: &mut Self) {
+    ///
+    /// # Returns
+    ///
+    /// `true` if `other`'s data was merged into `self`, `false` if the append was skipped (e.g.
+    /// diverging V1 tracer metadata). Callers that rely on `other` being fully drained must check
+    /// this return value rather than assuming success.
+    pub fn append(&mut self, other: &mut Self) -> bool {
         match self {
```
yannham: Nit: you could match on (self, other) directly, to avoid the nested ifs in each branch.

libdd-trace-utils/src/tracer_payload.rs
```diff
@@ -2,12 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::span::v05::dict::SharedDict;
-use crate::span::{v04, v05, BytesData, SharedDictBytes, TraceData};
-use crate::trace_utils::collect_trace_chunks;
+use crate::span::{v04, v05, v1, BytesData, SharedDictBytes, TraceData};
+use crate::trace_utils::convert_trace_chunks_v04_to_v05;
 use crate::{msgpack_decoder, trace_utils::cmp_send_data_payloads};
+use anyhow::Ok;
```
Aaalibaba42: I don't think that's a good practice. From what I can tell from the doc it can be helpful to use it for type deduction, but then I'd use it inline, shadowing `std::Result::Ok` feels a bit dirty

libdd-trace-utils/src/tracer_payload.rs
```diff
@@ -228,13 +258,66 @@ pub fn decode_to_trace_chunks(
...
+    .collect();
+
+    if !differing.is_empty() {
+        warn!(
+            "Skipping V1 TracerPayload append: diverging metadata fields {:?}",
+            differing
+        );
+        return false;
+    }
+    true
```
Aaalibaba42: This materializes the Vec even if it's not required. In those cases, it's better to optimize for the happy path:


```suggestion
    // let fields: [
    //     ("container_id", dest.container_id == src.container_id),
    //     ...
    // ];
    if fields.iter().any(|(_, eq)| !eq) {
        warn!(
            "Skipping V1 TracerPayload append: diverging metadata fields {:?}",
            fields
                .iter()
                .filter_map(|(label, eq)| (!eq).then_some(*label))
                .collect::<Vec<_>>()
        );
        return false;
    }
    true
```

There is 2 passes on the logging branch but it's cheaper on the happy path

libdd-trace-utils/src/tracer_payload.rs
```diff
@@ -228,13 +258,66 @@ pub fn decode_to_trace_chunks(
...
+            })?;
+            Ok((TraceChunks::V1(Box::new(data)), size))
+        }
     }
-    .map_err(|e| anyhow::format_err!("Error deserializing trace from request body: {e}"))?;
+}
 
-    Ok((collect_trace_chunks(data, encoding_type)?, size))
+/// Returns `true` iff every tracer-level metadata field (string fields and attributes) of `src`
+/// matches `dest`.
```
yannham: Very nit, but not sure everyone knows this math-y `iff`

```suggestion
/// Returns `true` if and only if every tracer-level metadata field (string fields and attributes) of `src`
/// matches `dest`.
```

libdd-trace-utils/src/tracer_payload.rs
```diff
@@ -228,13 +258,66 @@ pub fn decode_to_trace_chunks(
...
+        (
+            "language_version",
+            dest.language_version == src.language_version,
+        ),
+        ("tracer_version", dest.tracer_version == src.tracer_version),
+        ("runtime_id", dest.runtime_id == src.runtime_id),
+        ("env", dest.env == src.env),
+        ("hostname", dest.hostname == src.hostname),
+        ("app_version", dest.app_version == src.app_version),
+        ("attributes", dest.attributes == src.attributes),
```
yannham: Is this where you need `VecMap::Eq`? I _think_ we can get by with a `dedup_compare` method or something then, without enabling the `Eq` instance for everyone?

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+}
+
+/// Decodes a V1 msgpack payload from a borrowed slice into a [`TracerPayloadSlice`].
+/// The resulting payload borrows from the input buffer (same lifetime).
+pub fn from_slice(data: &[u8]) -> Result<(TracerPayloadSlice<'_>, usize), DecodeError> {
+    from_buffer(&mut Buffer::new(data))
+}
+
+/// Generic over the deserialization mode (owned `BytesData` or borrowed `SliceData`).
+#[allow(clippy::type_complexity)]
```
Aaalibaba42: The types here don't look too complex, it's probably due to some kind of bad practice in the types themselves that I don't see. Or if not it's better to type alias some things instead of clippy allow imho, like `type DecodedPayload<T> = (TracerPayload<T>, usize);` and use it in the return type

yannham (approved):
My knowledge around v1 is limited so it would be good to have other stamps

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+fn decode_chunks<T: DeserializableTraceData>(
+    buf: &mut Buffer<T>,
+    table: &mut StringTable<T>,
+) -> Result<Vec<TraceChunk<T>>, DecodeError>
+where
+    T::Text: Clone,
+{
+    let count = decode::read_array_len(buf.as_mut_slice())
+        .map_err(|_| DecodeError::InvalidFormat("V1 chunks array len read failure".to_owned()))?;
+    let mut chunks = Vec::with_capacity(count as usize);
```
Aaalibaba42: How trusted can we be about the inputs ? If a malformed/malicious payload sets the count to 0xffffffff we'd have problems. Maybe min between count and buf.len().

I'm noting it here but I've seen this pattern a few times elsewhere btw.

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+                }
+                chunk.spans = spans;
+                saw_spans = true;
+            }
+            chunk_key::ORIGIN => chunk.origin = read_interned_string(buf, table)?,
+            chunk_key::PRIORITY => {
+                let v: i64 = decode::read_int(buf.as_mut_slice()).map_err(|_| {
+                    DecodeError::InvalidFormat("V1 chunk priority read failure".to_owned())
+                })?;
+                chunk.priority = Some(v as i32);
```
Aaalibaba42: In these cases, i32::try_from is safer. If you are sure no wrapping can occur this is fine though.

Aaalibaba42: I've also seen this for `sampling_mechanism` and `flags`

Aaalibaba42 (commented):
I wager this is stacked with the encoder one ? I tried to not repeat myself and focus on the decoder part

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
ekump: Non-blocking since this is how v04 works today - I agree with @Aaalibaba42.  You could add a helper to `Buffer`

```rust
impl<T: DeserializableTraceData> Buffer<T> {                                                                                                                                                                                                                                           
    /// Caps a decoded element count at the bytes remaining in the buffer. Each msgpack                                                                                                                                                                                                
    /// element needs ≥1 byte on the wire, so a length prefix can't legitimately exceed                                                                                                                                                                                                
    /// the remaining bytes — this prevents a malicious count (e.g. 0xFFFFFFFF) from                                                                                                                                                                                                   
    /// forcing a huge pre-allocation before any element is read.                                                                                                                                                                                                                      
    pub(super) fn capped_capacity(&self, count: usize) -> usize {                                                                                                                                                                                                                      
        count.min(self.len())                                                                                                                                                                                                                                                          
    }                                                                                                                                                                                                                                                                                  
}
```

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
ekump: @yannham left a [similar comment](https://github.com/DataDog/libdatadog/pull/2174/changes#r3550199205) elsewhere.

The likelihood is low, so I don't consider it blocking. But, we should be consistent with how we handle out-of-range values and return an error (start and duration, for example, return an error).

ekump:
>Another thing we can do here to build confidence in the decoder is take the working dd-trace-go implementation and generate a bunch of various traces and put them just in to flat files, then decode them with this implementation and the trace-agent implementation to make sure they align? (Maybe there's an easier way to check this though)

@ajgajg1134 @anais-raison I think doing this locally as a one-off before merging the PR is a good idea. I would not recommended as part of this PR we have an automated CI workflow that spins up a real agent. If we can accomplish with just comparing files then we should do that.

libdd-trace-utils/src/msgpack_decoder/v1/mod.rs
ekump: I think this annotation is redundant. I don't get a warning when I remove it locally.

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
ekump: >where do they come from? If they come from tracers,

They do. And the expectation is that only dd-trace-php should be sending msgpack V1. Everyone else will be using native spans. 

I think this PR handles it correctly, and we don't need to require the tracers to do anything. The V1 RFC states that dupes are possible and last write wins.

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
ekump: IMO, this is should be elevated from `very nit` to `non-blocking but preferably fixed`.

I think this logic should live in `SpanKind` (which is already `#repr(u32)]`

You can do this in `SpanKind`:

```rust
impl From<u32> for SpanKind {                                                                                                                                                                                                                                                          
   /// OTEL SpanKind wire value → enum; unset/unknown → Internal (per OTEL spec).                                                                                                                                                                                                     
   fn from(kind: u32) -> Self {                                                                                                                                                                                                                                                       
       match kind {                                                                                                                                                                                                                                                                   
           2 => SpanKind::Server,                                                                                                                                                                                                                                                     
           3 => SpanKind::Client,                                                                                                                                                                                                                                                     
           4 => SpanKind::Producer,                                                                                                                                                                                                                                                   
           5 => SpanKind::Consumer,                                                                                                                                                                                                                                                   
           _ => SpanKind::Internal,                                                                                                                                                                                                                                                   
       }                                                                                                                                                                                                                                                                              
   }                                                                                                                                                                                                                                                                                  }
```  

and here you can just do `span.span_kind = SpanKind::from(kind);`

ekump (approved):
Approving as none of my suggestions are blocking. But, I think mine, @Aaalibaba42 and @yannham's suggestions are simple enough that they can be addressed before merging and will improve the decoder.

libdd-trace-utils/src/msgpack_encoder/v04/mod.rs
anais-raison: Didn't found any duplicate of this function, but I did moved it from another file so it's not an addition for sure.

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
anais-raison: `flat_len` comes from `read_array_len` which return a `u32` and `usize` is guaranteed to be at least 32 bits wide on every platform Rust supports, so it's a lossless conversion even on 32 bits.

yannham: I see! Then it's fine, but as a n(o|i)te, I would favor `usize::from` which makes it clear/guarantee that the conversion doesn't overflow.

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
yannham: Ah, if it's in the spec, then :+1:

libdd-trace-utils/src/msgpack_decoder/v1/span.rs
anais-raison: I'll keep that in mind (for `u8` and `u16`) but usize::from(u32) isn't actually implemented in std, so usize::from(flat_len / FLAT_ATTR_STRIDE) woudn't work in that case.

ajgajg1134 (approved):
From a V1 perspective this looks good! (And I saw some new rust things I don't see often so that was cool 😎 )

PR #2179, opened 2026-06-29
refactor(sidecar)!: Avoid a dedicated socket for crashtracker

bwoebi:
This allows us doing an unified authentication of the socket on the sidecar side, before handing off to crashtracker. Usually the parent process would be the one connecting to the crashtracker, here it is not. This will allow custom handling, and one less endpoint to secure.

The strategy chosen is upgrading to the crashtracker once an IPC socket is received. Currently we recreate a new socket, but we might also decide to reuse the existing IPC connection in future.

The changes to the crashtracker crate itself are kept minimal, just adding a new config for a custom socket connector, and using the existing one as default.

There's also a bit of refactoring on the sidecar side, to make it easier to take ownership of the asyncfd.

### DIFF
--- a/datadog-sidecar/src/crashtracker.rs
+++ b/datadog-sidecar/src/crashtracker.rs
@@ -1,18 +1,350 @@
 // Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
-use std::path::PathBuf;
+//! Crashtracker integration over the single sidecar IPC socket.
+//!
+//! Instead of a second dedicated listener, the crash-time collector connects to the sidecar IPC
+//! socket and sends an
+//! [`enter_crashtracker_receiver`](crate::service::SidecarInterface::enter_crashtracker_receiver)
+//! request as its first message — a normal codec-encoded message (see
+//! [`crashtracker_receiver_request_bytes`]), decoded by the regular serve loop with no marker or
+//! peeking — then streams the crash report over the same connection.
+//!
+//! The receiver consumes an `AsyncBufRead` byte stream; since the socket preserves message
+//! boundaries, [`SeqpacketStreamReader`] concatenates the collector's datagrams back into a
+//! contiguous stream (a zero-length datagram is EOF).
 
-use crate::primary_sidecar_identifier;
+use datadog_ipc::AsyncConn;
 
-pub fn crashtracker_unix_socket_path() -> PathBuf {
-    let base_path = format!(
-        concat!("libdatadog.ct.", crate::sidecar_version!(), "@{}.sock"),
-        primary_sidecar_identifier()
-    );
-    #[cfg(target_os = "linux")]
-    let ret = base_path.into();
-    #[cfg(not(target_os = "linux"))]
-    let ret = std::env::temp_dir().join(base_path);
-    ret
+/// Returns the abstract socket name the Linux listener binds.
+#[cfg(target_os = "linux")]
+pub fn crashtracker_ipc_socket_path(
+    master_pid: u32,
+    ipc_mode: crate::config::IpcMode,
+) -> std::path::PathBuf {
+    let liaison = if master_pid != 0 {
+        crate::setup::DefaultLiason::ipc_for_pid(master_pid)
+    } else {
+        crate::setup::liaison_for_ipc_mode(ipc_mode)
+    };
+    liaison.path().to_path_buf()
+}
+
+/// The codec-encoded `enter_crashtracker_receiver` request the collector sends as its first
+/// message. Fixed bytes, safe to send from a crash handler. This function must however be called
+/// once before the crash handler runs, to initialize.
+pub fn crashtracker_receiver_request_bytes() -> &'static [u8] {
+    use crate::service::sidecar_interface::SidecarInterfaceRequest;
+    static BYTES: std::sync::OnceLock<Vec<u8>> = std::sync::OnceLock::new();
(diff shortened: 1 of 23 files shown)


datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -130,22 +130,25 @@ struct ConnectionSidecarHandler {
     /// Used to auto-register metrics in newly-created telemetry clients when a metric point
     /// for a previously registered metric arrives for a new (service, env) combination.
     metric_registrations: Mutex<HashMap<String, MetricContext>>,
+    /// The connection this handler serves.
+    connection: OwnedServerConn,
 }
 
 impl ConnectionSidecarHandler {
-    fn new(server: SidecarServer) -> Self {
+    fn new(server: SidecarServer, connection: OwnedServerConn) -> Option<Self> {
```
gyuheon0h: This always returns `Some(..)`, right? Is the `Option` misleading then?

datadog-sidecar/src/crashtracker.rs
```diff
@@ -1,18 +1,350 @@
...
+
+            loop {
+                let mut guard = match this.conn.poll_read_ready(cx) {
+                    Poll::Ready(Ok(g)) => g,
+                    Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
+                    Poll::Pending => return Poll::Pending,
+                };
+
+                // One recv() returns one datagram; a max-sized buffer avoids truncation.
+                let read_result = guard.try_io(|inner| {
```
gyuheon0h, line 159: Every single poll read allocates and deallocs 4MiB. Is this okay?

datadog-sidecar/src/crashtracker.rs
```diff
@@ -1,18 +1,350 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::os::fd::{FromRawFd, OwnedFd};
+    use tokio::io::unix::AsyncFd;
+    use tokio::io::{AsyncReadExt, BufReader};
+
+    fn dgram_pair() -> (OwnedFd, OwnedFd) {
```
gyuheon0h, line 233: This test use SOCK_DGRAM but production is SOCK_SEQPACKET. Can we make tests consistent with our prod code or is this intentional?

gyuheon0h (approved):

datadog-sidecar/src/service/sidecar_server.rs
bwoebi: You're right, that code went through two iterations :-D

datadog-sidecar/src/crashtracker.rs
bwoebi: Not really a big problem for crashtracker, but yeah, I should move it out of the loop to save some syscalls.

datadog-sidecar/src/crashtracker.rs
bwoebi: It's intentional, yes, or I would have to rewrite the test for macos. In any case, both have the same semantics needed with respect to crashtracker.

libdd-crashtracker/src/shared/configuration/mod.rs
```diff
@@ -42,13 +42,56 @@ pub struct CrashtrackerConfiguration {
...
+        match std::os::unix::net::SocketAddr::from_abstract_name(unix_socket_path) {
+            Ok(addr) => UnixStream::connect_addr(&addr),
+            Err(e) => Err(e),
+        }
+    };
+    #[cfg(not(target_os = "linux"))]
+    let stream = UnixStream::connect(unix_socket_path);
+    match stream {
+        Ok(s) => s.into_raw_fd(),
+        Err(_) => -1,
```
danielsn, line 87: Should we document that this returns `-1` on error?

libdd-crashtracker/src/shared/configuration/mod.rs
```diff
@@ -42,13 +42,56 @@ pub struct CrashtrackerConfiguration {
     signals: Vec<i32>,
     timeout: Duration,
     unix_socket_path: Option<String>,
+    #[serde(skip, default = "default_unix_socket_connector_value")]
+    unix_socket_connector: fn(&str) -> std::os::fd::RawFd,
     use_alt_stack: bool,
 }
 
+impl PartialEq for CrashtrackerConfiguration {
```
danielsn, line 50: Why can't we just `derive` here?

libdd-crashtracker/src/shared/configuration/mod.rs
```diff
@@ -42,13 +42,56 @@ pub struct CrashtrackerConfiguration {
     signals: Vec<i32>,
     timeout: Duration,
     unix_socket_path: Option<String>,
+    #[serde(skip, default = "default_unix_socket_connector_value")]
+    unix_socket_connector: fn(&str) -> std::os::fd::RawFd,
```
danielsn, line 46: This is sent over the socket to the receiver as a json string.  What does it mean to send a function over the wire?

libdd-crashtracker/src/shared/configuration/mod.rs
bwoebi: miri will fail with `fn()` pointers.

libdd-crashtracker/src/shared/configuration/mod.rs
bwoebi: I think you've missed the `#[serde(skip)]`.

PR #2182, opened 2026-06-30
refactor(libdd-remote-config)!: hide Target inner properties so they are not leaked

hoolioh:
# What does this PR do?

Hide internal implementation so types are not leaked to the consumers

# Motivation

Avoid symol/crate duplication on downstream projects.

### DIFF
--- a/libdd-remote-config/src/lib.rs
+++ b/libdd-remote-config/src/lib.rs
@@ -37,9 +37,9 @@ impl Target {
     /// Creates a new `Target`. `tags` and `process_tags` are expected as
     /// already-formatted `"key:value"` strings.
     pub fn new(
-        service: &str,
-        env: &str,
-        app_version: &str,
+        service: String,
+        env: String,
+        app_version: String,
         tags: Vec<String>,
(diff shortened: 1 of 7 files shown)


libdd-remote-config/src/lib.rs
```diff
@@ -22,17 +22,35 @@ mod targets;
...
+    process_tags: Vec<String>,
+}
+
+impl Target {
+    /// Creates a new `Target`. `tags` and `process_tags` are expected as
+    /// already-formatted `"key:value"` strings.
+    pub fn new(
+        service: &str,
+        env: &str,
+        app_version: &str,
```
bwoebi: Why do you take &str, when you will always unconditionally transform to String? Just take String. It just causes callers to transform back and forth.

hoolioh: I did a first iteraton in which taking the references was a bit more flexible but It was a bit messy on the sidecar side so I changed it. I guess letting the caller decides when to clone or pass ownership is ok in this situation.

iunanua (approved):

PR #2183, opened 2026-06-30
feat(stats): emit canonical gRPC status name for OTLP rpc.response.status_code

mabdinur:
## Summary

The OTLP trace-metrics path emits the numeric gRPC status code as a string (e.g. `"0"`, `"14"`) for the `rpc.response.status_code` data-point attribute. The [OTel gRPC semantic conventions](https://opentelemetry.io/docs/specs/semconv/rpc/grpc/) require the canonical status **NAME** (e.g. `OK`, `NOT_FOUND`, `UNAVAILABLE`).

This PR maps the numeric code back to the canonical name at OTLP emission only:

- Adds `grpc_status_code_to_name` in `libdd-trace-stats` (`span_concentrator`), the inverse of the existing `grpc_status_str_to_int_value` name->code normalization, with the canonical list kept next to the forward table as a single source of truth.
- `libdd-data-pipeline` OTLP `build_attributes` now emits the name (omitting the attribute for empty/unmapped codes).

### Why map instead of pass through the raw string?

By the time stats reach the OTLP/agent boundary, `grpc.status.code` has already been normalized to a numeric `u8` in the aggregation key, so the raw name is gone. Mapping the numeric code back to the name reuses that existing normalization and yields the canonical name for every input form (`"OK"`, `"ok"`, `"StatusCode.OK"`, `"0"`, and numeric metrics) without changing the aggregation data model.

### Unchanged

- `/v0.6/stats` agent payload (`encode_grouped_stats` still emits the numeric code).
- The aggregation key (`grpc_status_code: Option<u8>`) and grouping/cardinality behavior.
- The SHM (`datadog-ipc`) path.

## Test plan

- [x] `cargo test -p libdd-trace-stats` (incl. `test_grpc_status_code_to_name`)
- [x] `cargo test -p libdd-data-pipeline --lib otlp::metrics` (incl. `emits_canonical_grpc_status_name_for_rpc_response_status_code`)
- [ ] system-tests `Test_FR06...::test_fr06_7_rpc_status_code` for python once this ships and the python OTLP suite is enabled

## Notes

Draft: depends on coordinating the system-tests `fr06_7` enablement for python (currently skipped via file-level `missing_feature`) and a libdatadog bump in dd-trace-py.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/libdd-trace-stats/src/span_concentrator/aggregation.rs
+++ b/libdd-trace-stats/src/span_concentrator/aggregation.rs
@@ -200,35 +200,6 @@ fn grpc_status_str_to_int_value(v: &str) -> Option<u8> {
     None
 }
 
-/// Canonical gRPC status names indexed by numeric code; the inverse of the name->code
-/// normalization in [`grpc_status_str_to_int_value`].
-/// See <https://github.com/grpc/grpc/blob/master/doc/statuscodes.md>.
-const GRPC_STATUS_NAMES: [&str; 17] = [
-    "OK",
-    "CANCELLED",
-    "UNKNOWN",
-    "INVALID_ARGUMENT",
-    "DEADLINE_EXCEEDED",
-    "NOT_FOUND",
-    "ALREADY_EXISTS",
-    "PERMISSION_DENIED",
-    "RESOURCE_EXHAUSTED",
-    "FAILED_PRECONDITION",
-    "ABORTED",
-    "OUT_OF_RANGE",
-    "UNIMPLEMENTED",
-    "INTERNAL",
-    "UNAVAILABLE",
-    "DATA_LOSS",
-    "UNAUTHENTICATED",
-];
-
-/// Map a numeric gRPC status code string (as carried on
-/// [`pb::ClientGroupedStats::grpc_status_code`]) to its canonical OTel status name. Returns `None`
-/// for empty/unparseable/out-of-range values.
-pub fn grpc_status_code_to_name(code: &str) -> Option<&'static str> {
-    GRPC_STATUS_NAMES.get(code.parse::<usize>().ok()?).copied()
-}
 
 impl<'a> BorrowedAggregationKey<'a> {
     /// Return an AggregationKey matching the given span.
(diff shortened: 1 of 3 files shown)


mabdinur:
## End-to-end validation (dd-trace-py + system-tests)

Validated this change end-to-end through the Python tracer and the system-tests OTLP trace-metrics parametric suite.

**Result: all green**
- `test_fr06_7_rpc_status_code[OK | NOT_FOUND | UNAVAILABLE]` -> 3 passed
- Full `tests/parametric/test_otlp_trace_metrics.py` -> 47 passed

The round-trip works as designed: dd-trace-py sets `grpc.status.code="OK"` -> libdatadog normalizes name->numeric (`0`) in aggregation -> the new `grpc_status_code_to_name` reverse lookup emits the canonical `"OK"` for `rpc.response.status_code` at OTLP emission (`/v0.6/stats` payload and aggregation key unchanged).

### Note on how it was built
`dd-trace-py` (`munir/otlp-trace-metrics-export`) currently pins libdatadog at the `v37.0.0` tag. This branch sits ~5 commits past `v37.0.0` (including breaking `feat(stats)!` changes), and pointing dd-trace-py's native ext directly at it caused transitive dependency drift (a broken `hyper 1.9.0`) unrelated to this change.

To isolate only this commit for validation, it was cherry-picked onto `v37.0.0` (`range-diff` confirms an identical diff to `8ec3bb982`) and dd-trace-py built cleanly against that. No change to the code in this PR is needed -- once this lands on a libdatadog release that dd-trace-py adopts, the behavior validated here applies as-is.

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -200,6 +200,36 @@ fn grpc_status_str_to_int_value(v: &str) -> Option<u8> {
...
+    "UNAUTHENTICATED",
+];
+
+/// Map a numeric gRPC status code string (as carried on
+/// [`pb::ClientGroupedStats::grpc_status_code`]) to its canonical OTel status name. Returns `None`
+/// for empty/unparseable/out-of-range values.
+pub fn grpc_status_code_to_name(code: &str) -> Option<&'static str> {
+    GRPC_STATUS_NAMES.get(code.parse::<usize>().ok()?).copied()
+}
+
```
VianneyRuhlmann: I think you should either update `grpc_status_str_to_int_value` to use`GRPC_STATUS_NAMES` or move `grpc_status_code_name` to data-pipeline/otlp. I have a slight preference for the latter since there's no real reason for `grpc_status_code_name` to live in libdd-trace-stats apart from sharing the list of names which is not likely to change anyway.

VianneyRuhlmann (approved):

PR #2184, opened 2026-07-01
feat(ci): run clippy on wasm target

Eldolfin:
# What does this PR do?

# Motivation

We don't run tests for wasm so we should at least ensure it's lint free to try to catch bugs early

# Additional Notes

# How to test the change?

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -200,11 +200,28 @@ jobs:
           # its chosen toolchain instead of falling back to the workspace MSRV.
           RUSTUP_TOOLCHAIN: ${{ steps.install-toolchain.outputs.version }}
         run: |
+          # Most crates pull in tokio/mio/socket2/getrandom/ring or native mmap
+          # code and cannot target wasm32. Restrict this job to the crates that
+          # are actually wasm-compatible, intersecting with the affected-crates
+          # list when one is available. --all-features is intentionally not
+          # used: it force-enables the native (non-wasm) code paths that these
+          # crates otherwise gate behind default features.
+          WASM_ALLOWLIST="build_common cc_utils datadog-ffe datadog-ffe-test-suite datadog-ipc-macros datadog-sidecar-macros libdd-capabilities libdd-ddsketch libdd-log libdd-otel-thread-ctx libdd-profiling-protobuf libdd-sampling libdd-tinybytes libdd-trace-normalization libdd-trace-protobuf sidecar_mockgen"
           if [[ -z "$CLIPPY_PACKAGES" ]]; then
-            cargo clippy --workspace --all-targets --all-features -- -D warnings
+            WASM_PACKAGES=$(for p in $WASM_ALLOWLIST; do echo "-p $p"; done | xargs)
+          else
+            WASM_PACKAGES=""
+            for pkg in $WASM_ALLOWLIST; do
+              if echo "$CLIPPY_PACKAGES" | grep -qw -- "$pkg"; then
+                WASM_PACKAGES="$WASM_PACKAGES -p $pkg"
+              fi
+            done
+          fi
+          if [[ -z "$WASM_PACKAGES" ]]; then
+            echo "No wasm-compatible crates affected by this change; skipping."
           else
             # shellcheck disable=SC2086
-            cargo clippy $CLIPPY_PACKAGES --all-targets --all-features -- -D warnings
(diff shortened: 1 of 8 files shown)


Aaalibaba42:
I'm not sure we actually want to do this. It's natural to have parameters that become unused or other kinds of lints that trigger on a secondary platform, and handling this would pollute the code a lot more that it would help imho.

For instance if we allow maybe_unused on stuff that isn't used only by wasm, it's error prone for native if every it really becomes unused, and forcing it be used by all platforms can be detrimental or require heavy refactor

Eldolfin:
I still have to clean it up a bit but I'll soon push the lint fixes, it's not that bad actually and some of it gets fixed by your PRs

For me the motivation is that we don't run tests for wasm so we should at least ensure it's lint free to try to catch bugs early.

.github/workflows/lint.yml
```diff
@@ -157,6 +158,67 @@ jobs:
...
+      - name: Run clippy on wasm ${{ matrix.rust_version }}
+        shell: bash
+        env:
+          CLIPPY_PACKAGES: ${{ needs.setup.outputs.clippy-packages }}
+          # Override rust-toolchain.toml so each matrix entry actually runs on
+          # its chosen toolchain instead of falling back to the workspace MSRV.
+          RUSTUP_TOOLCHAIN: ${{ steps.install-toolchain.outputs.version }}
+        run: |
+          # List of crates that are meant to be compilable for wasm.
+          WASM_ALLOWLIST="build_common cc_utils datadog-ffe datadog-ffe-test-suite datadog-ipc-macros datadog-sidecar-macros libdd-capabilities libdd-common libdd-data-pipeline libdd-ddsketch libdd-dogstatsd-client libdd-log libdd-otel-thread-ctx libdd-otel-thread-ctx-ffi libdd-profiling-protobuf libdd-remote-config libdd-sampling libdd-shared-runtime libdd-tinybytes libdd-trace-normalization libdd-trace-obfuscation libdd-trace-protobuf libdd-trace-stats libdd-trace-utils sidecar_mockgen"
```
danielsn, line 204: How will we keep this in sync?

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -24,15 +24,25 @@ use crate::worker::Worker;
...
 use std::{fmt, io};
 
+// Shared, reference-counted registry of live workers. On native targets it must be `Arc` because
+// runtimes and their `WorkerHandle`s are shared across threads. On wasm the runtime is
+// single-threaded and its workers are `!Send`, so `Rc` is both correct and avoids the
+// `arc_with_non_send_sync` lint.
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) type WorkerRegistry = std::sync::Arc<Mutex<Vec<WorkerEntry>>>;
+#[cfg(target_arch = "wasm32")]
+pub(crate) type WorkerRegistry = std::rc::Rc<Mutex<Vec<WorkerEntry>>>;
```
danielsn, line 37: Why do we need the mutex if single-threaded?

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -24,15 +24,25 @@ use crate::worker::Worker;
...
-use std::sync::{Arc, Mutex};
+use std::sync::Mutex;
 use std::{fmt, io};
 
+// Shared, reference-counted registry of live workers. On native targets it must be `Arc` because
+// runtimes and their `WorkerHandle`s are shared across threads. On wasm the runtime is
+// single-threaded and its workers are `!Send`, so `Rc` is both correct and avoids the
+// `arc_with_non_send_sync` lint.
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) type WorkerRegistry = std::sync::Arc<Mutex<Vec<WorkerEntry>>>;
```
danielsn, line 35: Will this pattern recur? Should this be a generic type?

.github/workflows/lint.yml
Eldolfin: Not sure honestly, the other idea I had instead of a fixed list would be to add a metadata field to each relevant crate's `Cargo.toml` to indicate if it should be compiling on wasm but that would add more complexity and I'm not even sure it solve the syncing problem, it just moves it to a per-crate toggle.

I'm open to suggestion if you have a better alternative in mind

PR #2187, opened 2026-07-01
fix: update rustls-webpki to 0.103.13

morrisonlevi:
# What does this PR do?

Updates rustls-webpki to v0.103.13.

# Motivation

Fixes RUSTSEC-2026-0098, RUSTSEC-2026-0099, and RUSTSEC-2026-0104.

# Additional Notes

I did not make any attempt to understand if we'd actually hit these advisories. I'm applying the fix out of good maintainership.

# How to test the change?

Test as normal. This is a Cargo.lock change only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2901,6 +2901,7 @@ dependencies = [
  "rustls",
  "rustls-native-certs",
  "rustls-platform-verifier",
+ "rustls-webpki",
  "serde",
  "static_assertions",
  "tempfile",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,9 +116,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.93"
+version = "1.0.103"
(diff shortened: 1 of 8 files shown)


gyuheon0h (approved):

Cargo.lock
```diff
@@ -4983,9 +4983,9 @@ checksum = "f87165f0995f63a9fbeea62b64d10b4d9d8e78ec6d7d51fb2125fda7bb36788f"
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.10"
+version = "0.103.13"
```
ekump, line 4987: This is a transitive dep. Won't this get overridden the next time someone updates Cargo.lock?

morrisonlevi: Yes, that can happen, but I'm not sure how to "fix" that. Claude made some suggestions but none of them have worked so far.

morrisonlevi: Using an optional dependency seems to do the trick. Not sure how guaranteed it is, though!

PR #2188, opened 2026-07-01
ci(miri): skip slow miri tests

morrisonlevi:
# What does this PR do?

Adds `#[cfg_attr(miri, ignore)]` to 34 SLOW tests.

# Motivation

The run-miri job partitions were taking 20-24 minutes each, dominated by 34 tests nextest flagged as SLOW (many 400-520s). None of these hang or exercise unsafe code Miri exists to check - they're slow because Miri must bytecode-interpret real TLS/HTTP client construction (rustls/ring crypto init, cert-store loading), tokio's executor park/wake machinery, large fixture-driven test loops, or regex-crate compilation/serde overhead on otherwise-trivial inputs.

# Additional Notes

I had claude sonnet 5 do all the analysis and judge the appropriate fixes, and had it leave justifications for why so we can audit that ourselves. They look reasonable to me.

# How to test the change?

Same as usual, should just be faster under miri.

### DIFF
--- a/datadog-sidecar/src/service/ffe_exposures_flusher.rs
+++ b/datadog-sidecar/src/service/ffe_exposures_flusher.rs
@@ -216,6 +216,7 @@ mod tests {
         .await;
     }
 
+    #[cfg_attr(miri, ignore)] // tokio executor park/wake overhead is prohibitively slow under Miri
     #[tokio::test]
     async fn timeout_returns_without_waiting_for_http_response() {
         let ep = Endpoint {
--- a/datadog-sidecar/src/service/ffe_metrics_flusher.rs
+++ b/datadog-sidecar/src/service/ffe_metrics_flusher.rs
@@ -169,6 +169,7 @@ mod tests {
         .await;
     }
 
+    #[cfg_attr(miri, ignore)] // tokio executor park/wake overhead is prohibitively slow under Miri
     #[tokio::test]
     async fn timeout_returns_without_waiting_for_http_response() {
         let ep = Endpoint {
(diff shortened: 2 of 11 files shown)


datadog-sidecar/src/service/ffe_exposures_flusher.rs
```diff
@@ -216,6 +216,7 @@ mod tests {
         .await;
     }
 
+    #[cfg_attr(miri, ignore)] // tokio executor park/wake overhead is prohibitively slow under Miri
```
danielsn, line 219: could we put this behind a flag so we can have a slow nightly check that does validate them?

morrisonlevi: I'm not opposed but I think that should be a different PR as there are already many `cfg_attr(miri, ignore)` attributes aside from these; `git grep "cfg_attr(miri, ignore)"` says 460, this PR added just 34 of them. I'm sure many of those are because of performance as well.

yannham: I believe most of the `cfg_attr(miri, ignore)` are rather for stuff that just doesn't work under Miri (there's a bunch of things Miri obviously doesn't support, such as syscalls, external C functions, etc.). Since this blocks CI, people have to add the flag. If we want to run slow tests at some point, we'll have to differentiate between "doesn't run under miri" and "run but is slow".

yannham (approved):
Looking at this, I wonder if we could have an allowlist approach instead of denylist for tests to run under Miri. Miri is super slow, and is really useful only for testing our own unsafe code, I guess. On the other hand, I can see people just not adding their test to Miri if it's not on by default...

Anyway, this looks reasonable.

PR #2191, opened 2026-07-02
chore(ci): shard and only run perf benchmarks on impacted crates in PRs

ekump:
# What does this PR do?

Speeds up the performance benchmark job that runs in Gitlab by doing 2 things:
1. Only run benchmarks for crates impacted by changes in the PR. Similar to what we do for tests in Github Actions.
2. Parallelize benchmarks by splitting them across two runners. A weighted list of crates is hardcoded in the shell script to attempt to balance load across the two shards. We'll need to figure out a way to keep this updated over time programatically. 

# Motivation

# Additional Notes

The actual performance improvements should be much bigger once this is merged. The candidate benchmarks are filtering to only impacted crates, but the baseline references main and is still running the full suite.  

# How to test the change?

Temporarily modified the branch to modify `libdd-trace-utils` and saw that it ran for only impacted crates on the candidate benchmarks. https://gitlab.ddbuild.io/DataDog/apm-reliability/libdatadog/-/pipelines/122493383

The script that posts the comment to PRs of benchmarks results has been broken prior to this PR. That will need to be fixed separately before we can fully evaluate this PR works as expected.

### DIFF
--- a/.gitlab/impacted-crates.yml
+++ b/.gitlab/impacted-crates.yml
@@ -20,9 +20,13 @@ compute_impacted_crates:
     name: $BASE_CI_IMAGE
   needs: []
   rules:
-    # main runs the full suite, so nothing to compute there; downstream jobs mark the dependency
-    # optional. Same for merge-queue branches (benchmarks is skipped there) and scheduled runs.
+    # Nothing to compute where benchmarks won't consume it: main runs the full suite, and
+    # release/hotfix/merge-queue/scheduled pipelines don't run benchmarks at all.
     - if: '$CI_COMMIT_BRANCH == "main"'
+      when: never
+    - if: '$CI_COMMIT_BRANCH =~ /^(release|hotfix)\//'
+      when: never
+    - if: '$CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_NAME =~ /^(release|hotfix)\//'
       when: never
     - if: '$CI_COMMIT_BRANCH =~ /^mq-working-branch-/'
       when: never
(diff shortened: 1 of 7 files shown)


.gitlab/impacted-crates.yml
```diff
@@ -0,0 +1,56 @@
...
+      when: never
+    - when: on_success
+  variables:
+    # Full history so the three-dot merge-base diff against origin/main resolves.
+    GIT_DEPTH: 0
+  # A failure leaves the outputs absent; downstream jobs then fall back to running everything.
+  allow_failure: true
+  script:
+    - git fetch --no-tags origin main
+    - (cd .github/actions && cargo build --release -p crates-reporter)
```
yannham: It's more of a nitpick / future work, but would there be a way to distribute a binary of crates-reporter instead? I suppose it doesn't change often. But sometimes getting a binary in a job is just so annoying that maybe building from source is simpler 🤷

benchmark/run_benchmarks_ci.sh
```diff
@@ -20,9 +20,41 @@ OUTPUT_DIR="${1:-}"
...
+    done
+  done
+  feature_args=()
+  if (( ${#features[@]} > 0 )); then
+    feature_args=(--features "$(IFS=,; echo "${features[*]}")")
+  fi
+  message "Benchmarking selected crates: ${BENCH_PACKAGES}"
+  cargo bench "${package_args[@]}" "${feature_args[@]}" -- --warm-up-time 1 --measurement-time 5 --sample-size=200
+else
+  cargo bench --workspace --features libdd-crashtracker/benchmarking,libdd-sampling/v04_span,libdd-sampling/bench-internals,libdd-trace-utils/bench-internals -- --warm-up-time 1 --measurement-time 5 --sample-size=200
```
yannham: Nit: this is probably simpler with this one liner, but I wonder if we shouldn't generate a `BENCH_PACKAGES` unconditionally (just putting all the crates that are known to bench when it's empty) and then use a single code path for the `cargo bench` command and package features. Otherwise there are two places where we define which features a specific crate needs for benchmarking (this line and in `bench_features_for_crate`), and they could disagree/drift.

yannham (approved):
So we get parallel _and_ "differential" benches, nice 👍

.gitlab/impacted-crates.yml
hoolioh: We experimented with that approach by having the binary in [releases](https://github.com/DataDog/libdatadog/releases?page=2#release-ci-crates-v0.1.0) (we need to remove it) but we decided that at some point it will be better to move all the CI code to a new repo, at least all the code that can be shared across repositories, and having the releases there.

hoolioh (approved):

.gitlab/impacted-crates.yml
ekump: Yes, we should pre-build the binaries. It's something we'll look into for the larger CI improvement project. At a minimum, `crates-reporter` should probably not live in `.github`, but we can address moving it in a separate PR.

benchmark/run_benchmarks_ci.sh
ekump: ah yes, good call.

PR #2193, opened 2026-07-03
fix(shared-runtime): wrap run() in select! so cancellation interrupts in-flight operations

r1viollet:
## What

`PausableWorker` wraps each `Worker` in a task loop. When `pause()` is called (e.g. from `before_fork()`), it cancels the stop token and awaits the task handle.

Before this change, the task loop checked the token **between** `trigger()` and `run()`, but not **during** `run()`:

```rust
loop {
    select! {
        biased;
        _ = cloned_token.cancelled() => { break; }  // ✓ checked here
        _ = worker.trigger() => {
            worker.run().await;  // ✗ token ignored here — run() blocks until done
        }
    }
}
```

If `run()` was executing when the token was cancelled, `pause()` had to wait for `run()` to return on its own.

## Contract

The `Worker` trait documents the expectation:

```rust
/// Code in this function must always use timeout on long-running await calls to avoid
/// blocking forks if an await call takes too long to complete.
async fn run(&mut self);
```

This put the burden on each implementation. This PR enforces it at the framework level instead: `run()` is now also wrapped in `select!` against the cancellation token. Implementations no longer need internal timeouts for fork safety — cancellation already happens at the right time and now propagates into `run()`.

## Concrete example

`AgentInfoFetcher::run()` makes an HTTP GET to `/info`. Without the fix, if a fork fires while the request is in flight:

```
before_fork()
  → pause() → stop_token.cancel()
  → handle.await
      task loop: select! → token cancelled? yes, but we're inside run().await
      AgentInfoFetcher::run() → HTTP GET /info → agent slow → blocks 2s
  → pause() returns after 2s
os.fork() returns after 2s
```

With the fix:

```
before_fork()
  → pause() → stop_token.cancel()
  → handle.await
      task loop: select! on run() → token already cancelled
      → exits at next .await inside HTTP client (~microseconds)
  → pause() returns in <1ms
os.fork() returns in <1ms
```

## Change

Two `select!` blocks added to the task loop in `PausableWorker::start()`, one for the initial run and one in the regular loop. No API changes.

## Testing

New test `test_pause_interrupts_long_running_run`: worker whose `run()` sleeps 10s. `pause()` must return in < 1s. Fails without the fix.

End-to-end in dd-trace-py against a fake agent that delays `/info` by 2s:

```
before_fork() without fix:   3005ms
before_fork() with fix:         1ms
```

### DIFF
--- a/libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
+++ b/libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
@@ -122,9 +122,16 @@ impl<T: Worker + MaybeSend + Sync + 'static> PausableWorker<T> {
                         _ = cloned_token.cancelled() => {
                             return worker;
                         }
-                        _ = worker.initial_trigger() => {
-                            worker.run().await;
-                        }
+                        _ = worker.initial_trigger() => {}
+                    }
+
+                    // First run — also cancellable so that a fork signal received
+                    // while run() is executing (e.g. mid-HTTP-request) exits
+                    // immediately rather than waiting for the request to complete.
+                    select! {
+                        biased;
+                        _ = cloned_token.cancelled() => { return worker; }
+                        _ = worker.run() => {}
                     }
 
                     // Regular iterations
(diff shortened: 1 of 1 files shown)


VianneyRuhlmann:
We don't want this `run` is not meant to be cancelled in the general case since it may cause us to lose data (e.g. telemetry flush being interrupted in-flight) `trigger` should be the cancellable function. What I mentioned in the incident channel was a possible improvement for some worker that are not at risks of losing data (e.g. agent info fetcher). Furthermore this can only be done for the shutdown as cancelling `run` can also corrupt the worker state preventing us from restarting it after the fork

r1viollet:
> We don't want this `run` is not meant to be cancelled in the general case since it may cause us to lose data (e.g. telemetry flush being interrupted in-flight) `trigger` should be the cancellable function. What I mentioned in the incident channel was a possible improvement for some worker that are not at risks of losing data (e.g. agent info fetcher). Furthermore this can only be done for the shutdown as cancelling `run` can also corrupt the worker state preventing us from restarting it after the fork

@VianneyRuhlmann true, closing this for now. I assume you will be putting up something more reasonable.

PR #2197, opened 2026-07-06
refactor(otel-thread-ctx): use inline assembly instead of C shim

yannham:
# What does this PR do?

This revives #2129, which has been merged in a different branch than main. Uses inline assembly instead of a C shim to enable TLSDESC access from Rust.

# Motivation

This get rids of all toolchain-related issues, at the cost of some inline ASM. See the original PR for more discussion.


libdd-otel-thread-ctx-ffi/tests/elf_properties.rs
```diff
@@ -6,29 +6,486 @@
...
+}
+
+fn compile_tls_shim_object(dir: &Path) -> PathBuf {
+    let source = Path::new(env!("CARGO_MANIFEST_DIR")).join("tests/tls_shim.c");
+    let object = dir.join("tls_shim.o");
+    let mut compile_object = Command::new("cc");
+    compile_object.args(["-O2", "-fPIC", "-fomit-frame-pointer", "-c"]);
+
+    #[cfg(target_arch = "x86_64")]
+    compile_object.arg("-mtls-dialect=gnu2");
```
yannham: I fear this brings back some toolchain constraints (basically that `-mtls-dialect=gnu2` is available), just moving them from the main crate to tests. I wonder if we should just eschew the test if the compiler doesn't support the GNU2 dialect? After all, what we need is that somewhere in CI there's _at least one_ amd64 target where this tests is run (which is currently the case).

libdd-otel-thread-ctx/src/lib.rs
```diff
@@ -64,23 +64,100 @@
...
+        // WARNING: do not change the assembly below. See the warning above for amd64.
+        core::arch::asm!(
+            "mrs   x1, tpidr_el0",
+            "adrp  x0, :tlsdesc:otel_thread_ctx_v1",
+            "ldr   x2, [x0, :tlsdesc_lo12:otel_thread_ctx_v1]",
+            "add   x0, x0, :tlsdesc_lo12:otel_thread_ctx_v1",
+            ".tlsdesccall otel_thread_ctx_v1",
+            // x1 is guaranteed not to be clobbered by the call
+            "blr   x2",
+            "add   x0, x1, x0",
```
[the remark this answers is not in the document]
yannham: Indeed, and what's more [ARM ABI](https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst#general-dynamic) actually mandates the exact uses of register as well, and we should use `x1` instead of `x2`. Swapping the two in dbb0eebbd.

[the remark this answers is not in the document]
yannham: Though it's surprising that the check for the exact sequence of instructions succeeded in the CI before :thinking:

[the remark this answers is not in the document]
yannham: Ok, so after investigation, the previous sequence was the exact one generated by a relatively oldish GCC version (the one we have on CI basically on Alpine and CentOS). The new one is the clang version, exactly. Some GCC versions also move the TP offset calculation around a bit.

For now I've rooted for the clang version, because Rust uses LLVM under the hood, it clobbers one less register, and respect the ABI described above byte per byte. However `cc` is often `gcc` in the test, so I've extended the test to make it a bit more lax and accommodate for different compilers.

But it starts to be _a lot_ of code, just for testing something "static" (and just because it's annoying to ensure clang in the CI). I start to question the usefulness of this check: somehow it only has to be tested "once".

pawelchcki (approved):
Drive By review

Looks good. I looked for some alternatives and this seems like a solid option

libdd-otel-thread-ctx/src/lib.rs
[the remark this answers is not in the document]
ivoanjo: TBH I wouldn't go too much into "try to support all compilers in CI"; I would instead suggest going the "gold/known version" approach, e.g. something like:

* In CI we only assert that "assembly -> matches some known hash". If you change the assembly, you need to update the "some known hash"

* And then we have some script/test that can recalculate the hash; and next to it is a comment saying "To get the latest assembly hash, we ran this on some standard docker image (ubuntu 26.06 with clang XX)"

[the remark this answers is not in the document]
yannham: I like that idea, implemented in latest version. It doesn't remove so much code (some of it has just moved to the gen_tls_shim_hash executable), but the annoying part (compiling the C shim with the exact right toolchain) has been offloaded to this run-once-for-a-long-time-outside-of-tests executable to generate hash, which is indeed win for tests.

libdd-otel-thread-ctx/src/test_utils/tls_shim_window.rs
```diff
@@ -0,0 +1,358 @@
...
+//!
+//! Both consumers must agree byte-for-byte, so the extraction lives here once and is shared.
+//!
+//! The target architecture is read from the ELF header at runtime, so the generator can process a
+//! cross-compiled reference object for either architecture from any host.
+//!
+//! The "window" is the full TLSDESC access sequence, in program order:
+//! - **x86-64**: `lea …@tlsdesc(%rip), %rax` / `call *…@tlscall(%rax)` / `add %fs:0x0, %rax`.
+//! - **aarch64**: `adrp`/`ldr`/`add`/`blr` (the relocation-bearing core) followed by the
+//!   thread-pointer computation `mrs x8, tpidr_el0` / `add x0, x8, x0`.
```
ivoanjo: Minor: I suggest simplifying this comment a bit to not inline the actual assembly, as it can get out of sync very fast (I suspect it already is on x86?)

libdd-otel-thread-ctx/src/test_utils/tls_shim_window.rs
```diff
@@ -0,0 +1,358 @@
...
+//! ELF object.
+//!
+//! This module has two consumers in `libdd-otel-thread-ctx-ffi`:
+//! - the `tlsdesc_inline_sequence` integration test, which extracts the sequence from our own
+//!   statically linked artifact (produced by the inline assembly in `libdd_otel_thread_ctx`) and
+//!   hashes it
+//! - the `gen_tls_shim_hash` binary, which extracts the sequence from a Clang-compiled reference
+//!   object and hashes it to produce the "golden" hash
+//!
+//! Both consumers must agree byte-for-byte, so the extraction lives here once and is shared.
```
ivoanjo: Minor: I'm a bit confused by this last claim -- why can't two different outputs of `libdd-otel-thread-ctx-ffi` share code that lives in `libdd-otel-thread-ctx-ffi` again?

libdd-otel-thread-ctx/src/test_utils/tls_shim_window.rs
```diff
@@ -0,0 +1,358 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared logic for locating and hashing the TLSDESC access sequence of `otel_thread_ctx_v1` in an
+//! ELF object.
```
ivoanjo: Minor: I think this would benefit from a bit more context, since it's kinda something that you don't to every day ;)

libdd-otel-thread-ctx/src/test_utils/tls_shim_window.rs
```diff
@@ -0,0 +1,358 @@
...
+        let mut out = String::with_capacity(digest.len() * 2);
+        for byte in digest {
+            out.push_str(&format!("{byte:02x}"));
+        }
+        out
+    }
+
+    /// Render the sequence bytes as hex, one 4-byte little-endian word per group on aarch64
+    /// (matching the fixed-width instruction encoding) or a single hex string on x86-64. Used in
+    /// failure messages so a human can eyeball which bytes differ.
```
ivoanjo: Minor: A bit too much info? I think it's kinda obvious from the code

```suggestion
```

libdd-otel-thread-ctx/src/test_utils/tls_shim_window.rs
```diff
@@ -0,0 +1,358 @@
...
+    /// Human-readable origin (file, archive member, …), used in messages.
+    pub label: String,
+    /// The detected architecture of the object the sequence came from.
+    pub arch: Arch,
+    /// The full sequence bytes, sliced out with the architecture-specific bounds.
+    pub bytes: Vec<u8>,
+}
+
+impl TlsDescWindow {
+    /// Lowercase hex SHA-256 of the sequence bytes.
```
ivoanjo: Minor: A bit too much info? I think it's kinda obvious from the code

```suggestion
```

libdd-otel-thread-ctx/src/lib.rs
```diff
@@ -64,23 +66,112 @@
...
+    #[cfg(target_arch = "x86_64")]
+    #[inline(always)]
+    unsafe fn tls_slot() -> *mut *mut ThreadContextRecord {
+        let ptr: usize;
+        // WARNING: keep the assembly below in the canonical compiler-emitted TLSDESC form. Linkers
+        // rely on these exact relocation-bearing instruction patterns for TLS relaxation,
+        // especially when this crate is linked statically. Harmless-looking rewrites can hide part
+        // of the sequence from the linker and produce a partially relaxed access that computes an
+        // invalid TLS address.
+        core::arch::asm!(
```
ivoanjo, line 139: Minor: Should this also include the "// This code match byte-per-byte what clang generates, and this is verified during tests." note we have below for arm64?

libdd-otel-thread-ctx/src/bin/gen_tls_shim_bytes.rs
ivoanjo, line 1: Minor: It's kinda weird to see this living in `libdd-otel-thread-ctx-ffi`, since it conceptually kinda belongs to `libdd-otel-thread-ctx` -- if we change the assembly there, we must always update this test here.

libdd-otel-thread-ctx/tests/tlsdesc_inline_sequence.rs
```diff
@@ -0,0 +1,94 @@
...
+/// Golden SHA-256 of the full x86-64 TLSDESC sequence (`lea …@tlsdesc` / `call …@tlscall` /
+/// `add %fs:0x0, %rax`), generated by the `gen_tls_shim_hash` binary. See [`regenerate_hint`].
+const EXPECTED_TLSDESC_HASH_X86_64: &str =
+    "5b38b6b2a151997785463a5d9d2b1f1cb9027660ea37f2aafe94867508cedc66";
+
+/// Golden SHA-256 of the full aarch64 TLSDESC sequence (`adrp`/`ldr`/`add`/`blr` followed by the
+/// thread-pointer computation `mrs x8, tpidr_el0` / `add x0, x8, x0`), generated by the
+/// `gen_tls_shim_hash` binary. See [`regenerate_hint`].
+const EXPECTED_TLSDESC_HASH_AARCH64: &str =
+    "f8adc6415409a726b51332a2c36931de9d3537dc33705c4a4aa160268e40ca97";
```
ivoanjo: Minor: Same as my note in `tls_shim_window.rs` -- I think writing out the actual assembly here as a comment is both redundant AND can get out of sync very easily.

ivoanjo: Minor: It occurred to me that we don't even need to particularly hash the sequence, we could just turn the sequence to hex bytes directly (funnily enough, I suspect we'll get _less_ bytes than with sha256)

ivoanjo (approved):
👍 Looks great, no concerns (only my usual smattering of small suggestions), and thanks @cataphract for all the work in the previous https://github.com/DataDog/libdatadog/pull/2129 .

libdd-otel-thread-ctx/src/test_utils/tls_shim_window.rs
yannham: Is it out of date? I expect those sequences to be very stable, but yeah, I hesitated myself to clean those AI-generated comments as well :+1: will try to simplify them

ivoanjo: The current assembly listed in the file uses leaq and addq instead of lea and add, although it's possible it's just a different encoding for the same instructions (e.g. just a way to encode flags or whatever), yet "why leave everyone wondering" is my thinking ;)

yannham: Ah, yeah, it' just an abuse of language to drop the final `q`. I believe the q was added from x32 to x64 to mean "operating on quadruple word / 8 bytes" and `leaq` is the only instruction that exists on x64. But people kept calling them `lea` (and respectively `mov`) out of habit. Anyway, point taken about it being confusing :stuck_out_tongue:

libdd-otel-thread-ctx/src/bin/gen_tls_shim_bytes.rs
yannham: Good point. I _think_ what happened is that I started with a test operating in the shared library (hence living in the FFI), and thus colocated the hash refresher with the test. But then Claude convinced me that it was sufficient and technically simpler to do it on a static build instead. Now that it is the case, I kinda agree that there's no obvious reason for both the test and the golden hash generator to live in the FFI instead of the original crate...

yannham: Moved to `otel-thread-ctx`; works as well with minimal adaptation :+1:

PR #2202, opened 2026-07-07
feat(profiling): add extensible profile type API for prototyping

r1viollet:
# What does this PR do?

Adds five configurable custom profile type slots for profile types that are not yet stable enough to add as dedicated `SampleType` variants.

- adds `SampleType::Custom1` through `SampleType::Custom5`
- keeps the normal profile constructors as the primary API
- adds setters to configure a custom slot with its concrete `(type, unit)` pair before serialization
  - Rust: `Profile::set_custom_sample_type`
  - C: `ddog_prof_Profile_set_custom_sample_type`
  - CXX: `Profile::set_custom_sample_type`
- rejects serialization if a used custom slot is not configured
- updates the replayer to map unknown pprof `(type, unit)` pairs onto `Custom1..Custom5`
- removes the previous fixed `Experimental*` variants
- adds a C example for `memory-breakdown` / `bytes`

# Motivation

#1450 moved the C API toward `SampleType`, which is still the right API for stable, agreed-upon profile types.

This PR adds @danielsn's slot-based prototyping path: profilers can use a bounded set of custom slots, configure their names/units at profile construction time, and avoid waiting for a libdatadog release while evaluating a new type such as `memory-breakdown`. Once the type is stable, it should be promoted to a dedicated `SampleType` variant and callers should migrate back to the fully typed API.

# Additional Notes

The old `ExperimentalCount`, `ExperimentalNanoseconds`, and `ExperimentalBytes` variants were fixed to names like `experimental-bytes`, so they could not express a meaningful type name and only provided three unit-specific slots. They are replaced by `Custom1..Custom5`.

For `memory-breakdown`, the sample type is configured as `("memory-breakdown", "bytes")`. Anonymous/file/JIT/etc. breakdowns are represented by separate stacks or labels within that type, not by separate sample types.

The period remains profile-level sampling metadata, not per-sample-type metadata. The custom example passes no period because `memory-breakdown` is a point-in-time byte measurement with no meaningful sampling cadence.

BREAKING CHANGE: `ExperimentalCount`, `ExperimentalNanoseconds`, and `ExperimentalBytes` are removed from `SampleType`. Use `Custom1..Custom5` and configure the selected slot with `set_custom_sample_type`.

# How to test the change?

Local macOS:

```bash
cargo +nightly-2026-02-08 fmt --all -- --check
cargo check -p libdd-profiling --features cxx
cargo test -p libdd-profiling -p libdd-profiling-ffi -p datadog-profiling-replayer --no-fail-fast
cargo +stable clippy --workspace --all-targets --all-features -- -D warnings
```

Linux workspace:

```bash
cargo check -p libdd-profiling --features cxx
cargo test -p libdd-profiling -p libdd-profiling-ffi -p datadog-profiling-replayer --no-fail-fast
cargo run --bin release --features profiling --release -- --out release
cd examples/ffi
cmake -S . -B build -DDatadog_ROOT=$HOME/go/src/github.com/DataDog/libdatadog-2/release
cmake --build build --target custom_profile_types -j2
LD_LIBRARY_PATH=$HOME/go/src/github.com/DataDog/libdatadog-2/release/lib ./build/custom_profile_types
cd ../..
cargo ffi-test --skip-build --filter custom_profile_types
```


libdd-profiling-ffi/src/profiles/datatypes.rs
```diff
@@ -105,6 +105,31 @@ pub use api::SampleType;
...
+pub struct CustomValueType<'a> {
+    pub type_str: CharSlice<'a>,
+    pub unit: CharSlice<'a>,
+}
+
+/// Period expressed as a raw (type, unit) string pair for use with
+/// [`ddog_prof_Profile_new_custom`].
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct CustomPeriod<'a> {
```
r1viollet: not sure we need this, I'll have to check what periods mean in OTel

libdd-profiling/src/api/sample_type.rs
```diff
@@ -74,11 +83,6 @@ pub enum SampleType {
     WallTime,
     /// Legacy: Use `WallTime` instead for consistency with naming scheme
     WallLegacy,
-
-    // Experimental sample types for testing and development.
-    ExperimentalCount,
```
r1viollet, line 79: I was going to remove these, though this is an API breaking change, so I can move this to a separate PR.

libdd-profiling-ffi/src/profiles/datatypes.rs
r1viollet: So after checking
- pprof only has one period, so this does not really need to exist
- otel has one period per sample type (as this is split per sample type), so if you want to profile every X system call, you might want to specify it.

datadog-profiling-replayer/src/replayer.rs
```diff
@@ -55,29 +55,48 @@ impl<'pprof> Replayer<'pprof> {
...
+    /// are promoted to `'static` via [`Box::leak`].  The replayer is a
+    /// short-lived tool; the bounded number of leaked allocations is acceptable.
+    fn to_static_value_type(type_str: &str, unit: &str) -> api::ValueType<'static> {
+        let borrowed = api::ValueType::new(type_str, unit);
+        // Fast path: resolve via the stable enum (returns 'static string literals).
+        if let Ok(st) = api::SampleType::try_from(borrowed) {
+            return api::ValueType::from(st);
+        }
+        // Slow path: custom type not yet in the enum – own the strings.
+        let ty: &'static str = Box::leak(type_str.to_string().into_boxed_str());
```
r1viollet: This is OK unless we continuously create new profile types, which should not happen ?

libdd-profiling/src/api/sample_type.rs
```diff
@@ -74,10 +83,14 @@ pub enum SampleType {
...
     WallLegacy,
-
-    // Experimental sample types for testing and development.
+    /// Deprecated: use `Profile::try_new_with_value_types` with a descriptive raw `(type, unit)`
+    /// pair instead. This fixed name cannot distinguish multiple custom count measurements.
     ExperimentalCount,
+    /// Deprecated: use `Profile::try_new_with_value_types` with a descriptive raw `(type, unit)`
+    /// pair instead. This fixed name cannot distinguish multiple custom time measurements.
     ExperimentalNanoseconds,
+    /// Deprecated: use `Profile::try_new_with_value_types` with a descriptive raw `(type, unit)`
```
danielsn: Is anyone currently using this or can we just remove it?

r1viollet:
So Daniel had an idea to avoid the current change in constructor, we can have custom types that we can specify

danielsn (approved):
As we discussed offline, a cleaner solution could be to add `Custom0` .. `CustomN` enums and an API for associating them with a name and value type

libdd-profiling/src/api/sample_type.rs
r1viollet: I wanted to remove these, and then I had semver check issues 🤔 
I don't think they are used, I'd favour removing them

libdd-profiling/src/cxx.rs
```diff
@@ -553,6 +582,17 @@ impl Profile {
         Ok(())
     }
 
+    pub fn set_custom_sample_type(
+        &mut self,
+        slot: ffi::SampleType,
+        type_: &str,
```
danielsn, line 582: In other parts of the code we spell this `typ` but this works too

danielsn (approved):

PR #2203, opened 2026-07-07
fix!: css obfuscation logic

Eldolfin:
# What does this PR do?
- fix the condition for CSS obfuscation (now needs the obfuscation_versions to be equal)
- ensure each bucket is composed only of obfuscated or un-obfuscated stats
- ensure each stats payload is composed only of obfuscated or un-obfuscated stats buckets

# Motivation
- CSS spec compliance

# Additional Notes

- removing the unused `results.xml` file is an unrelated change but making a PR for it would've been overkill.

# How to test the change?
- [ ] system tests

APMSP-2985


Eldolfin:
@codex review

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -603,11 +594,6 @@ mod tests {
             Endpoint::from_url(stats_url_from_agent_url(&server.url("/")).unwrap()),
             NativeCapabilities::new_client(),
             #[cfg(feature = "stats-obfuscation")]
-            Arc::new(ArcSwap::from_pointee(StatsComputationObfuscationConfig {
-                enabled: true,
-                ..Default::default()
-            })),
-            #[cfg(feature = "stats-obfuscation")]
             "1",
```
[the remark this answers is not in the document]
Eldolfin: fixed in 05d2364

Eldolfin (commented):
(testing claude review)

Overall the core fix looks correct (exact obfuscation-version match, per-bucket obfuscation locking, never mixing obfuscated/unobfuscated buckets in one payload). Main gap I'd flag before merging: no new tests cover the actual new behavior introduced here — a bucket locking its obfuscation state at creation, `drain_due_buckets` splitting a mixed set across two calls, or `StatsExporter::send(force=true)` issuing two separate HTTP requests when buckets are mixed. Given the PR body itself still has "system tests" unchecked, a unit test for the mixed-bucket force-flush path (in `span_concentrator/tests.rs` and `stats_exporter.rs`) would make this much safer to land.

One more note that didn't fit as an inline comment since it's on an unmodified line: `OtlpStatsExporter`'s shutdown path (`libdd-data-pipeline/src/otlp/metrics.rs:305`, `self.send(true, OTLP_SHUTDOWN_MAX_RETRIES)`) only force-flushes once, unlike `StatsExporter::send`'s new double-flush-on-force. That's safe today only because `builder.rs` always constructs that concentrator's `obfuscation_config` as `None` (so `obfuscated` can never be `true` there) — if that concentrator is ever reused for an obfuscation-enabled path, mixed buckets left behind on shutdown would be silently dropped. Left a few smaller inline notes too.

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -190,7 +198,23 @@ impl SpanConcentrator {
...
+                bucket_timestamp,
+                self.max_entries_per_bucket,
+                #[cfg(feature = "stats-obfuscation")]
+                self.obfuscation_config.load().enabled,
+            )
+        });
+        #[cfg(feature = "stats-obfuscation")]
+        let obfuscated_resource = Self::compute_obfuscated_span(
+            target_bucket.obfuscated,
+            self.obfuscation_config.load().sql_obfuscation_mode,
```
Eldolfin: `obfuscated` is locked at bucket creation, but `sql_obfuscation_mode` is still read live on every span insert here. If the SQL obfuscation mode changes mid-bucket-lifetime while `enabled` stays `true`, a single bucket could still aggregate spans obfuscated with different SQL modes — i.e. the bucket isn't fully obfuscation-consistent, just enabled/disabled-consistent. Is this covered by the APMSP-3720 FIXME in `trace_exporter/stats.rs`, or is it a separate gap worth its own ticket/comment here?

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -431,19 +431,29 @@ pub(super) struct StatsBucket {
     max_entries: usize,
     /// Number of spans collapsed into the overflow bucket due to cardinality limiting.
     collapsed_count: u64,
+    #[cfg(feature = "stats-obfuscation")]
```
Eldolfin: Nit: doc comment is sandwiched between the `#[cfg(...)]` attribute and the field. Usually the doc comment goes above the attribute (or docs+attrs grouped with docs first) for readability; not blocking since this isn't part of a public doc surface.

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -245,13 +248,13 @@ impl<Cap: HttpClientCapability + SleepCapability, Con: FlushableConcentrator>
     /// # Panic
     /// Will panic if another thread panicked while holding the concentrator lock in which
     /// case stats cannot be flushed since the concentrator might be corrupted.
-    fn flush(&self, force_flush: bool) -> (pb::ClientStatsPayload, u64) {
+    fn flush(&self, force_flush: bool) -> (pb::ClientStatsPayload, u64, bool) {
         let sequence = self.sequence_id.fetch_add(1, Ordering::Relaxed);
```
Eldolfin, line 229: With `force_flush=true` this now always loops twice via `send`, even when buckets aren't actually mixed — the second pass calls `flush` (and burns a sequence number via `fetch_add` here) just to find an empty payload and `continue`. Probably harmless if the agent only requires monotonic (not contiguous) sequence numbers, but worth double-checking that assumption holds agent-side.

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -107,7 +107,7 @@ fn is_obfuscation_active(agent_info: &AgentInfo) -> bool {
     agent_info
         .info
         .obfuscation_version
-        .is_some_and(|v| v >= 1 && v <= SUPPORTED_OBFUSCATION_VERSION)
+        .is_some_and(|v| v >= 1 && v == SUPPORTED_OBFUSCATION_VERSION)
 }
```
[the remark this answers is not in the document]
Eldolfin: No need, this is closer to what the spec says and the compiler will probably take care of it anyways

libdd-trace-stats/src/stats_exporter.rs
Eldolfin: I don't think that's an issue

libdd-trace-stats/src/span_concentrator/mod.rs
Eldolfin: Not a problem, the spec allows it and it's the same behavior as agent-side stats

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -21,13 +21,21 @@ pub use stat_span::StatSpan;
...
+    /// all buckets.
+    ///
+    /// Returns a triplet `(buckets, collapsed_spans, buckets_obfuscated)`
+    /// Where
+    ///   - `buckets` are the encoded stats bucket
+    ///   - `collapsed_spans` is the number of spans that were collapsed into the overflow sentinel
+    ///     bucket due to cardinality limiting
+    ///   - `buckets_obfuscated` indicates whether the returned buckets have been obfuscated
+    ///     client-side
+    fn flush_buckets(&mut self, force: bool) -> (Vec<pb::ClientStatsBucket>, u64, bool);
```
VianneyRuhlmann: I think this should return both obfuscated and unobfuscated buckets to avoid having to double-flush, something like

```rust
struct FlushResult {
  obfuscated_buckets: Vec<pb::ClientStatsBucket>,
  unobfuscated_buckets: Vec<pb::ClientStatsBucket>,
  collapsed_spans: u64
}
```

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -190,7 +198,23 @@ impl SpanConcentrator {
...
+            StatsBucket::new(
+                bucket_timestamp,
+                self.max_entries_per_bucket,
+                #[cfg(feature = "stats-obfuscation")]
+                self.obfuscation_config.load().enabled,
+            )
+        });
+        #[cfg(feature = "stats-obfuscation")]
+        let obfuscated_resource = Self::compute_obfuscated_span(
+            target_bucket.obfuscated,
```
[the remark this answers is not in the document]
VianneyRuhlmann: Well done codex you understood the goal of the PR :joy:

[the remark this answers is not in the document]
Eldolfin: That's the point of this PR fix

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -199,29 +223,27 @@ impl SpanConcentrator {
...
+    #[cfg(feature = "stats-obfuscation")]
     fn compute_obfuscated_span<'a>(
-        &self,
+        obfuscate: bool,
+        sql_obfuscation_mode: libdd_trace_obfuscation::sql::SqlObfuscationMode,
         #[allow(unused)] span: &'a impl StatSpan<'a>,
     ) -> Option<String> {
-        #[cfg(feature = "stats-obfuscation")]
-        if self.obfuscation_config.load().enabled {
+        if obfuscate {
```
VianneyRuhlmann: I think it's easier to follow if this alwasy applies obfuscation and you only call it if obfuscation is enabled else you pass the raw span

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -181,37 +173,61 @@ impl<Cap: HttpClientCapability + SleepCapability, Con: FlushableConcentrator>
...
-            return Ok(false);
+        // Obfuscated and un-obfuscated buckets must be sent in separate payloads because only
+        // the obfuscated one carries the `datadog-obfuscation-version` header.
+        let mut sent_stats = false;
+        if !flush.obfuscated_buckets.is_empty() {
+            self.send_payload(flush.obfuscated_buckets, true).await?;
+            sent_stats = true;
+        }
+        if !flush.unobfuscated_buckets.is_empty() {
+            self.send_payload(flush.unobfuscated_buckets, false).await?;
```
VianneyRuhlmann: You can use a FuturesUnordered here to allow both payloads to be sent concurrently

VianneyRuhlmann (approved):

PR #2204, opened 2026-07-07
test(data-pipeline): gate telemetry-only tests behind the telemetry feature

iunanua:
# What does this PR do?

Building the crate with `--no-default-features` (telemetry off) failed to compile because several tests and one example referenced `TelemetryConfig` and `TraceExporterBuilder::enable_telemetry`, both of which are gated behind the `telemetry` feature.

Gate the telemetry-dependent test code and the `send-traces-with-stats` example accordingly, and silence an unused-variable warning in stats.rs for non-telemetry builds. Non-telemetry tests in `test_new`/`test_new_defaults` still run.

### DIFF
--- a/libdd-data-pipeline/src/trace_exporter/mod.rs
+++ b/libdd-data-pipeline/src/trace_exporter/mod.rs
@@ -1118,11 +1118,7 @@ mod tests {
     use libdd_tinybytes::BytesString;
     use libdd_trace_utils::msgpack_encoder;
     use libdd_trace_utils::span::v04::SpanBytes;
-    use libdd_trace_utils::span::v05;
     use std::net;
-
-    // v05 messagepack empty payload -> [[""], []]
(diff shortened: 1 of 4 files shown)


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -2074,6 +2079,7 @@ mod tests {
         metrics_endpoint.assert_calls(1);
     }
 
+    #[cfg(feature = "telemetry")]
```
yannham: Nit: 3 tests is in the gray zone, but would that make sense to gather those tests under a separate submodule and gate the submodule once and for all?

yannham (approved):

libdd-data-pipeline/src/trace_exporter/mod.rs
iunanua: yes, they are better off in a submodule

PR #2205, opened 2026-07-07
feat(data-pipeline)!: add compression option for agentless export

paullegranddc:
# Motivation

Since we are sending data to the intake directly, we should be compressing requests as they can be multi MB.

The compression uses the zstd crate, which was already used in trace-utils.

It links to a C library, so it is not usable everyhwere (WASM for instance) and adds ~500KB of artifact size. For these reasons, it is hidden behind a feature flag and should be enabled as required (implementation using the agentless option mostly).

Another option would be rustzstd, an full rust port. Tge library is smaller (300KB) but it is notably slower.. (2x to 5x slower, with worse compression ratios at the same level)

# Changes

* Add a compression strategy parameter to send_with_retry
* implement zstd compression
* since compressio is implemented lower down, remove compression option from SendData
* Add a "compression" feature flag on data-pipeline to be used in languages where it is needed

BREAKING CHANGE: `libdd-trace-utils` public API changed — `send_with_retry` now
takes an additional `CompressionStrategy` argument, the `SendData` `Compression`
enum was replaced by `send_with_retry::CompressionStrategy`, and the implicit
`flate2` feature was removed from the `compression` feature set.


yannham (approved):
Beside some Codex comments that look relevant, the rest LGTM

PR #2207, opened 2026-07-07
fix(crashtracker): exclude crashing thread from ptrace stack collection

mabdinur:
## Problem

In v37, `stream_thread_contexts` began including `crashing_tid` in ptrace
collection, processing it first to guarantee it appears even when the thread
cap is reached.

The crashing thread is blocked in `poll()` inside its signal handler on the
alternate signal stack. When the receiver calls `PTRACE_INTERRUPT` on it,
`poll()` returns `EINTR`, the retry loop restarts `poll()`, and libunwind then
tries to unwind through the signal frame — which it cannot traverse correctly.
On Python 3.14 this `EINTR` loop is indefinite, hanging the receiver before the
crash report is delivered.

Discovered via `test_crashtracker_native_extension_crash` in dd-trace-py after
bumping to libdatadog v37. dd-trace-py is currently pinned to this fix commit
pending a tagged release.

## Fix

Restore the v35 behaviour: exclude `crashing_tid` from `stream_thread_contexts`.
The crash-site register context is already captured from `ucontext_t` in the
signal handler before ptrace collection runs, so no information is lost.

## Changes

- `libdd-crashtracker/src/receiver/ptrace_collector.rs`: filter out `crashing_tid`
  from the ptrace loop; update doc comment and two unit tests accordingly.

### DIFF
--- a/libdd-crashtracker/src/receiver/ptrace_collector.rs
+++ b/libdd-crashtracker/src/receiver/ptrace_collector.rs
@@ -558,47 +558,6 @@ mod tests {
         }
     }
 
-    #[test]
-    #[cfg_attr(miri, ignore)]
-    fn stream_excludes_crashing_tid() {
-        let barrier = Arc::new(Barrier::new(2));
-        let b: Arc<Barrier> = Arc::clone(&barrier);
-        let (tx, rx) = std::sync::mpsc::channel();
-
-        let handle = std::thread::spawn(move || {
-            tx.send(current_tid()).unwrap();
-            b.wait();
-        });
-
-        let worker_tid = rx.recv().unwrap();
-
-        let mut seen_worker = false;
-        let mut seen_self = false;
-        let self_tid = current_tid();
-        // Declare the current thread as the crashing TID; it must be skipped.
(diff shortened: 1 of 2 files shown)


mabdinur:
Fix was validated on this commit: https://github.com/DataDog/dd-trace-py/pull/18821/changes/52fa5a9f181798d461b2d1007b7191513d2657ab

mabdinur:
@codex review

gyuheon0h:
I think its a GIL issue when trying to retrieve runtime stacks from the python side given that when I tried myself and switched `CDLL` to `PyDLL`, the issue went away. Can you try with this commit on v37 on your dd-trace-py upgrade branch?

This worked for me

https://github.com/DataDog/dd-trace-py/commit/9d4346001701590b4be3e0308d0a2726ae9fb3bb

mabdinur:
Issue was resolved by: https://github.com/DataDog/dd-trace-py/pull/18821/changes/b441a95ddd60b9a79bc97ce2a4a09b3d5881679d

Thanks @gyuheon0h 🙏

PR #2212, opened 2026-07-08
feat(stats)!: add telemetry for per-field cardinality limits

Eldolfin:
# What does this PR do?

Send telemetry about per-field cardinality collapses.

# Motivation

CSS spec compliance.

# Additional Notes

Not a fan of how much I need to duplicate some parts of the code. But I wanted to avoid using HashMaps everywhere and keep the changes simple.

Maybe some sort of macro magic can solve this ? But I'm afraid it might hurt readability even more. I'm open to suggestions 

# How to test the change?

### DIFF
--- a/libdd-trace-stats/src/span_concentrator/aggregation.rs
+++ b/libdd-trace-stats/src/span_concentrator/aggregation.rs
@@ -434,7 +434,6 @@ pub(super) struct StatsBucket {
     /// Maximum number of distinct aggregation keys this bucket will hold before collapsing new
     /// ones into the overflow sentinel key.
     cardinality_limits: CardinalityLimitConfig,
-    // FIXME: optimize memory by storing hashes only
     distinct_resources: HashSet<u64>,
     distinct_http_endpoint: HashSet<u64>,
     distinct_peer_tags: HashSet<u64>,
@@ -486,6 +485,10 @@ impl StatsBucket {
         is_error: bool,
         is_top_level: bool,
     ) {
+        // Per field cardinality limiting
+        self.collapse_key_fields_cardinality(&mut key);
+
+        // Whole-key cardinality limiting
         if self.data.len() >= self.cardinality_limits.whole_key_limit
             && !self.data.contains_key(&key)
         {
@@ -497,8 +500,6 @@ impl StatsBucket {
             return;
         }
 
-        self.collapse_key_cardinality(&mut key);
-
         self.data
             .entry_ref(&key)
             .or_default()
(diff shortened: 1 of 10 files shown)


Eldolfin:
@codex review

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -370,21 +376,43 @@ impl SpanConcentrator {
...
+        if !self.cardinality_log_emitted && total_collapsed.whole_key > 0 {
+            self.cardinality_log_emitted = true;
             debug!(
                 max_entries_per_bucket = self.cardinality_limits.whole_key_limit,
-                total_collapsed,
+                total_whole_key_collapsed = total_collapsed.whole_key,
                 "Client-side stats values have been collapsed to 'tracer_blocked_value'. This is due to the cardinality exceeding DD_TRACE_STATS_CARDINALITY_LIMIT"
             );
         }
+        if !self.cardinality_log_emitted
```
paullegranddc: Not a blocker, but I think we should have one `cardinality_log_emitted` per field, since we can collapse mutliple times with different cardinalities

paullegranddc (approved):

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -370,21 +383,56 @@ impl SpanConcentrator {
...
                 #[cfg(not(feature = "stats-obfuscation"))]
                 let obfuscated = false;
                 Some((obfuscated, encode(bucket, self.bucket_size)))
             })
             .collect();
-        if total_collapsed > 0 {
-            debug!(
+
+        if total_collapsed.whole_key > 0 {
+            once!(debug!(
```
VianneyRuhlmann, line 396: It must be emited once _per flush period_ so no need to add a once macro

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -131,14 +145,54 @@ impl<Cap: HttpClientCapability + SleepCapability, Con: FlushableConcentrator>
...
+            let collapsed_additional_tags = handle.register_metric_context(
+                COLLAPSED_SPANS_TELEMETRY_METRIC.to_string(),
+                vec![libdd_common::tag!(
+                    "collapsed_spans",
+                    "additional_metric_tags"
+                )],
+                libdd_telemetry::data::metrics::MetricType::Count,
+                true,
+                libdd_telemetry::data::metrics::MetricNamespace::Tracers,
+            );
```
VianneyRuhlmann, line 185: From the RFC

> (Multiple `collapsed:` tags should be added when more than one field is collapsed)

I think it means we should send a single count metric with all the offending fields in the bucket instead of separate count metrics. I think the reason is because otherwise COLLAPSED_SPANS_TELEMETRY_METRIC will not count the number of collapsed spans but the total number of fields collapsed.

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -434,13 +459,13 @@ pub(super) struct StatsBucket {
     /// Maximum number of distinct aggregation keys this bucket will hold before collapsing new
     /// ones into the overflow sentinel key.
     cardinality_limits: CardinalityLimitConfig,
+    // HashSet of hashes of values so save memory, no need for total accuracy here
```
VianneyRuhlmann, line 462: Maybe detail a bit more why it's not accurate and how much collision we can escape as it can be a bit confusing we reading the code

Eldolfin:
will use #2227 instead

PR #2220, opened 2026-07-09
fix(shared-runtime): make ForkSafeRuntime teardown structurally TLS/poison-safe

quinna-h:
# What does this PR do?

Replaces the `shutdown()`-only TLS guard with two structural invariants so `ForkSafeRuntime` teardown never depends on the destruction order of other global/thread-local state during (embedded) interpreter finalization.

Also extends the same guard to `BlockingRuntime::block_on`, via a new opt-in `block_on_send` method, since `shutdown()` alone doesn't cover every path that can run during finalization (see Motivation).

# Motivation

During (embedded) interpreter finalization, a `ForkSafeRuntime` can be torn down after the calling thread's Tokio CONTEXT thread-local has already been destroyed. When that happens, teardown panics with `"The Tokio context thread-local variable has been destroyed"`:

- `shutdown()` calls `runtime.block_on(...)`, which enters the Tokio context via the thread-local — panicking if it's already gone.
- Even without an explicit `shutdown()`, a normal `Runtime` drop blocks the current thread to join its worker threads, and that join path also touches the context thread-local, so a **worker thread** panics during drop.

Because `lock_or_panic()` unwraps a poisoned mutex, that first panic then cascades: every other thread that later locks the shared runtime/workers mutex panics too with `PoisonError`.

This surfaces as a crash on every uWSGI worker shutdown with ddtrace ≥4.9.x (lazy-apps + `--die-on-term`). It's fundamentally an ordering problem; the process-global runtime, its worker threads, and thread-local state are destroyed
in an order the library doesn't control, and under uWSGI the teardown is driven from the C layer with no reliable point to shut down cleanly beforehand (the worker's SIGTERM is handled by uWSGI; Python's atexit runs only after TLS is
already destroyed). So the terminal teardown path must be safe unconditionally, rather than relying on detecting the finalization state at each call site.

Testing against this fix alone still reproduced the panic: `TraceExporter::shutdown()` (called from `TraceExporterPy::drop()` during interpreter finalization) goes through `BlockingRuntime::block_on()`, not `ForkSafeRuntime::shutdown()` — and `block_on()` had no TLS guard at all, so it hit the exact same thread-local panic this PR otherwise fixes.

`block_on`'s existing signature can't take a TLS-destruction guard directly without adding `Send + 'static` bounds that break other generic callers (e.g. the streaming send path, which isn't `Send`-bounded). Instead, a new `block_on_send` method is added to `BlockingRuntime` — `Send + 'static`-bounded, defaulting to forwarding to `block_on` — and `ForkSafeRuntime`'s override falls back to a helper thread (mirroring `shutdown_from_helper_thread`) when `can_block_on()` is false. Only `TraceExporter::shutdown()` was switched to call it; every other `block_on` call site is untouched.

# Additional Notes

**Known limitation — this is a stability fix, not a data-integrity fix.** The `block_on_send` helper-thread fallback stops the panic/crash, but it doesn't guarantee the in-flight flush actually succeeds during interpreter finalization — the runtime's I/O path (sockets, event loop state, etc.) may already be in a state where the send can't complete even without a panic. So this can still silently lose the same data `--skip-at-exit` loses; it just fails quietly instead of crashing the worker. Acceptable short-term since none of the current native shutdown hooks are essential yet (no CSS impact, some risk to native telemetry), but not a long-term fix for data loss on shutdown. Longer-term direction being investigated separately: hooking into worker teardown to run our atexit-equivalent logic before TLS is destroyed, and/or an upstream uwsgi change.

# How to test the change?

Reproduced locally with a standalone uWSGI worker harness (lazy-apps, `--die-on-term`) that sends a `TraceExporterPy` through interpreter finalization on SIGTERM:
- Without this fix: the worker panics on shutdown with `"The Tokio context thread-local variable has been destroyed"`.
- With this fix: the worker exits cleanly and is respawned normally — verified across multiple repeated runs.

`cargo test -p libdd-shared-runtime` (23/23 passing, including 3 new tests for `block_on_send` and its helper-thread fallback), plus `cargo fmt`/`clippy -D warnings`/full workspace `check` on the touched crates.


libdd-common/src/lib.rs
```diff
@@ -93,6 +99,11 @@ impl<T> MutexExt<T> for Mutex<T> {
         #[allow(clippy::unwrap_used)]
         self.lock().unwrap()
     }
+
+    #[inline(always)]
+    fn lock_or_recover(&self) -> MutexGuard<'_, T> {
+        self.lock().unwrap_or_else(|poisoned| poisoned.into_inner())
```
VianneyRuhlmann, line 105: The PoisonError should not be used, we have no guarantee over the state in which the value was left.

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
```diff
@@ -169,6 +196,34 @@ impl ForkSafeRuntime {
...
+    /// `shutdown_background` instead signals shutdown and returns immediately, detaching
+    /// the worker threads; the OS reclaims them on process exit. This is done
+    /// *unconditionally* — we do not try to detect whether we are in finalization —
+    /// so there is no ordering-dependent branch that a future teardown scenario can
+    /// slip past. If `shutdown` already took the runtime (the graceful path ran), this
+    /// is a no-op.
+    fn drop(&mut self) {
+        if let Some(runtime) = self.runtime.lock_or_recover().take() {
+            match Arc::try_unwrap(runtime) {
+                Ok(runtime) => runtime.shutdown_background(),
```
VianneyRuhlmann, line 261: This won't trigger the shutdown logic of the workers and therefore will not flush data before the fork this will cause the same issue as `--skip-atexit`

VianneyRuhlmann (changes requested):

danielsn:
I fundamentally worry we're going the wrong direction here. Tokio has undocumented global state, and may just be incompatable with forking.  We can play whack-a-mole with fork-based crashes on a multi-threaded runtime, but do we have a design where we believe we ought to eventually converge on true fork safety?

libdd-common/src/lib.rs
mabdinur: I narrowed lock_or_recover to only the terminal teardown paths. Let me know if we should remove it all together (I can use parking_lot::Mutex instead)

libdd-shared-runtime/src/shared_runtime/fork_safe.rs
mabdinur: Drop::shutdown_background() now only fires if shutdown() was never called at all before the runtime is dropped

VianneyRuhlmann:
> I fundamentally worry we're going the wrong direction here. Tokio has undocumented global state, and may just be incompatable with forking. We can play whack-a-mole with fork-based crashes on a multi-threaded runtime, but do we have a design where we believe we ought to eventually converge on true fork safety?

@danielsn This issue doesn't seem to be linked to fork-safety. The issue is uwsgi custom tear-down logic in workers, it would still happen if the worker was started with a proper spawn.

mabdinur:
closing in favor of: https://github.com/DataDog/dd-trace-py/pull/19086

PR #2159, opened 2026-06-24
feat(stats)!: send telemetry for cardinality limits

VianneyRuhlmann:
# What does this PR do?

Add telemetry for the collapsed stats group based on [this RFC](https://datadoghq.atlassian.net/wiki/spaces/APM/pages/6821151019/PENDING+Cardinality+Limits)

# Motivation

Follow-up of #2158 

# Additional Notes

Anything else we should know when reviewing?

# How to test the change?

Describe here in detail how the change can be validated.

PR #2222, opened 2026-07-10
refactor(statsd)!: make client clonable

VianneyRuhlmann:
# What does this PR do?

Refactor of the dogstatsd client crate.

Main changes are:

- Separated code in 3 modules, `action` `client` and `client/sink`
- Renamed `Client` to `DogStatDClient` to avoid renames when importing in other crates
- Move `new`, `create_client` and `do_send` to be method of the `DogStatsDClient`
- Replaced Mutex in `DogStatsDClient` to use a `OnceLock` and make the client handle clonable

# Motivation

I've encountered several issues with the current design when sharing the client with the StatsExporter in #2159 and when adding the SharedRuntime backed client #2224 .

# Additional Notes

# How to test the change?


VianneyRuhlmann:
* **#2224** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2222** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2222?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2222?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libdd-dogstatsd-client/src/client/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+pub struct DogStatsDClient {
+    inner: Arc<InnerClient>,
+}
+
+/// Inner struct of [DogStatsDClient] to be wrapped in [Arc].
+#[derive(Debug, Default)]
+struct InnerClient {
+    client: OnceLock<StatsdClient>,
+    endpoint: Endpoint,
+}
```
Aaalibaba42, line 34: nit: I understand why it's more legible this way but I think I'd have preferred a `struct DogStatsDClient(Arc<InnerClient>)` or even `struct DogStatsDClient(Arc<(OnceLock<StatsdClient>, Endpoint)>)`, I feel like this Inner struct is just for language reasons and not sementic

libdd-dogstatsd-client/src/client/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+mod sink;
+
+// Queue with a maximum capacity of 32K elements
+const QUEUE_SIZE: usize = 32 * 1024;
+
+/// A dogstatsd-client that flushes stats to a given endpoint.
+///
+/// This client can be cloned and shared between threads.
+#[derive(Debug, Default, Clone)]
+pub struct DogStatsDClient {
```
Aaalibaba42, line 25: nit: Previous `StatsdClient` seemed to not capitalize the same way, better be consistent in those things.

libdd-dogstatsd-client/src/client/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    fn get_or_init_client(&self) -> anyhow::Result<&StatsdClient> {
+        match self.inner.client.get() {
+            Some(client) => Ok(client),
+            None => {
+                let client = Self::create_client(&self.inner.endpoint)?;
+                Ok(self.inner.client.get_or_init(|| client))
+            }
+        }
```
Aaalibaba42, line 129: nit: slightly more legible imho

```suggestion
        if let Some(client) = self.inner.client.get() {
            return Ok(client);
        }

        let client = Self::create_client(&self.inner.endpoint)?;
        Ok(self.inner.client.get_or_init(|| client))
```

libdd-dogstatsd-client/src/client/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+        let mut tags_iter = tags.into_iter();
+        let mut tag_opt = tags_iter.next();
+        #[allow(clippy::unwrap_used)]
+        while tag_opt.is_some() {
+            builder = builder.with_tag_value(tag_opt.unwrap().as_ref());
+            tag_opt = tags_iter.next();
+        }
+        builder.try_send()?;
+        Ok(())
+    }
```
Aaalibaba42, line 159: nit:
```suggestion
    {
        let mut tags_iter = tags.into_iter();
        while let Some(tag) = tag_opt.next() {
            builder = builder.with_tag_value(tog.as_ref());
        }
        builder.try_send()?;
        Ok(())
    }
```

or even just this since V is `IntoIterator`


```suggestion
    {
        for tag in tags {
            builder = builder.with_tag_value(tag.as_ref());
        }
        builder.try_send()?;
        Ok(())
    }
```

libdd-dogstatsd-client/src/client/mod.rs
Aaalibaba42: Whelp apparently pre-existing `DogStatsDAction` is capitalized like you did now. Maybe it's out of scope, but I think either adding the capital D to the `StatsdClient` or removing it from this and the Action type thing would be nice

Aaalibaba42 (commented):
Mostly nits, I don't know dogstatsd enough to approve though, just looked at the code.

libdd-dogstatsd-client/src/client/mod.rs
VianneyRuhlmann: StatsdClient is actually a struct from cadence (rust statsd crate). I agree that's it's not great, I think rename DogStatsDAction isn't too disruptive.

libdd-dogstatsd-client/src/client/mod.rs
VianneyRuhlmann: Good catch this code has some really weird designs

libdd-dogstatsd-client/src/client/mod.rs
VianneyRuhlmann: I agree with `struct DogStatsDClient(Arc<InnerClient>)` but I'd rather have a proper struct to held Client fields rather than a tuple which is less readable in methods

libdd-dogstatsd-client/src/client/mod.rs
VianneyRuhlmann: Actually it changes in a lot of places, I'm not sure it's worth it since StatsdClient is not public anyway

libdd-dogstatsd-client/src/client/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+            }
+        }
+    }
+
+    fn get_or_init_client(&self) -> anyhow::Result<&StatsdClient> {
+        match self.inner.client.get() {
+            Some(client) => Ok(client),
+            None => {
+                let client = Self::create_client(&self.inner.endpoint)?;
+                Ok(self.inner.client.get_or_init(|| client))
```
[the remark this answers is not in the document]
VianneyRuhlmann: Considering there's only a few workers sharing the same client this should be quite rare, once it's stabilized we can start using [get_or_try_init](https://doc.rust-lang.org/beta/std/sync/struct.OnceLock.html#method.get_or_try_init)

libdd-dogstatsd-client/src/client/mod.rs
VianneyRuhlmann: I think the match is fine here

Aaalibaba42 (approved):

PR #2223, opened 2026-07-10
feat(metrics-v3): add Datadog V3 payload encoder

atanzu:
# What does this PR do?

Add a new library which allows to encode metric payloads using V3 columnar format.

# Motivation

This library enables users to encode their metric payloads using an efficient column-based protocol. For performance and compatibility reasons (we also want to keep this crate `no_std` so we could use it in very resource-constrained environments) this crate does manual protobuf serialization.

# Additional Notes

# How to test the change?

We do correctness test by comparing the resulting encoded payload with the one produced from protobuf-generated code.


atanzu:
Encoder will be moved to agent-payload repo.

PR #2225, opened 2026-07-10
chore(heap-profiling): remove `free` USDT when live-heap is not available [PROF-15190]

scottgerring:
# What does this PR do?
When the user does not opt into the `live-heap` feature, we should not emit the `free` USDT. This lets an attached profiler know that live heap is not available, making its life easier (e.g., it knows not to track sampled allocations). 

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-profiling-heap-sampler/src/allocation_freed.c
+++ b/libdd-profiling-heap-sampler/src/allocation_freed.c
@@ -22,7 +22,9 @@ dd_alloc_freed_t dd_allocation_freed_slow(void *ptr, void *raw, size_t size,
                                           size_t alignment) {
     /* Fire with the user-visible pointer, matching what was reported at alloc
      * time, so the profiler can correlate the two events by address. */
+#if DD_HEAP_LIVE_TRACKING
     dd_probe_free(ptr);
+#endif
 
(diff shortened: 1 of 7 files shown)


libdd-profiling-heap-sampler/src/allocation_freed.c
```diff
@@ -21,7 +22,9 @@
 dd_alloc_freed_t dd_allocation_freed_slow(void *ptr, void *raw, size_t size,
                                           size_t alignment) {
     /* Fire with the user-visible pointer, matching what was reported at alloc
-     * time, so the profiler can correlate the two events by address. */
+     * time, so the profiler can correlate the two events by address. Safe to
+     * call unconditionally: dd_probe_free is a no-op USDT-wise when live-heap
```
danielsn, line 26: is there a test for this

scottgerring: added

r1viollet: nice

nsavoire (approved):

PR #2227, opened 2026-07-10
feat(stats)!: add telemetry for per-field cardinality limits

Eldolfin:
# What does this PR do?

Send telemetry about per-field cardinality collapses. But unlike https://github.com/DataDog/libdatadog/pull/2212, now using a single count metric with all the offending fields in the bucket instead of separate count metrics.

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/datadog-ipc/src/shm_stats.rs
+++ b/datadog-ipc/src/shm_stats.rs
@@ -57,7 +57,7 @@ use zwohash::ZwoHasher;
 use libdd_ddsketch::DDSketch;
 use libdd_trace_protobuf::pb;
 use libdd_trace_stats::span_concentrator::{
-    FixedAggregationKey, FlushResult, FlushableConcentrator,
+    CollapsedFieldsMetrics, FixedAggregationKey, FlushResult, FlushableConcentrator,
 };
 
 use crate::platform::{FileBackedHandle, MappedMem, NamedShmHandle};
@@ -823,11 +823,12 @@ impl ShmSpanConcentrator {
 
 impl FlushableConcentrator for ShmSpanConcentrator {
     fn flush_buckets(&mut self, force: bool) -> FlushResult {
-        // The SHM concentrator does not perform client-side obfuscation.
+        // The SHM concentrator does not perform client-side obfuscation nor emits telemetry.
         FlushResult {
             obfuscated_buckets: vec![],
             unobfuscated_buckets: self.drain_buckets(force),
             collapsed_spans: 0,
+            collapsed_fields_metrics: CollapsedFieldsMetrics::zero(),
         }
     }
 }
--- a/datadog-ipc/src/shm_stats.rs
+++ b/datadog-ipc/src/shm_stats.rs
@@ -57,7 +57,8 @@ use zwohash::ZwoHasher;
 use libdd_ddsketch::DDSketch;
 use libdd_trace_protobuf::pb;
 use libdd_trace_stats::span_concentrator::{
-    CollapsedFieldsMetrics, FixedAggregationKey, FlushResult, FlushableConcentrator,
+    cardinality_limit_telemetry::CollapsedFieldsMetrics, FixedAggregationKey, FlushResult,
+    FlushableConcentrator,
 };
 
 use crate::platform::{FileBackedHandle, MappedMem, NamedShmHandle};
--- a/datadog-ipc/src/shm_stats.rs
+++ b/datadog-ipc/src/shm_stats.rs
@@ -823,7 +823,7 @@ impl ShmSpanConcentrator {
 }
 
 impl FlushableConcentrator for ShmSpanConcentrator {
-    fn flush_buckets(&mut self, force: bool) -> FlushResult {
+    fn flush_buckets(&mut self, force: bool) -> FlushResult<pb::ClientStatsBucket> {
         // The SHM concentrator does not perform client-side obfuscation nor emits telemetry.
         FlushResult {
             obfuscated_buckets: vec![],
--- a/libdd-trace-stats/src/span_concentrator/aggregation.rs
+++ b/libdd-trace-stats/src/span_concentrator/aggregation.rs
@@ -15,7 +15,7 @@ use std::{
 };
 use tracing::warn;
 
-use crate::span_concentrator::{cardinality_limit_telemetry::collapsed_field, StatSpan};
+use crate::span_concentrator::{cardinality_limit_telemetry::CollapsedFieldSet, StatSpan};
 
 use super::{
     cardinality_limit_telemetry::{self, CollapsedFieldsMetrics},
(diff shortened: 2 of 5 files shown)


Eldolfin:
@codex review

Eldolfin:
@codex review

libdd-trace-stats/src/span_concentrator/cardinality_limit_telemetry.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub const PEER_TAGS: usize = 1 << 3;
+    pub const ADDITIONAL_TAGS: usize = 1 << 4;
+    pub const COUNT: u8 = 5;
+}
+
+pub(super) const COLLAPSED_FIELD_METRIC_SIZE: usize = 1 << collapsed_field::COUNT;
+
+#[derive(Debug, Clone, Default, Copy)]
+// Note: slot 0 is a counter for non_collapsed spans. It's not used for emitting telemetry
+pub struct CollapsedFieldsMetrics(pub(super) [usize; COLLAPSED_FIELD_METRIC_SIZE]);
```
yannham: So if I understand correctly, you don't count each collapsed field individually, but you count _sets_ of collapsed fields instead. I'm curious about the reason behind it. I would naively expect that counting each field individually would be enough.

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -623,10 +638,12 @@ impl StatsBucket {
         if let Entry::Vacant(slot) = self.distinct_additional_tags.entry(additional_tags_hash) {
             if additional_tags_count >= self.cardinality_limits.additional_tags_limit {
                 key.additional_metric_tags = vec![(TRACER_BLOCKED_VALUE, "")];
+                collapsed_fields |= collapsed_field::ADDITIONAL_TAGS;
             } else {
                 slot.insert();
             }
         }
+        self.collapsed_fields_metrics.0[collapsed_fields] += 1;
```
yannham: I think the API of collapsed fields is leaking. The consuming code here shouldn't have to dabble with C-like fieldbits and accessing the internal array of `collapsed_fields_metrics`, IMHO.

Instead, you could define something like e.g.

```
struct CollapsedFieldSet(usize);

impl CollapsedFieldSet {
    const RESOURCE_NAME: usize = 1 << 1;
    // etc.

    fn add(&mut self, field: usize) {
        self.0 |= field;
    }
}
```

Then you could use `field_set.add(CollapsedFieldSet::RESOURCE_NAME)`, which doesn't depend on the implementation being a bit field. Similarly an `increment(&mut self, field_set: CollapsedFieldSet)` on `CollapsedFieldMetrics` could make the API nicer (whether you need to keep the internal array visible for sending the metrics is orthogonal, and is ok)

libdd-trace-stats/src/span_concentrator/cardinality_limit_telemetry.rs
```diff
@@ -0,0 +1,112 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Each collapsed field is assigned a bit; a span's mask is the OR of the bits for the
+//! fields collapsed on it, so masks double as an index into
+//! [`CollapsedFieldsMetrics`]'s counter array (one counter per distinct field combination).
```
yannham: This comment is hard to digest.

libdd-trace-stats/src/span_concentrator/cardinality_limit_telemetry.rs
```diff
@@ -0,0 +1,112 @@
...
+                collapsed_field::HTTP_ENDPOINT => {
+                    libdd_common::tag!("collapsed_spans", "http_endpoint")
+                }
+                collapsed_field::PEER_TAGS => {
+                    libdd_common::tag!("collapsed_spans", "peer_tags")
+                }
+                collapsed_field::ADDITIONAL_TAGS => {
+                    libdd_common::tag!("collapsed_spans", "additional_metric_tags")
+                }
+                // Unreachable: asserted just above that field is one of the 4 possible values
```
yannham: Nit: now the assert is debug only

```suggestion
                // Should be unreachable, but don't fail in prod if provided with an invalid field set
```

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -398,7 +406,7 @@ impl SpanConcentrator {
         now: SystemTime,
         force: bool,
         encode: impl Fn(StatsBucket, u64) -> T,
-    ) -> (Vec<(bool, T)>, u64) {
+    ) -> (Vec<(bool, T)>, u64, CollapsedFieldsMetrics) {
```
yannham: Nit: maybe this should be a proper struct with explicit field names? A 3 element tuple starts to be hard to read/guess.

datadog-ipc/src/shm_stats.rs
```diff
@@ -823,11 +824,12 @@ impl ShmSpanConcentrator {
 
 impl FlushableConcentrator for ShmSpanConcentrator {
     fn flush_buckets(&mut self, force: bool) -> FlushResult {
-        // The SHM concentrator does not perform client-side obfuscation.
+        // The SHM concentrator does not perform client-side obfuscation nor emits telemetry.
```
VianneyRuhlmann, line 827: That's an issue we want the sidecar to send metrics. The issue is pre-existing (that's on me) `flush_buckets` shouldn't return `collapsed_spans: 0` this can be addressed in follow-up but the ShmSpanConcentrator must send the metrics.

libdd-trace-stats/src/span_concentrator/cardinality_limit_telemetry.rs
VianneyRuhlmann: This comes from the [cardinality spec](https://datadoghq.atlassian.net/wiki/spaces/APM/pages/6821151019/APPROVED+Cardinality+Limits#5.-Telemetry). Since we apply tags to increment of a global count we want the global count to be the number of collapsed spans.   
  
This is easy to do if you increment the count when you collapse you just apply the correct tags to the increment. However this means adding a reference to the telemetry/statd client to the SpanConcentrator which is currently "pure computation logic" making it harder to reuse in serverless/sidecar.  
  
This was found as a way to allow the StatsExporter to emit the metric by giving it for each increment the offending fields. However this arguably not much better than adding a telemetry/dogstatsd handle to the concentrator.

libdd-trace-stats/src/span_concentrator/aggregation.rs
Eldolfin: Fixed in 43e494f, thanks it's a lot nicer now

datadog-ipc/src/shm_stats.rs
Eldolfin: Because `ShmSpanConcentrator` doesn't do stats cardinality limits (yet), this doesn't apply.

VianneyRuhlmann (approved):

PR #2176, opened 2026-06-29
feat(library-config): otel process context reader

cataphract:
[PROF-15479](https://datadoghq.atlassian.net/browse/PROF-15479)

Add `read()`, `threadlocal_attribute_key_map()`, and `read_threadlocal_attribute_key_map()` to `otel_process_ctx::linux`, along with the `find_otel_mapping()` / `is_named_otel_mapping()` / `parse_mapping_start()`.

# What does this PR do?

A brief description of the change being made with this pull request.

# Motivation

What inspired you to submit this pull request?

# Additional Notes

Anything else we should know when reviewing?

# How to test the change?

Describe here in detail how the change can be validated.

BREAKING CHANGE: unpublish becomes unsafe.

PR #2228, opened 2026-07-10
refactor(library-config): reorganize Linux process context

cataphract:
[PROF-15480](https://datadoghq.atlassian.net/browse/PROF-15480)

## What does this PR do?

This is **1/3** in the stacked process-context series.

It reorganizes the existing Linux process-context implementation without changing its behavior:

- separates shared reader and writer logic into dedicated modules;
- moves Linux-specific allocation, discovery, clock, and copy behavior behind platform modules;
- introduces independent reader and writer Cargo features while preserving the existing Linux defaults at this layer;
- adds platform-agnostic exports and retains the Linux compatibility path;
- preserves the existing functionality exactly, including the edge cases and bugs addressed by the next PR.

## Why?

This creates the module structure needed for additional operating systems without mixing a large code move with functional changes. Reviewers can evaluate the organization independently before the Linux implementation is updated and the new platforms are added.

## Impact

There is no intended runtime behavior change. Linux remains the only supported implementation in this PR, and existing Linux callers retain the same behavior.

## Validation

Validated independently at this commit with:

- `cargo check -p libdd-library-config --all-features`;
- `cargo check -p libdd-library-config-ffi`;
- all-feature process-context tests;
- reader-only process-context tests.

## Stack

1. **#2228 — reorganize the Linux process context**
2. #2237 — update the Linux implementation
3. #2238 — add macOS and Windows support

BREAKING CHANGE: moves `ProcessContextSelfReader` from the Linux-specific module to the platform-agnostic process-context API, while retaining the deprecated Linux compatibility path.


libdd-library-config/src/tracer_metadata.rs
```diff
@@ -220,7 +220,8 @@ mod linux {
     pub fn store_tracer_metadata(
         data: &super::TracerMetadata,
     ) -> anyhow::Result<super::AnonymousFileHandle> {
-        let _ = crate::otel_process_ctx::linux::publish(&data.to_otel_process_ctx());
+        #[cfg(feature = "process-context-writer")]
+        let _ = crate::otel_process_ctx::publish(&data.to_otel_process_ctx());
```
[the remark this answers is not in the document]
cataphract: Yes, I only made downstream changes; dd-trace-php is usign directly

```
    let context = metadata.to_otel_process_ctx();

    // libdatadog owns the process-wide writer and transparently updates an existing publication or
    // creates a fresh platform-specific publication after a fork.
    match otel_process_ctx::publish(&context) {
```

I'll change this part as well in a later commit

libdd-library-config/src/otel_process_ctx/reader/macos.rs
```diff
@@ -0,0 +1,59 @@
...
+        let header = process_context_global()?.load(Ordering::Acquire);
+        NonNull::new(header).ok_or_else(|| {
+            io::Error::new(io::ErrorKind::NotFound, "no process context is published")
+        })
+    }
+}
+
+#[cfg(feature = "process-context-writer")]
+fn process_context_global() -> io::Result<&'static AtomicPtr<u8>> {
+    Ok(&crate::otel_process_ctx::writer::macos::otel_process_ctx_v2)
```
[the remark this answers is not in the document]
cataphract: Kind of OK, but I'm making the writer the only default feature

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    fn copy(&self, addr: *const u8, len: usize) -> Result<Vec<u8>, PipeCopyError> {
+        let mut bytes: Vec<u8> = Vec::with_capacity(len);
+        let mut offset = 0;
+
+        while offset < len {
+            let chunk_len = (len - offset).min(self.chunk_size);
+            let chunk_addr = addr.wrapping_add(offset);
+
+            // SAFETY: write asks the kernel to copy from chunk_addr. Invalid user memory is
```
danielsn, line 42: Reading the [man page](https://man7.org/linux/man-pages/man2/write.2.html), it seems like error detection may occur but I don't see where it is guaranteed?  POSIX [does not appear to require](https://pubs.opengroup.org/onlinepubs/009696899/functions/write.html) EFAULT at all.

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -0,0 +1,298 @@
...
+        let (read_fd, write_fd, chunk_size) = create_pipe()?;
+
+        Ok(Self {
+            read_fd,
+            write_fd,
+            chunk_size,
+        })
+    }
+
+    fn copy(&self, addr: *const u8, len: usize) -> Result<Vec<u8>, PipeCopyError> {
```
danielsn, line 34: Does this function have SAFETY requirements?

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -0,0 +1,298 @@
...
+        let mut bytes: Vec<u8> = Vec::with_capacity(len);
+        let mut offset = 0;
+
+        while offset < len {
+            let chunk_len = (len - offset).min(self.chunk_size);
+            let chunk_addr = addr.wrapping_add(offset);
+
+            // SAFETY: write asks the kernel to copy from chunk_addr. Invalid user memory is
+            // reported as EFAULT or a short write without being dereferenced by Rust.
```
danielsn, line 43: What are the semantics in the case of a short write

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -0,0 +1,298 @@
...
+                        });
+                    }
+                }
+            }
+
+            offset += written;
+        }
+
+        // SAFETY: every byte was initialized by the pipe reads above.
+        unsafe { bytes.set_len(len) };
```
danielsn: would it be safer to `vec![0,len]`, or `std::io::BorrowedBuf`?

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: We're only targeting Linux and  macOS. Both Linux's and mac os man pages describe this return value.

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: The whole point of the function is that it should not generate a fault if attempting to read unmapped or otherwise inaccessible manner. So there is no safety contract that the caller must respect in order for the invocation of the function not to trigger UB.

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: If there is a short write, then something interrupted the function after it's already written some data. It could have been interrupted, or it could have found an unreadable page.

But it turns out that, despite what the mac os man page says, write() always returns EFAULT upon encountering a fault, even if it's written something. This requires marking its error with pipe_dirty: true, which I'll do in a bit.

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: I answered a similar comment when this code was introduced (it was just moved in this PR): https://github.com/DataDog/libdatadog/pull/2176/changes/BASE..12ca390b7683b6efaeae6cf06078b26eaf41a3d9#diff-e3f807d69a162e59e2e82257164c7642090139e9d90f6be2fecde5205ead3636 :

> What difference does it make if it's initialized with zeros or not? The read syscall writes directly into the buffer and it should fill it to capacity. And if it there was a bug and it didn't, zeros or uninitialized memory would be equally wrong (I'd probably even prefer uninitialized memory because it's more likely to be caught by MSan).

BorrowedBuf would indeed be a small improvement, but it doesn't appear to be stable.

ivoanjo:
This is very cool, and I definitely want to give it a pass ASAP. Having said that, I'd like to request a change in the way we refer to this.

Specifically, at this time, we're not trying or proposing upstream any way of using the OTel process context and thread context specifications on macOS and Windows.

Thus I humbly request:
* Let's drop the "otel" from the naming we use when we refer to macOS or Windows -- let's just call it "process context" and "thread context"
* In any files that include macOS or Windows support, let's explicitly document that macOS and Windows support is not part of the spec and is an experimental datadog extension

My intent here is to give us fully the space to experiment and validate -- if this goes really well it will be great to fold it into the spec; but not confuse outside-Datadog folks, since the specs explicitly call out process context and thread context as being Linux-specific mechanisms for the time being.

ivoanjo:
Furthermore (forgot to add to the above) -- not being yet tied to the spec gives us flexibility to go "Oh actually we want to do it another different way on Windows" (or macOS) while we're experimenting, whereas if we were folding this right away into the spec it'd be weird to flip-flop as we experiment.

cataphract:
OK, I've renamed the PR and added a few comments. Note that this follows the specification to the extent that it can be followed. I's only the part that can only be implemented on Linux, the discovery mechanism, that differs. All the rest -- the data itself, how it's laid out, the update protocol, follows the spec.

ivoanjo:
At this point, this PR is too big -- I humbly request that we break it down into smaller steps. E.g. first one with all changes for Linux, and then separate windows and macOS support.

With so many changes it'll be a lot of work to figure out what's only moving and what's actually changing.

cataphract:
@ivoanjo I've split it into 3 commits. The first is just moves code around and has no change in the linux functionality. The second contains the updates I did to the linux implementation. Finally, the third adds windows and mac os. Hope this helps.

ivoanjo:
I'd really like separate PRs -- I feel like we're creating a huge PR that will take a bunch of time to deliver, whereas small PRs have small steps that we can agree on and move quickly on.

ivoanjo:
To be clear, I am aware that split PRs are a pain for the submitter -- yet that's exactly what I'm asking for here, I'm asking you to help us ship this fast by making it easier on the reviewers to move fast on it.

cataphract:
As a reviewer, I'd rather have the three commits in one PR, since the reorganization and abstractions make more sense in light in the mac os/windows implementations in the later commits. But to each his own

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
yannham: The difference is that you would get rid of this unsafe `set_len` I believe. You initialize it with the right length from the beginning and just use it as a fixed-sized array `[u8; N]`. No need to uphold any safety condition anymore.

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -0,0 +1,240 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
```
yannham, line 3: Nit: add a short module description

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
yannham: I think this contract (`copy` won't fail even for unmapped addresses or unitialized memory) should be specified at the trait level, in the documentation of `copy`. It's not obvious for anyone who hasn't sufficient context (as demonstrated by Daniel's question)

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        // SAFETY: every byte was initialized by the pipe reads above.
+        unsafe { bytes.set_len(len) };
+        Ok(bytes)
+    }
+}
+
+#[cfg(target_os = "linux")]
+fn create_pipe() -> io::Result<(OwnedFd, OwnedFd, usize)> {
```
yannham: Should the return type be a proper struct instead of a triplet?

libdd-library-config/src/otel_process_ctx/linux.rs
```diff
@@ -0,0 +1,22 @@
...
+use core::ffi::CStr;
+
+#[cfg(feature = "process-context-reader")]
+pub mod self_reader {
+    pub use crate::otel_process_ctx::ProcessContextSelfReader;
+}
+
+// Re-exported for backwards compatibility.
+#[cfg(feature = "process-context-writer")]
+#[deprecated(note = "use libdd_library_config::otel_process_ctx directly")]
```
yannham: I would be inclined to ignore backward compatibility and just get rid of those. The process context implementation is rather recent, there's little to no code that uses the `ProcessContext` API directly outside of libadatadog to my knowledge, and even then, we control that code. Any opinion @ivoanjo ?

libdd-library-config/src/otel_process_ctx/reader.rs
```diff
@@ -0,0 +1,268 @@
...
+#[cfg(target_os = "linux")]
+type PlatformHeaderDiscovery = linux::HeaderDiscovery;
+
+pub(super) trait ReaderPlatform {
+    fn discover_header() -> io::Result<NonNull<u8>>;
+}
+
+pub(super) trait ProcessMemoryCopy: Sized {
+    fn new() -> io::Result<Self>;
+    fn copy(&self, addr: *const u8, len: usize) -> Result<Vec<u8>, PipeCopyError>;
```
yannham, line 64: Re `copy`: the contract should be clearly specified here.

libdd-library-config/src/otel_process_ctx/reader.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+#[derive(Debug)]
+pub(super) struct PipeCopyError {
```
yannham, line 68: Isn't this a bit Linux-specific (the pipe implementation)? Maybe it's just a minor naming issue, and this should be `MemoryCopyError` or sth. But it feels like Linux implementation is leaking a bit here.

libdd-library-config/src/otel_process_ctx/reader.rs
```diff
@@ -0,0 +1,268 @@
...
+
+/// Reader for the current process's OTel process context.
+///
+/// The platform implementation locates the header at construction. Reads use a kernel-mediated
+/// copy so unpublish and payload replacement races return errors instead of dereferencing freed
+/// memory.
+pub struct ProcessContextSelfReader {
+    pid: u32,
+    pub(in crate::otel_process_ctx) header_ptr: NonNull<u8>,
+    pipe: Cell<Option<PlatformCopyPipe>>,
```
yannham, line 82: Isn't this linux-specific as well? I think `PlatformCopyPipe` is only defined on Linux, right? In general I would expect this module to have a very simple copy interface -- basically `ProcessMemoryCopy` that guarantees fault-free copy, without talking about pipes or whatnot, which is an implementation detail.

Then either `ProcessContextSelfReader` should be parametrized by a `Copier: ProcessMemoryCopy`, or if that's too heavy, define a `ProcessMemoryCopyImpl` for each platform in a dedicated module, and import it unconditionally here.

libdd-library-config/src/otel_process_ctx/reader.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// SAFETY: the reader owns its copy pipe and only stores a process-global address. Cell keeps the
+// type !Sync, so the cached pipe cannot be used concurrently through shared references.
+unsafe impl Send for ProcessContextSelfReader {}
```
yannham, line 89: Same, for modularity purpose, we should probably require `ProcessMemoryCopy` to be `Send`, moving the burden to the platform-specific implementation. Maybe the implementation on Windows wouldn't use pipe but a totally different mechanism; it's hard for `ProcessContextSelfReader` to justify in advance that all implementations for `ProcessMemoryCopy` in the future will be `Send`.

libdd-library-config/Cargo.toml
```diff
@@ -15,7 +15,10 @@ crate-type = ["lib"]
 bench = false
 
 [features]
-otel-thread-ctx = []
+default = ["process-context-reader", "process-context-writer"]
```
yannham, line 18: I would not make process-context-reader the default. For now it seems quite specific to the PHP profiler.

libdd-library-config/src/otel_process_ctx/reader.rs
```diff
@@ -0,0 +1,268 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
```
yannham, line 1: I got quite confused by the fact that there's a top-level `reader.rs` file, but also submodules in a different `reader/xxx` subdirectory. Same for writer. Would you have anything against moving `reader.s` and `writer.rs` to respective `mod.rs` files in the corresponding subdirectories? We would have everything writer-related and reader-related at the same place.

yannham (approved):
(Sorry, I meant to approve - none of the comments above are blocking)

libdd-library-config/src/otel_process_ctx.rs
```diff
@@ -18,676 +18,226 @@
...
+
+        super::unpublish().expect("couldn't unpublish the context");
+
+        assert!(
+            !is_published(),
+            "process context should not be visible after unpublishing"
+        );
+    }
+
+    /// The only end-to-end test with `ProcessContextSelfReader`
```
ivoanjo: Minor: The "The only" part of this comment looks like it can be made out-of-date very easily the next time someone adds another test; maybe simply to only "End to end test"?

libdd-library-config/src/otel_process_ctx.rs
```diff
@@ -18,676 +18,226 @@
...
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn publish_read_update_read_and_unpublish() {
+        fn context(service_name: &str) -> ProcessContext {
+            ProcessContext {
                 resource: None,
```
ivoanjo: Minor: Maybe put a few things in the resource? An empty resource is actually something we don't expect will happen very often, and thus putting stuff here would make the end-to-end test a bit more realistic

libdd-library-config/src/otel_process_ctx/writer/linux.rs
```diff
@@ -0,0 +1,264 @@
...
+                            libc::PROT_WRITE | libc::PROT_READ,
+                            libc::MAP_PRIVATE,
+                            fd.as_raw_fd(),
+                            0,
+                        )
+                    },
+                    "mmap failed"
+                )?;
+
+                // We (implicitly) close the file descriptor right away, but this ok
```
ivoanjo: Minor: I know this comment already existed before, yet on a second read I think we can make it stronger even -- the spec does say that the fd is to be closed, it's not a "happy accident" that we do so here.

```suggestion
                // We (implicitly) close the file descriptor right away on purpose
```

libdd-library-config/src/otel_process_ctx/linux.rs
ivoanjo: Yeap +1 on this; I believe most writers are going through libdd-library-config, except maybe node, and we can update that one as well, so I think it's fine to clean these up.

Double so for the reader -- since it's basically new I'm pretty sure nobody's using it yet

libdd-library-config/src/otel_process_ctx/linux/self_reader.rs
```diff
@@ -1,745 +0,0 @@
...
-//! By design, this reader emulates an out-of-process reader: it does not take the publisher's
-//! mutex or hold any guard that prevents the mapping from being unpublished, the header from being
-//! rewritten, or the payload allocation from being replaced while it copies memory. The
-//! seqlock-style timestamp checks and fences only let it detect concurrent publication or update;
-//! unpublish can make the memory disappear entirely.
-//!
-//! To handle this safely, instead of dereferencing process memory directly, we ask the kernel to
-//! copy those bytes into a pipe and then read the pipe back. This gives us the same failure mode an
-//! out-of-process reader has: if memory is unmapped or invalid, the copy fails instead of
-//! segfaulting the process.
```
ivoanjo, line 15: Minor: This comment was kinda nice -- it explained why the reader looks like it does (it's weird), and what constraints we were working with. I think it's worth finding it a new home.

ivoanjo (approved):
👍 LGTM -- I +1 Yann's notes and I've left some tiny notes of my own but this is basically good to go!

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: Yes. `create_pipe()` now constructs and returns the existing `CopyPipe` type directly instead of returning a positional triplet.

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: I documented the contract on `ProcessMemoryCopy::copy`: callers do not need to establish that the range is valid, mapped, initialized, or remains mapped; inaccessible memory must return `WouldBlock`; success returns exactly `len` bytes; and the copy is not an atomic snapshot.

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: Added module documentation explaining the fault-safe Unix pipe mechanism and how inaccessible source memory becomes `WouldBlock`.

libdd-library-config/src/otel_process_ctx/linux.rs
cataphract: I removed the deprecated Linux compatibility module and re-exports. `MAPPING_NAME` remains because the Linux implementation still uses it.

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: I changed the destination to `vec![0; len]`, removed the unsafe `set_len`, and now construct a bounds-checked mutable slice before calling `read`. I also documented the slice-bounds proof.

libdd-library-config/Cargo.toml
cataphract: The reader intentionally remains enabled by default in this first PR because this commit promises to preserve `main`’s behavior. It is removed from the defaults in the next stacked PR.

libdd-library-config/src/otel_process_ctx/reader.rs
cataphract: macOS and Windows also use pipes, so `PlatformCopyPipe` describes the common mechanism across all supported platforms. I kept the current non-generic reader structure.

libdd-library-config/src/otel_process_ctx/reader.rs
cataphract: All three supported implementations—Linux, macOS, and Windows—use a pipe, so the pipe terminology is intentional rather than Linux-specific. I kept `PipeCopyError`.

libdd-library-config/src/otel_process_ctx/reader.rs
cataphract: Done. The trait now documents the caller’s lack of pointer-validity obligations, the exact `WouldBlock` error for inaccessible memory, the all-or-error result, and the lack of snapshot atomicity.

libdd-library-config/src/otel_process_ctx/reader.rs
cataphract: `ProcessMemoryCopy` now requires `Send`, making each platform pipe prove that property. I also updated the reader’s unsafe `Send` justification so it only needs to account for the process-global `NonNull` address.

libdd-library-config/src/otel_process_ctx.rs
cataphract: Done. The comment now describes it simply as an end-to-end test using `ProcessContextSelfReader`, without making a uniqueness claim.

libdd-library-config/src/otel_process_ctx/linux/self_reader.rs
cataphract: Restored at the top of `reader.rs`, adjusted to distinguish the shared mapping-lifetime constraints from the platform-specific same-process copy mechanism. The Unix pipe details are documented separately in `copy_pipe_unix.rs`.

libdd-library-config/src/otel_process_ctx.rs
cataphract: Done. The fixture now includes `service.name` and `telemetry.sdk.language` resource attributes, plus `datadog.process_tags` as an extra attribute.

libdd-library-config/src/otel_process_ctx/reader.rs
cataphract: I kept the current layout. `reader.rs` with children under `reader/` is the modern idiomatic Rust layout; `reader/mod.rs` is the older, still-supported form.

libdd-library-config/src/otel_process_ctx/writer/linux.rs
cataphract: Updated. The comment now says that the descriptor is closed immediately on purpose and explicitly notes that this does not invalidate the mapping.

libdd-library-config/src/otel_process_ctx/reader.rs
yannham: Do you have reference for this being the recommended way? I'm genuinely curious. Though maybe I'm just... old :sweat_smile: (

libdd-library-config/src/otel_process_ctx/reader.rs
yannham: TBF, I still find that at this level of abstraction on the reader code, whether we use a pipe or process_read_vm or whatnot is an implementation detail at this level of abstraction, and that what matters is that it's a fault-less copy. It could be switched for a different mechanism in the future, for example. But it's really not important

libdd-library-config/src/otel_process_ctx/reader.rs
yannham: Nevermind I found it in the doc. TIL.

cataphract: The Rust Book's ["Alternate File Paths" section](https://doc.rust-lang.org/stable/book/ch07-05-separating-modules-into-different-files.html#alternate-file-paths) explicitly calls foo/mod.rs the older style and presents foo.rs with foo/bar.rs as the idiomatic layout. The [Rust 2018 Edition Guide](https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs) also explains the move away from mod.rs -- to avoid having many indistinguishable mod.rs tabs open.

PR #2230, opened 2026-07-12
fix(sidecar): don't double-encode file:// telemetry endpoints

cataphract:
## What changed

- Add `Config::set_endpoint_uri` for integrations that already hold a parsed URI.
- Use the parsed-URI path in sidecar session telemetry and crashtracker telemetry instead of stringifying and reparsing endpoints.
- Preserve the telemetry-owned `TelemetryEndpoint` primitive configuration API.
- Document the non-idempotent behavior of `parse_uri` for file, Unix socket, and Windows pipe schemes.
- Add regression coverage for parsed `file://` endpoints.

## Why

Sidecar session setup converted an already-parsed endpoint URI back to a string and passed it through `parse_uri` again. For `file://` endpoints, the path is encoded into the URI authority, so parsing the string representation a second time encoded the authority again. Telemetry then opened the wrong file path and the worker failure tore down the sidecar IPC connection.

The fix gives parsed-URI callers an explicit path while keeping the public configuration boundary introduced for telemetry consumers.

## Validation

- `cargo check -p libdd-telemetry -p libdd-crashtracker -p datadog-sidecar`
- `cargo test -p libdd-telemetry config::tests::test_config_set_parsed_file_uri_does_not_reencode_path -- --exact`
- Full workspace formatting and clippy, targeted telemetry/FFI/data-pipeline/sidecar/crashtracker tests, and telemetry FFI examples were run before rebasing onto `origin/main`.

### DIFF
--- a/datadog-sidecar/src/service/sidecar_server.rs
+++ b/datadog-sidecar/src/service/sidecar_server.rs
@@ -724,13 +724,14 @@ impl SidecarInterface for ConnectionSidecarHandler {
                 &config.endpoint,
             );
             cfg.set_endpoint(TelemetryEndpoint {
-                url: Some(endpoint.url.to_string()),
                 api_key: endpoint.api_key.as_deref().map(str::to_owned),
                 test_token: endpoint.test_token.as_deref().map(str::to_owned),
                 timeout_ms: endpoint.timeout_ms,
                 use_system_resolver: endpoint.use_system_resolver,
+                ..Default::default()
             })
             .ok();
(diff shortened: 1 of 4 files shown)


cataphract:
This fixes a regression introduced by commit `b3144c676b73e157f9d563903c01df016882e8c4` in #2152. That refactor made telemetry endpoint configuration stringify an already-parsed `Uri`; for `file://` endpoints, passing that string through `parse_uri` again double-encoded the path stored in the URI authority.

bwoebi (approved):
Ugly workaround, but makes sense.

PR #2232, opened 2026-07-13
fix(rc): Make Target fields available again after eaf5ad06

bwoebi:
Otherwise the Target as returned by RemoteConfigManager cannot be read.

### DIFF
--- a/libdd-remote-config/src/lib.rs
+++ b/libdd-remote-config/src/lib.rs
@@ -51,6 +51,18 @@ impl Target {
             process_tags,
         }
     }
+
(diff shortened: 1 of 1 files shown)


Leiyks (approved):

cataphract:
SHouldn't this then remove the `RemoteConfigState.service` added in https://github.com/DataDog/dd-trace-php/pull/4006

See https://github.com/DataDog/dd-trace-php/blob/6e0aefafa2b105aae0f40eba36b4fc89a9eb6213/components-rs/remote_config.rs#L788-L794

PR #2237, opened 2026-07-14
fix(library-config): update Linux process context

cataphract:
[PROF-15481](https://datadoghq.atlassian.net/browse/PROF-15481)

## What does this PR do?

This is **2/3** in the stacked process-context series and depends on #2228.

It adapts the Linux implementation after the code-organization layer:

- makes `MADV_DONTFORK` best-effort instead of treating it as a publishing failure;
- detects whether the process-context mapping is available after a fork, rather than relying on a stored publisher PID;
- avoids reading or unmapping inherited state when the mapping is not present in the child;
- updates the Unix copy-pipe implementation and comments to match the current behavior;
- changes the default feature set to the writer, while keeping the reader available as an opt-in feature.

## Why?

#2228 intentionally preserves the old Linux behavior while introducing the shared reader/writer structure. This PR applies the Linux behavior changes separately so they can be reviewed without being mixed with either the file reorganization or the macOS/Windows implementation.

## Impact

Linux keeps the same platform-agnostic process-context API, with corrected fork handling and the final feature defaults. This layer does not add macOS or Windows support.

## Validation

Validated independently at this commit with:

- `cargo check -p libdd-library-config --all-features`;
- `cargo check -p libdd-library-config-ffi`;
- `cargo nextest run -p libdd-library-config --all-features --no-fail-fast`;
- `cargo nextest run -p libdd-library-config --no-default-features --features process-context-reader --no-fail-fast`.

## Stack

1. #2228 — reorganize the Linux process context
2. **#2237 — update the Linux implementation**
3. #2238 — add macOS and Windows support

BREAKING CHANGE: `process-context-reader` is no longer enabled by default; downstream users that require it must enable the feature explicitly.


libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -77,7 +77,11 @@ impl ProcessMemoryCopy for CopyPipe {
                                 io::ErrorKind::WouldBlock,
                                 "process context memory was unmapped during read",
                             ),
-                            pipe_dirty: false,
+                            // actually false on Linux: if EFAULT is returned, nothing was written;
+                            // should anything have been written already we would get a short write
+                            // However, this is not the case for macOS, despite what its manual
+                            // says: See https://github.com/apple-oss-distributions/xnu/blob/5c306bec31e314fa4d8bbdafb2f6f5a6b7e7b291/bsd/man/man2/write.2#L168-L186
+                            pipe_dirty: true,
```
yannham: Should it then be `cfg`-gated, instead of using `true` for both platforms? Or maybe we don't really care, because putting `true` for Linux just causes the pipe to be reinitialized which is fine?

libdd-library-config/src/otel_process_ctx/writer/linux.rs
```diff
@@ -63,7 +60,7 @@ impl super::HeaderMemoryHolder for MemMapping {
     }
 
     fn after_fork(self) {
```
yannham: What's the use-case for `after_fork`? If it's the same as `drop`, and `drop` is called automatically anyway when we exit a scope, it looks like it's a no-op

libdd-library-config/src/otel_process_ctx/writer.rs
```diff
@@ -111,12 +112,12 @@ impl<M: HeaderMemoryHolder, T: MonotonicTime> ProcessContextHandleGen<M, T> {
         unsafe {
             ptr::addr_of_mut!((*header).signature).write(*SIGNATURE);
             ptr::addr_of_mut!((*header).version).write(PROCESS_CTX_VERSION);
-            (*header)
-                .payload_ptr
-                .store(payload.as_ptr().cast_mut(), Ordering::Relaxed);
             (*header)
                 .payload_size
                 .store(payload_size, Ordering::Relaxed);
+            (*header)
```
yannham, line 114: Any reason for the re-ordering here? Not that it's important, since both accesses are `Relaxed`, this shouldn't change anything from the point of view of the reader.

libdd-library-config/src/otel_process_ctx/writer.rs
yannham: I wonder if all the comment and the `local_handler` dance is relevant anymore. Now that you've pushed the check to the `unmap` implementation, I think we can just use `swap` and call it a day.

yannham (approved):
Thanks for stress-testing and fixing the `fork` case! Mostly non-blocking comments. If  I understand correctly, if `MADV_DONTFORK` isn't supported and a fork happens, we don't have to do anything special because `update`, `publish` or `unmap` would touch the copy-on-write copy of the child. The only issue being that for some small time interval the child will appear as having the same context as its parent, instead of its own (which is why we are using `MADV_DONTFORK` in the first place). Is that more or less correct?

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
ivoanjo: Yeah I'd go for one simpler approach that works for both, this code is "weird enough" as-is ;)

libdd-library-config/src/otel_process_ctx/writer/linux.rs
```diff
@@ -19,12 +19,14 @@ use std::{
...
-///   `mapping_size()` and hasn't been unmmaped since.
+/// - when `only_for_pid` allows access from the current process, `start_addr` points to a live
+///   mapping of `mapping_size()` bytes created by `mmap`.
 /// - once `self` has been dropped, no memory access must be performed on the memory previously
-///   pointed to by `start`.
+///   pointed to by `start_addr`.
 pub(super) struct MemMapping {
     start_addr: NonNull<c_void>,
+    /// `Some(pid)` when `MADV_DONTFORK` succeeded, otherwise `None`.
+    only_for_pid: Option<u32>,
```
ivoanjo: Wait, when can `MADV_DONTFORK` fail on Linux? It's not clear from the comments/PR description why we're treating it as might-fail now.

In particular, the spec and other implementations are requiring `MADV_DONTFORK`; this is not to say that the spec is correct -- perhaps there's a good reason to change this, which is why I'm asking.

ivoanjo (changes requested):
I'd like to know more about the `MADV_DONTFORK` change before we apply it 👀

libdd-library-config/src/otel_process_ctx/writer/linux.rs
cataphract: Right, it's not needed anymore since the pid was moved to the `writer::linux::MemMapping` and the drop implementation became fork-safe.

libdd-library-config/src/otel_process_ctx/writer.rs
cataphract: I'll restore the previous order to avoid unnecessary churn

libdd-library-config/src/otel_process_ctx/writer.rs
cataphract: Yes, you're right (see also: `after_fork`)

libdd-library-config/src/otel_process_ctx/writer/linux.rs
cataphract: There are a few situations. Letting codex inspect the kernel source tree I got:

  - EINVAL: addr is not page-aligned, the rounded length overflows, or the requested subrange requires splitting a VMA that cannot be split—such as some
    special mappings, or a HugeTLB/dev-DAX mapping at an invalid boundary.

  - ENOMEM: some part of the range is unmapped or outside the process address space. Mapped portions are still processed.
  - EAGAIN: VMA splitting/merging could not allocate metadata, or the process reached vm.max_map_count. Internally these failures are ENOMEM, translated to
    EAGAIN (mm/vma.c:590, mm/madvise.c:1439).

  - EPERM: on current kernels, the range reaches a sealed, anonymous mapping that the caller cannot write—typically a read-only mseal()ed mapping
    (mm/madvise.c:1272, kernel documentation (https://docs.kernel.org/userspace-api/mseal.html)).

  - EINTR: interrupted while waiting for the mmap write lock (mm/madvise.c:1774).
  - ENOSYS: unusual kernel built without CONFIG_ADVISE_SYSCALLS.

We should prob retry on EINTR by the way

cataphract:
@yannham 

> If I understand correctly, if MADV_DONTFORK isn't supported and a fork happens, we don't have to do anything special because update, publish or unmap would touch the copy-on-write copy of the child. 

If it fails:
1) a reader opened in the parent will detect it's stale
2) an update requested through the writer will detect it's stale
3) a new reader will not detect the data is stale and will likely report stale data if it hasn't been updated in the child

If it doesn't fail (the normal case), the mapping will be gone so the new reader will report NotFound.

> The only issue being that for some small time interval the child will appear as having the same context as its parent, instead of its own (which is why we are using MADV_DONTFORK in the first place). Is that more or less correct?

Yes. If it succeeds, a new reader will not find the mapping before the child publishes.

libdd-library-config/src/otel_process_ctx/writer/linux.rs
ivoanjo: Wait but most of those reasons would mean we can't use our context for other things anyway?

My recommendation is to keep treating it as required, and fail the context publishing if we can't do this syscall, similar to how we fail publishing the context if some of the other syscalls fail.

Having a process fail to `MADV_DONTFORK` and then fork means keeping stale process for readers to find, and will be a _really_ hard to debug situation.

cataphract: Fair enough, I'll change it to fail for anything other than eintr

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: Yes, the optimized version would be to mark it non-dirty on linux to avoid an extra `pipe2` syscall. But an EFAULT should be rare case anyway, as it requires racing with the payload reclamation.

libdd-library-config/src/otel_process_ctx/writer/linux.rs
```diff
@@ -92,76 +83,107 @@ impl MemMapping {
...
+                unsafe {
+                    libc::madvise(
+                        mapping.start_addr.as_ptr(),
+                        super::mapping_size(),
+                        libc::MADV_DONTFORK,
+                    )
+                },
+                "madvise MADV_DONTFORK failed",
+            )
+        })?;
```
ivoanjo: Wait, the code seems correct but the comment is outdated I think?

libdd-library-config/src/otel_process_ctx/writer.rs
```diff
@@ -239,18 +249,10 @@ pub(super) fn publish_raw_payload(payload: Vec<u8>) -> io::Result<()> {
...
-            // normally, as it would try to unmap a region that isn't mapped anymore in the
-            // child process, or worse, could have been remapped to something else in the
-            // meantime.
-            //
-            // To do so, we get the old handler back in `local_handler` and prevent `mapping`
-            // from being dropped specifically.
-            swap(&mut local_handler, handler);
-            local_handler.mapping.after_fork();
+            let new_handler = ProcessContextHandleGen::publish(payload)?;
+            let _old_handler = replace(handler, new_handler);
```
ivoanjo, line 258: Minor: I don't think we need a 7 line comment, but now the logic is a bit cryptic e.g. you need to understand what `as_ptr().is_some()` does before understanding why we publish here, so a comment reminding readers would be useful.

libdd-library-config/src/otel_process_ctx/writer/linux.rs
cataphract: Yes, updated

libdd-library-config/src/otel_process_ctx/writer.rs
cataphract: I've added back a small comment

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -31,7 +32,7 @@ impl ProcessMemoryCopy for CopyPipe {
         create_pipe()
     }
 
-    fn copy(&self, addr: *const u8, len: usize) -> Result<Vec<u8>, PipeCopyError> {
+    fn copy(self, addr: *const u8, len: usize) -> (io::Result<Vec<u8>>, Option<Self>) {
```
ivoanjo, line 35: I'm curious (but not a blocker -- why move away from a `Result`?

ivoanjo (approved):
👍 LGTM

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: Before this commit, the caller had to behave like this:
1) if the result is Ok, then the pipe can be kept,
2) if it's an error, it has to inspect PipeCopyError's keep_pipe to decide

The commit gets rid of PipeCopyError:
1) as pointed out by another reviewer, it was a abstraction leak
2) by consuming self and only returning it back if the pipe doesn't need to be discarded, it eliminates the possibility that the caller forgets to check keep_pipe.

To use `Result`, it would have to be something like `Result<(Vec<u8>, Self), (Error, Option<Self>)>`, which is even more awkward to work with.

ivoanjo: Ah, I see, that's cool! I hadn't realized this "we take the pipe away from you if you can't have it anymore" was what's going on, nice!

PR #2238, opened 2026-07-14
feat(library-config): add macOS and Windows process context

cataphract:
## What does this PR do?

This is **3/3** in the stacked process-context series and depends on #2237.

It adds process-context publishing and same-process reading on macOS and Windows while reusing the shared protocol introduced by the preceding PRs:

- macOS exports `otel_process_ctx_v2` and discovers it with `dlsym`;
- Windows exports the same symbol and discovers it with `GetProcAddress`;
- adds platform-specific allocation, monotonic-clock, mapping-discovery, and safe memory-copy implementations;
- updates tracer-metadata publication and the FFI layer to use process contexts on all supported desktop platforms;
- adds reader-only CI coverage so the independently selectable feature configuration remains tested.

## Why?

The OpenTelemetry process-context discovery convention based on `/proc/self/maps` is Linux-specific. macOS and Windows therefore need an explicit exported-symbol discovery mechanism and platform-native memory-copy primitives, while the publication format and reader validation can remain shared.

Keeping this as the top layer isolates the new platform support from the Linux reorganization in #2228 and the Linux behavior updates in #2237.

## Impact

The platform-agnostic process-context API now works on Linux, macOS, and Windows. Linux behavior is unchanged by this layer. On macOS and Windows, tracer metadata can be published and read within the current process through the exported process-context symbol.

## Validation

Validated at this commit with:

- Linux all-feature and reader-only tests;
- Windows all-feature, reader-only, and FFI storage tests under Wine;
- targeted macOS tests;
- workspace formatting, Clippy, default/all-feature tests, and documentation tests.

## Stack

1. #2228 — reorganize the Linux process context
2. #2237 — update the Linux implementation
3. **#2238 — add macOS and Windows support**


libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -174,6 +174,64 @@ fn create_pipe() -> io::Result<(OwnedFd, OwnedFd, usize)> {
...
+    configure_fd(&write_fd)?;
+
+    // POSIX guarantees that an empty pipe accepts at least PIPE_BUF bytes without blocking.
+    // SAFETY: write_fd is a valid pipe descriptor.
+    let chunk_size = unsafe { libc::fpathconf(write_fd.as_raw_fd(), libc::_PC_PIPE_BUF) };
+    if chunk_size <= 0 {
+        return Err(last_error(
+            "failed to query process context copy pipe capacity",
+        ));
+    }
```
morrisonlevi, line 217: Making sure I understand here:

- Linux: use the full capacity of the pipe (`F_GETPIPE_SZ`) because we have it (Linux extension).
- macOS: use the atomically writeable size of the pipe  `_PC_PIPE_BUF`, as we do not have the full capacity of the pipe on this platform.

The practical effect is that on Linux, if we have larger contexts, we can copy it all in one go, whereas on macOS we'll have to loop with syscalls. Yes?

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
```diff
@@ -174,6 +174,64 @@ fn create_pipe() -> io::Result<(OwnedFd, OwnedFd, usize)> {
...
+
+    Ok((read_fd, write_fd, chunk_size as usize))
+}
+
+#[cfg(target_os = "macos")]
+fn configure_fd(fd: &OwnedFd) -> io::Result<()> {
```
morrisonlevi, line 223: This function is okay but I think making a `pipe2` polyfill would have come out shorter and kept more code on the same path/design.

libdd-library-config/src/otel_process_ctx/writer/macos.rs
```diff
@@ -0,0 +1,143 @@
...
+impl Drop for VmRegion {
+    fn drop(&mut self) {
+        let _ = self.unmap();
+    }
+}
+
+fn last_error(context: &'static str) -> io::Error {
+    let error = io::Error::last_os_error();
+    io::Error::new(error.kind(), format!("{context}: {error}"))
+}
```
morrisonlevi: There are at least 5 copies of this function, I'm going to consolidate this.

morrisonlevi (commented):
I've started to review. Note that I already pushed a commit to improve some safety comments. Probably going to focus on this PR as I can actually see the whole thing.

libdd-library-config/src/otel_process_ctx/reader/copy_pipe_unix.rs
cataphract: Yes, that's right.

cataphract:
Abandoned in light of later developments. See https://github.com/DataDog/libdatadog/pull/2272

PR #2240, opened 2026-07-15
feat: added file capability [APMSP-3780]

Aaalibaba42:
# What does this PR do?

Added Capability to read, write, and have metadata about files

# Motivation

dd-trace-js needing to add info from the `/proc` in their requests, it's better to have libdatadog drive the call

# Additional Notes

Nop

# How to test the change?

check, tests, check for wasm32


libdd-capabilities-impl/src/file.rs
```diff
@@ -0,0 +1,227 @@
...
+    async fn in_memory_mock_satisfies_trait() {
+        let fs = InMemoryFs::default();
+        fs.write("/proc/self/cgroup", Bytes::from_static(b"0::/"))
+            .await
+            .unwrap();
+        let out = fs.read("/proc/self/cgroup").await.unwrap();
+        assert_eq!(&out[..], b"0::/");
+        assert!(fs.exists("/proc/self/cgroup").await.unwrap());
+        assert!(!fs.exists("/nope").await.unwrap());
+    }
```
VianneyRuhlmann: What are you testing here ? I don't get why you need the InMemoryFs

VianneyRuhlmann (approved):

libdd-capabilities-impl/src/file.rs
Aaalibaba42: It's to show we can mock it for further testing, currently it's not used too much, but we can test things with this inmemoryfs

yannham: IMHO: if you don't need it right now, don't use it. Currently this tests is basically testing InMemoryFs, which is not the role for module tests for file capabilities. If we want to add more tests later, we can always add it then.

yannham (approved):

PR #2244, opened 2026-07-16
ci: avoid running out of space when miri workflow is run

hoolioh:
# What does this PR do?

Clean up unnecessary stuff from ubuntu runner in order to prevent running out of space when executing miri.

# Motivation

Lately this worflow [failed](https://github.com/DataDog/libdatadog/actions/runs/29132114832/job/86489319516)


Aaalibaba42 (approved):
It's sad that going through such hoops is necessary, github really sucks

PR #2247, opened 2026-07-17
docs: scope agent validation commands

pawelchcki:
Basically the default AGENTS.md cause a lot of full rebuilds on every checkpoint. The updated AGENTS.md teaches the agents to only rerun validations to actualy modified creates. 

At a risk of skipping some downstream crates. Which in either way will still be caught by the CI

## Summary

- scope the default Clippy and nextest commands to affected crates
- reserve workspace-wide build and test commands for repo-wide changes
- make doc tests conditional on documentation, example, or public API changes

## Why

The previous guidance required several exhaustive workspace checks for routine crate-level changes. This keeps validation proportional to the change while retaining workspace coverage for repo-wide work.

## Validation

- `git diff --check`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,7 +6,7 @@
 
 ### Validating after changes
 
-Iterate fastest with `cargo check -p <crate>` while editing; the full validation steps below are what should be green before declaring work done.
+Iterate fastest with `cargo check -p <crate>` while editing; validate each affected crate before declaring work done. Run workspace-wide checks only for repo-wide changes.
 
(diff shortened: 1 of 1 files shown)


pawelchcki:
Hey @codex will this changes improve iteration speed with local agents?

yannham (approved):
I wonder if an in-between could be to run tests on dependent crates as well. But it requires the agent to build the dependency graph (another way would be to use the same executable as the CI does). Anyway this can always be done later if the current changes are missing too many failures; for now I 👍 optimizing for iteration speed

pawelchcki:
Thanks @yannham for review!

PR #2251, opened 2026-07-17
feat(capabilities)!: Add streaming to http capabilities

bwoebi:
This allows moving live-debugger sender impl to use http-capabilities.

### DIFF
--- a/datadog-live-debugger/src/sender.rs
+++ b/datadog-live-debugger/src/sender.rs
@@ -7,8 +7,8 @@ use constcat::concat;
 use futures::future;
 use http::uri::PathAndQuery;
 use http::{Method, Uri};
-use http_body_util::BodyExt;
-use libdd_common::http_common;
+use libdd_capabilities::{BodySender, HttpClientCapability, ResponseFuture};
+use libdd_capabilities_impl::NativeHttpClient;
 use libdd_common::tag::Tag;
 use libdd_common::Endpoint;
 use libdd_data_pipeline::agent_info::schema::AgentInfoStruct;
@@ -19,6 +19,7 @@ use std::borrow::Cow;
 use std::fmt::Display;
 use std::hash::Hash;
 use std::str::FromStr;
+use std::time::Duration;
 use tokio::task::JoinHandle;
 use uuid::Uuid;
 
@@ -270,15 +271,16 @@ pub fn generate_tags(
 enum SenderFuture {
     #[default]
     Error,
-    Outstanding(http_common::ResponseFuture),
-    Submitted(JoinHandle<anyhow::Result<http_common::HttpResponse>>),
+    Outstanding(ResponseFuture),
+    Submitted(JoinHandle<anyhow::Result<http::Response<Bytes>>>),
 }
 
 pub struct PayloadSender {
     future: SenderFuture,
-    sender: http_common::Sender,
+    sender: BodySender,
     needs_boundary: bool,
     payloads: u32,
+    timeout_ms: u64,
 }
 
 const BOUNDARY: &str = "------------------------44617461646f67";
@@ -290,6 +292,20 @@ impl PayloadSender {
         debugger_type: DebuggerType,
         percent_encoded_tags: &str,
     ) -> anyhow::Result<Self> {
+        Self::new_with_client(
+            config,
+            debugger_type,
+            percent_encoded_tags,
+            NativeHttpClient::new_client(),
+        )
+    }
+
+    pub fn new_with_client<C: HttpClientCapability + Send + 'static>(
+        config: &Config,
+        debugger_type: DebuggerType,
+        percent_encoded_tags: &str,
+        http_client: C,
+    ) -> anyhow::Result<Self> {
         let endpoint = match debugger_type {
             DebuggerType::Diagnostics => &config.diagnostics_endpoint,
             DebuggerType::Snapshots => &config.snapshots_endpoint,
@@ -297,7 +313,12 @@ impl PayloadSender {
         }
         .as_ref()
         .ok_or_else(|| anyhow::anyhow!("missing endpoint for {debugger_type:?}"))?;
-        Self::new_to_endpoint(endpoint, debugger_type, percent_encoded_tags)
+        Self::new_to_endpoint_with_client(
+            endpoint,
+            debugger_type,
+            percent_encoded_tags,
+            http_client,
+        )
     }
 
     /// Creates a sender targeting a specific endpoint. Used to fan a payload out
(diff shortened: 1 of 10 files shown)


libdd-capabilities/src/maybe_send.rs
```diff
@@ -43,3 +45,7 @@ pub trait MaybeSend {}
 
 #[cfg(target_arch = "wasm32")]
 impl<T> MaybeSend for T {}
+
+/// A `Future` that is additionally `MaybeSend`, avoiding the need for a `cfg` split.
+pub trait MaybeSendFuture<T>: Future<Output = T> + MaybeSend {}
+impl<T, F: Future<Output = T> + MaybeSend> MaybeSendFuture<T> for F {}
```
Aaalibaba42, line 51: Is that because of type alias with non auto-traits ? It's the only situation I've encountered when you still need cfg gates

libdd-capabilities/src/http.rs
```diff
@@ -24,11 +26,54 @@ pub enum HttpError {
     Other(anyhow::Error),
 }
 
+pub type ChunkFuture<'a> = Pin<Box<dyn MaybeSendFuture<Result<(), HttpError>> + 'a>>;
```
yannham, line 29: Do we really need to spell this type out manually? I think this would be a use-case for `async_trait` that we use elsewhere in libdatadog (IIRC, the runtime result will be the same, but you just wouldn't have to write those types like `ChunkFuture` manually)

libdd-capabilities/src/http.rs
```diff
@@ -24,11 +26,54 @@ pub enum HttpError {
...
+
 pub trait HttpClientCapability: Clone + std::fmt::Debug {
     fn new_client() -> Self;
 
     fn request(
         &self,
         req: http::Request<bytes::Bytes>,
     ) -> impl Future<Output = Result<http::Response<bytes::Bytes>, HttpError>> + MaybeSend;
+
+    fn request_streamed(&self, req: http::Request<()>) -> (BodySender, ResponseFuture)
```
yannham, line 66: Nit: add a small description.

libdd-http-client/src/backend/hyper_backend.rs
```diff
@@ -183,20 +183,27 @@ impl super::Backend for HyperBackend {
...
-        let status = response.status().as_u16();
-        let headers = collect_response_headers(&response)?;
-
-        let body_bytes = response
-            .into_body()
-            .collect()
-            .await
-            .map_err(|e| HttpClientError::IoError(e.to_string()))?
-            .to_bytes();
+        let (status, headers, body_bytes) = tokio::time::timeout(timeout, async {
```
yannham, line 186: This change looks unrelated (or at least separate), since virtually no code is using libdd-http-client yet (beside Python is starting to do so). Is it to align with the behavior of `NativeHttpClient`? Also, shouldn't we also adapt the `reqwest` backend to behave the same?

libdd-capabilities/src/http.rs
Aaalibaba42, line 29: That's essentially a `MaybeSendBoxFuture`, right ? Do you think it'd be useful to have it in maybe_send.rs as `MaybeSend` and `MaybeSendFuture` ?

datadog-live-debugger/src/sender.rs
```diff
@@ -380,39 +414,43 @@ impl PayloadSender {
...
-                        anyhow::bail!(
-                            "Server did not accept debugger payload ({status}): {response_body}"
-                        );
-                    }
-                    Ok(self.payloads)
+            // Once the body is fully sent, bound the wait for the response headers and (if
+            // needed) the response body under a single timeout - a slow/stalled server must
+            // not be able to hang this indefinitely.
+            let result = tokio::time::timeout(Duration::from_millis(self.timeout_ms), async {
+                let response = future.await??;
```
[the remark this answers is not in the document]
yannham: I think the Codex assessment is correct, even if the issue hasn't been introduced by the PR but seems to be pre-existing. I suppose we'd have to call `abort` on timeout to make sure the task doesn't run in the background.

yannham (approved):
I initially thought it could be separated out of the capabilities impl, but indeed implementing efficient streaming requires http-backend-specific tools. Plus the default impl makes it ok for WASM that doesn't have to implement it for now 👍

Aaalibaba42 (approved):

libdd-capabilities/src/maybe_send.rs
bwoebi: Yes, you can't do `dyn Future<_> + MaybeSend`

libdd-http-client/src/backend/hyper_backend.rs
bwoebi: The reqwest backend already behaves this way, just that timeout is native to reqwest:
```
        if let Some(timeout) = request.timeout {
            builder = builder.timeout(timeout);
        }
```
This aligns it with reqwest, where the timeout is end-to-end.

libdd-capabilities/src/http.rs
bwoebi: I can't get the MaybeSend into the async_trait though, right? async_trait forces `Pin<Box<dyn Future<T> + Send + 'a>>`

bwoebi: I don't know... Let's add it once we have more usages? E.g. ChunkFuture and ResponseFuture are subtly different that the one has a lifetime and the other doesn't.

datadog-live-debugger/src/sender.rs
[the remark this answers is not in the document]
bwoebi: Yes, indeed. Had fixed it locally after codex complained, but forgot to commit.

libdd-capabilities/src/http.rs
yannham: In fact async trait is probably not even needed. What I mean instead of defining `ChunkFuture<'a>` explicitly, couldn't you just define `fn send_chunk(&mut self, data: bytes::Bytes) -> impl MaybeSendFuture<Result<(), HttpError>> + 'a`? Maybe that doesn't work for some reason, but just trying to understand if spelling this type out is really needed.

libdd-http-client/src/backend/hyper_backend.rs
yannham: Fair enough!

libdd-capabilities/src/http.rs
bwoebi: No, that's why async_trait exists in the first place, you can't do that on trait functions like that.

yannham: Ah, because `StreamingBodySender` needs to be returned and thus put in a box as a `dyn StreamingBodySender`, ack 👍 (because you can definitely use an `impl` in return position in a trait without `async_trait`, that's what `HttpClientCapability` does for `request`, but then you can't use as a `dyn Trait` anymore...)

PR #2252, opened 2026-07-17
fix(remote-config)!: Expose HttpClientCapability in remote config

bwoebi:
As a side-effect, this fixes the missing timeout on the remote config body awaiting.


libdd-capabilities-impl/src/http.rs
```diff
@@ -9,27 +9,47 @@ mod native {
...
         }
     }
 
+    impl NativeHttpClient {
+        /// Like [`HttpClientCapability::new_client`], but disables connection pooling.
+        ///
+        /// Intended for clients that issue requests on a fixed interval (e.g. remote
+        /// config polling): the agent's low keep-alive setting can close an idle
+        /// connection between polls, which turns a pooled/reused connection into
+        /// intermittent request failures.
```
yannham, line 42: I did something similar in libdd-http-client (disable pooling) because of similar comments in `Endpoint`, but after discussing with agent people, it seems there's no basis for this claim as of today. Might be worth checking if this whole periodic/connection pooling disabling business is really needed at all today.

libdd-capabilities-impl/src/http.rs
```diff
@@ -9,27 +9,47 @@ mod native {
...
+        pub fn new_periodic_client() -> Self {
```
yannham: Nit: I think calling it `periodic` is a bit too specific. At the level of an HTTP client, we don't really care why the caller disable connection pooling. IMHO we should be "lower" level in the description: what we do is that we disable connection pooling.

libdd-capabilities-impl/src/http.rs
```diff
@@ -39,7 +59,17 @@ mod native {
             req: http::Request<bytes::Bytes>,
         ) -> impl std::future::Future<Output = Result<http::Response<bytes::Bytes>, HttpError>> + MaybeSend
         {
-            let client = self.client.get_or_init(new_default_client).clone();
+            let periodic = self.periodic;
+            let client = self
+                .client
+                .get_or_init(|| {
+                    if periodic {
```
yannham: Nit: I wonder if only having `new()` for the client, and then `with_periodic(self) -> Self` or something builder-like wouldn't make the interface a bit lighter.

libdd-remote-config/src/fetch/single.rs
```diff
@@ -96,22 +106,28 @@ impl<S: FileStorage> SingleFetcher<S> {
...
+    pub fetcher: SingleFetcher<S, C>,
 }
 
-impl<S: FileStorage> SingleChangesFetcher<S>
+impl<S: FileStorage, C: HttpClientCapability> SingleChangesFetcher<S, C>
 where
     S::StoredFile: FilePath,
 {
-    pub fn new(sink: S, target: Target, runtime_id: String, options: ConfigOptions) -> Self {
+    pub fn with_client(
```
yannham: Nit: I would call that `new_with_client`. `with_client` sounds like it should have the signature `with_client(self, client: C) -> Self`

yannham (approved):

libdd-capabilities-impl/src/http.rs
bwoebi: So, what exactly would you propose as name then?

libdd-capabilities-impl/src/http.rs
bwoebi: I don't know, but feels safer to me?

libdd-capabilities-impl/src/http.rs
bwoebi: ... and the callers a bit heavier. I don't quite like this.

libdd-remote-config/src/fetch/single.rs
bwoebi: Agree, this went through some refactor, putting new back.

libdd-capabilities-impl/src/http.rs
yannham: something like `disable_connection_pooling` if you go for a builder API, or `new_no_connection_pooling` or something better in that style, which is more factual IMHO

libdd-capabilities-impl/src/http.rs
yannham: I don't think there's a huge difference between `new().with_periodic(true)` and `new_periodic()` (or whatever variation depending on the precise names you use). But it's a matter of taste and not very important, so do you as you feel is best 🙂

PR #2255, opened 2026-07-21
chore(profiling): allow user to customise default heap profiling sampling interval [PROF-15190]

scottgerring:
# What does this PR do?
Let's the user customise the sampling interval via either the gotter or allocator crates. This is done with a pair of free functions to reflect the lifecycle - i.e. this is a global change, and bolting it onto a field on e.g. the GlobalAllocator would feel weird. 

# Motivation
In the long term we're probably going to want to dynamically adjust the sampling interval so we can size it appropriately regardless of the workload's allocation patterns, but this is a bit more involved. In the short term, let's add a lever to override the default sampling interval. This will allow our users to set a workload-appropriate size by reflecting on allocation rate and desired sample count. For some of the sample workloads we have looked at so far we have seen allocation rate means above 40 MiB/s, which with a 512 KiB sample rate gives a sample every ~10ms, which is generally going to be _too much_. 

# Additional Notes

# How to test the change?


libdd-profiling-heap-sampler/src/tl_state.c
```diff
@@ -79,7 +79,16 @@ void dd_tl_state_init(void) {
     errno = saved_errno;
 }
 
+/* 64 KiB floor: intervals below this produce excessive overhead for
+ * negligible additional insight. The sampler's internal gap floor is 8
+ * bytes, but we clamp much higher here to protect callers from
+ * accidental misconfiguration. */
+#define DD_SAMPLING_INTERVAL_MIN (64u * 1024u)
```
r1viollet, line 86: 💭 having used ddprof as a valgrind alternative, having low values was sometimes useful.
Though I agree that preventing users from shooting themselves in the foot is probably higher priority 😄

r1viollet (approved):
LGTM
When we plug the auto-tuning of sampling distance we will have to think about how to document the behaviour. Though happy to think about that when we get there.

libdd-profiling-heap-sampler/src/tl_state.c
scottgerring: I figure lets dial this down when someone asks for a very low distance so we use that friction to capture the reasoning?

r1viollet:
removing from merge queue, just to let the fixes go in

PR #2256, opened 2026-07-21
test: stabilize flaky tests

r1viollet:
# What does this PR do?

Skips one ptrace thread-enumeration test under Miri.

Makes async test assertions wait for the expected request instead of relying on fixed sleeps.

Clears the macOS `target/panic-abort` cache before tests as a short-term workaround for stale cached build artifacts.

# Motivation

These flaky tests and the macOS cache issue are currently blocking builds.

# Additional Notes

The macOS cache cleanup is intended as a short-term unblocker. A separate draft PR moves panic-abort artifacts outside the cached target tree.

# How to test the change?

- `cargo +nightly-2026-02-08 fmt --all -- --check`
- `cargo test -q -p libdd-data-pipeline-ffi exporter_send_telemetry_test`
- `cargo test -q -p libdd-profiling exporter::exporter_manager::tests -- --nocapture`
- `cargo test -q -p libdd-common --features reqwest test_file_dump_captures_http_request -- --nocapture`
- Linux workspace: `cargo test -q -p libdd-data-pipeline-ffi exporter_send_telemetry_test` repeated 30 times
- Linux workspace: `RUSTUP_TOOLCHAIN=nightly-2026-02-08 MIRIFLAGS="-Zmiri-disable-isolation" cargo miri test -p libdd-crashtracker enumerate_discovers_spawned_thread -- --nocapture`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,6 +122,10 @@ jobs:
         with:
           cache-targets: true # cache build artifacts
           cache-bin: true # cache the ~/.cargo/bin directory
+      - name: "Clear macOS panic-abort cache"
+        if: runner.os == 'macOS'
+        shell: bash
+        run: rm -rf target/panic-abort
       - name: "Remove nextest CI report"
         shell: bash
         run: rm -rf target/nextest/ci/junit.xml
--- a/libdd-crashtracker/src/receiver/ptrace_collector.rs
+++ b/libdd-crashtracker/src/receiver/ptrace_collector.rs
@@ -557,6 +557,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(miri, ignore)]
     fn enumerate_discovers_spawned_thread() {
         let barrier = Arc::new(Barrier::new(2));
         let b = Arc::clone(&barrier);
--- a/libdd-common/src/dump_server.rs
+++ b/libdd-common/src/dump_server.rs
@@ -35,11 +35,18 @@ const HTTP_200_RESPONSE: &[u8] = b"HTTP/1.1 200 OK\r\nContent-Length: 0\r\n\r\n"
 pub fn spawn_dump_server(output_path: PathBuf) -> anyhow::Result<PathBuf> {
     use tokio::net::UnixListener;
 
-    // Create a temporary socket path with randomness to avoid collisions
-    // Retry if the path already exists (highly unlikely with 64-bit random IDs)
+    // Create a short temporary socket path with randomness to avoid collisions.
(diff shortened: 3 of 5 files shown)


r1viollet:
Link to failures:
https://github.com/DataDog/libdatadog/actions/runs/29804761537/job/88553008560
https://github.com/DataDog/libdatadog/actions/runs/29806802452/job/88558985923#step:8:3241

scottgerring (approved):

libdd-crashtracker/src/receiver/ptrace_collector.rs
```diff
@@ -557,6 +557,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(miri, ignore)]
```
yannham, line 560: I'm curious, why would miri behave differently wrt to flakiness? Isn't just a red herring (limiting the probability of hitting the issue if we run the test twice as less? Is it a new test that just fails under Miri consistently?

yannham (approved):
If this unblocks, let's merge this 👍

.github/workflows/test.yml
```diff
@@ -122,6 +122,10 @@ jobs:
         with:
           cache-targets: true # cache build artifacts
           cache-bin: true # cache the ~/.cargo/bin directory
+      - name: "Clear macOS panic-abort cache"
```
r1viollet, line 125: @gyuheon0h this is a workaround, I think we should not be caching this.

r1viollet:
More failures...
https://github.com/DataDog/libdatadog/actions/runs/29813190571/job/88578758174?pr=2256

libdd-crashtracker/src/receiver/ptrace_collector.rs
r1viollet: So I could not repro this. Clanker has theories about the test that could be an issue:

 For this specific test:

 ```rust
   enumerate_discovers_spawned_thread
 ```

 the test spawns a thread, receives its TID over a channel, then enumerates /proc/self/task and expects that TID to be present.

 The channel guarantees the child thread sent its TID, but it does not necessarily prove that Miri/host /proc view will expose that interpreted thread exactly like a normal native Linux thread at that point. Under Miri, threads are simulated/interpreted, while /proc
 is an OS interface. With -Zmiri-disable-isolation, Miri allows more host access, but that does not make every host-observable thread assumption robust.

r1viollet:
New failure: https://gitlab.ddbuild.io/DataDog/apm-reliability/libddprof-build/-/jobs/1877379843

PR #2263, opened 2026-07-22
feat(trace_exporter)!: add `restart_after_fork` configuration option [APMSP-3843]

VianneyRuhlmann:
# What does this PR do?

This adds a config option to the trace exporter to allow background workers to not be restarted in child. This used to be the default behavior and is now opt-in. This feature is only used in dd-trace-py so it will only be a breaking change there.

Note: This change also exposed that both the StatsExporter and OtlpStatsExporter were missing the `reset` implementation. It 's added as part of this PR.

# Motivation

In python, the TraceExporter is currently recreated on forks, this is a legacy of the old thread architecture in python. To avoid restarting the workers in the child just to shut them down when recreating, we skip the restart on the trace exporter background workers. We want to change this behavior in dd-trace-py (https://github.com/DataDog/dd-trace-py/pull/18929) as well as allow dd-trace-rb to use the workers in the intended way. Once dd-trace-py has transitioned this can be removed.

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-data-pipeline/src/otlp/metrics.rs
+++ b/libdd-data-pipeline/src/otlp/metrics.rs
@@ -8,6 +8,7 @@ use super::config::OtlpMetricsConfig;
 use super::exporter::{send_otlp_http, OTLP_MAX_RETRIES, OTLP_SHUTDOWN_MAX_RETRIES};
 use async_trait::async_trait;
 use libdd_capabilities::{HttpClientCapability, MaybeSend, SleepCapability};
+use libdd_common::MutexExt;
 use libdd_ddsketch::DDSketch;
 use libdd_shared_runtime::Worker;
 use libdd_trace_protobuf::pb;
@@ -257,8 +258,7 @@ impl<C: HttpClientCapability + SleepCapability> OtlpStatsExporter<C> {
     /// Flush the concentrator and export stats; returns `Ok(true)` if anything was sent.
     async fn send(&self, force_flush: bool, max_retries: u32) -> anyhow::Result<bool> {
         let buckets = {
-            #[allow(clippy::unwrap_used)]
-            let mut c = self.concentrator.lock().unwrap();
+            let mut c = self.concentrator.lock_or_panic();
             c.flush_with_otlp_exact(SystemTime::now(), force_flush)
         };
         if buckets.is_empty() {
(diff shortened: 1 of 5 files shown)


libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -344,6 +348,16 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
         self
     }
 
+    /// Sets whether background workers spawned by this exporter  are restarted after the process
+    /// `fork()`s. Defaults to `true`.
+    ///
+    /// Set to `false` if the trace exporter is recreated after the fork to avoid restarting workers
+    /// which are going to be discarded anyway.
+    pub fn set_restart_after_fork(&mut self, restart_after_fork: bool) -> &mut Self {
```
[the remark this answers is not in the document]
VianneyRuhlmann: This is meant to be used only by dd-trace-py so there's no need for ffis

libdd-data-pipeline/src/otlp/metrics.rs
```diff
@@ -301,6 +301,13 @@ impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Wor
         }
     }
 
+    fn reset(&mut self) {
+        let _ = self
+            .concentrator
+            .lock_or_panic()
+            .flush(SystemTime::now(), true);
```
yannham, line 308: Should the child flush the concentrator though? I would expect to leave the original concentrator entirely to the parent, and just re-initialize it here to something new.

yannham (approved):

yannham:
Oh, and I forget: how do we make sure we clean this once dd-trace-py doesn't need it anymore? Should we have KTLO ticket or sth?

libdd-data-pipeline/src/otlp/metrics.rs
VianneyRuhlmann: This effectively resets the concentrator state by force flushing, the buckets are dropped not sent. Re-initializing the concentrator from the stats exporter would require to store all the parameters required to create the concentrator. Since the concentrator config can be updated by the TraceExporter this also means we have to update these parameters and update the concentrator handle in the TraceExporter.

PR #2266, opened 2026-07-23
feat(profiling): add USDT semaphores to skip slow path around heap profile sampler

scottgerring:
When no profiler is attached, we should use USDT semaphores to skip the slow path. This was a helpful suggestion from the profiling SIG. 

Criterion benchmarks below (this is ofc without a profiler attached / USDT is a `NOP`)  - in the case where we are being loaded through a shared library and forced to use TLSDESC the cost saving is almost 40%! Said otherwise, almost half of our cost in a dynamic scenario is the TLS access itself.  The TLSDESC test involves a lot of arduous plumbing and I don't imagine we're going to touch this code much so i've opted out of comitting it. 

Some ARM64 results: 

| Benchmark | Description | No semaphore | With semaphore | Improvement |
|---|---|---|---|---|
| **Static TLS (Criterion)** | | | | |
| `sampler_only/fast_path/64` | Raw sampler calls only (no allocator), local-exec TLS | 3.33 ns | 2.82 ns | **-15%** |
| `sampled_noop_fast_path/64` | `SampledAllocator<NoopAlloc>` alloc+free, isolates sampler overhead | 3.04 ns | 2.71 ns | **-11%** |
| `sampled_system_fast_path/64` | `SampledAllocator<System>` alloc+free, real-world overhead | 9.52 ns | 9.61 ns | ~same |
| **Dynamic TLS (forcing usage through a `.so`)** | | | | |
| size=64 | Raw sampler calls through shared library, TLSDESC (dlopen/LD_PRELOAD scenario) | 6.96 ns | 4.30 ns | **-38%** |

**Cost of live heap**
While i'm here, i've added a test "sampled_noop_slow_path_force_attach", which lets uses an API i've added to flip the semaphore on in tests to measure the cost of the "real slow path". On ARM64 this yields 14.79 ns as the slow path  live heap cost.

# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-profiling-heap-allocator/benches/sampler_overhead.rs
+++ b/libdd-profiling-heap-allocator/benches/sampler_overhead.rs
@@ -17,7 +17,7 @@ mod linux_bench {
     use libdd_profiling_heap_allocator::SampledAllocator;
     use libdd_profiling_heap_sampler::{
         dd_allocation_created, dd_allocation_freed, dd_allocation_requested,
-        dd_tl_state_get_or_init,
+        dd_test_set_profiler_active, dd_tl_state_get_or_init,
     };
     use std::alloc::{GlobalAlloc, Layout, System};
     use std::hint::black_box;
@@ -210,6 +210,25 @@ mod linux_bench {
         group.finish();
     }
 
+    fn bench_sampled_noop_slow_path_force_attach(c: &mut Criterion) {
+        let alloc = SampledAllocator::new(NoopAllocator);
+        let mut group = c.benchmark_group("alloc_free/sampled_noop_slow_path_force_attach");
+        unsafe { dd_test_set_profiler_active(true) };
+        for &size in SIZES {
+            let layout = Layout::from_size_align(size, ALIGN).unwrap();
+            group.bench_with_input(BenchmarkId::from_parameter(size), &layout, |b, &layout| {
+                b.iter(|| unsafe {
+                    force_next_allocation_to_sample();
+                    let ptr = alloc.alloc(layout);
+                    black_box(ptr);
+                    alloc.dealloc(ptr, layout);
+                });
+            });
+        }
+        group.finish();
+        unsafe { dd_test_set_profiler_active(false) };
+    }
+
     fn bench_sampler_only_slow_path(c: &mut Criterion) {
         let mut group = c.benchmark_group("sampler_only/slow_path");
         for &size in SIZES {
@@ -235,6 +254,7 @@ mod linux_bench {
         bench_sampler_only,
         bench_sampled_system_slow_path,
         bench_sampled_noop_slow_path,
+        bench_sampled_noop_slow_path_force_attach,
         bench_sampler_only_slow_path,
     );
 } // mod linux_bench
--- a/libdd-profiling-heap-allocator/benches/sampler_overhead.rs
+++ b/libdd-profiling-heap-allocator/benches/sampler_overhead.rs
@@ -15,10 +15,7 @@ criterion::criterion_main!(linux_bench::benches);
 mod linux_bench {
     use criterion::{criterion_group, BenchmarkId, Criterion};
     use libdd_profiling_heap_allocator::SampledAllocator;
-    use libdd_profiling_heap_sampler::{
-        dd_allocation_created, dd_allocation_freed, dd_allocation_requested,
-        dd_test_set_profiler_active, dd_tl_state_get_or_init,
-    };
+    use libdd_profiling_heap_sampler::{dd_test_set_profiler_active, dd_tl_state_get_or_init};
     use std::alloc::{GlobalAlloc, Layout, System};
     use std::hint::black_box;
     use std::ptr;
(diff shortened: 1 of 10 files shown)


libdd-profiling-heap-sampler/src/lib.rs
```diff
@@ -52,6 +52,33 @@ pub fn set_default_sampling_distance(distance_bytes: u64) {
     unsafe { sys::dd_set_default_sampling_interval(distance_bytes) }
 }
 
+#[cfg(target_os = "linux")]
+extern "C" {
+    #[link_name = "dd_heap_profiler_attached"]
+    fn dd_heap_profiler_attached_ffi() -> bool;
+}
```
nsavoire: This duplicates bindgen declaration

libdd-profiling-heap-sampler/src/lib.rs
```diff
@@ -52,6 +52,33 @@ pub fn set_default_sampling_distance(distance_bytes: u64) {
...
+///
+/// This is a point-in-time read of the alloc probe's USDT semaphore. A profiler
+/// can attach or detach immediately after this returns. Use it as a
+/// diagnostic/readiness signal, not as a synchronization primitive.
+#[cfg(target_os = "linux")]
+#[inline]
+pub fn is_profiler_attached() -> bool {
+    // SAFETY: dd_heap_profiler_attached performs a single semaphore read and
+    // has no preconditions.
+    unsafe { dd_heap_profiler_attached_ffi() }
```
nsavoire:
```suggestion
    unsafe { sys::dd_heap_profiler_attached() }
```

libdd-profiling-heap-sampler/include/datadog/heap/allocation_requested.h
```diff
@@ -107,6 +113,9 @@ static inline __attribute__((always_inline, warn_unused_result))
 dd_alloc_req_t dd_allocation_requested(size_t size, size_t alignment) {
     dd_alloc_req_t out = { size, size, alignment, 0 };
 
+    // If no tracer is attached to ddheap:alloc, skip all sampling logic.
+    if (__builtin_expect(!USDT_IS_ACTIVE(ddheap, alloc), 1)) return out;
```
nsavoire: nit: `USDT_IS_ACTIVE` in a header can lead to unexpected behaviour if included in another translsation unit. It invokes `__usdt_sema_implicit`, which emits a .weak/.hidden zero-initialized definition into the including TU.

nsavoire: Using an explicit semaphore would at least cause a link error (`USDT_WITH_EXPLICIT_SEMA`).

libdd-profiling-heap-sampler/src/probes.c
```diff
@@ -38,3 +39,17 @@ void dd_probe_free(void *ptr) {
...
+bool dd_heap_profiler_attached(void) {
+    return USDT_IS_ACTIVE(ddheap, alloc);
+}
+
+void dd_test_set_profiler_active(bool active) {
+    /* The implicit semaphore symbol emitted by USDT_WITH_SEMA(ddheap, alloc, ...)
+     * is __usdt_sema_ddheap__alloc (see __usdt_sema_name macro in usdt.h).
+     * Declare it extern and poke its .active field directly - this is exactly
+     * what the kernel does when an eBPF program attaches to the probe. */
+    extern struct usdt_sema __usdt_sema_ddheap__alloc
```
nsavoire: Using an explicit semaphore would make the code easier to read here:
```
 USDT_SEMA(ddheap).active = active ? 1 : 0;
```

libdd-profiling-heap-sampler/src/probes.c
nsavoire: Maybe share the semaphore with `dd_probe_alloc` ? (not sure if this is useful)

libdd-profiling-heap-allocator/benches/sampler_overhead.rs
```diff
@@ -235,6 +254,7 @@ mod linux_bench {
         bench_sampler_only,
         bench_sampled_system_slow_path,
```
nsavoire: With the semaphore being off by default, slow/fast benches basically measure the same thing, right ?

scottgerring: There's these variants now:

   | Benchmark | Semaphore | Sampling forced? | What it measures |
   |---|---|---|---|
   | `sampler_only/fast_path` | off (default) | No, pinned to fast path | Sampler fast-path check cost in isolation |
   | `sampler_only/slow_path` | off (default) | Yes (`remaining_bytes=0`) | Slow path logic in isolation, probe skipped |
   | `sampled_noop_slow_path` | off (default) | Yes | Same slow path but through the full allocator wrapper |
   | `sampled_noop_slow_path_force_attach` | **on** | Yes | Slow path + probe/we make it to the NOP  |

   So `sampler_only/slow_path` and `sampled_noop_slow_path` are _almost_ capturing the same thing, but the former is serving as a regression
 benchmark for the sampler logic in isolation (no allocator overhead). The `force_attach` variant is the one that shows the delta when a
 profiler is actually listening.

   It took me a hot minute to re-grok the differences so I'll beef up the commenting a bit.

libdd-profiling-heap-sampler/include/datadog/heap/allocation_requested.h
scottgerring: TIL! how did you even realise this 😄 
Let me change to an explicit semaphore.

libdd-profiling-heap-sampler/src/probes.c
scottgerring: I _think_ in practice it doesn't make much of a difference. If you never sample an allocation you bail out here pretty quickly anyway (almost zero cost ARM, a little for unaligned pages on AMD64 to check).

I'm not convinced either way with this one (what's the downside of sharing a semaphore if any?) and propose we leave it for this change and revisit?

libdd-profiling-heap-allocator/benches/sampler_overhead.rs
scottgerring: Actually, let's do this:

* Set of tests with the sempahore forced on (fast path, fast path system allocator, slow path, slow path system allocator) 
* _ONE_ test with the semaphore off to provide a regression test around the short circuit

scottgerring: @nsavoire , here's what we've got now:

| Benchmark | Allocator | Semaphore | Path | What it measures |
|---|---|---|---|---|
| `alloc_free/system` | System (no sampler) | n/a | n/a | Baseline system allocator cost |
| `profiler_attached/fast_path_system` | SampledAllocator\<System\> | ON | Fast | Full realistic per-alloc overhead (not sampled) |
| `profiler_attached/fast_path_noop` | SampledAllocator\<NoopAllocator\> | ON | Fast | Sampler+wrapper overhead isolated from allocator |
| `profiler_attached/slow_path_system` | SampledAllocator\<System\> | ON | Slow | Full realistic cost when allocation is sampled |
| `profiler_attached/slow_path_noop` | SampledAllocator\<NoopAllocator\> | ON | Slow | Sampling overhead isolated from allocator |
| `no_profiler/short_circuit` | SampledAllocator\<System\> | OFF | Short-circuit | Regression: near-zero cost when no profiler attached |

nsavoire (approved):

PR #2267, opened 2026-07-23
feat(stats)!: string field length limit

Eldolfin:
# What does this PR do?

Implement the `String length limits` section of the cardinality limit RFC.

# Motivation
CSS spec compliance.

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-data-pipeline/src/agent_info/schema.rs
+++ b/libdd-data-pipeline/src/agent_info/schema.rs
@@ -24,7 +24,8 @@ pub struct AgentInfoStruct {
     /// List of available endpoints
     pub endpoints: Option<Vec<String>>,
     /// List of feature flags
-    pub feature_flags: Option<Vec<String>>,
+    #[serde(default)]
+    pub feature_flags: Vec<String>,
     pub client_drop_p0s: Option<bool>,
     pub span_meta_structs: Option<bool>,
     pub long_running_spans: Option<bool>,
--- a/libdd-data-pipeline/src/agent_info/schema.rs
+++ b/libdd-data-pipeline/src/agent_info/schema.rs
@@ -123,6 +123,16 @@ pub struct MemcachedObfuscationConfig {
     pub keep_command: bool,
 }
 
+impl AgentInfo {
+    /// Return true if the agent advertises support for the extended (15 KB) resource length limit.
+    pub fn is_big_resource_enabled(&self) -> bool {
+        self.info
+            .feature_flags
+            .iter()
+            .any(|flag| flag == "big_resource")
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::AgentInfoStruct;
--- a/libdd-data-pipeline/src/trace_exporter/stats.rs
+++ b/libdd-data-pipeline/src/trace_exporter/stats.rs
@@ -277,6 +277,13 @@ pub(crate) async fn handle_stats_enabled(
         let mut concentrator = stats_concentrator.lock_or_panic();
         concentrator.set_span_kinds(get_span_kinds_for_stats(agent_info));
         concentrator.set_peer_tags(agent_info.info.peer_tags.clone().unwrap_or_default());
+        concentrator.set_big_resource(
+            agent_info
+                .info
+                .feature_flags
+                .iter()
+                .any(|flag| flag == "big_resource"),
+        );
         #[cfg(feature = "stats-obfuscation")]
         update_obfuscation_config(agent_info, client_side_stats);
     } else {
--- a/libdd-data-pipeline/src/trace_exporter/stats.rs
+++ b/libdd-data-pipeline/src/trace_exporter/stats.rs
@@ -107,6 +107,15 @@ fn is_obfuscation_active(agent_info: &AgentInfo) -> bool {
         .is_some_and(|v| v >= 1 && v == SUPPORTED_OBFUSCATION_VERSION)
 }
 
+/// Return true if the agent advertises support for the extended (15 KB) resource length limit.
+fn is_big_resource_enabled(agent_info: &AgentInfo) -> bool {
+    agent_info
+        .info
+        .feature_flags
+        .iter()
+        .any(|flag| flag == "big_resource")
+}
+
 /// Get span kinds for stats computation with default fallback
 fn get_span_kinds_for_stats(agent_info: &Arc<AgentInfo>) -> Vec<String> {
     agent_info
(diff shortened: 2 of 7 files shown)


Eldolfin:
@codex review

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -277,6 +277,13 @@ pub(crate) async fn handle_stats_enabled(
         let mut concentrator = stats_concentrator.lock_or_panic();
         concentrator.set_span_kinds(get_span_kinds_for_stats(agent_info));
         concentrator.set_peer_tags(agent_info.info.peer_tags.clone().unwrap_or_default());
+        concentrator.set_big_resource(
+            agent_info
+                .info
+                .feature_flags
+                .iter()
+                .any(|flag| flag == "big_resource"),
+        );
```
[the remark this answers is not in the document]
Eldolfin: fixed in d5467cd

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -283,6 +291,13 @@ impl SpanConcentrator {
                 self.additional_metric_tag_keys.as_slice(),
             ),
         };
+        // Apply field truncation only when obfuscation was applied
+        #[cfg(feature = "stats-obfuscation")]
+        let mut agg_key = agg_key;
+        #[cfg(feature = "stats-obfuscation")]
+        if target_bucket.obfuscated {
+            agg_key.truncate(self.big_resource);
```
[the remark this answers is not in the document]
Eldolfin: Spec says otherwise

libdd-data-pipeline/src/agent_info/schema.rs
```diff
@@ -24,7 +24,8 @@ pub struct AgentInfoStruct {
     /// List of available endpoints
     pub endpoints: Option<Vec<String>>,
     /// List of feature flags
-    pub feature_flags: Option<Vec<String>>,
+    #[serde(default)]
+    pub feature_flags: Vec<String>,
```
yannham, line 28: Is this related to the PR? Also, `endpoints` above is still an `Option<Vec<String>>`. Is that ok?

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -107,6 +107,15 @@ fn is_obfuscation_active(agent_info: &AgentInfo) -> bool {
         .is_some_and(|v| v >= 1 && v == SUPPORTED_OBFUSCATION_VERSION)
 }
 
+/// Return true if the agent advertises support for the extended (15 KB) resource length limit.
+fn is_big_resource_enabled(agent_info: &AgentInfo) -> bool {
```
yannham: Not sure, but do you think this could be a method on `AgentInfo` instead? Seems to not be really related to `trace_exporter::stats` specifically

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -329,6 +329,30 @@ impl<'a> BorrowedAggregationKey<'a> {
...
+        self.fixed.span_type = slice_up_to(self.fixed.span_type, 100);
+    }
+}
+
+/// Truncate `s` to at most `max_len` bytes
+fn slice_up_to(s: &str, max_len: usize) -> &str {
+    if max_len >= s.len() {
+        return s;
+    }
+    let mut idx = max_len;
```
yannham, line 353:
```suggestion
    // TODO: use `floor_char_boundary` once our MSRV is 1.91 or higher
    // (https://doc.rust-lang.org/std/primitive.str.html#method.floor_char_boundary)
    let mut idx = max_len;
```

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -283,6 +291,13 @@ impl SpanConcentrator {
...
+        }
```
yannham, line 300:
```suggestion
        #[cfg(feature = "stats-obfuscation")]
        {
            let mut agg_key = agg_key;
            
            if target_bucket.obfuscated {
                agg_key.truncate(self.big_resource);
            }
        }
```

libdd-trace-stats/src/span_concentrator/tests.rs
```diff
@@ -2275,3 +2276,106 @@ fn test_normalize_additional_metric_tag_keys_limit() {
     assert_eq!(result, vec!["aaa", "bbb", "ccc", "ddd"]);
     assert_eq!(result.len(), 4);
 }
+
+/// Test that the SpanConcentrator truncates inserted span's fields after obfuscation is applied
+#[duplicate_item(
```
yannham: Is this just to make a matrix of test? What's the advantage over just making `big_resource` and `obfuscation` parameters of one general test function `my_test(big_resource: bool, obfuscation: bool)`, and then have 3 tests that just trivially call to this function with the expected parameters? It looks like a lot of magic and an additional dependency for something simple tbh (but maybe I'm misunderstanding what it does).

libdd-trace-stats/src/span_concentrator/mod.rs
Eldolfin: This doesn't works because `agg_key` then becomes shadowed in the block, so the outer one is not actually truncated when used just below in `target_bucket.insert`

libdd-data-pipeline/src/agent_info/schema.rs
Eldolfin: This is a pattern I used before previously for tags filter:
When we don't need to differentiate between a missing field or its default value, we might as well put 
```rust
#[serde(default)]
```
to avoid an extra `.is_some_and`/`.unwrap_or_default` when reading it.

Here for example, if the agent doesn't send its `feature_flags`, it's the same as if there was no feature flags enabled.

As for `endpoints`, it could also be refactored to have it but it would be unrelated to this PR.

libdd-trace-stats/src/span_concentrator/tests.rs
Eldolfin: I don't have a strong opinion on that but I find it more readable 🤷🏽‍♂️ .

Also note that `duplicate` is already used in other crates that `libdd-trace-stats` depends on, like `libdd-trace-obfuscation`.

VianneyRuhlmann: I'm not a fan of the duplicate syntax either and I also feel like this makes it a bit harder to read (no types on the parameters) but this is mostly a personal taset

VianneyRuhlmann (approved):

libdd-trace-stats/src/span_concentrator/tests.rs
Eldolfin: Ok given that there is a consensus that duplicate_item makes it worse so I removed it 👍🏽

PR #2268, opened 2026-07-23
feat(crashtracking)!: retrieve c assert message for linux when `__assert_fail` is dynamically loaded

gyuheon0h:
Stacked above: [feat(got-hook): add DT_HASH fallback, relocation type guard, and hook_symbol](https://github.com/DataDog/libdatadog/pull/2297)

[PROF-15482](https://datadoghq.atlassian.net/jira/software/c/projects/PROF/boards/11?selectedIssue=PROF-15482)

# What does this PR do?

When a process crashes due to a C assert() failure, the crash report now includes the assertion expression string (`Assertion failed: (x > 0), function foo, file bar.c, line 42`) in error.message, instead of just reporting a bare SIGABRT.

When a shared library calls an external function like `__assert_fail`, it doesn't jump directly to the target. Instead, it    jumps through the global offset table. Every shared library has its own GOT with its own slot for each external function it calls.                                 
We overwrites the slots at runtime for `__assert_fail` so calls are redirected to our hook function. My initial thought was to use `LD_PRELOAD` but this approach works no matter the load order.

During `crashtracker::init()`, `install_assert_hook()`
1. Resolves the original assertion handler
2. Enumerate loaded ELF objects
3. Inspect dynamic relocation tables
4. parses each object’s `PT_DYNAMIC `segment to locate its relocation tables
5. Patch references to __assert_fail

When a C `assert()` fails:
The hook captures the assertion expression, file name, line no, func name, and stores it through an atomic pointer:
1. The hook calls the original `__assert_fail()` implementation.
2. The original implementation calls `abort()`, which raises SIGABRT.
3. The crashtracker signal handler checks the stored assertion metadata
4. The assertion details are included in the resulting crash report.

Unfortunately, this doesn't work everywhere
This only works on ARM64 and AMD64 Linux in processes using glibc or musl runtimes where the assert fail function has been been dynamically linked                                     
                                                                                     
# Motivation

# Additional Notes

# How to test the change?

```
{
  "counters": {
    "profiler_collecting_sample": 1,
    "profiler_inactive": 0,
    "profiler_serializing": 0,
    "profiler_unwinding": 0
  },
  "data_schema_version": "1.8",
  "error": {
    "is_crash": true,
    "kind": "UnixSignal",
    "message": "Assertion failed: (test_value > 0), function trigger_c_assert, file src/c/trigger_assert.c, line 12.",
    "source_type": "Crashtracking",
    "stack": {
      "format": "Datadog Crashtracker 1.0",
      "frames": [
        {
          "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
          "build_id_type": "GNU",
          "column": 10,
          "file": "./nptl/pthread_kill.c",
          "file_type": "ELF",
          "function": "__GI___pthread_kill",
          "ip": "0x727c09d7b9fc",
          "line": 89,
          "module_base_address": "0x727c09ce5000",
          "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
          "relative_address": "0x00000000000969fc",
          "sp": "0x7ffc91dce470",
          "symbol_address": "0x727c09d7b8d0"
        },
        {
          "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
          "build_id_type": "GNU",
          "column": 6,
          "file": "./signal/../sysdeps/posix/raise.c",
          "file_type": "ELF",
          "function": "__GI_raise",
          "ip": "0x727c09d27476",
          "line": 27,
          "module_base_address": "0x727c09ce5000",
          "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
          "relative_address": "0x0000000000042476",
          "sp": "0x7ffc91dce530",
          "symbol_address": "0x727c09d27460"
        },
        {
          "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
          "build_id_type": "GNU",
          "column": 7,
          "file": "./stdlib/abort.c",
          "file_type": "ELF",
          "function": "__GI_abort",
          "ip": "0x727c09d0d7f3",
          "line": 81,
          "module_base_address": "0x727c09ce5000",
          "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
          "relative_address": "0x00000000000287f3",
          "sp": "0x7ffc91dce540",
          "symbol_address": "0x727c09d0d720"
        },
        {
          "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
          "build_id_type": "GNU",
          "column": 9,
          "file": "./intl/loadmsgcat.c",
          "file_type": "ELF",
          "function": "_nl_load_domain",
          "ip": "0x727c09d0d71b",
          "line": 1177,
          "module_base_address": "0x727c09ce5000",
          "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
          "relative_address": "0x000000000002871b",
          "sp": "0x7ffc91dce680"
        },
        {
          "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
          "build_id_type": "GNU",
          "comments": [
            "resolve_names failed with Couldn't symbolize 125877066264214: address not found in symbolization source"
          ],
          "file_type": "ELF",
          "ip": "0x727c09d1ee96",
          "module_base_address": "0x727c09ce5000",
          "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
          "relative_address": "0x0000000000039e96",
          "sp": "0x7ffc91dce6d0"
        },
        {
          "build_id": "a7ba23793697db2e70bfbe6648326398eed71003",
          "build_id_type": "GNU",
          "column": 5,
          "file": "/home/bits/go/src/github.com/DataDog/libdatadog/libdd-crashtracker/src/collector/assert_interceptor.rs",
          "file_type": "ELF",
          "function": "hook_assert_fail",
          "ip": "0x63a727372d46",
          "line": 92,
          "module_base_address": "0x63a727324000",
          "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
          "relative_address": "0x000000000004ed46",
          "sp": "0x7ffc91dce700"
        },
        {
          "build_id": "a7ba23793697db2e70bfbe6648326398eed71003",
          "build_id_type": "GNU",
          "file_type": "ELF",
          "function": "trigger_c_assert",
          "ip": "0x63a7273b319b",
          "module_base_address": "0x63a727324000",
          "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
          "relative_address": "0x000000000008f19b",
          "sp": "0x7ffc91dce7e0"
        },
        {
          "build_id": "a7ba23793697db2e70bfbe6648326398eed71003",
          "build_id_type": "GNU",
          "column": 5,
          "file": "/home/bits/go/src/github.com/DataDog/libdatadog/bin_tests/src/bin/crashtracker_bin_test.rs",
          "file_type": "ELF",
          "function": "main",
          "ip": "0x63a727351f78",
          "line": 9,
          "module_base_address": "0x63a727324000",
          "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
          "relative_address": "0x000000000002df78",
          "sp": "0x7ffc91dce800"
        },
        {
          "build_id": "a7ba23793697db2e70bfbe6648326398eed71003",
          "build_id_type": "GNU",
          "column": 2,
          "file": "/home/bits/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs",
          "file_type": "ELF",
          "function": "__rust_begin_short_backtrace<fn() -> core::result::Result<(), anyhow::Error>, core::result::Result<(), anyhow::Error>>",
          "ip": "0x63a72735283b",
          "line": 172,
          "module_base_address": "0x63a727324000",
          "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
          "relative_address": "0x000000000002e83b",
          "sp": "0x7ffc91dcf000"
        },
        {
          "build_id": "a7ba23793697db2e70bfbe6648326398eed71003",
          "build_id_type": "GNU",
          "file_type": "ELF",
          "function": "main",
          "ip": "0x63a727353cd9",
          "module_base_address": "0x63a727324000",
          "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
          "relative_address": "0x000000000002fcd9",
          "sp": "0x7ffc91dcf010"
        },
        {
          "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
          "build_id_type": "GNU",
          "column": 16,
          "file": "./csu/../sysdeps/nptl/libc_start_call_main.h",
          "file_type": "ELF",
          "function": "__libc_start_call_main",
          "ip": "0x727c09d0ed90",
          "line": 58,
          "module_base_address": "0x727c09ce5000",
          "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
          "relative_address": "0x0000000000029d90",
          "sp": "0x7ffc91dcf110"
        },
        {
          "build_id": "4f7b0c955c3d81d7cac1501a2498b69d1d82bfe7",
          "build_id_type": "GNU",
          "column": 5,
          "file": "./csu/../csu/libc-start.c",
          "file_type": "ELF",
          "function": "__libc_start_main_impl",
          "ip": "0x727c09d0ee40",
          "line": 379,
          "module_base_address": "0x727c09ce5000",
          "path": "/usr/lib/x86_64-linux-gnu/libc.so.6",
          "relative_address": "0x0000000000029e40",
          "sp": "0x7ffc91dcf1b0",
          "symbol_address": "0x727c09d0edc0"
        },
        {
          "build_id": "a7ba23793697db2e70bfbe6648326398eed71003",
          "build_id_type": "GNU",
          "file_type": "ELF",
          "function": "_start",
          "ip": "0x63a727350b25",
          "module_base_address": "0x63a727324000",
          "path": "/home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
          "relative_address": "0x000000000002cb25",
          "sp": "0x7ffc91dcf200"
        }
      ],
      "incomplete": false
    },
    "thread_name": "crashtracker_bi"
  },
  "experimental": {},
  "files": {
    "/proc/self/maps": [
      "63a727324000-63a727350000 r--p 00000000 103:03 19137659                  /home/bits/go/src/github.com/DataDog/libdatadog/target/release/crashtracker_bin_test",
...
      "7ffc91dd6000-7ffc91dda000 r--p 00000000 00:00 0                          [vvar]",
      "7ffc91dda000-7ffc91ddc000 r-xp 00000000 00:00 0                          [vdso]",
      "ffffffffff600000-ffffffffff601000 --xp 00000000 00:00 0                  [vsyscall]",
      ""
    ]
  },
  "incomplete": false,
  "log_messages": [
    "Error resolving frames: normalize_ips: Ok(())\tresolve_names: Err(Failed to resolve names, see frame comments for details)"
  ],
  "metadata": {
    "family": "native",
    "library_name": "libdatadog",
    "library_version": "1.0.0",
    "tags": [
      "service:foo",
      "service_version:bar",
      "runtime-id:xyz",
      "language:native"
    ]
  },
  "os_info": {
    "architecture": "x86_64",
    "bitness": "64-bit",
    "os_type": "Ubuntu",
    "version": "22.4.0"
  },
  "proc_info": {
    "pid": 3004337,
    "tid": 3004337
  },
  "sig_info": {
    "si_code": -6,
    "si_code_human_readable": "SI_TKILL",
    "si_signo": 6,
    "si_signo_human_readable": "SIGABRT"
  },
  "timestamp": "2026-07-27 15:40:47.932376685 UTC",
  "ucontext": {
    "arch": "x86_64",
    "raw": ...,
    "registers": {
      "r10": "0x0000000000000008",
...
      "rsp": "0x00007ffc91dce470"
    }
  },
  "uuid": "be0eefee-dd24-4588-b377-a5c144dc1177"
}
```

[PROF-15482]: https://datadoghq.atlassian.net/browse/PROF-15482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gyuheon0h:
* **#2268** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2268?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2268?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

yannham:
Have yet to look at it, but as a preamble, do you think there's some way to reuse or share the code from libdd-profiling-heap-gotter? It seems some type definitions at least, and probably some functions as well are very similar, if not the same.

gyuheon0h:
> Have yet to look at it, but as a preamble, do you think there's some way to reuse or share the code from libdd-profiling-heap-gotter? It seems some type definitions at least, and probably some functions as well are very similar, if not the same.

@yannham Yeah. Maybe I shouldn't have opened this yet.

libdd-crashtracker/src/collector/assert_interceptor.rs
```diff
@@ -0,0 +1,201 @@
...
+        return;
+    }
+
+    let mut orig_addr: usize = 0;
+    // SAFETY: hook_assert_fail has the same signature as __assert_fail.
+    let patched = unsafe {
+        libdd_got_hook::hook_symbol(
+            c"__assert_fail",
+            hook_assert_fail as *const () as usize,
+            &mut orig_addr,
```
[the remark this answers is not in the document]
gyuheon0h: AFAIK, tracers do not unload libdatadog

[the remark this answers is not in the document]
scottgerring: > needs an unpatch/disable-on-unload strategy or the library must be made non-unloadable before patching external GOTs.

For some extra context here from the alloc tracking side - we intentionally cannot unpatch alloc tracking, and the original API design for the GOT stuff reflects that. 

The reason that is that in some cases we manipulate the pointers passed through to the underlying allocator to flag them as sampled allocations, which means that any allocations that _have_ been flagged have to go back through our patched functions on the `free` side. If they don't - because we remove the GOT override - the process will fault when the underlying allocator sees an address it has no clue about.

For crash tracking I don't think you have the same constraint? So you could, if you wanted, allow un-patching. You'd just have to do so in a way that doesn't allow it to happen for the memory profiling side, but that shouldn't be a big deal.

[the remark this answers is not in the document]
gyuheon0h: True, but I wonder if it is worth doing because AFAIK, there is no "unload libdatadog" for any of the tracer

libdd-crashtracker/src/collector/assert_interceptor.rs
```diff
@@ -0,0 +1,210 @@
...
+static ASSERT_MESSAGE: AtomicPtr<String> = AtomicPtr::new(ptr::null_mut());
+
+/// Store an assert-failure message for later retrieval by the signal handler.
+///
+/// Intentionally leaks the old message (if any): this runs on a path
+/// that leads immediately to `abort()`, so freeing memory is unnecessary
+/// and calling `free` in the signal handler is not async-signal-safe.
+fn store_assert_message(message: String) {
+    let new_ptr = Box::into_raw(Box::new(message));
+    let _ = ASSERT_MESSAGE.swap(new_ptr, SeqCst);
```
yannham: (Honestly it's not very important but it was itching 😛 )

```suggestion
    let _ = ASSERT_MESSAGE.swap(new_ptr, Ordering::Release);
```

libdd-crashtracker/src/collector/assert_interceptor.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/// Atomically take the stored assert message pointer, leaving null.
+///
+/// Async-signal-safe (only an atomic swap). The caller borrows the
+/// returned pointer without reconstructing the `Box`, avoiding `free`
+/// inside the signal handler.
+pub(crate) fn take_assert_message_ptr() -> *mut String {
+    ASSERT_MESSAGE.swap(ptr::null_mut(), SeqCst)
```
yannham:
```suggestion
    ASSERT_MESSAGE.swap(ptr::null_mut(), Ordering::Acquire)
```

libdd-crashtracker/src/collector/assert_interceptor.rs
```diff
@@ -0,0 +1,210 @@
...
+    } else {
+        alloc::format!(
+            "Assertion failed: ({assertion}), function {function}, file {file}, line {line}."
+        )
+    }
+}
+
+/// # Safety
+/// `ptr` must be null or point to a valid null terminated C string.
+#[cfg(all(target_os = "linux", target_pointer_width = "64"))]
```
yannham: If the whole module is linux-only and 64-bits only, we should probably gate the whole module instead of repeating the `cfg`. Another solution, if you need the current module to still exist, is to put what's linux64 specific in a `linux64` submodule that is gated, and then do a `use linux64::*` (cfg-gated as well), so you have only two `cfg` gates.

libdd-crashtracker/src/collector/assert_interceptor.rs
```diff
@@ -0,0 +1,201 @@
...
+    file: *const libc::c_char,
+    line: libc::c_uint,
+    function: *const libc::c_char,
+) -> ! {
+    let assertion_str = unsafe { cstr_to_str(assertion, "<unknown>") };
+    let file_str = unsafe { cstr_to_str(file, "<unknown>") };
+    let function_str = unsafe { cstr_to_str(function, "") };
+
+    let message = format_assert_message(assertion_str, file_str, line, function_str);
+    store_assert_message(message);
```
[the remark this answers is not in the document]
yannham: This is a good point. Maybe we should have a static buffer of x bytes and truncate the message if it's too long?

libdd-crashtracker/src/collector/assert_interceptor.rs
```diff
@@ -0,0 +1,210 @@
...
+
+/// Install the `__assert_fail` GOT hook across all currently loaded
+/// libraries.
+///
+/// Safe to call multiple times; only the first call patches.
+///
+/// Only supported on 64-bit Linux. On other platforms this is a no-op.
+#[cfg(all(target_os = "linux", target_pointer_width = "64"))]
+pub(crate) fn install_assert_hook() {
+    if ORIG_ASSERT_FN.load(core::sync::atomic::Ordering::Acquire) != 0 {
```
yannham: Nit: a zero check doesn't need any special ordering constraint

```suggestion
    if ORIG_ASSERT_FN.load(core::sync::atomic::Ordering::Relaxed) != 0 {
```

libdd-crashtracker/src/collector/assert_interceptor.rs
```diff
@@ -0,0 +1,210 @@
...
+        return;
+    }
+
+    // SAFETY: hook_assert_fail has the same signature as __assert_fail.
+    let result = unsafe {
+        libdd_gotter::hook_symbol(c"__assert_fail", hook_assert_fail as *const () as usize)
+    };
+
+    if let Ok(hook) = result {
+        ORIG_ASSERT_FN.store(hook.orig_addr, core::sync::atomic::Ordering::Release);
```
yannham: Is it possible that two `install_assert_hook()` calls race? I suppose all of this is not _theoretically_ multithreaded, but it seems on paper nothing prevents a strange caller to have two threads racing to install the hook. In this case the hook might be installed two times, and the `ORIG_ASSERT` will randomly either be the actual original `assert_fail` or the other hook's wrapper. In the latter case I wonder if it can then cause an infinite loop.

It's rather theoretical, so I'm not entirely sure what to do about it. One way would just to be make it clear in `install_assert_hook`'s documentation that it's not thread-safe: it must not be called from different threads. Another simple check would be to see if `hook.orig_addr` is actually the address of our own `hook_assert_fail`, and skip the update in that case. Yet another solution would be to not use atomics but locks instead (because hook installation doesn't look perf sensitive, and a mutex should be uncontended and thus almost free in this case), so that we can lock the whole sequence "install hook; update orig_assert_fn pointer".

libdd-crashtracker/src/collector/crash_handler.rs
```diff
@@ -322,7 +322,16 @@ fn handle_posix_signal_impl(
     // Get the panic message pointer but don't dereference or deallocate in signal handler.
     // The collector child process will handle converting this to a String after forking.
     // Leak of the message pointer is ok here.
-    let message_ptr = PANIC_MESSAGE.swap(ptr::null_mut(), SeqCst);
+    let panic_message_ptr = PANIC_MESSAGE.swap(ptr::null_mut(), SeqCst);
```
yannham: (Actually it depends on what the setter of the panic message does, but this should be sufficient if the store is either SeqCst or Release)

```suggestion
    let panic_message_ptr = PANIC_MESSAGE.swap(ptr::null_mut(), Ordering::Acquire);
```

libdd-crashtracker/src/collector/assert_interceptor.rs
gyuheon0h: >Another simple check would be to see if hook.orig_addr is actually the address of our own hook_assert_fail, and skip the update in that case

I like this approach. We should also expliciotly recommend against calling concurrently across multiple threads

yannham (approved):

libdd-crashtracker/src/collector/assert_interceptor.rs
gyuheon0h: Ah. Because there is no need for global order if its just one producer one consumer

PR #2272, opened 2026-07-24
feat: add caller owned process/thread context storage

morrisonlevi:
# What does this PR do?

This PR supports the dd-trace-php PR needed to add Process and Thread contexts. This includes adjusting visibility and storage and a few similar things.

# Motivation

PHP needs to support allocating its own storage. It also supports these contexts on macOS and Windows even though it's non-standard (without discovery, just the in-process mechanism). So we needed to change some things.

# Additional Notes

I have reviewed the code just a tiny bit. I've been focused on the observable behaviors so far.

# How to test the change?

Mostly, this should be tested through PHP: https://github.com/DataDog/dd-trace-php/pull/4062.


morrisonlevi:
Closing for a different direction.

PR #2273, opened 2026-07-24
feat(data-pipeline): add fork-safe OTLP gRPC trace transport

bm1549:
# What does this PR do?

Adds a plaintext gRPC-over-HTTP/2 (h2c) trace-export transport for OTLP to `libdd-data-pipeline`. It's a custom `GrpcService` (`H2Service`) plugged into tonic's generic `Grpc<T>` client, so the exporter holds no persistent connection or background task: each send opens a fresh connection driven by per-request tasks that are torn down when the send completes.

This is the transport primitive only. The OTLP export path is wired to it in a follow-up PR that rebases on top of this one.

# Motivation

The earlier OTLP gRPC approach built a tonic `Channel`, which spawns a persistent background task. Because libdatadog rebuilds its runtime across `fork(2)`, that task dies silently after a fork, which forced a `Worker` plus `watch` channel to rebuild it. Making the transport per-request and task-free removes that machinery and is fork-safe by construction.

# Additional Notes

- `default-features = false` on tonic (no `transport`, `channel`, or `server`); a minimal prost codec replaces `ProstCodec`, which moved to the separate `tonic-prost` crate.
- Plaintext only: `https://` endpoints are rejected at build time; terminate TLS in a proxy in front of the endpoint if encryption is needed.
- Deferred to the wiring PR: retry parity with the HTTP path, default-port handling for portless endpoints, and preserving the gRPC status code in the error mapping.

# How to test the change?

- `cargo nextest run -p libdd-data-pipeline` covers unit tests plus in-process `h2` integration tests: a real unary send decoded by a mock gRPC server, and connection-refused, timeout, and post-connect-reset cases that all map to `TraceExporterError::Io`.
- Run `cargo clippy -p libdd-data-pipeline --all-targets -- -D warnings`, `cargo +nightly-2026-07-26 fmt --all -- --check`, and `cargo check -p libdd-data-pipeline --target wasm32-unknown-unknown --no-default-features`.


libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -85,6 +85,19 @@ pub struct OtlpTraceConfig {
     pub otel_trace_semantics_enabled: bool,
 }
 
+/// Per-request OTLP gRPC trace exporter configuration.
+// Not yet wired to the trace exporter's send loop; exercised by tests only.
+#[allow(dead_code)]
```
bm1549, line 90: To be wired in #2171

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,684 @@
...
+    pub(crate) config: OtlpGrpcTraceConfig,
+    origin: http::Uri,
+    service: H2Service,
+    /// Custom headers parsed to gRPC metadata once at build time; invalid entries are dropped.
+    metadata_headers: Vec<(AsciiMetadataKey, AsciiMetadataValue)>,
+}
+
+/// Validate a gRPC endpoint (plaintext `http://` only) and build the transport.
+// Not yet wired to the trace exporter's send loop; exercised by tests only.
+#[allow(dead_code)]
```
bm1549, line 214: To be wired in #2171

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,684 @@
...
+
+const GRPC_EXPORT_PATH: &str = "/opentelemetry.proto.collector.trace.v1.TraceService/Export";
+const CLIENT_COMPUTED_STATS_HEADER: &str = "datadog-client-computed-stats";
+
+type ExportCodec =
+    prost_codec::ProstCodecImpl<ExportTraceServiceRequest, ExportTraceServiceResponse>;
+
+/// Send one OTLP trace export request over gRPC. Bounds connect + RPC with a single timeout.
+// Not yet wired to the trace exporter's send loop; exercised by tests only.
+#[allow(dead_code)]
```
bm1549, line 297: To be wired in #2171

libdd-data-pipeline/src/otlp/config.rs
```diff
@@ -120,6 +133,19 @@ mod tests {
...
+        let cfg = OtlpGrpcTraceConfig {
+            headers: vec![("k".to_string(), "v".to_string())],
+            timeout: Duration::from_secs(3),
+            otel_trace_semantics_enabled: true,
+        };
+        let clone = cfg.clone();
+        assert_eq!(clone.headers, cfg.headers);
+        assert_eq!(clone.timeout, Duration::from_secs(3));
+        assert!(clone.otel_trace_semantics_enabled);
+    }
```
yannham: We're basically testing the builtin `Clone` derive implementation.

```suggestion
```

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,686 @@
...
+        fn default() -> Self {
+            Self(PhantomData)
+        }
+    }
+    impl<T: ProstMessage + Default> ProstDecoder<T> {
+        // See `ProstEncoder::encode_into`: generic over `Buf` so it is testable without tonic's
+        // private `DecodeBuf` constructor.
+        fn decode_from(src: &mut impl bytes::Buf) -> Result<Option<T>, Status> {
+            // copy_to_bytes drains the whole buffer even if the backing store is non-contiguous.
+            let buf = src.copy_to_bytes(src.remaining());
```
yannham: I'm not sure to understand the point of this line. Sure we get a contiguous buffer out of it, but we need to go through the original one anyways (and copy it). So why not read directly read from the original one?

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,686 @@
...
+                .await?;
+            // The Connection (request dispatcher) must be driven or send_request deadlocks.
+            // Spawn it ephemerally; it ends when `sender` drops at the close of this future.
+            let driver = tokio::spawn(async move {
+                let _ = conn.await;
+            });
+            let resp = sender.send_request(req).await?;
+            let (parts, incoming) = resp.into_parts();
+            let collected = incoming.collect().await?;
+            driver.abort();
```
yannham: It seems `driver.abort()` kinda ends the connection abruptly, instead of a graceful shutdown. On one hand, we already got our response, so we should be good on our side. I just wonder if this is an issue on the intake side.

One alternative is to await the driver after dropping the sender (`drop(sender); let _ = driver.await;`). This is a graceful shutdown. However this is now subject to network hangs or whatnot, so it probably needs a timeout.

I'm not 100% sure what's the best solution here. `abort` is not graceful shutdown, but on the other hand it ensures we don't block the calling task for unimportant reasons.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,686 @@
...
+}
+
+/// A gRPC transport for OTLP trace export: per-request config, request origin, and the dial
+/// service. Holds no live connection (nothing to rebuild across fork).
+#[derive(Clone, Debug)]
+pub(crate) struct OtlpGrpcTransport {
+    pub(crate) config: OtlpGrpcTraceConfig,
+    origin: http::Uri,
+    service: H2Service,
+    /// Custom headers parsed to gRPC metadata once at build time; invalid entries are dropped.
```
yannham: Nit: "invalid entries are dropped" sounds like it should go on the function that builds `metadata_headers`, not on the field itself.

```suggestion
    /// Custom headers parsed to gRPC metadata once at build time.
```

libdd-data-pipeline/src/otlp/grpc_exporter.rs
```diff
@@ -0,0 +1,686 @@
...
+            let (parts, incoming) = resp.into_parts();
+            let collected = incoming.collect().await?;
+            driver.abort();
+            Ok(http::Response::from_parts(parts, collected))
+        })
+    }
+}
+
+/// A gRPC transport for OTLP trace export: per-request config, request origin, and the dial
+/// service. Holds no live connection (nothing to rebuild across fork).
```
yannham: Nit: connections in itself is not necessarily the issue, but background async tasks that we don't control are.

```suggestion
/// service. Holds no live connection and thus no background task (nothing to rebuild across fork).
```

yannham (commented):
Overall LGTM, nothing major blocking.

Though the same caveat as in the original PR applies about the size issue:

> A first naive question: could/should this be feature-gated? I assume you'd want all tracers to support otlp out of the box, so the answers might be no, bu asking just > in case.
>
> A good test for the impact of the binary size increase: would you mind trying to  create a PR against dd-trace-py pointing to this PR's libdatadog ref, and see if it triggers the size gates for e.g. serverless?

libdd-data-pipeline/src/otlp/config.rs
mabdinur: Removed this redundant test; the derived `Clone` implementation doesn't need a behavioral test. Addressed in ad17de325.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
mabdinur: Agreed. `prost::Message::decode` accepts the existing `Buf`, so this now decodes directly without allocating or copying. Addressed in ad17de325.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
mabdinur: Reworked this so no detached driver task is spawned or aborted. The request and hyper connection are polled together; after the response completes and drops the sender, the connection is polled to completion for graceful shutdown. The existing exporter timeout bounds the whole operation. Addressed in ad17de325.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
mabdinur: Updated the field doc and kept the invalid-entry behavior documented at the build-time parsing site. Addressed in ad17de325.

libdd-data-pipeline/src/otlp/grpc_exporter.rs
mabdinur: Updated the doc to state that the transport holds neither a live connection nor a background task. Addressed in ad17de325.

mabdinur (approved):
Reviewed the OTLP gRPC transport and follow-up fixes. All review threads are resolved, local validation passes, and all CI checks are green on ad17de325.

yannham:
I am still worrying about the binary size explosion, which could trigger SLOs on the serverless side, as quoted in https://github.com/DataDog/libdatadog/pull/2273#pullrequestreview-4788985676. Has any of the suggestions been tried (either feature gating if possible, or trying the little experiment to see if it triggers a binary size SLO down the line)?

bm1549:
> I am still worrying about the binary size explosion, which could trigger SLOs on the serverless side, as quoted in [#2273 (review)](https://github.com/DataDog/libdatadog/pull/2273#pullrequestreview-4788985676). Has any of the suggestions been tried (either feature gating if possible, or trying the little experiment to see if it triggers a binary size SLO down the line)?

@yannham just confirmed that dd-trace-py is [still under the limit](https://mosaic.us1.ddbuild.io/change-request/3531313703708520147?taskId=gitlab&taskExecutionId=2000253517) after adding in this dependency https://github.com/DataDog/dd-trace-py/pull/19971

yannham (approved):
Great, thanks for checking. Review suggestion implementation LGTM

PR #2276, opened 2026-07-27
feat(sampling): OTel consistent-probability rv/th derivation (APMAPI-2181)

MilanGarnier:
Adds a public API to emit OpenTelemetry consistent-probability-sampling tracestate (`ot.th`/`ot.rv`). https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/

Part of [APMAPI-2170] (OTel consistent probability sampling across dd-trace-*).

## Changes

- `SamplingMechanism::is_probability()` — classifies rate-driven mechanisms
- `OtelConsistentSampling { rv, th }` (threshold, and rv as source of randomness derived from the trace id)
- `TraceRootSamplingInfo::otel_consistent_sampling(trace_id)` returns `Some(OtelConsistentSampling)` for a probability decision

## Tests

Tests from system-tests #7372
`cargo test -p libdd-sampling`

[APMAPI-2170]: https://datadoghq.atlassian.net/browse/APMAPI-2170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libdd-sampling/src/datadog_sampler.rs
```diff
@@ -210,13 +211,45 @@ fn format_sampling_rate(rate: f64) -> Option<String> {
...
+    pub fn otel_consistent_sampling<T: TraceIdLike>(
+        &self,
+        trace_id: &T,
+    ) -> Option<OtelConsistentSampling> {
+        if !self.mechanism.is_probability() || self.rl_effective_rate.is_some() {
+            return None;
+        }
+        let low64 = trace_id.to_u128() as u64;
+        let rv = (!low64.wrapping_mul(KNUTH_FACTOR)) >> 8;
+        let th = ((1.0 - self.rate) * (1u64 << 56) as f64).round() as u64;
```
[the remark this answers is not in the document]
MilanGarnier: Fixed in 132296c6b / rebased as 62673128c: `th` now clamped to `U56_MAX` (2^56-1), so rate=0.0 no longer overflows into a 57-bit value.

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -210,13 +211,45 @@ fn format_sampling_rate(rate: f64) -> Option<String> {
...
+    /// * a non-probability mechanism (manual keep, ASM, single-span, ...), or
+    /// * a probability mechanism whose keep was overturned by the trace rate limiter
+    ///   (`rl_effective_rate.is_some()`): the limiter, not the rate, is what dropped the trace, so
+    ///   per the RFC the threshold is erased while an inherited `rv` is still forwarded by the
+    ///   caller.
+    pub fn otel_consistent_sampling<T: TraceIdLike>(
+        &self,
+        trace_id: &T,
+    ) -> Option<OtelConsistentSampling> {
+        if !self.mechanism.is_probability() || self.rl_effective_rate.is_some() {
```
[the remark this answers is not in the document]
MilanGarnier: intermediary step to address this proposed in https://github.com/DataDog/libdatadog/pull/2288

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -210,13 +212,63 @@ fn format_sampling_rate(rate: f64) -> Option<String> {
...
+            return None;
+        }
+        let rv = derive_rv(trace_id);
+        let th = derive_th(self.rate);
+        let rv = reconcile_rv(rv, th, self.is_keep);
+        Some(OtelConsistentSampling { rv, th })
```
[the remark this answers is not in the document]
MilanGarnier: th must be retained on DROP too (matches OTEP 235) — only rate-limiter-overturned drops erase it; a plain probability-rule drop keeps th, so current Some return here is correct, not a bug

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -213,13 +215,63 @@ fn format_sampling_rate(rate: f64) -> Option<String> {
...
+/// Derives the 56-bit OTel `rv` from the low 64 bits of a trace id.
+fn derive_rv<T: TraceIdLike>(trace_id: &T) -> u64 {
+    let low64 = trace_id.to_u128() as u64;
+    (!low64.wrapping_mul(KNUTH_FACTOR)) >> 8
+}
+
+/// Derives the 56-bit OTel `th` (rejection threshold) from a DD sample rate.
+fn derive_th(rate: f64) -> u64 {
+    const U56_MOD: u64 = 1u64 << 56;
+    const U56_MAX: u64 = U56_MOD - 1;
```
bantonsson, line 243: I'm a bit confused here. The comment in the PRs says `computed in
exact u128 integer arithmetic`, but this is `u64` arithmetic.

MilanGarnier: Yes sorry for that, the description was stale but there's no need for `u128`, I wasn't sure about rounding issues at first

bantonsson (changes requested):
The PR description is messy and confusing. It says:
```
th = round(2^56 * (1 - rate_micros / 1_000_000)) (56-bit), on the rate rounded to 6 decimals so it
matches the RFC appendix's imprecision table rather than a float
(1.0 - rate) * 2^56
```

But the code does exactly `(1.0 - rate) * 2^56`, or am I missing something?

Also, the test code in https://github.com/DataDog/system-tests/pull/7372 does truncation, ie `int(...)` but this code does `.round()`.

Please clean up the description and change the code.

libdd-sampling/src/datadog_sampler.rs
```diff
@@ -213,13 +215,63 @@ fn format_sampling_rate(rate: f64) -> Option<String> {
...
+    (((1.0 - rate) * U56_MOD as f64).round() as u64).clamp(0, U56_MAX)
```
paullegranddc, line 244: Are you sure this is a good idea? 
f64 have 53 bits of precision, and you are doing 56 bits arithmetic. So when you cast back to u64, if `1.0 - rate` is bigger than 1/8th you will collapse the lower bits of your result together

bantonsson: I think that the system tests also suffer from this issue. This is the code that the OTel spec for this points to as an example https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/sampling/probability.go

MilanGarnier: The spec we used as a basis (and also for system tests) is https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/
The code you linked also uses round and then the part about adding precision I'm not exactly sure to understand why we would need this. 

About system-tests, although the comment says int(), the values are calculated with round() and the go tracer opted for round() as well https://github.com/DataDog/dd-trace-go/pull/5060/changes

MilanGarnier: About doing  `1.0 - rate`, I understand it would be better to do the operations in 56 bit arithmetic (converting rate to a 56 bit int before), this is just a different implem from what we went for with the go tracer and system tests already @paullegranddc

MilanGarnier: An alternative would be to do 
```rust
fn derive_th(rate: f64) -> u64 {
    const U56_MOD: u64 = 1u64 << 56;
    const U56_MAX: u64 = U56_MOD - 1;
    U56_MAX - ((U56_MOD as f64 * rate) as u64).clamp(0, U56_MAX)
}
```
This would not pass the current system tests in their current form
WDYT ?

bantonsson: The thing I'm most confused about is why the code and system tests don't do what the [spec](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/) says:

> OpenTelemetry SDKs are recommended to use 4 digits of precision by default.

Why do we compare the full hex representation and not only to 4 digits of precision?

Also, the comments in system tests must be very explicit (and correct) about how the values are computed and what precision is expected of the tracers.

MilanGarnier: > Also, the comments in system tests must be very explicit (and correct) about how the values are computed and what precision is expected of the tracers.

I'm changing the comment in system-tests: https://github.com/DataDog/system-tests/pull/7482
This is only one recommendation we are not following but it is technically respecting the spec. 

If it's needed to make small gains on the tracestate size one day, it will still be possible to implement it but since it's already implemented (and merged) this way in dd-trace-go, I prefer consistency

bantonsson: > I'm changing the comment in system-tests: [DataDog/system-tests#7482](https://github.com/DataDog/system-tests/pull/7482)

Thanks

> This is only one recommendation we are not following but it is technically respecting the spec.

Sure, it's optional.

bantonsson (approved):

PR #2278, opened 2026-07-27
feat(remote-config)!: add AsmRawResponseBody capability

estringana:
# What does this PR do?

Add AsmRawResponseBody (49) to RemoteConfigCapabilities to allow tracers to advertise support for raw HTTP response body collection to the WAF via remote config.

RFC-115

### DIFF
--- a/libdd-remote-config/src/lib.rs
+++ b/libdd-remote-config/src/lib.rs
@@ -79,6 +79,7 @@ impl Target {
     strum_macros::IntoStaticStr,
     strum_macros::Display,
 )]
+#[non_exhaustive]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
 pub enum RemoteConfigCapabilities {
     AsmActivation = 1,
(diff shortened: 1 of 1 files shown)


libdd-remote-config/src/lib.rs
```diff
@@ -79,6 +79,7 @@ impl Target {
     strum_macros::IntoStaticStr,
     strum_macros::Display,
 )]
+#[non_exhaustive]
```
estringana, line 82: I added this to fix the semver-check CI failure

yannham: I think the issue was rather the title (misplaced `!`), right? It's ok to have breaking changes, the CI just enforces that the title matches. Also I believe adding `#[non_exhaustive]` is technically a breaking change as well?

That being said, if this C-like enum is likely to be extended in the future, it looks like a good candidate for `#[non_exhaustive]`. This will make the next changes non-breaking. I'll let you decide if it's worth keeping or not (typically if this enum is expected to be very stable from now, maybe not worth it). Both options are reasonable.

yannham (approved):

cataphract (approved):

PR #2279, opened 2026-07-27
refactor(trace-utils)!: split generic TracerHeaderTags

anais-raison:
# What does this PR do?

Splits `TracerHeaderTags` into two parts: the tracer-identity strings (`lang`, `lang_version`, `tracer_version`, etc.) stay as-is, and the generic bool/int flags (`client_computed_stats`, `client_computed_top_level`, `dropped_p0_traces`, `dropped_p0_spans`) are pulled into a new standalone `TracerGenericTags` struct, embedded as `TracerHeaderTags::generic`.

# Motivation

`TracerHeaderTags` mixed tracer-identity strings with generic bool/int flags. The latter are meaningful independently of the string metadata, so pulling them into their own struct lets them be carried on their own in contexts where the string fields would be redundant (e.g. payload formats that already embed the tracer identity), without forcing a full split of the type.

# Additional Notes

Raised during review of #2156. Prepared as a separate PR to keep #2156 scoped to the sidecar-specific V04→V1 changes.

This PR alone is a no-op refactor — `TracerGenericTags` is still only used embedded in `TracerHeaderTags`. The actual payoff comes in #2156: the V1 trace-sending path already carries the tracer identity in the V1 payload itself, so `send_trace_v1` will be able to pass just `TracerGenericTags` (a small `Copy` struct, no serialization needed) across the sidecar's IPC boundary instead of the full `TracerHeaderTags`/`SerializedTracerHeaderTags`, which currently re-transmits redundant string data.

### DIFF
--- a/libdd-trace-utils/src/tracer_header_tags.rs
+++ b/libdd-trace-utils/src/tracer_header_tags.rs
@@ -19,14 +19,12 @@ macro_rules! parse_string_header {
     }
 }
 
-#[derive(Default, Debug, Serialize, Deserialize, Clone)]
-pub struct TracerHeaderTags<'a> {
-    pub lang: &'a str,
-    pub lang_version: &'a str,
-    pub lang_interpreter: &'a str,
-    pub lang_vendor: &'a str,
-    pub tracer_version: &'a str,
-    pub container_id: &'a str,
+/// The subset of tracer header tags that are plain booleans/integers, with no string/lifetime
+/// data. Unlike the rest of [`TracerHeaderTags`], these are meaningful regardless of whether the
+/// tracer payload format carries its own lang/version metadata (e.g. the V1 msgpack payload),
+/// so they can be transferred on their own without a dedicated serialized envelope.
+#[derive(Default, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq)]
+pub struct TracerGenericTags {
     // specifies that the client has marked top-level spans. Follows Go's isHeaderTrue rule:
     // absent/empty → false; "0"/"f"/"F"/"false"/"False"/"FALSE" → false;
     // every other non-empty value (including unparseable values like "yes") → true.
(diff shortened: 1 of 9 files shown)


bwoebi (approved):

libdd-trace-utils/src/tracer_header_tags.rs
```diff
@@ -19,14 +19,12 @@ macro_rules! parse_string_header {
...
-    pub lang_version: &'a str,
-    pub lang_interpreter: &'a str,
-    pub lang_vendor: &'a str,
-    pub tracer_version: &'a str,
-    pub container_id: &'a str,
+/// The subset of tracer header tags that are plain booleans/integers, with no string/lifetime
+/// data. Unlike the rest of [`TracerHeaderTags`], these are meaningful regardless of whether the
+/// tracer payload format carries its own lang/version metadata (e.g. the V1 msgpack payload),
+/// so they can be transferred on their own without a dedicated serialized envelope.
+#[derive(Default, Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq)]
```
yannham, line 26: If we're going to transmit them without serialization (if I understood correctly) through e.g. shared memory, does this struct need to be `repr(C)` to guarantee that whoever is reading it will have a stable layout? Or does the IPC mechanism already takes care of that in any way?

bwoebi: These are serialized and sent separately to SHM. So no, it doesn't need repr(C).

libdd-trace-utils/src/tracer_header_tags.rs
```diff
@@ -42,6 +40,17 @@ pub struct TracerHeaderTags<'a> {
...
 
+#[derive(Default, Debug, Serialize, Deserialize, Clone)]
+pub struct TracerHeaderTags<'a> {
+    pub lang: &'a str,
+    pub lang_version: &'a str,
+    pub lang_interpreter: &'a str,
+    pub lang_vendor: &'a str,
+    pub tracer_version: &'a str,
+    pub container_id: &'a str,
+    pub generic: TracerGenericTags,
```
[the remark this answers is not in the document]
matheus-souza: ✅ Fixed: Added LIMIT/OFFSET pagination to ListTopologyLinks and split graph node/edge fetching into ListActiveDevicesForGraph and ListActiveTopologyLinksForGraph with pagination loops in PgTopologyRepository.GetGraphData.

VianneyRuhlmann (approved):

libdd-trace-utils/src/tracer_header_tags.rs
yannham: Ok, thanks! I misread the comment as "those headers won't even need to be serialized at all".

yannham (approved):

PR #2280, opened 2026-07-27
feat: add v1 isolated bricks for trace exporter

anais-raison:
# What does this PR do?

Adds v1-native equivalents of existing v0.4 pipeline helpers, as groundwork for making `v1::Span`/`v1::TraceChunk` the exporter's canonical internal type:

- `trace_utils_v1::drop_chunks` and `TraceFilterer::filter_traces_v1` (libdd-trace-utils)
- `stats::add_spans_to_stats_v1` / `stats::process_traces_for_stats_v1` (libdd-data-pipeline)

# Motivation

Part of [APMSP-2812](https://datadoghq.atlassian.net/browse/APMSP-2812) : migration of the exporter from v0.4 to v1 with isolated bricks first before one final breaking PR that will handle the actual swap.

# Additional Notes

Pure addition, no behavior change — these functions aren't called anywhere yet, so there's no regression risk.

# State of the ticket 

## Done so far (in this PR)

- StatSpan implemented for v1::Span<T>
- trace_filter::Span implemented for v1::Span<T>
- get_root_span_index, compute_top_level_span (v1)
- drop_chunks v1 + filter_traces_v1
- add_spans_to_stats_v1 / process_traces_for_stats_v1

## Remaining (in a follow up PR)

- otlp_encoder::map_traces_to_otlp v1
- LogSpan/log_writer.rs v1
- agentless_encoder::encode_payload v1

## Swap PR (with breaking changes)

- Retype send_trace_chunks/send_trace_chunks_async to Vec<v1::TraceChunk<T>>
- Remove send/send_async and all msgpack decoding from TraceExporter
- Update TraceChunks (tracer_payload.rs) and trace_serializer.rs
- Actually wire process_traces_for_stats_v1/add_spans_to_stats_v1/filter_traces_v1 into the live pipeline (drop the #[allow(dead_code)] markers)
- Remove the v04::Span model and dead v0.4 code (might do this in a separate cleanup PR)


libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -360,6 +360,80 @@ pub(crate) fn process_traces_for_stats<
...
+    >,
+) {
+    let status = client_side_stats.load();
+    if let StatsComputationStatus::Enabled {
+        stats_concentrator, ..
+    } = &**status
+    {
+        let dropped_by_trace_filter = trace_filterer.filter_traces_v1(traces);
+        #[cfg(not(all(not(target_arch = "wasm32"), feature = "telemetry")))]
+        let _ = dropped_by_trace_filter;
```
ajgajg1134, line 426: Hmm, what's the point of this line here? 🤔

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -360,6 +360,80 @@ pub(crate) fn process_traces_for_stats<
...
+        stats_concentrator.add_span(span);
+    }
+}
+
+/// V1 counterpart of [`process_traces_for_stats`], operating on
+/// [`libdd_trace_utils::span::v1::TraceChunk`] chunks instead of flat v0.4 span vectors.
+// Not yet called from the live pipeline; will be wired in once the exporter's public API is
+// swapped to v1 native structs.
+#[allow(dead_code)]
+pub(crate) fn process_traces_for_stats_v1<T: libdd_trace_utils::span::TraceData>(
```
ajgajg1134: I know these functions are dead code but should there be any unit tests for these or do you want to include that in the next steps?

libdd-trace-stats/src/span_concentrator/stat_span.rs
```diff
@@ -93,6 +95,67 @@ impl<'a, T: TraceData> StatSpan<'a> for Span<T> {
...
+
+    fn is_partial_snapshot(&'a self) -> bool {
+        v1_trace_utils::is_partial_snapshot(self)
+    }
+
+    fn has_top_level(&'a self) -> bool {
+        v1_trace_utils::has_top_level(self)
+    }
+
+    fn get_meta(&'a self, key: &str) -> Option<&'a str> {
```
ajgajg1134, line 145: Are these get meta and metrics methods expected to go away once we move to v1 internally? I wouldn't expect to continue to see references to meta nor metrics once we have attributes

libdd-trace-utils/src/span/trace_utils_v1.rs
```diff
@@ -0,0 +1,460 @@
...
+
+//! Trace-utils functionalities implementation for V1 spans.
+
+use crate::span::trace_utils::DroppedP0Stats;
+use crate::span::v1::{AttributeValue, Span, TraceChunk};
+use crate::span::{SpanText, TraceData};
+use std::collections::HashSet;
+use tracing::debug;
+
+/// Span metric the mini agent must set for the backend to recognize top level span
```
ajgajg1134: mini agent?

libdd-trace-utils/src/span/trace_utils_v1.rs
```diff
@@ -0,0 +1,460 @@
...
+}
+
+/// Updates all the spans top-level attribute.
+/// A span is considered top-level if:
+///   - it's a root span
+///   - OR its parent is unknown (other part of the code, distributed trace)
+///   - OR its parent belongs to another service (in that case it's a "local root" being the highest
+///     ancestor of other spans belonging to this service and attached to it).
+pub fn compute_top_level_span<T: TraceData>(trace: &mut [Span<T>]) {
+    let mut span_id_idx: std::collections::HashMap<u64, usize> = std::collections::HashMap::new();
```
ajgajg1134: perf nit: Can we initialize this hash map with capacity of len(trace) to avoid re-sizing?

ajgajg1134: Or maybe instead it might read a little nicer to use something like `collect`?
```
    let span_id_idx: HashMap<u64, usize> = trace
        .iter()
        .enumerate()
        .map(|(i, span)| (span.span_id, i))
        .collect();
```

But maybe a rust person should double check me here

ajgajg1134 (changes requested):

libdd-trace-utils/src/span/trace_utils_v1.rs
```diff
@@ -0,0 +1,460 @@
...
+                if !(parent_span_service == &trace[span_idx].service) {
+                    // parent is not in the same service
+                    set_top_level_span(&mut trace[span_idx])
+                }
+            }
+            None => {
+                // span has no parent in chunk
+                set_top_level_span(&mut trace[span_idx])
+            }
+        }
```
Aaalibaba42: nit: just slightly easier to read on my eyes, nothing inherently wrong with what you did
```suggestion
        let span = &trace[span_idx];

        let is_top_level = span.parent_id == 0
            || span_id_idx
                .get(&span.parent_id)
                .is_none_or(|&parent_idx| trace[parent_idx].service != span.service);

        if is_top_level {
            set_top_level_span(&mut trace[span_idx]);
        }
```

libdd-trace-utils/src/span/trace_utils_v1.rs
```diff
@@ -0,0 +1,460 @@
...
+    match value {
+        AttributeValue::Float(v) => Some(*v),
+        AttributeValue::Int(v) => Some(*v as f64),
+        _ => None,
+    }
+}
+
+fn set_top_level_span<T: TraceData>(span: &mut Span<T>) {
+    span.attributes.insert(
+        T::Text::from_static_str(TOP_LEVEL_KEY),
```
Aaalibaba42, line 32: This could be built once as `Bytes` in `compute_top_level_span`, and cloned (ref count bump) into here, avoids allocating for each tls

libdd-trace-utils/src/span/trace_utils_v1.rs
```diff
@@ -0,0 +1,460 @@
...
+        dropped_p0_spans += chunk.spans.len() - sampled_indexes.len();
+        if sampled_indexes.is_empty() {
+            // If no spans were sampled we can drop the whole chunk
+            dropped_p0_traces += 1;
+            return false;
+        }
+        let sampled_spans = sampled_indexes
+            .iter()
+            .map(|i| std::mem::take(&mut chunk.spans[*i]))
+            .collect();
```
Aaalibaba42: This looks like a fancy and complicated `Vec::retain`

libdd-trace-utils/src/span/trace_utils_v1.rs
Aaalibaba42: This could even be rewritten as a two pass thing afterwards, which reads more idiomatically and lets us build the index map directly via collect:
```rs
    let span_id_idx: HashMap<u64, usize> = trace
        .iter()
        .enumerate()
        .map(|(i, span)| (span.span_id, i))
        .collect();
    let top_level: Vec<usize> = trace
        .iter()
        .enumerate()
        .filter(|(_, span)| {
            span.parent_id == 0
                || span_id_idx
                    .get(&span.parent_id)
                    .is_none_or(|&p| trace[p].service != span.service)
        })
        .map(|(i, _)| i)
        .collect();
    for i in top_level {
        set_top_level_span(&mut trace[i]);
    }
```

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -360,6 +360,80 @@ pub(crate) fn process_traces_for_stats<
...
+#[allow(dead_code)]
+fn add_spans_to_stats_v1<T: libdd_trace_utils::span::TraceData>(
+    stats_concentrator: &Mutex<SpanConcentrator>,
+    traces: &[libdd_trace_utils::span::v1::TraceChunk<T>],
+) {
+    let mut stats_concentrator = stats_concentrator.lock_or_panic();
+
+    let spans = traces.iter().flat_map(|chunk| chunk.spans.iter());
+    for span in spans {
+        stats_concentrator.add_span(span);
```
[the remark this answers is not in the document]
VianneyRuhlmann: I wonder what's the semantic of chunk attributes. Can any attribute currently stored in spans be moved to chunk  if they apply to all the spans i.e. if I'm looking for "key" on a span should I always also look for it on the chunk attributes ?

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -360,6 +360,80 @@ pub(crate) fn process_traces_for_stats<
...
+    }
+}
+
+/// V1 counterpart of [`process_traces_for_stats`], operating on
+/// [`libdd_trace_utils::span::v1::TraceChunk`] chunks instead of flat v0.4 span vectors.
+// Not yet called from the live pipeline; will be wired in once the exporter's public API is
+// swapped to v1 native structs.
```
VianneyRuhlmann, line 402: You should mark all these duplicate functions that need some cleanup once we move to v1 and link them to a jira ticket

libdd-trace-stats/src/span_concentrator/stat_span.rs
```diff
@@ -5,6 +5,8 @@
 //! support both trace-utils' Span and pb::Span.
 
 use libdd_trace_protobuf::pb;
+use libdd_trace_utils::span::trace_utils_v1 as v1_trace_utils;
```
VianneyRuhlmann: That looks quite confusing, is there any reason to do it ?

libdd-data-pipeline/src/trace_exporter/stats.rs
anais-raison: It's meant for clippy so it does not panick because `dropped_by_trace_filter` is used only with `wasm32`. I mirrored what was done in the `process_traces_for_stats` function (v04 version of the function).

libdd-trace-stats/src/span_concentrator/stat_span.rs
VianneyRuhlmann: Even then we will probably have to support the pb::Span (which are v04 like) for serverless. Maybe one solution that would also adress fields that were moved out of meta (e.g. span_kind) is to update the trait to provide explicit getters for each field e.g. `get_span_kind` instead of the generic `get_meta`

libdd-trace-utils/src/span/trace_utils_v1.rs
VianneyRuhlmann: I don't see how that would work, SpanText doesn't implement `clone`

libdd-trace-utils/src/span/trace_utils_v1.rs
VianneyRuhlmann: Just a note, this file is a duplicate of trace_utils for Span which is already a duplicate of trace_utils for pb::Span. This already quite confusing so they should be kept as similar as possible. If you end up doing some changes to it make sure to replicate it everywhere (and maybe in a separate PR).

libdd-trace-stats/src/span_concentrator/stat_span.rs
anais-raison: ~~Yup totally 👍
But it's true I could already add a `get_attributes` in this PR, don't think it'll break anything.~~

For now I'll leave `get_meta`/`get_metrics` as the generic accessors and not add a separate `get_attributes` cause it doesn't map cleanly onto v0.4/pb's split meta/metrics maps, and not all attribute keys are known at compile time (e.g. agent-configured peer_tag_keys), so a key-based lookup stays necessary either way.

Once v0.4 gets swapped out in favor of v1 as the canonical representation, we'll only have one span shape to support, and swapping to `get_attributes` at that moment we'll be cleaner and easier.

libdd-trace-utils/src/span/trace_utils_v1.rs
anais-raison: I mirrored the `trace_utils.rs` file that has the v04 functions, I'll take a deeper look at the comments.

anais-raison: Mini agent refers to the serverless mini agent component

libdd-data-pipeline/src/trace_exporter/stats.rs
anais-raison: Added unit tests for `add_spans_to_stats_v1` and a no-op test for `process_traces_for_stats_v1` when stats are disabled. I didn't cover the full Enabled pipeline path since it requires spawning a real `WorkerHandle` (the only precedent is the heavy `test_shutdown` integration test) cause it felt disproportionate for currently dead/unwired code. Happy to add it if you think it's needed.

libdd-trace-stats/src/span_concentrator/stat_span.rs
anais-raison: Yes, `trace_utils` (v04), `trace_utils_v1`, and the crate-level `trace_utils` (`pb::Span`) all expose functions with identical names, so aliasing was needed to avoid collisions. Renamed for consistency: `pb_utils` → `trace_utils_pb`, and dropped the unnecessary `v1_trace_utils` alias since `trace_utils_v1` is already the module's real name.

libdd-data-pipeline/src/trace_exporter/stats.rs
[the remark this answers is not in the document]
anais-raison: Good question, I decided to solve this by doing : span value wins, and falls back to the chunk's attributes if the span doesn't have its own value.

Added a `ChunkSpanView` wrapper in `stat_span.rs` that pairs a span with its chunk and implements the fallback for `get_meta` ans `get_metrics`. Applied the same pattern in `trace_filter.rs` for `filter_traces_v1`. Added tests covering both directions: span-level value wins, and chunk-level value is used when the span has none.

libdd-trace-utils/src/trace_filter.rs
```diff
@@ -229,6 +254,25 @@ impl TraceFilterer {
...
 
+    /// V1 counterpart of [`Self::filter_traces`]: removes chunks that fail filter checks
+    /// in-place. Returns the number of chunks dropped.
+    pub fn filter_traces_v1<T: TraceData>(&self, traces: &mut Vec<TraceChunk<T>>) -> usize {
+        let traces_count_before = traces.len();
+        traces.retain(|chunk| {
+            let Ok(root_span_index) = trace_utils_v1::get_root_span_index(&chunk.spans) else {
+                return true;
+            };
+            let should_drop = self.should_drop(&chunk.spans[root_span_index]);
```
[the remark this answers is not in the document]
anais-raison: Fixed : `filter_traces_v1` now evaluates the root span through a `ChunkSpanView` that falls back to `TraceChunk.attributes` when the span itself has no value for the tag, with span-level values still taking precedence.

ajgajg1134 (approved):
Looks like all my V1 related feedback was addressed :)

PR #2282, opened 2026-07-27
refactor(got-patching): extract shared GOT-patching primitives into libdd-got-hook

gyuheon0h:
Stacked under [feat(got-hook): add DT_HASH fallback, relocation type guard, and hook_symbol](https://github.com/DataDog/libdatadog/pull/2297)

# What does this PR do?

Moves the ELF GOT-patching infrastructure out of libdd-profiling-heap-gotter into a new shared crate libdd-got-hook, so multiple crates can reuse the same machinery for runtime function interposition.

This is a pure code move. I tried to have no behavioral changes here.

What moves to `libdd-got-hook`
- `DynamicInfo::from_phdr`: parse PT_DYNAMIC from a loaded ELF object
- `gnu_hash` / `gnu_hash_symbol_count` / `gnu_hash_lookup` / `check_sym`: GNU hash table utilities
- `iterate_libraries`: this is a  `dl_iterate_phdr` wrapper with panic-safe trampoline
- `PageProtGuard` / `read_proc_maps` / `MapEntry`: RELRO-aware page protection management
- `lookup_symbol` / `LookupResult`: used for resolving a symbol across all loaded objects
- `elf64_r_sym`: relocation info helper
-  ELF types and constants

What stays in `libdd-profiling-heap-gotter`
- `SymbolOverrides`: multi-symbol registry
- `hooks.rs`: the actual `malloc`/`free`/`calloc`/`realloc` hook functions
- lib.rs: `install_heap_overrides` / `update_heap_overrides` public API

# Motivation

# Additional Notes
Gated on @scottgerring's approval

# How to test the change?


gyuheon0h:
* **#2312** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2312?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2297** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2297?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2282** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2282?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2282?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

scottgerring:
Hey @gyuheon0h cool - this is a good idea!  We don't want to end up maintaining N variants of this sort of thing. 
Happy to review this but it will have to wait til next week; i'll regress it against all my stuff just to be extra sure before we merge and this week is hectic. For the new hashing modes, you'll want to make sure you test against real binaries linked in this fashion.

libdd-got-hook/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! ELF GOT-patching primitives for runtime function interposition.
+//!
+//! Walks each loaded ELF object via `dl_iterate_phdr`, parses its
+//! `PT_DYNAMIC` for the symbol/string/hash tables and the relocation
+//! arrays, and provides utilities to rewrite GOT entries.
+//!
+//! Scope:
+//! * 64-bit ELF only (`Elf64_*`). Other targets are compile-time gated.
```
yannham:
```suggestion
//! * 64-bit Linux ELF only (`Elf64_*`). Other targets are compile-time gated.
```

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,950 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
```
yannham, line 1: In order to better review, where's the big size difference between this file (950LoC) and the code removed in libdd-profiling-lib-gotter/src/elf.rc (500~ish LoC removed) coming from? From the PR description it looks like it's mostly moving code plus small differences, but here it's almost 2x times the code.

libdd-gotter/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[package]
+name = "libdd-got-hook"
+version = "1.0.0"
+description = "ELF GOT-patching primitives for runtime function interposition on 64-bit Linux."
+homepage = "https://github.com/DataDog/libdatadog/tree/main/libdd-got-hook"
+repository = "https://github.com/DataDog/libdatadog/tree/main/libdd-got-hook"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
```
yannham, line 12: Unless you need it to be published soon, it's a good default to start like this.

```suggestion
license.workspace = true
publish = false
```

libdd-profiling-heap-gotter/Cargo.toml
```diff
@@ -23,6 +23,9 @@ live-heap = ["libdd-profiling-heap-sampler/live-heap"]
 # used, not just that nothing crashed. Not for production use.
 test-support = []
 
+[target.'cfg(all(target_os = "linux", target_pointer_width = "64"))'.dependencies]
```
yannham: Just to follow the rest of this file's convention

```suggestion
[target.'cfg(target_os = "linux")'.dependencies]
```

libdd-got-hook/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+homepage = "https://github.com/DataDog/libdatadog/tree/main/libdd-got-hook"
+repository = "https://github.com/DataDog/libdatadog/tree/main/libdd-got-hook"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+
+[lib]
+bench = false
+
+[target.'cfg(all(target_os = "linux", target_pointer_width = "64"))'.dependencies]
```
yannham: If it's gated elsewhere, I don't think it's worth the trouble reproducing all the conditions here, as long as the dependency exists for Linux targets.

```suggestion
[target.'cfg(target_os = "linux")'.dependencies]
```

libdd-gotter/src/elf.rs
gyuheon0h: The 400 line growth is from
- new tests(`sysv` hash integration test with real `.so`, three `gnu_hash_symbol_count unit` tests,
  `test_all_loaded_libraries_have_valid_sym_count`)
- `hook_symbol` convenience function
- `elf64_r_type` + `is_got_pointer_reloc` + the relocation type constants
- accessor methods on `DynamicInfo` (`base_address()`, `rels()`, `relas()`, `jmprels()`)
- doc comments

yannham: Thanks! Will focus on those more specifically

libdd-gotter/Cargo.toml
scottgerring: It will need to be marked published so it doesn't break the libdd-profiling-heap- packages that now depend on it

scottgerring:
Nice - I wanted to do this! Will review first thing Monday 💪  Please don't merge beforehand as I'd like to double check all the allocation profiling stuff that rests on it.

libdd-gotter/src/elf.rs
scottgerring: @gyuheon0h , can you add some words to each of the changes that _isnt_ the refactoring so we know what problem you are addressing? Some short spiel in the description would be fine. 

FWIW for these sort of changes I find it helpful t to have separate commits with the mechanical part + the new part separate , but it might be a bit late for that (unless you are using `jj` :D )

gyuheon0h:
@yannham @scottgerring Ill just stack another PR on top of this with the new changes, so that this PR strictly only focuses on the initial "no new feature" refactor 👍

libdd-got-hook/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+repository = "https://github.com/DataDog/libdatadog/tree/main/libdd-got-hook"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+
+[lib]
+bench = false
+
+[dependencies]
+libc = "0.2"
```
yannham: (We've are moving most dependencies to workspace level)

```suggestion
libc.workspace = true
```

yannham (approved):

libdd-got-hook/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+# Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "libdd-got-hook"
```
scottgerring: Naming things is hard, but I wonder if we can think of a better one? It's inconsistent with the common pattern of the others `libdd-{domain}-{subdomain}` , and its inconsistent with `libdd-profiling-heap-**gotter**`. 

How about `libdd-gotter` ? We already prosecuted the "gotter" thing for the original crate.

libdd-gotter/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+# Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
```
scottgerring, line 1: Unrelated to this point in the file, but I think we also want a README.md with a basic TL;DR and maybe some example usage pointing at the heap profiling stuff? Speaking to @nsavoire , it would be great to have a big terrifying warning banner at the top, something like:

```
> [!WARNING]
> This library does runtime function interposition. This is a substantial intervention in a running process and it should only be used in cases where there is no other reasonable mechanism available. If you are not sure what other options might be available to you, reach out to the profiling team!

File: libdd-gotter/src/elf.rs [hunk 9]
```diff
@@ -0,0 +1,601 @@
...
+///
+/// Returns the `Elf64_Sym` entry if found and valid (non-zero value,
+/// function/object/notype binding).
+///
+/// # Safety
+/// `info` must have been produced by [`DynamicInfo::from_phdr`] for a
+/// currently-loaded ELF object.
+pub unsafe fn gnu_hash_lookup(info: &DynamicInfo, name: &[u8]) -> Option<Elf64_Sym> {
+    let hashtab = info.gnu_hash;
+    if hashtab.is_null() || info.gnu_hash_words < 4 {
```
scottgerring, line 312: Extra defensive check welcome!

libdd-profiling-heap-gotter/src/elf.rs
```diff
@@ -810,36 +301,23 @@ impl SymbolOverrides {
...
+        let r = r.unwrap();
+        assert!(r.address != 0);
+    }
 
-        assert_eq!(guard.original_prot(0x1000), Some(PROT_READ));
-        assert_eq!(guard.original_prot(0x1fff), Some(PROT_READ));
-        assert_eq!(guard.original_prot(0x2000), Some(PROT_READ | PROT_EXEC));
-        assert_eq!(guard.original_prot(0x3000), None);
+    #[test]
+    #[cfg_attr(miri, ignore)] // miri doesn't support dl_iterate_phdr
```
scottgerring: This is duped in libdd-got-hook now; suggest removing this one, but keeping the miri ignore comment to explain why it is ignored

libdd-profiling-heap-gotter/src/elf.rs
```diff
@@ -810,36 +301,23 @@ impl SymbolOverrides {
...
-                MapEntry {
-                    start: 0x2000,
-                    end: 0x3000,
-                    prot: PROT_READ | PROT_EXEC,
-                },
-            ],
-            touched: HashMap::new(),
-        };
+    #[cfg_attr(miri, ignore)]
+    fn can_lookup_malloc() {
```
scottgerring: This is duped in libdd-got-hook; suggest removing one or the other (in this case perhaps the _other_ one), just because its malloc ... 

Or rename it to "test_lookup_known_symbol", use malloc as an incidental detail, and chuck it in libdd-got-hook 🤷

libdd-got-hook/src/elf.rs
```diff
@@ -0,0 +1,601 @@
...
+        Some(CStr::from_ptr(self.strtab.add(off)))
+    }
+
+    /// The base load address of this ELF object.
+    pub fn base_address(&self) -> usize {
+        self.base_address
+    }
+
+    /// Access to REL relocations (pointer, count).
+    pub fn rels(&self) -> (*const Elf64_Rel, usize) {
```
scottgerring: I'm not sure the accessors here are adding much apart from indirection; this is just pushing the unsafe outwards.

I think something like this would be better as it adapts to a safe API and unburdens the caller, but I am also interested in what @yannham thinks:

```rust
pub unsafe fn rels(&self) -> Option<&[Elf64_Rel]>
```

.. basically if we're going to change this in the mechanical refactoring bit, we may as well deal with the null check and slice construction in the API itself.

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,601 @@
...
+        data: *mut c_void,
+    ) -> c_int {
+        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+            let ctx = &mut *(data as *mut Ctx);
+            let is_exe = ctx.is_first;
+            ctx.is_first = false;
+            (ctx.callback)(&*info, is_exe)
+        }));
+
+        // Never unwind a Rust panic through libc's dl_iterate_phdr callback.
```
scottgerring, line 415: I think the rest of the original comment here is important as it justifies swallowing the error: 

```suggestion
        // Never unwind a Rust panic through libc's dl_iterate_phdr callback.
       // Treat patching as best-effort and stop iteration on panic.

```

libdd-got-hook/src/elf.rs
```diff
@@ -0,0 +1,601 @@
...
+}
+
+impl Default for PageProtGuard {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Drop for PageProtGuard {
+    /// Restore every touched page to its original protection.
```
scottgerring: Again here i'd preserve the original comment as it establishes intent and indicates that the author has thought through the edge cases:

```suggestion
    /// Restore every touched page to its original protection. Runs on
    /// scope exit - including panic or early return - so page protections
    /// are never left weakened even if a patching pass bails out midway.
```

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,601 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Drop for PageProtGuard {
+    /// Restore every touched page to its original protection.
+    fn drop(&mut self) {
+        for (aligned, orig) in self.touched.drain() {
+            unsafe { mprotect(aligned as *mut c_void, self.page_size, orig) };
```
scottgerring, line 550: Let's keep the original comment:

```suggestion
            // Best-effort: nothing sensible to do on failure other than
            // leave the page RW, which is the pre-fix behavior.
            unsafe { mprotect(aligned as *mut c_void, self.page_size, orig) };
```

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,601 @@
...
+    if bloom_size == 0 || buckets_start > hashtab_words || chains_start > hashtab_words {
+        return None;
+    }
+    if nbuckets == 0 {
+        return None;
+    }
+
+    let buckets = core::slice::from_raw_parts(hashtab.add(buckets_start), nbuckets as usize);
+    let mut idx = *buckets.iter().max()?;
+    if idx == STN_UNDEF {
```
scottgerring, line 282: As i've spent a pile of time digging through "why is Symbol XYZ not being found?" issues, any hints in the code that suggest next steps when they are _not_ found is gold. Let's restore the comment so it is easier to debug this in the future!

```suggestion
        // All buckets empty: hash covers zero defined symbols, but the
        // symtab may still have undefined imports. Signal the caller to
        // use a fallback.
    if idx == STN_UNDEF {
```

scottgerring (commented):
Hey @gyuheon0h , thanks for taking the time to split this out and make it re-usable! 🙌 
Apart from a handful of nits, I think the two things to agree on are the name of the crate (some suggestions inline) and the shape of the API introduced around `DynamicInfo`

libdd-got-hook/src/elf.rs
yannham: Yeah, I agree a slice API would be nice. I guess we can even make it safe if the ELF ABI guarantees that a well-formed Elf file will have an array-like layout for `self.rels` (and check for nulls here).

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,601 @@
+// Copyright 2025-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! GOT-table interposition primitives.
+//!
+//! Scope:
+//! * 64-bit Linux ELF only (`Elf64_*`).
+//! * GNU hash tables only (`DT_GNU_HASH`). `DT_HASH` is not parsed; objects without a GNU hash
```
danielsn, line 8: do we have a way of tracking metrics on how often this happens?

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,601 @@
...
+use std::collections::HashMap;
+use std::io::{BufRead, BufReader};
+
+use libc::{
+    dl_iterate_phdr, dl_phdr_info, mprotect, sysconf, Elf64_Rel, Elf64_Rela, Elf64_Sym,
+    _SC_PAGESIZE, PROT_EXEC, PROT_READ, PROT_WRITE, PT_DYNAMIC, PT_LOAD,
+};
+
+// ELF dynamic-section tags. The `libc` crate doesn't export these
+// (they're processor-independent ELF spec constants). Values from `<elf.h>`.
```
danielsn, line 22: Do we have any sort of check that they stay in sync?

libdd-got-hook/Cargo.toml
gyuheon0h: Sure

libdd-gotter/README.md
```diff
@@ -0,0 +1,39 @@
+# libdd-gotter
+
+> [!WARNING]
+> This library does runtime function interposition by patching the Global Offset Table (GOT) of loaded ELF objects. This is a substantial intervention in a running process — it modifies function pointers that the dynamic linker has already resolved, affecting all code that calls through those GOT entries. Incorrect use can cause crashes, infinite recursion, heap corruption, or silent data loss. Understand the ELF dynamic linking model before using this crate.
+
+## What it does
+
+When a shared library calls an external function like `malloc`, it jumps through a pointer in its **Global Offset Table** -- a writable table that the dynamic linker fills at load time. This crate walks every loaded ELF object via `dl_iterate_phdr`, parses its `PT_DYNAMIC` segment, and rewrites GOT entries so calls are redirected to a hook function. The original function address is resolved and returned so the hook can forward to it.
```
danielsn, line 8: Is there good documentation on how ELF / GOT works that we can link here?

libdd-gotter/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Usage
+
+### Single-symbol hook (crashtracker intercepting `__assert_fail`)
+
+```rust
+use libdd_got_hook::hook_symbol;
+
+static ORIG_FN: AtomicUsize = AtomicUsize::new(0);
+
+unsafe extern "C" fn my_hook(/* same signature as target */) {
```
danielsn, line 19: Is there a best practice for ensuring this remains in sync / checking this at runtime?

libdd-gotter/README.md
```diff
@@ -0,0 +1,39 @@
...
+    // ... do work ...
+    // forward to original via ORIG_FN
+}
+
+let mut orig_addr: usize = 0;
+unsafe {
+    hook_symbol(c"__assert_fail", my_hook as *const () as usize, &mut orig_addr);
+}
+ORIG_FN.store(orig_addr, Ordering::Release);
```
danielsn, line 30: If `Release` is required, document why

libdd-gotter/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+### Multi-symbol registry (heap profiling hooking malloc/free/calloc/realloc)
+
+See [`libdd-profiling-heap-gotter`](../libdd-profiling-heap-gotter) which builds a `SymbolOverrides` registry on top of the primitives exported by this crate.
+
+## Platform support
+
+- **64-bit Linux (glibc)**: Full support.
```
danielsn: how do statically linked symbols work on glibc?

libdd-profiling-heap-gotter/src/elf.rs
```diff
@@ -729,26 +222,16 @@ impl SymbolOverrides {
...
         }
-        for slice_ptr_and_len in [
-            (dyn_info.relas, dyn_info.relas_count),
-            (dyn_info.jmprels, dyn_info.jmprels_count),
-        ] {
-            if slice_ptr_and_len.0.is_null() {
-                continue;
-            }
-            let relocs = std::slice::from_raw_parts(slice_ptr_and_len.0, slice_ptr_and_len.1);
+        for relocs in [dyn_info.relas(), dyn_info.jmprels()] {
```
danielsn, line 234: would it be clearer to have a helper function that gave `all_relocs`?

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,628 @@
...
+                DT_SYMTAB => symtab = correct(v) as *const Elf64_Sym,
+                DT_GNU_HASH => gnu_hash = correct(v) as *const u32,
+                DT_REL => rels = correct(v) as *const Elf64_Rel,
+                DT_RELA => relas = correct(v) as *const Elf64_Rela,
+                DT_JMPREL => jmprels = correct(v) as *const Elf64_Rela,
+                DT_RELSZ => rels_size = v as usize,
+                DT_RELASZ => relas_size = v as usize,
+                DT_PLTRELSZ => jmprels_size = v as usize,
+                DT_PLTREL => pltrel_type = v as i64,
+                _ => {}
```
danielsn, line 122: I get worried when I see a catchall, especially one that just ignores the unexpected value.  Why is this safe?

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,628 @@
...
+            unsafe { core::slice::from_raw_parts(self.rels, self.rels_count) }
+        }
+    }
+
+    /// RELA relocations for this object, or empty if none.
+    pub fn relas(&self) -> &[Elf64_Rela] {
+        if self.relas.is_null() || self.relas_count == 0 {
+            &[]
+        } else {
+            unsafe { core::slice::from_raw_parts(self.relas, self.relas_count) }
```
danielsn, line 223: safety comment

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,628 @@
...
+        }
+    }
+
+    /// JMPREL (PLT) relocations for this object, or empty if none.
+    pub fn jmprels(&self) -> &[Elf64_Rela] {
+        if self.jmprels.is_null() || self.jmprels_count == 0 {
+            &[]
+        } else {
+            unsafe { core::slice::from_raw_parts(self.jmprels, self.jmprels_count) }
```
danielsn, line 234: safety comment

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,628 @@
...
+/// `info` must have been produced by [`DynamicInfo::from_phdr`] for a
+/// currently-loaded ELF object.
+pub unsafe fn gnu_hash_lookup(info: &DynamicInfo, name: &[u8]) -> Option<Elf64_Sym> {
+    let hashtab = info.gnu_hash;
+    if hashtab.is_null() || info.gnu_hash_words < 4 {
+        return None;
+    }
+
+    let nbuckets = *hashtab;
+    let symbias = *hashtab.add(1);
```
danielsn, line 325: where are these offsets documented

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,628 @@
...
+    maps: Vec<MapEntry>,
+    // Aligned page base -> original prot flags read from /proc/self/maps.
+    touched: HashMap<usize, i32>,
+}
+
+impl PageProtGuard {
+    pub fn new() -> Self {
+        // sysconf can return -1 on error; fall back to a conservative
+        // 4 KiB default if the query fails.
+        let raw = unsafe { sysconf(_SC_PAGESIZE) };
```
danielsn, line 492: safety comment

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,628 @@
...
+    pub end: usize,
+    pub prot: i32,
+}
+
+/// Parse /proc/self/maps into a list of (range, prot) entries.
+///
+/// Used to remember each GOT page's original protection so we can restore
+/// it after patching, rather than leaving Full-RELRO pages read-write for
+/// the lifetime of the process.
+pub fn read_proc_maps() -> Vec<MapEntry> {
```
danielsn, line 438: are there any tests for this?

libdd-gotter/README.md
scottgerring: AFAIK statically linking glibc is strongly discouraged and doesn't work in the general case because a bunch of its functionality requires it to `dlopen` other bits of itself. Having said that, I think this point kind of conflates a couple things - its only a problem to GOT patch statically linked libraries if the symbol you are trying to intercept is statically linked into the target module - it's not _really_ about glibc/musl at all, unless those contain the symbols you are after.

Maybe something like the following is clearer:
```md
## Support 
This library can be used on ARM64 and AMD64 Linux in processes using glibc or musl runtimes. 
Only symbols that have been dynamically linked can be intercept. 
For instance, if you want to intercept the `malloc` of your C runtime, 
you _cannot_ do so if the application has been statically linked against musl.
```

libdd-profiling-heap-gotter/src/elf.rs
scottgerring: Yann and I asked @gyuheon0h to break his change out into 1/ a mechanical splitting of crates and 2/ an improvements PR. This one is intended to be the "I'm just moving code around" one, and the latter is stacked on top of #2297 . Can we address the "let's make this clearer" feedback in the latter one to maintain clarity?

scottgerring (approved):
LGTM! Thanks for addressing comments. Suggest addressing continued feedback on "how we can make this better" in the stacked PR.

libdd-gotter/README.md
scottgerring: The ancient dark arts passed down from engineer to engineer 😆 😱  
I found an [article in phrack](https://phrack.org/issues/56/7)  which is probably not corporate-appropriate but is pretty great. @nsavoire i'm not sure if you've ever seen anything good  ?

libdd-gotter/src/elf.rs
gyuheon0h: I don't think these will ever really drift

libdd-gotter/src/elf.rs
gyuheon0h: Sure. A follow up could be to include metric on the success of the hook / what fallbacks were used

libdd-gotter/src/elf.rs
gyuheon0h: We aren't catching unexpected values, `DT_*` tags are an open set. A typical ELF object has dozens of dynamic tags that we simply don't care about. We only extract the subset needed for GOT patching.

The loop is bounded by DT_NULL which the ELF spec requires as the sentinel terminator of the dynamic section. Every well-formed ELF object has it. If the dynamic section were truncated without a `DT_NULL`, we'd walk off the end but that would be a corrupt binary, and `dl_iterate_phdr` itself would have already had problems loading it.

libdd-gotter/README.md
gyuheon0h: Not really. the `Release`/`Acquire` pairing is a convention enforced by the programmer. The closest you could get is wrapping the pattern in a struct so the store and load orderings are co-located and can't drift independently but that is just moving the comment to code. Its up to the consumers of the library to do what they wish. 

The real protection is that `hook_symbol` writes `orig_out` before returning, and the caller stores it before any hook can fire (since GOT entries are already patched by that point on the same thread). A hook on another thread that races the store will see 0 and take the fallback path `libc::abort()`, which is safe if not ideal

libdd-gotter/src/elf.rs
```diff
@@ -0,0 +1,674 @@
...
+    let mut found: Option<LookupResult> = None;
+    iterate_libraries(|info, _is_exe| unsafe {
+        let lib_name = if info.dlpi_name.is_null() {
+            ""
+        } else {
+            CStr::from_ptr(info.dlpi_name).to_str().unwrap_or("")
+        };
+        if lib_name.contains("linux-vdso") || lib_name.contains("/ld-linux") {
+            return false;
+        }
```
realFlowControl, line 580: Something like this is missing from a dd-trace-php PoV:

https://github.com/DataDog/dd-trace-php/blob/8b890b89902789114a37ae8f779c8a5f1647cbcb/profiling/src/io/got_elf64.rs#L186-L197

Basically: when we hook the `read()` function, we do not want to hook the `read()` function in the profiler itself, but only in the binary and all other libs loaded.

realFlowControl (commented):
Had one comment, that might be dd-trace-php specific and is not a blocker to merging this PR.

libdd-gotter/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "libdd-gotter"
+version = "1.0.0"
```
iunanua: Please set `0.1.0` so that is considered as an initial release by the release proposal workflow.

```suggestion
version = "0.1.0"
```

libdd-gotter/src/elf.rs
gyuheon0h: Ill impl this on this PR: [feat(gotter): add fn to hook but exclude self](https://github.com/DataDog/libdatadog/pull/2312)

thanks for the guidance

iunanua (approved):

PR #2285, opened 2026-07-28
refactor(datadog-ffe)!: rename crate so it can be published to crates.io

hoolioh:
# What does this PR do?

This PR makes the proper changes to publish FFE crate, for that the crate name has to be renamed in order to adhere to the repo publishing policies and dependants references updated.

# Motivation

Allow downstream projects to use independent versions.


Aaalibaba42 (approved):

libdd-ffe-ffi/Cargo.toml
```diff
@@ -22,7 +22,7 @@ build_common = { path = "../build-common" }
 
 [dependencies]
 anyhow = "1.0.103"
-datadog-ffe = { path = "../datadog-ffe", version = "=1.0.0" }
+libdd-ffe = { path = "../libdd-ffe", version = "0.1.0" }
```
dd-oleksii: major: this probably needs to pin the exact version of `libdd-ffe`

libdd-ffe-ffi/CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 ## 1.0.1
 
 ### Patch changes
-- Bundle datadog-ffe-ffi with the rest of libdatadog release. In preparation for this, we also fixed a couple of smaller issues and inconsistencies.
+- Bundle libdd-ffe-ffi with the rest of libdatadog release. In preparation for this, we also fixed a couple of smaller issues and inconsistencies.
```
dd-oleksii: minor: it doesn't make sense to retroactively modify changelog

libdd-ffe/Cargo.toml
```diff
@@ -1,17 +1,19 @@
 [package]
-name = "datadog-ffe"
-version = "1.0.0"
+name = "libdd-ffe"
+version = "0.1.0"
```
dd-oleksii, line 3: major: any reason this is downgraded to sub 1?

libdd-ffe/Cargo.toml
```diff
@@ -1,17 +1,19 @@
...
 rust-version.workspace = true
 edition.workspace = true
 license.workspace = true
 autobenches = false
-publish = false
+homepage = "https://github.com/DataDog/libdatadog/tree/main/libdd-ffe"
+repository = "https://github.com/DataDog/libdatadog/tree/main/libdd-ffe"
+description = "Library that implements Feature Flag Evaluation"
```
dd-oleksii:
```suggestion
description = "Feature Flagging and Experimentation"
```

libdd-ffe/Cargo.toml
hoolioh: It's to avoid that the release system will publish a 2.0.0 version. The release system works on the assumption that every crate on the workspace stays in beta until it's published because it doesn't track which crates are published and which aren't so if I leave the 1.0.0 the system will bump the major and a 2.0.0 version will appear as a the first release in crates.io.
It's something we could improve in the future though

libdd-ffe-ffi/CHANGELOG.md
hoolioh: the changelog hasn't been published to crates.io yet so I thought that it would make sense to align it with the crate name.

libdd-ffe-ffi/Cargo.toml
hoolioh: In this case, since the crates are part of the same workspace, an `path` takes precedence over `version` the versions are algined in every commit. However I might be missing something, What would be the benefits of pinning the version?

dd-oleksii: Primarily to have easier provenance of what code is running in user's system. Ideally, one version of top-level library/tracers maps to exactly one version of ffe library. This helps with debugging and making sure users don't get unexpected features.

`path` is ignored for crates.io releases, so the only way to keep crates.io consumers aligned is via version constraints.

libdd-ffe/Cargo.toml
dd-oleksii: Then we need to decrease `libdd-ffe-ffi` crate version as well?

libdd-ffe-ffi/CHANGELOG.md
dd-oleksii: Yeah... we've had an unsuccessful attempt at versioning before realizing libdatadog is primarily consumed via git. I think the sensible thing to do is erase the changelog here and start from scratch

libdd-ffe-ffi/Cargo.toml
hoolioh: Fair enough, I can leave it pinned but since the ffi layer is consumed by using the released artifacts of by tag references it won't make much of a difference.

libdd-ffe-ffi/CHANGELOG.md
hoolioh: Make sense. It eventually the crate is published to crates.io we will create an initial release and start from there.

libdd-ffe/Cargo.toml
hoolioh: It's not needed, at the moment, we're not publishing the ffi crates since the ffi library is consumed via the released artifacts or by tag reference to libdd-profiling-ffi crate (which is the one that manages the inclusion of the rest of the crates). 
In the future if we publish the ffi crates we will need to do it but we're in thinking about how to rearchitecture the whole thing.

dd-oleksii (approved):

PR #2286, opened 2026-07-28
fix(telemetry)!: stop sending Connection: close to the Agent

cataphract:
Connection is hop-by-hop, but the Agent telemetry proxy currently copies it into a newly constructed Go request without copying the parsed Request.Close state. The forwarded request therefore carries Connection: close while the Agent transport still considers its connection reusable.

An HTTP/1.1 server that receives the close option must close after the response and SHOULD (but need not) repeat the option in that response. If it omits the response option, the Agent can hand the closing connection to a queued telemetry POST, which then fails with EOF or ECONNRESET.

Remove the header to work-around this bug.

See 73f23a2c03be39971c966e444785d63b6fd52e81

See also https://github.com/DataDog/system-tests/pull/7394

BREAKING CHANGE: here to satisfy the fantasy that libdatadog has a stable API; this adds two public functions

### DIFF
--- a/libdd-capabilities-impl/src/http.rs
+++ b/libdd-capabilities-impl/src/http.rs
@@ -8,7 +8,6 @@ mod native {
     use std::future::Future;
     use std::io::Write;
     use std::sync::{Arc, OnceLock};
-    use std::time::Duration;
 
     use libdd_capabilities::http::{
         BodySender, ChunkFuture, HttpClientCapability, HttpError, ResponseFuture,
@@ -17,23 +16,15 @@ mod native {
     use libdd_capabilities::maybe_send::MaybeSend;
     use libdd_common::connector::Connector;
     use libdd_common::http_common::{
-        new_client_periodic, new_client_with_connection_pool_idle_timeout, new_default_client,
-        Body, GenericHttpClient,
+        new_client_periodic, new_default_client, Body, GenericHttpClient,
     };
 
     use http_body_util::BodyExt;
 
-    #[derive(Clone, Copy, Debug)]
-    enum ConnectionPooling {
-        Disabled,
-        Default,
-        IdleTimeout(Duration),
-    }
-
     #[derive(Clone)]
     pub struct NativeHttpClient {
         client: Arc<OnceLock<GenericHttpClient<Connector>>>,
-        connection_pooling: ConnectionPooling,
+        connection_pooling: bool,
     }
 
     pub struct NativeBodySender(libdd_common::http_common::Sender);
@@ -63,24 +54,7 @@ mod native {
         pub fn new_without_connection_pooling() -> Self {
             Self {
                 client: Arc::new(OnceLock::new()),
-                connection_pooling: ConnectionPooling::Disabled,
-            }
-        }
-
-        pub fn new_with_connection_pool_idle_timeout(idle_timeout: Duration) -> Self {
-            Self {
-                client: Arc::new(OnceLock::new()),
-                connection_pooling: ConnectionPooling::IdleTimeout(idle_timeout),
-            }
-        }
-    }
-
-    fn build_client(connection_pooling: ConnectionPooling) -> GenericHttpClient<Connector> {
-        match connection_pooling {
-            ConnectionPooling::Disabled => new_client_periodic(),
-            ConnectionPooling::Default => new_default_client(),
-            ConnectionPooling::IdleTimeout(idle_timeout) => {
-                new_client_with_connection_pool_idle_timeout(idle_timeout)
+                connection_pooling: false,
             }
         }
     }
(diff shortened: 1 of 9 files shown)


estringana (approved):

libdd-telemetry/src/worker/http_client.rs
```diff
@@ -59,11 +59,6 @@ pub fn request_builder(c: &Config) -> anyhow::Result<HttpRequestBuilder> {
             );
             let mut builder =
                 e.to_request_builder(concat!("telemetry/", env!("CARGO_PKG_VERSION")));
```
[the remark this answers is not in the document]
cataphract: The review is materially correct, but its mechanism is misstated.

The problem is not "heartbeat exceeds the Agent timeout". Once the Agent's FIN is fully observed, Hyper discards the pooled connection and reconnects successfully. The failure occurs when the heartbeat and idle-close deadlines coincide (currently both default to 60 seconds) and the POST races the Agent closing the connection.

Since the 60 s timeout is hardcoded in the agent, a better workaround for this agent's bug is to reduce the timeout of the idle connections.

cataphract:
@Aaalibaba42 @ekump  assigning you as reviewers as this is partially reversing a commit you are the authors of (https://github.com/DataDog/libdatadog/commit/73f23a2c03be39971c966e444785d63b6fd52e81  )

libdd-telemetry/src/worker/http_client.rs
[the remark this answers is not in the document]
cataphract: Reduced to 45 s in https://github.com/DataDog/libdatadog/pull/2286/commits/f2aca25fb02933e7be31bbfc409343d45540a153 . I'm now unable to reproduce the previous failures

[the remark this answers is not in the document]
cataphract: Actually, we can't really call it a bug of the agent. The race is ultimately unavoidable if there is connection reuse and the client doesn't have a  smaller timeout (with a safety margin). Perhaps libdatadog could also automatically retry the request if it fails for this reason, though we'd need to make sure there is no risk of it being doubly submitted.

libdd-capabilities-impl/src/http.rs
```diff
@@ -54,7 +63,24 @@ mod native {
         pub fn new_without_connection_pooling() -> Self {
             Self {
                 client: Arc::new(OnceLock::new()),
-                connection_pooling: false,
+                connection_pooling: ConnectionPooling::Disabled,
+            }
+        }
+
+        pub fn new_with_connection_pool_idle_timeout(idle_timeout: Duration) -> Self {
```
ekump: For telemetry, why not just use [new_without_connection_pooling()](https://github.com/DataDog/libdatadog/blob/f2aca25fb02933e7be31bbfc409343d45540a153/libdd-capabilities-impl/src/http.rs#L63)? It accomplishes the same goal.

cataphract: Yes, it would work, but I'm not sure it's preferable. Connection pooling is generally desirable.

paullegranddc (commented):
IIRC this worked before the telemetry was passed to using capabilities in this PR https://github.com/DataDog/libdatadog/pull/2172

The telemetry client called to [`new_client_periodic` ](https://github.com/DataDog/libdatadog/blob/802f06a842848ba81b4fed9587a0ba7904cb7830/libdd-telemetry/src/worker/http_client.rs#L131C29-L131C48) which [did not perform connection pooling](https://github.com/DataDog/libdatadog/blob/main/libdd-common/src/http_common.rs#L124-L128).

Having a connection timeout less than 60 seconds (the period for telemetry sending) is worse than not maintaining, as we __never__ reuse the connection. So we will just keep an idle socket around for 45s, drop it, and re-open a new one anyway after 60s.

I would strongly prefer if you used `new_without_connection_pooling` and removed the code [here](https://github.com/DataDog/libdatadog/blob/19987a197aa46fda53019f1bdf6dd05504e9263b/libdd-telemetry/src/worker/http_client.rs#L62-L66) as your changes do not bring advantages over it

libdd-capabilities-impl/src/http.rs
ekump: @paullegranddc's [comment](https://github.com/DataDog/libdatadog/pull/2286#pullrequestreview-4808848697) explains for me.

cataphract: OK changed to NativeHttpClient::new_without_connection_pooling()

bwoebi (approved):

leoromanovsky (approved):

Aaalibaba42 (approved):

ekump (approved):

PR #2297, opened 2026-07-29
feat(gotter)!: add DT_HASH fallback, relocation type guard, and hook_symbol

gyuheon0h:
Stacked above [refactor(got-patching): extract shared GOT-patching primitives into libdd-got-hook](https://github.com/DataDog/libdatadog/pull/2282)
Stacked under [feat(crashtracking): retrieve c assert message for linux when __assert_fail is dynamically loaded](https://github.com/DataDog/libdatadog/pull/2268)

[PROF-15629](https://datadoghq.atlassian.net/jira/software/c/projects/PROF/boards/11?selectedIssue=PROF-15629)

# What does this PR do?
Adds three general improvements to `libdd-got-hook` for crash tracking (and beneficial to heap profiling):
- `DT_HASH` (SysV) fallback: `DynamicInfo::from_phdr` no longer requires `DT_GNU_HASH`. Objects linked with `--hash-style=sysv`are now parsed by reading `nchain` from the SysV hash header. We jjust skipped these objects previously.
- Relocation type guard: `elf64_r_type` + `is_got_pointer_reloc` now only patches `GLOB_DAT` and `JUMP_SLOT` relocations (pointer-sized GOT slots) on both `x86_64` and `aarch64`. Non-pointer relocation types are now skipped, preventing potential corruption of adjacent code/data from writing 8 bytes into a 4-byte relocation field. Applied to both `libdd-got-hook`'s `patch_got_entries` and `libdd-profiling-heap-gotter`'s `process_relocation`. (this was a codex recommendation)
- `hook_symbol`: Single-symbol convenience wrapper that combines `dlsym` + `iterate_libraries` + `patch_got_entries` into one call.

# Motivation

# Additional Notes

Gated on @scottgerring's approval

# How to test the change?

[PROF-15629]: https://datadoghq.atlassian.net/browse/PROF-15629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libdd-gotter/src/elf.rs
+++ b/libdd-gotter/src/elf.rs
@@ -858,39 +858,34 @@ unsafe fn patch_got_entries(
     guard: &mut PageProtGuard,
     patched: &mut bool,
 ) {
-    // NOTE: the SysV x86-64 ABI specifies that only RELA entries are
-    // used on AMD64 (spec page 64). ARM64 appears similar. REL
-    // processing is kept for defensive completeness but may be
-    // dead code on both architectures. We should revisit this.
-    for reloc in dyn_info.rels() {
-        if !is_got_pointer_reloc(elf64_r_type(reloc.r_info)) {
-            continue;
+    // Both REL and RELA relocations carry r_info (symbol + type) and
+    // r_offset (GOT slot address). RELA has an additional r_addend we
+    // don't use. This helper processes one relocation by those two fields.
+    let mut try_patch = |r_info: u64, r_offset: u64| {
+        if !is_got_pointer_reloc(elf64_r_type(r_info)) {
+            return;
         }
-        let sym_idx = elf64_r_sym(reloc.r_info);
+        let sym_idx = elf64_r_sym(r_info);
         if let Some(cstr) = dyn_info.sym_name(sym_idx) {
             if cstr.to_bytes() == symbol_name {
-                let addr = reloc.r_offset as usize + dyn_info.base_address();
+                let addr = r_offset as usize + dyn_info.base_address();
                 if guard.override_entry(addr, hook_fn) {
                     *patched = true;
                 }
             }
         }
-    }
+    };
 
+    // NOTE: the SysV x86-64 ABI specifies that only RELA entries are
+    // used on AMD64 (spec page 64). ARM64 appears similar. REL
+    // processing is kept for defensive completeness but may be
+    // dead code on both architectures. We should revisit this.
+    for reloc in dyn_info.rels() {
+        try_patch(reloc.r_info, reloc.r_offset);
+    }
     for relocs in [dyn_info.relas(), dyn_info.jmprels()] {
         for reloc in relocs {
-            if !is_got_pointer_reloc(elf64_r_type(reloc.r_info)) {
-                continue;
-            }
-            let sym_idx = elf64_r_sym(reloc.r_info);
-            if let Some(cstr) = dyn_info.sym_name(sym_idx) {
-                if cstr.to_bytes() == symbol_name {
-                    let addr = reloc.r_offset as usize + dyn_info.base_address();
-                    if guard.override_entry(addr, hook_fn) {
-                        *patched = true;
-                    }
-                }
-            }
+            try_patch(reloc.r_info, reloc.r_offset);
         }
     }
 }
--- a/libdd-gotter/src/elf.rs
+++ b/libdd-gotter/src/elf.rs
@@ -839,10 +839,20 @@ pub struct HookResult {
     /// Resolved address of the original symbol. Store this so the hook
     /// function can forward calls to the real implementation.
     pub orig_addr: usize,
-    /// Whether at least one GOT entry was patched. `false` means the
-    /// symbol was found but no loaded library had a matching GOT
-    /// relocation for it (e.g. statically linked libc on musl).
-    pub patched: bool,
+    /// Number of GOT entries successfully rewritten.
+    pub entries_patched: usize,
+    /// Number of GOT entries that matched the symbol but could not be
+    /// patched (`mprotect` failed to make the page writable).
+    pub entries_failed: usize,
+}
+
+/// Error returned by [`hook_symbol`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum HookError {
+    /// `symbol_name` was not valid.
+    InvalidSymbolName,
+    /// No loaded library exports a definition for this symbol.
+    SymbolNotFound,
 }
 
 /// Hook a single symbol across all loaded ELF objects by patching their
@@ -856,29 +866,37 @@ pub struct HookResult {
 /// [`PageProtGuard`] primitives to build a registry that re-scans on
 /// `dlopen` (see `libdd-profiling-heap-gotter`'s `SymbolOverrides`).
 ///
-/// - `symbol_name`: the symbol to hook (e.g. `c"__assert_fail"`)
+/// - `symbol_name`: the symbol to hook (`c"__assert_fail"`)
 /// - `hook_fn`: address of the replacement function
 ///
-/// Returns `None` if the symbol could not be found in any loaded
-/// library. Returns `Some(HookResult)` if the symbol was resolved,
-/// with `orig_addr` set to the original function address and `patched`
+/// Returns `Ok(HookResult)` if the symbol was resolved, with
+/// `orig_addr` set to the original function address and `patched`
 /// indicating whether any GOT entries were actually rewritten.
+/// Returns `Err(HookError)` if the symbol name is invalid or the
+/// symbol could not be found.
 ///
 /// # Safety
 ///
 /// `hook_fn` must point to a function with the same calling convention
 /// and signature as the symbol being hooked. The patching is permanent.
-pub unsafe fn hook_symbol(symbol_name: &CStr, hook_fn: usize) -> Option<HookResult> {
+pub unsafe fn hook_symbol(
+    symbol_name: &CStr,
+    hook_fn: usize,
+) -> Result<HookResult, HookError> {
     let symbol_name_bytes = symbol_name.to_bytes();
-    let name_str = symbol_name.to_str().ok()?;
+    let name_str = symbol_name
+        .to_str()
+        .map_err(|_| HookError::InvalidSymbolName)?;
 
-    let result = lookup_symbol(name_str, hook_fn)?;
+    let result = lookup_symbol(name_str, hook_fn).ok_or(HookError::SymbolNotFound)?;
 
-    let mut patched_any = false;
+    let mut entries_patched: usize = 0;
+    let mut entries_failed: usize = 0;
     let mut guard = PageProtGuard::new();
 
     let guard_ptr = &mut guard as *mut PageProtGuard;
-    let patched_ptr = &mut patched_any as *mut bool;
+    let patched_ptr = &mut entries_patched as *mut usize;
+    let failed_ptr = &mut entries_failed as *mut usize;
 
     iterate_libraries(|info, _is_exe| {
         let lib_name = if info.dlpi_name.is_null() {
@@ -900,8 +918,9 @@ pub unsafe fn hook_symbol(symbol_name: &CStr, hook_fn: usize) -> Option<HookResu
             return false;
         };
         // SAFETY: dyn_info was just produced from a currently-loaded
-        // library. guard_ptr/patched_ptr are valid for the duration of
-        // iterate_libraries (they point to locals in the enclosing fn).
+        // library. guard_ptr/patched_ptr/failed_ptr are valid for the
+        // duration of iterate_libraries (they point to locals in the
+        // enclosing fn).
         unsafe {
             patch_got_entries(
                 &dyn_info,
@@ -909,14 +928,16 @@ pub unsafe fn hook_symbol(symbol_name: &CStr, hook_fn: usize) -> Option<HookResu
                 hook_fn,
                 &mut *guard_ptr,
                 &mut *patched_ptr,
+                &mut *failed_ptr,
             );
         }
         false
     });
 
-    Some(HookResult {
+    Ok(HookResult {
         orig_addr: result.address,
-        patched: patched_any,
+        entries_patched,
+        entries_failed,
     })
 }
 
@@ -935,7 +956,8 @@ unsafe fn patch_got_entries(
     symbol_name: &[u8],
     hook_fn: usize,
     guard: &mut PageProtGuard,
-    patched: &mut bool,
+    patched: &mut usize,
+    failed: &mut usize,
 ) {
     // Both REL and RELA relocations carry r_info (symbol + type) and
     // r_offset (GOT slot address). RELA has an additional r_addend we
@@ -949,7 +971,9 @@ unsafe fn patch_got_entries(
             if cstr.to_bytes() == symbol_name {
                 let addr = r_offset as usize + dyn_info.base_address();
                 if guard.override_entry(addr, hook_fn) {
-                    *patched = true;
+                    *patched += 1;
+                } else {
+                    *failed += 1;
                 }
             }
         }
--- a/libdd-gotter/src/elf.rs
+++ b/libdd-gotter/src/elf.rs
@@ -52,9 +52,10 @@ pub struct DynamicInfo {
     /// Used by [`gnu_hash_lookup`] for symbol resolution.
     gnu_hash: *const u32,
     gnu_hash_words: usize,
-    /// Pointer to the `DT_HASH` (sysv) table, if present.
+    /// Pointer and word-count for the `DT_HASH` (sysv) table, if present.
     /// Used by [`sysv_hash_lookup`] as a fallback when `DT_GNU_HASH` is absent.
     sysv_hash: *const u32,
+    sysv_hash_words: usize,
     rels: *const Elf64_Rel,
     rels_count: usize,
     relas: *const Elf64_Rela,
@@ -148,6 +149,17 @@ impl DynamicInfo {
             return None;
         }
 
+        // Compute sysv_hash_words from the containing load segment.
+        let sysv_hash_words = if !sysv_hash.is_null() {
+            let addr = sysv_hash as usize;
+            containing_load_segment_end(addr)
+                .and_then(|end| end.checked_sub(addr))
+                .map(|bytes| bytes / core::mem::size_of::<u32>())
+                .unwrap_or(0)
+        } else {
+            0
+        };
+
         // Determine sym_count and gnu_hash metadata.
         let (sym_count, gnu_hash_words) = if !gnu_hash.is_null() {
             let gnu_hash_addr = gnu_hash as usize;
@@ -174,6 +186,7 @@ impl DynamicInfo {
             gnu_hash,
             gnu_hash_words,
             sysv_hash,
+            sysv_hash_words,
             rels,
             rels_count: rels_size / core::mem::size_of::<Elf64_Rel>(),
             relas,
@@ -301,6 +314,15 @@ pub fn sysv_hash(name: &[u8]) -> u32 {
 
 /// Look up a symbol by name in an object's `DT_HASH` (sysv) table.
 ///
+/// [nbucket] [nchain] [bucket[0..nbucket]] [chain[0..nchain]]
+///
+/// Each bucket holds the index of the first symbol in that bucket's
+/// chain (or `STN_UNDEF` if empty). Each chain entry at position `i`
+/// holds the index of the next symbol after symbol `i` in the same
+/// bucket (or `STN_UNDEF` to end the chain). `nchain` equals the
+/// total number of dynamic symbols, so chain indices double as
+/// symbol indices into `.dynsym`.
+///
 /// Returns the `Elf64_Sym` entry if found and valid (per [`check_sym`]).
 ///
 /// # Safety
@@ -308,26 +330,40 @@ pub fn sysv_hash(name: &[u8]) -> u32 {
 /// currently-loaded ELF object.
 pub unsafe fn sysv_hash_lookup(info: &DynamicInfo, name: &[u8]) -> Option<Elf64_Sym> {
     let hashtab = info.sysv_hash;
-    if hashtab.is_null() {
+    if hashtab.is_null() || info.sysv_hash_words < 2 {
         return None;
     }
 
+    // Read the header: nbucket and nchain.
     let nbucket = *hashtab;
-    // nchain at hashtab[1] == sym_count, already validated by from_phdr
+    let nchain = *hashtab.add(1);
     if nbucket == 0 {
         return None;
     }
 
-    let buckets = hashtab.add(2);
-    let chains = buckets.add(nbucket as usize);
+    // Validate the table fits within the mapped region before computing
+    // any pointers into the bucket/chain arrays.
+    let buckets_start: usize = 2;
+    let chains_start = buckets_start.checked_add(nbucket as usize)?;
+    let table_end = chains_start.checked_add(nchain as usize)?;
+    if table_end > info.sysv_hash_words {
+        return None;
+    }
+
+    let buckets = hashtab.add(buckets_start);
+    let chains = hashtab.add(chains_start);
 
     let h = sysv_hash(name);
     let mut idx = *buckets.add((h % nbucket) as usize);
 
-    // Walk the chain. Guard with sym_count to avoid infinite loops on
-    // malformed tables
+    // Follow the chain from the bucket's head symbol, comparing names
+    // at each step. The chain terminates at STN_UNDEF (0). We also
+    // cap iterations at nchain to guard against malformed cycles.
     let mut steps = 0u32;
-    while idx != STN_UNDEF && steps < info.sym_count {
+    while idx != STN_UNDEF && steps < nchain {
+        if (idx as usize) >= nchain as usize {
+            break;
+        }
         if let Some(sname) = info.sym_name(idx) {
             let sym = &*info.symtab.add(idx as usize);
             if sname.to_bytes() == name && check_sym(sym) {
--- a/libdd-gotter/src/elf.rs
+++ b/libdd-gotter/src/elf.rs
@@ -9,6 +9,7 @@
 //! * REL / RELA / JMPREL relocation arrays.
 
 use core::ffi::{c_char, c_int, c_void, CStr};
+use core::slice;
 use std::collections::HashMap;
 use std::io::{BufRead, BufReader};
 
@@ -51,6 +52,9 @@ pub struct DynamicInfo {
     /// Used by [`gnu_hash_lookup`] for symbol resolution.
     gnu_hash: *const u32,
     gnu_hash_words: usize,
+    /// Pointer to the `DT_HASH` (sysv) table, if present.
+    /// Used by [`sysv_hash_lookup`] as a fallback when `DT_GNU_HASH` is absent.
+    sysv_hash: *const u32,
     rels: *const Elf64_Rel,
     rels_count: usize,
     relas: *const Elf64_Rela,
@@ -74,7 +78,7 @@ impl DynamicInfo {
     /// # Safety
     /// `info` must point to a valid `dl_phdr_info` from `dl_iterate_phdr`.
     pub unsafe fn from_phdr(info: &dl_phdr_info) -> Option<Self> {
-        let phdrs = core::slice::from_raw_parts(info.dlpi_phdr, info.dlpi_phnum as usize);
+        let phdrs = slice::from_raw_parts(info.dlpi_phdr, info.dlpi_phnum as usize);
         let dyn_phdr = phdrs.iter().find(|p| p.p_type == PT_DYNAMIC)?;
         let dyn_begin = (info.dlpi_addr as usize + dyn_phdr.p_vaddr as usize) as *const Elf64_Dyn;
         let base = info.dlpi_addr as usize;
@@ -168,6 +172,7 @@ impl DynamicInfo {
             sym_count,
             gnu_hash,
             gnu_hash_words,
+            sysv_hash,
             rels,
             rels_count: rels_size / core::mem::size_of::<Elf64_Rel>(),
             relas,
@@ -213,7 +218,7 @@ impl DynamicInfo {
             // entries of a mapped ELF object; the array is valid for the
             // object's lifetime (held by dl_iterate_phdr's loader lock or by
             // the caller's dlopen handle).
-            unsafe { core::slice::from_raw_parts(self.rels, self.rels_count) }
+            unsafe { slice::from_raw_parts(self.rels, self.rels_count) }
         }
     }
 
@@ -239,33 +244,14 @@ impl DynamicInfo {
         }
     }
 
-    /// Linear scan of the dynsym table for a symbol by name.
-    ///
-    /// This is the fallback for objects that have `DT_HASH` but no
-    /// `DT_GNU_HASH`. The scan is bounded by `sym_count` (this is calc from
-    /// `DT_HASH` nchain or the symtab/strtab distance heuristic).
-    ///
-    /// # Safety
-    /// The `DynamicInfo` must have been produced by [`DynamicInfo::from_phdr`]
-    /// for a currently-loaded ELF object.
-    pub unsafe fn linear_sym_lookup(&self, name: &[u8]) -> Option<Elf64_Sym> {
-        for idx in 0..self.sym_count {
-            let sym = &*self.symtab.add(idx as usize);
-            let off = sym.st_name as usize;
-            if off >= self.strtab_size {
-                continue;
-            }
-            let sname = CStr::from_ptr(self.strtab.add(off));
-            if sname.to_bytes() == name && check_sym(sym) {
-                return Some(*sym);
-            }
-        }
-        None
-    }
-
     /// Whether this object has a usable GNU hash table.
     pub fn has_gnu_hash(&self) -> bool {
         !self.gnu_hash.is_null() && self.gnu_hash_words >= 4
+    }
+
+    /// Whether this object has a usable sysv (`DT_HASH`) hash table.
+    pub fn has_sysv_hash(&self) -> bool {
+        !self.sysv_hash.is_null()
     }
 }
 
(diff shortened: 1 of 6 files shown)


gyuheon0h:
* **#2312** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2312?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2297** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2297?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2282** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2282?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

libdd-gotter/src/elf.rs
```diff
@@ -211,6 +215,63 @@ impl DynamicInfo {
     pub fn jmprels(&self) -> (*const Elf64_Rela, usize) {
         (self.jmprels, self.jmprels_count)
     }
+
+    /// Linear scan of the dynsym table for a symbol by name.
+    ///
+    /// This is the fallback for objects that have `DT_HASH` but no
+    /// `DT_GNU_HASH`. The scan is bounded by `sym_count` (this is calc from
```
yannham: So I'm a bit confused, shouldn't the sysv hash still allow to make faster lookups than a linear scan? It looks like we don't use it at all here. For the record I have no idea how the SysV hash interface works.

libdd-gotter/src/elf.rs
```diff
@@ -477,10 +535,6 @@ impl PageProtGuard {
     pub unsafe fn override_entry(&mut self, addr: usize, new_value: usize) -> bool {
         let aligned = addr & !(self.page_size - 1);
         if !self.touched.contains_key(&aligned) {
-            // If /proc/self/maps isn't available (or the page isn't in
-            // it, which shouldn't happen for a mapped GOT page) fall
-            // back to PROT_READ - the RELRO'd default. That's tighter
-            // than the previous behavior of leaving pages RW.
```
yannham, line 519: Any reason to remove those comments?

libdd-gotter/src/elf.rs
```diff
@@ -541,7 +627,14 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
         let Some(dyn_info) = DynamicInfo::from_phdr(info) else {
             return false;
         };
-        if let Some(sym) = gnu_hash_lookup(&dyn_info, needle) {
+        // Try GNU hash first (O(1) average), fall back to linear scan
+        // for sysv-hash-only objects.
```
yannham:
```suggestion
```

libdd-gotter/src/elf.rs
```diff
@@ -555,6 +648,126 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
...
+/// # Safety
+///
+/// `hook_fn` must point to a function with the same calling convention
+/// and signature as the symbol being hooked. The patching is permanent.
+pub unsafe fn hook_symbol(symbol_name: &CStr, hook_fn: usize, orig_out: &mut usize) -> bool {
+    let symbol_name_bytes = symbol_name.to_bytes();
+    let name_str = match symbol_name.to_str() {
+        Ok(s) => s,
+        Err(_) => return false,
+    };
```
yannham:
```suggestion
    let Ok(name_str) = symbol_name.to_str() else {
        return false;
    };
```

libdd-gotter/src/elf.rs
```diff
@@ -555,6 +648,126 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
...
+    patched_any
+}
+
+/// Patch GOT entries in one library for the target symbol.
+///
+/// Only patches relocations of type `GLOB_DAT` or `JUMP_SLOT` — the
+/// pointer-sized GOT slots that the dynamic linker fills with resolved
+/// symbol addresses. Other relocation types (e.g. `R_X86_64_PC32`) have
+/// different widths or addend semantics and are skipped.
+unsafe fn patch_got_entries(
```
yannham, line 970: An unsafe function should have a safety comment about pre-conditions.

libdd-gotter/src/elf.rs
```diff
@@ -555,6 +648,126 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
...
+                .to_str()
+                .unwrap_or("")
+        };
+        if lib_name.contains("linux-vdso") || lib_name.contains("/ld-linux") {
+            return false;
+        }
+        let Some(dyn_info) = (unsafe { DynamicInfo::from_phdr(info) }) else {
+            return false;
+        };
+        unsafe {
```
yannham, line 940: Missing a Safety comment to say why it's ok to call patch_got_entries.

libdd-gotter/src/elf.rs
```diff
@@ -555,6 +648,126 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
...
+                }
+            }
+        }
+    }
+
+    for (ptr, count) in [dyn_info.relas(), dyn_info.jmprels()] {
+        if ptr.is_null() {
+            continue;
+        }
+        let relocs = core::slice::from_raw_parts(ptr, count);
```
yannham: Nit: import `core::slice` if it's used several times.

libdd-gotter/src/elf.rs
```diff
@@ -555,6 +648,126 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
...
+    patched: &mut bool,
+) {
+    let (rels_ptr, rels_count) = dyn_info.rels();
+    if !rels_ptr.is_null() {
+        let relocs = core::slice::from_raw_parts(rels_ptr, rels_count);
+        for reloc in relocs {
+            if !is_got_pointer_reloc(elf64_r_type(reloc.r_info)) {
+                continue;
+            }
+            let sym_idx = elf64_r_sym(reloc.r_info) as u32;
```
yannham: Isn't this already a `u32`?

```suggestion
            let sym_idx = elf64_r_sym(reloc.r_info);
```

libdd-gotter/src/elf.rs
```diff
@@ -555,6 +648,126 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
...
+        }
+        let relocs = core::slice::from_raw_parts(ptr, count);
+        for reloc in relocs {
+            if !is_got_pointer_reloc(elf64_r_type(reloc.r_info)) {
+                continue;
+            }
+            let sym_idx = elf64_r_sym(reloc.r_info) as u32;
+            if let Some(cstr) = dyn_info.sym_name(sym_idx) {
+                if cstr.to_bytes() == symbol_name {
+                    let addr = reloc.r_offset as usize + dyn_info.base_address();
```
yannham: Nit: avoid `as` conversion. If we're on x64, I _think_ there should be a direct `From` instance?
```suggestion
                    let addr = usize::from(reloc.r_offset) + dyn_info.base_address();
```

libdd-gotter/src/elf.rs
```diff
@@ -598,4 +888,118 @@ mod tests {
         let r = lookup_symbol("definitely_not_a_real_libc_symbol_xyzzy", 0);
         assert!(r.is_none());
     }
+
+    /// Compile a tiny shared library with `--hash-style=sysv` (no
+    /// DT_GNU_HASH), dlopen it, and verify that `DynamicInfo::from_phdr`
+    /// parses it with a valid sym_count calculated from DT_HASH nchain.
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn test_sysv_hash_library_parsed_correctly() {
```
yannham, line 1188: I wonder if it's reasonable to recompile a shared library each time just for a test. Would that be possible to pre-compile it instead as an asset and include it in the repo? Or maybe as an additional target of the crate and try to find it at test time using environment variable? Though maybe on a very small lib it's fast enough, but...

libdd-gotter/src/elf.rs
```diff
@@ -584,6 +797,83 @@ mod tests {
...
+            false
+        });
+        assert!(count > 0);
+    }
+
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn test_dynamic_info_parses_loaded_library() {
+        let mut found = false;
+        iterate_libraries(|info, _| {
```
yannham, line 1102: On all those tests that use `iterate_libraries`, do you rely on `libc` being loaded at least for them to not be flaky?

libdd-gotter/src/elf.rs
gyuheon0h: You're correct here

libdd-gotter/src/elf.rs
gyuheon0h: Ugh making the PR stack when my commits were not atomic. :sad

libdd-gotter/src/elf.rs
gyuheon0h: Nope, `Elf64_Xword` is u64

gyuheon0h: The `elf64_r_sym` returns u64 and `sym_name` takes u32 because symbol indices are 32-bit values in practice.

So the as u32 cast is always lossless so perhaps we _could_ just have `elf64_r_sym` return u32 directly since the result is always 32 bits

libdd-gotter/src/elf.rs
gyuheon0h: 😢 

```
error[E0277]: the trait bound `usize: From<u64>` is not satisfied
   --> libdd-gotter/src/elf.rs:816:28
    |
816 |                 let addr = usize::from(reloc.r_offset) + dyn_info.base_address();
    |                            ^^^^^ the trait `From<u64>` is not implemented for `usize`
    ```

File: libdd-gotter/src/elf.rs (hunk 11 above)
Comment (gyuheon0h) [reply libdd-gotter/src/elf.rs:1102]: Yes. `iterate_libraries` always returns at least the main executable (which has `PT_DYNAMIC` on dynamically linked binaries), so the iteration itself isn't flaky. But a couple of things here:
- `test_dynamic_info_parses_loaded_library` needs at least one object with PT_DYNAMIC + valid strtab/symtab/hash. The main executable could theoretically be statically linked with no dynamic section, making `from_phdr` return `None` for everything.
- `test_all_loaded_libraries_have_valid_sym_count` is the same as abov                                                    
- `test_can_lookup_known_symbol ` needs `malloc` (or whatever symbol) to exist in a dynamically loaded library.              
                                                                                                                            
On our test runners, all of these seem to be okay, which seems to me like as long as our test runner images don't change, these should never be flaky. 

But if you want these to be defensive, `test_iterate_libraries_finds_loaded_objects` could be the only hard assertion ( checks count > 0), and the others could skip gracefully if no library parses.

WDYT?

File: libdd-gotter/src/elf.rs [hunk 12]
```diff
@@ -65,10 +71,14 @@ impl DynamicInfo {
...
     ///
+    /// Supports both `DT_GNU_HASH` and `DT_HASH` (sysv) for determining
+    /// the dynsym entry count. Objects with neither hash table fall back
+    /// to a symtab/strtab distance heuristic.
+    ///
     /// # Safety
     /// `info` must point to a valid `dl_phdr_info` from `dl_iterate_phdr`.
     pub unsafe fn from_phdr(info: &dl_phdr_info) -> Option<Self> {
-        let phdrs = core::slice::from_raw_parts(info.dlpi_phdr, info.dlpi_phnum as usize);
+        let phdrs = slice::from_raw_parts(info.dlpi_phdr, info.dlpi_phnum as usize);
```
danielsn: I know Rust doesn't require `unsafe` blocks inside `unsafe` functions, but I consider that a design flaw.  Consider putting the actually unsafe parts inside blocks with safety comments to make it easier to audit

libdd-gotter/src/elf.rs
```diff
@@ -130,32 +142,28 @@ impl DynamicInfo {
...
+                if let Some(count) = gnu_hash_symbol_count(gnu_hash, words) {
+                    (count, words)
+                } else {
+                    (sym_count_fallback(symtab, strtab, sysv_hash), words)
+                }
             } else {
-                // Can't estimate; allow any index and rely on strtab
-                // bounds checking in sym_name to catch bad accesses.
-                u32::MAX
+                (sym_count_fallback(symtab, strtab, sysv_hash), 0)
```
danielsn, line 228: nice :)

libdd-gotter/src/elf.rs
```diff
@@ -540,19 +640,52 @@ impl Drop for PageProtGuard {
...
+///
+/// Accepted types:
+/// - `GLOB_DAT` / `JUMP_SLOT` -- GOT entries filled by the dynamic linker.
+/// - `R_X86_64_64` / `R_AARCH64_ABS64` -- absolute pointer-width relocations used for data-section
+///   function pointers (`void *(*fn)(size_t) = malloc;`).
+///
+/// Narrow or PC-relative types (`R_X86_64_PC32`, `R_AARCH64_TLSDESC`, etc.)
+/// are excluded since they have different widths and addend semantics
+pub fn is_got_pointer_reloc(r_type: u32) -> bool {
+    // x86_64
```
danielsn, line 781: Should these be `cfg` guarded by platform?

libdd-gotter/src/elf.rs
danielsn: Agreed.  Its a 🦶 🔫 that unchecked casts are easier than checked.

libdd-profiling-heap-gotter/src/elf.rs
```diff
@@ -226,17 +229,20 @@ impl SymbolOverrides {
             Self::process_relocation(
```
scottgerring, line 237: I think we should we guard here against `is_got_pointer_reloc` as well, same as down on :239

As an aside, after a bit of googling i'm not sure that REL relocations are used on AMD64/ARM64 _at all_:

[SysV X86-64 ABI](https://gitlab.com/x86-psABIs/x86-64-ABI/-/jobs/artifacts/master/raw/x86-64-ABI/abi.pdf?job=build) page 64:
> The AMD64 LP64 ABI architecture uses only Elf64_Rela relocation entries with explicit addends

I found similar noise for ARM64, but a bit less definitive. I think we might want to research a bit more if this code is needed at all, but we can probably do that async to this change.

Paging also @nsavoire for his wisdom

libdd-gotter/src/elf.rs
```diff
@@ -164,6 +172,7 @@ impl DynamicInfo {
             sym_count,
             gnu_hash,
             gnu_hash_words,
+            sysv_hash,
```
scottgerring, line 260: I reckon we also want `sysv_hash_words` here so that down in `sysv_hash_lookup` we can do bounds checking like `gnu_hash_lookup` does

libdd-gotter/src/elf.rs
```diff
@@ -227,9 +236,103 @@ impl DynamicInfo {
...
+    if hashtab.is_null() {
+        return None;
+    }
+
+    let nbucket = *hashtab;
+    // nchain at hashtab[1] == sym_count, already validated by from_phdr
+    if nbucket == 0 {
+        return None;
+    }
+
```
scottgerring, line 400: I think we should do bounds checking here, same as `gnu_hash_lookup` does (`buckets_start > info.gnu_hash_words || chains_start > info.gnu_hash_words`) before computing `buckets`.

Once `sysv_hash_words` is on `DynamicInfo`, check should be something like `2 + nbucket <= info.sysv_hash_words` here before computing `chains`, and add the equivalent of `chain_idx >= chain_count` inside the loop below - right now `nbucket`/`idx` are trusted unbounded, so `buckets.add(...)`/`chains.add(...)` can read past the actual table.

libdd-gotter/src/elf.rs
```diff
@@ -582,6 +722,123 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
...
+/// currently-loaded ELF object. `guard` must belong to the current
+/// patching pass.
+unsafe fn patch_got_entries(
+    dyn_info: &DynamicInfo,
+    symbol_name: &[u8],
+    hook_fn: usize,
+    guard: &mut PageProtGuard,
+    patched: &mut bool,
+) {
+    for reloc in dyn_info.rels() {
```
scottgerring: Nit: this loop is the same loop duplicated twice, which makes it _seem_ like there should be some rels / relas specific behaviour, but there's actually not - in both cases all you need is `r_info` and `r_offset`.

Suggest refactoring into a helper; will make it easier to read intent and ultimately maintain in the future

libdd-gotter/src/elf.rs
```diff
@@ -540,19 +640,52 @@ impl Drop for PageProtGuard {
...
+/// that is safe to overwrite with a function pointer.
+///
+/// Accepted types:
+/// - `GLOB_DAT` / `JUMP_SLOT` -- GOT entries filled by the dynamic linker.
+/// - `R_X86_64_64` / `R_AARCH64_ABS64` -- absolute pointer-width relocations used for data-section
+///   function pointers (`void *(*fn)(size_t) = malloc;`).
+///
+/// Narrow or PC-relative types (`R_X86_64_PC32`, `R_AARCH64_TLSDESC`, etc.)
+/// are excluded since they have different widths and addend semantics
+pub fn is_got_pointer_reloc(r_type: u32) -> bool {
```
scottgerring, line 780: this could use some tests! It's the main safety thing the PR adds, and it is a simple function over a single input type - ripe for testing :D 

Maybe one to cover the 6 pointer width variants => true, and one for garbage => false

libdd-gotter/src/elf.rs
```diff
@@ -625,4 +976,118 @@ mod tests {
...
+    /// Compile a tiny shared library with `--hash-style=sysv` (no
+    /// DT_GNU_HASH), dlopen it, and verify that `DynamicInfo::from_phdr`
+    /// parses it with a valid sym_count calculated from DT_HASH nchain.
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn test_sysv_hash_library_parsed_correctly() {
+        use std::io::Write;
+        use std::process::Command;
+
+        let dir = std::env::temp_dir().join("libdd_got_hook_test_sysv");
```
scottgerring: I think using the same temp_dir across tests might be fragile in some of our varied CI environments (dir re-use across tests/runs/feature-combos). Suggest using `tempfile`, which is already a workplace dep:

```
let dir = tempfile::tempdir().expect("create temp dir");
```

libdd-gotter/src/elf.rs
```diff
@@ -625,4 +976,118 @@ mod tests {
...
+        let status = Command::new("cc")
+            .args(["-shared", "-fPIC", "-Wl,--hash-style=sysv", "-o"])
+            .arg(&so_path)
+            .arg(&c_path)
+            .status();
+
+        let status = match status {
+            Ok(s) => s,
+            Err(e) => {
+                eprintln!("note: cc not available ({e}), skipping sysv hash test");
```
scottgerring: Silent failure feels a bit off here, as changes to our CI environments that remove access to a cc would mean we never know the test has become essentially a no-op.

Maybe we can run this through CI _failing_ on a lack of a cc, then use the resulting failure matrix to work out some narrower scoping of the test run to a job that we _do_ require to succeed?

scottgerring:
Thanks for this one, it adds some nice robustness improvements!
Also out of curiosity, did you see the older sysv hashing in the wild somewhere? I'm sure its still around in our customer base _somewhere_ and think its worth adding, but am genuinely curious as its getting pretty old now.

libdd-gotter/src/elf.rs
gyuheon0h: Not sure if this is necessary; ive seen this pattern before, and there shouldn't be any confusion by nature because of the names of the vars

libdd-gotter/src/elf.rs
gyuheon0h: All pass with hard failure on

gyuheon0h:
@scottgerring 

>Also out of curiosity, did you see the older sysv hashing in the wild somewhere?

No. Unfortunately, its hard to see what customers are running for their elf hashing :( but as part of just making this library more robust, I thought it was worth to do haha

libdd-gotter/src/elf.rs
```diff
@@ -595,10 +774,126 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
...
+}
+
+/// Hook a single symbol across all loaded ELF objects by patching their
+/// GOT entries.
+///
+/// - `symbol_name`: the symbol to hook (`c"__assert_fail"`)
+/// - `hook_fn`: address of the replacement function
+/// - `orig_out`: on success, receives the address of the original symbol
+///
+/// Returns `true` if at least one GOT entry was patched.
```
danielsn: Would there be value to returning the actual number of entries that are patched?

libdd-gotter/src/elf.rs
```diff
@@ -595,10 +774,126 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
...
+///
+/// # Safety
+///
+/// `hook_fn` must point to a function with the same calling convention
+/// and signature as the symbol being hooked. The patching is permanent.
+pub unsafe fn hook_symbol(symbol_name: &CStr, hook_fn: usize, orig_out: &mut usize) -> bool {
```
danielsn: How does this function return errors if any?

libdd-gotter/src/elf.rs
gyuheon0h: Well, its intended to be used [as such](https://github.com/DataDog/libdatadog/blob/3d864c21375012d065d7655c51648a73cd2f5505/libdd-crashtracker/src/collector/assert_interceptor.rs#L120-L127)

```
let patched = unsafe {
    libdd_gotter::hook_symbol(
        c"__assert_fail",
        hook_assert_fail as *const () as usize,
        &mut orig_addr,
    )
};
```

I am not sure if getting `patched_count` changes anything with how we can use it. Maybe for some extra metrics on how many times we are patching, but for our usecases (patching `read`/`write` for php, patching `malloc`/`free`/`calloc`/`realloc` for heap profiling, etc), I don't think that is particularly useful

libdd-gotter/src/elf.rs
gyuheon0h: It returns false for any failure (symbol not found, no GOT entries patched). The callers treat hooking as best-effort. If it fails, the feature degrades gracefully ( crash reports just show `SIGABRT` instead of the assert expression). We could change to `Result` if we want diagnostics, but none of the current callers would act on specific error variants. Happy to add it if you feel strongly. WDYT?

libdd-gotter/src/elf.rs
danielsn: Thought: these could be validated slices, so that users could be pure safe code

danielsn: Option<slice> would let us distinguish between "we couldn't find this" and "we found it but it was empty" if that is a useful distinction to make

libdd-gotter/src/elf.rs
```diff
@@ -209,7 +232,7 @@ impl DynamicInfo {
             // entries of a mapped ELF object; the array is valid for the
             // object's lifetime (held by dl_iterate_phdr's loader lock or by
             // the caller's dlopen handle).
-            unsafe { core::slice::from_raw_parts(self.rels, self.rels_count) }
+            unsafe { slice::from_raw_parts(self.rels, self.rels_count) }
```
danielsn: >data must be non-null and aligned even for zero-length slices or slices of ZSTs. One
reason for this is that enum layout optimizations may rely on references
(including slices of any length) being aligned and non-null to distinguish
them from other data. You can obtain a pointer that is usable as data
for zero-length slices using NonNull::dangling().

https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html



Do we ensure that

danielsn: we do check that above, it was hidden in the diff

libdd-gotter/src/elf.rs
```diff
@@ -130,66 +144,119 @@ impl DynamicInfo {
...
-                // bounds checking in sym_name to catch bad accesses.
-                u32::MAX
+                (sym_count_fallback(symtab, strtab, sysv_hash), 0)
             }
-        });
+        } else {
+            (sym_count_fallback(symtab, strtab, sysv_hash), 0)
+        };
+
+        // SAFETY (applies to all `slice::from_raw_parts` calls below):
```
gyuheon0h: These safety points come from `from_raw_parts` contract

libdd-gotter/src/elf.rs
```diff
@@ -671,4 +1104,167 @@ mod tests {
...
+                "from_phdr should succeed for sysv-hash library at {lib_name}"
+            );
+            let dyn_info = dyn_info.unwrap();
+            assert!(
+                !dyn_info.symtab.is_empty(),
+                "sym_count should be > 0 for sysv-hash library"
+            );
+
+            // Verify we can look up our exported symbol by walking
+            // relocations isn't needed
```
scottgerring:
```suggestion
            // relocations aren't needed
```

libdd-gotter/src/elf.rs
```diff
@@ -198,42 +265,142 @@ impl DynamicInfo {
...
+    // Follow the chain from the bucket's head symbol, comparing names
+    // at each step. The chain terminates at STN_UNDEF (0). We also
+    // cap iterations at nchain to guard against malformed cycles.
+    let mut steps = 0u32;
+    while idx != STN_UNDEF && steps < nchain {
+        if (idx as usize) >= nchain as usize {
+            break;
+        }
+        if let Some(sname) = info.sym_name(idx) {
+            let sym = &info.symtab[idx as usize];
```
scottgerring: I think we could use `get` here to avoid the indexing, or, restructure to avoid double indexing - `sym_name` indexes into symtab anyway but then discards it; we could add a `sym_entry` accessor that returns the tuple (Elf64_Sym, &Cstr) and then use it to get both here

libdd-gotter/src/elf.rs
```diff
@@ -595,10 +812,132 @@ pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult>
     found
 }
 
+/// Result of a symbol lookup.
+#[derive(Clone, Copy)]
+pub struct LookupResult {
+    pub address: usize,
+}
+
+/// Hook a single symbol across all loaded ELF objects by patching their
```
scottgerring, line 877: I think the doc comment (or crate-level module docs) should note when you'd each for `hook_symbol` vs the lower-level building blocks.

 Something like: `hook_symbol` is a one-shot API for cases where you know the target symbol is already loaded when the hook is installed. It only patches libraries loaded at call time; libraries `dlopen`'d afterwards won't be patched. For hooks that need to cover dynamically-loaded libraries, use the `SymbolOverrides` registry in `libdd-profiling-heap-gotter`, which tracks patched libraries and re-patches on `dlopen`.

We should also point out that this API isn't useful for hooking symbols in _dynamically loaded libraries_. Returning to the `__assert_fail` case, if your app loads some other `so` via `dlopen` and _that_ calls `_assert_fail`, that call will not be intercepted.

libdd-gotter/src/elf.rs
gyuheon0h: Ah, good idea. Will help with possibility of diverging also

libdd-gotter/src/elf.rs
gyuheon0h: Yup good idea. 

>I think the doc comment (or crate-level module docs) should note when you'd each for hook_symbol vs the lower-level building blocks

I think you gave me this feedback on the [PR above this one](https://github.com/DataDog/libdatadog/pull/2312/changes), but I should include it here, since the two levels of abstraction exist here too.

scottgerring: Huh! I spent a hot minute trying to work out where i'd written that before, then convinced myself I never had.

libdd-gotter/src/elf.rs
```diff
@@ -42,33 +58,44 @@ const STN_UNDEF: u32 = 0;
...
+    /// the dynsym entry count. Objects with neither hash table fall back
+    /// to a symtab/strtab distance heuristic.
+    ///
     /// # Safety
     /// `info` must point to a valid `dl_phdr_info` from `dl_iterate_phdr`.
-    pub unsafe fn from_phdr(info: &dl_phdr_info) -> Option<Self> {
-        let phdrs = core::slice::from_raw_parts(info.dlpi_phdr, info.dlpi_phnum as usize);
+    pub unsafe fn from_phdr(info: &'a dl_phdr_info) -> Option<Self> {
+        // SAFETY: info is valid for the lifetime of the program.
+        let phdrs = unsafe { slice::from_raw_parts(info.dlpi_phdr, info.dlpi_phnum as usize) };
```
yannham:
```suggestion
        let phdrs = unsafe { slice::from_raw_parts(info.dlpi_phdr, usize::from(info.dlpi_phnum) };
```

libdd-gotter/src/elf.rs
```diff
@@ -42,33 +58,44 @@ const STN_UNDEF: u32 = 0;
...
         let dyn_phdr = phdrs.iter().find(|p| p.p_type == PT_DYNAMIC)?;
         let dyn_begin = (info.dlpi_addr as usize + dyn_phdr.p_vaddr as usize) as *const Elf64_Dyn;
```
yannham: I think this library is 64-bits only, but we should avoid silently-failing-as-conversions in sensitive code like this IMHO.

```suggestion
        let dyn_begin = (usize::try_from(info.dlpi_addr).ok()? + usize::try_from(dyn_phdr.p_vaddr as usize).ok()?) as *const Elf64_Dyn;
```

libdd-gotter/src/elf.rs
```diff
@@ -130,66 +159,121 @@ impl DynamicInfo {
...
+        // - Non-null: guarded by the `is_null()` check in each branch; the null/zero-length case
+        //   returns `&[]` without calling `from_raw_parts`.
+        // - Properly aligned: ELF sections are required by the spec to be aligned to their entry
+        //   size (e.g. `Elf64_Sym` is 8-byte aligned, `Elf64_Rela` is 8-byte aligned). The dynamic
+        //   linker enforces this at load time.
+        // - Valid for `len * size_of::<T>()` reads within a single allocation: the entire section
+        //   is mapped contiguously from the ELF file by the kernel via `mmap`. The size/count
+        //   values come from DT_STRSZ, sym_count (from DT_GNU_HASH or DT_HASH nchain), DT_RELSZ,
+        //   DT_RELASZ, and DT_PLTRELSZ respectively.
+        // - Not mutated for lifetime `'a`: the ELF object is held mapped by `dl_iterate_phdr`'s
```
yannham: This looks like this has to be ensured by the caller of `from_phdr`, right? If so, this should move to be a safety pre-condition of `from_phdr` in the Safety section, and here just we can just say it's a safety precondition. `from_phdr` is public and can't enforce it's called in a specific context (lock or dlopen handle held), so we have to pass the verification burden to the caller.

libdd-gotter/src/elf.rs
```diff
@@ -130,66 +159,121 @@ impl DynamicInfo {
...
+        let relas_slice = if relas.is_null() || relas_count == 0 {
+            &[]
+        } else {
+            slice::from_raw_parts(relas, relas_count)
+        };
+        let jmprels_slice = if jmprels.is_null() || jmprels_count == 0 {
+            &[]
+        } else {
+            slice::from_raw_parts(jmprels, jmprels_count)
+        };
```
yannham: Should we have a small helper to build slices? Like

```rust
unsafe fn try_as_slice<'a, T>(ptr: *const T, len: usize) -> &'a [T] {
    if ptr.is_null() || len == 0 {
        &[]
    } else {
        slice::from_raw_parts(ptr, len)
    }
}
```

libdd-gotter/src/elf.rs
```diff
@@ -130,66 +159,121 @@ impl DynamicInfo {
...
             base_address: base,
         })
     }
 
+    /// Look up the symbol entry and its name at index `idx`.
+    /// Returns the `Elf64_Sym` and its name from the string table,
+    /// or `None` if the index is out of bounds or the name is invalid.
+    pub fn sym_entry(&self, idx: u32) -> Option<(&Elf64_Sym, &CStr)> {
+        let sym = self.symtab.get(idx as usize)?;
+        let off = sym.st_name as usize;
```
yannham:
```suggestion
        let off = usize::from(sym.st_name);
```

libdd-gotter/src/elf.rs
```diff
@@ -198,42 +282,141 @@ impl DynamicInfo {
...
+            h ^= g >> 24;
+        }
+        h &= !g;
+    }
+    h
+}
+
+/// Look up a symbol by name in an object's `DT_HASH` (sysv) table.
+///
+/// [nbucket] [nchain] [bucket[0..nbucket]] [chain[0..nchain]]
```
yannham, line 373: Should this be a code block, given than `[]` is semantically significant in markdown?

libdd-gotter/src/elf.rs
```diff
@@ -545,30 +731,69 @@ impl Drop for PageProtGuard {
...
 /// non-zero-sized definition whose address is not `not_this_symbol`.
 /// Null-sized symbols are ignored so hooks resolve to callable definitions.
+///
+/// Uses `gnu_hash_lookup` for objects with `DT_GNU_HASH`, and falls
+/// back to a bounded linear dynsym scan for objects that only have
+/// `DT_HASH` (sysv).
 pub fn lookup_symbol(name: &str, not_this_symbol: usize) -> Option<LookupResult> {
     let needle = name.as_bytes();
     let mut found: Option<LookupResult> = None;
+    // SAFETY: the callback runs synchronously inside dl_iterate_phdr;
```
yannham: Same remark: I think this should be a safety pre-condition of this function, moving the responsibility to the caller instead. This function alone can't guarantee that it will necessarily be called within `dl_iterate_phdr`.

libdd-gotter/src/elf.rs
```diff
@@ -198,42 +282,141 @@ impl DynamicInfo {
...
+        u32::MAX
+    }
+}
+
+/// Compute the ELF sysv hash used by `DT_HASH` tables.
+/// From <https://refspecs.linuxfoundation.org/elf/elf.pdf>
+pub fn sysv_hash(name: &[u8]) -> u32 {
+    let mut h: u32 = 0;
+    for &c in name {
+        h = (h << 4).wrapping_add(c as u32);
```
yannham: (Because I wondered myself why the explicit `wrapping_add` here vs the spec)

```suggestion
        // The algorithm is described in C in the spec, and C standard defines arithmetic
        // to be defined modulo `2^N` for N-bits unsigned integers; that is, to be wrapping.
        h = (h << 4).wrapping_add(c as u32);
```

yannham (approved):
I haven't flagged them all, but there are a bunch of `as xxx` integer conversions that might silently wrap. Since this code is quite unsafe/sensitive, I think it could be nice to either:

- use `from/into` functions where the conversion is infallible
- use `try_from/try_into` if it is actually fallible
- or justify with a comment if it's fallible in theory but won't in practice, typically stuff like `(h >> 32) as u32`.

If this is annoying for usize conversions (given that it's platform dependent), one possibility would be to write a small helper once and for all:

```rust
#[cfg(pointer_width = "64")]
const fn usize_to_u64(n: usize) -> u64 {
    // Overflow: on 64 bits, the conversion is lossless
    n as u64
}

// etc.
```

It can even be a small trait (just a copy of `Into/From`). It's a bit more verbose, but IMHO as a reader you can see immediately that it's not a risky conversion, while with all these `as` it's hard to tell without knowing the type of the value being converted.

PR #2298, opened 2026-07-29
feat(otel-thread-ctx)!: publish W3C trace flags

morrisonlevi:
# What does this PR do?

Store the W3C trace-flags byte at offset 25 and pass it through the Rust and C APIs during construction and in-place updates.

BREAKING CHANGES: these functions now require a trace_flags argument:
 - `ThreadContext::new`
 -  `ThreadContext::update`
 - `ddog_otel_thread_ctx_new`
 - `ddog_otel_thread_ctx_update`

# Motivation

The spec changed after the context work was initially done in libdatadog.

# Additional Notes

PHP will use this in an upstream PR.

# How to test the change?

Update your APIs for the above functions to include the trace flags, then proceed as normal.

### DIFF
--- a/libdd-otel-thread-ctx/src/lib.rs
+++ b/libdd-otel-thread-ctx/src/lib.rs
@@ -27,8 +27,8 @@
 //! let local_root_span_id = [2u8; 8];
 //!
 //! // First call allocates a record and attaches it.
-//! ThreadContext::new(trace_id, span_id, local_root_span_id, &[(0, "first")]).attach();
-//! ThreadContext::update(trace_id, span_id, local_root_span_id, &[(0, "second")]);
+//! ThreadContext::new(trace_id, span_id, 1, local_root_span_id, &[(0, "first")]).attach();
+//! ThreadContext::update(trace_id, span_id, 1, local_root_span_id, &[(0, "second")]);
 //! ThreadContext::detach();
 //! # }
 //! # #[cfg(not(all(target_os = "linux", any(target_arch = "x86_64", target_arch = "aarch64"))))]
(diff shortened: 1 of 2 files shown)


ivoanjo (approved):
👍 LGTM thanks for the addition

(For extra context -- the W3C trace flags field was a late addition to the spec, we had one leftover reserved byte and it was decided that we'd repurpose it for carrying those flags. The eBPF Profiler doesn't yet do anything with this data, but in the future it might...)

libdd-otel-thread-ctx/src/lib.rs
```diff
@@ -612,7 +628,7 @@ pub mod linux {
             let span_id = [2u8; 8];
             let root_span_id = [3u8; 8];
 
-            ThreadContext::new(trace_id, span_id, root_span_id, &[]).attach();
+            ThreadContext::new(trace_id, span_id, 0, root_span_id, &[]).attach();
```
yannham: Nit: should `0` be a named constant like `NO_FLAGS` or something? To make the default value more discoverable. Another possibility, if those flags are rarely set, is to keep the original interface but have a `with_flags` API. Though it doesn't really work for updates as well...

yannham (approved):

libdd-otel-thread-ctx/src/lib.rs
morrisonlevi: Ergonomic improvements e.g. `NO_FLAGS` are possible but kind of limited because these things go over into FFI as well, and then you end up with having to name something like `DDOG_OTEL_THREAD_CTX_NO_TRACE_FLAGS` or something similarly horrible.

But if you mean for the sake of readability in tests, sure, `NO_FLAGS` is fine.

yannham: Right, agreed on the FFI part - and I suppose it's so idiomatic in C that it probably doesn't deserve a constant on the C side... Yeah, I was thinking test but also the few potential pure Rust API consumers, like dd-trace-rs, or tracers that re-use libdatadog _à la carte_

PR #2306, opened 2026-07-30
chore(rc)!: add DEBUG product

KowalskiThomas:
# What does this PR do?

This PR adds the `DEBUG` product to the list of known Remote Configuration products.

# Motivation

I _need_ this in order to test a new Remote Configuration integration in `dd-trace-py` (see https://github.com/DataDog/dd-trace-py/pull/19357). Because the list of known Remote Configuration products in `dd-trace-py` is defined as a Python enum derived from the native Remote Configuration Product enum defined in `libdatadog`, the only way I can use the  `DEBUG` product in `dd-trace-py` is to add the `DEBUG` product to `libdatadog`.

I don't think this need is a one time thing or specific to me. Anyone trying to add a new integration in a tracing library that uses the native Remote Configuration client would have the same issue. The only way not to need it would be not to use the `libdatadog` client, which isn't the way of the future as far as I can tell.

### DIFF
--- a/libdd-remote-config/src/path.rs
+++ b/libdd-remote-config/src/path.rs
@@ -40,6 +40,7 @@ pub enum RemoteConfigProduct {
     FfeFlags,
     LiveDebugging,
     LiveDebuggingSymbolDb,
+    Debug,
 }
 
 #[derive(Clone)]
--- a/libdd-remote-config/src/path.rs
+++ b/libdd-remote-config/src/path.rs
@@ -36,6 +36,7 @@ pub enum RemoteConfigProduct {
     FfeFlags,
     LiveDebugger,
     LiveDebuggerSymbolDb,
+    Debug,
 }
 
 impl Display for RemoteConfigProduct {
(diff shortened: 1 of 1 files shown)


libdd-remote-config/src/path.rs
```diff
@@ -40,6 +40,7 @@ pub enum RemoteConfigProduct {
     FfeFlags,
     LiveDebugging,
     LiveDebuggingSymbolDb,
+    Debug,
```
yannham, line 43: Should we make this debug-build only, or something to avoid having it go into production? Or is using a libdatadog debug build going to be a pain for what you're trying to do?

KowalskiThomas: I'm not exactly sure what implications making this debug build-only would have... What do we gain by making this debug build-only? If there are strong advantages I can look into it, but otherwise, and based on what I see, I'm really unsure what we'd get from that.

yannham: Mostly avoiding `Debug` leaking past the integration/prototype phase in production. But I'm not entirely sure it fits what you need, so please do as you think is best :+1:

yannham (approved):

PR #2308, opened 2026-07-31
feat: add github shared caching

Aaalibaba42:
# What does this PR do?

Add shared caching

# Motivation

Have shared caching

# Additional Notes

The caching is done from main, so on the branch we won't see a real difference, that's expected

# How to test the change?

Dashboard

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -112,10 +112,14 @@ jobs:
           rustup install ${{ needs.setup.outputs.nightly-version }}
           rustup component add rustfmt --toolchain ${{ needs.setup.outputs.nightly-version }}
       - name: Cache [rust]
+        id: rust-cache
         uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # 2.8.1
         with:
           cache-targets: true # cache build artifacts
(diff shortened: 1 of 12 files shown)


.github/workflows/lint.yml
```diff
@@ -135,10 +139,14 @@ jobs:
           rustup component add clippy --toolchain "$VERSION"
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
       - name: Cache [rust]
+        id: rust-cache
         uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # 2.8.1
         with:
           cache-targets: true # cache build artifacts
           cache-bin: true # cache the ~/.cargo/bin directory
+          shared-key: all-features
```
hoolioh: wouldn't it be better in jobs like clippy to add the rustc version in the shared key? That way each flavour would have their own cache and does not produce contention for the cache entry. Something like:
```
shared-key: clippy-${{ matrix.rust_version }}

Aaalibaba42: I'm a little worried about store size, I seem to recall it's limited, but I agree it's way better like this

hoolioh (approved):

PR #2309, opened 2026-08-03
feat(data-pipeline)!: add agentless stats export

paullegranddc:
# Motivation

Perform stats aggregation and export directly to the public intake when agentless tracing is enabled

# Changes

* Add paramateres on the TraceExporterBuilder
  * agentless_stats_endpoint - the fully formed URL. The API key is derived from the agentless export parameter
* Refactor the stats exporter to add a path that wraps tje ClientStatsPayload into a top-level StatsPayloas
* Hardcode the agent_version to libdd version + a libdatadog suffix to be able to differentiate metrics origin


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -813,6 +802,10 @@ impl<
             }
         }
 
+        if let Some(ref config) = self.agentless_config {
+            return self.send_agentless_traces_inner(traces, config).await;
```
hoolioh: Shouldn't we check if the trace payload is empty? It could happen that all those chunks were dropped and we still will be sending an empty payload

libdd-data-pipeline/src/agentless/stats.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+/// The libdatadog version reported in the `StatsPayload.agent_version` field.
```
hoolioh: Is the intention to get the workspace or crate version? IMO the crate version is fine but the doc is misleading.

libdd-data-pipeline/src/trace_exporter/mod.rs
hoolioh, line 608: Shouldn't we also skip checking the agent info if agentless is enabled?

paullegranddc: We already do not launch the agent info fetcher, so `check_agent_info` is going to stop at the first line of the check.

I'll add the skip check when agentless is enabled for uniformity, but this is not functionnal. It will just be a very very marginal perf gain (one atomic ptr dereference)

libdd-data-pipeline/src/agentless/stats.rs
paullegranddc: crate version. I think that's the most interesting one for us 🤔 
Or I could put the tracer version maybe?

hoolioh: Hmm you have a point there, maybe from the intake point of view it makes more sense to report the tracer version rather than the crate version.

paullegranddc:
@codex review

ekump (commented):
This doesn't include fuzz and benchmark jobs. Please create tickets for those under this milestone to update those as well. You'll need to update the images in benchmarking-platform. Also please update the macOS AMI in ci-platform-machine-images before merging this PR. Please attach performance changes to the description of the PR

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -236,9 +338,46 @@ impl<
...
+        obfuscated: bool,
     ) -> anyhow::Result<()> {
         let sequence = self.sequence_id.fetch_add(1, Ordering::Relaxed);
+        let request = match &self.destination {
+            StatsDestination::Agent { endpoint } => {
+                self.build_agent_request(endpoint.clone(), sequence, buckets, obfuscated)?
+            }
+            StatsDestination::Agentless(target) => {
+                build_agentless_request(&self.meta, sequence, buckets, target)?
+            }
```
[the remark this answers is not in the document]
ekump: Not an issue. That header is only relevant for sending to the agent.

libdd-trace-stats/src/stats_exporter.rs
ekump, line 560: I know we send `container_id` as a header. But, does the stats endpoint know about the header? Should we be setting `container_id` for agentless? [We do for agentless trace export](https://github.com/DataDog/libdatadog/pull/2309/changes#diff-655ab2a3c4f35c19b2f40933143c5c7ed7668afcfc9b8de214b70f5394d525d1L139-L141).

libdd-trace-utils/src/stats_payload_encoder/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Encoding utilities for the top-level [`pb::StatsPayload`] message.
+//!
+//! This module wraps a single client [`pb::ClientStatsPayload`] into a
+//! [`pb::StatsPayload`] and serializes it as msgpack, matching the wire format
+//! the Datadog Agent uses when sending stats to the `/api/v0.2/stats` intake.
+//!
+//! Unlike the Agent's stats writer, we do **not** split payloads at 4000
+//! entries: a single client cannot exceed that limit, so we always emit one
```
ekump: > a single client cannot exceed that limit

How is that enforced?

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -815,6 +839,118 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
...
+            // `dropped`, `dropped_bytes` (queue-full counts).
+            let stats_exporter = StatsExporter::<C, _>::new_agentless(
+                bucket_size,
+                concentrator.clone(),
+                meta,
+                target,
+                capabilities.clone(),
+                #[cfg(feature = "telemetry")]
+                None,
+                dogstatsd.clone(),
```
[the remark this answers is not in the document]
ekump: Additionally, it's consistency issue. We respect `health_metrics_enabled` for agent mode, we should also do it for agentless.

libdd-data-pipeline/Cargo.toml
```diff
@@ -87,6 +87,14 @@ tokio = { version = "1.23", features = [
 duplicate = "2.0.1"
 zstd = { version = "0.13", default-features = false }
 
+[[test]]
+name = "test_agentless_stats"
+required-features = ["stats-obfuscation"]
```
ekump, line 101: I think we have a real test coverage gap. This isn't a default feature, which means it only runs on CI in the --all-features job on the Github [test workflow](https://github.com/DataDog/libdatadog/blob/4554db3237470e413ebe4e907e39b89af255e2a9/.github/workflows/test.yml#L166). It's not running in Gitlab.

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -452,6 +638,59 @@ mod tests {
         mock.assert_async().await;
     }
 
+    #[cfg_attr(miri, ignore)]
+    #[tokio::test]
+    async fn test_send_agentless_stats() {
```
ekump, line 695: Is it feasible to have version of [test_send_stats_fail](https://github.com/DataDog/libdatadog/pull/2309/changes#diff-8b66f0533e46aaa4be373d31458efa65f7aebc07fac8cc743e668fc2afefebe6R696) for agentless that also covers retry logic?

libdd-data-pipeline/src/agentless/stats.rs
bantonsson: Is there any way for the intake to know that it's not a _normal_ agent sending the stats? Will it mess things up somewhere if we start reporting our versions that are not aligned to proper agent versions?

libdd-trace-utils/src/stats_payload_encoder/mod.rs
paullegranddc: It was not enforced. I assumed it was true because of the cardinality limits but it turns out that I am wrong

Splitting is correctly implemented now

libdd-data-pipeline/src/agentless/stats.rs
paullegranddc: No, this should just be used for internal backend metrics

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -815,6 +839,118 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
...
         }
 
+        // Agentless stats: when a stats intake endpoint and stats computation are both
+        // configured, start the concentrator unconditionally (bypass the agent gate, which is
+        // never reached in agentless mode) and send the top-level `StatsPayload` directly to the
+        // intake. Mutually exclusive with OTLP stats above (both set `stats` to `Enabled`).
+        if let (Some(stats_url), Some(bucket_size)) = (
+            self.agentless_stats_endpoint.as_ref(),
+            self.stats_bucket_size,
+        ) {
```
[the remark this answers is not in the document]
paullegranddc: resolved

libdd-data-pipeline/Cargo.toml
paullegranddc: What stops us from enabling `all-features` on the gitlab CI ?

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -831,6 +859,121 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
...
+                    ..Default::default()
+                },
+            )));
+            let concentrator = Arc::new(Mutex::new(SpanConcentrator::new(
+                bucket_size,
+                web_time::SystemTime::now(),
+                span_kinds,
+                self.peer_tags.clone(),
+                self.stats_cardinality_limits,
+                vec![],
```
ekump: shouldn't you pass along `self.additional_metric_tag_keys` here? OTLP and agent modes do. The builder allows you to set it in agentless mode.

libdd-data-pipeline/src/trace_exporter/builder.rs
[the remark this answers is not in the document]
ekump: I believe this comment is still relevant

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -815,6 +839,118 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
...
+                    TraceExporterError::Builder(BuilderErrorKind::InvalidConfiguration(
+                        "Agentless stats export requires an API key; call set_agentless_endpoint \
+                         before set_agentless_stats_endpoint"
+                            .to_string(),
+                    ))
+                })?;
+
+            let stats_uri = parse_uri(stats_url).map_err(|e: anyhow::Error| {
+                TraceExporterError::Builder(BuilderErrorKind::InvalidUri(e.to_string()))
+            })?;
```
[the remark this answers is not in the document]
ekump: This is valid. We expect the full URI with scheme, but we aren't erroring on a relative path in the builder.

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -229,16 +333,74 @@ impl<
...
-        #[cfg_attr(not(feature = "stats-obfuscation"), allow(unused))] obfuscated: bool,
+        obfuscated: bool,
     ) -> anyhow::Result<()> {
+        let groups = split_stats_buckets(buckets, MAX_GROUPED_STATS_PER_PAYLOAD);
+        let split = groups.len() > 1;
+        // All fragments of one flush share a single sequence id, as the Agent does.
         let sequence = self.sequence_id.fetch_add(1, Ordering::Relaxed);
+        for group in groups {
+            self.send_single_payload(group, obfuscated, split, sequence)
+                .await?;
```
ekump: The `?` is going to abort the whole loop if one group fails, which is different behavior to the agent. Should we instead collect the errors and return it after the for loop so each group can attempt to send?

libdd-data-pipeline/src/agentless/stats.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+/// The `StatsPayload.agent_version` value: the tracer version passed to the
+/// exporter suffixed with the language, so the backend can tell libdatadog
+/// tracers from the Agent.
+pub fn agentless_stats_version(tracer_version: &str, language: &str) -> String {
+    format!("{tracer_version}-{language}")
```
ekump, line 7: >/// The `StatsPayload.agent_version` value: the tracer version passed to the
/// exporter suffixed with the language, so the backend can tell libdatadog
/// tracers from the Agent.

In a previous [commit](https://github.com/DataDog/libdatadog/blob/4554db3237470e413ebe4e907e39b89af255e2a9/libdd-data-pipeline/src/agentless/stats.rs#L8-L9) you had a `libdatadog` suffix, but that's gone now. Is that intentional? If you aren't going to append `libdatadog` then the rustdoc is inaccurate now.

libdd-data-pipeline/Cargo.toml
ekump: We should run a matrix of features on CI (or at least default and all-features). I don't want to overly complicate this PR or blowout our CI pipeline times. For now, in Gitlab you could just add a step that builds the agentless features and runs only the tests gated on those features on each OS/arch.

ekump (commented):
Added some new comments in addition to many of my existing unaddressed comments

libdd-data-pipeline/src/agentless/stats.rs
paullegranddc: In a precedent cmment with Julio, we said it'd be more interesting to know the actual trace, which gives us the libdatadog version rather than just us.
Fixed the comment

libdd-trace-stats/src/stats_exporter.rs
paullegranddc: Added a test

libdd-data-pipeline/src/trace_exporter/builder.rs
paullegranddc: Done

libdd-data-pipeline/src/trace_exporter/builder.rs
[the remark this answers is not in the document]
paullegranddc: Added some validation

libdd-trace-stats/src/stats_exporter.rs
paullegranddc: Yeah, I changed it so that we do all exports, and return the last error + other ones as additional context.

I was a bit afraid of hitting timeouts repeatedly for each of the split payload, leading to a very high total time but since there is the cardinality limit on buckets the number of split playload per flush is bounded and this runs in a background task so it should be oka

libdd-data-pipeline/Cargo.toml
ekump: @paullegranddc, we have all-features coverage in github, which I think is ok for now. Just create a Jira ticket to follow up for gitlab.

ekump (approved):
made a commit to add missing test coverage for `additional_metric_tags`. LGTM

libdd-data-pipeline/Cargo.toml
paullegranddc: Adding a ticket

PR #2310, opened 2026-08-03
feat(sidecar)!: support appsec helper-rust integration with sidecar

cataphract:
See https://github.com/DataDog/dd-trace-php/pull/3725

Also fix zigbuild windows builds, for easier testing of windows when running Linux (zigbuild + wine).


datadog-sidecar-ffi/src/lib.rs
```diff
@@ -1781,6 +1781,72 @@ pub unsafe extern "C" fn ddog_sidecar_send_garbage(transport: &mut Box<SidecarTr
...
+    transport: &mut Box<SidecarTransport>,
+    session_id: ffi::CharSlice,
+    client_id: u64,
+    data: ffi::CharSlice,
+) -> AppsecCResponse {
+    match blocking::send_appsec_message(
+        transport,
+        session_id.as_bytes(),
+        client_id,
+        data.as_bytes(),
```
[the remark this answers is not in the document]
bwoebi: sidecar-ffi is abort, no handling needed.

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -158,6 +175,19 @@ impl ConnectionSidecarHandler {
...
 
+        // Notify the appsec helper of the per-client disconnect before session
+        // teardown. This lets it evict the exact ClientKey for this connection
+        // without touching sibling workers that share the same session_id.
+        #[cfg(unix)]
+        {
+            let client_id_opt = *self.appsec_client_id.lock_or_panic();
+            if let (Some(session_id), Some(client_id)) = (self.session_id.get(), client_id_opt) {
+                if let Some(appsec) = self.server.appsec.as_ref() {
+                    appsec.disconnect(session_id, client_id);
```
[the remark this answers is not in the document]
bwoebi: Won't happen.

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -251,6 +306,17 @@ impl SidecarServer {
...
         debug!("Successfully shut down session: {}", session_id);
+
+        // Failsafe: evict any remaining appsec clients for this session.
+        // cleanup() already fires per-client notifications (client_id != 0)
+        // before decrementing the session counter, so in the normal path this
+        // is a no-op. client_id == 0 signals a session-wide sweep.
+        #[cfg(unix)]
+        {
+            if let Some(appsec) = self.appsec.as_ref() {
+                appsec.disconnect(session_id, 0);
```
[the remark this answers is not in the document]
cataphract: True, I suppose, but rechecking the session counter would not eliminate the hazard. There could still be a TOCTOU race, where another connection increments the session counter after our checking it in `stop_session`. So for this to be safe we'd probably need to hold both the sessions lock and the counter lock, which raises the possibility of accidentally deadlocking.

A similar problem seems to already exist where the session counter can be incremented after calling `stop_session` but before the `SessionInfo` is removed.

I think the solution would be to combine both locks. Instead of:

```
    /// A `Mutex` guarded `HashMap` that stores active sessions.
    sessions: Arc<Mutex<HashMap<String, SessionEntry>>>,
    /// A `Mutex` guarded `HashMap` that keeps a count of each session.
    session_counter: Arc<Mutex<HashMap<String, u32>>>,
```

have something like `Arc<Mutex<HashMap<String, SessionEntry>>`

where

```
enum SessionEntry {
    Active {
        connection_count: u32,
        session: SessionInfo,
    },
    // block the session from being recreated until shutdown completes
    ShuttingDown(tokio::sync::watch::Receiver<bool>)
}
```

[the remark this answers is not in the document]
cataphract: Won't do because of pending refactoring.

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -1178,6 +1244,98 @@ impl SidecarInterface for ConnectionSidecarHandler {
...
+            //  stored=Some(P), incoming=0  → extension abandons client: evict P, clear
+            //  stored=Some(P), incoming=N≠P → bug: log, evict P, tell client to reconnect
+            let (evict, reject) = match (*self.appsec_client_id.lock_or_panic(), client_id) {
+                (None, 0) => (None, false),
+                (None, _) => {
+                    warn!(
+                        "appsec: extension sends client_id {client_id} without this \
+                        connection having memory of it (sidecar restarted?)"
+                    );
+                    (Some(client_id), true)
```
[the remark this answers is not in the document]
cataphract: changed it so it no longer evicts

datadog-sidecar/src/service/sidecar_server.rs
```diff
@@ -1178,6 +1244,118 @@ impl SidecarInterface for ConnectionSidecarHandler {
...
+        #[cfg(not(unix))]
+        {
+            _ = (log_file_path, log_level);
+            false
+        }
+    }
+
+    async fn send_appsec_message(
+        &self,
+        session_id: String,
```
bwoebi: What's this for? Why not simply rely on `self.session_id`?

It would make more sense to assert that the first message on a connection is set_session_config than resubmitting the session id here.

datadog-sidecar/src/appsec.rs
```diff
@@ -0,0 +1,336 @@
...
+
+/// Registers the AppSec backend linked by the sidecar's embedding application.
+///
+/// Only the first registration in a process is retained. This allows inverting
+/// the dependency between sidecar and appsec helper-rust.
+pub fn register_backend_factory(factory: AppSecBackendFactory) {
+    _ = APPSEC_BACKEND_FACTORY.set(factory);
+}
+
+/// Publishes one AppSec backend and coordinates its one-way lifecycle.
```
bwoebi, line 26: I find this construction quite peculiar.

What you want, is essentially a `trait AppSecManager { ... }` with a `Arc<tokio::sync::OnceCell<dyn AppSecManager>>` with a `.get_or_init(async || APPSEC_BACKEND_FACTORY())` (sort of)?
And then have a shutdown function on the trait which expects `self`, and using `Arc::into_inner()` in shutdown.

To support no-op (i.e. failed startup) mode, you can just have an impl with trait methods which do nothing / return disconnect.

Like:
```
    // Calling send_message/disconnect is allowed to race with shutdown. It either
    // completes successfully or observes a closed channel in the backend and fails,
    // returning a message requesting a disconnect.
```
is just not going to happen anyway - when appsec shutdown happens, all sidecar connections are already shut down.

This is an overly defensive design with unncessary complexity from there.

datadog-sidecar/src/log.rs
bwoebi, line 1: Do we really want a dedicated way to log?
For me, common extension also should mean common log. (Certainly, we can choose different log levels per component, but log file should be the same).

Is there a major reason to split this?

datadog-sidecar/src/service/sidecar_server.rs
cataphract: Yeah, it's a leftover from the code before the integration. I'll remove it.

datadog-sidecar/src/appsec.rs
cataphract: > trait AppSecManager { ... } with a Arc<tokio::sync::OnceCell<dyn AppSecManager>> with a .get_or_init(async || APPSEC_BACKEND_FACTORY()) (sort of)

Yes, I think so, for the initialization part. Of course, I'd rather not replace

```
    appsec: Option<Arc<AppSecManager>>,
 ```
 
 on `struct SidecarServer` with _that_, plus expose the factory and force it to use `get_or_init`. So in practice this would be done in a public encapsulating type inside the appsec.rs that would save callers from it. In fact, I had an earlier version more or less along those lines.

The "peculiarity" (it's just a state machine after all) is all about the shutdown. This is already a bit of a simplified version -- previously I had  also synchronization around the shutdown (not just initialization) to ensure no messages were sent to the backend after a shutdown was requested -- which I then replaced with that comment to make it a bit simpler. And if OnceCell can't help all the lifecycle state I need (it only does uninitialized and initialized), and it need to track it in _other_ fields, AND if I need an extra wrapper type... well, at some point this design became (to me) clearer.

Of course, this all hinges on the complexity for shutdown actually being necessary -- for instance, it handles cases where shutdown is called before and during initialization, where shutdown is called more than once, discards messages after shutdown or after failed initialization, etc.

> And then have a shutdown function on the trait which expects self, and using Arc::into_inner() in shutdown.

I don't think that this, at least literally, would work. Maybe `Mutex<Option<Box<dyn AppSecManager>>>`, and then:

```
pub async fn shutdown(server) {
    if let Some(m) = server.appsec.lock().unwrap().take() {
        m.shutdown().await;   // fn shutdown(self: Box<Self>) -> impl Future
    }
}
```

but then you have the mutex

File: datadog-sidecar/src/log.rs
Comment (cataphract) [reply datadog-sidecar/src/log.rs:1]: For now it keeps compatibility with whatever configuration the clients already have, plus the testing infrastructure. But TBH the main reason is that I like to see them separately from all the noise, with my own format.

In any case, the default will be that the appsec messages go to the sidecar. Maybe it's not clear from this libdatadog part of the changes, but when the value is `<sidecar log>` (the default), it all goes to the sidecar log. Or at least should.

File: datadog-sidecar/src/appsec.rs (hunk 6 above)
Comment (bwoebi) [reply datadog-sidecar/src/appsec.rs:26]: > Of course, this all hinges on the complexity for shutdown actually being necessary -- for instance, it handles cases where shutdown is called before and during initialization, where shutdown is called more than once, discards messages after shutdown or after failed initialization, etc.

Yes, that's basically what I'm saying. I don't think most of that is necessary to handle at all, what's the actual use case? The sidecar as is literally cannot shut down while there are currently actively processed payloads, which initialization depends on.

And no, `Arc::into_inner()` should just work, it gives you an owned value which you can then just orderly shut down. Because, at least assuming the shutdown in the entrypoint shutdown is the only one, that should work out fine, no?

File: datadog-sidecar/src/log.rs
Comment (bwoebi) [reply datadog-sidecar/src/log.rs:1]: Okay. (for now at least.)

File: datadog-sidecar/src/appsec.rs (hunk 6 above)
Comment (cataphract) [reply datadog-sidecar/src/appsec.rs:26]: Arc::into_inner() requires that no other shared owner exist. I don't think that can be guaranteed with all the cloning of `SidecarServer`.

But I'll otherwise try to simplify this

Comment (cataphract) [reply datadog-sidecar/src/appsec.rs:26]: I've pushed a simplified version in f5a4ba04c

File: datadog-sidecar/src/service/sidecar_interface.rs [hunk 7]
```diff
@@ -247,6 +247,21 @@ pub trait SidecarInterface {
...
+    /// cannot overtake initialization.
+    async fn ensure_appsec_started(log_file_path: Vec<u8>, log_level: String) -> bool;
+
+    /// Forwards an AppSec message from the PHP extension to the registered helper.
+    ///
+    /// Returns the response bytes from the helper and a flag indicating whether
+    /// the extension session should be disconnected.
+    async fn send_appsec_message(
+        client_id: u64,
+        #[ClientType(&'request [u8])] data: Vec<u8>,
```
bwoebi, line 313: Note: The way this is architected is not optimal, and we really _could_ just take `&[u8]` on server side too. But it'd require more changes to libdd_ipc. Not needed to do right now.

bwoebi (approved):
Looks good for now. I think we should eventually port the appsec backend code to libdatadog as well, and then we can clean this up even more.

cataphract:
Blocked by https://github.com/DataDog/libdatadog/pull/2410

PR #2312, opened 2026-08-04
feat(gotter)!: add fn to hook symbol but exclude self

gyuheon0h:
Stacked above [feat(got-hook)!: add DT_HASH fallback, relocation type guard, and hook_symbol](https://github.com/DataDog/libdatadog/pull/2297)

[PROF-15630](https://datadoghq.atlassian.net/jira/software/c/projects/PROF/boards/11?selectedIssue=PROF-15630)
# What does this PR do?

PHP needs to be able to hook libs and bins but skip its own exported fn. 

# Motivation

# Additional Notes

# How to test the change?

[PROF-15630]: https://datadoghq.atlassian.net/browse/PROF-15630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libdd-gotter/src/elf.rs
+++ b/libdd-gotter/src/elf.rs
@@ -135,13 +135,8 @@ impl<'a> DynamicInfo<'a> {
         // On 64-bit (this crate's cfg gate), Elf64_Addr (u64) -> usize is lossless.
         let base = u64_to_usize(info.dlpi_addr);
         let dyn_begin = (base + u64_to_usize(dyn_phdr.p_vaddr)) as *const Elf64_Dyn;
-        let containing_load_segment_end = |addr: usize| -> Option<usize> {
-            phdrs.iter().filter(|p| p.p_type == PT_LOAD).find_map(|p| {
-                let start = base.checked_add(u64_to_usize(p.p_vaddr))?;
-                let end = start.checked_add(u64_to_usize(p.p_memsz))?;
-                (addr >= start && addr < end).then_some(end)
-            })
-        };
+        let containing_load_segment_end =
+            |addr: usize| -> Option<usize> { find_containing_load_segment(phdrs, base, addr) };
         let correct = |a: u64| -> usize {
             let a = u64_to_usize(a);
             if a > base {
@@ -587,24 +582,36 @@ pub fn check_sym(sym: &Elf64_Sym) -> bool {
         matches!(stt, 0 | 1 | 2 | 10)
 }
 
+/// Find the `PT_LOAD` segment in `phdrs` that contains `addr` (using
+/// `base` as the ELF object's load bias) and return the segment's end
+/// address.
+///
+/// on non-PIE executables `base` is 0 but the segments still have
+/// the correct absolute virtual addresses once `base` is added.
+fn find_containing_load_segment(
+    phdrs: &[libc::Elf64_Phdr],
+    base: usize,
+    addr: usize,
+) -> Option<usize> {
+    phdrs.iter().filter(|p| p.p_type == PT_LOAD).find_map(|p| {
+        let start = base.checked_add(u64_to_usize(p.p_vaddr))?;
+        let end = start.checked_add(u64_to_usize(p.p_memsz))?;
+        (addr >= start && addr < end).then_some(end)
+    })
+}
+
 /// Check whether `addr` falls within any of a loaded ELF object's
-/// `PT_LOAD` segments. Works regardless of PIE vs non-PIE: on non-PIE
-/// executables `dlpi_addr` is 0 but the segments still have the correct
-/// absolute virtual addresses once `dlpi_addr` is added.
+/// `PT_LOAD` segments.
 ///
 /// # Safety
 /// `info` must point to a valid `dl_phdr_info` from `dl_iterate_phdr`.
 unsafe fn phdr_contains_addr(info: &dl_phdr_info, addr: usize) -> bool {
+    // SAFETY: caller guarantees `info` is a valid `dl_phdr_info` for a
+    // currently-loaded ELF object. `dlpi_phnum` is u16, so the
+    // conversion to usize is lossless.
     let phdrs = slice::from_raw_parts(info.dlpi_phdr, info.dlpi_phnum as usize);
-    let base = info.dlpi_addr as usize;
-    phdrs.iter().any(|p| {
-        if p.p_type != PT_LOAD {
-            return false;
-        }
-        let start = base + p.p_vaddr as usize;
-        let end = start + p.p_memsz as usize;
-        addr >= start && addr < end
-    })
+    let base = u64_to_usize(info.dlpi_addr);
+    find_containing_load_segment(phdrs, base, addr).is_some()
 }
 
 /// Visit each loaded ELF object once. `is_exe` is true only on the
@@ -1111,6 +1118,7 @@ unsafe fn patch_got_entries(
 mod tests {
     use super::*;
     use tempfile::TempDir;
+
     #[test]
     fn page_prot_guard_finds_original_mapping_protection() {
         let guard = PageProtGuard {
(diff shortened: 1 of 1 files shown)


gyuheon0h:
* **#2312** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2312?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2312?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2297** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2297?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2282** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2282?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

libdd-gotter/src/elf.rs
```diff
@@ -821,8 +852,17 @@ pub unsafe fn hook_symbol(symbol_name: &CStr, hook_fn: usize, orig_out: &mut usi
         if lib_name.contains("linux-vdso") || lib_name.contains("/ld-linux") {
             return false;
         }
+
+        // Skip the hook's own library if requested.
+        if let Some(base) = skip_base {
+            if info.dlpi_addr as usize == base {
```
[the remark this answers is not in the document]
gyuheon0h: Not sure if this is common enough to do. Php tracer's GOT patching doesn't seem to care for this. WDYT @realFlowControl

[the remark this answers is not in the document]
scottgerring: Non-PIE executables are uncommon but still in use; we should handle this.

libdd-gotter/src/elf.rs
```diff
@@ -809,12 +836,49 @@ pub struct LookupResult {
 /// `hook_fn` must point to a function with the same calling convention
 /// and signature as the symbol being hooked. The patching is permanent.
 pub unsafe fn hook_symbol(symbol_name: &CStr, hook_fn: usize, orig_out: &mut usize) -> bool {
+    hook_symbol_impl(symbol_name, hook_fn, orig_out, None)
+}
+
+/// Like [`hook_symbol`], but skips the library that contains `hook_fn`.
```
scottgerring, line 961: As this is all rather esoteric, I think the doc comment needs to explain the motivation and usage pattern more clearly.

As I understand it, `excluding_self` is an ergonomic improvement that enables a simpler hook pattern - because the hook library's own GOT isn't patched, the hook can forward to the original through normal linkage rather than capturing a stored function pointer. 

If that's the intent, can we spell it out in the doc comment? Something like:

> Use this instead of `hook_symbol` when the hook function forwards calls to the original via normal linkage (e.g. `libc::malloc(size)`) rather than through the stored `orig_out` pointer. By skipping the hook's own library during patching, its GOT entries remain pointed at the real symbol, so normal calls from within the hook don't recurse.


I think it would also be worth noting in the module-level docs where you'd reach for the "lower-level lever". In your case you know that the symbol you want to override is there when the code runs, so a one-shot patch is perfect. In other cases you need to patch symbols in libraries as they are loaded, and the underlying `SymbolOverrides` machinery is necessary - e.g. the allocation stuff.

libdd-gotter/src/elf.rs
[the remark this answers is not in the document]
gyuheon0h: Got it. Im gonna be honest, I am not familiar with PIE, but I came to this commit: 520be6b3a5a627d19e66b5fb1c6e6abb7210cac0 off of reading https://one2bla.me/the-dark-arts/exploit-mitigations/position-independent-executables.html and some stack overflow blogs, with the help of my buddy claude. 

Would appreciate special review over the above commit. Thanks :D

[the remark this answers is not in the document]
scottgerring: Hey hey - had a look - I like that it is a generalisation and not a specialisation for non-PIE!
Comments on the commit itself

allanchen511:
nice

libdd-gotter/src/elf.rs
```diff
@@ -587,6 +587,26 @@ pub fn check_sym(sym: &Elf64_Sym) -> bool {
...
 }
 
+/// Check whether `addr` falls within any of a loaded ELF object's
+/// `PT_LOAD` segments. Works regardless of PIE vs non-PIE: on non-PIE
+/// executables `dlpi_addr` is 0 but the segments still have the correct
+/// absolute virtual addresses once `dlpi_addr` is added.
+///
+/// # Safety
+/// `info` must point to a valid `dl_phdr_info` from `dl_iterate_phdr`.
+unsafe fn phdr_contains_addr(info: &dl_phdr_info, addr: usize) -> bool {
```
scottgerring, line 608: IIRC we already have `containing_load_segment_end` in `from_phdr` (introduced in #2297) doing the same computation but with `checked_add` and `u64_to_usize` - can we extract a shared helper? This reimplements without the bounds checking

scottgerring: Maybe something like:

```rust
  fn find_containing_load_segment(phdrs: &[Elf64_Phdr], base: usize, addr: usize) -> Option<usize> {
```

Then it'll be nicely testable too

libdd-gotter/src/elf.rs
```diff
@@ -1308,6 +1407,50 @@ mod tests {
         assert!(!is_got_pointer_reloc(u32::MAX));
     }
 
+    /// Verify that `hook_symbol_excluding_self` skips the library
+    /// containing the hook function. Uses `phdr_contains_addr` to identify
+    /// the hook's own library.
```
scottgerring:
```suggestion
    /// the hook's own library. Due to rust's testing infrastructure, this test only
    /// covers PIE executables. 
```

libdd-gotter/src/elf.rs
```diff
@@ -803,9 +823,31 @@ pub fn is_got_pointer_reloc(r_type: u32) -> bool {
...
+    lookup_symbol_impl(name, not_this_symbol, None)
+}
+
+/// Like [`lookup_symbol`], but also skips the library whose `PT_LOAD`
+/// segments contain `skip_addr`.
+///
+/// Used by [`hook_symbol_excluding_self`] to ensure `orig_out` resolves
+/// to the external definition rather than a same-name export from the
+/// hook's own library. Uses segment containment rather than `dlpi_addr`
+/// comparison so it works for non-PIE executables (where `dlpi_addr` is 0).
```
scottgerring: Suggest removing TMI / implementation detail 
```suggestion
/// hook's own library. 
```

libdd-gotter/src/elf.rs
gyuheon0h: Ooo great idea. Thanks for the catch

libdd-gotter/src/elf.rs
```diff
@@ -899,12 +955,54 @@ pub enum HookError {
...
 
-    let result = lookup_symbol(name_str, hook_fn).ok_or(HookError::SymbolNotFound)?;
+    // Resolve the original symbol, excluding both the hook_fn address
+    // and (if excluding self) the entire hook library so we don't
+    // accidentally resolve to a different export from the same object.
+    let result = if let Some(addr) = skip_addr {
+        lookup_symbol_excluding_addr(name_str, hook_fn, addr)
+    } else {
+        lookup_symbol(name_str, hook_fn)
+    }
```
yannham, line 1004: Nit: isn't this just

```suggestion
    let result = look_symbol_impl(name_str, hook_fn, skip_addr)
```

yannham (approved):

libdd-gotter/src/elf.rs
gyuheon0h: Lol yeah. I guess my I thought my version made it explicit to read but no strong opinions on this.

PR #2313, opened 2026-08-04
feat(data-pipeline)!: add flush_and_close to the trace buffer

brettlangdon:
# What does this PR do?

Add `TraceBuffer::flush_and_shutdown(timeout)`, a synchronous method to flush the trace buffer and reject further writes.

# Motivation

dd-trace-py stops and joins its background export thread before every fork. The trace buffer had no equivalent operation: force_flush only queues a flush request, and `TraceBuffer::new()` gives the caller a `Sender`, not a `WorkerHandle` or the `SharedRuntime`. Without `flush_and_shutdown`, dd-trace-py could not get a bounded, synchronous guarantee that buffered chunks reach the exporter and that no new chunk can arrive before a fork.

`flush_and_shutdown` triggers a flush of the current batch, blocks until the exporter exports that batch or timeout elapses, then marks the channel shut down. The channel shuts down even on timeout, so a caller that must tear down before a fork always gets the "no further chunks" guarantee, whether or not the flush finished in time.

This also fixes a separate drop bug in `Worker::shutdown`: `PausableWorker::pause`'s biased `select!` cancels the trigger-and-run loop without letting it hand off a pending batch, so `Worker::shutdown` used to ack without exporting. It now drains and exports the pending batch before it acks.

# Additional Notes

### Changes

- `Sender::trigger_flush` returns the `BatchGeneration` of the batch it flushes, so a caller can wait for that exact batch.
- `Sender::flush_and_shutdown` / `TraceBuffer::flush_and_shutdown(timeout)`: flush, wait, then shut down. Shuts down even when the flush times out.
- `Receiver::drain_pending`: take whatever sits in the batch, regardless of flush_needed. `Worker::shutdown` calls this before it acks, so a paused worker no longer drops its last batch.
- `Waiter::mark_shutdown` and `TraceExporterWorker::export_and_ack`: extracted from duplicated code paths.

# How to test the change?

- New tests cover: `flush_and_shutdown` exporting a pending chunk, marking the channel shut down even when the flush times out, and returning `AlreadyShutdown` on a second call.
- `cargo test -p libdd-data-pipeline trace_buffer`, `cargo +stable clippy -p libdd-data-pipeline --all-targets -- -D warnings`, and `cargo +nightly-2026-07-26 fmt --all -- --check all` pass.


libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -391,6 +435,7 @@ fn channel<T>(
         state: Mutex::new(SharedState {
             flush_needed: false,
             last_flush_generation: BatchGeneration::default(),
+            accepting_chunks: true,
             has_shutdown: false,
```
Eldolfin: NIT: Instead of having 2 bools here we could instead have an enum with 3 states: `Running, Stopping, Stopped` to avoid the inconsistent state `has_shutdown: true, accepting_chunks: true` being possible.

libdd-data-pipeline/src/trace_buffer/mod.rs
```diff
@@ -484,11 +541,42 @@ impl<T> Sender<T> {
...
+        self.wait_flush_done(flush_gen, timeout)
+    }
+
+    /// Flush and wait as [`Sender::flush_and_wait`] does, then mark the channel as shut down so
+    /// that the channel refuses further chunks.
+    ///
+    /// `flush_and_shutdown` marks the channel as shut down even when the flush times out. A
+    /// caller that tears down the runtime before a fork needs the guarantee that no chunk can
+    /// be queued after this call returns, whether or not the flush finished in time.
+    fn flush_and_shutdown(&self, timeout: Option<Duration>) -> Result<(), TraceBufferError> {
```
ekump: minor: This function name might be confusing since it doesn't actually shutdown, like other functions with `shutdown` in them. The caller is handling that with `worker.stop()`, right? Would `flush_and_close` be a better name?

ekump (approved):
one suggestion for a function name change, but LGTM

libdd-data-pipeline/src/trace_buffer/mod.rs
brettlangdon: thanks this is good feedback, I have updated the function names/enums/etc to be more aligned on differences between "shutdown" and "close"

PR #2314, opened 2026-08-04
fix(libdd-trace-utils): apply SpanLink flags masking when v0.5 json encoding

brettlangdon:
# What does this PR do?

A tracer flags a `SpanLink` with a 32-bit flags value. Bit 31 is a sentinel: it marks that the tracer explicitly set the sampling decision, as opposed to leaving `flags` at its default of zero. The remaining bits carry the sampling decision itself (bit 0: kept or dropped).

# Motivation

v0.4 and v0.5 disagree on whether the wire value includes this sentinel.

- In v0.4's native msgpack encoding, dd-trace-py adds the sentinel bit before it writes flags. A kept link becomes `0x8000_0001`; a link the tracer explicitly dropped becomes `0x8000_0000`.
- v0.5 has no native span-link field, so it encodes links as a JSON array under `meta["_dd.span_links"]`. dd-trace-py's v0.5 JSON encoder never adds the sentinel. A kept link's flags in this JSON is plain 1; the sentinel bit never appears there.

libdatadog's v0.5 encoder builds this JSON from the same `SpanLink` struct that v0.4 uses, so flags may already carry the sentinel bit. Before this fix, the v0.5 serializer wrote that raw value straight into the JSON, so a kept link produced `"flags": 2147483649` instead of `1` — a value no v0.5 producer would ever emit, and one that downstream consumers checking against small integers would not recognize as "kept".

This PR masks bit 31 off the value the v0.5 serializer writes into flags, while it still decides whether to emit the flags key at all from the raw, unmasked value. Deciding presence from the masked value would break the case of a link the tracer explicitly dropped: raw `0x8000_0000` masks to `0`, indistinguishable from "flags never set" if presence were decided post-mask. With this fix, an explicitly dropped link still emits `"flags": 0`; a link that never set flags emits no flags key at all.

# Additional Notes

# How to test the change?

- Added `span_link_flags_sentinel_bit_masked_test`: covers the unset, kept, and explicitly-dropped states of the sentinel bit.
- `cargo test -p libdd-trace-utils`, `cargo fmt --check`, and `cargo clippy -D warnings` pass.


libdd-trace-utils/src/span/mod.rs
```diff
@@ -19,6 +19,18 @@ use std::marker::PhantomData;
...
+/// A `SpanLink`'s `flags` field reserves bit 31 to mean "a value was explicitly set", separate
+/// from the sampling decision carried in the low bits. The sentinel bit distinguishes
+/// `flags == 0` (never set) from an explicit decision of `0`, for example a dropped context,
+/// which would otherwise look identical on the wire.
+///
+/// Only the native v0.4 msgpack wire format (`msgpack_encoder::v04::span_v04`) carries this bit
+/// on the wire, matching the encoding datadog-agent and other tracers already expect. Every
+/// other encoder must mask off this bit before it emits `flags`. These encoders include the
+/// v0.5 JSON dictionary, OTLP protobuf, agentless JSON, and structured JSON logging. Each of
+/// them treats `flags` as the real W3C trace-flags value.
```
bwoebi: Nitpick: _All_ non-JSON formats except otlp expect it. Protobuf, v04 and v1 all expect the bit to be set.

bwoebi (approved):
Looks correct to me, thanks for the improvement.

libdd-trace-utils/src/span/mod.rs
brettlangdon: good callout on phrasing, I started with doing only v0.4 and v0.5

PR #2315, opened 2026-08-04
fix(wasm)!: finish the WASM port of remote-config

bwoebi:
Fix tokio in Cargo.toml and adds missing SleepCapability.


morrisonlevi (approved):
It's a lot of plumbing but basically it moves `tokio::time::timeout` to a `select!` which uses a `SleepCapability::new()` which in this case is coming from setTimeout from a JsFuture, I think? (The other part of the select is the same workload that was under the timeout)

mellon85 (approved):

PR #2316, opened 2026-08-04
feat(data-pipeline): adds new attributes to OTLP trace metrics export

mabdinur:
# What does this PR do?

Adds the libdatadog OTLP trace-metrics support used by ddtrace-py.

- Exports canonical SMC dimensions plus available HTTP, RPC, and Datadog data-point attributes.
- Emits trace-root and top-level values as booleans and peer tags as an array.
- Adds runtime ID, process tags, and tracer tags to resource attributes.
- Adds builder configuration for tracer tags and additional metric tag keys.
- Omits optional method, route, span type, and service-source attributes when unset.
- Keeps trace-metrics output independent of OTel-semantics mode without changing configuration parsing.

# Motivation

ddtrace-py needs the shared exporter to implement the contract defined in [system-tests#7363](https://github.com/DataDog/system-tests/pull/7363) and refined in [system-tests#7466](https://github.com/DataDog/system-tests/pull/7466).

# Additional Notes

Includes a release note announcing ddtrace-py support.

# How to test the change?

Rust formatting, `cargo check`, Clippy, data-pipeline tests, trace-utils tests, and ddtrace-py OTLP system tests passed.

### DIFF
--- a/libdd-data-pipeline/src/otlp/metrics.rs
+++ b/libdd-data-pipeline/src/otlp/metrics.rs
@@ -13,7 +13,7 @@ use libdd_ddsketch::DDSketch;
 use libdd_shared_runtime::Worker;
 use libdd_trace_protobuf::pb;
 use libdd_trace_stats::span_concentrator::{OtlpStatsBucket, SpanConcentrator};
-use libdd_trace_utils::otlp_encoder::mapper::status_code;
+use libdd_trace_utils::otlp_encoder::mapper::tag_to_otlp_kind_str_name;
 use libdd_trace_utils::otlp_encoder::OtlpResourceInfo;
 use serde_json::{json, Value};
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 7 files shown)


libdd-data-pipeline/src/otlp/metrics.rs
```diff
@@ -150,77 +172,84 @@ fn build_attributes(
...
-    otel_trace_semantics_enabled: bool,
 ) -> Vec<Value> {
-    let mut attrs = Vec::new();
-    let mut push = |k: &str, v: &str| {
+    fn push(attrs: &mut Vec<Value>, k: &str, v: &str) {
         if !v.is_empty() {
             attrs.push(kv_str(k, v));
         }
-    };
+    }
```
vpellan: From my understanding, this does not add overhead and only done for scoping purpose. But would it make sense to still extract it as its own function so that it could be used in other functions too in the future?

vpellan: Also, maybe the naming is slightly confusing as I'd expect a `push` function to always push if it can (or throw an error)? Lower, line 189, we have `attrs.push` which will push even if the value is empty

vpellan (approved):
LGTM with minor comments

libdd-data-pipeline/src/otlp/metrics.rs
mabdinur: Moved the helper to module scope and renamed it push_if_non_empty in refactor(data-pipeline): clarify optional OTLP attributes (https://github.com/DataDog/libdatadog/commit/9890ff2c5)

PR #2322, opened 2026-08-05
fix(crashtracking): check fields and exclude uuid for `has_data`

gyuheon0h:
# What does this PR do?

Excludes `uuid` field from `CrashInfoBuilder` when checking for `has_data`

# Motivation

Previously, the `has_data` check would compare the whole struct against `Self::default()`. However, uuid is generated when the builder is created, so this would always return true

# Additional Notes

# How to test the change?

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
 channel = "1.87.0"
-components = ["rustfmt", "clippy", "rust-analyzer"]
+components = ["rustfmt", "clippy"]
 profile = "minimal"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
 channel = "1.87.0"
-components = ["rustfmt", "clippy"]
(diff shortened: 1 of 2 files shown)


gyuheon0h:
* **#2322** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2322?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2322?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

danielsn (approved):
But please set a description and motivation for this PR

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
 channel = "1.87.0"
-components = ["rustfmt", "clippy"]
+components = ["rustfmt", "clippy", "rust-analyzer"]
```
danielsn: why

gyuheon0h: Local linter? Something added this. Reverted

PR #2323, opened 2026-08-05
fix(trace-stats): read OTel HTTP names for the status and method dimensions

link04:
### What does this PR do?

Preserves OpenTelemetry HTTP semantics in libdatadog's in-process consumers and OTLP output.

#### Trace stats and OTLP trace metrics

- reads canonical Datadog HTTP method/status keys first, then falls back to `http.request.method` and `http.response.status_code`
- accepts valid numeric or string status values without allowing empty, malformed, non-finite, or non-integral values to shadow a valid fallback
- emits OTel trace-metric dimensions for method, status, route, span name, and kind
- leaves successful `status.code` unset and prevents additional tags from overriding the computed status

### HTTP tag handling

libdatadog does not rename tracer span tags in this PR. It accepts the Datadog and OpenTelemetry names below as equivalent inputs to trace-stat aggregation and emits the canonical OpenTelemetry name in OTLP trace metrics.

| Concept | Datadog span input | OpenTelemetry span input | OTLP trace-metric output |
| --- | --- | --- | --- |
| HTTP method | `http.method` | `http.request.method` | `http.request.method` |
| HTTP response status | `http.status_code` | `http.response.status_code` | `http.response.status_code` as an integer |
| Span error status | Span error flag | Span error flag | `status.code=STATUS_CODE_ERROR` for errors; omitted for successful spans |

When both naming variants are present, the Datadog name wins. Empty or malformed values fall through to the alternate name instead of hiding a valid value.

#### OTLP sampled flags

Sampling priority is a trace-level decision, but v0.4/v0.5 inputs may attach `_sampling_priority_v1` to only one span. The mapper now uses the same first-priority policy as `drop_chunks`, treats a missing priority as sampled, and applies one decision to every span in the chunk.

This prevents sampled traces from exporting child spans with `flags=0` and keeps OTLP flags consistent with the existing chunk-retention policy.

No public signature, wire-format, setting, or environment-variable change is introduced.

### Motivation

Under `DD_TRACE_OTEL_SEMANTICS_ENABLED`, spans use OTel HTTP names before client-side stats are computed. Stats and OTLP metrics must preserve those dimensions, and all spans in an exported trace must carry the same sampled decision.

Companion changes:

- Historical full Python PR: [DataDog/dd-trace-py#19501](https://github.com/DataDog/dd-trace-py/pull/19501)
- Native Python stack: [foundation #20008](https://github.com/DataDog/dd-trace-py/pull/20008), [client #20009](https://github.com/DataDog/dd-trace-py/pull/20009), [server #20010](https://github.com/DataDog/dd-trace-py/pull/20010)
- [DataDog/dd-trace-js#9688](https://github.com/DataDog/dd-trace-js/pull/9688)
- [DataDog/system-tests#7455](https://github.com/DataDog/system-tests/pull/7455)

### Testing

- `cargo fmt --check -p libdd-trace-utils`
- mapper unit tests: **31 passed**
- OTLP trace-exporter integration tests: **3 passed**
- trace-stats tests: **43 passed**

End-to-end re-run from libdatadog head `0ed7fd9bd` with the final Python server layer and companion system tests:

| Scenario | Result |
| --- | ---: |
| `OTEL_SEMANTICS_OTLP_TRACE_METRICS` | **4 passed** |
| `OTEL_SEMANTICS_OTLP_SAMPLING_RULES` | **2 passed** |


libdd-data-pipeline/src/otlp/metrics.rs
```diff
@@ -242,7 +242,7 @@ fn build_attributes(
     }
     for tag in &group.additional_metric_tags {
         if let Some((k, v)) = tag.split_once(':') {
-            if attrs.iter().all(|attr| attr["key"].as_str() != Some(k)) {
+            if k != "status.code" && attrs.iter().all(|attr| attr["key"].as_str() != Some(k)) {
```
ichinaski: On line 192 we have
```rs
attrs.push(kv_str(
        "status.code",
        if is_error {
            STATUS_CODE_ERROR
        } else {
            STATUS_CODE_OK
        },
    ));
```

isn't this change dead code?

ichinaski (commented):
Changes regarding the OTel tags for method and status_code extraction look good.

libdd-data-pipeline/src/otlp/metrics.rs
link04: Yes seems I missed that when trying to fix setting the wrong code, I've removed it, thank you!

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -158,6 +167,49 @@ fn float_to_int(f: f64) -> Option<u8> {
     Some(f as u8)
 }
 
+fn float_to_u32(f: f64) -> Option<u32> {
+    if !f.is_finite() || f.floor() != f || f < 0.0 || (u32::MAX as f64) < f {
```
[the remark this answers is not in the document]
link04: Addressed in [1a7d38e](https://github.com/DataDog/libdatadog/pull/2323/commits/1a7d38ec190db88f1c78642fe3191a0d399ac5c6).

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -158,6 +167,49 @@ fn float_to_int(f: f64) -> Option<u8> {
...
+        return None;
+    }
+    Some(f as u32)
```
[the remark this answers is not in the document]
link04: Addressed in [1a7d38e](https://github.com/DataDog/libdatadog/pull/2323/commits/1a7d38ec190db88f1c78642fe3191a0d399ac5c6).

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -90,6 +90,15 @@ fn chunk_trace_id_high<T: TraceData>(chunk: &[Span<T>]) -> u64 {
         .unwrap_or(0)
 }
 
+/// Resolve the chunk's trace-level sampling decision using the same first-priority policy as
+/// v04 `drop_chunks`.
+fn chunk_sampling_flags<T: TraceData>(chunk: &[Span<T>]) -> u32 {
+    let priority = chunk
+        .iter()
+        .find_map(|span| span.metrics.get("_sampling_priority_v1"));
+    priority.is_none_or(|priority| *priority > 0.0) as u32
```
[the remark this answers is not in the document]
link04: Addressed in [1a7d38e](https://github.com/DataDog/libdatadog/pull/2323/commits/1a7d38ec190db88f1c78642fe3191a0d399ac5c6).

ichinaski (approved):

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -69,6 +73,39 @@ const V04_TRACES_ENDPOINT: &str = "/v0.4/traces";
...
+    for chunk in traces {
+        if !libdd_trace_utils::span::trace_utils::chunk_sampling_decision(chunk) {
+            for span in chunk.iter().filter(|span| {
+                libdd_trace_utils::span::trace_utils::is_span_sampled_independently(span)
+            }) {
+                rejected_sampled_spans.insert((span.trace_id, span.span_id));
+            }
+        }
+    }
+    rejected_sampled_spans
```
yannham: Nit: this can be written slightly more idiomatically using iterator/collect()

```suggestion
    traces
        .iter()
        .filter(|chunk| {
            !libdd_trace_utils::span::trace_utils::chunk_sampling_decision(chunk)
        })
        .flatten()
        .filter(libdd_trace_utils::span::trace_utils::is_span_sampled_independently)
        .map(|span| (span.trace_id, span.span_id))
        .collect()
```

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -753,6 +793,14 @@ impl<
             return self.send_agentless_traces_inner(traces, config).await;
         }
 
+        // Filtering can remove the span carrying the trace-level priority while retaining a
+        // single sampled span. Remember those independently sampled spans from rejected chunks so
+        // each retained chunk can still emit the correct OTLP sampled flag.
+        let rejected_sampled_spans = self
+            .otlp_config
+            .as_ref()
+            .map(|_| capture_rejected_sampled_spans(&traces));
```
yannham:
```suggestion
        let rejected_sampled_spans = self
            .otlp_config
            .is_some()
            .then(|| capture_rejected_sampled_spans(&traces));
```

yannham (approved):

bantonsson (approved):

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -157,6 +166,62 @@ fn float_to_int(f: f64) -> Option<u8> {
...
+    if !f.is_finite() || f.floor() != f || f < 0.0 || (u32::MAX as f64) < f {
+        return None;
+    }
+    // The checks above guarantee the finite integral value is representable without loss.
+    Some(f as u32)
+}
+
+fn is_valid_http_status_code(code: &u32) -> bool {
+    (100..600).contains(code)
+}
```
bm1549, line 179: Should we preserve the existing DD behavior here? `http.status_code` currently accepts any non-negative `u32`, which matches the Agent. Restricting it to `100..600` means a legacy value like `600` now becomes `0` or falls through to the OTel alias. That seems separate from adding alias support.

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -271,20 +336,17 @@ impl<'a> BorrowedAggregationKey<'a> {
             vec![]
         };
 
-        let http_method = span.get_meta("http.method").unwrap_or_default();
+        // The Datadog name wins when both are present. An empty value counts as absent, so a
+        // tracer that writes the key unconditionally cannot shadow a real value under the other
+        // name, which is the mistake get_grpc_status_code above already avoids.
+        let http_method = get_http_method(span);
 
         let http_endpoint = span
```
bm1549, line 347: Existing code, but should `http.route` win when it is non-empty? We later emit this field as `http.route`, so preferring `http.endpoint` can relabel the DD-only value as the canonical route. I would use non-empty `http.route`, then `http.endpoint` as the fallback.

libdd-trace-utils/src/span/trace_utils.rs
```diff
@@ -135,6 +135,30 @@ const SAMPLING_PRIORITY_KEY: &str = "_sampling_priority_v1";
 const SAMPLING_SINGLE_SPAN_MECHANISM: &str = "_dd.span_sampling.mechanism";
 const SAMPLING_ANALYTICS_RATE_KEY: &str = "_dd1.sr.eausr";
 
+/// Returns the trace-level sampling decision stored on the first span carrying a priority.
+///
+/// A missing priority is treated as sampled, matching the agent's no-priority policy.
+pub fn chunk_sampling_decision<T>(chunk: &[Span<T>]) -> bool
```
bm1549: This is ahead of the current RFC, which still says single-span sampling support is deferred. It also makes missing priority `flags=1`, while Node and Go use `0` without an explicit keep decision. The behavior makes sense with `drop_chunks`, but can we make this contract explicit in the RFC and cross-tracer tests?

bm1549 (approved):
Left a few small comments about cross-repo consistency that I think should be considered. Otherwise LGTM

libdd-trace-stats/src/span_concentrator/aggregation.rs
link04: Good catch. I preserved the existing Datadog-key conversion and precedence behavior for `http.status_code`, including its numeric cast and metadata parse-or-zero behavior. The 100.. 600 validation now applies only to the OTel `http.response.status_code` key. I also added regression coverage for Datadog values outside the HTTP range and malformed values.

libdd-trace-stats/src/span_concentrator/aggregation.rs
link04: Agreed. The aggregation now prefers a non-empty `http.route` and falls back to `http.endpoint`. Tests cover both the two-key case and an empty route falling back to the retained endpoint.

libdd-trace-utils/src/span/trace_utils.rs
link04: Agreed. I removed the new single-span-specific pre-filter bookkeeping. In OTel semantics mode, the first explicit trace priority is applied consistently to every span; without an explicit priority, flags remains 0, matching Node and Go. Outside OTel semantics mode, the prior per-span >= 1 behavior is preserved.

link04: I've updated the section where we speak about this in the RFC.

PR #2330, opened 2026-08-07
refactor(ci): release proposal dispatch wf

iunanua:
# What does this PR do?

Move inline bash script in the workflow into separated scripts:
- ~~`semver-level.sh`: now handles correctly failures~~ fixed in another PR
- `commits-since-release.sh`: now includes 'tag_in_local_branch', 'latest_tag' and 'range' in the output json needed by release-version-bumps.sh
- `major-bumps-level.sh`: removed not used code
- `check_cargo_metadata.sh`: clean up
- `release-version-bumps.sh`: Replaces 'Release version bumps' step
- `release-version-major-bumps.sh`: Replaces 'Update version for crates with libdd-* …' step
- `release-generate-changelogs.sh`: Replaces 'Generate CHANGELOGS' step

## Changes
- `set -euo pipefail` in the extracted scripts. The steps ran under GitHub's default bash -e {0}: errexit on, but pipefail and nounset off, so a pipeline whose first stage failed (`git tag -l … | head -1`) reported success and a typo'd variable expanded to empty instead of aborting. It does not cover the loops' row read though: `done < <(jq -c '.[]' …)` runs jq in a process substitution, whose exit status neither errexit nor pipefail sees — a failing jq left the loop with no input and the script exited 0 having done nothing. That needed its own fix, materializing the rows into a variable first, since a command substitution's status *is* checked.
- A `semver-level.sh` failure now prints why. Its output was captured with 2>&1 into a variable that errexit then discarded.
- `api-changes.json` now includes the deferred crates. It is in the release-dispatch-data artifacts.

## Notes
You can compare the result of the refactored wf with the previous one:
- refactor: https://github.com/DataDog/libdatadog/actions/runs/31375384070
- previous: https://github.com/DataDog/libdatadog/pull/2332

crates: `libdd-ffe,libdd-http-client,libdd-tracer-flare`
commit: `ea75b04c3547037937730a14cf8a72a5ebf702d7`


hoolioh (approved):
LGTM, while reviewing the code I've detected a possible bug but I will open a subsequent PR based on this work.

PR #2333, opened 2026-08-07
feat(shared-runtime): add block_on_with_timeout to BlockingRuntime

brettlangdon:
# What does this PR do?

Lets consumers bound a `block_on` wait with a deadline without adding tokio as a direct dependency — the executor is whatever the runtime already runs on.

Returns `BlockOnTimeoutError::TimedOut`/`Io` to distinguish a timeout from an executor failure.

# Motivation

# Additional Notes

# How to test the change?


libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -75,13 +75,99 @@ pub trait SharedRuntime {
...
+        f: F,
+        timeout: std::time::Duration,
+    ) -> Result<F::Output, BlockOnTimeoutError> {
+        // When the executor has no timer driver (e.g. a caller-supplied runtime built without
+        // `enable_time`), tokio::time::timeout panics instead of returning an error. The
+        // default block_on implementations are tokio-backed, so the panic reaches this
+        // catch_unwind. Catching the panic here converts it to an error, so a misconfigured
+        // executor cannot abort a process that calls block_on_with_timeout across the FFI
+        // boundary.
+        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
yannham, line 157: I'm wondering if we need to go through this machinery. While it's true that it could panic in theory, there are a dozen usages of `timeout` in various part of libdatadog that just use `timeout` and expect it to be there and to not panic, so I think it's reasonable to assume that there's always a timer driver. Unless you do not have one in your setup?. Even if we were to protect against that, I suppose the panic catching should rather be handled at the FFI boundary. Any thought @VianneyRuhlmann ?

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -75,13 +75,99 @@ pub trait SharedRuntime {
...
+            self.block_on(async move { tokio::time::timeout(timeout, f).await })
```
yannham, line 158:
```suggestion
            self.block_on(tokio::time::timeout(timeout, f))
```

libdd-shared-runtime/src/shared_runtime/mod.rs
brettlangdon: My main push back on this is that the shared runtime should be abstracting away the Tokio runtime wherever possible, if we allow a tokio specific panic to be raised up to called (Rust or C FFI or etc) then we are exposing the tokio specifics to the caller, and if we ever swap out the runtime then we'll break their call sites.

yannham: Hmm, it's a good point. In general I'm a bit wary of using `catch_unwind` as an exception catching mechanism; it was not really designed for that (or at least it did not became that), it is moot if `panic=abort` is set somewhere in the final build, and the string message matching is a bit fragile.

That being said, the perf impact is minimal in the happy path and unfortunately tokio doesn't expose an easy way to know if it was built with the timer capability, which is a bummer. So maybe it's the safest way to go for now yeah...

yannham: Well, at least I'm trying: https://github.com/tokio-rs/tokio/issues/8361 :man_shrugging: in the meantime let's not block this PR,

yannham (approved):

libdd-shared-runtime/src/shared_runtime/mod.rs
brettlangdon: oh, nice!

PR #2337, opened 2026-08-10
fix(library-config): omit absent process context attributes

morrisonlevi:
# What does this PR do?

If these items are not set, they should be missing, rather than sent with no value.

# Motivation

In PHP, these will not be set at all because they live in the OTel thread context instead of the process context.

# Additional Notes

Needed for DataDog/dd-trace-php#4077.

# How to test the change?

Tests the same, there just won't be "empty" items and they'll be missing instead.

### DIFF
--- a/libdd-library-config/src/tracer_metadata.rs
+++ b/libdd-library-config/src/tracer_metadata.rs
@@ -109,10 +109,8 @@ impl TracerMetadata {
             }
         }
 
-        // Even if there's no value, we still set the key to let the reader know that we do support
-        // and emit this specific attribute, which happens to be empty in this case.
-        fn key_value_opt(key: &'static str, val: &Option<String>) -> KeyValue {
-            key_value(key, val.as_ref().cloned().unwrap_or_default())
+        fn key_value_if_some(key: &'static str, val: &Option<String>) -> Option<KeyValue> {
+            val.as_ref().map(|val| key_value(key, val.clone()))
         }
 
         // Every field of `self` should get propagated to the otel context.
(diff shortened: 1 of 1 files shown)


bwoebi (approved):

yannham:
The original implementation did drop undefined attributes, but @ivoanjo reverted this decision explicitly, as per this review comment: https://github.com/DataDog/libdatadog/pull/1658#discussion_r2896429123, and as hinted at by the presence of the comment. It could be good to double check with him that this is ok.

cataphract (approved):

libdd-library-config/src/tracer_metadata.rs
```diff
@@ -133,21 +131,29 @@ impl TracerMetadata {
...
+        .into_iter()
+        .flatten()
+        .collect();
 
         // `mut` is only needed when `otel-thread-ctx` is enabled.
         #[allow(unused_mut)]
-        let mut extra_attributes = vec![key_value_opt("datadog.process_tags", process_tags)];
+        let mut extra_attributes: Vec<_> = key_value_if_some("datadog.process_tags", process_tags)
+            .into_iter()
+            .collect();
```
ivoanjo, line 156: I agree that for a multi-tenant case (like we're doing for PHP), it's kinda weird to have empty `service.name`/`service.instance.id`/`service.version`/`deployment.environment.name` since those are properties of a given tenant, not for the whole process.

Yet, `container.id` and `datadog.process_tags` I think still make sense as being process-wide.

Do you mind if we keep the old `key_value_opt` for those two, and use the new `key_value_if_some` only for those that are truly per-tenant?

ivoanjo (commented):
Thanks for the ping @morrisonlevi and @yannham . I think for the multi-tenant use-case this change is fine, and for all others, our libraries usually have defaults for most of them so in practice I don't think this empty vs omitted will show up very often there.

libdd-library-config/src/tracer_metadata.rs
bwoebi: For container.id, I agree. Sadly process tags are service specific with the `svc` tags...

ivoanjo: :+1: I wasn't aware of that, thanks for catching it!

bwoebi: Given that I don't think it makes much sense to have particular behaviour for just this single container.id, I'm now just merging this as is.

PR #2339, opened 2026-08-11
fix(ffe): report rejected flags as parse errors

leoromanovsky:
## Motivation

Malformed per-flag UFC configuration is already isolated during ingestion: libdatadog stores a rejection error for that key while compiling valid neighboring flags normally. The FFI translation incorrectly exposed that rejection as a normal DEFAULT result with no error code, hiding a configuration bug from callers.

This implements the contract merged in [DataDog/ffe-system-test-data#26](https://github.com/DataDog/ffe-system-test-data/pull/26): evaluating a rejected key returns the caller default with ERROR / PARSE_ERROR, while valid neighboring flags remain usable.

## Changes

- Map FlagConfigurationInvalid to the FFI ERROR reason and PARSE_ERROR error code.
- Add regression coverage proving an invalid regex flag does not poison a valid neighboring flag.
- Bump the canonical fixture submodule to the revision merged in [DataDog/ffe-system-test-data#26](https://github.com/DataDog/ffe-system-test-data/pull/26).
- Update the canonical adapter to expect PARSE_ERROR for rejected keys.
- Compare semantic versions by precedence so build metadata does not change equality or ordering.
- Correct the regex-ingestion documentation to match the parser behavior.

## Decisions

- Keep validation in libdatadog so the same Rust regex and UFC parsers used for evaluation own rejection decisions.
- Retain only the per-key rejection error in the compiled configuration, not the malformed flag object.
- Preserve FLAG_NOT_FOUND for keys absent from the configuration.
- Use SemVer precedence comparisons, which ignore build metadata as required by the specification.

## Validation

- [x] cargo test -p libdd-ffe
- [x] cargo test -p libdd-ffe-ffi rejected_flag_is_exposed_as_parse_error
- [x] cargo test -p libdd-ffe-test-suite --test canonical_fixtures
- [x] git diff --check

### DIFF
--- a/libdd-ffe/src/rules_based/error.rs
+++ b/libdd-ffe/src/rules_based/error.rs
@@ -33,8 +33,10 @@ pub enum EvaluationError {
     ConfigurationParseError,
 
     /// A requested flag exists in the configuration payload, but its per-flag configuration is
-    /// invalid or unsupported by this SDK. The SDK should return the caller default and expose
-    /// this as a parse error for that flag without invalidating the rest of the configuration.
+    /// invalid or unsupported by this SDK. Failures encountered while compiling an individual
+    /// flag are converted to `FlagConfigurationInvalid` at the per-flag ingestion boundary. The
+    /// SDK should return the caller default and expose this as a parse error for that flag without
+    /// invalidating the rest of the configuration.
     #[error("flag configuration is invalid or unsupported")]
     FlagConfigurationInvalid,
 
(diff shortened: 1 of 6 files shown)


libdd-ffe/src/rules_based/error.rs
sameerank, line 31: I feel like this comment could also be updated with something like "Failures encountered while compiling an individual flag are converted to `FlagConfigurationInvalid` at the per-flag ingestion boundary."

sameerank (approved):

libdd-ffe/src/rules_based/error.rs
leoromanovsky: Yea good call

PR #2342, opened 2026-08-11
feat(gotter)!: unpatch symbol functionality

gyuheon0h:
# What does this PR do?

This pr adds `unhook_symbol` API to `libdd-gotter` that reverses a previous `hook_symbol` call by restoring GOT entries that currently point at the hook function back to the original address. This only restores entries whose current value matches `hook_fn`, making it safe to call even if libraries were loaded/unloaded after the original hook or entries were independently modified.

# Motivation

Embedders that load libdatadog as a shared library and later `dlclose` it would leave other modules' GOT slots pointing at `hook_assert_fail` inside the now-unmapped image. A subsequent C `assert()` in any patched module would jump to stale code.

# Additional Notes

# How to test the change?

Unit tests

### DIFF
--- a/libdd-gotter/src/elf.rs
+++ b/libdd-gotter/src/elf.rs
@@ -1185,6 +1185,138 @@ unsafe fn patch_got_entries(
     }
 }
 
+/// Restore GOT entries in one library for the target symbol, undoing a
+/// previous [`patch_got_entries`] call.
+///
+/// Only restores entries whose current value matches `hook_fn`, so this
+/// is safe to call even if some entries were never patched or were
+/// independently overwritten.
+///
+/// # Safety
+/// Same as [`patch_got_entries`].
+unsafe fn restore_got_entries(
+    dyn_info: &DynamicInfo,
+    symbol_name: &[u8],
+    hook_fn: usize,
+    orig_addr: usize,
+    guard: &mut PageProtGuard,
+    restored: &mut usize,
+    failed: &mut usize,
+) {
+    let mut try_restore = |r_info: u64, r_offset: u64| {
+        if !is_got_pointer_reloc(elf64_r_type(r_info)) {
+            return;
+        }
+        let sym_idx = elf64_r_sym(r_info);
+        if let Some(cstr) = dyn_info.sym_name(sym_idx) {
+            if cstr.to_bytes() == symbol_name {
+                let addr = u64_to_usize(r_offset) + dyn_info.base_address();
+                // Only restore entries that currently point at our hook.
+                // SAFETY: `addr` is a GOT slot address from a valid
+                // relocation entry in a currently-loaded library.
+                let current = unsafe { core::ptr::read_unaligned(addr as *const usize) };
+                if current == hook_fn {
+                    if guard.override_entry(addr, orig_addr) {
+                        *restored += 1;
+                    } else {
+                        *failed += 1;
+                    }
+                }
+            }
+        }
+    };
+
+    for reloc in dyn_info.rels() {
+        try_restore(reloc.r_info, reloc.r_offset);
+    }
+    for relocs in [dyn_info.relas(), dyn_info.jmprels()] {
+        for reloc in relocs {
+            try_restore(reloc.r_info, reloc.r_offset);
+        }
+    }
+}
+
+/// Reverse a previous [`hook_symbol`] call by restoring all GOT entries
+/// for `symbol_name` that currently point at `hook_fn` back to
+/// `orig_addr`.
+///
+/// Only entries whose current value equals `hook_fn` are touched, so
+/// this is safe even if some libraries were loaded after the original
+/// hook (their entries were never patched and won't be modified).
+///
+/// Returns the number of entries restored and failed.
+///
+/// # Safety
+///
+/// * `orig_addr` must be the original function address returned by the corresponding `hook_symbol`
+///   call (`HookResult::orig_addr`).
+/// * The caller must ensure no other thread is concurrently calling through a GOT entry being
+///   restored. In practice this means the hooked symbol must not be called concurrently with
+///   unhooking.
+pub unsafe fn unhook_symbol(
+    symbol_name: &CStr,
+    hook_fn: usize,
+    orig_addr: usize,
+) -> Result<HookResult, HookError> {
+    let symbol_name_bytes = symbol_name.to_bytes();
+    // Validate the symbol name but we don't need to resolve it. The
+    // caller provides orig_addr directly.
+    let _ = symbol_name
+        .to_str()
+        .map_err(|_| HookError::InvalidSymbolName)?;
+
+    let mut entries_restored: usize = 0;
+    let mut entries_failed: usize = 0;
+    let mut guard = PageProtGuard::new();
+
+    let guard_ptr = &mut guard as *mut PageProtGuard;
+    let restored_ptr = &mut entries_restored as *mut usize;
+    let failed_ptr = &mut entries_failed as *mut usize;
+
+    iterate_libraries(|info, _is_exe| {
+        let lib_name = if info.dlpi_name.is_null() {
+            ""
+        } else {
+            // SAFETY: dl_iterate_phdr guarantees dlpi_name is a valid
+            // NUL-terminated C string for the callback's duration.
+            unsafe { CStr::from_ptr(info.dlpi_name) }
+                .to_str()
+                .unwrap_or("")
+        };
+        if lib_name.contains("linux-vdso") || lib_name.contains("/ld-linux") {
+            return false;
+        }
+        // SAFETY: `info` points to a valid `dl_phdr_info` provided by
+        // `dl_iterate_phdr`; the library is mapped for the callback's
+        // duration.
+        let Some(dyn_info) = (unsafe { DynamicInfo::from_phdr(info) }) else {
+            return false;
+        };
+        // SAFETY: dyn_info was just produced from a currently-loaded
+        // library. guard_ptr/restored_ptr/failed_ptr are valid for the
+        // duration of iterate_libraries (they point to locals in the
+        // enclosing fn).
+        unsafe {
+            restore_got_entries(
+                &dyn_info,
+                symbol_name_bytes,
+                hook_fn,
+                orig_addr,
+                &mut *guard_ptr,
+                &mut *restored_ptr,
+                &mut *failed_ptr,
+            );
+        }
+        false
+    });
+
+    Ok(HookResult {
+        orig_addr,
+        entries_patched: entries_restored,
+        entries_failed,
+    })
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -1522,6 +1654,78 @@ mod tests {
             "excluding self should skip at least one library \
              (total={total_libs}, excluding_self={libs_excluding_self})"
         );
+    }
+
+    /// Hook `getpid` and then unhook it, verifying that:
+    /// 1. hook_symbol patches at least one GOT entry
+    /// 2. unhook_symbol restores the same number of entries
+    /// 3. a second unhook_symbol is a no-op (entries already restored)
+    ///
+    /// We use `getpid` because it is a known symbol that we are okay with patching for the
+    /// duration of the test.
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn test_unhook_symbol_restores_got_entries() {
+        unsafe extern "C" fn dummy_getpid() -> libc::pid_t {
+            -1
+        }
+
+        let hook_fn = dummy_getpid as *const () as usize;
+
+        // SAFETY: dummy_getpid matches getpid's signature.
+        let hook_result = unsafe { hook_symbol(c"getpid", hook_fn) };
+
+        let Ok(hook_result) = hook_result else {
+            eprintln!("note: getpid not found, skipping unhook test");
+            return;
+        };
+
+        assert!(
+            hook_result.entries_patched > 0,
+            "getpid should have at least one GOT entry"
+        );
+
+        let orig_addr = hook_result.orig_addr;
+        let patched_count = hook_result.entries_patched;
+
+        // Verify the hook took effect.
+        assert_eq!(
+            unsafe { libc::getpid() },
+            -1,
+            "hooked getpid should return -1"
+        );
+
+        // SAFETY: orig_addr is the real getpid address returned by
+        // hook_symbol, and hook_fn is the dummy we installed.
+        let unhook_result = unsafe { unhook_symbol(c"getpid", hook_fn, orig_addr) }.unwrap();
+
+        assert_eq!(
+            unhook_result.entries_patched, patched_count,
+            "unhook should restore exactly the entries that were patched"
+        );
+
+        // Verify the original behavior is restored.
+        let real_pid = unsafe { libc::getpid() };
+        assert_ne!(real_pid, -1, "unhooked getpid should return the real PID");
+        assert!(real_pid > 0, "real PID should be positive");
+
+        // A second unhook should be a no-op. Entries already point at
+        // orig_addr, not hook_fn.
+        let second = unsafe { unhook_symbol(c"getpid", hook_fn, orig_addr) }.unwrap();
+        assert_eq!(
+            second.entries_patched, 0,
+            "second unhook should find nothing to restore"
+        );
+    }
+
+    /// Verify that unhook_symbol returns InvalidSymbolName for invalid
+    /// UTF-8 symbol names.
+    #[test]
+    #[allow(clippy::manual_c_str_literals)] // intentionally invalid UTF-8
+    fn test_unhook_symbol_invalid_name() {
+        let invalid = CStr::from_bytes_with_nul(b"\xff\0").unwrap();
+        let result = unsafe { unhook_symbol(invalid, 0, 0) };
+        assert_eq!(result.unwrap_err(), HookError::InvalidSymbolName);
     }
 
     /// Sanity check against real loaded libraries: the filter should
--- a/libdd-gotter/src/elf.rs
+++ b/libdd-gotter/src/elf.rs
@@ -986,7 +986,7 @@ pub struct LookupResult {
 }
 
 /// Result of a successful [`hook_symbol`] call.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
 pub struct HookResult {
     /// Resolved address of the original symbol. Store this so the hook
     /// function can forward calls to the real implementation.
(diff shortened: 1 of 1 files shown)


gyuheon0h:
* **#2492** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2492?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2496** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2496?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2342** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2342?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2342?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

libdd-gotter/src/elf.rs
```diff
@@ -1185,6 +1259,50 @@ unsafe fn patch_got_entries(
...
+/// are left untouched.
+///
+///
+/// # Safety
+///
+/// * Every address in `hook_result.slots` must still point into a currently-mapped region. A
+///   library whose GOT was patched must not be `dlclose`d before this function returns.
+/// * The caller must ensure no other thread is concurrently calling through one of the GOT entries
+///   being restored.
+pub unsafe fn unhook_symbol(hook_result: &HookResult, hook_fn: usize) -> HookResult {
```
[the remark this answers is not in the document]
gyuheon0h: Use PR is stacked on top

libdd-gotter/src/elf.rs
```diff
@@ -1185,6 +1264,40 @@ unsafe fn patch_got_entries(
...
+/// Designed to be called **once**, in a controlled context (e.g. a library
+/// unload handler), when no other thread is concurrently patching the same
+/// GOT entries. Threads concurrently *calling through* a hooked slot are
+/// safe: the write is pointer-sized and naturally aligned, so the CPU sees
+/// either the old or new pointer atomically.
+///
+/// # Safety
+/// Every address in `hook_result.slots` must still point into a
+/// currently-mapped region. No library whose GOT was patched may be
+/// `dlclose`d before this function returns.
```
gyuheon0h, line 1310: this is a hard to satisfy requirement. Need to protect against when this happens in the API explicitly

libdd-gotter/src/elf.rs
```diff
@@ -1185,6 +1264,40 @@ unsafe fn patch_got_entries(
...
+pub unsafe fn unhook_symbol(hook_result: &HookResult) {
+    let mut guard = PageProtGuard::new();
+
+    for &(addr, original_value) in &hook_result.slots {
+        // SAFETY: caller guarantees `addr` is in a currently-mapped region.
+        if unsafe { guard.make_writable(addr) } {
```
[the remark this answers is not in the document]
gyuheon0h: Good point

libdd-gotter/src/elf.rs
```diff
@@ -1185,6 +1264,40 @@ unsafe fn patch_got_entries(
...
+            // Use write_unaligned to match patch_got_entries. GOT slots for
+            // JUMP_SLOT/GLOB_DAT are always pointer-aligned, but
+            // R_X86_64_64/R_AARCH64_ABS64 entries in data sections may not be.
+            // SAFETY: `addr` points to a writable GOT slot in mapped memory.
+            unsafe { core::ptr::write_unaligned(addr as *mut usize, original_value) };
```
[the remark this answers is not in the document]
gyuheon0h: Yeah... I guess there isnt really anything we can do here since we can't lock function calls through the GOT... and we physically cannot have atomic guarantees on non-aligned ptrs

libdd-gotter/src/elf.rs
```diff
@@ -817,28 +817,43 @@ impl PageProtGuard {
...
+    /// `addr` must point into a currently-mapped region.
+    pub unsafe fn make_writable(&mut self, addr: usize) -> bool {
+        let aligned = addr & !(self.page_size - 1);
+        if self.touched.contains_key(&aligned) {
+            return true;
+        }
+        // If /proc/self/maps isn't available (or the page isn't in it,
+        // which shouldn't happen for a mapped GOT page) fall back to
+        // PROT_READ, the RELRO'd default. Tighter than leaving pages RW.
+        let orig = self.original_prot(aligned).unwrap_or(PROT_READ);
```
yannham, line 836: This means that if something else (imagine a libdatadog-like library, or the host program itself) is relying on this mapping being writeable, this might create strange memory protection errors if we restore a stronger permission than what was here originally. Now I assume this will be an exceptional situation (writeable required + no `/proc/self/maps`), and I still think restoring `READ` only is the right thing to do security-wise, but I wonder if that will come to bite us some day with a strange bug. Maybe we could document this behavior, e.g. in the README or somewhere else?

libdd-gotter/src/elf.rs
```diff
@@ -1007,6 +1025,31 @@ pub enum HookError {
     SymbolNotFound,
 }
 
+/// Result of an [`unhook_symbol`] call.
+///
+/// If [`slots_failed`] is non-zero, one or more GOT entries still point at
+/// the hook function because `mprotect` was rejected (e.g. a seccomp policy
+/// installed after hooking). Those slots remain live pointers into the
+/// library being unloaded; the caller **must not** unload until all slots
```
yannham, line 1033: It's a bit strange to talk about "the caller" for a struct documentation. Would the actual function be a better place to put this documentation? Should this be a safety condition?

libdd-gotter/src/elf.rs
```diff
@@ -1030,7 +1073,7 @@ pub enum HookError {
 /// # Safety
 ///
 /// `hook_fn` must point to a function with the same calling convention
-/// and signature as the symbol being hooked. The patching is permanent.
+/// and signature as the symbol being hooked.
```
yannham, line 1076: Nit: mention in this function's documentation the unhook function?

libdd-gotter/src/elf.rs
```diff
@@ -1061,6 +1104,9 @@ pub unsafe fn hook_symbol_excluding_self(
 /// contain `addr`. Used by `hook_symbol_excluding_self` to identify the
 /// hook's own library regardless of PIE vs non-PIE (where `dlpi_addr`
 /// may be 0 for the main executable).
+///
+/// We MUST NOT dlopen a library that contains the hook symbol during the duration
```
yannham, line 1108: Nit: should that be part of a proper `# Safety` section? Or is it not really unsafe but just leads to unwanted behavior

libdd-gotter/src/elf.rs
```diff
@@ -1105,14 +1159,19 @@ unsafe fn hook_symbol_impl(
...
-        // `dl_iterate_phdr`
+        // `dl_iterate_phdr`.
         let Some(dyn_info) = (unsafe { DynamicInfo::from_phdr(info) }) else {
             return false;
         };
+
+        // Reserve before patching so push is guaranteed not to allocate
+        // after any GOT entry in this library has been overwritten.
+        let count = count_matching_got_entries(&dyn_info, symbol_name_bytes);
+        unsafe { (*slots_ptr).reserve(count) };
```
yannham, line 1170: Does this have to be unsafe code? Could we share a mutable reference, use `reserve` safely, and only convert it into a pointer just in time when calling `patch_got_entries`? (maybe not, but asking just in case)

libdd-gotter/src/elf.rs
```diff
@@ -1130,15 +1190,48 @@ unsafe fn hook_symbol_impl(
...
+        }
+    };
+    for reloc in dyn_info.rels() {
+        try_count(reloc.r_info);
+    }
+    for relocs in [dyn_info.relas(), dyn_info.jmprels()] {
+        for reloc in relocs {
+            try_count(reloc.r_info);
+        }
+    }
```
yannham, line 1218: Nit (style): instead of having to make a closure here, I think you could just chain iterators instead, something like `for reloc in dyn_info.rels().into_iter().chain(dyn_info.relas()).chain(dyn_info.jmprels()) { ...count... }`.

libdd-gotter/src/elf.rs
yannham: In fact, `patch_got_entries` takes an `&mut`, so I suspect `slots` can stay a safe reference all along?

libdd-gotter/src/elf.rs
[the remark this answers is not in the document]
yannham: Doesn't that contradict the documentation of this function (writes are aligned and pointer-sized thus atomic)? Or at least it's an addendum that could be added? Just to be sure to understand the situation: if a thread goes through the GOT slot, then all good (aligned, atomic, sees either the old or the new function). However for direct relocations, this is not the case. So a concurrent thread reading the target of a relocation could observe an inconsistent state, right?

yannham (commented):
What is the scenario for unloading the crashtracker lib? Can it happen in the middle of a normal run, or is it something like at exit (if we're unloaded in the wrong order, this causes spurious crashes)?

The issue with unhooking is that the lifecycle for pure relocations that don't go through the GOT/PLT seems to be quite delicate:

## Overwriting with the wrong value

For example, let's say you have static like

```
static void *(*allocate)(size_t) = malloc;
```

Then, during the lifetime of the program, there's a new assignment `allocate = my_personal_allocator`. During unhooking, we might accidentally restore `malloc` because we look at the original relocations, but those that aren't modified if a static variable gets updated. So it's not because we've initially patched a relocation target that it still contains the same thing. I guess we can just check the value of the target first and compare it to `hook_symbol`, although it introduces new fun data races.

## Data races

Since we might overwrite mutable memory that is concurrently used, with misaligned pointer writes, there's no protection at all against data races. This is UB and could very well happen if another thread is trying to read the variable. Maybe we could just skip patching unaligned locations? I really don't expect this to happen a lot, especially for something like `assert`.

## Stale data

Another possibility is that the value has been copied somewhere else. For example in a `Vec<callbacks>` from the static, or even just in a register. Those values can't be traced at all, and thus can't be "unhooked". They will continue to point back to the original hook, which might not exist anymore.

## Should we patch relocations?

I don't know all the tradeoffs involved here, but I feel if we can get our way in practice without touching relocations at all, our life will be easier/safer. It means we'll miss some calls, and I guess we would have to see if it matters or not in practice (like if 0.01% of calls to `assert` are done through direct relocations, then 🤷‍♂️ ).

Not patching unaligned relocation target would already get rid of the data race issue (at least most of it, because there's still a TOCTOU when some threads might have already loaded the old hook address locally in a register as we unhook, but those are 1. unavoidable I guess and 2. the window for those is much, much smaller - we would have to unhook and unload the module between a thread loading the address and actually jumping to the function).

gyuheon0h:
@yannham Yeah. I spent (too much -- IMO shouldve bailed earlier) time trying to force an "unhook" API, but making unhooks that work all the time for all cases and is always consistent and reliable after a "permanent" GOT modification has been a struggle.

The reason I started this is because PHP apparently sometimes unloads the tracer, which the customer application dynamically links to. So, if we ever patch something that gets called later on, it is a segfault as our patch function is no longer mapped. I think I will change the direction of this to implement an unhook that is only to be called if crashtracker is unloaded -- making it clear in the docs that its not intended to be called regularly, and GOT patching through libdd-gotter shouldbe considered permanent.

libdd-gotter/src/elf.rs
```diff
@@ -1185,6 +1289,46 @@ unsafe fn patch_got_entries(
...
+    let mut result = UnhookResult::default();
+
+    for &(addr, original_value) in &hook_result.slots {
+        // SAFETY: caller guarantees `addr` is in a currently-mapped region.
+        if unsafe { guard.make_writable(addr) } {
+            // Use write_unaligned to match patch_got_entries. GOT slots for
+            // JUMP_SLOT/GLOB_DAT are always pointer-aligned, but
+            // R_X86_64_64/R_AARCH64_ABS64 entries in data sections may not be.
+            // SAFETY: `addr` points to a writable GOT slot in mapped memory.
+            unsafe { core::ptr::write_unaligned(addr as *mut usize, original_value) };
```
scottgerring, line 1322: The PR description says:

> This only restores entries whose current value matches hook_fn

But here we are doing an unconditional write - not checking what is in the slot? I reckon we need to guard against making changes that are _not_ undoing changes we made.

scottgerring: looking at the stacked hooks test, I think the contract is also "a caller must LIFO their changes when stacking around a particular symbol", which is not enforced or captured in this signature

libdd-gotter/src/elf.rs
scottgerring: Yeah, I think this needs protection in the implementation rather than just an unsafe precondition, especially because it is exposed through a safe `uninstall_hooks()` wrapper. Maybe each saved slot needs enough ELF object identity and its relocation offset to verify that it still belongs to the same loaded object before writing; otherwise address reuse could cause us to corrupt an unrelated mapping.

Beyond that, there’s another rather gnarly race too:

A thread can load the hook address from the GOT *before* `unhook_symbol` restores the slot, then jump to that saved address after the hook library has been unloaded. Making the slot replacement atomic doesn't help with a pointer that's already been read from the table.

I’m not sure how we can address this. Conceptually, we would need to:

1. prevent new calls through the affected slots
2. wait for callers that may already have loaded or entered the hook to quiesce
3. unpatch the slots
4. complete `dlclose`
5. resume callers

I don’t know how libdatadog can guarantee that sequence inside an arbitrary host process.

libdd-gotter/src/elf.rs
```diff
@@ -986,7 +1001,7 @@ pub struct LookupResult {
 }
 
 /// Result of a successful [`hook_symbol`] call.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
```
scottgerring, line 1004: The API says call once, but `HookResult` is cloneable and `unhook_symbol` only borrows it, so you can clone and restore over and over. This could lead to overwriting changes made after the first unhook, and means that any retry logic is going to be more error prone. 

COuld we make this non-cloneable, and consume it in unhook_symbol?

libdd-gotter/src/elf.rs
```diff
@@ -1105,14 +1135,19 @@ unsafe fn hook_symbol_impl(
...
-        // `dl_iterate_phdr`
+        // `dl_iterate_phdr`.
         let Some(dyn_info) = (unsafe { DynamicInfo::from_phdr(info) }) else {
             return false;
         };
+
+        // Reserve before patching so push is guaranteed not to allocate
+        // after any GOT entry in this library has been overwritten.
+        let count = count_matching_got_entries(&dyn_info, symbol_name_bytes);
+        unsafe { (*slots_ptr).reserve(count) };
```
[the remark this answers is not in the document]
scottgerring: I think this still applies in the current impl: `reserve(count)` prevents allocation whilst patching _the current_ library, but a later library can still grow the vector after earlier libraries have been patched. Maybe i'm missing something here?

libdd-gotter/src/elf.rs
```diff
@@ -1007,6 +1025,31 @@ pub enum HookError {
...
+/// library being unloaded; the caller **must not** unload until all slots
+/// are restored or the process will fault on the next call through them.
+///
+/// [`slots_failed`]: UnhookResult::slots_failed
+#[derive(Clone, Copy, Debug, Default)]
+pub struct UnhookResult {
+    /// Number of GOT slots successfully restored to their pre-hook value.
+    pub slots_restored: usize,
+    /// Number of GOT slots that could not be restored because `mprotect`
+    /// was rejected. These slots still point at the hook function.
```
scottgerring, line 1042: I don't think `slots_failed` can currently guarantee that these slots still point at our hook, because `unhook_symbol` never reads their current values

libdd-gotter/src/elf.rs
```diff
@@ -1130,15 +1190,48 @@ unsafe fn hook_symbol_impl(
...
 /// Patch GOT entries in one library for the target symbol.
 ///
 /// Only patches relocations of type `GLOB_DAT`, `JUMP_SLOT`, or
 /// pointer-width absolute (`R_X86_64_64` / `R_AARCH64_ABS64`).
 /// Narrow or PC-relative relocation types are skipped.
 ///
+/// For each slot that is successfully patched, the original slot value is
+/// read before overwriting and appended to `slots` as `(addr,
+/// original_value)`. This captures IFUNC-resolved addresses and
+/// RELA-addend-adjusted values that differ from the raw symbol address,
```
scottgerring, line 1231: I rabbit holed myself a bit on this one 😆  I don't think we can support non-zero addend relocations (nor should we need to); see #2526

libdd-gotter/src/elf.rs
```diff
@@ -1524,6 +1668,137 @@ mod tests {
...
+    ///    `HookResult::slots`.
+    /// 2. `unhook_symbol` restores exactly those entries using the captured per-slot values
+    /// 3. A second `unhook_symbol` call is a no-op
+    ///
+    /// We use `getpid` because it is a well-known symbol that we can safely
+    /// patch for the duration of the test.
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn test_unhook_symbol_restores_got_entries() {
+        unsafe extern "C" fn dummy_getpid() -> libc::pid_t {
```
scottgerring, line 1682: I think there's a race in the tests - we've got two tests patching the same thing (getpid)
IIRC we're using nextest here, which would solve this in the CI environment, but it would be nice for these to work without a race with regular cargo test running locally I think

libdd-gotter/src/elf.rs
```diff
@@ -1524,6 +1668,137 @@ mod tests {
...
+        assert_ne!(real_pid, -1, "unhooked getpid should return the real PID");
+        assert!(real_pid > 0, "real PID should be positive");
+
+        // A second unhook writes the same original values back — safe and
+        // idempotent since unconditional write_unaligned is used.
+        unsafe { unhook_symbol(&hook_result) };
+        let pid_after_second = unsafe { libc::getpid() };
+        assert!(pid_after_second > 0 && pid_after_second != -1);
+
+        let _ = patched_count; // used above in assertion on hook_result
```
scottgerring, line 1730: > // used above in assertion on hook result
Is it - it looks like the `UnhookResult` values are discarded?

I think there's a few missing assertions - slots_restored is patched_count, slots_failed is zero, is_complete is true.

scottgerring (commented):
Some comments inline! Zooming out for a minute, I have a couple of high-level points too. First, I think the PR description has drifted from the implementation; it would be good to update it to reflect the current state.

More fundamentally, is unpatching the right solution here? There are several moderately gnarly conditions we would need to guarantee to make this safe, and in at least one case I cannot see how we could do so inside an arbitrary host process. See the inline comments for details.

As an alternative, could we ensure that the DSO containing the crashtracker hooks remains mapped? For example, could we use `RTLD_NODELETE` and, on shutdown, disable crashtracker while leaving the hook forwarding path and its state alive, without trying to unpatch anything? We'd need to make sure this does what it says, but this seems like it would be easier to reason about (and is the sort of thing we need to rely on elsewhere - for the memory profiling).

PR #2343, opened 2026-08-11
fix(ffe): report malformed flags as parse errors

leoromanovsky:
## Motivation

Ruby, PHP, and Python share libdatadog's FFE evaluator. The updated canonical corpus requires malformed flag keys to report parse errors without blocking valid siblings, and exposes a SemVer precedence bug for build metadata.

## Changes

- bump `ffe-system-test-data` to `ea8b5cc5ce335109f11f3efbc5fd608f98a3ca54`
- expose the existing per-flag invalid-configuration sentinel as `ERROR/PARSE_ERROR` through the FFI
- keep unrecognized flags on `ERROR/FLAG_NOT_FOUND`
- compare SemVer values by precedence so build metadata is ignored
- add focused FFI and SemVer regressions and update the canonical harness

## Decisions

- retain `FlagConfigurationInvalid` as the per-flag sentinel; reserve `ConfigurationParseError` for whole-payload failures
- keep independent flag compilation and immutable full-map replacement as the isolation boundary
- put shared behavior here instead of duplicating evaluator logic in Ruby, PHP, and Python
- keep the separate regex-fixture proposal in ffe-system-test-data#21 out of scope

## Validation

- `cargo test -p libdd-ffe -p libdd-ffe-ffi -p libdd-ffe-test-suite --no-fail-fast`
- `cargo +stable clippy -p libdd-ffe -p libdd-ffe-ffi -p libdd-ffe-test-suite --all-targets -- -D warnings`
- `cargo ffi-test` built the release FFI artifacts and 14 example targets before an intentional stop during unrelated examples
- `git diff --check`

### DIFF
--- a/libdd-ffe-ffi/src/assignment.rs
+++ b/libdd-ffe-ffi/src/assignment.rs
@@ -370,7 +370,7 @@ impl From<&ResolutionDetails> for Reason {
             Ok(assignment) => assignment.reason.into(),
             Err(EvaluationError::FlagDisabled) => Reason::Disabled,
             Err(EvaluationError::DefaultAllocationNull) => Reason::Default,
-            Err(EvaluationError::FlagConfigurationInvalid) => Reason::Default,
+            Err(EvaluationError::FlagConfigurationInvalid) => Reason::Error,
             Err(_) => Reason::Error,
         }
     }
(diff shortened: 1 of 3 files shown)


leoromanovsky:
Superseded by #2339, which already implements the same canonical fixture bump, malformed-flag parse-error mapping, and SemVer precedence fix with green CI.

PR #2344, opened 2026-08-12
feat(profiling): replace weight with per-allocation unbiased byte …

scottgerring:
### What

 Redefines weight in the ddheap:alloc USDT from nsamples * sampling_interval to a proper per-allocation unbiased byte estimator. Matches [proposed changes to design](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1743). 

 ### Why

 nsamples * sampling_interval is a coarse measure that counts how many sampling-interval boundaries an allocation crosses, then multiplies by the interval. This has two problems:

 1. Excessive variance for large allocations. A 4 MiB allocation with a 512 KiB interval might cross 7 or 9 boundaries depending on where the random sampling points fall, producing a weight of 3.5-4.5 MiB instead of
    consistently ~4 MiB. The allocation is virtually certain to be sampled, so its weight should be close to its actual size.
 2. Size-bias distortion. Downstream consumers that count alloc_objects = 1 per emitted event systematically overstate large-allocation stacks relative to small ones, since large allocations are more likely to be sampled.

 The new weight enables consumers to derive both unbiased byte totals (alloc_space = weight) and unbiased object counts (alloc_objects = weight / size) from a single value.

 This brings the sampler in line with [jemalloc's approach to correcting sampling bias https://github.com/jemalloc/jemalloc/blob/dev/doc_internal/PROFILING_INTERNALS.md), which the design already references as the model for the in-process sampler.

 ### Changes

 - weight semantics: now the unbiased estimate of allocated bytes represented by this sampled allocation, computed from its inclusion probability under Poisson sampling: p = 1 - exp(-size / interval), weight = size / p.
   Uses expm1 for numerical stability.
 - sample() return type: changed from uint64_t (nsamples * interval) to bool (was this allocation sampled?). The nsamples counter is removed; the sampling loop now only advances remaining_bytes state.
 - Weight calculation on slow path only: allocation_weight() is called after the sampling decision, so no floating-point work is added to the unsampled fast path.
 - Zero-size allocations: treated as unsampled (weight = 0) to preserve the weight == 0 means "not sampled" invariant and avoid leaving the reentry guard open.
 - Documentation: all references to nsamples * interval updated across headers, comments, and README.
 - Tests: focused deterministic tests for weight at small/interval/large sizes, monotonic convergence, edge cases, and a statistical regression test verifying sum(weights) ≈ total_allocated.

### DIFF
--- a/libdd-profiling-heap-sampler/include/datadog/heap/allocation_requested.h
+++ b/libdd-profiling-heap-sampler/include/datadog/heap/allocation_requested.h
@@ -19,9 +19,12 @@
  *                the profiler via the ddheap:alloc USDT.
  *   - alignment: passed through so dd_allocation_created can place the
  *                user pointer correctly relative to the raw pointer.
- *   - weight:    0 if not sampled; otherwise the unbiased size
- *                estimator (nsamples * interval) to attribute to this
- *                allocation.
+ *   - weighted_bytes: 0 if not sampled; otherwise the estimated allocated
+ *                bytes represented by this sampled allocation, computed
+ *                from the allocation size and its probability of being
+ *                sampled under Poisson sampling:
+ *                  p = 1 - exp(-size / interval)
+ *                  weighted_bytes = size / p
  *
  * Always pair with dd_allocation_created(), even if the allocator fails.
  */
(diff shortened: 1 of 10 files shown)


libdd-profiling-heap-sampler/include/datadog/heap/allocation_requested.h
```diff
@@ -19,9 +19,12 @@
  *                the profiler via the ddheap:alloc USDT.
  *   - alignment: passed through so dd_allocation_created can place the
  *                user pointer correctly relative to the raw pointer.
- *   - weight:    0 if not sampled; otherwise the unbiased size
- *                estimator (nsamples * interval) to attribute to this
- *                allocation.
+ *   - weight:    0 if not sampled; otherwise the estimated allocated
```
r1viollet: I know that weight is standard, though it feels like a poor name ? This is the extrapolated_size considering probability of being sampled.

scottgerring: Big ACK. Naming is hard! i've gone with `weighted-bytes` - i feel like this is a sweet spot between re-using the language and making the units & intention clear

PR #2350, opened 2026-08-13
refactor: migrate HTTP & networking deps to workspace level (phase 4bis)

yannham:
# What does this PR do?

Moves the HTTP & networking dependencies to `[workspace.dependencies]`:
`http`, `bytes`, `http-body-util`, `httpmock`, `reqwest`, `httparse`,
`mime`, `sha2`, `base64`, `percent-encoding`, `rustls-platform-verifier`,
`flate2`, `zstd`, `fastrand`, `once_cell` and `regex`. This is an additional phase before phase 5 of the original plan.

All workspace entries are version-only with `default-features = false`; each
leaf crate opts into exactly the features it needs (mostly `std`). Two
documented exceptions where the no-default-features build is unusable:

- `flate2` needs a compression backend; `rust_backend` (pure Rust, no C
  dependency) is the baseline.
- `httpmock` 0.8.0-alpha.1 does not compile with `default-features = false`:
  its own `readers.rs` calls `req.cookies()` unconditionally while that
  method is gated behind the `cookies` feature (upstream alpha bug).

# Motivation

Single source of truth per dependency version, no accidental skew, shorter
leaf-crate manifests.

# Additional Notes

- `Cargo.lock` only bumps `headers` 0.4.0 -> 0.4.1 and `sha1` 0.10.6 ->
  0.10.7 (transitive of httpmock's `cookies` feature).
- Dev-only duplicates (`serial_test`, `proptest`, `duplicate`,
  `pretty_assertions`) are left for a later step.

# How to test the change?

- `cargo check --workspace`
- `cargo nextest run` on the affected crates
- `cargo ffi-test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2176,11 +2176,11 @@ dependencies = [
 
 [[package]]
 name = "headers"
-version = "0.4.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "322106e6bd0cba2d5ead589ddb8150a13d7c4217cf80d7c4f682ca994ccc6aa9"
+checksum = "b3314d5adb5d94bcdf56771f2e50dbbc80bb4bdf88967526706205ac9eff24eb"
 dependencies = [
- "base64 0.21.7",
+ "base64 0.22.1",
  "bytes",
  "headers-core",
  "http 1.1.0",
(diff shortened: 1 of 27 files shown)


leoromanovsky (approved):

hoolioh (approved):

morrisonlevi (approved):
Approved for profiling.

PR #2356, opened 2026-08-17
chore(trace-protobuf)!: sync span.proto with datadog-agent

Aaalibaba42:
# What does this PR do?

Agent changed the proto, we must sync

# Motivation

CI passing, and having the correct proto

# Additional Notes

-

# How to test the change?

ci

### DIFF
--- a/.github/workflows/verify-proto-files.yml
+++ b/.github/workflows/verify-proto-files.yml
@@ -3,7 +3,7 @@ on:
   pull_request:
     types: [ opened, synchronize, reopened ]
 env:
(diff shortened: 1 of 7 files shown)


.github/workflows/verify-proto-files.yml
bantonsson: Shouldn't this diff the generated `pb.idx.rs` as well?

Aaalibaba42: Yes you are right

bantonsson (approved):

PR #2362, opened 2026-08-17
perf(trace-exporter): forward encoded v0.4 payloads

BridgeAR:
Already-finalized v0.4 payloads currently pay for a full decode and serialization before agent transport. This forwards them unchanged when no configured consumer needs span objects.

Stats, trace filters, alternate formats, log output, agentless export, and OTLP keep the decoded path.


paullegranddc (commented):
This is not great as you are basically copying all spans to obfuscate them.
This is going to be superseded by this PR 
https://github.com/DataDog/libdatadog/pull/2418

PR #2369, opened 2026-08-19
feat(trace-utils)!: add v1-native OTLP encoder brick

anais-raison:
# What does this PR do?

Adds a v1-native equivalent of the existing v0.4 OTLP mapper, as groundwork for making `v1::Span`/`v1::TraceChunk` the exporter's canonical internal type:

- `otlp_encoder::map_traces_to_otlp_v1` (libdd-trace-utils), mapping `v1::TraceChunk`/`v1::Span` directly to the prost OTLP IR

# Motivation

Part of [APMSP-2812](https://datadoghq.atlassian.net/browse/APMSP-2812): migration of the exporter from v0.4 to v1 with isolated bricks first before one final breaking PR that will handle the actual swap.

# Additional Notes

Pure addition, no behavior change — this function isn't called anywhere yet, so there's no regression risk. Also bumps `build_resource`/`proto_kv` in the existing v0.4 mapper to pub(super) so the v1 mapper can reuse them without duplicating that logic.


libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+                .iter()
+                .map(|(k, v)| proto_kv(k.borrow().to_string(), attr_value_to_proto(v)))
+                .collect(),
+        }),
+    }
+}
+
+/// Merge chunk-level and span-level attributes into one ordered, deduplicated key/value list,
+/// span wins on key collision. Both `VecMap`s are deduplicated first (last value wins).
+fn merged_attrs_v1<'a, T: TraceData>(
```
ajgajg1134, line 67: This function seems a bit inefficient with multiple passes over the structure, does this implementation read cleaner to you? My rust sense seems to think so but I'm not confident
```
  fn merged_attrs_v1<'a, T: TraceData>(
      chunk: &'a TraceChunk<T>,
      span: &'a Span<T>,
  ) -> Vec<(&'a str, &'a AttributeValue<T>)> {
      let mut merged: Vec<(&'a str, &'a AttributeValue<T>)> =
          Vec::with_capacity(chunk.attributes.len() + span.attributes.len());
      let mut index: std::collections::HashMap<&'a str, usize> =
          std::collections::HashMap::with_capacity(merged.capacity());

      for (k, v) in chunk.attributes.iter().chain(span.attributes.iter()) {
          let key = (*k).borrow();
          match index.get(key) {
              Some(&i) => merged[i].1 = v,
              None => {
                  index.insert(key, merged.len());
                  merged.push((key, v));
              }
          }
      }
      merged
  }
```

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+            order.push(key);
+        }
+    }
+    order.into_iter().map(|k| (k, merged[k])).collect()
+}
+
+/// OTLP status (code, optional message) for a span: ERROR with the error message when
+/// `span.error` is set, otherwise UNSET. See [`super::mapper::span_status`] for the v0.4
+/// equivalent and the `error.msg`/`error.message` rationale.
+fn span_status_v1<T: TraceData>(
```
ajgajg1134, line 93: Would it be possible to do this status lookup on the hashmaps before they're flattened to arrays to have faster lookups? or does that make this harder to read / deduce on?

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+                spans: all_spans,
+                schema_url: String::new(),
+            }],
+            schema_url: String::new(),
+        }],
+    }
+}
+
+#[cfg(test)]
+mod tests_v1 {
```
ajgajg1134, line 343: Are there any tests that validate the promoted fields like `env, version, component, _dd.origin, _dd.p.dm` are still available in otlp? I don't see where these get translated from v1 but I might be missing something!

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+        AttributeValue::String(s) => ProtoValue::StringValue(s.borrow().to_string()),
+        AttributeValue::Bool(b) => ProtoValue::BoolValue(*b),
+        AttributeValue::Int(i) => ProtoValue::IntValue(*i),
+        AttributeValue::Float(f) => ProtoValue::DoubleValue(*f),
+        AttributeValue::Bytes(b) => ProtoValue::BytesValue(b.borrow().to_vec()),
+        AttributeValue::List(items) => ProtoValue::ArrayValue(ProtoArrayValue {
+            values: items
+                .iter()
+                .map(|it| ProtoAnyValue {
+                    value: Some(attr_value_to_proto(it)),
```
ajgajg1134, line 51: I think for the recursion here we probably want a safety limit to avoid stack overflow (e.g. for the datadog agent it uses 64 which is the recommended default in otel)

Not needed in this PR but we should validate a similar limit is used in the decoders (and probably encoders too to be safe in case a native span / non-decoded trace has too deep an attribute value)

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+            .collect(),
+        dropped_attributes_count: 0,
+        // See `super::mapper::map_span_link`: bit 31 is an internal "explicitly set" sentinel
+        // that must not leak into OTLP's flags field.
+        flags: link.flags & !SPAN_LINK_FLAGS_SET_SENTINEL,
+    }
+}
+
+fn map_span_events_v1<T: TraceData>(events: &[SpanEvent<T>]) -> (Vec<ProtoEvent>, usize) {
+    const MAX_EVENTS_PER_SPAN: usize = 128;
```
ajgajg1134, line 214: hmm where did this limit come from?

ajgajg1134 (changes requested):

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! v1-native analog of [`super::mapper::map_traces_to_otlp`]: maps `v1::TraceChunk`/`v1::Span`
+//! directly to the prost OTLP IR.
+//!
+//! Unlike the v0.4 mapper, this needs no hex/decimal round trip *and* no meta/metrics flatten:
+//! v1's `trace_id: [u8; 16]` is already the full 128-bit id, `span_kind: SpanKind` already carries
+//! the OTLP span kind, sampling priority lives on the chunk, and `AttributeValue` maps onto OTLP's
+//! `AnyValue` losslessly (string/bool/int/double/bytes/array/kvlist), so attributes are converted
+//! recursively with no flattening or type coercion.
```
yannham, line 11: This information might have its role somewhere, but I find it a bit out of place at the module-level documentation (too detailed/specific).

```suggestion
```

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
yannham: Unless I'm missing something, if we're just merging two lists and deduping, why can't we just do  `chunks.attribute.iter().chain(span.attributes.iter()).collect::<HashMap<&'a str, &'a AttributeValue<T>>().into_iter().collect:<Vec<_>>()` ? (you can introduce intermediate variables with type annotation if you don't like the turbofish operator, but my point is, I don't think we need any imperative logic beyond iterator combinators here)

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+                .defensive_dedup()
+                .iter()
+                .map(|(k, v)| proto_kv(k.borrow().to_string(), attr_value_to_proto(v)))
+                .collect(),
+        }),
+    }
+}
+
+/// Merge chunk-level and span-level attributes into one ordered, deduplicated key/value list,
+/// span wins on key collision. Both `VecMap`s are deduplicated first (last value wins).
```
yannham, line 66: Does this mean that the vecmap are deduped before this function, or during this function? If the former, this is useless work, since we're deduping here anyway. If the latter please ignore my comment.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+/// capped at `MAX_ATTRIBUTES_PER_SPAN`.
+fn collect_span_attributes_v1<T: TraceData>(
+    span: &Span<T>,
+    merged: &[(&str, &AttributeValue<T>)],
+    resource_service: &str,
+    otel_trace_semantics_enabled: bool,
+) -> (Vec<ProtoKeyValue>, usize) {
+    // Pre-size to avoid reallocations as attributes accumulate. Upper bound is the 4 synthetic
+    // attrs plus every merged chunk+span attribute, clamped to the per-span cap.
+    let capacity = (4 + merged.len()).min(MAX_ATTRIBUTES_PER_SPAN);
```
yannham, line 124: It's really a nitpick, but we could size it exactly I believe; we typically computed `promoted` later in this function.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+    (attrs, dropped)
+}
+
+fn map_span_link_v1<T: TraceData>(link: &SpanLink<T>) -> ProtoLink {
+    ProtoLink {
+        trace_id: link.trace_id.to_vec(),
+        span_id: link.span_id.to_be_bytes().to_vec(),
+        trace_state: {
+            let ts = link.tracestate.borrow();
+            if ts.is_empty() {
```
yannham, line 194: Is this branch really needed? Wouldn't `to_string` take care of that?

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+            name: ev.name.borrow().to_string(),
+            attributes: ev
+                .attributes
+                .defensive_dedup()
+                .iter()
+                .map(|(k, v)| proto_kv(k.borrow().to_string(), attr_value_to_proto(v)))
+                .collect(),
+            dropped_attributes_count: 0,
+        });
+    }
```
yannham, line 228: (Note: thanks to the magic of ExactSizedIterator, this should build the `Vec` with only one allocation, being equivalent to the `with_capacity` solution)

```suggestion
    let out = events.iter().take(MAX_EVENTS_PER_SPAN).map(|ev| ProtoEvent {
        time_unix_nano: ev.time_unix_nano,
        name: ev.name.borrow().to_string(),
        attributes: ev
            .attributes
            .defensive_dedup()
            .iter()
            .map(|(k, v)| proto_kv(k.borrow().to_string(), attr_value_to_proto(v)))
            .collect(),
        dropped_attributes_count: 0,
      }).collect();
```

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+    };
+    let merged = merged_attrs_v1(chunk, span);
+    let (attributes, dropped_attributes_count) = collect_span_attributes_v1(
+        span,
+        &merged,
+        resource_service,
+        otel_trace_semantics_enabled,
+    );
+    let (code, message) = span_status_v1(span, &merged);
+    let trace_state = merged
```
yannham, line 253: Same question as @ajgajg1134 above: could we keep merged attributes as a hashmap a bit longer, so that those kind of lookup are constant time instead of linear time?

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+        trace_state,
+        parent_span_id,
+        flags,
+        name: span.resource.borrow().to_string(),
+        // v1's `span_kind` discriminants (Internal=1, Server=2, Client=3, Producer=4,
+        // Consumer=5) already match the OTLP SpanKind enum, so no meta lookup/fallback is
+        // needed like the v0.4 mapper's `span_kind()`/`tag_to_otlp_kind`/`dd_type_to_otlp_kind`.
+        kind: span.span_kind as i32,
+        // OTLP timestamps are unsigned; clamp negatives to 0 so the `as u64` cast can't wrap.
+        start_time_unix_nano: span.start.max(0) as u64,
```
yannham, line 278:
```suggestion
        start_time_unix_nano: u64::try_from(span.start.max).unwrapr_or(0),
```

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+        end_time_unix_nano: span.start.saturating_add(span.duration).max(0) as u64,
```
yannham, line 279:
```suggestion
        end_time_unix_nano: u64::try_from(span.start.saturating_add(span.duration)).unwrap_or(0),
```

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+        attributes,
+        dropped_attributes_count: dropped_attributes_count as u32,
```
yannham, line 281: Can, in a pathological case / adversarial case, `dropped_attributes_count` be greater than `u32::MAX`? If yes, should rather do `u32::try_from(dropped_attributes_count).unwrap_or(u32::MAX)`. If no, I would encourage to write a small comment justifying the `as` conversion is fine because blablabla.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+        events,
+        dropped_events_count: dropped_events_count as u32,
```
yannham, line 283: Same as above.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+        let flags = chunk.priority.map(|p| (p >= 1) as u32).unwrap_or(0);
+        for span in &chunk.spans {
+            all_spans.push(map_span_v1(
+                span,
+                chunk,
+                &resource_info.service,
+                flags,
+                otel_trace_semantics_enabled,
+            ));
+        }
```
yannham, line 322: Matter of taste, but I find iterators more idiomatic than imperatively pushing to a vec. Feel free to ignore if disagreed 🙂 

```suggestion
        all_spans.extend(chunk.spans.iter().map(|span| map_span_v1(
            span,
            chunk,
            &resource_info.service,
            flags,
            otel_trace_semantics_enabled,
        ));
```

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,694 @@
...
+/// Merge chunk-level and span-level attributes into one ordered, deduplicated key/value list,
+/// span wins on key collision. Both `VecMap`s are deduplicated first (last value wins).
+fn merged_attrs_v1<'a, T: TraceData>(
+    chunk: &'a TraceChunk<T>,
+    span: &'a Span<T>,
+) -> Vec<(&'a str, &'a AttributeValue<T>)> {
+    let mut order: Vec<&'a str> =
+        Vec::with_capacity(chunk.attributes.len() + span.attributes.len());
+    let mut merged: std::collections::HashMap<&'a str, &'a AttributeValue<T>> =
+        std::collections::HashMap::with_capacity(order.capacity());
```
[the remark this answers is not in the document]
yannham: We do perform one allocation per span (in `O(chunk attrs + span attrs)` memory), which is not great, but the solution proposed by Codex is probably a dumb `O(n²)` dedup, which is not great either, unless we are sure to have very, very few attributes.

I don't see an obvious, simple solution here. We could pre-compute a hashmap for chunks attrs with a bit of additional capacity and clone it for every span as a base, but it's a bet: if we didn't allocate enough spare capacity and we need to resize, we have to re-hash all elements, which is more costly than what's done here.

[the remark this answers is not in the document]
yannham: Maybe another solution would be to sort both vecs, and then merge-sort-dedup :thinking: ? If attribute order is not significant, this would avoid the hashmap allocation and hashing. And we could merge the chunk attrs vec once and for all.

Honestly this shouldn't really be in this PR, I'm just thinking out loud for potential follow-ups.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: Clarified it in the comment but neither `VecMap` input is assumed pre-deduplicated, so this function handles that too (last value wins).

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: Fixed 👍
Moved the total count computation before allocating the `Vec`

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: Yup, that's fixed, `merged_attrs_v1` returns the index `HashMap` that it builds internally with the `Vec` so that it use it for O(1) lookup instead.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: Done 👍

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: Added the test `dedicated_fields_are_promoted_to_otlp_attributes` covering it !

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: Created [APMSP-4000](https://datadoghq.atlassian.net/browse/APMSP-4000) to handle this separately 👍

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: I just reused the existing `MAX_EVENTS_PER_SPAN` constant that come from the existing v04 otlp encoder, so this stays consistent with that encoder (and the same applies to `MAX_ATTRIBUTES_PER_SPAN` constant as well).

For the question "where does 128 comes from" : I found out it matches the OpenTelemetry SDK's default SpanLimits (EventCountLimit and AttributeCountLimit), it's not an OTLP protocol constraint. So it looks like a defensive copy of the OTel SDK defaults rather than a documented requirement. 

Worth noting it's also a partial implementation of that model cause both limits are hardcoded (and thus not configurable via `OTEL_SPAN_*_COUNT_LIMIT` like the spec seem to expect).

I can open a ticket if we want to revisit this, but keeping both as is here for consistency with the existing v04 encoder.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: Yes, in a pathological/adversarial case (very very very large payload) it could exceed `u32::MAX`. Applied your suggestion to both, and also fixed the same pre-existing pattern in the v04 mapper for consistency.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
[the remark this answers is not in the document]
anais-raison: Agreed, added a point to [APMSP-3786](https://datadoghq.atlassian.net/browse/APMSP-3786) (point 6) to track this merge sort dedup idea as a follow up.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: Went with a single-pass HashMap-index version instead of `chain()` cause HashMap has no defined iteration order, so converting it back to a Vec doesn't preserve the span-before-chunk ordering we need (span attributes must survive truncation at `MAX_ATTRIBUTES_PER_SPAN` before chunk attributes do). The chain approach gets the value precedence right (span overwrites chunk) but not the final ordering, and collecting through a HashMap loses it entirely. Also, `merged_attrs_v1` now returns the index alongside the merged vec, so we need to build and keep that map either way.

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
ajgajg1134: I think this is fine for now, I was mostly curious since I didn't recall seeing this number / limit used before!

ajgajg1134 (approved):

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,840 @@
...
+    }
+}
+
+/// Ordered, deduplicated chunk+span attribute list produced by [`merged_attrs_v1`].
+type MergedAttrs<'a, T> = Vec<(&'a str, &'a AttributeValue<T>)>;
+/// Key to position index into a [`MergedAttrs`] list, for O(1) lookup of one specific key.
+type MergedAttrsIndex<'a> = std::collections::HashMap<&'a str, usize>;
+
+/// Merge chunk-level and span-level attributes into one ordered, deduplicated key/value list,
+/// span wins on key collision. Neither input `VecMap` is assumed pre-deduplicated: a key repeated
```
yannham:
```suggestion
/// Merge chunk-level and span-level attributes into one ordered, deduplicated key/value list.
/// Span wins on key collision. Neither input `VecMap` is assumed pre-deduplicated: a key repeated
```

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
```diff
@@ -0,0 +1,840 @@
...
+    let span_claimed = merged.len();
+    for (k, v) in chunk.attributes.iter() {
+        let key = (*k).borrow();
+        match index.get(key) {
+            Some(&i) if i < span_claimed => {}
+            Some(&i) => merged[i].1 = v,
+            None => {
+                index.insert(key, merged.len());
+                merged.push((key, v));
+            }
```
yannham: You can use the [entry API](https://doc.rust-lang.org/beta/std/collections/struct.HashMap.html#entry-api) of hashmaps to avoid doing a double lookup in the `None` case: you lookup once, and then you can use the entry to insert or read the existing value accordingly.

yannham (approved):

libdd-trace-utils/src/otlp_encoder/mapper_v1.rs
anais-raison: Good idea 👍

mabdinur (approved):

PR #2370, opened 2026-08-19
feat(trace-utils): add v1-native agentless JSON encoder brick

anais-raison:
# What does this PR do?

Adds a v1-native equivalent of the existing v0.4 agentless JSON encoder, as groundwork for making `v1::Span`/`v1::TraceChunk` the exporter's canonical internal type:

- `agentless_encoder::encode_payload_from_v1` and its helpers (`collect_attrs_v1`, `flatten_attr_into_v1`, `encode_span_link_v1`, `encode_span_event_v1`) (libdd-trace-utils)

# Motivation

Part of [APMSP-2812](https://datadoghq.atlassian.net/browse/APMSP-2812): migration of the exporter from v0.4 to v1 with isolated bricks first before one final breaking PR that will handle the actual swap.

# Additional Notes

Pure addition, no behavior change — these functions aren't called anywhere yet, so there's no regression risk.


libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,587 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+    }
+}
+
+/// Leaves collected by [`collect_attrs_v1`]: `meta` entries, `metrics` entries, and raw
+/// `meta_struct`-bound `Bytes` entries.
+type CollectedAttrsV1<'a, T> = (
+    Vec<(String, String)>,
+    Vec<(String, f64)>,
+    Vec<(&'a <T as TraceData>::Text, &'a <T as TraceData>::Bytes)>,
+);
```
paullegranddc: it's inferable from the typing, but I would prefer and actual struct with named field rather than an aliased tuple here

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,587 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+) -> CollectedAttrsV1<'a, T> {
+    let span_attrs_dd = span.attributes.defensive_dedup();
+    let chunk_attrs_dd = chunk.attributes.defensive_dedup();
+    let merged_attrs = span_attrs_dd
+        .iter()
+        .filter(|(k, _)| !PROMOTED_ATTR_KEYS_V1.contains(&(*k).borrow()))
+        .chain(chunk_attrs_dd.iter().filter(|(k, _)| {
+            !PROMOTED_ATTR_KEYS_V1.contains(&(*k).borrow())
+                && !span_attrs_dd.iter().any(|(k2, _)| k2 == *k)
+        }));
```
paullegranddc, line 559: Is this really necessary?

We perfom `len(PROMOTED_ATTR_KEYS_V1) * len(span_attr) + (len(PROMOTED_ATTR_KEYS_V1) +  len(span_attr) ) * len(chunk_attr)` string comparison 🤔 

Span level attributes should not collide with chunk level attribute, and if they do they should be promoted before we reach the encoder IMO.

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,587 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+    match v {
+        AttributeValueV1::String(s) => meta_out.push((key.clone(), s.borrow().to_owned())),
+        AttributeValueV1::Bool(b) => {
+            meta_out.push((key.clone(), if *b { "true" } else { "false" }.to_owned()))
+        }
+        AttributeValueV1::Int(i) => metrics_out.push((key.clone(), *i as f64)),
+        AttributeValueV1::Float(f) => metrics_out.push((key.clone(), *f)),
+        AttributeValueV1::Bytes(_) => {
+            // Callers filter `Bytes` out before recursing; unreachable in practice.
+        }
```
paullegranddc: This is not true
If one of the nested `AttributeValueV1::List` item is a Bytes attribute it is not filteredt

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,587 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+    for (k, v) in merged_attrs {
+        match v {
+            AttributeValueV1::Bytes(b) => bytes_attrs.push((k, b)),
+            _ => {
+                key_buf.clear();
+                key_buf.push_str(k.borrow());
+                flatten_attr_into_v1(&mut key_buf, v, &mut meta_leaves, &mut metrics_leaves);
+            }
+        }
+    }
```
paullegranddc: I think we can assume that most of of attributes are scalars.
This means that the attribute key will be a single segment, that does not need flattening and we can borrow it as is from the `merged_attrs`.

So in order to not copy keys in the non-nested case we can use a Vec<(Cow<'a, str>, Cow<'a, str>)> and borrow the key (and values if the scalar is a string, or a value we can turn in a &'static like  booleans)

Something like this I guess

```suggestion
    let mut meta_leaves: Vec<(Cow<'a, str>, Cow<'a, str>)> = Vec::new();
    let mut metrics_leaves: Vec<(Cow<'a, str>, f64)> = Vec::new();
    let mut bytes_attrs: Vec<(&T::Text, &T::Bytes)> = Vec::new();
    let mut key_buf = String::new();
    for (k, v) in merged_attrs {
        match v {
            AttributeValueV1::Bytes(b) => bytes_attrs.push((k, b)),
            AttributeValueV1::String(s) => meta_leaves.push(Cow::Borrowed(k.as_str(), Cow::Borrowed(s)),
            // same for metrics/ bools/ints/floats
            _ => {
                key_buf.clear();
                key_buf.push_str(k.borrow());
                flatten_attr_into_v1(&mut key_buf, v, &mut meta_leaves, &mut metrics_leaves);
            }
        }
    }
```

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,587 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+    (
+        dedup_first_wins_v1(meta_leaves),
+        dedup_first_wins_v1(metrics_leaves),
```
paullegranddc: why is the dedup here necessary if we already filtered out trace chunks and span attrs with the same paths?

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,587 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+            for (key, val) in metrics_leaves.iter() {
+                match key.as_str() {
+                    "_trace_root" => trace_root_seen = true,
+                    "_top_level" => {
+                        metrics.serialize_entry(key, &(*val as u32))?;
+                        continue;
+                    }
+                    _ => {}
+                }
+                metrics.serialize_entry(key, val)?;
```
paullegranddc, line 844: We should skip `NaN` and `Inf` metric values as is done for the v04 encoder

libdd-trace-utils/src/agentless_encoder/mod.rs
anais-raison: Agreed this logic probably shouldn't live duplicated inside the encoder. _Following up on the Slack discussion_, the plan is to extract this into a shared attribute-promotion pass used by both this encoder and the msgpack encoder (tracked in this follow-up ticket [APMSP-3980](https://datadoghq.atlassian.net/browse/APMSP-3980)). Keeping it as-is here for now to keep this PR scoped.

libdd-trace-utils/src/agentless_encoder/mod.rs
anais-raison: This dedup isn't guarding against span/chunk key collisions cause those are already filtered above. It's for keys that only collide after flattening, (e.g. a literal attribute named `a.0` next to a List named a whose first item flattens to the same dotted key `a.0`). It's the same rationale as `dedup_first_wins` in `span_v1.rs`.

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,615 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+    let has_attributes = attrs_dd.iter().any(|(_, v)| is_supported_event_attr_v1(v));
+    if has_attributes {
+        map.serialize_entry(
+            "attributes",
+            &ser_fn!(<T: TraceData> |ser, attrs_dd: &'a crate::span::vec_map::DedupedVecMap<'a, T::Text, AttributeValueV1<T>>| {
+                let mut attrs = ser.serialize_map(None)?;
+                for (k, v) in attrs_dd.iter().filter(|(_, v)| is_supported_event_attr_v1(v)) {
+                    let key: &str = k.borrow();
+                    attrs.serialize_entry(key, &ser_fn!(<T: TraceData> |ser, v: &'a AttributeValueV1<T>| {
+                        encode_event_attr_value_v1(ser, v)
```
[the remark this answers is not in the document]
paullegranddc: This is a real concern. This is not how the other agentless encoder works

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -27,13 +27,19 @@
 //! TODO: span normalization (service/name/resource/type truncation + defaults)
 
 use crate::span::v04::{AttributeAnyValue, AttributeArrayValue, Span, SpanEvent, SpanLink};
+use crate::span::v1::{
```
yannham: Very nit: it's a matter of taste, but if there's a lot of conflicting structures in `v1` and `v04`, another possibility is to keep the prefix (which is short here) and use `v1::Span` instead of `SpanV1`, etc. (also it's more consistent since `v04::Span` here is just `Span`, instead of `SpanV04`)

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,615 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+        SpanKind::Server => Some("server"),
+        SpanKind::Client => Some("client"),
+        SpanKind::Producer => Some("producer"),
+        SpanKind::Consumer => Some("consumer"),
+    }
+}
+
+/// Drops entries whose key was already seen, keeping the first occurrence: two distinct
+/// attributes can flatten to the same dotted key.
+fn dedup_first_wins_v1<'a, V>(mut leaves: Vec<(Cow<'a, str>, V)>) -> Vec<(Cow<'a, str>, V)> {
```
yannham: If we mutate `leaves` in place, any reason to take a `Vec` and return a `Vec`, instead of just operating on `&mut Vec` in-place?

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,615 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+    let mut keep = keep.into_iter();
+    leaves.retain(|_| keep.next().unwrap_or(false));
+    leaves
+}
+
+/// Recursively flattens a `List`/`KeyValue` attribute into dotted-key leaf entries for the
+/// `meta` (string-valued), `metrics` (numeric), and `meta_struct` (`Bytes`) buckets. Only called
+/// once an attribute is known to be nested: the resulting key is always a freshly built dotted
+/// string, so it's always owned — unlike the top-level scalar fast path in [`collect_attrs_v1`],
+/// which can borrow the key/value directly from the source attribute.
```
yannham: I find those overly technical AI comments a bit hard to digest. Maybe adding a simple example would help make the role of this function clear?

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,615 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+fn flatten_attr_into_v1<'a, T: TraceData>(
+    key: &mut String,
```
yannham, line 496: Is `key` mutated in place? It's not entirely clear from the function description. Should we take a copy instead? I wonder if it can be surprising for the caller to get `k` to be changed after this function.

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,615 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+    SerializeMap::end(map_ser)?;
+    Ok(bytes)
+}
+
+fn encode_trace_v1<T: TraceData, S: Serializer>(
+    ser: S,
+    chunk: &TraceChunk<T>,
+    metadata: &TracerMetadata,
+) -> Result<S::Ok, S::Error> {
+    let mut map = ser.serialize_map(None)?;
```
yannham, line 651: Nit: I think we can pre-compute the size here, avoiding useless allocations: `(!metadata.env.is_empty() as usize) + (....) + 1`

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,615 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+    map.end()
+}
+
+fn encode_span_v1<'a, T: TraceData, S: Serializer>(
+    ser: S,
+    chunk: &'a TraceChunk<T>,
+    span: &'a SpanV1<T>,
+    is_first_in_trace: bool,
+) -> Result<S::Ok, S::Error> {
+    let mut map = ser.serialize_map(None)?;
```
yannham, line 696: Same: is it doable to precompute the size?

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,615 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+    }))
+    .ok()?;
+    Some(truncate_with_ellipsis(s, MAX_META_VALUE_LEN))
+}
+
+fn encode_span_link_v1<T: TraceData, S: Serializer>(
+    ser: S,
+    link: &SpanLinkV1<T>,
+) -> Result<S::Ok, S::Error> {
+    let mut map = ser.serialize_map(None)?;
```
yannham, line 895: Same question: could we easily pre-compute the size?

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,615 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+            | AttributeValueV1::Int(_)
+            | AttributeValueV1::Float(_)
+    )
+}
+
+fn encode_span_event_v1<T: TraceData, S: Serializer>(
+    ser: S,
+    event: &SpanEventV1<T>,
+) -> Result<S::Ok, S::Error> {
+    let mut map = ser.serialize_map(None)?;
```
yannham, line 986: I think we can precompute the size here: something like `2 + (has_attributes as usize)`

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -431,6 +437,615 @@ fn serialize_scalar<S: serde::Serializer, T: TraceData>(
...
+                            }
+                            seq.end()
+                        }),
+                    )?;
+                    m.end()
+                }),
+            )?;
+            map.end()
+        }
+        other => encode_event_scalar_v1(ser, other),
```
yannham, line 1034: It's a matter of taste, but I wonder if we shouldn't just inline `encode_event_scalar_v1` here. It's still relatively small and avoid an `unreachable`. Though one could argue that keep smaller function is better, so no strong opinion either way

yannham (approved):

libdd-trace-utils/src/agentless_encoder/mod.rs
anais-raison: Updated the doc comment to make it clearer 👍

libdd-trace-utils/src/agentless_encoder/mod.rs
anais-raison: Done 👍

libdd-trace-utils/src/agentless_encoder/mod.rs
anais-raison: @yannham I looked into this but actually for `serde_json`, the `len` passed to `serialize_map` doesn't actually help. The `len` only matter in the special-case of `len == Some(0)` (where it closes the object immediately). For any other value it takes the exact same path as `None` and starts writing fields one by one directly to the output writer. There's no length-prefixed header, no upfront `Vec::with_capacity` sized from the lenght and `Compound::Map` doesn't store the length either. So `Some(3)` vs `None` produce byte-identical output, no allocation avoided (verified that with a small test comparing the two to be 100% sure).

yannham: Ah, I think I had msgpack in head, but you're right that it doesn't make a difference for JSON which is streaming-style anyway. It could though if in the future agentless move from JSON to a different encoding, so I would say if it's trivial to pre-compute (just re-order some declaration), it's a sanity step that doesn't cost much to do anyway. If it's too complicated or if it does mangle the code, then it's probably not worth it :+1:

anais-raison: You're right 👍 
It's done, precomputed the size, it was in fact trivial here.

libdd-trace-utils/src/agentless_encoder/mod.rs
anais-raison: Keeping it separate cause `encode_event_scalar_v1` has two call sites (array-element loop and scalar fallback), so inlining would duplicate the match rather than remove the `unreachable!`.

PR #2373, opened 2026-08-19
feat(ffe): expose observeFullEvaluationData config-level FFI getter

vjfridge:
## Summary

Adds the top-level UFC `observeFullEvaluationData` boolean end-to-end through libdatadog and exposes a config-level FFI getter so SDKs can read it without re-parsing the raw JSON string. This is the libdatadog half of the "Protecting PII in flagevaluations track" initiative; the server-side UFC field ships via ddoghq/dd-source#22826 and the RC schema via ddoghq/dd-go#3627.

- When `true`: SDK emits unhashed `targeting_key` + full evaluation context (today's behavior).
- When `false` (default): SDK emits a SHA-256 hex hash of the targeting key and omits the context.

Default is `false` (privacy-preserving). A fresh deploy is a no-op for customers until an SDK ships the read-side change and someone toggles the field on for an env.

## Changes

- `UniversalFlagConfigWire` (`libdd-ffe`): new `#[serde(default)] bool observe_full_evaluation_data` field, a sibling of `environment` at the UFC top level. `#[serde(default)]` keeps older configurations (without the key) parsing to `false`.
- `CompiledFlagsConfig`: carries the field through the wire -> compiled transform.
- `Configuration::observe_full_evaluation_data()`: Rust getter, mirroring the existing `created_at()` / `environment()` accessors.
- `ddog_ffe_configuration_get_observe_full_evaluation_data` (`libdd-ffe-ffi`): new `#[no_mangle]` C accessor returning `bool`. The `datadog/ffe.h` header regenerates via cbindgen.
- `examples/ffi/ffe.c`: prints the flag after `ddog_ffe_configuration_new` to demonstrate usage.

## Test plan

- [x] `cargo check -p libdd-ffe -p libdd-ffe-ffi`
- [x] `cargo +nightly-2026-07-26 fmt --all -- --check`
- [x] `cargo +nightly-2026-07-26 clippy -p libdd-ffe -p libdd-ffe-ffi --all-targets -- -D warnings`
- [x] `cargo nextest run -p libdd-ffe -p libdd-ffe-ffi` (incl. 4 new unit tests: field defaults to false when absent, parses true when present, exposed via `Configuration` true/false)
- [x] `cargo test -p libdd-ffe --doc`
- [x] `cargo ffi-test` — `ffe` C example compiles, links against the regenerated header (new symbol present), and runs green against the `ffe-system-test-data` fixture.

Refs: FFL-3038

### DIFF
--- a/examples/ffi/ffe.c
+++ b/examples/ffi/ffe.c
@@ -223,6 +223,9 @@ int main(int argc, char* argv[]) {
     ddog_ffe_Handle_Configuration config = config_result.ok;
     printf("  Configuration loaded successfully\n");
 
+    bool observe_full_evaluation_data = ddog_ffe_configuration_get_observe_full_evaluation_data(config);
+    printf("  observeFullEvaluationData: %s\n", observe_full_evaluation_data ? "true" : "false");
+
     // Step 2: Create evaluation context with targeting key and attributes
     printf("\nStep 2: Creating evaluation context...\n");
 
--- a/examples/ffi/ffe.c
+++ b/examples/ffi/ffe.c
@@ -226,6 +226,13 @@ int main(int argc, char* argv[]) {
     bool observe_full_evaluation_data = ddog_ffe_configuration_get_observe_full_evaluation_data(config);
     printf("  observeFullEvaluationData: %s\n", observe_full_evaluation_data ? "true" : "false");
 
+    // The test fixture omits `observeFullEvaluationData`, so the privacy-preserving default applies.
+    if (observe_full_evaluation_data) {
+        fprintf(stderr, "Expected observeFullEvaluationData to default to false\n");
+        ddog_ffe_configuration_drop(&config);
+        return 1;
+    }
+
     // Step 2: Create evaluation context with targeting key and attributes
     printf("\nStep 2: Creating evaluation context...\n");
 
--- a/examples/ffi/ffe.c
+++ b/examples/ffi/ffe.c
@@ -226,13 +226,6 @@ int main(int argc, char* argv[]) {
     bool observe_full_evaluation_data = ddog_ffe_configuration_get_observe_full_evaluation_data(config);
     printf("  observeFullEvaluationData: %s\n", observe_full_evaluation_data ? "true" : "false");
 
-    // The test fixture omits `observeFullEvaluationData`, so the privacy-preserving default applies.
-    if (observe_full_evaluation_data) {
-        fprintf(stderr, "Expected observeFullEvaluationData to default to false\n");
-        ddog_ffe_configuration_drop(&config);
-        return 1;
-    }
-
     // Step 2: Create evaluation context with targeting key and attributes
     printf("\nStep 2: Creating evaluation context...\n");
 
--- a/libdd-ffe-ffi/src/configuration.rs
+++ b/libdd-ffe-ffi/src/configuration.rs
@@ -48,3 +48,19 @@ pub unsafe extern "C" fn ddog_ffe_configuration_drop(config: *mut Handle<Configu
     // SAFETY: the caller must ensure that config is a valid handle
     unsafe { Handle::free(config) };
 }
+
+/// Get the config-level `observeFullEvaluationData` flag.
+///
+/// Opt-in boolean for emitting full flag evaluation data from SDKs to Datadog. Flag evaluation data may 
+/// contain PII; defaults to `false` for privacy.
+///
+/// # Safety
+///
+/// `config` must be a valid `Configuration` handle created by `ddog_ffe_configuration_new`.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_ffe_configuration_get_observe_full_evaluation_data(
+    config: Handle<Configuration>,
+) -> bool {
+    // SAFETY: the caller must ensure that config is a valid handle.
+    unsafe { config.as_ref() }.observe_full_evaluation_data()
+}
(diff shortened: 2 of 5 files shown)


vjfridge:
## Note to self: Follow-ups to propagate to SDK repos 

This PR adds the `ddog_ffe_configuration_get_observe_full_evaluation_data` symbol to libdatadog, but it only takes effect once SDKs wrap it so they dont need to re-parsing the raw JSON. Required follow-ups, in merge order:

1. **libdatadog release** Nothing downstream can consume the getter until a release goes out for this PR.
2. **datadog/libdatadog-rb** bump the vendored libdatadog to the release from step 1, then add `observe_full_evaluation_data` to the `Core::FeatureFlags::Configuration` C extension wrapping `ddog_ffe_configuration_get_observe_full_evaluation_data`
3. **dd-trace-rb** consume the new `libdatadog-rb` release and read `observe_full_evaluation_data` from the C extension instead of parsing the config string.

### Other language SDKs to check
The same getter should be exposed by the other libdatadog FFI consumers once they pick up the release:
- Python (`dd-trace-py` / libdatadog Python bindings) — wrap the new symbol if it ships FFE bindings.
- Any other SDK whose C extension links `libdd-ffe-ffi` and currently parses the UFC JSON for this field.

libdd-ffe-ffi/src/configuration.rs
```diff
@@ -48,3 +48,19 @@ pub unsafe extern "C" fn ddog_ffe_configuration_drop(config: *mut Handle<Configu
     // SAFETY: the caller must ensure that config is a valid handle
     unsafe { Handle::free(config) };
 }
+
+/// Get the config-level `observeFullEvaluationData` flag.
+///
+/// Opt-in flag for emitting full flag-evaluation data from SDKs to Datadog. May contain PII;
+/// defaults to `false` for privacy. Not a per-flag evaluation result.
```
vjfridge:
```suggestion
/// Opt-in boolean for emitting full flag evaluation data from SDKs to Datadog. Flag evaluation data may 
/// contain PII; defaults to `false` for privacy.
```

libdd-ffe/src/rules_based/ufc/compiled_flag_config.rs
```diff
@@ -26,6 +26,9 @@ pub(crate) struct CompiledFlagsConfig {
     pub created_at: Timestamp,
     /// Environment this configuration belongs to.
     pub environment: Environment,
+    /// Opt-in flag for emitting full flag-evaluation data from SDKs to Datadog. May contain PII;
+    /// defaults to `false` for privacy.
```
vjfridge:
```suggestion
    /// Opt-in boolean for emitting full flag evaluation data from SDKs to Datadog. Flag evaluation data may
    /// contain PII; defaults to `false` for privacy.
```

libdd-ffe/src/rules_based/ufc/models.rs
```diff
@@ -16,6 +16,10 @@ pub(crate) struct UniversalFlagConfigWire {
     pub created_at: Timestamp,
     /// Environment this configuration belongs to.
     pub environment: Environment,
+    /// Opt-in flag for emitting full flag-evaluation data from SDKs to Datadog. May contain PII;
+    /// defaults to `false` for privacy.
```
vjfridge:
```suggestion
    /// Opt-in boolean for emitting full flag evaluation data from SDKs to Datadog. Flag evaluation data may
    /// contain PII; defaults to `false` for privacy.
```

pavlokhrebto (approved):

pavlokhrebto:
lgtm, please address Codex review comments

libdd-ffe-ffi/src/configuration.rs
```diff
@@ -48,3 +48,94 @@ pub unsafe extern "C" fn ddog_ffe_configuration_drop(config: *mut Handle<Configu
...
+///
+/// # Safety
+///
+/// `config` must be a valid `Configuration` handle created by `ddog_ffe_configuration_new`.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_ffe_configuration_get_observe_full_evaluation_data(
+    config: Handle<Configuration>,
+) -> bool {
+    // SAFETY: the caller must ensure that config is a valid handle.
+    unsafe { config.as_ref() }.observe_full_evaluation_data()
```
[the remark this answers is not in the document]
vjfridge: Thanks for flagging this. The reported panic only occurs if a caller passes a zeroed handle or reuses one after dropping it. Both cases violate the function’s documented requirement that config must be a valid handle. The current code creates the configuration, calls the getter while it is valid, and drops it afterward. 

This getter also follows the existing handle-access pattern used by the other FFE getters. Adding panic handling only here would be inconsistent and would turn an invalid-handle programming error into a normal false result. I think that is unnecessary for this PR. 

If we want stronger protection against invalid handles, we should address it consistently across the full FFE FFI API in a separate change.

examples/ffi/ffe.c
```diff
@@ -223,6 +223,16 @@ int main(int argc, char* argv[]) {
     ddog_ffe_Handle_Configuration config = config_result.ok;
     printf("  Configuration loaded successfully\n");
 
+    bool observe_full_evaluation_data = ddog_ffe_configuration_get_observe_full_evaluation_data(config);
+    printf("  observeFullEvaluationData: %s\n", observe_full_evaluation_data ? "true" : "false");
+
+    // The test fixture omits `observeFullEvaluationData`, so the privacy-preserving default applies.
+    if (observe_full_evaluation_data) {
```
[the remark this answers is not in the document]
vjfridge: Fixed in [fix(ffe): allow opt-in configs in C example](https://github.com/DataDog/libdatadog/pull/2373/commits/f1baf071f9be3aa10904149c2965658d228fdd3a)

Aaalibaba42 (approved):
lgtm o/

PR #2382, opened 2026-08-20
perf(trace-utils)!: use pooled spans on the send path

paullegranddc:
# Motivation

Allocation of collections is one of the most expensive things we perform in the tracer (with allocation of Strings). Being able to reuse allocated capacity should improve performance.

This wires the span pool added in the stacked prerequisite PR (#2477) into the live send path so recycled span allocations skip alloc churn on flush.

# Changes

 - Add `PooledChunks` through the send path in libdd-data-pipeline:
     - `TraceExporter::send_trace_chunks` / `send_trace_chunks_async` now take `PooledChunks<'_, T>` instead of `Vec<Vec<Span<T>>>`.
     - Internal dispatch functions (`send_agentless_traces_inner`, `send_otlp_traces_inner`, `prepare_traces_payload`, `serialize_payload`) now borrow the spans via `&[Vec<Span<T>>]` instead of consuming them, keeping them alive for recycling when the `PooledChunks` drops.
     - The OTLP mapper (`map_traces_to_otlp`) and the v0.5 converter (`convert_trace_chunks_v04_to_v05`, `from_v04_span`) likewise move to borrowed inputs.
     - `drop_chunks` now operates on `PooledChunks<T>`, returning dropped chunks (and dropped sub-spans) to the attached pool instead of freeing them.
     - Adds an optional `span_pool` to `DefaultExport` that wraps chunks before passing them to the exporter.
 - Call sites that do not want to pool use `PooledChunks::unpooled`, a zero-overhead wrapper: the FFI entry point and the examples pass `unpooled` so their behavior is unchanged.

# Additional Notes

Stacked on #2477 (which adds the pool types); merge that first, then retarget this PR to `main`.

This is a breaking API change for `send_trace_chunks` / `send_trace_chunks_async` (signature moves from owned `Vec<Vec<Span>>` to `PooledChunks`). Downstream callers can wrap with `PooledChunks::unpooled` for unchanged behavior, or adopt the pool to recycle allocations.


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -841,15 +849,17 @@ impl<
             if traces.is_empty() {
                 return Ok(AgentResponse::Unchanged);
             }
-            return self.send_otlp_traces_inner(traces, config).await;
+            // The OTLP mapper transforms spans into a different representation and consumes
```
ekump: I don't think this comment is accurate? 

```suggestion
            // The OTLP mapper borrows the spans and builds a separate OTLP 
           // representation, the original spans are recycled to the pool when `traces` drops. 
```

libdd-trace-utils/src/span/span_pool.rs
```diff
@@ -0,0 +1,236 @@
...
+/// * We will keep the **maximum** number of spans ever used by the program alive, even if memory
+///   usage scales down
+/// * As spans get reused and atributes data structure are pushed and popped, they will tend to grow
+///   to have the maximum size of attributes
+///
+/// Dropping a fixed pct of spans returned ensures that we eventually free memory
+/// if span usage spikes, and then goes down.
+fn drop_policy() -> bool {
+    const PCT_OF_SPANS_RETURNED_DROPPED: f64 = 0.1;
+    rand::thread_rng().gen_bool(PCT_OF_SPANS_RETURNED_DROPPED)
```
ekump: This is going to get called on every span being flushed, right? Isn't `thread_rng()` going to be expensive? Could you do it at the chunk level instead? Or, does it have to be random at all? Could you just drop every X spans deterministically in order to shrink the pool?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -793,8 +796,13 @@ impl<
 
     async fn send_trace_chunks_inner<T: TraceData>(
         &self,
-        mut traces: Vec<Vec<Span<T>>>,
+        mut traces: PooledChunks<'_, T>,
     ) -> Result<AgentResponse, TraceExporterError> {
+        // `traces` is a `PooledChunks`: keeping it owned (rather than moving its inner `Vec`
```
ekump, line 810: Not sure if this is the right place, but should we document somewhere that only sampled spans make it to the pool?

libdd-trace-utils/src/span/v05/mod.rs
ekump: I think this comment is stale now, span isn't consumed anymore.

libdd-data-pipeline/src/trace_exporter/mod.rs
ekump: Ah, I guess you note it [here](https://github.com/DataDog/libdatadog/pull/2382/changes#diff-a9338b4e6c08a5a82d881a9d6f191f366f8c5c5155d110173594ce4a4453c575R93-R94).

libdd-trace-utils/src/span/span_pool.rs
```diff
@@ -0,0 +1,236 @@
...
+            name: BytesString::from_slice(name.as_bytes()).unwrap(),
+            ..Default::default()
+        }
+    }
+
+    #[test]
+    fn returned_spans_are_recycled_and_reset() {
+        let pool = SpanPool::<crate::span::BytesData>::new(100);
+        {
+            // No drop policy control here, but with a single span it is very likely retained.
```
ekump: "very likely" is another way of saying flaky. 

What is this test actually covering? won't `s.name` be the same whether it comes from the pool or is a new span?

Should we have a `DropPolicy` that's injectable? Something like...

 ```rust                                                                                                                                                                                                                                                                                  
   #[derive(Debug)]                                                                                                                                                                                                                                                                       
   enum DropPolicy {                                                                                                                                                                                                                                                                      
       Random(f64),                                                                                                                                                                                                                           
       EveryN { n: usize, counter: AtomicUsize },     // deterministic, no global                                                                                                                                                                                                         
       Never,                                         // tests only?                                                                                                                                                                                                                            
   }                                                                                                                                                                                                                                                                                      
   impl DropPolicy {                                                                                                                                                                                                                                                                      
       fn should_drop(&self) -> bool {                                                                                                                                                                                                                                                    
           match self {                                                                                                                                                                                                                                                                   
               DropPolicy::Random(p) => rand::thread_rng().gen_bool(*p),                                                                                                                                                                                                                  
               DropPolicy::EveryN { n, counter } =>                                                                                                                                                                                                                                       
                   counter.fetch_add(1, Ordering::Relaxed) % n == 0,                                                                                                                                                                                                                      
               DropPolicy::Never => false,                                                                                                                                                                                                                                                
           }                                                                                                                                                                                                                                                                              
       }                                                                                                                                                                                                                                                                                  
   }                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                          
   #[derive(Debug, Clone)]                                                                                                                                                                                                                                                                
   pub struct SpanPool<T: TraceData> {                                                                                                                                                                                                                                                    
       queue: crossbeam_channel::Sender<Span<T>>,                                                                                                                                                                                                                                         
       receiver: crossbeam_channel::Receiver<Span<T>>,                                                                                                                                                                                                                                    
       drop_policy: Arc<DropPolicy>,                                                                                                                                                                                                              
   }  
```

**Review by ekump — APPROVED** [review]
My only real concern is the potential flaky test, otherwise LGTM

**Comment by ekump** [comment]
Should we have benchmark tests too?

File: libdd-trace-utils/src/span/span_pool.rs (hunk 2 above)
Comment (paullegranddc) [reply libdd-trace-utils/src/span/span_pool.rs]: I changed it to run on every "chunk", which is less often and I changed the random generator to rand::SmallRng (with the state living in a thread local) for performance.

After a bit of benchmarking, the cost is kind of negligeable (a few %) vs a no-op

File: libdd-data-pipeline/src/trace_exporter/mod.rs (hunk 3 above)
Comment (paullegranddc) [reply libdd-data-pipeline/src/trace_exporter/mod.rs:810]: Updated so that `drop_chunks` and `filter_traces` now return dropped spans and chunks to the pool

**Comment by paullegranddc** [comment]
> Should we have benchmark tests too?

Added benchmarks, as it turns out beating the allocator is not that easy and I had to add a thread local cache from which to pull chunks to make the perf gain obvious.
Now creating/populating/dropping and populating spans is about 30% faster than doing it through allocations

File: libdd-trace-utils/src/span/span_pool.rs (hunk 4 above)
Comment (paullegranddc) [reply libdd-trace-utils/src/span/span_pool.rs]: So the thing is, this test is not "flaky" in the sense it fail from time to time. If there were a bug it could pass.
I added more spans so that the likelihood it passes flakily is ridiculously low.

File: libdd-trace-utils/src/span/trace_utils.rs [hunk 5]
```diff
@@ -186,15 +188,25 @@ where
...
+            let drop = if sampled_indices.peek().copied() == Some(i) {
+                sampled_indices.next();
+                false
+            } else {
+                true
+            };
+            i += 1;
+            drop
+        });
+        pool.add_spans(dropped_spans);
```
VianneyRuhlmann: It seems a bit weird that drop_chunks has to handle pooling. Mayb spans can be wrapped with a custom Drop logic that handles it. Also having to rely on add_spans to poll the iterator for the filtering to be applied seems error prone.

paullegranddc: That's not ideal as that means each chunk has to be a special type that has a reference to the pool.

If we want to abstract it a bit better, I could add a PooledChunkRef type handed out in the inner iterator with a  "retain_if" method that returns dropped spans but all of that seems like a lot of abstraction to remove two calls to the pool

libdd-data-pipeline/src/trace_exporter/mod.rs
paullegranddc: adressed

libdd-trace-utils/src/span/trace_utils.rs
paullegranddc: Ok, added the abstraction mentionned. Now changes to drop_spans and filter_traces are minimal

VianneyRuhlmann (approved):

PR #2384, opened 2026-08-20
fix(crashtracker): hold the collector connection open through symbolization

gyuheon0h:
[PROF-15799](https://datadoghq.atlassian.net/jira/software/c/projects/PROF/boards/11?selectedIssue=PROF-15799)

# What does this PR do?

The crashing process blocks in `wait_for_pollhup` until the receiver closes the collector connection. `receive_report_from_stream` owned that stream and dropped it on return, so the process was released *before* symbolization ran.

Symbolization's normalization phase reads `/proc/<pid>/maps`, so it can race the crashing process's teardown. When it lost, normalization failed, and a frame that fails to normalize has no ELF location to retry against it

`receive_report_from_stream` now borrows the stream instead of consuming it. `receiver_entry_point` owns it, holds it open across symbolization, and drops it immediately after.

Only the Unix-socket sidecar path is affected. Under `receiver_entry_point_stdin` the stream wraps `tokio::io::stdin()`, and dropping that handle never closed fd 0.

`test_receive_report_keeps_collector_connection_open` asserts the peer sees no hangup after the report is returned, and does see one once the receiver drops the stream. Verified it fails against the old ownership model.

Ran the sidecar bin_test 1000 times and no failures -- it failed ~8% of the time before.

# Motivation

# Additional Notes

# How to test the change?

[PROF-15799]: https://datadoghq.atlassian.net/browse/PROF-15799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/RFCs/0015-crashtracker-all-thread-collection.md
+++ b/docs/RFCs/0015-crashtracker-all-thread-collection.md
@@ -231,24 +231,28 @@ The process source needs `/proc/{pid}` to exist. The ELF retry reads the
 binary from disk and needs nothing from the live process, which is what
 keeps function names available once the process is gone.
 
-Whether that matters depends on how the receiver was reached:
+Normalization still depends on that process, since it reads
+`/proc/{pid}/maps`, and a frame that fails to normalize has no ELF
+location to retry against and stays unsymbolized. The ELF retry
+therefore does not by itself make symbolization independent of the
+crashing process; it only covers the second phase.
 
-- **Unix-socket sidecar** (`async_receiver_entry_point_unix_listener` and
-  `async_receiver_entry_point_unix_socket`): the crashing process is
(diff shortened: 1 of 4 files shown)


gyuheon0h:
* **#2384** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Aaalibaba42 (approved):

PR #2386, opened 2026-08-20
feat(compression): use zrip for wasm targets

rochdev:
# What does this PR do?

Keeps the existing zstd backend for native targets and selects the pure-Rust zrip backend for wasm32 compression paths in profiling and trace utilities.

It also adds a small WASM-only runtime test package. The harness compiles the standalone production compression modules directly and runs two round-trip tests under Node, covering trace compression plus the profiling profile and observation codecs.

# Motivation

Native zstd remains substantially faster, while zrip is lighter and performs better for WASM. Selecting the backend by target preserves native performance and removes the need for a C toolchain when compiling these compression paths to WASM.

# Additional Notes

The wire format and Content-Encoding remain zstd-compatible.

This PR was generated by Codex.

# How to test the change?

The existing WASM CI job now installs the matching wasm-bindgen test runner and executes:

- env CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=wasm-bindgen-test-runner cargo test -p libdd-wasm-tests --target wasm32-unknown-unknown

Additional validation performed locally:

- cargo test -p libdd-trace-utils --no-default-features --features compression zstd_compression_roundtrips
- cargo clippy -p libdd-wasm-tests --target wasm32-unknown-unknown --tests -- -D warnings
- cargo +nightly-2026-07-26 fmt --all -- --check

The WASM runtime suite executes two tests and both pass. The repository-wide cargo deny check still reports existing default-branch advisory and license-policy failures, including time 0.3.41 and spin 0.9.8.


paullegranddc (approved):

libdd-profiling/src/pprof/test_utils.rs
```diff
@@ -6,11 +6,11 @@ use libdd_profiling_protobuf::prost_impls::{Profile, Sample};
 fn deserialize_compressed_pprof(encoded: &[u8]) -> anyhow::Result<Profile> {
     use prost::Message;
 
-    // The zstd bindings use FFI so they don't work under miri. This means the
-    // buffer isn't compressed, so simply convert to a vec.
+    // Native zstd uses FFI and is unavailable under Miri, where the default
+    // profile codec remains uncompressed.
```
bengl, line 10: Why change this comment? The related config change seems to be about wasm, but I don't see that in the comment.

libdd-profiling/src/profiles/compressor.rs
```diff
@@ -223,7 +267,7 @@ impl<C: ProfileCodec> Compressor<C> {
     /// - `size_hint`: beginning capacity for the output buffer. This is a hint for the starting
     ///   size, and the implementation may use something different.
     /// - `max_capacity`: the maximum size for the output buffer (hard limit).
-    /// - `compression_level`: see [`zstd::Encoder::new`] for the valid range.
+    /// - `compression_level`: must be supported by the target's zstd encoder.
```
bengl, line 270: Are the ranges different for each encoder? Should they be codified here as enums?

libdd-trace-utils/src/send_with_retry/compression.rs
```diff
@@ -27,12 +27,19 @@ pub fn compress(data: Vec<u8>, strategy: CompressionStrategy) -> (Vec<u8>, Compr
...
             let writer = Vec::with_capacity((data.len() / 10).max(256));
-            zstd::Encoder::new(writer, level)
+            #[cfg(not(target_arch = "wasm32"))]
+            let result = zstd::Encoder::new(writer, level).and_then(|mut e| {
+                e.write_all(&data)?;
+                Ok((e.finish()?, strategy))
+            });
+            #[cfg(target_arch = "wasm32")]
+            let result = zrip::FrameEncoder::new(writer, level)
+                .map_err(std::io::Error::other)
```
bengl, line 37: It looks like this map_err is the only difference between the two. In either case we do the write_all call. Can you shorten the length of these branches, since there's very little different between the two? For example, initialization should probably be abstracted out into a helper function elsewhere.

PR #2389, opened 2026-08-21
feat(data-pipeline): add runtime-independent agentless sending

rochdev:
# What does this PR do?

Adds a runtime-independent agentless data-pipeline send path. Rust owns trace
preparation, JSON encoding, compression, retry policy, timeouts, backoff, and
response handling; the host provides only HTTP and sleep capabilities.

The existing `libdd-data-pipeline` exporter now uses the same high-level send
path, so native and WASM consumers share the behavior without requiring a
specific async runtime.

# Motivation

Node.js needs equivalent native and WASM data-pipeline implementations without
embedding Tokio and Hyper in the native extension. Keeping the send lifecycle
in Rust preserves retry behavior and makes the capability boundary reusable by
future targets.

# Additional Notes

- Request bodies are compressed once before the retry loop.
- Native HTTP/runtime dependencies remain optional.
- Node-specific v0.4 trace decoding remains in `libdatadog-nodejs`.

_This PR was generated by Codex._

# How to test the change?

- `cargo test -p libdd-data-pipeline-core --features compression`
- `cargo test -p libdd-trace-utils send_with_retry`
- `cargo test -p libdd-data-pipeline agentless`
- `cargo clippy -p libdd-data-pipeline-core --all-targets --features compression -- -D warnings`
- `cargo clippy -p libdd-trace-utils --lib --tests -- -D warnings`
- `cargo clippy -p libdd-data-pipeline --lib -- -D warnings`


libdd-trace-utils/src/send_with_retry/mod.rs
```diff
@@ -21,6 +22,65 @@ pub type Attempts = u32;
...
+/// call to [`Self::request`] creates a fresh request with a cheaply cloned
+/// [`Bytes`] body, allowing a host runtime to implement retries without using
+/// libdatadog's HTTP client or async executor.
+#[derive(Clone, Debug)]
+pub struct PreparedRequest {
+    target: Endpoint,
+    payload: Bytes,
+    headers: HeaderMap,
+    compression_strategy: CompressionStrategy,
+}
```
paullegranddc: How is this different from an `http::Request<Bytes>` ?

bengl (dismissed):
Seems fine to me, but @DataDog/apm-common-components-core should probably review.

libdd-data-pipeline-core/src/agentless.rs
```diff
@@ -0,0 +1,244 @@
...
+use thiserror::Error;
+
+const AGENTLESS_MAX_RETRIES: u32 = 2;
+const AGENTLESS_RETRY_DELAY_MS: u64 = 1000;
+
+/// Default timeout for one agentless request attempt.
+pub const DEFAULT_AGENTLESS_TIMEOUT: Duration = Duration::from_secs(15);
+
+/// Agentless trace request configuration.
+#[derive(Clone, Debug)]
```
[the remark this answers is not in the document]
rochdev: Comment is for an old implementation that has since been replaced.

libdd-data-pipeline-core/src/agentless.rs
```diff
@@ -0,0 +1,244 @@
...
+    );
+
+    #[cfg(feature = "compression")]
+    let compression_strategy = CompressionStrategy::Zstd { level: 1 };
+    #[cfg(not(feature = "compression"))]
+    let compression_strategy = CompressionStrategy::None;
+
+    Ok(PreparedAgentlessRequest {
+        request: PreparedRequest::new(target, json_body, headers, compression_strategy),
+        retry_strategy,
```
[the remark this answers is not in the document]
rochdev: Comment is for an old implementation that has since been replaced.

libdd-data-pipeline-core/src/agentless.rs
```diff
@@ -0,0 +1,244 @@
...
+
+/// Prepares already-decoded v0.4 traces for host transport.
+pub fn prepare_agentless_traces_request<T: TraceData>(
+    traces: Vec<Vec<libdd_trace_utils::span::v04::Span<T>>>,
+    metadata: &TracerMetadata,
+    config: &AgentlessTraceConfig,
+) -> Result<PreparedAgentlessRequest, PrepareAgentlessError> {
+    let trace_count = traces.len();
+    let json_body = libdd_trace_utils::agentless_encoder::encode_payload(&traces, metadata)
+        .map_err(PrepareAgentlessError::Serialization)?;
```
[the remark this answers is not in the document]
rochdev: Comment is for an old implementation that has since been replaced.

libdd-common/src/lib.rs
```diff
@@ -21,8 +21,10 @@ pub mod azure_app_services;
 #[cfg(not(target_arch = "wasm32"))]
 pub mod cc_utils;
 #[cfg(not(target_arch = "wasm32"))]
+#[cfg(feature = "http-client")]
 pub mod connector;
```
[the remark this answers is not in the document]
rochdev: Comment is for an old implementation that has since been replaced.

rochdev:
@codex review

libdd-common/src/lib.rs
```diff
@@ -33,6 +34,7 @@ pub mod cstr;
 pub mod bench_utils;
 pub mod config;
 pub mod error;
+#[cfg(feature = "http-client")]
 pub mod http_common;
```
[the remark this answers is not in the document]
rochdev: Addressed in 366521870. The mini_agent and test-utils features now explicitly enable libdd-common/http-client, and CI checks both no-default-features configurations. Both cargo check configurations, their unit and documentation tests (excluding the Docker-backed integration tests), Rust 1.87 Clippy, formatting, and workflow validation pass. Reply generated by Codex.

libdd-common/Cargo.toml
```diff
@@ -55,15 +55,15 @@ hyper-rustls = { version = "0.27.7", default-features = false, features = [
...
+hyper = { workspace = true, features = ["http1", "client"], optional = true }
+hyper-util = { workspace = true, features = ["http1", "client", "client-legacy"], optional = true }
+http-body = { version = "1.0", optional = true }
+http-body-util = { version = "0.1", optional = true }
+tower-service = { version = "0.3", optional = true }
 cc = "1.1.31"
 pin-project = "1"
 libc.workspace = true
-tokio = { workspace = true, features = ["rt", "rt-multi-thread", "macros", "net", "io-util", "fs", "time"] }
+tokio = { workspace = true, features = ["rt", "rt-multi-thread", "macros", "net", "io-util", "fs", "time"], optional = true }
```
paullegranddc, line 68: Why are these marked as optional. The are already not added for wasm and the codepaths providing the http client are already gated as to compile the crate.

 So I don't see how refactoring libdd-common in necessary

libdd-trace-utils/src/send_with_retry/retry_strategy.rs
```diff
@@ -93,26 +93,51 @@ impl RetryStrategy {
...
-            capabilities
-                .sleep(delay + Duration::from_millis(jitter))
-                .await;
+            let jitter_ms = u64::try_from(jitter.as_millis()).unwrap_or(u64::MAX);
+            if jitter_ms == 0 {
+                delay
+            } else {
+                let randomized_ms = rand::random::<u64>() % jitter_ms;
+                delay.saturating_add(Duration::from_millis(randomized_ms))
+            }
```
paullegranddc: This branch here is kind of useless. It is equivalent if the `jitter_ms == 0 ` branch is removed and we gain just a call to `random` in a case that basically never happens

libdd-data-pipeline-core/src/agentless.rs
```diff
@@ -0,0 +1,312 @@
...
+        self.retry_strategy.retry_delay(attempt)
+    }
+}
+
+/// Converts v0.4 MessagePack traces into a host-transported request.
+pub fn prepare_agentless_v04_request(
+    data: &[u8],
+    metadata: &TracerMetadata,
+    config: &AgentlessTraceConfig,
+) -> Result<PreparedAgentlessRequest, PrepareAgentlessError> {
```
paullegranddc: This is very specific to the way you want to integrate libdatadog in NodeJS, so I would prefer if this function lived in libdatadog-nodejs rather than here.

paullegranddc (commented):
Overall, I'd like to push back against this pull request.
First without the matching libdatadog-nodejs PR it is very hard to reason about which APIs are necessary and which are not.

But more than this, I don't understand why we need the `PreparedRequest` level of abstraction, and this is not really motivated in the PR description.

What is the goal here
1) Is it getting just a bag of data with the request and then doing the http request fully in NodeJS?
2) Or is it to have a `send_agentless` function that takes messagepack &[u8] serialized trace chunk and uses the existing capabilities implementation to drive calling to the node http stack from rust?


While I am not completely against 1), I would much prefer option 2) as it doesn't expose as much details of libdatadog to nodeJS and you should just be able to hook the exisitng WASM capabilities implementation.

We should just do something like 
```
NodeJS                      |   Rust                                               | NodeJS
send(chunk) -> encode msp   -> send(traces: &[u8]) -> encode() -> send_with_retry ->  http capability impl
```
 and not
 
 ```
NodeJS                        |   Rust                                              | NodeJS                         | Rust             | NodeJS           
prepare(chunk) -> encode msp  -> prepare(traces: &[u8]) -> returns prepared request -> send_rust(prepared_request)   -> send_with_retry ->  http capability 
```

File: libdd-trace-utils/src/send_with_retry/mod.rs [hunk 10]
```diff
@@ -105,66 +177,72 @@ pub async fn send_with_retry<C: HttpClientCapability + SleepCapability>(
...
-            target.set_standard_headers(builder, concat!("Tracer/", env!("CARGO_PKG_VERSION")));
-        for (key, value) in headers {
-            builder = builder.header(key, value);
-        }
-        // headers_mut is only None if the builder is in an error state
-        if let Some(h) = builder.headers_mut() {
-            compression::add_headers(h, compression_strategy);
-        }
-        let req = match builder.body(payload.clone()) {
+        let req = match prepared.request() {
```
paullegranddc: This is not good. 
Calling to `.request()` will do compression on the body but this is inside of the retry loop.

libdd-common/Cargo.toml
rochdev: WASM is only half the equation. Without making them explicitly optional, NAPI-RS was still including them. While some things are build target specific, some other things are not and need to be controllable regardless of the target.

libdd-trace-utils/src/send_with_retry/mod.rs
rochdev: Just a small abstraction but it should be effectively equivalent, let me see if I can simplify.

libdd-trace-utils/src/send_with_retry/retry_strategy.rs
rochdev: I'll try to clean this up.

libdd-data-pipeline-core/src/agentless.rs
rochdev: Let me try to see if this can be removed.

libdd-trace-utils/src/send_with_retry/mod.rs
rochdev: PreparedRequest handles this so it's not an issue with the current implementation, but I'll try to see if we can remove PrepareRequest completely since half the comments are about it.

paullegranddc (approved):
LGTM!

PR #2392, opened 2026-08-21
ci(release): semver-level review

iunanua:
# What does this PR do?

- Handle `async` changes in the method signature
- Improve output truncation


yannham (approved):
It's a bit out of topic for this PR, but I wonder if those kind of non-trivial scripts shouldn't be moved to either Rust (since our CI tooling seems to be going that direction), or at least Python, since Bash is quite bad at handling those kind of non trivial stuff 🤔

iunanua:
> It's a bit out of topic for this PR, but I wonder if those kind of non-trivial scripts shouldn't be moved to either Rust (since our CI tooling seems to be going that direction), or at least Python, since Bash is quite bad at handling those kind of non trivial stuff 🤔

@yannham That's the idea. Someday it'll happen :)

PR #2393, opened 2026-08-21
fix(rc)!: fixup rc to be wasm compatible

bwoebi:
Agentless addition was not fully wasm compilable. Adding it in CI to avoid wasm target regressions.


paullegranddc (approved):

libdd-remote-config/src/fetch/agentless/integration_tests.rs
```diff
@@ -96,6 +96,21 @@ impl MockHttp {
...
+    fn new() -> Self {
+        <Self as HttpClientCapability>::new_client()
+    }
+
+    fn sleep(
+        &self,
+        _duration: std::time::Duration,
+    ) -> impl std::future::Future<Output = ()> + libdd_capabilities::MaybeSend {
+        std::future::pending()
+    }
```
paullegranddc, line 111: Not sure about this. It's only used for timeouts so this works  but if we ever use it to wait it's going to fail the tests 🤔

Maybe use tokio sleep?

bwoebi: Well, the point is _not_ using tokio sleep for wasm?
But well, it's just for the test. When they fail eventually we can switch to a better solution.

paullegranddc: Ah my bad, I'm dumb... didn't realize you also ran the tests on wasm target

libdd-remote-config/Cargo.toml
iunanua, line 64: a quick note: could tokio-util be moved to the `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]`?

PR #2396, opened 2026-08-24
feat: thin lto for non consuming artifacts

Aaalibaba42:
Fat LTO + codegen-units=1 makes the Generate profiling FFI step slow for no reason, this doesn't ship an artifact anyone consumes. Override to thin LTO / codegen-units=16 via CARGO_PROFILE_RELEASE_* env vars for this build only

### DIFF
--- a/.github/workflows/test-ffi.yml
+++ b/.github/workflows/test-ffi.yml
@@ -67,10 +67,13 @@ jobs:
         include:
           - platform: "ubuntu-latest"
             flags: "-C relocation-model=pic"
+            profile: "release-ci"
           - platform: "macos-15"
             flags: "-C relocation-model=pic"
+            profile: "release"
           - platform: "windows-latest"
             flags: "-C target-feature=+crt-static"
+            profile: "release-ci"
     steps:
       - name: Checkout sources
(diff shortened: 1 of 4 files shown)


.github/workflows/test-ffi.yml
```diff
@@ -118,6 +118,8 @@ jobs:
         shell: bash
         env:
           RUSTFLAGS: "${{ matrix.flags }}"
+          CARGO_PROFILE_RELEASE_LTO: "thin"
+          CARGO_PROFILE_RELEASE_CODEGEN_UNITS: "16"
```
hoolioh: Instead of hardcoding these values in the workflow files we could setup a profile for the CI in the Cargo.toml the downside is that you would need to include some plumbering in the builder crate in order to pass the profile. That way we could reuse that profile in libdatadog, libddprof-build and also downstream projects that use the builder. Wdyt? It can be done in a subsequent PR if you want to test this approach first.
Also another advantage is that you can include the profile in the cache key rather than manually setting them depending on the job.

gyuheon0h (approved):
nice this seems clean. thx

hoolioh (approved):

PR #2398, opened 2026-08-24
WIP: fix(crashtracker): unwind from saved crash context

pawelchcki:
## Summary

- seed the stopped crashing thread's registers from the saved signal `ucontext` before remote libunwind collection
- restore the stopped registers before detaching and promote the causal unwind to the primary error stack
- retain all-thread data, synthesize module+offset names for unresolved native frames, and trim leading libc delivery frames for software-generated signals

## Validation

- `cargo check -p libdd-crashtracker --features collector,receiver`
- Linux musl `cargo check -p libdd-crashtracker --features collector,receiver --tests --target x86_64-unknown-linux-musl`
- dd-trace-c optimized `-fomit-frame-pointer` crash integration test

### DIFF
--- a/libdd-crashtracker/src/receiver/entry_points.rs
+++ b/libdd-crashtracker/src/receiver/entry_points.rs
@@ -159,9 +159,128 @@ fn resolve_frames(
             .as_ref()
             .context("Unable to resolve frames: No PID specified")?
             .pid;
-        crash_info.enrich_callstacks(pid)?;
+        let enrichment = crash_info.enrich_callstacks(pid);
+        finish_native_stacks(crash_info);
+        enrichment?;
     }
     #[cfg(not(target_os = "linux"))]
     let _ = (config, crash_info);
     Ok(())
+}
+
+#[cfg(target_os = "linux")]
+fn finish_native_stacks(crash_info: &mut CrashInfo) {
+    // A software-generated fatal signal is captured while raise/pthread_kill is
+    // still on the stack. Those libc delivery frames describe how the signal
+    // arrived, not what application operation requested it. Remove only the
+    // contiguous leading libc segment and leave the first application frame as
+    // the grouping frame.
+    if crash_info
+        .sig_info
+        .as_ref()
+        .is_some_and(|sig_info| sig_info.si_code <= 0)
+    {
+        trim_signal_delivery_frames(&mut crash_info.error.stack);
+    }
+
+    synthesize_module_offsets(&mut crash_info.error.stack);
+    if let Some(threads) = crash_info.error.threads.as_mut() {
+        for thread in threads {
+            synthesize_module_offsets(&mut thread.stack);
+        }
+    }
+}
+
+#[cfg(target_os = "linux")]
+fn trim_signal_delivery_frames(stack: &mut crate::crash_info::StackTrace) {
+    let delivery_frames = stack
+        .frames
+        .iter()
+        .take_while(|frame| {
+            frame.path.as_deref().is_some_and(|path| {
+                let basename = std::path::Path::new(path)
+                    .file_name()
+                    .and_then(|name| name.to_str())
+                    .unwrap_or(path);
+                basename == "libc.so.6"
+                    || basename.starts_with("libc-")
+                    || basename.starts_with("ld-musl-")
+            })
+        })
+        .count();
+    stack.frames.drain(..delivery_frames);
+}
+
+#[cfg(target_os = "linux")]
+fn synthesize_module_offsets(stack: &mut crate::crash_info::StackTrace) {
+    for frame in &mut stack.frames {
+        if frame.function.is_some() {
+            continue;
+        }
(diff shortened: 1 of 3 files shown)


gyuheon0h:
Do you want this to be taken a look at? or is this still WIP

pawelchcki:
@gyuheon0h still WIP. 

I was looking at a crash today. first surfaced through crashtracking, then digging through coredump - I found hte actual crashing stack pointed to elsewhere. IDK how widespread this issue is or was it just my C handler implemetnation doing something really wrong.  I need some time to properly understand what is up

gyuheon0h:
> @gyuheon0h still WIP.
> 
> I was looking at a crash today. first surfaced through crashtracking, then digging through coredump - I found hte actual crashing stack pointed to elsewhere. IDK how widespread this issue is or was it just my C handler implemetnation doing something really wrong. I need some time to properly understand what is up

@pawelchcki 

I noticed this too. `error.stack` is from the crashing thread context at crash point and the stacks look fine there from the crashes I look at day to day. 

However, we also collect the stack for the crashing thread *again* in the receiver as part of all thread collection, and at this point, the two stacks may disagree, sometimes including signal handler and crashtracker frames. Confirming this is what you noticed also?

For example, this crash: https://app.datadoghq.com/error-tracking?query=&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%227e175306-a14a-11f1-b97d-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&from_ts=1787767961397&to_ts=1787854361397&live=true


The main `error.stack` is application provided Python stack. But in `error.threads`, there is the same thread `python` that crashes but with the stack being the "native" stack captured by receiver side all thread collection. And that stack contains a bunch of frames we ideally dont want to include.

raphaelgavache:
Nice `ucontext` would help a lot to identify this crash https://app.datadoghq.com/error-tracking?query=source%3Acrashtracker%20language_name%3Anative&fromUser=true&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22ebac424a-81b1-11f1-9d9e-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&from_ts=1789674261654&to_ts=1790279061654&live=true

any blockers

gyuheon0h:
> Nice `ucontext` would help a lot to identify this crash https://app.datadoghq.com/error-tracking?query=source%3Acrashtracker%20language_name%3Anative&fromUser=true&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22ebac424a-81b1-11f1-9d9e-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&from_ts=1789674261654&to_ts=1790279061654&live=true
> 
> any blockers

@raphaelgavache Hey! QQ, is the service `ds_ray_cluster` using the full e2e libdatadog crashtracker flow? (libdatadog collector + libdatadog receiver)? If so, ucontext should be included in the crash report, but in the telemetry log, I do not see it: [LOG](https://app.datadoghq.com/logs?query=%40instrumented_service%3Ads-ray-cluster%20is_crash%3Atrue%20uuid%3A93c1e4f7-b4fb-4a27-8c92-987892e4bdc6&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AwAAAaDHXLV5mbs_qwAAABhBYURIWExWNUFBRC01RzVGR3NHOHpRQnkAAAAkMTFhMGM3ODktYmMzMS00Yzk3LTkyNWUtM2RmMGJkZWUzOThkAAlgmA&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1789675755513&to_ts=1790280555513&live=true)

We collect the ucontext in the collector and pipe it over to the receiver [here](https://github.com/DataDog/libdatadog/blob/9b339e98215443e3764b741eef06565f3f91a6a6/libdd-crashtracker/src/collector/emitters.rs#L107-L115)

Given that I don't see it, I suspect y'all are using the native crashtracker, which is separate from libdatadog crashtracker?

pawelchcki:
> Given that I don't see it, I suspect y'all are using the native crashtracker, which is separate from libdatadog crashtracker?

We're currently using 1/2 of libdatadog crashtracker.
since we're statically compiling, and we don't want to bundle large blobs in the main binary. In addition the current crashtracker collector is not signal-safe and might crash or deadlock in unexpected places.

The plan was to contribute signal-safe implementation of the collector to libdatadog but that fell through some cracks. 

So for now we're using C reimplementation, that also is not doing the unwinding on its own, so it needs to send the ucntext to libdatadog

gyuheon0h (approved):
I see no issues -- seems like a general improvement. 

It would be nice if we can write a regression `bin_test` for this though

PR #2402, opened 2026-08-24
feat(ffe)!: send the split serial id on exposure events [EX-3425]

danyal002:
# What does this PR do?

Sends the split serial id on FFE exposure events, and adds it to the exposure deduplication cache value.

| File | Change |
|---|---|
| `libdd-ffe/src/telemetry/exposures.rs` | `serial_id` on `FfeExposure`, on the cache value, and on the serialized event |
| `datadog-sidecar-ffi/src/lib.rs` | `serial_id` and `has_serial_id` on the FFI struct, mapped to an `Option` |
| `datadog-sidecar/src/service/{sidecar_server,ffe_exposures_flusher}.rs` | test helpers |

The key is `serial_id` at the top level of each exposure event. An absent value omits the key. The value is not validated here; the flag configuration layer owns validation.

# Motivation

The exposures intake uses the serial id to find the holdout that an allocation comes from. The compiler rewrites a holdout into an ordinary allocation before an SDK receives the flag configuration, so an exposure event records no holdout. The serial id is the only link back to it.

The deduplication cache keyed on allocation key and variant only. A serial id that appeared, or that changed, while both stayed the same was swallowed, and no exposure was sent. This occurs in production: the `ffe.ufc.vac_miss` metric is continuously non-zero, and those misses resolve on a later configuration refresh.

# Additional Notes

Serial ids are zero-based per organization, so `0` is a real value and is the most frequent value across the fleet. A sentinel cannot signal absence, so the FFI carries a value plus a `has_serial_id` flag, matching the existing `FfeResult` convention.

`skip_serializing_if` is on the serialized event only, never on `FfeExposure`, which crosses the bincode IPC boundary to the sidecar.

`dd-trace-php` reaches the intake through this crate, so it needs a submodule bump before it can send the field.

# How to test the change?

```
cargo test -p libdd-ffe --lib --features exposure-events telemetry::exposures
cargo test -p datadog-sidecar-ffi --lib ffe_exposure
```

New tests pin the wire key, a serial id of `0`, omission when absent, an exposure when the serial id appears or changes on an otherwise unchanged assignment, and no second exposure when nothing changes.

*This description was generated by Claude.*

### DIFF
--- a/libdd-ffe/src/telemetry/exposures.rs
+++ b/libdd-ffe/src/telemetry/exposures.rs
@@ -419,7 +419,7 @@ mod tests {
             exposure_with_serial_id("user", "alloc", "variant", 0),
         );
 
-        assert!(first.is_some());
+        assert_eq!(first.unwrap()["exposures"][0].get("serial_id"), None);
         assert_eq!(appeared.unwrap()["exposures"][0]["serial_id"], 0);
     }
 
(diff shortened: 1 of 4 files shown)


libdd-ffe/src/telemetry/exposures.rs
```diff
@@ -66,6 +70,7 @@ impl ExposureDeduplicator {
         let value = ExposureCacheValue {
             allocation_key: exposure.allocation_key.clone(),
             variant: exposure.variant.clone(),
+            serial_id: exposure.serial_id,
```
leoromanovsky, line 73: Please add test cases for:

```
Some(A) -> None
Some(A) -> Some(B) -> Some(A)
```

There should be some

leoromanovsky (approved):

libdd-ffe/src/telemetry/exposures.rs
danyal002: Added in new revision. Thanks!

PR #2408, opened 2026-08-24
fix(ipc)!: gracefully handle SHM errors instead of unwrapping

bwoebi:
This prevents errors especially around ENOSPC on customers environments from stopping the sidecar alltogether.

BREAKING CHANGE: `OneWayShmWriter::write`, `MappedMem::ensure_space`, `ShmLimiterMemory::alloc`, and `ShmLimiterMemory::alloc_with_granularity` in `libdd-ipc` now return `io::Result` instead of panicking on failure.

### DIFF
--- a/datadog-sidecar/src/service/agent_info.rs
+++ b/datadog-sidecar/src/service/agent_info.rs
@@ -132,20 +132,11 @@ impl AgentInfoFetcher {
                             // `SameState` and other-process shm readers would never observe the
                             // update. Leaving `state` unchanged makes us fetch and retry the write
                             // on the next iteration.
-                            let published = match writer {
-                                Some(ref writer) => {
-                                    match writer.write(&serde_json::to_vec(&status.info).unwrap()) {
-                                        Ok(()) => true,
-                                        Err(e) => {
-                                            error!("Failed to write agent info shm segment: {e}");
-                                            false
-                                        }
-                                    }
+                            if let Some(writer) = &writer {
+                                match writer.write(&serde_json::to_vec(&status.info).unwrap()) {
+                                    Ok(()) => state = Some(status.state_hash),
+                                    Err(e) => error!("Failed to write agent info shm segment: {e}"),
                                 }
-                                None => false,
-                            };
-                            if published {
-                                state = Some(status.state_hash);
                             }
                             if let Some(completer) = completer {
                                 complete_fut = Some(completer.complete(status.info));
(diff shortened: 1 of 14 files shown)


datadog-sidecar/src/service/agent_info.rs
```diff
@@ -128,8 +127,25 @@ impl AgentInfoFetcher {
...
+                                            error!("Failed to write agent info shm segment: {e}");
+                                            false
+                                        }
+                                    }
+                                }
+                                None => false,
+                            };
+                            if published {
+                                state = Some(status.state_hash);
                             }
```
yannham: Nit: shorter version (might need to adapt `writer` to `&mut writer` or sth, not sure without the LSP)

```suggestion
                            if let Some(writer) = writer {
                                    match writer.write(&serde_json::to_vec(&status.info).unwrap()) {
                                        Ok(()) => state = Some(status.state_hash),
                                        Err(e) => error!("Failed to write agent info shm segment: {e}"),
                                    }
                                }
}
```

datadog-sidecar/src/service/telemetry.rs
```diff
@@ -398,23 +400,48 @@ impl TelemetryCachedClient {
...
-        } else {
-            warn!("Failed to serialize telemetry data for shared memory");
+    pub fn write_shm_file(&mut self) {
+        let buf = match bincode::serialize(&self.shared) {
+            Ok(buf) => buf,
+            Err(_) => {
+                warn!("Failed to serialize telemetry data for shared memory");
+                return;
+            }
+        };
```
yannham: Nit style: shorter version

```suggestion
        let Ok(buf) = bincode::serialize(&self.shared) else {
            warn!("Failed to serialize telemetry data for shared memory");
            return;
        };
```

datadog-sidecar/src/tracer.rs
```diff
@@ -3,27 +3,40 @@
...
 use libdd_trace_utils::config_utils::trace_intake_url_prefixed;
 use std::borrow::Cow;
 use std::ffi::CString;
-use std::mem::ManuallyDrop;
 use std::str::FromStr;
 use std::sync::{LazyLock, Mutex};
+use tracing::warn;
 
-pub static SHM_LIMITER: LazyLock<Mutex<ManuallyDrop<ShmLimiterMemory<()>>>> = LazyLock::new(|| {
+pub static SHM_LIMITER: LazyLock<Mutex<Option<ShmLimiterMemory<()>>>> = LazyLock::new(|| {
```
yannham, line 15: Why was it `ManuallyDrop` before, and why is that ok to make it dropped automatically now?

yannham (approved):

datadog-sidecar/src/tracer.rs
bwoebi: It was only ManuallyDrop to allow dropping a value directly. With the Option<> to take it, it's not an issue.

Leiyks (approved):

PR #2415, opened 2026-08-25
prototype(trace-utils): honest-lifetime decoder trait scoped for v1

anais-raison:
# What does this PR do?

Adds DeserializableTraceDataLt<'x>, an additive supertrait of DeserializableTraceData carrying the buffer's genuine lifetime instead of the erased 'static that get_mut_slice/intern_skipped_str report. SliceData<'a> implements it with no unsafe code at all, since Self::Bytes is already &'a [u8] - the exact type the method claims to return. BytesData implements it as 'static, reusing its existing (unsafe) methods unchanged.

Scoped to the v1 decoder only: skip_unknown_value/ record_strings_in_value_ref and their v1 callers now require DeserializableTraceDataLt<'x>, giving record_strings_in_value_ref a real lifetime on rmpv::ValueRef instead of a lied 'static. v04/v05 and the original trait/Buffer impl are untouched.

# Motivation

Prototype for the follow-up to PR #2295's review discussion.

# Additional Notes

# How to test the change?

### DIFF
--- a/libdd-trace-utils/src/span/mod.rs
+++ b/libdd-trace-utils/src/span/mod.rs
@@ -210,6 +210,53 @@ impl<'a> DeserializableTraceData for SliceData<'a> {
     }
 }
 
+/// Prototype: an honest-lifetime counterpart to [`DeserializableTraceData`].
+///
+/// `get_mut_slice`/`intern_skipped_str` report `'static` for every implementor, even though for
+/// `BytesData` the backing `Bytes` is heap-owned and never truly `'static` — the lifetime is
+/// erased via `transmute` and "laundered" by immediately deriving an owned `Self::Text` before
+/// it can escape into safe code. For `SliceData<'a>` this erasure is pure overhead: `Self::Bytes`
+/// is already `&'a [u8]`, the real target type, so the transmute exists only to satisfy a trait
+/// signature that can't express `'a` generically.
+///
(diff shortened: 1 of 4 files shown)


libdd-trace-utils/src/span/mod.rs
```diff
@@ -210,6 +210,53 @@ impl<'a> DeserializableTraceData for SliceData<'a> {
...
+/// `BytesData` the backing `Bytes` is heap-owned and never truly `'static` — the lifetime is
+/// erased via `transmute` and "laundered" by immediately deriving an owned `Self::Text` before
+/// it can escape into safe code. For `SliceData<'a>` this erasure is pure overhead: `Self::Bytes`
+/// is already `&'a [u8]`, the real target type, so the transmute exists only to satisfy a trait
+/// signature that can't express `'a` generically.
+///
+/// This trait carries the genuine lifetime `'x` explicitly, so `SliceData<'a>` can implement it
+/// with no unsafe code at all. It extends `DeserializableTraceData` rather than replacing it, and
+/// `Buffer<T>` gains an additive method (`as_mut_slice_lt`) in a new `impl` block — every existing
+/// call site on `DeserializableTraceData`/`Buffer::as_mut_slice` is untouched.
```
VianneyRuhlmann, line 225: I think we want to fully replace `DeserializableTraceData` with a `DeserializableTraceData<'a>` otherwise you keep implementing `DeserializableTraceData` for SliceData which means you have to keep the problematic transmute.

PR #2418, opened 2026-08-25
feat(data-pipeline)!: Obfuscate v04 spans in agentless context

paullegranddc:
# What does this PR do?

* Add a function to obfuscate v04 span
* Hook the obfuscation configuration in the trace exporter
* Force obfuscation if agentless is enabled
* Use the agent default values for the span obfuscation configuration


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -655,9 +660,22 @@ impl<
     /// Sends trace chunks to the Datadog agentless intake (`/v1/input`) as JSON.
     async fn send_agentless_traces_inner<T: TraceData>(
         &self,
-        traces: Vec<Vec<Span<T>>>,
+        mut traces: Vec<Vec<Span<T>>>,
         config: &AgentlessTraceConfig,
     ) -> Result<AgentResponse, TraceExporterError> {
+        // Obfuscate every span we are about to send to the intake
+        #[cfg(feature = "stats-obfuscation")]
```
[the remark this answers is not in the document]
paullegranddc: Fixed, added an explicit agentless feature

paullegranddc:
@codex review

libdd-trace-obfuscation/src/obfuscate.rs
```diff
@@ -225,6 +221,163 @@ fn obfuscate_attribute_array(v: &mut pb::AttributeArray, config: &ObfuscationCon
...
+
+    match as_str(&span.r#type) {
+        "web" | "http" if !span.meta.is_empty() => {
+            if let Some(url) = span.meta.get_mut(TAG_HTTPURL) {
+                apply(url, |u| {
+                    obfuscate_url(
+                        u,
+                        config.http.remove_query_string,
+                        config.http.remove_paths_with_digits,
+                    )
```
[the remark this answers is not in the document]
paullegranddc: To be fixed in a followup PR by using `defensive_dedup`

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -609,6 +623,117 @@ pub unsafe extern "C" fn ddog_trace_exporter_config_default_stats_cardinality_li
     CardinalityLimitConfig::default()
 }
 
+/// Enables agentless APM trace export and sets the intake URL and API key.
+///
+/// When set, APM trace spans are sent directly to the Datadog HTTP intake in JSON format
```
ichinaski, line 624: Why do we use the JSON endpoint when sending data to the Intake? Is this the recommended approach by the Intake team (instead of using the protobuf serialization)?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -804,15 +785,21 @@ impl<
...
         if let Some(ref config) = self.agentless_config {
-            // For agentless we want to tag top level spans, but not perform
-            // stats aggregation or span drops
-            if !self.client_computed_top_level {
-                for chunk in traces.iter_mut() {
-                    libdd_trace_utils::span::trace_utils::compute_top_level_span(chunk);
+            #[cfg(feature = "agentless")]
+            {
+                // For agentless we want to tag top level spans, but not perform
+                // stats aggregation or span drops
```
ichinaski: Why doesn't agentless aggregate stats and drop spans? Is this intended to be implemented further down the line?

libdd-data-pipeline/Cargo.toml
```diff
@@ -108,6 +112,8 @@ stats-obfuscation = [
     "libdd-trace-obfuscation",
     "libdd-trace-stats/stats-obfuscation"
 ]
+
+agentless = ["stats-obfuscation", "libdd-trace-obfuscation", "https"]
```
ichinaski: This dependency is odd, since agentless can never activate CSS (it never reads `/info` endpoint from an agent). Why do we have `stats-obfuscation` here?

libdd-trace-obfuscation/src/obfuscation_config.rs
```diff
@@ -1,35 +1,56 @@
...
 #[serde(default, deny_unknown_fields)]
 pub struct MemcachedConfig {
     pub enabled: bool,
     pub keep_command: bool,
 }
 
-#[derive(Debug, Default, Deserialize)]
+/// Mirrors the Datadog Agent defaults
+/// see `pkg/config/schema/yaml/apm_config.yaml`
+impl Default for MemcachedConfig {
```
ichinaski, line 22: Commenting here but applicable to all Configs below: It seems we are flipping the defaults to actually enable many of the obfuscators? It seems like a breaking change? Is this intended?

ichinaski (commented):
It's somewhat unclear how much on parity this approach is with the Trace Agent obfuscation config. We should document this somewhere, and also test the schemas are the same.

I also don't fully understand why we want to keep 2 separate implementations here, one for protobuf, the other for raw spans (using JSON endpoint). These 2 paths look also inconsistent at first glance (pb implementation being more complex and more tested).

libdd-data-pipeline/src/trace_exporter/mod.rs
paullegranddc: Yes, this is implemented in this PR (which does stats quantization and obfuscation unconditionaly since there is no /info) https://github.com/DataDog/libdatadog/pull/2309

libdd-data-pipeline-ffi/src/trace_exporter.rs
paullegranddc: This is the the intake used by among others RUM to send traces from browser. It is already accepting trafic that doesn't go through the agent, which is why it was picked.

libdd-trace-obfuscation/src/obfuscation_config.rs
paullegranddc: I have confirmed that the default impl was not used anywhere.
`ObfuscationConfig` is only passed to `obfuscate_span` which is used
* not used in libdatadog
* used in the serverless mini agent, but the config is read from env variables and does not depend on the new Default impl https://github.com/DataDog/serverless-components/blob/main/crates/datadog-trace-agent/src/config.rs#L216-L220

libdd-data-pipeline/Cargo.toml
paullegranddc: That's a leftover from a previous attempt at feature gating things. We want `stats-obfuscation` for this PR https://github.com/DataDog/libdatadog/pull/2309 , but I'll add it there

paullegranddc:
>I also don't fully understand why we want to keep 2 separate implementations here, one for protobuf, the other for raw spans (using JSON endpoint). These 2 paths look also inconsistent at first glance (pb implementation being more complex and more tested).

The protobuf span struct is obsolete and used only by serverless for their extension. I cannot remove the `obfuscate` method though as I don't want to break their extension completely.

If we ever want to send protobuf spans to the intake APM clients using libdatadog, we will do it by adding a protobuf serializer to the v1 span struct.
The paths are from what I can tell consistent and I have added tests that run the same cases against both functions.

libdd-data-pipeline/Cargo.toml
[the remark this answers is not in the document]
paullegranddc: fixed

bwoebi (approved):
I'm not too happy that it double obfuscates common operations, which have already been obfuscated once for stats computation.
But that's a future optimization, not a blocker. The implementation looks right to me: trivially obfuscating every span, and avoiding reallocating/copying for stuff which clearly doesn't need to be obfuscated.

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -609,6 +619,108 @@ pub unsafe extern "C" fn ddog_trace_exporter_config_default_stats_cardinality_li
...
+        } else {
+            gen_error!(ErrorCode::InvalidArgument)
+        },
+        gen_error!(ErrorCode::Panic)
+    )
+}
+
+/// Sets the span obfuscation configuration from a JSON string.
+///
+/// `json` is deserialized into `ObfuscationConfig` (the same struct the Datadog Agent uses),
```
ekump, line 675: The JSON here isn't actually the same as the agent, and we're setting `#[serde(deny_unknown_fields)]` so if someone tried to use the same JSON schema it'd break. 

For HTTP, the agent has [remove_path_digits](https://github.com/DataDog/datadog-agent/blob/6a44f563356c0339a62473f8d07b7ba9b6e5871b/pkg/obfuscate/obfuscate.go#L264), we have [remove_path_with_digits](https://github.com/DataDog/libdatadog/pull/2418/changes#diff-874c0ab809a94dffae95f7f5cc2dcf39b263473cefe8f9b4454bb4b571b4054cR125)

For ES, OpenSearch, Mongo, the agent has [keep_values](https://github.com/DataDog/datadog-agent/blob/6a44f563356c0339a62473f8d07b7ba9b6e5871b/pkg/obfuscate/obfuscate.go#L305), libdatadog has [keep_keys](https://github.com/DataDog/libdatadog/pull/2418/changes#diff-874c0ab809a94dffae95f7f5cc2dcf39b263473cefe8f9b4454bb4b571b4054cR59)

The JSON fields weren't added as part of this PR, I'm ok with a follow-up PR to fix this, but I do think we should be consistent with the agent on this.

libdd-trace-obfuscation/src/obfuscate.rs
```diff
@@ -225,6 +221,163 @@ fn obfuscate_attribute_array(v: &mut pb::AttributeArray, config: &ObfuscationCon
...
+/// immutable [`SpanText`] type. Each obfuscator has a cheap precheck, so unmodified fields don't
+/// allocate.
+// TODO(APMSP-2764): return parsing errors in a vec to log them ?
+pub fn obfuscate_v04_span<T: TraceData>(span: &mut v04::Span<T>, config: &ObfuscationConfig) {
+    for span_event in &mut span.span_events {
+        obfuscate_v04_span_event(span_event, config);
+    }
+
+    if config.credit_cards.enabled {
+        if let Some(credit_card) = span.meta.get_mut(TAG_CARD_NUMBER) {
```
ajgajg1134, line 249: This isn't how the trace-agent does CC obfuscation 🤔 , when enabled it looks at _all_ attributes (meta/metrics) except those in an explicit skip-list. https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/obfuscate.go#L152

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -736,6 +848,35 @@ pub unsafe extern "C" fn ddog_trace_exporter_new(
                 builder.set_output_to_log(config.log_max_line_size);
             }
 
+            // Agentless export path (enables span obfuscation). Mutually exclusive with an
+            // agent URL and an OTLP endpoint; the builder validates this and returns an
+            // error we map to ExporterError below.
+            if let Some(ref url) = config.agentless_endpoint {
```
ekump, line 854: [The builder will return an error](https://github.com/DataDog/libdatadog/blob/107977d1ed7b0adc1135829cc287d451ec732f59/libdd-data-pipeline/src/trace_exporter/builder.rs#L979-L986) if timeout is set, but the endpoint isn't. But with the FFI interface we won't return an error, we just fallback to the default timeout. We should probably match the Rust builder's behavior?

I don't consider this blocking as long as we follow up in a separate PR.

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -609,6 +623,117 @@ pub unsafe extern "C" fn ddog_trace_exporter_config_default_stats_cardinality_li
...
+    url: CharSlice,
+    api_key: CharSlice,
+) -> Option<Box<ExporterError>> {
+    catch_panic!(
+        if let Some(handle) = config {
+            handle.agentless_endpoint = match sanitize_string(url) {
+                Ok(s) => Some(s),
+                Err(e) => return Some(e),
+            };
+            handle.agentless_api_key = match sanitize_string(api_key) {
```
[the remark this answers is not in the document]
ekump: This looks like a legit concern. We should probably sanitize both inputs first before writing them to the handle.

Non-blocking as long as we follow up quickly.

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -736,6 +848,35 @@ pub unsafe extern "C" fn ddog_trace_exporter_new(
...
+            // error we map to ExporterError below.
+            if let Some(ref url) = config.agentless_endpoint {
+                let api_key = config.agentless_api_key.as_deref().unwrap_or("");
+                builder.set_agentless_endpoint(url, api_key);
+                if let Some(timeout_ms) = config.agentless_timeout_ms {
+                    builder.set_agentless_timeout(Duration::from_millis(timeout_ms));
+                }
+            }
+            if let Some(ref json) = config.obfuscation_config_json {
+                match serde_json::from_str::<
```
ekump, line 862: minor: Looks like we're parsing the JSON twice, the first time is [here](https://github.com/DataDog/libdatadog/pull/2418/changes#diff-6ef5922c131461351b8d0f7c176abc5cd803b83c2ed7bb14d1df0f3191b8e242R704).

libdd-data-pipeline-ffi/src/trace_exporter.rs
ekump, line 1: There should probably be tests for the new setters?

libdd-trace-obfuscation/src/obfuscation_config.rs
ekump, line 207: I don't think the defaults here all match the new default impl you defined? For example, luhn is true here, keep_command is true, but the default impl is the opposite.

libdd-trace-obfuscation/src/sql.rs
```diff
@@ -2208,6 +2208,18 @@ pub fn obfuscate_sql(s: &str, config: &SqlObfuscateConfig, dbms: DbmsKind) -> St
     }
 }
 
+/// Obfuscates a SQL string, returning `None` for empty input.
+///
+/// Non-empty SQL is always obfuscated and mirrored into `sql.query` by callers, so empty input is
+/// the only case skipped.
+#[must_use]
+pub fn obfuscate_sql_opt(s: &str, config: &SqlObfuscateConfig, dbms: DbmsKind) -> Option<String> {
```
ekump, line 2216: nit: Doesn't `_opt` imply None means "unmodified, don't allocate"? But for this we return None only when `s` is empty. Should we drop the `_opt` suffix?

ekump (approved):
A few things that need fixing, but not blocking. Please create Jira tickets for any follow up work.

libdd-trace-obfuscation/src/obfuscate.rs
paullegranddc: Adding a ticket to fix this

libdd-data-pipeline-ffi/src/trace_exporter.rs
paullegranddc: Adding a ticket to fix this

libdd-trace-obfuscation/src/obfuscation_config.rs
paullegranddc: This codepath is used in the serverless extension, so I kept it for backward compatibility. I'll rename it to `new_serverless` and mark it as deprecated so they can keep using it if they want but they'll get a warning when they bump

libdd-data-pipeline-ffi/src/trace_exporter.rs
paullegranddc: Will make a follow up in a separate PR

libdd-data-pipeline-ffi/src/trace_exporter.rs
paullegranddc: Will make a follow up in a separate PR

libdd-trace-obfuscation/src/sql.rs
paullegranddc: I wanted to make a non-allocating version of the sql obfuscator but doing a pre-check, or a "lazy until rewrite is needed" was a bit complex, and not really the point of this PR.
I'll add a followup ticket

morrisonlevi (approved):
I looked only at the `libdd-profiling-ffi/Cargo.toml` changes, which are fine.

PR #2420, opened 2026-08-26
ci(APMSP-3083): fix security issue by splitting the job

hoolioh:
# What does this PR do?

This PR hardens the workflow by improving its security model:

Use`pull_request_target` which runs the definition from the default branch instead, which also pins the OIDC `ref` and `job_workflow_ref` claims to `refs/heads/main`, so the trust policy can require a protected ref like the repository's other STS policies do.
The trade-off of `pull_request_target` is that the job context is privileged, so the work is split in two: 
1.`rustfmt` checks out the pull request and runs cargo fmt with no token and no `id-token` permission.
2. `rustfmt-commit` only applies the resulting patch and pushes it only using verified tools from the image, in this case 'git'. Note that `--force-with-lease` is used which makes the push a no-op if the branch moved on from the revision that was formatted. Not using third-party tools in this job is deliberate as it prevents using uncontrolled code with elevated permissions.

### DIFF
--- a/.github/chainguard/self.write.rustfmt.sts.yaml
+++ b/.github/chainguard/self.write.rustfmt.sts.yaml
@@ -1,9 +1,5 @@
 issuer: https://token.actions.githubusercontent.com
 
-# Pin policy to proper guards:
-# - Code went through review on main so 'ref' reflect that.
-# - The reference is protected.
-# - Assure the workflow provenance.
 subject_pattern: "repo:DataDog/libdatadog.*"
 
 claim_pattern:
--- a/.github/chainguard/self.write.rustfmt.sts.yaml
+++ b/.github/chainguard/self.write.rustfmt.sts.yaml
@@ -1,9 +1,15 @@
 issuer: https://token.actions.githubusercontent.com
 
-subject_pattern: "repo:DataDog/libdatadog:pull_request"
+# Pin policy to proper guards:
+# - Code went through review on main so 'ref' reflect that.
+# - The reference is protected.
+# - Assure the workflow provenance.
+subject_pattern: "repo:DataDog/libdatadog.*"
 
 claim_pattern:
-  job_workflow_ref: DataDog/libdatadog/\.github/workflows/rustfmt-auto\.yml@.+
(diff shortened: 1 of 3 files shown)


.github/chainguard/self.write.rustfmt.sts.yaml
```diff
@@ -1,9 +1,15 @@
 issuer: https://token.actions.githubusercontent.com
 
-subject_pattern: "repo:DataDog/libdatadog:pull_request"
+# Pin policy to proper guards:
+# - Code went through review on main so 'ref' reflect that.
+# - The reference is protected.
+# - Assure the workflow provenance.
```
ekump: Should this comment be here? It reads more like AI explaining what it did.

.github/workflows/rustfmt-auto.yml
```diff
@@ -1,42 +1,145 @@
...
+# Security model:
+# Use`pull_request_target` which runs the definition from the default branch instead, which 
+# also pins the OIDC `ref` and `job_workflow_ref` claims to `refs/heads/main`, so
+# the trust policy can require a protected ref like the repository's other STS policies do.
+# The trade-off of `pull_request_target` is that the job context is privileged, so the
+# work is split in two: 
+# 1.`rustfmt` checks out the pull request and runs cargo fmt with no token and no `id-token`
+# permission.
+# 2. `rustfmt-commit` only applies the resulting patch and pushes it only using verified tools
+# from the image, in this case 'git'. Not that --force-with-lease is used which makes the push
```
ekump:
```suggestion
# from the image, in this case 'git'. Note that --force-with-lease is used which makes the push
```

.github/workflows/rustfmt-auto.yml
```diff
@@ -1,42 +1,145 @@
 name: Rustfmt Auto
+
+# Security model:
```
ekump: Can we make this comment block more concise? It reads like AI explaining what it did, rather than something useful for future readers.

ekump (approved):
Code is LGTM, but could you do an editorial pass on the comments before merging?

PR #2421, opened 2026-08-26
feat(common)!: add HTTPS_PROXY support for hyper_backend

bwoebi:
We make sure to fetch the env vars only once at init, at basically the same time than the TLS envs are read too.

This allows users to redirect requests going to the datadog backend through egress proxies, rather than having to hack around with DNS and MITM TLS certificates.

### DIFF
--- a/libdd-common/src/connector/proxy.rs
+++ b/libdd-common/src/connector/proxy.rs
@@ -15,22 +15,30 @@ use super::{https, Connector};
 // Read getenv only once at init to avoid concurrency issues with the environment.
 static PROXY_MATCHER: LazyLock<Matcher> = LazyLock::new(Matcher::from_env);
 
-/// Wraps a direct `Connector` to additionally honor the `HTTPS_PROXY`/`https_proxy`
+/// Wraps a direct `Connector` to additionally honor the `HTTP(S)_PROXY`/`http(s)_proxy`
 /// and `NO_PROXY`/`no_proxy` environment variables.
 #[derive(Clone)]
 pub(super) struct HttpsProxyConnector {
     matcher: &'static Matcher,
     tls_config: Option<rustls::ClientConfig>,
-    http: HttpConnector,
+    proxy_dialer: Option<hyper_rustls::HttpsConnector<HttpConnector>>,
     direct: Connector,
 }
 
 impl HttpsProxyConnector {
     pub(super) fn new(direct: Connector) -> Self {
+        let tls_config = https::build_tls_config().ok();
+        let proxy_dialer = tls_config.clone().map(|tls_config| {
+            hyper_rustls::HttpsConnectorBuilder::new()
+                .with_tls_config(tls_config)
+                .https_or_http()
+                .enable_http1()
+                .build()
+        });
         Self {
             matcher: &PROXY_MATCHER,
-            tls_config: https::build_tls_config().ok(),
-            http: HttpConnector::new(),
+            tls_config,
+            proxy_dialer,
             direct,
         }
     }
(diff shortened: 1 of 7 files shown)


libdd-http-client/Cargo.toml
```diff
@@ -19,6 +19,8 @@ default = ["https", "reqwest-backend"]
 https = ["dep:reqwest", "reqwest?/rustls-no-provider", "libdd-common?/https"]
 reqwest-backend = ["dep:reqwest", "reqwest?/hickory-dns", "reqwest?/multipart"]
 hyper-backend = ["dep:libdd-common", "dep:hyper", "dep:hyper-util", "dep:http-body-util"]
+# Honor HTTPS_PROXY/NO_PROXY environment variables in the hyper backend.
+hyper-proxy = ["hyper-backend", "libdd-common?/hyper-proxy"]
```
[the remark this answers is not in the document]
bwoebi: Added an Err.

paullegranddc:
Why not `hyper_http_proxy` which is already used in the serverless bottlecap project if we need to support this? 

https://github.com/DataDog/datadog-lambda-extension/blob/fec395f5b7546d82ebde6aa149e6362bd050829b/bottlecap/src/traces/http_client.rs#L28-L29

libdd-http-client/tests/hyper_proxy_test.rs
```diff
@@ -0,0 +1,99 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! The proxy environment variables are read only once, ever, on the first
+//! `HttpsProxyConnector` constructed in the process (see `PROXY_MATCHER` in
+//! `libdd_common::connector::proxy`) to avoid racing a concurrent `setenv`.
+//! This file must therefore stay the only test that touches `HTTPS_PROXY` in
+//! this binary, and must set it before constructing its first `HttpClient`.
```
paullegranddc: Good to mention works because each top-level file in `/tests` is compiled as a separate binary (https://doc.rust-lang.org/cargo/reference/cargo-targets.html#integration-tests)

So it's not that this file should be the only one touching HTTPS_PROXY, but that there can be only one test in this file.

bwoebi:
@paullegranddc hyper_http_proxy doesn't really help us get rid of of most of the plumping done to abstract the connection stream away (proxy vs non-proxy stream) as it is supposed to be an optional dependency.

It would only replace the `HttpsProxyConnector` impl in this PR (which is pretty straighforward I'd say?) basically - and instead be wired in there. 

Everything we could get from the crate is minimal, given that it would have to be integrated into our own abstractions including the tls config builder. I don't quite see that being worth the dependency.

libdd-common/src/connector/proxy.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub(super) fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), ConnStreamError>> {
+        tower_service::Service::poll_ready(&mut self.direct, cx)
+    }
+
+    pub(super) fn build_conn_stream(
+        &mut self,
+        uri: hyper::Uri,
+        require_tls: bool,
+    ) -> BoxFuture<'static, Result<ConnStream, ConnStreamError>> {
+        if require_tls {
```
paullegranddc: If the url is not https we do not respect the `HTTP(S)_PROXY` which is weird 🤔

bwoebi: Yeah, makes sense. I was targeting https traffic to the intake specifically, but really, this should be generic. Will change.

bwoebi: Changed.

libdd-http-client/tests/hyper_proxy_test.rs
bwoebi: Adjusted :-)

paullegranddc (approved):

PR #2428, opened 2026-08-27
fix(crashtracking): filter out frames above faulting frame

gyuheon0h:
# What does this PR do?

When `collect_all_threads` is enabled, the receiver ptraces the crashing thread while it is parked inside our own signal handler waiting for the receiver to finish. Its entry in `error.threads` therefore starts inside libdatadog rather than at the faulting instruction:

from running bin_test crash output on a contained crash:
```
__GI___poll 
ProcessHandle::finish 
handle_posix_sigaction 
__restore_rt cause_segfault <- the actual crash site, where error.stack starts
```

`error.stack` never shows these because libunwind seeds it straight from the kernel-saved registers. This makes the crashed thread's stack match.

Those same registers are already on the wire in the ucontext block, so the receiver drops every frame above the one whose `ip`/`sp` match `rip`/`rsp` (`pc`/`sp` on aarch64). This makes sure the crashing thread stack in `error.threads` outputs a clear stack that correctly represents the frames leading up to the crash, not including crashtracker and signal handler frames.

# Testing

-  unit tests in `receive_report.rs`: trimming, alternate-signal-stack layout, zero-padded vs unpadded register formatting, per-arch register names, etc
- `test_crash_tracking_multi_thread_collection` now asserts the crashed thread starts at the same `ip` as `error.stack` and contains no `libdd_crashtracker::collector` frames.
- I also verified the new e2e assertion fails without the fix (crashed thread starts at `__GI___poll`).

# Motivation

Noticed crash reports with crashing thread in `error.threads` containing unwanted frames.

# Additional Notes

### DIFF
--- a/bin_tests/tests/crashtracker_bin_test.rs
+++ b/bin_tests/tests/crashtracker_bin_test.rs
@@ -444,6 +444,46 @@ fn test_crash_tracking_multi_thread_collection() {
             );
         }
 
+        // The receiver unwinds the crashing thread while it is parked inside our own
+        // signal handler, so its raw stack starts inside libdatadog. Those frames are
+        // trimmed back to the frame the kernel-saved registers point at, which is where
+        // error.stack starts too.
+        //
+        // The remote unwind has to cross the signal trampoline to reach that frame, and
+        // in some cases the trampoline carries no DWARF unwind info, so it stops short. There
+        // is nothing to trim in that case and the receiver leaves the stack alone, so
+        // only assert the invariant once the crash site is actually present.
+        let crashed_frames = crashed_threads[0]["stack"]["frames"]
+            .as_array()
+            .expect("crashed thread stack.frames should be an array");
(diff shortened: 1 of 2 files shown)


gyuheon0h:
* **#2428** <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2428?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/libdatadog/2428?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

gyuheon0h:
Maybe a better solution here is to seed libunwind in the receiver with the crashing thread registers directly 🤷

KowalskiThomas (approved):
Little knowledge on that but based on the PR description it makes sense to me.

gleocadie (approved):
LGTM (makes total sense)

PR #2430, opened 2026-08-28
feat(ffe)!: send flagevaluation producer identity

vjfridge:
# What does this PR do?

Adds explicit producer identity metadata to libdatadog flag-evaluation EVP requests.

- requires and validates non-empty `origin` and `origin_version` values in `FlagEvaluationEvpSendConfig`
- sends `DD-EVP-ORIGIN` and `DD-EVP-ORIGIN-VERSION` on every request
- attributes sidecar-forwarded evaluations to `dd-trace-php` using the originating session's tracer version
- keeps the sidecar `User-Agent` independent from the originating SDK identity
- includes the send configuration in the coalescer destination key so different producer identities cannot share a request
- adds sender and sidecar tests for exact headers, invalid configuration, identity-specific batching, and originating tracer attribution

# Motivation

Implements [FFL-3019](https://datadoghq.atlassian.net/browse/FFL-3019). The Agent now forwards EVP origin metadata, so libdatadog needs to provide the originating SDK identity for flag-evaluation intake attribution.

# Additional Notes

This follows the same transport contract as [dd-trace-rb#6254](https://github.com/DataDog/dd-trace-rb/pull/6254) and [dd-trace-go#5283](https://github.com/DataDog/dd-trace-go/pull/5283): request-level SDK identity headers use the producer's stable name and release version while preserving existing routing headers and event payloads. The shared libdatadog sender accepts explicit producer identity; the current sidecar call site supplies `dd-trace-php` and the tracer version already stored in its session configuration.

The sidecar is a transport rather than the source SDK, so it remains represented only by `User-Agent`. This does not add an event-level `source` field or change payload limits, endpoint routing, timeout handling, metrics, or delivery behavior.

# How to test the change?

```bash
cargo +nightly-2026-07-26 fmt --all -- --check
cargo check -p libdd-ffe --features flagevaluation-evp
cargo nextest run -p libdd-ffe --features flagevaluation-evp
cargo clippy -p libdd-ffe --all-targets --all-features -- -D warnings
cargo test -p libdd-ffe --doc --features flagevaluation-evp
cargo check -p datadog-sidecar
cargo nextest run -p datadog-sidecar
cargo clippy -p datadog-sidecar --all-targets -- -D warnings
cargo test -p datadog-sidecar --doc
```

[FFL-3019]: https://datadoghq.atlassian.net/browse/FFL-3019

### DIFF
--- a/datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
+++ b/datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
@@ -22,6 +22,8 @@ use std::time::Duration;
 use tokio::sync::Mutex as AsyncMutex;
 
 const USER_AGENT: &str = concat!("ddtrace-sidecar/", env!("CARGO_PKG_VERSION"));
+const EVP_ORIGIN: &str = "ddtrace-sidecar";
+const EVP_ORIGIN_VERSION: &str = env!("CARGO_PKG_VERSION");
 const COALESCE_DELAY: Duration = Duration::from_millis(250);
 
 pub(crate) const FLAG_EVALUATION_DROPPED_EVALUATIONS_METRIC: &str =
@@ -38,13 +40,19 @@ pub(crate) const FLAG_EVALUATION_REASON_PAYLOAD_LIMIT: &str = "payload_limit";
 struct DestinationKey {
     endpoint: Endpoint,
     context: FfeTelemetryContext,
+    send_config: FlagEvaluationEvpSendConfig,
 }
 
 impl DestinationKey {
-    fn new(endpoint: Endpoint, context: &FfeTelemetryContext) -> Self {
+    fn new(
+        endpoint: Endpoint,
+        context: &FfeTelemetryContext,
+        send_config: FlagEvaluationEvpSendConfig,
+    ) -> Self {
         Self {
             endpoint,
             context: context.clone(),
+            send_config,
         }
     }
 }
@@ -62,7 +70,27 @@ impl FlagEvaluationCoalescer {
         endpoint: Endpoint,
         batch: FfeFlagEvaluationBatch,
     ) {
-        let destination_key = DestinationKey::new(endpoint, &batch.context);
+        let send_config =
+            match FlagEvaluationEvpSendConfig::new(USER_AGENT, EVP_ORIGIN, EVP_ORIGIN_VERSION) {
+                Ok(config) => config,
+                Err(error) => {
+                    tracing::error!(
+                        "invalid sidecar flagevaluation EVP send configuration: {error}"
+                    );
+                    return;
+                }
+            };
+        self.enqueue_with_config(client, endpoint, batch, send_config);
+    }
+
+    fn enqueue_with_config(
+        &self,
+        client: NativeCapabilities,
+        endpoint: Endpoint,
+        batch: FfeFlagEvaluationBatch,
+        send_config: FlagEvaluationEvpSendConfig,
+    ) {
+        let destination_key = DestinationKey::new(endpoint, &batch.context, send_config);
         if self.inner.enqueue(destination_key, batch) {
             let coalescer = self.clone();
             tokio::spawn(async move {
@@ -82,8 +110,14 @@ impl FlagEvaluationCoalescer {
             let client = client.clone();
             let coalescer = self.inner.clone();
             async move {
-                send_batch_with_writer_stats(&client, &destination.endpoint, batch, &coalescer)
-                    .await
+                send_batch_with_writer_stats(
+                    &client,
+                    &destination.endpoint,
+                    batch,
+                    &destination.send_config,
+                    &coalescer,
+                )
+                .await
             }
         }))
         .await;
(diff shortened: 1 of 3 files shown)


vjfridge:
@codex review

datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
```diff
@@ -22,6 +22,8 @@ use std::time::Duration;
 use tokio::sync::Mutex as AsyncMutex;
 
 const USER_AGENT: &str = concat!("ddtrace-sidecar/", env!("CARGO_PKG_VERSION"));
+const EVP_ORIGIN: &str = "ddtrace-sidecar";
+const EVP_ORIGIN_VERSION: &str = env!("CARGO_PKG_VERSION");
```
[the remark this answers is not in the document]
vjfridge: 🤖 Done — the EVP origin-version header now uses `crate::sidecar_version!()`, so release builds honor `SIDECAR_VERSION` instead of reporting the package's fixed `0.0.1`. The existing sidecar request test now asserts against the configured sidecar version; it was verified with `SIDECAR_VERSION=42.0.0`.

Commit: https://github.com/DataDog/libdatadog/commit/b6e6abd717744cc058ec1bacaac8c9929b9688c5

vjfridge:
@codex review

libdd-ffe/src/telemetry/flagevaluation/sender.rs
```diff
@@ -2,18 +2,22 @@
...
 use std::time::Duration;
+use thiserror::Error;
 
 /// EVP proxy path for FFE flag evaluation intake.
 pub const EVP_FLAGEVALUATION_PATH: &str = "/evp_proxy/v2/api/v2/flagevaluation";
 /// EVP subdomain header name.
 pub const EVP_SUBDOMAIN_HEADER: &str = "X-Datadog-EVP-Subdomain";
 /// EVP subdomain that routes requests to event-platform intake.
 pub const EVP_SUBDOMAIN_VALUE: &str = "event-platform-intake";
+const EVP_ORIGIN_HEADER: &str = "DD-EVP-ORIGIN";
```
yannham, line 18: Nit, but curious why the previous `const` are public and not those ones? (I'm not saying we should make those `pub` if not needed, just wondering about the inconsistency - if anything, we should rather make the other non-pub if it's not needed)

yannham (approved):
Can't say about the FFE logic, but otherwise LGTM 👍

libdd-ffe/src/telemetry/flagevaluation/sender.rs
vjfridge: Good question! The existing public constants predate this PR and are already re-exported as part of the `libdd-ffe` API. Some also have consumers outside this module. 

The new origin header names are intentionally private because callers configure identity through `with_origin` / `with_origin_version` rather than constructing the request headers themselves. 

Making the older constants private would be a separate breaking API cleanup, so I’d prefer not to touch it in this PR, but something for our team to think about in the future, thanks!

vjfridge:
@codex review

datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
```diff
@@ -34,17 +35,33 @@ pub(crate) const FLAG_EVALUATION_REASON_DEGRADED_CAP: &str = "degraded_cap";
 pub(crate) const FLAG_EVALUATION_REASON_CARDINALITY_CAP: &str = "cardinality_cap";
 pub(crate) const FLAG_EVALUATION_REASON_PAYLOAD_LIMIT: &str = "payload_limit";
 
+pub(crate) fn evp_origin_from_language(language: &str) -> Cow<'_, str> {
+    match language {
+        "ruby" => Cow::Borrowed("dd-trace-rb"),
+        "python" => Cow::Borrowed("dd-trace-py"),
+        "javascript" | "nodejs" => Cow::Borrowed("dd-trace-js"),
+        "rust" => Cow::Borrowed("dd-trace-rs"),
+        language => Cow::Owned(format!("dd-trace-{language}")),
```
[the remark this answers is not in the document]
vjfridge: 🤖 Addressed in https://github.com/DataDog/libdatadog/commit/3bd77ef754c9ab2aca7bae50b20762f401066e4e. `evp_origin_from_language` now returns `None` for empty or whitespace-only language, and the coalescer omits `DD-EVP-ORIGIN` in that case while keeping `DD-EVP-ORIGIN-VERSION` independent. Added a sidecar request regression test that verifies the origin header is absent for whitespace-only language.

datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
```diff
@@ -34,17 +35,37 @@ pub(crate) const FLAG_EVALUATION_REASON_DEGRADED_CAP: &str = "degraded_cap";
...
+    })
+}
+
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 struct DestinationKey {
     endpoint: Endpoint,
     context: FfeTelemetryContext,
+    send_config: FlagEvaluationEvpSendConfig,
```
btthomas, line 56: You might be able to cache `FlagEvaluationEvpSendConfig` with `Arc` since it shouldn't change between batches. Then below you can use `Arc::clone()`. If performance matters, this might save you from reallocating the configuration (which might not change).

btthomas (approved):

datadog-sidecar/src/service/ffe_flagevaluation_flusher.rs
vjfridge: Thanks! After digging into this a little more, I decided to only add the tiny change to reduce the intermediate string allocations (see 6f0955187).

This PR does move config construction from 'once per coalesced outbound batch' to 'once per incoming SDK batch', so there is some additional fixed per-batch work. It isn’t per evaluation, though, and should be small relative to aggregation, serialization, and HTTP delivery.

I don't think I'll add the session-scoped Arc cache in this PR, since that would introduce cache lifetime and invalidation state for what is likely a small optimization. If this path becomes noticeably slow, caching would be a reasonable follow-up.

PR #2440, opened 2026-09-01
feat: do not entirely disable connection pooling for periodic connections

yannham:
# What does this PR do?

This PR slightly changes the HTTP behavior for periodic connections (most flushes to the agent/agentless intake). Instead of disabling connection pooling to avoid data races entirely, it restricts the lifetime of pooled connections.

This PR aims to be a backward-compatible hotfix.  A subsequent PR is coming with renaming (since `no_connection_pooling` isn't really true anymore) and mirroring the change in libdd-http-client and libdd-agent-client as well.

# Motivation

Mitigates [APMS-20441](https://datadoghq.atlassian.net/browse/APMS-20441?atlOrigin=eyJpIjoiZjcyY2U2Zjk3Njg2NGNlZGE5YjNhNTllNzU5Nzk1MDgiLCJwIjoiaiJ9) / https://github.com/DataDog/dd-trace-py/issues/19915: some telemetry events can generate several requests for short-lived scripts. In the agentless case, this means opening several new HTTPS connection in a row, which is slow.

# Additional Notes

Applying the setting to all periodic connections sounds reasonable, as multiple requests could reasonably be issued for other things than telemetry. It should impact single-request workflows.

# How to test the change?

This was tested with the repro in 20441, reducing the shutdown delay to 0.750ms locally, which indicates there's indeed a single HTTPS connection (vs double before the change).

[APMS-20441]: https://datadoghq.atlassian.net/browse/APMS-20441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libdd-capabilities-impl/src/http.rs
+++ b/libdd-capabilities-impl/src/http.rs
@@ -45,12 +45,9 @@ mod native {
     }
 
     impl NativeHttpClient {
-        /// Like [`HttpClientCapability::new_client`], but disables connection pooling.
-        ///
-        /// Intended for clients that issue requests on a fixed interval (e.g. remote
-        /// config polling): the agent's low keep-alive setting can close an idle
-        /// connection between polls, which turns a pooled/reused connection into
-        /// intermittent request failures.
+        /// Like [`HttpClientCapability::new_client`], but sets a small lifetime on pooled
+        /// connections. See [`HttpClientCapability::new_without_connection_pooling`] for the
+        /// rationale.
         pub fn new_without_connection_pooling() -> Self {
             Self {
                 client: Arc::new(OnceLock::new()),
--- a/libdd-common/src/http_common.rs
+++ b/libdd-common/src/http_common.rs
@@ -3,6 +3,7 @@
 
 use core::convert::Infallible;
 use core::fmt;
+use core::time::Duration;
 
 use thiserror::Error;
 
@@ -80,6 +81,7 @@ mod native {
     use crate::connector::Connector;
     use http_body_util::BodyExt;
     use hyper::body::Incoming;
+    use hyper_util::rt::TokioTimer;
     use pin_project::pin_project;
 
     impl From<hyper::Error> for ClientError {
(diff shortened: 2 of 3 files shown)


libdd-common/src/http_common.rs
```diff
@@ -115,15 +117,25 @@ mod native {
...
+    /// This client pools connections with a small timeout (smaller than any potential keep-alive on
+    /// the receiver side), because otherwise we would get a pipe closed every second connection
+    /// because of the keep alive in the agent or the backend.
     ///
-    /// This is on general not a problem if we use the client once every tens of seconds.
+    /// This is in general not a problem if we use the client once every tens of seconds.
     pub fn new_client_periodic() -> GenericHttpClient<Connector> {
         hyper_util::client::legacy::Client::builder(hyper_util::rt::TokioExecutor::default())
-            .pool_max_idle_per_host(0)
+            .pool_timer(TokioTimer::new())
```
ekump, line 140: minor: Do you think it's also worth setting `.pool_max_idle_per_host()` to something reasonable?

libdd-capabilities-impl/src/http.rs
```diff
@@ -45,12 +45,9 @@ mod native {
...
-        /// Like [`HttpClientCapability::new_client`], but disables connection pooling.
-        ///
-        /// Intended for clients that issue requests on a fixed interval (e.g. remote
-        /// config polling): the agent's low keep-alive setting can close an idle
-        /// connection between polls, which turns a pooled/reused connection into
-        /// intermittent request failures.
+        /// Like [`HttpClientCapability::new_client`], but sets a small lifetime on pooled
+        /// connections. See [`HttpClientCapability::new_without_connection_pooling`] for the
+        /// rationale.
         pub fn new_without_connection_pooling() -> Self {
```
ekump, line 51: minor: This function name isn't quite accurate anymore

ekump (approved):

libdd-capabilities-impl/src/http.rs
yannham: Yup

> This PR aims to be a backward-compatible hotfix. A subsequent PR is coming with renaming (since no_connection_pooling isn't really true anymore) and mirroring the change in libdd-http-client and libdd-agent-client as well.

libdd-common/src/http_common.rs
yannham: Hmm, I would say there's a low risk of accumulating many idle connections given the timeout is quite short. On the other hand it doesn't hurt to cap resources, so why not

yannham: I've also applied the bound to the normal `new_client()` constructor, since bounding the resources for the connection pool is orthogonal to the keep-alive race condition thing.

yannham: Implemented in [1444009](https://github.com/DataDog/libdatadog/pull/2440/commits/14440099139775854f659c6f1f213075b94452d3)

libdd-capabilities-impl/src/http.rs
ekump: Damn, you caught me not reading the description throughly

PR #2441, opened 2026-09-01
fix(obfuscation): prevent crash and cap recursion in SQL/HTTP obfuscation

ajgajg1134:
## Summary
- Cap recursion depth for nested dollar-quoted SQL strings so pathologically nested input can no longer exhaust the stack and abort the process.
- Fix a fallback path in URL obfuscation that could return an unparseable URL with userinfo still attached instead of stripping it.

Fixes APMSP-3086 and APMSP-3085

## Test plan
- [x] `cargo nextest run -p libdd-trace-obfuscation` (343 tests, incl. new regression tests for both fixes)
- [x] `cargo fmt --check` / `cargo clippy -D warnings` clean

### DIFF
--- a/libdd-trace-obfuscation/src/http.rs
+++ b/libdd-trace-obfuscation/src/http.rs
@@ -71,6 +71,35 @@ fn encode_char(out: &mut String, c: char) {
     }
 }
 
+/// Best-effort userinfo stripping for URLs that fail strict RFC 3986 parsing (or contain control
+/// chars) and would otherwise be returned unchanged: credentials must never survive obfuscation
+/// even when we can't fully parse the rest of the URL.
+///
+/// `prefix_end` bounds the scan to the path portion (before query/fragment) so a `@` inside the
+/// query or fragment is never mistaken for userinfo.
+///
+/// This is a heuristic, not a parser: it may occasionally strip more than necessary on
+/// non-URL-like malformed input, but it never leaves real credentials in the output.
+fn strip_userinfo_best_effort(url: &str, prefix_end: usize) -> String {
+    let prefix = &url[..prefix_end];
+    let Some(marker) = prefix.find("//") else {
+        return url.to_string();
+    };
+    let authority_start = marker + 2;
+    let authority_end = prefix[authority_start..]
+        .find('/')
+        .map_or(prefix.len(), |i| authority_start + i);
+    let authority = &prefix[authority_start..authority_end];
+    let Some(at_pos) = authority.rfind('@') else {
+        return url.to_string();
+    };
+    let mut out = String::with_capacity(url.len());
+    out.push_str(&url[..authority_start]);
+    out.push_str(&authority[at_pos + 1..]);
+    out.push_str(&url[authority_end..]);
+    out
+}
+
 fn redact_path_digits(path: &str) -> String {
     path.split('/')
         .map(|seg| {
@@ -187,7 +216,7 @@ pub fn obfuscate_url_string(
         return if remove_query_string || remove_path_digits {
             "?".to_string()
         } else {
-            url.to_string()
+            strip_userinfo_best_effort(url, path_query_end)
         };
     }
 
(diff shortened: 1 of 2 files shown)


libdd-trace-obfuscation/src/sql.rs
```diff
@@ -1056,19 +1062,34 @@ impl<'a> Tokenizer<'a> {
...
-                                    // Obfuscate the content inside dollar quotes
+                                    // Obfuscate the content inside dollar quotes.
+                                    // Past MAX_DOLLAR_QUOTE_DEPTH, use a placeholder directly
+                                    // rather than recursing, so the nested literal can't leak.
                                     let tag_str = &self.s[start..inner_start];
                                     let inner = &self.s[inner_start..inner_end];
                                     let close_tag = &self.s[inner_end..outer_end];
-                                    let obfuscated_inner =
-                                        obfuscate_sql(inner, self.config, self.dbms);
+                                    if self.depth + 1 >= MAX_DOLLAR_QUOTE_DEPTH {
```
yannham: Very nit, but since the constant is arbitrary to begin with, the `+ 1` looks a bit superfluous. We could decrement `MAX_DOLLAR_QUOTE_DEPTH` if needed.

```suggestion
                                    if self.depth >= MAX_DOLLAR_QUOTE_DEPTH {
```

libdd-trace-obfuscation/src/sql.rs
```diff
@@ -206,13 +206,18 @@ fn find_dollar_quote_end(bytes: &[u8], start: usize) -> Option<(usize, usize, us
...
+
 struct Tokenizer<'a> {
     s: &'a str,
     bytes: &'a [u8],
     pos: usize,
     result: String,
     dbms: DbmsKind,
     config: &'a SqlObfuscateConfig,
+    // Nesting depth of dollar-quote recursion; see MAX_DOLLAR_QUOTE_DEPTH.
+    depth: usize,
```
yannham: Nit: this field could have a slightly more telling name, like `dollar_recursion_depth`. Similarly for the corresponding function arguments.

libdd-trace-obfuscation/src/http.rs
```diff
@@ -187,7 +216,7 @@ pub fn obfuscate_url_string(
         return if remove_query_string || remove_path_digits {
```
yannham, line 216: unrelated, but this `return` is unnecessary.

yannham (approved):

libdd-trace-obfuscation/src/http.rs
ajgajg1134: hmm am I mis-reading those code? Doesn't the function continue below? So removing this return would mean the function no long short circuits?

libdd-trace-obfuscation/src/http.rs
```diff
@@ -187,7 +216,7 @@ pub fn obfuscate_url_string(
...
             "?".to_string()
         } else {
-            url.to_string()
+            strip_userinfo_best_effort(url, path_query_end)
```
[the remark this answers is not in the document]
ekump: @ajgajg1134 I think this is a valid concern?

libdd-trace-obfuscation/src/sql.rs
ajgajg1134: I think we actually need this since otherwise we create a gap where we're off by 1 (if you try removing this you can see this happens with failing the test even if you lower the constant by 1).

libdd-trace-obfuscation/src/http.rs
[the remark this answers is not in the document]
ajgajg1134: yeah I agree, pushing a fix now

libdd-trace-obfuscation/src/http.rs
yannham: My bad, the diff view of Github made me think this was the end of the function. You're totally right, sorry for the noise.

Eldolfin (approved):

PR #2442, opened 2026-09-01
feat(profiling)!: add gc-time sample type (PROF-15894)

vlad-scherbich:
# What does this PR do?

Adds first-class `GcTime` / `GcSamples` to `SampleType` (Rust + CXX FFI):

- `GcTime` → `("gc-time", "nanoseconds")`
- `GcSamples` → `("gc-samples", "count")` (same pairing as wall/cpu)

Encode and decode both ways. House round-trip test covers the new variants. C headers are generated from `cxx.rs` in CI.

If this merges and nothing else lands: the type exists in libdatadog. No tracer emits it. No intake/CompView listing.

| Layer | Meaning | PR |
| --- | --- | --- |
| Compiled into artifact | `SampleType` + FFI | this PR |
| Armed at runtime | tracer `push_*` | dd-trace-py after a libdatadog bump |
| Observable in product | CompView / intake allow-list | follow-up |

# Motivation

https://github.com/DataDog/dd-trace-py/pull/19940 piggybacks GC pauses on wall-time (default-off). It stays that way until this type exists here and the tracer can `push_*` it.

# Additional Notes

Timeline (STW) and intake/CompView `availableProfileTypes` are follow-up. Not eBPF, not DogStatsD, not T #19190 CPU leaf frames. Frames stay `gc.collect[gen=N]`.

BREAKING CHANGE: additive variants on a public exhaustive `SampleType` enum. Existing discriminants unchanged (appended after `Custom5`).

# How to test the change?

`cargo test -p libdd-profiling --lib api::sample_type`


KowalskiThomas (approved):

ivoanjo:
So I'm curious about this one -- multiple languages using pprof are already expressing GC time in the timeline profile type, so what does that mean for those languages?

Should we move to this new profile type and stop using timeline for that? Keep both? Something else?

vlad-scherbich:
> So I'm curious about this one -- multiple languages using pprof are already expressing GC time in the timeline profile type, so what does that mean for those languages?
> 
> Should we move to this new profile type and stop using timeline for that? Keep both? Something else?

@ivoanjo 

One of the reasons for this PR, besides making GC events work for Python, is to codify shared GC types (currently main has no such types). As a result, runtimes are using runtime-specific mechanisms (.NET) or generic types such as timeline.

As such, I would propose making `gc-time` / `gc-samples` a shared interface for (all?) consumers:

gc-time: elapsed GC pause duration in nanoseconds
gc-samples: GC event count

Existing runtime behavior would remain unchanged and unaffected. Once the shared types are supported in BE and FE, teams can migrate to the new types on their own timeline.

This PR unblocks Python use of GC events, as I assumed this is the right direction to take during implementation.

Thoughts on this?

ivoanjo:
My big worry here is trying to avoid the classic xkcd:

> 

In particular, I think it would be nice to try producing a migration guide or even ask our AI friends to try to make PRs for everyone to use the new format and see if it's good for all runtimes we need to support.

Otherwise, if we validate the new format with only python, I fear that we might not ensure that we can represent everything that other runtimes already need, and so rather than producing the nice effect "we have one format for everyone to use" we'll just have created yet-another-format which only one runtime uses.

vlad-scherbich:
> My big worry here is trying to avoid the classic xkcd:
> 
> > 
> 
> In particular, I think it would be nice to try producing a migration guide or even ask our AI friends to try to make PRs for everyone to use the new format and see if it's good for all runtimes we need to support.
> 
> Otherwise, if we validate the new format with only python, I fear that we might not ensure that we can represent everything that other runtimes already need, and so rather than producing the nice effect "we have one format for everyone to use" we'll just have created yet-another-format which only one runtime uses.

@ivoanjo I agree with your concern in principle. My only reservation is making Python GC profiler depend on buy-ins from all runtimes. I’d suggest making cross-runtime migration as an explicit follow-up, with a concrete migration guide and a prototype for e.g .NET and / or Java. That lets Python proceed, while also giving us a fast way to validate the shared format. I also don’t want this to become the `15th standard`, so I’d be happy to help drive that effort rather than just "let it happen organically".

ivoanjo:
Fair enough -- my concern was indeed having that drive to finish the migration beyond Python, so if you're up for making it happen, that clears it, thank you!

(For what it's worth Java doesn't use pprof so I think you can ignore that one; and I guess as soon as the backend lands support for this format we should be good to go to start moving the other profilers!)

vlad-scherbich:
> Fair enough -- my concern was indeed having that drive to finish the migration beyond Python, so if you're up for making it happen, that clears it, thank you!
> 
> (For what it's worth Java doesn't use pprof so I think you can ignore that one; and I guess as soon as the backend lands support for this format we should be good to go to start moving the other profilers!)

@ivoanjo https://datadoghq.atlassian.net/browse/PROF-15949

PR #2443, opened 2026-09-01
feat(otel-thread-ctx): add update-and-attach operation

y9v:
# What does this PR do?

It adds `ddog_otel_thread_ctx_update_and_attach` operation for updating a named thread context and attaching it.

# Motivation

During [a review](https://github.com/DataDog/dd-trace-rb/pull/6102#discussion_r3872434524) of a Ruby implementation for the OTel Thread Context we realized that it would be nice to have one operation that updates the context and attaches it, to avoid attaching a context record with potentially stale data before updating it.

# Additional Notes

We want this change for the Ruby tracer, since Ruby threads can migrate between OS threads under M:N scheduler (which was added in Ruby 3.3).

# How to test the change?

CI.


yannham (approved):
Reasonable addition, and implementation LGTM 👍

PR #2449, opened 2026-09-03
refactor!: apply small timeout pooling strategy to libdd-http-client as well

yannham:
# What does this PR do?

Follow-up of #2440. Rename `new_no_connection_pooling` to reflect the more nuanced semantics of the function. Apply the same treatment to the equivalent functions in libdd-http-client and libdd-agent-client.

# Motivation

See #2440.

# How to test the change?

Tested locally on the repro mentioned in #2440.

### DIFF
--- a/libdd-agent-client/src/builder.rs
+++ b/libdd-agent-client/src/builder.rs
@@ -180,10 +180,10 @@ impl AgentClientBuilder {
     /// Set whether this client is used for periodic one-shot communication (typically regularly
     /// flushing to the agent). Defaults to `true`.
     ///
-    /// Depending on the capabilities of the HTTP backend of [libdd_http_client], this setting
-    /// either sets the lifetime of pooled connections to a timeout much smaller than 60s (e.g.
-    /// 5s), disables connection pooling entirely, or does nothing if the backend has no
-    /// connection pooling to begin with. See [libdd_http_client::HttpClientBuilder::periodic].
+    /// Depending on the capabilities of the HTTP backend of [libdd_http_client], this setting sets
+    /// the lifetime of pooled connections to a timeout much smaller than 60s (e.g. 5s), or does
+    /// nothing if the backend has no connection pooling support. See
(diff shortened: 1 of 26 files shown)


libdd-agent-client/src/builder.rs
```diff
@@ -161,20 +177,21 @@ impl AgentClientBuilder {
...
-    /// second connection. The default of `false` is correct for all periodic-flush writers (traces,
-    /// stats, data streams). Set to `true` only for high-frequency continuous senders (e.g. a
-    /// streaming profiling exporter).
-    pub fn allow_connection_pooling(mut self, enabled: bool) -> Self {
-        self.allow_connection_pooling = enabled;
+    /// The Datadog agent has a low keep-alive timeout, and reusing an idle pooled connection that
+    /// the agent has closed on its side causes "pipe closed" errors on every second connection.
+    /// The default of `true` is correct for all periodic-flush writers (traces, stats, data
+    /// streams). Set to `false` only for high-frequency continuous senders (e.g. a streaming
+    /// profiling exporter).
```
VianneyRuhlmann, line 192: Do we still need a non-peridodic case ? Since we are mostly talking to the agent, we are always limited by the agent keep-alive so all client should use the agent timeout.

VianneyRuhlmann (approved):

libdd-agent-client/src/builder.rs
yannham: It's a good question. I _think_ I added that to keep some compatibility with existing API/behavior in dd-trace-py (though the agent client isn't integrated yet). Note that it's set to `true` by default, so it's just an additional knob, but maybe it's useless... If I can't find a good reason in the next 30min or so I will get rid of it as part of this PR :slightly_smiling_face:

yannham: Removed in the end

typotter (approved):
LGTM for FFE - this change just affects a test mock

PR #2452, opened 2026-09-03
perf(trace): encode agentless IDs without formatting

BridgeAR:
# What does this PR do?

Replaces formatting-backed agentless trace, span, parent, and link ID serialization with fixed-width lowercase hexadecimal buffers. Exact wire tests and lasting Criterion workloads cover common HTTP and span-link traces.

# Motivation

Formatting allocated temporary strings for every serialized ID in the agentless hot path.

# Additional Notes

With production-deduplicated spans, position-balanced Criterion runs improved common HTTP traces by 32.35% and span-link traces by 44.48% on macOS arm64.

# How to test the change?

Run `cargo test -p libdd-trace-utils --lib agentless_encoder` and `cargo bench -p libdd-trace-utils --bench main -- agentless_encoding`.

### DIFF
--- a/libdd-trace-utils/src/hex.rs
+++ b/libdd-trace-utils/src/hex.rs
@@ -8,14 +8,12 @@ const HEX_DIGITS: &[u8; 16] = b"0123456789abcdef";
 pub(crate) struct FixedHex<const N: usize>([u8; N]);
 
 impl<const N: usize> Serialize for FixedHex<N> {
-    #[inline]
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         let value = std::str::from_utf8(&self.0).map_err(S::Error::custom)?;
         serializer.serialize_str(value)
     }
 }
 
-#[inline]
 fn fixed_hex<const N: usize>(mut value: u128) -> FixedHex<N> {
     let mut encoded = [b'0'; N];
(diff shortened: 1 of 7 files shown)


yannham:
Is this PR superseded by https://github.com/DataDog/libdatadog/pull/2454? Or still worth reviewing?

BridgeAR:
@yannham it is additional to the other PR, it just needs a rebase after it lands

libdd-trace-utils/src/hex.rs
```diff
@@ -0,0 +1,46 @@
...
+        index -= 1;
+        // The mask guarantees that the conversion fits and indexes HEX_DIGITS.
+        encoded[index] = HEX_DIGITS[(value & 0x0f) as usize];
+        value >>= 4;
+    }
+
+    FixedHex(encoded)
+}
+
+#[inline]
```
yannham: Same nitpick on `#[inline]` annotations: I would be very surprised if LLVM doesn't _always_ inline those functions. If it's not the case and it's measured, I would then use `#[inline(always)]` to really force the decision.

yannham (approved):

libdd-trace-utils/src/hex.rs
BridgeAR: I will remove them

PR #2453, opened 2026-09-03
perf(trace): cache repeated SQL obfuscation

BridgeAR:
# What does this PR do?

Adds a payload-scoped two-hit SQL cache keyed by the exact raw resource and parsed DBMS. The agentless exporter reuses obfuscation only after a key repeats.

# Motivation

Agentless batches commonly contain adjacent spans with identical SQL resources, which currently repeat the full tokenizer work.

# Additional Notes

For 100 production-shaped spans, repeated SQL improved from 138.46 µs to 10.67 µs. Balanced unique-query medians remained equal at 155.74 µs versus 155.08 µs.

# How to test the change?

Run `cargo test -p libdd-trace-obfuscation -p libdd-data-pipeline-core` and `cargo bench -p libdd-trace-obfuscation --bench trace_obfuscation -- v04_sql`.


libdd-trace-obfuscation/src/obfuscate.rs
```diff
@@ -334,6 +362,69 @@ pub fn obfuscate_v04_span<T: TraceData>(span: &mut v04::Span<T>, config: &Obfusc
...
+    let dbms: DbmsKind = span
+        .meta
+        .get(TAG_DBMS)
+        .map(as_str)
+        .and_then(|dbms| TryInto::try_into(dbms).ok())
+        .unwrap_or_default();
+    if let Some(last_sql) = last_sql {
+        let resource = as_str(&span.resource);
+        let cached = last_sql.as_ref().filter(|cached| {
+            mem::discriminant(&cached.dbms) == mem::discriminant(&dbms)
```
Eldolfin: Can we instead derive `PartialEq` on `DbmsKind` instead of using `mem::discriminant` ?

libdd-trace-obfuscation/src/obfuscate.rs
```diff
@@ -334,6 +362,69 @@ pub fn obfuscate_v04_span<T: TraceData>(span: &mut v04::Span<T>, config: &Obfusc
...
+            span.meta.insert(
+                T::Text::from_static_str(TAG_SQLQUERY),
+                T::Text::from_owned(obfuscated.clone()),
+            );
+            return;
+        }
+
+        let cache_hit = cached.is_some();
+        let obfuscated = obfuscate_sql(resource, &config.sql, dbms);
+        let resource = mem::take(&mut span.resource);
```
Eldolfin, line 377: nit: `mem:swap` with the new value could save 1 allocation here

Eldolfin (commented):
Code looks correct but I feel like this adds significant complexity and it could instead be extracted to a separate cache component.

That way it could be tested by itself, be reusable and `obfuscate_v04_sql_resource` would be easier to read.

Eldolfin (commented):
Side question:
> Agentless batches commonly contain adjacent spans with identical SQL resources

Is this something measured ? Why is this happening ? I just want to make sure the optimisation is needed

---
Now for the review: the design better but it's not quite what I had in mind: the cache is still strongly coupled with the sql obfuscation.

I was more thinking of something like a generic `RepeatCacheEntry` struct that would have a function that takes a key and a function to compute the result taking the key. 
Something like:
```rust
fn resolve_with<K, V>(key: K, compute: impl FnOnce(&K) -> V)
```
This has the advantage that it can be well tested with a function that count the number of times it was called, and if we want to reuse it somewhere else later we have the option.

This is not blocking however, I'm ok with how the code looks now

Eldolfin (approved):
Code looks good now, I'm still not sure if the case we're optimizing for is realistic but fine.

BridgeAR:
@Eldolfin this might indeed be more of an edge case and I am looking into doing alternatives

PR #2454, opened 2026-09-03
perf(trace): write agentless JSON directly

BridgeAR:
# What does this PR do?

Writes fixed agentless JSON framing and field names directly into the output buffer. serde_json still owns every dynamic key, value, escape, and MessagePack conversion.

# Motivation

The generic serializer state machine adds overhead for a fixed wire schema.

# Additional Notes

This draft is stacked on the fixed-ID branch and contains only the JSON writer layer. Position-balanced Criterion runs improved that baseline by 35.44% for common HTTP traces and 20.18% for span-link traces.

# How to test the change?

Run `cargo test -p libdd-trace-utils --lib agentless_encoder` and `cargo bench -p libdd-trace-utils --bench main -- agentless_encoding`.

### DIFF
--- a/libdd-trace-utils/src/agentless_encoder/mod.rs
+++ b/libdd-trace-utils/src/agentless_encoder/mod.rs
@@ -31,8 +31,10 @@ use crate::span::v04::{AttributeAnyValue, AttributeArrayValue, Span, SpanEvent, 
 use crate::span::v1;
 use crate::span::{TraceData, SPAN_LINK_FLAGS_SET_SENTINEL};
 use crate::tracer_metadata::TracerMetadata;
-use serde::ser::{SerializeMap, SerializeSeq};
-use serde::Serializer;
+use serde::{
+    ser::{SerializeMap, SerializeSeq},
+    Serialize, Serializer,
+};
 use std::borrow::{Borrow, Cow};
 use std::collections::HashSet;
 use std::fmt::Write as _;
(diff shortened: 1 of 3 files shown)


libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -26,21 +26,44 @@
...
+struct FixedHex<const N: usize>([u8; N]);
+
+impl<const N: usize> serde::Serialize for FixedHex<N> {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let value = std::str::from_utf8(&self.0).map_err(S::Error::custom)?;
+        serializer.serialize_str(value)
+    }
+}
+
+fn fixed_hex<const N: usize>(mut value: u128) -> FixedHex<N> {
```
yannham: Nit: it's not this PR's fault, but I feel like we are writing multiple variants of this function again and again. I wonder if in each crate at least, we shouldn't write it once and for all and share it across components. Not needed for this PR, just thinking about a follow-up out loud.

libdd-trace-utils/src/agentless_encoder/mod.rs
```diff
@@ -100,192 +123,247 @@ pub fn encode_payload<T: TraceData>(
...
+        }
+        encode_trace(&mut bytes, chunk, metadata, client_side_stats)?;
+    }
+    bytes.extend_from_slice(b"]}");
     Ok(bytes)
 }
 
-fn encode_trace<T: TraceData, S: Serializer>(
-    ser: S,
+#[inline]
```
yannham: I'm not sure those inline annotations are really useful. Usually the main reason for such annotations is for cross-crate inlining (without the annotation, it's just not possible). On a private function, they are just a hint at best and LLVM is generally smart enough to inline such small functions. I would remove them and let the compiler decides if that doesn't make a difference on the benches.

If for some reason the heuristics trip up on a function specifically, and adding the annotation does make a difference, I would use `#[inline(always)]` to make sure it's inlined in the future as well (applies to the other inline annotations below on non-public functions).

yannham (approved):
Overall LGTM. The code is not much more complicated than the serde one. I am surprised by the result, I would somehow wishfully assume that the compiler and LLVM would make the serde bloat somehow less noticeable than that.

libdd-trace-utils/src/agentless_encoder/mod.rs
BridgeAR: Seems like that does already exist on main by now. I will rebase

libdd-trace-utils/src/agentless_encoder/mod.rs
BridgeAR: There is about 2-3% difference, so I will mark them as always inlined

PR #2455, opened 2026-09-03
perf(common): add full regex ASCII mode

BridgeAR:
The Node WASM pipeline evaluates a closed set of ASCII patterns. `regex-lite` removes Unicode tables, but also gives up the full engine's performance optimizations.

This adds `regex-ascii` for the full engine with `perf` and `std`, without Unicode tables. Default builds and consumers that accept user-provided replacement patterns remain Unicode-capable.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -186,6 +186,28 @@ jobs:
           fi
         env:
           RUST_BACKTRACE: full
+      - name: "[${{ steps.rust-version.outputs.version}}] cargo test (ASCII full regex)"
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          if [[ -z "$PACKAGES" ]] || echo "$PACKAGES" | grep -q "libdd-common"; then
+            cargo test \
+              --manifest-path libdd-common/tests/regex-ascii/Cargo.toml \
+              --verbose
+          fi
+        env:
+          RUST_BACKTRACE: full
+      - name: "[${{ steps.rust-version.outputs.version}}] cargo nextest run (default Unicode regex)"
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          if [[ -z "$PACKAGES" ]] || echo "$PACKAGES" | grep -q "libdd-data-pipeline-core"; then
+            cargo nextest run -p libdd-data-pipeline-core \
+              applies_unicode_replacement_rules \
+              --no-tests=fail
+          fi
+        env:
+          RUST_BACKTRACE: full
       - name: "[${{ steps.rust-version.outputs.version}}] cargo nextest run --all-features"
         shell: bash
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -150,13 +150,14 @@ jobs:
         run: |
           if [[ -z "$PACKAGES" ]] || echo "$PACKAGES" | \
             grep -Eq "libdd-(common|dogstatsd-client|http-client|trace-utils)"; then
-            cargo check -p libdd-dogstatsd-client --no-default-features
+            cargo check -p libdd-dogstatsd-client --no-default-features \
+              --features regex-unicode
             cargo check -p libdd-http-client --no-default-features \
-              --features hyper-backend
+              --features hyper-backend,regex-unicode
             cargo check -p libdd-trace-utils --no-default-features \
-              --features mini_agent
+              --features mini_agent,regex-unicode
             cargo check -p libdd-trace-utils --no-default-features \
-              --features test-utils
+              --features regex-unicode,test-utils
           fi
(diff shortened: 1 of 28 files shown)


libdd-common/Cargo.toml
Aaalibaba42, line 1: lite->ascii switch is not behavior-preserving for user-provided patterns. A pattern like `token\d+` works today under lite, and becomes a hard error under ascii-full in `libdd-trace-obfuscation` for example.

libdd-common/src/entity_id/unix/container_id.rs
```diff
@@ -15,11 +15,11 @@ const UUID_SOURCE: &str =
 /// Distinct from `UUID_SOURCE` (8-4-4-4-12) because the last group is 4 hex.
 const PCF_UUID_SOURCE: &str = r"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}";
 const CONTAINER_SOURCE: &str = r"[0-9a-f]{64}";
-const TASK_SOURCE: &str = r"[0-9a-f]{32}-\d+";
+const TASK_SOURCE: &str = r"[0-9a-f]{32}-[0-9]+";
 
 pub(crate) static LINE_REGEX: LazyLock<Regex> = LazyLock::new(|| {
     #[allow(clippy::unwrap_used)]
-    Regex::new(r"^\d+:[^:]*:(.+)$").unwrap()
+    Regex::new(r"^[0-9]+:[^:]*:(.+)$").unwrap()
```
Aaalibaba42, line 22: nit: This looks like a bug fix, but is this related to this PR ?

libdd-common/src/regex_engine.rs
```diff
@@ -3,17 +3,62 @@
...
+    not(any(feature = "regex-ascii", feature = "require-regex-full"))
+))]
 pub use regex_lite::{escape, Captures, Error, Regex, RegexBuilder, Replacer};
 
-#[cfg(not(all(feature = "regex-lite", not(feature = "require-regex-full"))))]
+#[cfg(not(all(
+    feature = "regex-lite",
+    not(any(feature = "regex-ascii", feature = "require-regex-full"))
+)))]
 pub use regex::{escape, Captures, Error, Regex, RegexBuilder, Replacer};
```
Aaalibaba42, line 38: The tradeoff is not just a question of performance. The difference between `regex` and `regex-lite` was originally made for consumers to choose between efficiency (`regex`) or compiled size (`regex-lite`). The in-between you have trades 15% more perf for more than 60% bigger artifact. That's a real trade-off that should be fully considered.

libdd-common/Cargo.toml
Aaalibaba42: BTW, this shows a gap in the ci, the ascii mode is never exercised against data-pipeline itself. The wasm job only checks data-pipeline with `regex-ascii`, nothing runs the tests in this mode.

libdd-common/tests/regex-ascii/Cargo.toml
Aaalibaba42, line 1: The standalone tests/regex-ascii crate has its own workspace so it is invisible to cargo fmt --all, workspace clippy, and cargo deny (license CSV is built from the workspace lock)

.github/workflows/test.yml
```diff
@@ -167,22 +168,63 @@ jobs:
...
+      - name: "[${{ steps.rust-version.outputs.version}}] reject missing regex mode"
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          if [[ -z "$PACKAGES" ]] || echo "$PACKAGES" | grep -q "libdd-common"; then
+            if output=$(cargo check -p libdd-common --no-default-features 2>&1); then
+              echo "::error::libdd-common compiled without a regex mode"
+              exit 1
+            fi
+            if ! grep -Fq "enable one regex engine feature" <<< "$output"; then
```
Aaalibaba42, line 204: This doesn't seem very future proof, if we every change the error message, this test will be rendered moot.

libdd-common/src/regex_engine.rs
Aaalibaba42: Btw, doing a little investigation myself about this, I'm pretty sure the change could not be enabled in libdatadog-nodejs currently, the inlined JS would go from ~230K to ~385K, 1.8 times the 210KiB budget.

One angle you might want to look at though, is `regex/unicode-perl` feature, it adds only 34K of optimized wasm, and 14.4K of inlined JS, which would stay under the current budget. I've not been to rigorous with the perf checks, but it looks promising.

Aaalibaba42 (changes requested):
This change is not really well designed, even if the idea is what you want to go with, regex-engine and feature selection is starting to look really increasingly messy.

I know it's already not the case, but Rust features were meant to be additive, not mutually exclusive like we are doing. Adding more scaffolding like this is not the direction we ought to be going.

PR #2456, opened 2026-09-03
feat(remote-config): add an agentless fetcher constructor

BridgeAR:
# What does this PR do?

Adds an explicit agentless constructor for the single-target remote-config fetcher.

# Motivation

Consumers with an agentless-only contract should not express it through an optional mode flag or retain generic transport dispatch.

# Additional Notes

The libdatadog-nodejs consumer pins commit `3bb89cfd2ce3c6f4ee3e4009c74a77e4cf89adf8`.

# How to test the change?

Run the remote-config tests, all-feature Clippy, and the no-default-features agentless check.

### DIFF
--- a/libdd-remote-config/src/fetch/single.rs
+++ b/libdd-remote-config/src/fetch/single.rs
@@ -1,6 +1,8 @@
 // Copyright 2021-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 
+#[cfg(feature = "agentless")]
+use crate::fetch::AgentlessConfig;
 use crate::fetch::{
     random_uuid_string, ConfigApplyState, ConfigClientState, ConfigFetcher, ConfigFetcherState,
     ConfigInvariants, ConfigProductCapabilities, FileStorage,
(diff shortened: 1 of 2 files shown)


libdd-remote-config/src/fetch/single.rs
```diff
@@ -171,10 +206,33 @@ where
...
+    /// Returns an error when TUF initialization fails.
+    #[cfg(feature = "agentless")]
+    pub async fn new_agentless(
+        sink: S,
+        target: Target,
+        runtime_id: String,
+        options: ConfigOptions,
+        agentless_config: AgentlessConfig,
+        http_client: C,
+    ) -> anyhow::Result<Self> {
```
paullegranddc, line 225: Enabling agentless is a matter of configuration, and in your these two constructs are strictly equivalent in term of behaviour.
 
```rust
SingleChangesFetcher::new(
  // ...,
  ConfigOptions {
    // ... whatever options,
    agentless_config: Some(AgentlessConfig {
      // ....
    }),
  },
)
```

and 

```rust
SingleChangesFetcher::new_agentless(
   // ...,
  ConfigOptions {
    // ... whatever options,
  },
  AgentlessConfig {
    // ....
  },
)
```

I don't see why it is necessary to extend our API surface to add a method that doesn't even make the code faster.

BridgeAR: Makes sense, it is probably sufficient to do this in libdatadog-nodejs. I will just check once if that is sufficent

PR #2459, opened 2026-09-03
ci(fuzz): restrict fuzz job to protected refs

yannham:
# What does this PR do?

The internal fuzzing job only needs to run on scheduled pipelines and merges to main. Keep manual runs to protected refs so the job doesn't show up on branch and merge request pipelines where it isn't needed.

# Motivation

[APSMP-3123](https://datadoghq.atlassian.net/browse/APMSP-3123)


yannham:
I totally missed this is a Gitlab job, so this doesn't apply.

PR #2464, opened 2026-09-04
fix(shared-runtime): allow disabling worker fork restart

juanjux:
# What does this PR do?

Adds `WorkerHandle::set_fork_restart(bool)`, allowing integrations to control whether an existing worker restarts in a fork child without stopping it.

# Motivation

Python-managed and native forks require different worker behavior. Before a Python fork, ddtrace must configure the telemetry worker so the child drops its inherited copy without running shutdown logic. The parent then restores normal native-fork restart behavior.

Changing this policy before the fork avoids locking or mutating the inherited worker registry in the child.

# Additional Notes

The [dependent ddtrace-py change](https://github.com/DataDog/dd-trace-py/pull/20066) requires this API and a libdatadog release containing it.

# How to test the change?

- `cargo fmt --all -- --check`
- `cargo check -p libdd-shared-runtime --all-targets --all-features`
- Tests cover dropping a worker without shutdown when child restart is disabled.

### DIFF
--- a/libdd-shared-runtime/src/shared_runtime/fork_safe.rs
+++ b/libdd-shared-runtime/src/shared_runtime/fork_safe.rs
@@ -432,6 +432,30 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_set_fork_restart_drops_worker_without_shutdown() {
+        let shared_runtime = ForkSafeRuntime::new().unwrap();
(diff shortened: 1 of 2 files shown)


yannham (approved):
I wonder if this will interact with https://github.com/DataDog/dd-trace-py/pull/20015?

PR #2465, opened 2026-09-04
fix!: benchmark was doing many samples for deterministic heap usage

Aaalibaba42:
# What does this PR do?

Make a deterministic benchmark do less samples.

# Motivation

Time it takes for this benchmark to complete (about 80s), when you have the deterministic results in 2s.

# Additional Notes

It was because the args overrode the inline config by how the bench was setup.

# How to test the change?

pipeline logs

### DIFF
--- a/libdd-common/src/bench_utils.rs
+++ b/libdd-common/src/bench_utils.rs
@@ -3,7 +3,7 @@
 
 //! Scaffolding for memory usage benchmarks.
 //!
-//! See the `ReportingAllocator` type and `memory_allocated_measurement` for usage.
+//! See the `ReportingAllocator` type and `memory_allocated_criterion` for usage.
 
 #![allow(missing_docs)]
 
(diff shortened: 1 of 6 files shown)


libdd-common/src/bench_utils.rs
```diff
@@ -22,11 +22,17 @@ impl MeasurementName for criterion::measurement::WallTime {
...
+/// deterministic, so extra samples cannot refine a constant.
+///
+/// Not usable as `criterion_group!`'s `config =` -- that macro appends `.configure_from_args()`,
+/// which would let command-line flags replace the settings below.
+pub fn memory_allocated_criterion(
     global_alloc: &'static ReportingAllocator<System>,
 ) -> Criterion<AllocatedBytesMeasurement<System>> {
     Criterion::default()
         .with_measurement(AllocatedBytesMeasurement(Cell::new(false), global_alloc))
+        .configure_from_args()
```
yannham, line 35: I'm a bit confused: aren't we saying just above that we _don't_ want `configure_from_args()`?

Aaalibaba42: It's the "appending" that is problematic, here we use args, and then change measurement time and warm up time

yannham: Ah, so here it's ok because our settings come _after_ the `configure_from_args()`, and thus override them if needed?

Aaalibaba42: Yep exactly

yannham (approved):

PR #2469, opened 2026-09-04
fix(remote-config): refresh fetcher identity

BridgeAR:
A resumed process changes its client and runtime identifiers, but an existing fetcher keeps the snapshot identity. Updating the identity in place preserves client and file state for later polls. Refs: https://github.com/DataDog/dd-trace-js/pull/10067#discussion_r3934876183


BridgeAR:
For MicroVMs, we need to be able to refresh the IDs, thus the PR.

bwoebi (approved):
I left a small change dropping the Arc<> altogether. Looks good.

PR #2472, opened 2026-09-07
fix(trace-obfuscation): scan all span meta for credit-card obfuscation

paullegranddc:
# Motivation
Credit-card obfuscation only checked the single `card.number` tag, but the trace-agent scans all span meta attributes (with an explicit skip-list)

# Changes

* Replace the single `TAG_CARD_NUMBER` fast-path in `obfuscate_pb_span` and     `obfuscate_v04_span` with a full scan of all `span.meta` entries
* Skip keys in the built-in skip-list (`should_obfuscate_cc_key`) and the user-configured `keep_values` set

Followup for #2418


yannham (approved):

Eldolfin (approved):
LGTM

note to myself: this needs to be wired up in client side stats and the config exposed in agent's /info

PR #2475, opened 2026-09-07
fix(sidecar)!: stabilize the sidecar for macos usage

bwoebi:
Macos more or less worked, but with quite a bunch of edge cases, except for weakening (which is fixed here).

This makes it mostly non-flaky.

As a side-note: Working on this, I encountered an issue where telemetry to file:// was causing interleaved data. Guarded it.


libdd-ipc/src/platform/unix/mem_handle_macos.rs
```diff
@@ -46,12 +46,15 @@ pub(crate) fn mmap_handle<T: FileBackedHandle>(mut handle: T) -> io::Result<Mapp
         }
     }
 
+    // Handle transiently not yet assigned size the same than a non-existing mapping
+    let size = NonZeroUsize::new(shm.size)
+        .ok_or_else(|| io::Error::other("shared memory mapping size not yet committed"))?;
```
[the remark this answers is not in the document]
bwoebi: > When a macOS reader opens a named segment before its writer commits the size,

won't happen

Leiyks (approved):

libdd-capabilities-impl/src/http.rs
```diff
@@ -91,6 +92,11 @@ mod native {
             std::fs::rename(&tmp, &dest)
                 .map_err(|e| HttpError::Other(anyhow::anyhow!("renaming to {dest:?}: {e}")))?;
         } else {
+            // Serialize writes to avoid large writes interleaving. This is a global lock,
+            // but totally acceptable for the debug-case of file://.
```
ekump, line 96: comment nit: This module is for generic HTTP. I'm not sure you can make the claim that any use of `file://` will be a debug-case.

libdd-ipc/src/platform/unix/sockets/macos.rs
```diff
@@ -29,10 +29,60 @@ use crate::PeerCredentials;
...
+    extern "C" {
+        fn pthread_chdir_np(path: *const libc::c_char) -> libc::c_int;
+        fn pthread_fchdir_np(fd: libc::c_int) -> libc::c_int;
+    }
+    let dir = path.parent().unwrap_or(Path::new("."));
+    let name = path.file_name().ok_or_else(|| {
+        io::Error::new(io::ErrorKind::InvalidInput, "socket path has no filename")
+    })?;
+    // Save the calling thread's CWD as an fd so we can restore it unconditionally.
+    let saved = unsafe { libc::open(c".".as_ptr(), libc::O_RDONLY) };
```
ekump: minor: should you also set `O_CLOEXEC` to prevent the fd from leaking during a fork?
```suggestion
    let saved = unsafe { libc::open(c".".as_ptr(), libc::O_RDONLY | libc::O_CLOEXEC) };
```

tools/sidecar_mockgen/src/lib.rs
```diff
@@ -124,12 +126,20 @@ pub fn weaken_object_symbols(target: &Path, binary: &Path) -> Result<(), String>
         let so_file = File::parse(bin_data.as_slice())
             .map_err(|e| format!("parse {}: {e}", binary.display()))?;
         let mut result = HashSet::new();
-        for sym in so_file.dynamic_symbols() {
+        // `dynamic_symbols()` is exported symbols only on macos.
+        #[cfg(target_os = "macos")]
```
ekump: minor: Candidate selection and the underscore stripping are gated on OS, but the [write](https://github.com/DataDog/libdatadog/pull/2475/changes#diff-aaa3f6d2f3b1d42176da909fe45d0509386bfc9111f01744c0c8a19ba5e8d760R161) dispatches  on `FileKind`. Are we opening ourselves up to an edge-case on cross-compilation?

ekump (approved):
Nothing blocking, just a couple of questions

libdd-ipc/src/platform/unix/sockets/macos.rs
bwoebi: It doesn't do harm, but it's really a short-lived fd (you see it gets fed into saved_owned right after, i.e. it'll be dropped at function end).

libdd-capabilities-impl/src/http.rs
bwoebi: Let's put it this way: *currently* all uses of it are debug case.
If this is no longer going to be debug-case, we probably don't want to serialize this on a global lock for any file write. So that comment is accurate for now.

tools/sidecar_mockgen/src/lib.rs
bwoebi: Makes sense, not an actual problem for what we do, but changed :-)

PR #2476, opened 2026-09-07
chore: move all remaining external deps to workspace-level dependencies

yannham:
# What does this PR do?

Follow-up of #2350. Migrate all of the remaining external dependencies to the workspace level. `duplicate` was used with incompatible major versions. Mostly the macro syntax changed. Files are updated to use the latest version's syntax.

# Motivation

As for the other workspace-level deps: use a single source of truth and version, for each dependency, in the whole workspace. Force leaf crates to opt-in features they need instead of blindly enabling default features.

### DIFF
--- a/libdd-trace-obfuscation/src/http.rs
+++ b/libdd-trace-obfuscation/src/http.rs
@@ -302,509 +302,78 @@ mod tests {
     use super::obfuscate_url_string;
 
     #[duplicate_item(
-        [
-            test_name           [remove_query_string_1]
-            remove_query_string [true]
-            remove_path_digits  [false]
-            input               ["http://foo.com/"]
-            expected_output     ["http://foo.com/"];
-        ]
-        [
-            test_name           [remove_query_string_2]
-            remove_query_string [true]
-            remove_path_digits  [false]
-            input               ["http://foo.com/123"]
-            expected_output     ["http://foo.com/123"];
-        ]
-        [
-            test_name           [remove_query_string_3]
-            remove_query_string [true]
-            remove_path_digits  [false]
-            input               ["http://foo.com/id/123/page/1?search=bar&page=2"]
-            expected_output     ["http://foo.com/id/123/page/1?"];
-        ]
-        [
-            test_name           [remove_query_string_4]
-            remove_query_string [true]
-            remove_path_digits  [false]
-            input               ["http://foo.com/id/123/page/1?search=bar&page=2#fragment"]
-            expected_output     ["http://foo.com/id/123/page/1?#fragment"];
-        ]
-        [
-            test_name           [remove_query_string_5]
-            remove_query_string [true]
-            remove_path_digits  [false]
-            input               ["http://foo.com/id/123/page/1?blabla"]
-            expected_output     ["http://foo.com/id/123/page/1?"];
-        ]
-        [
-            test_name           [remove_query_string_6]
-            remove_query_string [true]
-            remove_path_digits  [false]
-            input               ["http://foo.com/id/123/pa%3Fge/1?blabla"]
-            expected_output     ["http://foo.com/id/123/pa%3Fge/1?"];
-        ]
-        [
-            test_name           [remove_query_string_7]
-            remove_query_string [true]
-            remove_path_digits  [false]
-            input               ["http://user:password@foo.com/1/2/3?q=james"]
-            expected_output     ["http://foo.com/1/2/3?"];
-        ]
-        [
-            test_name           [remove_path_digits_1]
-            remove_query_string [false]
-            remove_path_digits  [true]
-            input               ["http://foo.com/"]
-            expected_output     ["http://foo.com/"];
-        ]
-        [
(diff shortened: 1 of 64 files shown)


typotter (approved):
lgtm for FFE

libdd-trace-obfuscation/src/http.rs
```diff
@@ -302,509 +302,78 @@ mod tests {
     use super::obfuscate_url_string;
 
     #[duplicate_item(
-        [
```
VianneyRuhlmann, line 305: nit: I think it's better to avoid unrelated changes especially in PRs that change many files

libdd-trace-protobuf/Cargo.toml
```diff
@@ -13,9 +13,11 @@ bench = false
 
 [dependencies]
 bolero = { workspace = true, features = ["std"], optional = true }
-prost = "0.14.1"
+prost = { workspace = true, default-features = true }
 serde = { workspace = true, features = ["derive"] }
-serde_bytes = "0.11.9"
+# `std` is required: it gates the alloc-backed `ByteBuf`/`Vec<u8>` serde impls used by
+# `#[serde(with = "serde_bytes")]` in the generated code.
```
VianneyRuhlmann: I'm not sure this is needed, I don't think the crate is supposed to no-std anyway

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -674,7 +674,7 @@ impl StatsBucket {
                     return;
                 }
                 // Within the max entry limit, admit key as a new distinct entry.
-                e.insert(GroupedStats::default())
+                e.insert_with_key(OwnedAggregationKey::from(&key), GroupedStats::default())
```
VianneyRuhlmann, line 677: I think it would be a little bit more efficient to implement `ToOwned` on `BorrowedAggregationKey` since this avoids [the `equivalent` check](https://docs.rs/hashbrown/latest/src/hashbrown/map.rs.html#4612). This can be done in follow-up.

VianneyRuhlmann (approved):

libdd-trace-obfuscation/src/http.rs
yannham: Very fair point, especially given that I'm the first one to complain of big, unfocused PR :stuck_out_tongue: I thought about splitting it but I left it there because the changes were entirely mechanical here (migrating to the new macro format)

libdd-trace-protobuf/Cargo.toml
yannham: Sorry, what do yo think is not needed? The comment? The default-features?

libdd-trace-stats/src/span_concentrator/aggregation.rs
yannham: Ah, good catch! Will do in a follow up.

libdd-trace-obfuscation/src/http.rs
VianneyRuhlmann: I thought it was just a formatting change, since it's caused by the major update it's fine. That's on me for not reading the PR description 😅

libdd-trace-protobuf/Cargo.toml
VianneyRuhlmann: The comment is mentioning `std` which is a bit confusing (not clear it's talking about the feature). I guess it's fine to explain why we add the default feature it just seems weird that it's the only dependency explaining it.

yannham: Ah, got it. I think it's because elsewhere I've avoided `std` features on many crates, preferring `alloc`, in the spirit of the lint we have (make it easier to move crates to `no-std` in the future). Though the comment is dubious because `serde-bytes` does have an `alloc` feature. I'll either remove it if I can get by with `alloc`, or reword it to make things clearer :+1:

paullegranddc (approved):

bwoebi (approved):

gyuheon0h (approved):

vpellan (approved):

yannham:
@codex review

libdd-trace-stats/src/span_concentrator/aggregation.rs
yannham: So, I took a stab at this, and it's more complicated than anticipated. Implementing `ToOwned` requires `type Borrowed: Borrow<Self>`, but we can't really implement the latter (ie `OwnedAggregationKey: Borrow<BorrowedAggregationKey>`) because `OwnedAggregationKey` and `BorrowedAggregationKey` just have a different layout (typically we need to turn a `&Vec<_, _>` into `Vec<&_, _>` which is not possible without allocating). So there's no cheap conversion. The last resort would be to use an unsafe escape hatch, but it's not worth it IMHO. So dropping the idea for now (I wishfully hope that it might be elided after inlining, but I haven't checked).

PR #2477, opened 2026-09-07
feat(trace-utils)!: add span pool for recycling span allocations

paullegranddc:
# Motivation

Allocation is expensive, and spans use a lot of small collections (meta, metrics, span links, events...). Being able to recycle the allocation should be beneficial in term of perf.

This PR adds the pool types only. A stacked follow-up PR (#2382) switches the data-pipeline send path over to them.

# Changes

 - Add `SpanPool<T>` in `libdd-trace-utils/src/span/span_pool.rs`
     - Backed by a bounded crossbeam channel so the pool never grows without limit.
     - A thread-local chunk cache keeps the single-producer (exporter) path lock-free and gives each thread a local chunk under contention.
 - Add `PooledChunks<'_, T>`, which wraps `Vec<Vec<Span<T>>>` and returns its spans to a `SpanPool` on drop. `PooledChunks::unpooled()` is a zero-overhead wrapper for callers that do not use the pool.
 - Add `MaybePool` so call sites can feed spans/chunks back to the pool only when one is attached, and otherwise drop them.
 - Add `Send` bounds to `SpanText`/`SpanBytes` so pooled `Vec<Span<T>>` can flow through the crossbeam channel and thread-local cache.
 - Add criterion benchmarks for the pool (recycle vs. allocate) under `libdd-trace-utils/benches/span_pool.rs`.

# Performance

This makes the alloc/populate/dealloc cycle about 20% to 30% faster

| Name | Allocation (ns) | Pooled (ns) | Increase |
| - | ----------- | ----------- | - |
| small chunk (1x10) | 842.73 | 616.29 | x1.36
| big chunks (20x100) | 171560 | 144270  | x1.18 

# Additional Notes

Pure addition, no behavior change. The new types are not used anywhere yet. The stacked PR #2382 wires them into the trace exporter.

### DIFF
--- a/libdd-trace-utils/src/span/span_pool.rs
+++ b/libdd-trace-utils/src/span/span_pool.rs
@@ -1,0 +1,421 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+use super::v04::Span;
+use super::TraceData;
+use rand::{Rng as _, SeedableRng as _};
+use std::cell::RefCell;
+use std::ops::{Deref, DerefMut};
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+use thread_local::ThreadLocal;
+
+/// When this function returns true, do not add the returned chunk to the queue.
+///
+/// Why are we doing this?
+///
+/// If we keep recylcing spans forever, two things are going to happen
+/// * We will keep the **maximum** number of spans ever used by the program alive, even if memory
+///   usage scales down
+/// * As spans get reused and atributes data structure are pushed and popped, they will tend to grow
+///   to have the maximum size of attributes
+///
+/// The policy operates at chunk granularity: `add_chunks` draws once per chunk and either
+/// recycles the whole chunk or drops it. Dropping a fixed pct of chunks returned ensures that
+/// we eventually free memory if span usage spikes, and then goes down.
+fn drop_policy() -> bool {
+    const PCT_OF_SPANS_RETURNED_DROPPED: f64 = 0.1;
+    thread_local! {
+        static RNG: RefCell<rand::rngs::SmallRng> = RefCell::new(rand::rngs::SmallRng::from_entropy());
+    }
+    RNG.with_borrow_mut(|r| r.gen_bool(PCT_OF_SPANS_RETURNED_DROPPED))
+}
+
+/// Reset fields to default, keeping collection capacity for reuse.
+fn reset_span<T: TraceData>(
+    Span {
+        service,
+        name,
+        resource,
+        r#type,
+        trace_id,
+        span_id,
+        parent_id,
+        start,
+        duration,
+        error,
+        meta,
+        metrics,
+        meta_struct,
+        span_links,
+        span_events,
+    }: &mut Span<T>,
+) {
+    *service = Default::default();
+    *name = Default::default();
+    *resource = Default::default();
+    *r#type = Default::default();
+    *trace_id = Default::default();
+    *span_id = Default::default();
+    *parent_id = Default::default();
+    *start = Default::default();
+    *duration = Default::default();
+    *error = Default::default();
+    meta.clear();
+    metrics.clear();
+    meta_struct.clear();
+    span_links.clear();
+    span_events.clear();
+}
+
+fn reset_chunk<T: TraceData>(chunk: &mut Vec<Span<T>>) {
+    for span in chunk {
+        reset_span(span);
+    }
+}
+
+/// Max spans per recycled chunk. Larger chunks are split so no thread hoards a big chunk in its
+/// local cache.
+const MAX_CHUNK_SIZE: usize = 20;
+
+/// Split a chunk into pieces of at most [`MAX_CHUNK_SIZE`] spans, keeping each piece's spans'
+/// buffer capacity for reuse
+fn split_chunk<T: TraceData>(chunk: Vec<Span<T>>) -> impl Iterator<Item = Vec<Span<T>>> {
+    let mut remaining = chunk;
+    std::iter::from_fn(move || {
+        if remaining.is_empty() {
+            return None;
+        }
+        if remaining.len() <= MAX_CHUNK_SIZE {
+            return Some(std::mem::take(&mut remaining));
+        }
+        let at = remaining.len() - MAX_CHUNK_SIZE;
+        Some(remaining.split_off(at))
+    })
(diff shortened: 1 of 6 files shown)


libdd-trace-utils/src/span/span_pool.rs
```diff
@@ -0,0 +1,421 @@
...
+///
+/// Backed by an unbounded crossbeam channel of chunks (one send per chunk, not per span). The
+/// capacity (in spans) bounds the channel only. Thread-local caches are not counted, so idle
+/// threads holding cached spans don't reduce the pool's headroom.
+///
+/// The capacity bound is best-effort: `add_chunks` checks-and-increments `len` with relaxed
+/// atomics, so concurrent producers can briefly exceed `capacity` (the channel itself is
+/// unbounded).
+/// Since there is a single task returning chunks to the queue (the exporter task) the bound is
+/// exact. `get_span` first hits a per-thread
```
VianneyRuhlmann: Aren't these two lines in contradiction ?

libdd-trace-utils/src/span/span_pool.rs
```diff
@@ -0,0 +1,421 @@
...
+    last_chunk: Option<Vec<Span<T>>>,
+    /// One recycled empty `Vec` kept for `pull_empty_chunk`; never returned to the pool.
+    empty_chunk: Option<Vec<Span<T>>>,
+}
+
+impl<T: TraceData> SpanPool<T> {
+    /// New pool holding at most `pool_capacity` recycled spans.
+    pub fn new(pool_capacity: usize) -> Self {
+        Self::with_capacity(pool_capacity)
+    }
```
VianneyRuhlmann: I'm not sure this function is useful since `with_capacity` is already public

libdd-trace-utils/src/span/span_pool.rs
```diff
@@ -0,0 +1,421 @@
...
+            if chunk.is_empty() || drop_policy() {
+                continue;
+            }
+            reset_chunk(&mut chunk);
+            for piece in split_chunk(chunk) {
+                let piece_len = piece.len();
+                // Reserve span-count atomically against the cap; drop the piece if it won't fit.
+                let current = self.inner.len.load(Ordering::Relaxed);
+                if current + piece_len > self.inner.capacity {
+                    return;
```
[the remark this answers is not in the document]
paullegranddc: Capacity should never be this low. Going to document this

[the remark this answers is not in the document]
iunanua: but, shouldn't it be a `continue` instead of a `return` to see if the next chunk fits?

libdd-trace-utils/src/span/span_pool.rs
```diff
@@ -0,0 +1,437 @@
...
+/// * As spans get reused and atributes data structure are pushed and popped, they will tend to grow
+///   to have the maximum size of attributes
+///
+/// The policy operates at chunk granularity: `add_chunks` draws once per chunk and either
+/// recycles the whole chunk or drops it. Dropping a fixed pct of chunks returned ensures that
+/// we eventually free memory if span usage spikes, and then goes down.
+fn drop_policy() -> bool {
+    const PCT_OF_SPANS_RETURNED_DROPPED: f64 = 0.1;
+    thread_local! {
+        // https://xkcd.com/221/
```
iunanua: xDD

libdd-trace-utils/src/span/span_pool.rs
```diff
@@ -0,0 +1,421 @@
...
+            }
+            reset_chunk(&mut chunk);
+            for piece in split_chunk(chunk) {
+                let piece_len = piece.len();
+                // Reserve span-count atomically against the cap; drop the piece if it won't fit.
+                let current = self.inner.len.load(Ordering::Relaxed);
+                if current + piece_len > self.inner.capacity {
+                    return;
+                }
+                self.inner.len.fetch_add(piece_len, Ordering::Relaxed);
```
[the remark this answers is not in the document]
paullegranddc: There is usually only one task returning spans at a time, since this happens in the trace exporter which is a one of.
Changing to a compare swap is more expensive

libdd-trace-utils/src/span/span_pool.rs
```diff
@@ -0,0 +1,421 @@
...
+
+    #[test]
+    fn returned_spans_are_recycled_and_reset() {
+        let pool = SpanPool::<crate::span::BytesData>::with_capacity(100);
+        {
+            // No drop-policy control here, but a single span is very likely retained.
+            // If we drop 10% of spans, the likelyhood all spans are dropped is 1/10**100
+            // which is basically never happening if we ran this test until the heat death of
+            // this universe
+            let chunks = pool.wrap_chunks(vec![vec![span("a"); 100]]);
```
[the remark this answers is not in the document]
paullegranddc: Fixed

libdd-trace-utils/src/span/span_pool.rs
paullegranddc: The first line refers to the fact there is a thred cache, the second to the fact that we do a fetch + fetch add instead of a compare swap when retuning spans. I'll clarify the comment

libdd-trace-utils/src/span/span_pool.rs
```diff
@@ -0,0 +1,437 @@
...
+            None => Vec::new(),
+        }
+    }
+}
+
+/// Owned trace chunks that return to a [`SpanPool`] on drop. Deref's to the inner
+/// `Vec<Vec<Span<T>>>` for in-place processing. With no pool ([`PooledChunks::unpooled`]) spans
+/// just drop normally.
+#[derive(Debug)]
+pub struct PooledChunks<'a, T: TraceData> {
```
VianneyRuhlmann, line 288: I find it confusing that chunks can mean both "TraceChunk" i.e. related spans from the same trace and "SpanChunk" i.e. Buffer of uninitialized spans that can be used for different traces

VianneyRuhlmann (approved):

libdd-trace-utils/src/span/span_pool.rs
paullegranddc: renamed the inner type to SpabBatch, to reduce the confusion between logical chunks and list of spans

libdd-trace-utils/src/span/span_pool.rs
paullegranddc: removed

VianneyRuhlmann:
Merged as part of !2382

PR #2478, opened 2026-09-08
fix(shared-runtime): discard inherited MicroVM runtime data

litianningdatadog:
## Why

AWS Lambda MicroVM snapshots can restore a process with trace, stats, and telemetry data accumulated under a previous runtime identity. This adds a no-flush identity-refresh path so `/run` callers can discard that inherited exporter state before creating a replacement exporter.

It does not eliminate the small window between process restore and completion of the `/run` refresh. Data emitted in that window may still carry the snapshotted runtime ID. The RFC accepts that MVP gap; closing it requires a pre-snapshot worker hold and the associated lifecycle and accounting changes.

## How

Adds a discard-only exporter teardown for the MicroVM identity-refresh path.

`TraceExporter::shutdown_without_flush()` races each worker's in-flight async run against cancellation rather than waiting for normal pause/shutdown behavior. It does not preempt synchronous work already running inside a worker or remove the pre-`/run` race. It removes the worker through `Worker::discard()`. `discard()` defaults to `reset()`, but lets a worker override cleanup that must differ from its fork-continuation `reset()`; the trace-buffer worker uses this to close its channel instead of reopening it. That worker is not wired into an exporter yet, but the generic discard contract needs to handle it correctly before it is. The exporter also clears the global agent-info cache so the replacement exporter does not skip its first `/info` fetch. The FFI exposes this as `ddog_trace_exporter_free_without_flush`.

This is intentionally scoped to the identity-refresh API. Existing `WorkerHandle::stop`, runtime shutdown, and fork lifecycle behavior are unchanged: `pause()` still waits for an in-flight run to finish, and `reset()` still reopens a worker for continued use. Callers use discard in the MicroVM startup identity-refresh path, not concurrently with ordinary shutdown or a real fork.

## References

- [RFC](https://docs.google.com/document/d/17lde4Zak2YRBuDvf32KFanXj9TdpOv24GYaEY8swA1w/edit?usp=sharing)
- [AWS Lambda](https://aws.amazon.com/lambda/lambda-microvms/)
- [PRNG uniqueness in AWS Lambda MicroVM](https://github.com/aws/agent-toolkit-for-aws/blob/847f477649252b98f8fb828bbfeaf109b57b8cac/skills/specialized-skills/serverless-skills/aws-lambda-microvms/references/snapshots-and-uniqueness.md)

## Verification

- `cargo check -p libdd-shared-runtime -p libdd-data-pipeline -p libdd-data-pipeline-ffi`
- `cargo test -p libdd-shared-runtime`
- `cargo test -p libdd-data-pipeline test_shutdown_without_flush`
- `cargo test -p libdd-data-pipeline test_worker_discard`
- `cargo test -p libdd-data-pipeline-ffi exporter_free_without_flush`
- `cargo +stable clippy -p libdd-shared-runtime --all-targets -- -D warnings`


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -375,6 +395,37 @@ impl<
...
+        if let Some(telemetry) = self.workers.telemetry {
+            handles.push(telemetry);
+        }
+
+        let mut futures: FuturesUnordered<_> = handles.into_iter().map(|h| h.discard()).collect();
+
+        while let Some(result) = futures.next().await {
+            if let Err(e) = result {
+                error!("Worker failed to discard: {:?}", e);
+            }
```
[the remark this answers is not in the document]
litianningdatadog: The current implementation propagates discard failures rather than treating them as success. `discard_workers` waits for every `WorkerHandle::discard()` future, retains the first error, and returns it to `shutdown_without_flush`.

hoolioh (commented):
have you considered controlling the closing behavior through a variable in the exporter builder?
If the case is that you don't have to use both methods (flus and no-flush) in the same flow, IMO, it would be better to encapsulate the behaviour through the exporter handle. That way we will avoid duplicating the external API which comes with some downsides in terms on maintability and binary size.

litianningdatadog:
> have you considered controlling the closing behavior through a variable in the exporter builder? If the case is that you don't have to use both methods (flus and no-flush) in the same flow, IMO, it would be better to encapsulate the behaviour through the exporter handle. That way we will avoid duplicating the external API which comes with some downsides in terms on maintability and binary size.

@hoolioh Thanks for the comments. My thought is this should stay as an explicit close API because the no-flush path has a different ownership/error contract.

On discard failure, the FFI path keeps the exporter handle alive so the caller can retry. The existing `ddog_trace_exporter_free` consumes the handle and has no way to report an error. If we move this behind a builder flag, `free` either has to swallow discard failures or we still need a second fallible close function.

I’m fine sharing more internal plumbing to avoid duplicated shutdown code, but I think keeping public discard path explicit is the better and safer approach.

libdd-data-pipeline-ffi/src/tracer.rs
```diff
@@ -1554,8 +1554,8 @@ mod tests {
             assert!(event.0.attributes.is_empty());
 
             // Misaligned pointer
-            let buf = [0u8; 32];
-            let misaligned = buf.as_ptr().add(1).cast::<i64>();
```
VianneyRuhlmann, line 1558: Is this related to the PR ?

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -257,6 +257,31 @@ impl WorkerHandle {
...
+                .iter()
+                .position(|entry| entry.id == self.worker_id)
+            else {
+                return Err(WorkerHandleError::AlreadyStopped);
+            };
+            let WorkerEntry { worker, .. } = workers_lock.swap_remove(position);
+            worker
+        };
+        worker.pause().await?;
+        worker.discard();
```
VianneyRuhlmann: I don't think you need to reset here since the worker will be dropped anyway.

libdd-shared-runtime/src/worker.rs
```diff
@@ -40,6 +40,14 @@ pub trait Worker: std::fmt::Debug + MaybeSend {
 
     /// Hook called when the app is shutting down. Can be used to flush remaining data.
     async fn shutdown(&mut self) {}
+
+    /// Discard buffered data without performing the normal shutdown flush.
+    ///
+    /// By default, discard has the same state-clearing behavior as [`Worker::reset`]. Workers
+    /// that need additional discard-only cleanup can override this method.
+    fn discard(&mut self) {
```
VianneyRuhlmann, line 49: Do you have an example of a worker that would require this hook ? I'd rather be conservative here and only add the hook if there's a need for it.

VianneyRuhlmann (commented):
The approach is looking fine. However just a heads up, I'm working on making the runtime_id (and other metadata) mutable in the workers. Would it work for you if you were able to update the runtime_id and reset the worker state (we could expose `reset` on the shared runtime) ?

litianningdatadog:
> The approach is looking fine. However just a heads up, I'm working on making the runtime_id (and other metadata) mutable in the workers. Would it work for you if you were able to update the runtime_id and reset the worker state (we could expose `reset` on the shared runtime) ?

@VianneyRuhlmann Thanks for your input. Here is my thought: mutable `runtime_id` would help avoid rebuilding workers just to refresh future metadata, but we still prefer a no-flush reset boundary for pre-`/run` state.

Our baseline is that we must not emit data with a stale `runtime_id`. Retagging buffered data satisfies that mechanically, but it can still produce misleading data. Anything in the buffer before `/run` may come from snapshot/build-time state, validation traffic, stale worker schedules, or aggregation windows that no longer line up with the new MicroVM instance. The label would be current, but the data would not really belong to that runtime.

That also creates a duplication problem: if the same snapshot contains buffered stats or telemetry, each restored MicroVM could emit the same buffered data under its own fresh `runtime_id`.

So I think the reset approach in this PR is still the right first step: pause the worker, clear buffered and aggregate state without flushing, then let future work run under the refreshed identity. If mutable worker metadata becomes available later and gives us cleaner lifecycle handling, we can switch to that API in a future release. But I’d still want the MicroVM refresh path to reset/drop pre-refresh state rather than only retag it.

libdd-data-pipeline-ffi/src/tracer.rs
litianningdatadog: It was exposed by this PR. The original assumption was invalid where `[u8; 32]` only has alignment 1, so Rust/Miri may place it at any byte address. If `buf.as_ptr()` happens to be `7 mod 8`, adding 1 produces an address that is aligned for `i64`, so `!misaligned.is_aligned()` fails. So this is not a public/API breaking change from the PR. It is a test input construction bug surfaced by PR CI/Miri because this PR caused `libdd-data-pipeline-ffi` tests to run.

.gitignore
```diff
@@ -15,7 +15,7 @@ release
 target
 tmp
 __fuzz__
-
+graphify-out
```
litianningdatadog: this is ignore [Graphify](https://github.com/Graphify-Labs/graphify) output

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
```diff
@@ -125,32 +139,54 @@ impl<T: Worker + MaybeSend + Sync + 'static> PausableWorker<T> {
...
-                                worker.run().await;
                             }
                         }
+                        select! {
+                            biased;
+                            _ = cloned_discard_token.cancelled() => {
+                                break;
+                            }
+                            _ = worker.run() => {}
+                        }
```
VianneyRuhlmann, line 178: Threads are restored when the MicroVM is starting from a snapshot right ? This doesn't guarantee that we wont send traces in the new instance:

1. the thread could send before you call discard
2. `select!` can only abort the task when await is called so if `run` has been called and it doesn't `await` before flushing (which is often the case since the main reason we need async is for io calls) the stale data will be flushed.

Would it be an option to pause all the threads in `/ready` or `/validate` hook like it's done for forks this way in the new image all threads are stopped and there's no risks of flushing stale data.

.gitignore
```diff
@@ -15,7 +15,7 @@ release
 target
 tmp
 __fuzz__
-
+/graphify-out
```
litianningdatadog, line 18: it is to ignore the output of [graphify](https://github.com/Graphify-Labs/graphify). But we can remove it if it is decided to use it as a formal tool

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -375,6 +393,47 @@ impl<
...
+        let mut handles: Vec<WorkerHandle> = Vec::new();
+
+        if let StatsComputationStatus::Enabled { worker_handle, .. } =
+            &**self.client_side_stats.status.load()
+        {
+            handles.push(worker_handle.clone());
+        }
+
+        if let Some(info_fetcher) = &self.workers.info_fetcher {
+            handles.push(info_fetcher.clone());
```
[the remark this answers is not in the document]
litianningdatadog: Resolved as `shutdown_without_flush_async` now clears `AGENT_INFO_CACHE` after discard, and the test asserts it. The fresh `AgentInfoFetcher` therefore fetches immediately rather than waiting for its refresh interval.

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
litianningdatadog: You're correct. Removing this race entirely would require pausing workers before the MicroVM snapshot is created when the event `/ready` is fired, then resuming them only after `/run` refreshes the identity. We are treating that as follow-up work. The [RFC accepts the small gap](https://docs.google.com/document/d/17lde4Zak2YRBuDvf32KFanXj9TdpOv24GYaEY8swA1w/edit?tab=t.0#bookmark=id.wy1gul594azu) between process restore and completion of `/run`, because a global hold would also require decisions around queue growth, lost startup diagnostics, lifecycle duplication, and writer/drop accounting.  This PR only adds the no-flush path used during `/run` to discard inherited exporter state before creating the replacement exporter.

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -901,6 +901,44 @@ pub unsafe extern "C" fn ddog_trace_exporter_free(handle: Box<TraceExporter>) {
...
+/// non-snapshot shutdown callers should continue using [`ddog_trace_exporter_free`].
+///
+/// Returns `None` when the inherited state was discarded. Once a non-null handle is accepted,
+/// `*handle` is always set to null, including when an ordinary error is returned.
+///
+/// # Arguments
+///
+/// * handle - A non-null pointer to the TraceExporter handle. The pointee must be non-null.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_trace_exporter_free_without_flush(
```
VianneyRuhlmann: nit: It should be `ddog_trace_exporter_free_without_shutdown` to make it clear that all worker `shutdown` logic is skipped

libdd-data-pipeline/src/agent_info/mod.rs
```diff
@@ -29,12 +29,17 @@ pub fn get_agent_info() -> Option<Arc<schema::AgentInfo>> {
...
+
 pub use fetcher::{
     fetch_info, fetch_info_with_state, AgentInfoFetcher, FetchInfoStatus, ResponseObserver,
 };
 
 #[cfg(test)]
 /// Clear the global agent info cache for test isolation in envs where nextest isn't used.
 pub fn clear_cache_for_test() {
-    AGENT_INFO_CACHE.store(None);
+    clear_cache_for_runtime_identity_refresh();
```
VianneyRuhlmann, line 44: use the new function in the tests directly

libdd-data-pipeline-ffi/src/trace_exporter.rs
```diff
@@ -901,6 +901,44 @@ pub unsafe extern "C" fn ddog_trace_exporter_free(handle: Box<TraceExporter>) {
     let _ = catch_panic!(handle.shutdown(None), Ok(()));
 }
 
+/// Discards runtime-identity-bound state without sending buffered data, then frees the exporter.
```
VianneyRuhlmann, line 904:
```suggestion
/// Stop workers without sending buffered data, then frees the exporter.
```

It isn't runtime_id specific it deletes all state

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -321,6 +322,24 @@ impl<
         runtime.block_on(self.shutdown_async(timeout))?
     }
 
+    /// Discard inherited runtime-identity state without sending buffered worker data.
```
VianneyRuhlmann, line 325:
```suggestion
    /// Stop workers without sending buffered data.
```

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -344,6 +363,16 @@ impl<
         }
     }
 
+    /// Async version of [`Self::shutdown_without_flush`].
+    ///
+    /// # Cancel safety
+    /// This function is *NOT* cancel safe. If cancelled, workers can be left in an invalid state.
+    pub async fn shutdown_without_flush_async(self) -> Result<(), TraceExporterError> {
+        let discard_result = self.discard_workers_without_flush().await;
+        agent_info::clear_cache_for_runtime_identity_refresh();
```
VianneyRuhlmann, line 372: I don't think this should be called here the method should either be called in the python handler which handles all the snapshot-reset logic or in the agent info fetcher discard.

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -375,6 +404,50 @@ impl<
         }
     }
 
+    async fn discard_workers_without_flush(&self) -> Result<(), TraceExporterError> {
```
VianneyRuhlmann, line 407: To stay coherent with the other method names this should be `discard_workers` or `shutdown_workers_without_flush`

libdd-data-pipeline/src/trace_exporter/mod.rs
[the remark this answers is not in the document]
VianneyRuhlmann: We don't want to propagate the errors here, these are mostly not actionnable by the tracer and don't matter much in this case since the data is meant ot be discarded anyway. This method should replicate `shutdown_workers` error handling i.e. logging and returning `Ok`. Regarding the sync issue with shutdown_async, if a normal shutdown have been triggered data will be flushed and we can't prevent it, there's no reason for this method to wait for the flush to be over.

libdd-shared-runtime/src/shared_runtime/mod.rs
```diff
@@ -259,6 +259,45 @@ impl WorkerHandle {
...
+    /// choose discard semantics when they meant ordinary shutdown.
+    ///
+    /// # Errors
+    /// Returns an error if the worker has already been stopped.
+    ///
+    /// # Cancel safety
+    /// This function is *NOT* cancel safe and should not be called from
+    /// [`Worker::trigger`](crate::worker::Worker::trigger). If cancelled, the discarded worker can
+    /// end up in an invalid state.
+    pub async fn discard_worker_without_flush(
```
VianneyRuhlmann, line 282: This shouldn't be it's own module and should be a method of `WorkerHandle` also it should be named `discard` as discard already implies without_flush.

libdd-shared-runtime/src/shared_runtime/pausable_worker.rs
VianneyRuhlmann: I'm not sure I see the point of this change then. Either sending stale data with the wrong runtime_id is acceptable, or it's not. If it's acceptable then you don't need to cancel inflight requests or even prevent the flush on shutdown. If it isn't then you need to stop the threads before the snapshot. Otherwise the unwanted flush will happen and since it depends on the state of the workers before the snapshot, an image encountering it will send stale data on every run.

PR #2483, opened 2026-09-08
chore(deps): bump zrip to 0.8.7

BridgeAR:
zrip 0.8.7 requires Rust 1.89, so this moves the dependency and workspace MSRV together. It lets downstream WASM consumers use one zrip version instead of retaining the 0.6.0 stack.

Refs: https://github.com/DataDog/libdatadog-nodejs/pull/270

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7503,9 +7503,9 @@ dependencies = [
 
 [[package]]
 name = "zrip"
-version = "0.6.0"
+version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "964fe9f1ea10a0d183fc143a7525266cbe16978883dbe304725da34b314c04cf"
+checksum = "edd02f990082191d398ad4ffb08dc3d92174f275b0b5e8b76d12eddd3e486bcf"
 dependencies = [
  "zrip-core",
  "zrip-decode",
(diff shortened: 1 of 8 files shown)


ekump:
You can't just arbitrarily bump libdatadog's MSRV. We currently have customers using dd-trace-rs with an MSRV below 1.89.

BridgeAR:
@ekump that makes sense. Do we have a documentation for what is considered internal and what must stay as is and what support guarantees we have?

paullegranddc:
> Do we have a documentation for what is considered internal and what must stay as is and what support guarantees we have?

We have a policy of being at least 1 year behind the latest rust version. So 1.89 would be fine, except for the fact that we already bumped the MSRV 3 months ago.
Bumping the MSRV would require a new major version of dd-trace-rs and we're not ready to do it that often

ekump:
> Do we have a documentation for what is considered internal and what must stay as is and what support guarantees we have?

@BridgeAR - The documentation is a bit disjointed at the moment. I'll fix that so we have something clear. We also need to officially publish the dd-trace-rs support policy. As @paullegranddc said, when we bump the MSRV of dd-trace-rs we bump to at most one year behind latest. I'll discuss with the libdatadog guild, and our team and figure out how soon we can update the MSRV again, hopefully in Q4.

PR #2485, opened 2026-09-08
chore(deps): update thiserror to 2

BridgeAR:
Updating libdatadog-nodejs to thiserror 2 leaves project-owned thiserror 1 dependencies in its WASM graph, and the artifact validator rejects both majors.


VianneyRuhlmann (approved):

yannham (changes requested):
Could we wait on https://github.com/DataDog/libdatadog/pull/2476 instead  ? It does bump thiserror, and would avoid a rebase

PR #2488, opened 2026-09-08
feat(data-pipeline)!: generate agentless trace stats

BridgeAR:
Agentless v0.4 export now generates trace stats from the decoded trace payload. Traces and stats share one exporter and one WASM artifact.


bengl (approved):
Overall looks fine, but I'd prefer if someone from @DataDog/apm-common-components-core approved.

libdd-trace-stats/src/span_concentrator/aggregation.rs
```diff
@@ -674,7 +674,8 @@ impl StatsBucket {
                     return;
                 }
                 // Within the max entry limit, admit key as a new distinct entry.
-                e.insert_with_key(OwnedAggregationKey::from(&key), GroupedStats::default())
+                e.insert(OwnedAggregationKey::from(&key), GroupedStats::default())
+                    .1
```
paullegranddc, line 678: I don't think this is necessary

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -126,6 +127,251 @@ impl From<TracerMetadata> for StatsMetadata {
...
+    async fn send_payload(
+        &self,
+        capabilities: &Cap,
+        meta: &StatsMetadata,
+        sequence_id: &AtomicU64,
+        buckets: Vec<pb::ClientStatsBucket>,
+        obfuscated: bool,
+        #[cfg(feature = "stats-obfuscation")] supported_obfuscation_version: &'static str,
+    ) -> anyhow::Result<()>;
+}
```
paullegranddc, line 322: I don't really get why we need to separate just agentless vs agent/agentless with a trait. Does it allow reducing package size?

paullegranddc (commented):
Approved with some nits. Overall I have a gut feeling that it shouldn't be this hard to do what you are doing.

Also, you're basically rebuilding a small TracerExporter which target just agentless. Instead of maintaining the same pipeline twice we should see if we can refactor the TraceExporter even if we have to feature gate some code for package size

paullegranddc (approved):

libdd-trace-stats/src/span_concentrator/mod.rs
```diff
@@ -23,6 +24,10 @@ pub use stat_span::{ChunkSpanView, StatSpan};
 
 const ADDITIONAL_METRIC_TAGS_MAX_KEYS: usize = 4;
 
+/// Span kinds that are eligible for stats computation without Agent configuration.
+pub const DEFAULT_STATS_ELIGIBLE_SPAN_KINDS: [&str; 4] =
```
danielsn, line 28: should these be enums?

danielsn (approved):

libdd-data-pipeline-core/src/agentless/v04.rs
```diff
@@ -0,0 +1,498 @@
...
+) -> Result<Option<AgentlessStatsExporter<C>>, AgentlessV04Error>
+where
+    C: HttpClientCapability + SleepCapability,
+{
+    let _ = (capabilities, metadata, trace_config);
+    match stats_config {
+        Some(_) => Err(AgentlessV04Error::StatsObfuscationDisabled),
+        None => Ok(None),
+    }
+}
```
paullegranddc: Should we even have this function? It should be fine to not compile if the feature is not enabled

paullegranddc (approved):
LGTM, just a nit about leaving APIs that are going to be unusable anyway when not enabling certain features.
If it's more practical for you to leave it this way feel free to ignore it though

libdd-trace-stats/src/span_concentrator/mod.rs
paullegranddc: We receive the list of eligible span kinds as opaque strings from the agent, and they are compared to the "span.kind" attribute (in the dd span model). These are only used as defaults if we cannot retrieve them.

Having an enum means we would loose forward compatibility with the agent adding new span kinds

libdd-data-pipeline-core/src/agentless/v04.rs
BridgeAR: Agreed, I removed that

bengl (approved):

PR #2490, opened 2026-09-09
feat(data-pipeline)!: refactor agent's /info obfuscation config format

Eldolfin:
# What does this PR do?
Parse the obfuscation config from the agent `/info` endpoint with a new, more complete format.
It now includes the missing parts, these are not used in stats computation yet it will come in a follow up PR.

Agent side PR: https://github.com/DataDog/datadog-agent/pull/56014

# Motivation
2 main motivation:

`Client side stats obfuscates resources without using the agent config. This is a breaking change for customers enabling CSS when they have a custom obfuscation config`.
See https://docs.google.com/document/d/1i-CQfkF-5B_8vLYepIJ6A16aRISH-Px7XIBSq0HKMaY/edit?tab=t.0#heading=h.i4txi29o2zao

We might want to obfuscate span derived primary tags in client side stats in the future.

# Additional Notes

# How to test the change?
I tried the `fetch_info` function manually using a released agent version and with the version in the [mentioned PR](ttps://github.com/DataDog/datadog-agent/pull/56014).

### DIFF
--- a/libdd-data-pipeline/src/agent_info/schema.rs
+++ b/libdd-data-pipeline/src/agent_info/schema.rs
@@ -97,7 +97,7 @@ pub struct ObfuscationConfig {
     pub elasticsearch: libdd_trace_obfuscation::json::JsonObfuscator,
     pub opensearch: libdd_trace_obfuscation::json::JsonObfuscator,
     pub mongodb: libdd_trace_obfuscation::json::JsonObfuscator,
-    pub tag_replace_rules: Option<Vec<ReplaceRule>>,
+    pub tag_replace_rules: Vec<ReplaceRule>,
 }
 
 impl AgentInfo {
--- a/libdd-data-pipeline/src/agent_info/schema.rs
+++ b/libdd-data-pipeline/src/agent_info/schema.rs
@@ -1,6 +1,11 @@
 // Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 //! This module provides struct representing the info endpoint response
+use libdd_trace_obfuscation::{
+    json::JsonObfuscator,
+    obfuscation_config::{self, SqlObfuscationMode},
+    replacer::ReplaceRule,
+};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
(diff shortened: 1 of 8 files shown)


libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -265,13 +265,15 @@ fn update_obfuscation_config(
...
-                .as_ref()?
-                .sql_obfuscation_mode
+            Some(
+                agent_info
+                    .info
+                    .config
+                    .as_ref()?
+                    .obfuscation
+                    .as_ref()?
+                    .sql_obfuscation_mode,
```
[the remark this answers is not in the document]
Eldolfin: We won't break existing fields in /info to the "legacy" sql_obfuscation_mode is there to stay

libdd-trace-obfuscation/src/obfuscation_config.rs
```diff
@@ -76,6 +78,16 @@ impl Default for JsonObfuscatorConfig {
     }
 }
 
+impl PartialEq for JsonObfuscatorConfig {
+    fn eq(&self, other: &Self) -> bool {
+        self.enabled == other.enabled
+            && self.keep_keys == other.keep_keys
+            && self.transform_keys == other.transform_keys
+            && self.transformer.is_none()
+            && other.transformer.is_none()
```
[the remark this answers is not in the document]
Eldolfin: This is intentional because
> comparing function pointers is unreliable: pointers to the same function can compare inequal

https://doc.rust-lang.org/std/primitive.fn.html

So let's just disallow transformers to be present to make it more intuitive.
Also obfuscation config comparison is only used in the agent info fetcher and /info never populate transformers anyway.

libdd-data-pipeline/src/agent_info/schema.rs
```diff
@@ -82,45 +83,21 @@ pub struct Config {
...
 pub struct ObfuscationConfig {
-    pub elastic_search: bool,
-    pub mongo: bool,
-    pub sql_exec_plan: bool,
-    pub sql_exec_plan_normalize: bool,
-    #[cfg(feature = "stats-obfuscation")]
-    // Option because it might not exist with old agents
-    pub sql_obfuscation_mode: Option<libdd_trace_obfuscation::sql::SqlObfuscationMode>,
-    pub http: HttpObfuscationConfig,
+    // Old format from the agent, now present under sql->obfuscation_mode directly
```
yannham: What does `sql->obfuscation_mode` mean in this context? Is this a go struct field or something?

libdd-data-pipeline/src/agent_info/schema.rs
```diff
@@ -133,6 +110,31 @@ impl AgentInfo {
...
+impl From<ObfuscationConfig> for libdd_trace_obfuscation::obfuscation_config::ObfuscationConfig {
+    fn from(value: ObfuscationConfig) -> Self {
+        let sql_config = match value.sql {
+            Some(sql_config) => sql_config,
+            // Fallback for the previous /info config format
+            None => obfuscation_config::SqlConfig {
+                obfuscation_mode: value.sql_obfuscation_mode,
+                ..Default::default()
+            },
+        };
```
yannham: Nit: use `unwrap_or_else(|| )` instead.

libdd-data-pipeline/src/agent_info/schema.rs
```diff
@@ -82,45 +83,21 @@ pub struct Config {
...
+    pub sql: Option<obfuscation_config::SqlConfig>,
+    pub http: obfuscation_config::HttpConfig,
+    pub redis: obfuscation_config::RedisConfig,
+    pub valkey: obfuscation_config::RedisConfig,
+    pub credit_cards: obfuscation_config::CreditCardConfig,
+    pub memcached: obfuscation_config::MemcachedConfig,
+    pub elasticsearch: libdd_trace_obfuscation::json::JsonObfuscator,
+    pub opensearch: libdd_trace_obfuscation::json::JsonObfuscator,
+    pub mongodb: libdd_trace_obfuscation::json::JsonObfuscator,
+    pub tag_replace_rules: Vec<ReplaceRule>,
```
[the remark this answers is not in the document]
Eldolfin: fixed in a03920e0d2a5

ichinaski: I am unsure how the agents serializes this field when no rule is defined, would the decoder end properly handle both `null` and `[]` from that `/info` response? I suspect the response will contain `null` per Go's default marshaling, but I'm not 100% sure about this. Worth checking and adding a test for it.

ichinaski (commented):
Consider adding test coverage specially for edge cases

libdd-data-pipeline/src/agent_info/schema.rs
Eldolfin: Good catch, you're right that it did indeed not work, I added a test for it and fixed it in 3ac1f553f07a

libdd-data-pipeline/src/agent_info/schema.rs
```diff
@@ -82,45 +83,21 @@ pub struct Config {
 
 #[allow(missing_docs)]
 #[derive(Clone, Serialize, Deserialize, Default, Debug, PartialEq)]
+#[serde(default)]
 pub struct ObfuscationConfig {
```
VianneyRuhlmann: Maybe add something like ObfuscationConfigRaw or ObfuscationConfigPayload to avoid confusion with the `obfuscation_config::ObfuscationConfig`

VianneyRuhlmann (approved):

libdd-data-pipeline/src/agent_info/schema.rs
Eldolfin: Fixed in 164a508aff5b

yannham (approved):

libdd-data-pipeline/src/agent_info/schema.rs
Eldolfin: It's meant to be the `obfuscation_mode` under the `SqlConfig` struct of the `sql` field. Instead of accessing it with `cfg.sql_obfuscation_mode` it's now also duplicated in `cfg.sql.obfuscation_mode`.

Would it me clearer if I replaced the arrow with a dot ?

ichinaski (approved):

PR #2497, opened 2026-09-09
feat(profiling)!: expose missing API in CXX

danielsn:
# What does this PR do?

Updates the CXX API to expose APIs for 
* dictionary-backed samples
* timestamped samples
* split profile uploading 

# Motivation

I would like to use this API for Python, but currently we're missing this functionality so we have to use the old C api.

# Additional Notes

# How to test the change?

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -63,7 +63,17 @@ Iterate fastest with `cargo check -p <crate>` while editing; validate each affec
 
 ### Code exploration
 
-When searching for code with `grep` or `find`, always exclude the `./target` directory. Only search in it if specifically looking for build artifacts
+When searching for code with `grep` or `find`, always exclude the `./target` directory. Only search in it if specifically looking for build artifacts.
+
+### Output hygiene
+
+Keep tool output concise to preserve conversation context:
+- For noisy builds/tests, redirect output to a log file and print only a short pass/fail summary.
+- On failure, show the relevant error excerpt or a short tail, not full logs or long skipped-test lists.
+- Prefer quiet flags such as `-q`, `--success-output never`, or equivalent when available.
+- Avoid dumping full diffs unless explicitly requested; use `git diff --stat`, `git diff --name-only`, or targeted snippets instead.
+- Avoid inspecting generated artifacts under `target/` unless validating generated output specifically.
+- During iteration, prefer narrow checks such as `cargo check -p <crate>`; batch slower tests, examples, and linting once the intended API/code shape has settled.
 
 ## Key Conventions
 
--- a/examples/cxx/README.md
+++ b/examples/cxx/README.md
@@ -97,6 +97,36 @@ By default, the example saves the profile to `profile.pprof`. To export to Datad
 
 See [`profiling.cpp`](profiling.cpp) for a complete example showing profile creation, sample collection, and exporting to Datadog with optional attachments and metadata.
 
+### Profiling api2 (`profiling_api2.cpp`)
+
+Demonstrates building profiling data with the dictionary-backed api2 CXX API.
+This is closer to the low-overhead profiler hot path used by language tracers:
+strings, functions, and mappings are inserted into a `ProfilesDictionary`, and
+samples carry opaque ids instead of string payloads for locations and label keys.
+
+**Build and run:**
+
+Unix (Linux/macOS):
+```bash
+./build-profiling-api2.sh
+```
+
+Windows:
+```powershell
+.\build-profiling-api2.ps1
+```
+
+**Key features:**
+- `ProfilesDictionary` creation and dictionary string insertion
+- Dictionary-backed `Mapping2`, `Function2`, `Location2`, `Label2`, and `Sample2`
+- `Profile::create_with_dictionary`
+- `Profile::add_sample2`
+- Timestamped samples via nonzero `endtime_ns`
+- Non-timestamped samples via `endtime_ns == 0`
+- Pprof serialization to `profile_api2.pprof`
+
+See [`profiling_api2.cpp`](profiling_api2.cpp) for a focused dictionary-backed api2 sample creation and serialization example.
+
 **Requirements:**
 - C++20 compiler
 - For agent mode: Datadog agent running (default: localhost:8126)
@@ -112,6 +142,7 @@ The examples use a consolidated build system:
 Convenience wrappers are provided for each example:
 - `build-and-run-crashinfo.sh` / `build-and-run-crashinfo.ps1`
 - `build-profiling.sh` / `build-profiling.ps1`
+- `build-profiling-api2.sh` / `build-profiling-api2.ps1`
 
 ## Requirements
 
--- a/examples/cxx/README.md
+++ b/examples/cxx/README.md
@@ -119,8 +119,8 @@ Windows:
 **Key features:**
 - `ProfilesDictionary` creation and dictionary string insertion
 - Dictionary-backed `Mapping2`, `Function2`, `Location2`, `Label2`, and `Sample2`
-- `Profile::create_with_dictionary`
-- `Profile::add_sample2`
+- `Profile::create_with_dictionary_or_throw`
+- `Status` / typed-result hot-path calls such as `ProfilesDictionary::insert_string` and `Profile::add_sample2`
 - Timestamped samples via nonzero `endtime_ns`
 - Non-timestamped samples via `endtime_ns == 0`
 - Pprof serialization to `profile_api2.pprof`
--- a/examples/cxx/build-profiling-api2.sh
+++ b/examples/cxx/build-profiling-api2.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+# SPDX-License-Identifier: Apache-2.0
+
+# Build and run the CXX profiling api2 example
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+exec "$SCRIPT_DIR/build-and-run.sh" libdd-profiling profiling_api2
--- a/libdd-profiling/benches/add_samples.rs
+++ b/libdd-profiling/benches/add_samples.rs
@@ -203,7 +203,7 @@ pub fn bench_add_sample_vs_add2(c: &mut Criterion) {
                 let labels_iter = [Ok(api2::Label::num(thread_id_key, thread_id, ""))].into_iter();
                 // SAFETY: all ids come from the profile's dictionary.
                 black_box(unsafe {
-                    profile.try_add_sample2(&locations, &values, labels_iter, None)
+                    profile.try_add_sample2(locations.iter().copied(), &values, labels_iter, None)
                 })
                 .unwrap();
             }
--- a/libdd-profiling/src/cxx.rs
+++ b/libdd-profiling/src/cxx.rs
@@ -260,8 +260,52 @@ pub mod ffi {
         data: &'a [u8],
     }
 
+    /// Explicit no-exception status for CXX callers.
+    ///
+    /// These status-returning APIs are intended for runtimes that cannot allow
+    /// Rust errors to cross the CXX bridge as C++ exceptions. Success uses an
+    /// empty message; failure carries a human-readable error string.
+    ///
+    /// #[must_use] catches ignored statuses in Rust tests/helpers, but cxx does
+    /// not currently translate it to C++ [[nodiscard]] in generated headers.
+    #[must_use]
+    struct Status {
+        ok: bool,
+        message: String,
+    }
+
+    /// No-exception result for ProfilesDictionary::insert_string.
+    ///
+    /// If status.ok is false, value is the null/default string id.
+    #[must_use]
+    struct StringId2Result {
+        status: Status,
+        value: StringId2,
+    }
+
+    /// No-exception result for ProfilesDictionary::insert_function.
+    ///
+    /// If status.ok is false, value is the null/default function id.
+    #[must_use]
+    struct FunctionId2Result {
+        status: Status,
+        value: FunctionId2,
+    }
+
+    /// No-exception result for ProfilesDictionary::insert_mapping.
+    ///
+    /// If status.ok is false, value is the null/default mapping id.
+    #[must_use]
+    struct MappingId2Result {
+        status: Status,
+        value: MappingId2,
+    }
+
     // Opaque Rust types
     extern "Rust" {
+        // CXX exposes Rust Result<T> errors as C++ exceptions. Keep any Result-returning
+        // C++ API explicitly named *_or_throw; hot-path mutation APIs below return Status
+        // or typed result structs instead.
         type Profile;
         type ProfileExporter;
         type EncodedProfile;
@@ -280,21 +324,24 @@ pub mod ffi {
 
         // Static factory methods for Profile
         #[Self = "Profile"]
+        #[cxx_name = "create_or_throw"]
         fn create(sample_types: Vec<SampleType>, period: &Period) -> Result<Box<Profile>>;
 
         /// Create a profile without a sampling period.
         #[Self = "Profile"]
+        #[cxx_name = "create_no_period_or_throw"]
         fn create_no_period(sample_types: Vec<SampleType>) -> Result<Box<Profile>>;
 
         // Static factory methods for ProfilesDictionary
         #[Self = "ProfilesDictionary"]
+        #[cxx_name = "create_or_throw"]
         fn create() -> Result<Box<ProfilesDictionary>>;
 
         /// Inserts value into this dictionary and returns an opaque id.
         ///
         /// The returned id must only be used with this dictionary or Profiles
         /// created from it.
-        fn insert_string(self: &ProfilesDictionary, value: &str) -> Result<StringId2>;
+        fn insert_string(self: &ProfilesDictionary, value: &str) -> StringId2Result;
 
         /// Inserts function into this dictionary and returns an opaque id.
         ///
@@ -302,14 +349,14 @@ pub mod ffi {
         /// in function must have been produced by this dictionary. The returned
         /// id must only be used with this dictionary or Profiles created from
         /// it.
-        fn insert_function(self: &ProfilesDictionary, function: &Function2) -> Result<FunctionId2>;
+        fn insert_function(self: &ProfilesDictionary, function: &Function2) -> FunctionId2Result;
 
         /// Inserts mapping into this dictionary and returns an opaque id.
         ///
         /// Null/default ids in mapping represent empty strings. Non-null ids in
         /// mapping must have been produced by this dictionary. The returned id
         /// must only be used with this dictionary or Profiles created from it.
-        fn insert_mapping(self: &ProfilesDictionary, mapping: &Mapping2) -> Result<MappingId2>;
+        fn insert_mapping(self: &ProfilesDictionary, mapping: &Mapping2) -> MappingId2Result;
 
         /// Creates a Profile backed by dictionary.
         ///
@@ -318,6 +365,7 @@ pub mod ffi {
         /// or unknown values, but all non-null ids must be produced by this
         /// same dictionary.
         #[Self = "Profile"]
+        #[cxx_name = "create_with_dictionary_or_throw"]
         fn create_with_dictionary(
             sample_types: Vec<SampleType>,
             period: &Period,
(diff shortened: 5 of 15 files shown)


libdd-profiling/src/cxx.rs
```diff
@@ -137,6 +151,105 @@ pub mod ffi {
...
+    ///
+    /// String ids may be null/default to represent empty strings. Non-null ids
+    /// must come from the same ProfilesDictionary receiving the insertion.
+    struct Function2 {
+        name: StringId2,
+        system_name: StringId2,
+        // NOTE: api2's Rust/C FFI datatype uses file_name, while pprof and
+        // the string-based CXX Function use filename. Keep file_name for now
+        // to match api2; revisit the CXX-facing name during PR review.
+        file_name: StringId2,
```
danielsn, line 200: Anyone have any thoughts here?

libdd-profiling/benches/add_samples.rs
```diff
@@ -203,7 +203,7 @@ pub fn bench_add_sample_vs_add2(c: &mut Criterion) {
                 let labels_iter = [Ok(api2::Label::num(thread_id_key, thread_id, ""))].into_iter();
                 // SAFETY: all ids come from the profile's dictionary.
                 black_box(unsafe {
-                    profile.try_add_sample2(&locations, &values, labels_iter, None)
+                    profile.try_add_sample2(locations.iter().copied(), &values, labels_iter, None)
```
morrisonlevi, line 206: Why did this have to change, by the way? I assume it's because `&[Location]` doesn't implement an iterator, but rather `IntoIterator`?

ekump (approved):
LGTM for the AGENTS.md change

AGENTS.md
taegyunkim, line 1: I'm not sure why this file needs to be updated as a part of this PR.

libdd-profiling/src/cxx.rs
```diff
@@ -162,22 +324,80 @@ pub mod ffi {
...
+
+        /// Adds a dictionary-backed sample.
+        ///
+        /// Null/default ids in sample represent empty or unknown values. All
+        /// non-null ids in sample must have been produced by the
+        /// ProfilesDictionary used to create this Profile. The sample slices are
+        /// borrowed only for the duration of this call. endtime_ns is an
+        /// optional end timestamp in nanoseconds; pass 0 to record the sample
+        /// without a timestamp.
+        fn add_sample2(self: &mut Profile, sample: &Sample2, endtime_ns: i64) -> Status;
```
taegyunkim, line 391: Since libdd_profiling::cxx::ffi is not intended for Rust callers, please make the bridge module and its Rust implementation surface crate-private. The current public visibility accidentally exposes an unsound safe API to downstream Rust, as demonstrated by this reproducer (https://gist.github.com/taegyunkim/8d0b2ff9fe8e95262299fd7ad40c3a1f), which uses `#![forbid(unsafe_code)]` but still segfaults through `add_sample2`.

examples/cxx/build-profiling-api2.sh
taegyunkim, line 1: Can this also be tested in CI by adding it to .github/workflows/test-ffi.yml?

examples/cxx/build-profiling-dictionary.sh
danielsn: done

libdd-profiling/benches/add_samples.rs
danielsn: The CXX binding does a map on an iterator, and this avoids us needing to collect that into a second vector just to get the slicd.

libdd-profiling/src/cxx.rs
morrisonlevi: I think "file name" and "function name" and such are multiple words, "filename" sometimes is compounded but I'd probably match api2.

libdd-profiling/src/cxx.rs
danielsn: I added "unsafe" in more places, but can also make the apis private

libdd-profiling/src/cxx.rs
```diff
@@ -145,37 +249,238 @@ pub mod ffi {
...
+    ///
+    /// These status-returning APIs are intended for runtimes that cannot allow
+    /// Rust errors to cross the CXX bridge as C++ exceptions. Success uses an
+    /// empty message; failure carries the operation and a human-readable error
+    /// string. Prefer ok() or check_and_*() methods over inspecting fields
+    /// directly.
+    ///
+    /// #[must_use] catches ignored statuses in Rust tests/helpers, but cxx does
+    /// not currently translate it to C++ [[nodiscard]] in generated headers.
+    #[must_use]
```
taegyunkim: The documentation says `#[must_use]` is intended to catch ignored `Status` values, but the attribute is currently attached to `Operation`. As written, a Rust caller can still ignore a returned `Status` without a compiler warning. Please move `#[must_use]` to the `Status` struct.

examples/cxx/README.md
```diff
@@ -27,7 +29,7 @@ Windows:
 - Support for stack traces, frames, and metadata
 - Process and OS information
 - Automatic memory management
-- Exception-based error handling
+- Explicit error handling
```
KowalskiThomas: Is that right for _all_ CXX bindings or only profiling?

libdd-profiling/include/datadog/profiling.hpp
```diff
@@ -0,0 +1,84 @@
...
+        .labels = slice(labels),
+    };
+}
+
+}  // namespace datadog::profiling::views
+
+namespace datadog::profiling::strings {
+
+// Convert a std::string_view to rust::Str. Throws std::invalid_argument if
+// the input is not valid UTF-8.
```
KowalskiThomas: Can we also not have exceptions here?

libdd-profiling/src/cxx/cancellation.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// This is useful for multi-threaded scenarios where one thread performs the
+    /// send operation while another thread can cancel it.
+    #[allow(clippy::should_implement_trait)]
+    pub fn clone(&self) -> Box<CancellationToken> {
+        Box::new(CancellationToken {
+            inner: self.inner.clone(),
+        })
+    }
+
+    /// Cancels the token.
```
KowalskiThomas: **nit** I'm not sure this is cancelling _the token_

libdd-profiling/src/cxx/cancellation.rs
```diff
@@ -0,0 +1,38 @@
...
+    ///
+    /// Note that cancellation is a terminal state; calling cancel multiple times
+    /// has no additional effect.
```
KowalskiThomas:
```suggestion
    /// has no additional effect and is safe.
```

libdd-profiling/src/cxx/profile_exporter.rs
```diff
@@ -0,0 +1,262 @@
...
+                    endpoint,
+                    timeout_ms,
+                    use_system_resolver,
+                ))
+            })(),
+        )
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn create_agentless_exporter(
```
KowalskiThomas, line 143: Wow we support agentless in profiling? 🫨

libdd-profiling/src/cxx/profile.rs
```diff
@@ -0,0 +1,310 @@
...
+        let label_strs: Vec<_> = sample
+            .labels
+            .iter()
+            .map(|l| String::from_utf8_lossy(l.str_bytes))
+            .collect();
+        let label_units: Vec<_> = sample
+            .labels
+            .iter()
+            .map(|l| String::from_utf8_lossy(l.num_unit))
+            .collect();
```
taegyunkim: This adds two staging `Vec` allocations to every dictionary-backed sample, even when every label is already valid UTF-8. `add_dictionary_sample` is the profiler hot path, and `num_unit` is usually empty, so this can add two heap allocations per recorded event solely for conversion lifetimes.

Can we avoid unconditional staging, for example by adding a byte-aware internal label path or at least using a valid-UTF-8 fast path and one combined backing allocation? Please also benchmark this path against the previous implementation.

libdd-profiling/src/cxx/dictionary.rs
```diff
@@ -0,0 +1,130 @@
...
+            Operation::InternDictionaryString,
+            out,
+            self.inner.try_insert_str2(value).map(Into::into),
+        )
+    }
+
+    pub fn intern_string_lossy(&self, value: &[u8], out: &mut ffi::DictionaryStringId) -> bool {
+        // Returns Cow::Borrowed (zero-copy) when valid UTF-8, Cow::Owned (one
+        // allocation with U+FFFD replacements) only when invalid.
+        let s = String::from_utf8_lossy(value);
```
taegyunkim, line 90: `String::from_utf8_lossy` performs an infallible allocation for invalid input, so allocation failure aborts the host process instead of being returned through this bool/error API. The previous C FFI conversion deliberately used the fallible `try_from_utf8_lossy` helper in `libdd-profiling-ffi/src/profiles/utf8.rs`.

Please preserve that behavior by moving or sharing a fallible lossy-conversion helper in `libdd-profiling`, and use it here and in the dictionary-label conversion path.

libdd-profiling/src/cxx.rs
```diff
@@ -130,9 +138,107 @@ pub mod ffi {
...
+    /// Dictionary-backed label.
+    ///
+    /// key may be null/default to represent an empty key, though callers should
+    /// normally use a meaningful dictionary string. Non-null key ids must come
+    /// from the same ProfileDictionary used to create the Profile receiving
+    /// this label. str and num_unit are borrowed only for the duration of the
+    /// add_dictionary_sample call.
+    struct DictionaryLabel<'a> {
+        key: DictionaryStringId,
+        /// Arbitrary bytes; the Rust side applies lossy UTF-8 conversion.
```
taegyunkim, line 227: **Blocking: the CXX dictionary example was not migrated to these renamed fields.**

`examples/cxx/profiling_dictionary.cpp` still initializes `DictionaryLabel` with `.str` and `.num_unit` strings. Running `./examples/cxx/build-profiling-dictionary.sh` now fails because `DictionaryLabel` no longer has a `str` member.

Please update the example to use `str_bytes` and byte-slice `num_unit` values, and rerun the CXX example build.

morrisonlevi: I was about to report the same. Do we not run this example in CI anywhere?

libdd-profiling/src/cxx.rs
```diff
@@ -166,8 +285,61 @@ pub mod ffi {
...
+        /// The Profile keeps dictionary storage alive internally. Future
+        /// add_sample2 calls on the Profile may use null/default ids for empty
+        /// or unknown values, but all non-null ids must be produced by this
+        /// same dictionary.
+        #[Self = "Profile"]
+        fn create_with_dictionary(
+            sample_types: Vec<SampleType>,
+            period: &Period,
+            dictionary: &ProfilesDictionary,
+        ) -> Result<Box<Profile>>;
```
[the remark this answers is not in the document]
danielsn: there isn't currently such a consumer, so following YAGNI

libdd-profiling/src/cxx.rs
danielsn: done

libdd-profiling/src/cxx.rs
danielsn: fixed

libdd-profiling/include/datadog/profiling.hpp
danielsn: done

libdd-profiling/src/cxx/profile_exporter.rs
danielsn: kinda.  We have code paths for it but its not a supported thing in prod

libdd-profiling/src/cxx.rs
danielsn: We do.  For some reason github actions are not running on this PR

morrisonlevi: Try resolving the conflicts from main, and see if they'll run on that new commit.

examples/cxx/README.md
danielsn: fixed

libdd-profiling/src/cxx.rs
danielsn: that fixed it thanks

libdd-profiling/src/cxx/dictionary.rs
danielsn: Going to keep this for a future improvement, the PR is getting large already

danielsn: https://github.com/DataDog/libdatadog/issues/2550

libdd-profiling/src/cxx/profile.rs
danielsn: fixed using `Cow`

libdd-profiling/src/cxx/cancellation.rs
danielsn: fixed

libdd-profiling/src/cxx/cancellation.rs
danielsn: fixed

libdd-profiling/src/cxx/cancellation.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
```
gyuheon0h, line 1: Why do some new files have 2024 while other new files have 2026?

gyuheon0h (approved):
I find no glaring issues

libdd-profiling/src/api2.rs
```diff
@@ -22,40 +23,40 @@ pub struct Location2 {
...
 }
 
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Clone, Debug, Default)]
 pub struct Label<'a> {
     pub key: StringId2,
 
     /// At most one of `.str` and `.num` should not be empty.
-    pub str: &'a str,
+    pub str: Cow<'a, str>,
```
taegyunkim: **Blocking API-scope issue:** this changes the existing public Rust `api2::Label` from `Copy` with `&str` fields to non-`Copy` with `Cow<str>` fields. Downstream initializers such as `Label { str: "value", ... }`, code relying on `Copy`, and function signatures using the old shape will stop compiling, even though this PR is adding a C++ bridge rather than changing the Rust profiling API.\n\nPlease keep `api2::Label` unchanged and add a bridge-private Cow-backed label path for the lossy CXX conversion. The adjacent public `StringTable::try_intern(&str)` change to a generic `impl AsRef<str>` also appears unnecessary: the private Cow path can call `cow.as_ref()`. That avoids both public API changes while retaining the no-allocation valid-UTF-8 path.

libdd-profiling/src/internal/profile/mod.rs
```diff
@@ -192,16 +192,17 @@ impl Profile {
     ///
     /// All MappingId2, FunctionId2, and StringId2 values should be coming
     /// from the same profiles dictionary used by this profile internally.
-    pub unsafe fn try_add_sample2<
-        'a,
-        L: ExactSizeIterator<Item = anyhow::Result<api2::Label<'a>>>,
-    >(
+    pub unsafe fn try_add_sample2<'a, Locations, Labels>(
```
taegyunkim: This is another source-breaking change to the existing public Rust path: callers that pass `&[Location2]` to `try_add_sample2` no longer compile because a slice reference is not an `Iterator`. Please preserve the current public signature and delegate to a new crate-private iterator-based implementation for the CXX zero-staging path.

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
 channel = "1.87.0"
-components = ["rustfmt", "clippy"]
+components = ["rustfmt", "clippy", "rust-analyzer"]
```
taegyunkim: This adds `rust-analyzer` to every pinned toolchain installation but is unrelated to the CXX or Cow changes in this PR. Please remove it from this branch along with the unrelated `AGENTS.md` edit.

libdd-profiling/src/cxx.rs
```diff
@@ -1,13 +1,21 @@
 // Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
```
taegyunkim: **Blocking panic-containment issue:** these CXX exports bypass libdatadog's release-artifact `catch_panic` policy. `libdd-profiling` has no `catch_panic` feature, and none of the bridge implementations catch unwinds. CXX v1.0.192 wraps generated `extern "Rust"` calls in `cxx::private::prevent_unwind`, whose documented behavior is to turn any panic into a deterministic process abort, not an error return (`cxx-1.0.192/src/unwind.rs:9-20`).

As a result, an unexpected panic in any of these profiling methods aborts the embedding tracer process even when builder's default `catch_panic` feature is enabled. dd-trace-py is especially exposed because it links `libdd-profiling` directly with only the `cxx` feature, so the existing C FFI containment layer is not involved.

Please add feature-aware `catch_unwind` containment inside the CXX entry points and translate panics into the existing `Status`/result/error-policy paths, propagate that feature from builder and direct consumers, and add a release-artifact C++ test that deliberately panics in Rust and verifies the host survives. Calls such as `take_value()` that explicitly abort also need to be considered in the API's containment contract.

rust-toolchain.toml
gyuheon0h, line 3: I've seen agents do this when local rust toolchain doesn't match the workspace version. So it adds it manually. We should probably remove this since this is an unnecessary change

libdd-profiling/src/cxx/errors.rs
```diff
@@ -0,0 +1,266 @@
...
+            pub fn is_ok(&self) -> bool {
+                self.status.ok()
+            }
+
+            #[cfg(test)]
+            #[track_caller]
+            #[allow(dead_code)]
+            pub fn is_err(&self) -> bool {
+                !self.status.ok()
+            }
```
gyuheon0h, line 158: Are these `#[track_caller]` macros necessary? These don't panic, right?

libdd-profiling/src/internal/profile/mod.rs
```diff
@@ -1017,8 +1018,8 @@ impl Profile {
     /// Interns the `str` as a string, returning the id in the string table.
     /// The empty string is guaranteed to have an id of [StringId::ZERO].
     #[inline]
-    fn try_intern(&mut self, item: &str) -> Result<StringId, string_table::Error> {
-        self.strings.try_intern(item)
+    fn try_intern(&mut self, item: impl AsRef<str>) -> Result<StringId, string_table::Error> {
+        self.strings.try_intern(item.as_ref())
```
gyuheon0h, line 1022: is this `as_ref()` necessary if `StringTable::try_intern` takes `impl AsRef<str>`

libdd-profiling/src/cxx.rs
```diff
@@ -132,52 +140,359 @@ pub mod ffi {
...
+        fn is_null(self: &DictionaryMappingId) -> bool;
+
         // CancellationToken factory and methods
-        fn new_cancellation_token() -> Box<CancellationToken>;
-        fn clone_token(self: &CancellationToken) -> Box<CancellationToken>;
+        #[Self = "CancellationToken"]
+        fn create() -> Box<CancellationToken>;
+        fn clone(self: &CancellationToken) -> Box<CancellationToken>;
         fn cancel(self: &CancellationToken);
-        fn is_cancelled(self: &CancellationToken) -> bool;
```
gyuheon0h, line 161: Lets document this public API removal in the PR description especially following the libdatadog guild discussion we had about being more communicative about breaking changes in libdatadog

libdd-profiling/src/cxx/cancellation.rs
danielsn: fixed

libdd-profiling/src/cxx/errors.rs
danielsn: done

libdd-profiling/src/internal/profile/mod.rs
danielsn: removed

rust-toolchain.toml
danielsn: done

AGENTS.md
danielsn: technically it could be seperate, but it was approved by edmund

libdd-profiling/src/cxx.rs
danielsn: done

PR #2498, opened 2026-09-09
chore(benchmarks): remove batched-loop sampling noise

BridgeAR:
Criterion times output collection inside batched loops. The glob matcher and sampler benchmarks used that API with an empty setup, which made harness bookkeeping part of their short measurements.

On Apple Silicon with Rust 1.87.0, ten fresh-process measurements of the worst wildcard case narrowed from a 10.4-19.2 ns non-outlier range to 10.7-12.5 ns. The remaining outlier fell from 80.1 ns to 27.4 ns.

### DIFF
--- a/libdd-sampling/benches/glob_matcher_bench.rs
+++ b/libdd-sampling/benches/glob_matcher_bench.rs
@@ -7,7 +7,7 @@
 use std::alloc::System;
 use std::hint::black_box;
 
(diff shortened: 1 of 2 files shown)


yannham (approved):
LGTM. I don't know why `iter_batched` has been used without any setup code/init value, this doesn't make a lot of sense to me.

PR #2500, opened 2026-09-09
refactor(benchmarks): measure allocator thread CPU time

BridgeAR:
# What does this PR do?

Measure allocator benchmarks with Linux thread CPU time instead of wall time. Retire the old series IDs under a `thread_cpu` prefix and use the no-op allocator for the no-profiler short-circuit benchmark.

# Motivation

Wall time charged host descheduling to nanosecond-scale sampler paths. The short-circuit benchmark also included libc allocator variance instead of isolating the semaphore gate.

Across five Linux ARM64 processes using the CI settings, system allocation medians spanned 6.2%, the no-op fast path spanned 9.7%, and the short-circuit case reported 2 ns every time.

# Additional Notes

Benchmark-only; no production code changes. `libc` is a Linux-only development dependency. The forced-sample path still spanned 12.2%.

# How to test the change?

- `cargo clippy -p libdd-profiling-heap-allocator --all-targets -- -D warnings`
- `cargo test -p libdd-profiling-heap-allocator`
- Run all 30 `sampler_overhead` benchmark cases on Linux.
- `cargo check -p libdd-profiling-heap-allocator --all-targets` on macOS.

### DIFF
--- a/libdd-profiling-heap-allocator/benches/sampler_overhead.rs
+++ b/libdd-profiling-heap-allocator/benches/sampler_overhead.rs
@@ -18,7 +18,9 @@ mod linux_bench {
     use libdd_profiling_heap_sampler::{dd_test_set_profiler_active, dd_tl_state_get_or_init};
     use std::alloc::{GlobalAlloc, Layout, System};
     use std::hint::black_box;
+    use std::mem::MaybeUninit;
     use std::ptr;
+    use std::time::Duration;
 
     const SIZES: &[usize] = &[16, 64, 256, 4096, 65_536];
     const ALIGN: usize = 8;
@@ -27,6 +29,30 @@ mod linux_bench {
     struct AlignedBuffer([u8; 128 * 1024]);
 
     static mut NOOP_BUFFER: AlignedBuffer = AlignedBuffer([0; 128 * 1024]);
+
+    // Do not attribute time when this benchmark thread is descheduled to the sampler.
+    fn thread_cpu_time() -> Duration {
+        let mut time = MaybeUninit::<libc::timespec>::uninit();
+        let result =
+            unsafe { libc::clock_gettime(libc::CLOCK_THREAD_CPUTIME_ID, time.as_mut_ptr()) };
+        assert_eq!(
+            result,
+            0,
+            "read thread CPU time: {}",
+            std::io::Error::last_os_error()
+        );
+        let time = unsafe { time.assume_init() };
+        Duration::new(time.tv_sec as u64, time.tv_nsec as u32)
+    }
+
+    #[inline]
+    fn measure_thread_cpu_time(iterations: u64, mut routine: impl FnMut()) -> Duration {
+        let start = thread_cpu_time();
+        for _ in 0..iterations {
+            routine();
+        }
+        thread_cpu_time() - start
+    }
 
     struct NoopAllocator;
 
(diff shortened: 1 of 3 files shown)


libdd-profiling-heap-allocator/benches/sampler_overhead.rs
```diff
@@ -28,6 +30,30 @@ mod linux_bench {
...
+            result,
+            0,
+            "read thread CPU time: {}",
+            std::io::Error::last_os_error()
+        );
+        let time = unsafe { time.assume_init() };
+        Duration::new(time.tv_sec as u64, time.tv_nsec as u32)
+    }
+
+    #[inline]
```
yannham:
```suggestion
    #[inline(always)]
```

scottgerring (approved):
Seems like a sensible improvement to me! Wondering how this came up - is variance in the benchmark results showing up as PR gate noise?

BridgeAR:
@scottgerring if you look at the benchmark comment and see the reported improvements and regressions and `Unstable benchmarks` there are many listed that are flaky. Next to those, a few come up as improvement or regression when working on independent things. Thus, I had a look at making a few more stable and these were just a first simple improvement.

I also worked on making dd-trace-js benchmarks stable (excluding memory usage, see e.g., https://github.com/DataDog/dd-trace-js/pull/10310#issuecomment-5645550532) and that helped to know immediately when to look at regressions. Having flaky benchmarks otherwise just makes it hard to really rely on the data and knowing when to trust the outcome or not 😅

libdd-profiling-heap-allocator/benches/sampler_overhead.rs
```diff
@@ -28,6 +30,30 @@ mod linux_bench {
...
+        let result =
+            unsafe { libc::clock_gettime(libc::CLOCK_THREAD_CPUTIME_ID, time.as_mut_ptr()) };
+        assert_eq!(
+            result,
+            0,
+            "read thread CPU time: {}",
+            std::io::Error::last_os_error()
+        );
+        let time = unsafe { time.assume_init() };
+        Duration::new(time.tv_sec as u64, time.tv_nsec as u32)
```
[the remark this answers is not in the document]
BridgeAR: Guarding against a broken syscall is not something we should have to do

scottgerring:
> @scottgerring if you look at the benchmark comment and see the reported improvements and regressions and `Unstable benchmarks` there are many listed that are flaky. Next to those, a few come up as improvement or regression when working on independent things. Thus, I had a look at making a few more stable and these were just a first simple improvement.
> 
> I also worked on making dd-trace-js benchmarks stable (excluding memory usage, see e.g., [DataDog/dd-trace-js#10310 (comment)](https://github.com/DataDog/dd-trace-js/pull/10310#issuecomment-5645550532)) and that helped to know immediately when to look at regressions. Having flaky benchmarks otherwise just makes it hard to really rely on the data and knowing when to trust the outcome or not 😅

@BridgeAR sounds good - and appreciate the cleanup! 🤝

yannham (approved):

libdd-profiling-heap-allocator/benches/sampler_overhead.rs
BridgeAR: Could we use a linter for that?

yannham: It's a good point! Maybe we could lint for non-public function with an `inline` annotation. I'll first have to see if other Rust folks agree with me :sweat_smile:

BridgeAR: Thank you :)

ivoanjo (approved):
I took a look at this (since it needed a reviewer from profiling) and it seems reasonable to use cpu-time for this kind of measurement, no concerns.

PR #2503, opened 2026-09-10
refactor(trace)!: extract shared span types

Eldolfin:
# What does this PR do?

Move v0.4 and v1 models into a new crate.

# Motivation

~Normalizing traces before they enter the stats computation pipeline for CSS correctness.~ (This won't be done client-side in the end.)

- Simplify the dependency graph
- Allow crates that are dependencies of `trace-utils` to depend on trace types
- Break down `trace-utils`, currently the largest crate in libdatadog

# Additional Notes

The original approach was to re-export moved types from `libdd-trace-utils` to maintain backward compatibility on the consumers, and have less renaming to do in the repo too.
But I find it cleaner to have no re-export, only one way to import it. And this way we might realize we don't need `libdd-trace-utils` but only its types for some crates.

# How to test the change?
- No logic change, if CI pass it should be enough

[Generated using codex]


libdd-trace-types/src/span/v05/dict.rs
```diff
@@ -51,6 +53,21 @@ impl<T: SpanText> SharedDict<T> {
...
+    /// Gets the index of borrowed span text, inserting an owned copy only when it is absent.
+    pub fn get_or_insert_span_text<T: SpanText>(
+        &mut self,
+        value: &T,
+    ) -> Result<u32, std::num::TryFromIntError> {
+        let entry = self.map.raw_entry_mut_v1().from_key(value.borrow());
+        let index = entry.index();
+        entry.or_insert_with(|| (value.to_bytes_string(), ()));
+        index.try_into()
+    }
```
VianneyRuhlmann: This seems like a good optimization, but we should avoid doing unrelated changes especially in PRs such as this one which changes many crates as it makes it harder to know what requires in-depth review. Also naming is a bit confusing since the other `get_or_insert` also inserts a SpanText .

VianneyRuhlmann (approved):

libdd-trace-types/src/span/v05/dict.rs
Eldolfin: Actually this is not new code, it was [here](https://github.com/DataDog/libdatadog/pull/2503/changes#diff-baa85492b74d89f59d40c91a34b24d3b2759524a0bf0ccbebec5b4633092d274L191) previously but it now needs to be near its type definition because the inner map is private, which I find pretty clean.

I made it clearer with better doc explaining the difference between get_or_insert vs get_or_insert_span_text and renamed it to get_or_insert_from_span_text which is more like a `From` conversion

PR #2506, opened 2026-09-11
feat(release): score dependency floors and cargo features in semver-level

iunanua:
# What does this PR do?

`semver-level.sh` combined cargo-semver-checks and cargo-public-api. Both read rustdoc and neither reads a manifest, so two semver-relevant manifest changes came out as `patch`:

- a **raised dependency requirement floor** (`http = "1"` → `"1.1"`): every rustdoc signature is byte-identical, yet a consumer pinned to the old version can no longer resolve the crate — conventionally a minor;
- an **added cargo feature**: cargo-semver-checks has no lint for an addition at all, and a feature adds no rustdoc item unless it happens to gate one.

Two passes are added, both reading the manifest through `cargo metadata`:

| pass | reports |
| --- | --- |
| 2b) dependency requirement floors | `minor` when the lowest version a requirement admits goes up |
| 2c) cargo feature surface | `minor` on an added feature; `major` on one removed, or dropped from the set `default` reaches |

The second commit is a consequence of the first: `build` moves into the permissive PR-title type list, because scoring a dependency floor as minor would otherwise make `build(deps): bump foo from 1.0 to 1.1` — the conventional title for a dependency change — fail the type rule.

The third commit is what lets CI reach any of it. `detect-changes` selected a crate only when a file under its own directory changed, so a PR editing just the root manifest's `[workspace.dependencies]` selected nothing: `has_rust_changes` came out `false`, both the `semver-check` and `validate` jobs were skipped, and every PR title was accepted — a `docs:` title included. Since this workspace declares dependencies version-only at the root and members inherit them with `workspace = true`, that is the ordinary shape of a floor raise here, which left the pass above unreachable from CI. `semver-level.sh` grows a `--list-affected` mode naming every member whose dependency requirements, dependency feature selections or own feature surface moved between two revisions, and `detect-changes` adds those crates to the ones it finds by path.

The last three commits answer review findings against the manifest readers. None of them changes what any pass scores on this workspace today:

- **Default membership is the closure, not a direct listing.** 2c asked whether `default` *lists* a feature. So a behaviour-preserving refactor — `default = ["foo"]` becoming `default = ["bundle"]` with `bundle = ["foo"]` — read as `foo` losing its default and scored major, while the inverse, emptying a `bundle` that `default` still lists, is a real break that read as no change at all. Rows now carry the closure over the feature graph, which is what a consumer writing `default-features = true` actually gets.
- **Dependency rows now carry the feature selection.** They recorded only the requirement, so a root entry that merely gained a feature, or stopped disabling default features, left every inheriting member's facts identical: `--list-affected` named nobody and, with only the root manifest touched, the jobs were skipped again — the same shape of gap the third commit fixes for floors.
- **Exclusive bounds are normalized, not tagged.** `req_min` appended an exclusivity bit to the version a bound states rather than advancing past it, so `>` was read as admitting the version it excludes. `>1.2` admits nothing below 1.3.0, so `>=1.2.1` → `>1.2` read as a lowering rather than the raise it is; and `>1.2.3` → the equivalent `>=1.2.4` read as a raise. Bounds now advance to the first version they admit, which drops the flag and returns `version_gt` to a plain triple.

# Motivation

Reviewing release proposal #2482 surfaced `libdd-capabilities` proposed as `patch`, when #2350 had moved its `http` requirement from `^1` to the workspace entry at `^1.1`. That is a minor, and nothing in the pipeline could see it: the version had moved *into* `[workspace.dependencies]`, so even a textual diff of the crate's own `Cargo.toml` shows only `{ workspace = true }` with no version in it.

# Additional Notes

Points reviewers may want to weigh:

- **Read through `cargo metadata`, not `Cargo.toml`.** That is what resolves `{ workspace = true }` to the version the root manifest declares, and what surfaces the implicit feature an `optional = true` dependency creates.
- **Dependencies are keyed by `.rename` as well as name/kind/target.** A crate aliasing two versions of one package emits an identical name, kind and target for both; keyed without the alias, the second matches the first's baseline requirement and an unchanged pair reads as a raise.
- **An exclusive bound is advanced to the first version it admits**, `>` excluding everything up to and including the components it states: `>1.2.3` starts at 1.2.4, `>1.2` at 1.3.0 and `>1` at 2.0.0. So `>=1.2.3` → `>1.2.3` reads as the narrowing it is, `>=1.2.1` → `>1.2` as the raise it is, and `>1.2.3` → the equivalent `>=1.2.4` as no change at all. A pre-release bound is the exception and stays where it is: `>1.2.3-rc.1` admits 1.2.3 itself, so the release version this reader already compares is that bound's exact floor, and advancing it would overstate the requirement. `req_min` was checked against `semver` 1.0.28's own `VersionReq::matches` over 33 requirement forms — every operator at one, two and three components, the four shapes this workspace uses, and both pre-release cases — and agrees on all of them.
- **A raised *major* floor is capped at `minor` on purpose.** Whether that forces the dependent to major is `release-version-major-bumps.sh`'s decision, made with the dependency graph this script cannot see; `max_level()` lets it win from there.
- **The two passes stop at different points.** 2b can only report minor, so minor ends it. 2c can report major, so it runs on until major — which matters for a crate with no library target, where cargo-semver-checks is skipped entirely and nothing else watches the feature table.
- **Feature removals are a backstop, not the primary check.** For a library crate `feature_missing` and `feature_not_enabled_by_default` get there first; both were verified to *fail* the run rather than merely warn, using a two-crate repro against the pinned invocation. The closure above is what keeps the backstop from contradicting them: cargo-semver-checks resolves the default set transitively, so keyed on direct membership this pass would have reported major over a refactor the lint passes.
- **Dependency feature facts are selected on, not scored.** Enabling a feature in a dependency can change this crate's own API — an item reached through a glob re-export appears — but the manifest cannot say whether it did, and the passes that read rustdoc can. So the columns widen what `--list-affected` names and assign no level of their own; 2b's key stops before them, so a feature move never reads as a requirement change.
- **Deliberately not scored:** dev-dependencies, a widened requirement, a dependency added or removed outright, and a change to what a feature *enables* for its own sake (the two `*_enables_feature` lints' job) — that last reaches 2c only where it moves the set `default` reaches.

On the detection commit:

- **`--list-affected` lives in `semver-level.sh` rather than a script of its own**, because it reuses `manifest_facts_at_rev` — now able to read the whole workspace in one extraction — so what the detector considers a manifest fact cannot drift from what the passes score. A detector considering *fewer* facts silently selects nothing and reports "no change", which is the failure this commit fixes; there are no shell tests to catch that drift. 19 lines are new, the 51 they lean on are shared.
- **The mode compares from `git merge-base`**, matching the `base...HEAD` the changed-file search uses. Measured from the baseline tip instead, a branch behind the baseline reports every crate that moved *on* the baseline since the fork, putting another branch's changes on this one's report — and since scoring uses the tip, a feature added on the baseline would read as a removal, i.e. a spurious major.
- **Publishability stays the caller's business.** The mode lists workspace members; `detect-changes` keeps its own `publish` filter, so a `publish = false` member such as `libdd-agent-client` shows up in the list and is dropped where that filter already lives.
- **Cost.** A crate selected only by moved facts still gets the full treatment, cargo-semver-checks and cargo-public-api both building rustdoc, so a one-line root bump now costs ~12 of those instead of zero, and a root entry that only gains a feature costs one per inheritor too — 20 for `serde_json`. That is deliberate: a changed floor or a newly enabled dependency feature can alter a crate's API through re-exports. If it bites, the cheap follow-up is a mode running only passes 2b/2c for those crates. Release-proposal PRs are unaffected — they carry `skip-pr-title-semver-check`, and each member's `version` is literal, so the path search already finds them.
- **The release path reports the same change rather than releasing for it.** `commits-since-release.sh` selects a crate's commits with `git log "$COMMIT_RANGE" -- "$CRATE_PATH"`, so a root-only floor raise leaves every inheriting crate with no commits and `release-version-bumps.sh` defers it. That is the right outcome, not a second instance of the bug above: the crate's code is unchanged, the requirement its published version states is still true of that code, and requirements are minimums, so a consumer combining it with a freshly published sibling that asks for the raised floor resolves the newer dependency anyway. Deferring moves neither the version nor the tag, so the raise stays in range and the crate's next release is still scored a minor for it — late, not wrong — while the case where a crate's code does need the raised version always arrives with a change to a file of its own, which the path search already finds. What is missing there is only the saying so: a deferred crate is recorded as level `none` and reads as "nothing happened". A follow-up reports those moves at the deferral point instead of releasing 12 crates that needed nothing.

Selection, each case a root-manifest edit and nothing else:

| root edit | before | after |
| --- | --- | --- |
| `bytes` floor `1.11` → `1.12` (12 members inherit it) | 0, both jobs skipped | exactly the 12 inheritors |
| + a feature added to one crate | that crate, by path | same, plus the 12 |
| `serde_json` gains a feature (20 members inherit it outside dev-dependencies) | 0, both jobs skipped | exactly those 20 |
| `serde_json` stops disabling default features | 0, both jobs skipped | exactly those 20 |
| baseline unchanged | — | none |

Scoring, over the same revisions, is unmoved by any of the three follow-ups: 2b's projection is byte-identical across all 698 dependency rows of the workspace, the 217 feature rows are byte-identical too, and all 149 distinct requirement strings in the workspace keep the floors they had — every bound here being inclusive with a full triple, the normalization has nothing to advance. So the floor and feature passes see exactly what they saw before. Both manifest passes also produce byte-identical output after the field shift in the shared reader (`bytes (normal): ^1.11 -> ^1.12`; `Cargo feature added / added: new-thing`). Scoring one of the 20 selected crates end to end reports `patch` — "No public API changes detected" — rather than a level invented from the new columns.

The closure was checked against cargo-semver-checks 0.47.0 and 0.48.0, the pinned version, on a two-crate repro: `feature_not_enabled_by_default` passes the behaviour-preserving refactor and fails the inverse, exactly as the closure does, and on `libdd-common`'s real feature graph the pass and the lint name the same eight features. The refactor now scores `minor` for the genuinely new feature name rather than `major`.

Everything above runs under `RUSTUP_TOOLCHAIN=1.92.0` as the workflow does. `shellcheck` on the script and on the extracted `run:` block (actionlint's `-e SC2086` profile) reports nothing that was not already there.

### DIFF
--- a/scripts/semver-level.sh
+++ b/scripts/semver-level.sh
@@ -159,6 +159,115 @@ crate_target_kinds_at_rev() {
     echo "$kinds"
 }
 
+# Echo one tab-separated row per manifest fact about crate $1 at revision $2, read once
+# per revision for both manifest passes below:
+#
+#   dep      <name> <alias> <kind> <target> <req>   one per dependency a consumer resolves
+#   feature  <name> default|optional                one per declared feature
+#
+# Read via `cargo metadata` rather than the crate's own Cargo.toml, which would see
+# neither the version behind a `{ workspace = true }` inheritance marker nor the implicit
+# feature an `optional = true` dependency creates.
+#
+# <alias> is `.rename`, defaulting to the package name. It is in the row because a crate
+# may alias two versions of one package (`foo1`/`foo2` both `package = "foo"`), for which
+# cargo metadata reports an identical name, kind and target; keyed without it, the second
+# alias matches the first's baseline requirement and an unchanged pair reads as a raise.
+crate_manifest_facts_at_rev() {
+    local crate=$1 rev=$2
+    local tree meta cargo_status
+    tree=$(mktemp -d) || return 1
+    if ! git archive "$rev" | tar -x -C "$tree"; then
+        echo "Error: could not extract $rev to read the manifest for $crate" >&2
+        rm -rf "$tree"
+        return 1
+    fi
+    # Captured before jq: without pipefail a failing cargo would surface as jq's clean
+    # exit over empty input, and a broken manifest would read as "declares nothing".
+    meta=$(cargo metadata --format-version=1 --no-deps --manifest-path "$tree/Cargo.toml" 2>/dev/null)
+    cargo_status=$?
+    rm -rf "$tree"
+    if [[ $cargo_status -ne 0 || -z "$meta" ]]; then
+        echo "Error: could not read the manifest for $crate at $rev" >&2
+        return 1
+    fi
+    jq -r --arg crate "$crate" '
+        .packages[]
+        | select(.name == $crate)
+        | . as $p
+        | (
+            ( $p.dependencies[]
+              | select((.kind // "normal") != "dev")
+              | ["dep", .name, (.rename // .name), (.kind // "normal"), (.target // "any"), .req] ),
+            ( ($p.features // {}) as $f
+              | ($f.default // []) as $d
+              | ($f | keys[]) as $n
+              | select($n != "default")
+              | ["feature", $n, (if ($d | index($n)) then "default" else "optional" end)] )
+          )
+        | @tsv' <<< "$meta"
+}
+
+# version_gt A B — true when the dotted numeric tuple A is strictly greater than B.
+# Four components: req_min appends an exclusivity flag to the X.Y.Z triple.
+# Compared in the shell rather than with `sort -V`, a GNU extension.
+version_gt() {
+    local -a a=() b=()
+    local i x y
+    IFS='.' read -r -a a <<< "$1"
+    IFS='.' read -r -a b <<< "$2"
+    for i in 0 1 2 3; do
+        x=${a[i]:-0}
+        y=${b[i]:-0}
+        if (( 10#$x > 10#$y )); then
+            return 0
+        elif (( 10#$x < 10#$y )); then
+            return 1
+        fi
+    done
+    return 1
+}
+
+# Echo the lowest version requirement $1 admits, as `X.Y.Z.E` with E=1 when the bound
+# EXCLUDES that version. `>1.2.3` and `>=1.2.3` are different floors, so without E the
+# two compare equal and narrowing one into the other passes as a patch. E is an ordering
+# device only; the report quotes requirements verbatim.
+#
+# Echo nothing when there is no lower bound (`*`, `<2`) or it cannot be parsed: callers
+# treat that as "no opinion", so an exotic requirement can never invent a bump. A
+# pre-release bound is compared as its release version (`1.0.0-rc.1` -> `1.0.0`), which
+# can only understate a raise.
+req_min() {
+    local req=$1
+    local -a parts=() v=()
+    local part bound best="" exclusive
+    read -r -a parts <<< "${req//,/ }"
+    for part in "${parts[@]}"; do
+        exclusive=0
+        case "$part" in
+            ''|'*'|'<'*|'!='*) continue ;;
+            '>='*)  bound=${part#>=} ;;
+            '>'*)   bound=${part#>}; exclusive=1 ;;
+            '^'*)   bound=${part#^} ;;
+            '~'*)   bound=${part#\~} ;;
+            '='*)   bound=${part#=} ;;
+            [0-9]*) bound=$part ;;
+            *)      continue ;;
+        esac
+        bound=${bound%%[-+]*}   # drop pre-release and build metadata
+        bound=${bound//\*/0}    # `1.*` admits 1.0.0
+        [[ "$bound" =~ ^[0-9]+(\.[0-9]+)*$ ]] || continue
+        IFS='.' read -r -a v <<< "$bound"
+        bound="${v[0]:-0}.${v[1]:-0}.${v[2]:-0}.${exclusive}"
+        if [[ -z "$best" ]] || version_gt "$bound" "$best"; then
+            best=$bound
+        fi
+    done
+    if [[ -n "$best" ]]; then
+        printf '%s\n' "$best"
+    fi
+}
+
 compute_semver_results() {
     local crate=$1
     local baseline=$2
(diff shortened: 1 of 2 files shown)


yannham:
I'm starting to look at at this, but there's something I don't yet understand in the description:

> Reviewing release proposal #2482 surfaced libdd-capabilities proposed as patch, when #2350 had moved its http requirement from ^1 to the workspace entry at ^1.1. That is a minor, and nothing in the pipeline could see it: the version had moved into [workspace.dependencies], so even a textual diff of the crate's own Cargo.toml shows only { workspace = true } with no version in it.

Why would raising a dependency from ^1 to ^1.1 be a minor or major change? Even if some dependent pins the version to `1.0` , they will have at worst a duplicated version, but everything should still compile and work fine, right? (unless we re-export some types from this dependency but this is somehow a contrived case).

Also I would expect the semver script to only check for breaking changes. The difference between minor and patch sounds rather subjective (in this case you could fix an urgent security issue by upgrading a dependency 1.0 -> 1.1 and making that a patch release sounds reasonable IMHO). But this one is not only for CI, it is for release? Does it decide if the release should be major or minor automatically, or does it just double check?

iunanua:
> I'm starting to look at at this, but there's something I don't yet understand in the description:
> 
> > Reviewing release proposal #2482 surfaced libdd-capabilities proposed as patch, when #2350 had moved its http requirement from ^1 to the workspace entry at ^1.1. That is a minor, and nothing in the pipeline could see it: the version had moved into [workspace.dependencies], so even a textual diff of the crate's own Cargo.toml shows only { workspace = true } with no version in it.
> 
> Why would raising a dependency from ^1 to ^1.1 be a minor or major change? Even if some dependent pins the version to `1.0` , they will have at worst a duplicated version, but everything should still compile and work fine, right? (unless we re-export some types from this dependency but this is somehow a contrived case).
> 
It is based on an "agreement" regarding the bumps specified [here](https://doc.rust-lang.org/cargo/reference/semver.html#cargo-dep-add)
At the moment, `semver-level.sh` is not able to detect dependency version or feature changes and this PR is trying to address it. With the added difficulty that dependencies may be defined in the workspace root `Cargo.toml`

> Also I would expect the semver script to only check for breaking changes. The difference between minor and patch sounds rather subjective (in this case you could fix an urgent security issue by upgrading a dependency 1.0 -> 1.1 and making that a patch release sounds reasonable IMHO). But this one is not only for CI, it is for release? Does it decide if the release should be major or minor automatically, or does it just double check?

The script decides the level. It checks for breaking changes but if there are no breaking it tries to resolve if the changes correspond to a `minor` or to a `patch`.
As you said, sometimes the level is subjective and in most cases there would be no difference between releasing a `minor` or a `patch` from the consumer point of view.
`semver-level.sh` is used by the CI to check if the PR's title/description is correct (if there is a breaking it must contain a breaking indicator) and also it is used by the release workflow to get the semver level to be applied to a crate.

That said, this PR is kind of optional to increase `semver-level.sh` coverage in order to detect those changes in the dependencies. 
Although I admit that the script tries to solve increasingly complicated cases and that, in the end, it ends up becoming a fiefdom of AI agents.

@yannham

yannham:
Thanks for the detailed answer!

> It is based on an "agreement" regarding the bumps specified [here](https://doc.rust-lang.org/cargo/reference/semver.html#cargo-dep-add)

This section is about adding a new dependency, not bumping an existing one to a new minor version, right? In the end it's not very important as you mention it, but I don't know if I would flag a semver-compatible update as a necessary being at least a minor version bump.

> Although I admit that the script tries to solve increasingly complicated cases and that, in the end, it ends up becoming a fiefdom of AI agents

I probably said it elsewhere already, but... rewrite it in Rust :stuck_out_tongue: ?

scripts/semver-level.sh
```diff
@@ -159,15 +178,218 @@ crate_target_kinds_at_rev() {
...
+# below, which ask for one crate, cut it back off.
+#
+# Read via `cargo metadata` rather than the crate's own Cargo.toml, which would see
+# neither the version behind a `{ workspace = true }` inheritance marker nor the implicit
+# feature an `optional = true` dependency creates.
+#
+# <alias> is `.rename`, defaulting to the package name. It is in the row because a crate
+# may alias two versions of one package (`foo1`/`foo2` both `package = "foo"`), for which
+# cargo metadata reports an identical name, kind and target; keyed without it, the second
+# alias matches the first's baseline requirement and an unchanged pair reads as a raise.
```
yannham: Doc: this documentation is a bit hard to digest

scripts/semver-level.sh
```diff
@@ -159,15 +178,218 @@ crate_target_kinds_at_rev() {
...
+                 # Not `// true`: jq treats a false left-hand side as absent, so that
+                 # spelling would report every dependency as taking the defaults.
+                 (if .uses_default_features == false then "no-defaults" else "defaults" end)] ),
+            ( ($p.features // {}) as $f
+              | default_closure($f) as $d
+              | ($f | keys[]) as $n
+              | select($n != "default")
+              | [$p.name, "feature", $n, (if ($d | index($n)) then "default" else "optional" end)] )
+          )
+        | @tsv' <<< "$meta"
```
yannham, line 278: This starts to be really non trivial logic in jq. I'm not sure it's the right programming environment to compute the transitive closure of a graph... if that helps for the short term be it, but I really think we should rewrite all those scripts in proper language.

yannham (approved):
I can't say I'm very confident in my review, given it's a several hundred lines of complicated bash. But I guess the impact is limited (our own ci/release process) and the overall approach is ok

scripts/semver-level.sh
iunanua: Agree.
At least, I guess, all of this will come in handy for identifying all the cases we need to cover when we decide to rewrite it.

scripts/semver-level.sh
iunanua: I hope it is clearer now

yannham: Thanks, this is already better yes. I think in general the main problem with those AI-generated comments is that they entangle the general description of a function with the motivation/technical specifics of why and how it is used, which makes it really hard to follow. The description of a function should really just, well, describe what the functions does, in a simple and straightforward way. How it is used should be commented at callsite IMHO, or at least in a very high-level way in the function's documentation when that makes sense.

PR #2508, opened 2026-09-11
feat(release): allow overriding a crate's bump level from the dispatch

iunanua:
# What does this PR do?

Adds a `level_overrides` input to the release-proposal dispatch, so a wrong bump level can be corrected by re-running the workflow from the Actions tab instead of checking the proposal branch out and running cargo-release by hand.

```
level_overrides: libdd-capabilities=minor, libdd-ipc=major
```

`release-version-bumps.sh` gains `--level-overrides CRATE=LEVEL,...` and uses the given level for those crates instead of calling `semver-level.sh` — which is also where the run's time goes, so an overridden crate gets cheaper as well as corrected.

It also records the dispatch inputs in an HTML comment at the top of the generated PR body:

```html
<!-- release-proposal-inputs: {"crates":"libdd-capabilities-impl, libdd-ipc","main_start_ref":"","level_overrides":"libdd-capabilities=minor","bypass_standard_checks":false} -->
```

so a follow-up run can reconstruct them without an operator retyping the crate list.

The second commit is unrelated to the feature and can be reviewed on its own: the `cargo-release` job installed its four tools with `taiki-e/cache-cargo-install-action`, which runs `cargo install` and caches the result. On the last dispatch those four steps cost **703s of the job's 3660s**, all cache misses. `pr-title-semver-check` installs two of the same tools with `taiki-e/install-action` in **77s**. Switched to install-action, and pinned all four — `cargo-release` and `git-cliff` were unpinned, which let the tools that choose version numbers and render changelogs drift under the workflow that cuts releases.

# Motivation

Reviewing release proposal #2482 turned up `libdd-capabilities` proposed as `patch` when its `http` requirement had moved from `^1` to `^1.1`. Fixing one crate's level meant a local checkout, a manual `cargo release version`, regenerating that crate's changelog, and a push — for a branch whose whole diff is `Cargo.toml` and `CHANGELOG.md` files.

Overriding the *input* rather than patching the branch is the point: the cascade, the changelogs and the publication order are all recomputed from the corrected level, so a level raised to `major` can pull further crates into the release, which hand-editing the branch would miss.

# Additional Notes

- **Names are validated against the computed release set, not the `crates` input.** The crate you most often need to correct was never named in `crates` — in #2482, `libdd-capabilities`, `libdd-ipc-macros` and `libdd-tinybytes` were all pulled in transitively. A name that is not a candidate fails with the candidate list rather than silently shipping the computed level.
- **`level_overrides` is carried in the inputs block** so successive corrections accumulate instead of the newest one reverting the last. That is a correctness requirement, not a convenience.
- **The block is `-->`-proof.** `>` is escaped to the JSON sequence `\u003e`, which decodes to the same character for a parser but can no longer terminate the HTML comment. `main_start_ref` is the one value with no charset validation upstream.
- **An override on the deferred or initial-release path warns** rather than passing unnoticed, since it cannot apply there.
- **A raised *major* floor is still capped by `semver-level.sh`**; this input can set `major` explicitly, and the row records the *requested* level so the PR body can show that a level was set by hand even if the cascade later raises it.

⚠️ **Not yet usable on an open proposal.** `check-proposal-ongoing` rejects a dispatch while any proposal branch exists — including the branches of the proposal being corrected — so today this works on a fresh proposal, or after cancelling the open one by hand. A follow-up branch makes the guard permit replacing a proposal it can prove is the same release, and retires it once the replacement exists. Worth knowing when judging whether this is complete.

Also not covered by tests: `release-version-bumps.bats` exists on a separate branch (#2507) and has no case for the override path yet.

# How to test the change?

```bash
shellcheck -e SC2086 --severity=warning scripts/*.sh
```

The validation runs before any `cargo release`, so it can be exercised directly against a hand-written candidate file:

| `--level-overrides` | result |
| --- | --- |
| `libdd-alpha` | ERROR: not `CRATE=LEVEL` |
| `libdd-alpha=minr` | ERROR: must be major, minor or patch |
| `=minor` | ERROR: has no crate name |
| `nope=minor` | ERROR + lists the candidates |
| `libdd-alpha=minor` on a deferred crate | WARNING, run continues |
| `" a=minor ,  b=major "` | both parsed |
| omitted | behaviour unchanged |

The PR-body block was rendered through the real jq filter, including a hostile `main_start_ref` of `ref--> <!-- evil`, and round-trips back through `jq` with exactly one `-->` on the line.

### DIFF
--- a/.github/workflows/release-proposal-dispatch.yml
+++ b/.github/workflows/release-proposal-dispatch.yml
@@ -231,18 +231,15 @@ jobs:
       - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c #stable
         with:
           toolchain: ${{ env.RUSTUP_TOOLCHAIN }}
-      - uses: taiki-e/cache-cargo-install-action@7447f04c51f2ba27ca35e7f1e28fab848c5b3ba7 # 2.3.1
+      
+      - name: Install dependencies
+        run: |
+          sudo apt update && sudo apt install -y libssl-dev  # cargo-public-api dependency
+
+      - name: Install cargo tools
+        uses: taiki-e/install-action@2c41309d51ede152b6f2ee6bf3b71e6dc9a8b7df # 2.49.27
         with:
-          tool: cargo-public-api@0.52.0
-      - uses: taiki-e/cache-cargo-install-action@7447f04c51f2ba27ca35e7f1e28fab848c5b3ba7 # 2.3.1
-        with:
-          tool: cargo-release
-      - uses: taiki-e/cache-cargo-install-action@7447f04c51f2ba27ca35e7f1e28fab848c5b3ba7 # 2.3.1
-        with:
-          tool: git-cliff
-      - uses: taiki-e/cache-cargo-install-action@7447f04c51f2ba27ca35e7f1e28fab848c5b3ba7 # 2.3.1
-        with:
-          tool: cargo-semver-checks@0.48.0
+          tool: cargo-public-api@0.52.0, cargo-release@1.1.5, git-cliff@2.14.1, cargo-semver-checks@0.48.0
       
       - uses: DataDog/dd-octo-sts-action@acaa02eee7e3bb0839e4272dacb37b8f3b58ba80 # v1.0.3
         id: octo-sts
--- a/.github/workflows/release-proposal-dispatch.yml
+++ b/.github/workflows/release-proposal-dispatch.yml
@@ -25,6 +25,14 @@ on:
         required: false
         type: boolean
         default: false
+      level_overrides:
+        description: >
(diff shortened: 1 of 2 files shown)


yannham (approved):
My bash fu is rusty, but I understand the need.

Nit:

> The second commit is unrelated to the feature and can be reviewed on its own

This should probably have been a separate PR.

iunanua:
> My bash fu is rusty, but I understand the need.
> 
> Nit:
> 
> > The second commit is unrelated to the feature and can be reviewed on its own
> 
> This should probably have been a separate PR.

The idea behind that commit was to prepare the data for another workflow which would consume it (in a follow-up PR), but I'm going to put it on hold until we see if it is really necessary. In any case, that change provides useful information when debugging, so I've decided to leave it here.

@yannham

PR #2514, opened 2026-09-11
ci: enforce workspace-level dependency declarations

yannham:
# What does this PR do?

Adds a CI check that enforces our new dependency declaration policy:

1. External dependencies of a workspace member must be declared in the root `[workspace.dependencies]` and inherited with `workspace = true`.
2. Workspace entries must have `default-features = false` and no `features`.

Exceptions are allowed per dependency, clippy-allow style. Rule 1 needs a `# allow(workspace-deps): <justification>` comment above the dependency, rule 2 needs a `# allow(workspace-deps-features): <justification>` comment above the workspace entry explaining why every member gets that feature.

The check is a new linter in the existing `.github/actions` CI workspace, run from a new job in `lint.yml`. It reports file and line, and annotates the PR diff. A handful of manifests are brought in line with the rules so the check passes.

# Motivation

The workspace-dependency migration is essentially done, but nothing stops it from regressing: a new crate can quietly pin its own version, and a workspace entry can quietly turn a feature on for everyone. We agreed to enforce a workspace-level dependency level policy, but no existing tool or set of tools entirely fit the bill. Cargo-autoinherit and friends only cover rule 1 and have no notion of documented exceptions, and clippy has no manifest lints.

Since we already have custom CI rules in Rust, the cost to implement it directly seems reasonable.

# Additional Notes

A few pre-existing cases genuinely cannot inherit yet (Windows crates pinned to older major versions); they are documented exceptions rather than migrations, which would need a Windows build to validate. Resolved feature sets are unchanged on both Linux and Windows targets.

# How to test the change?

CI: the new "Workspace dependency declarations" job must pass.

Locally:

```bash
(cd .github/actions && cargo run -p workspace-deps-lint)
```

To see it fail, drop a dependency with its own version into any member manifest, or enable a feature on a workspace entry, and re-run.

The linter has unit tests for its own rules, run with `(cd .github/actions && cargo test -p workspace-deps-lint)` when editing it. Like the other crates under `.github/actions`, they are not wired into CI.

### DIFF
--- a/.github/actions/workspace-deps-lint/src/lint.rs
+++ b/.github/actions/workspace-deps-lint/src/lint.rs
@@ -1,0 +1,487 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+//! Linter for how dependencies are declared across the libdatadog workspace. See the crate-level
+//! documentation for the rules.
+
+use anyhow::{Context, Result};
+use std::fmt;
+use toml_edit::{ImDocument, Item, Key, TableLike};
+
+/// Marker that waives a rule for the dependency declared right below it:
+/// `# allow(workspace-deps): <justification>`.
+pub const ALLOW_MARKER: &str = "allow(workspace-deps)";
+
+/// The dependency tables a manifest can declare, both at the top level and
+/// under `[target.<cfg>]`.
+const DEP_TABLES: [&str; 3] = ["dependencies", "dev-dependencies", "build-dependencies"];
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Rule {
+    /// External dependencies of a member must be inherited from
+    /// `[workspace.dependencies]` with `workspace = true`.
+    Inherit,
+    /// `[workspace.dependencies]` entries must not turn features on for every
+    /// member: `default-features = false` and no `features`.
+    Features,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Violation {
+    /// Manifest path, as passed in (kept relative for readable reports).
+    pub file: String,
+    /// 1-based line of the offending dependency key.
+    pub line: usize,
+    pub dep: String,
+    pub rule: Rule,
+    /// What is wrong.
+    pub problem: String,
+    /// How to fix it.
+    pub hint: String,
+}
+
+impl fmt::Display for Violation {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        writeln!(
+            f,
+            "{}:{}: `{}` {}",
+            self.file, self.line, self.dep, self.problem
+        )?;
+        write!(f, "    {}", self.hint)
+    }
+}
+
+/// Outcome of looking for a justification above a dependency.
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum Allow {
+    /// No comment at all above the dependency.
+    Absent,
+    /// A comment is there, but it is only an `allow(workspace-deps)` marker
+    /// with an empty justification.
+    MarkerWithoutJustification,
+    /// A free-form comment (no marker) sits above the dependency.
+    Comment,
+    /// An `allow(workspace-deps): <justification>` marker with a justification.
+    Justified,
+}
+
+/// Lint `[workspace.dependencies]` of a workspace root manifest (rule 2). Checks that no feature is
+/// enabled unless there's a comment justification for it.
+pub fn lint_workspace_dependencies(file: &str, src: &str) -> Result<Vec<Violation>> {
+    let doc = ImDocument::parse(src).with_context(|| format!("failed to parse {file}"))?;
+    let Some(deps) = doc
+        .as_table()
+        .get("workspace")
+        .and_then(Item::as_table_like)
+        .and_then(|workspace| workspace.get("dependencies"))
+        .and_then(Item::as_table_like)
+    else {
+        return Ok(Vec::new());
+    };
+
+    let violations = deps.iter().filter_map(|(name, _)| {
+        let (key, item) = deps.get_key_value(name)?;
+
+        // Path entries point at crates of this repository, not at external
+        // dependencies, so the feature policy does not apply to them.
+        if is_path_dependency(item) {
+            return None;
+        }
+
+        let problem = feature_problem(item)?;
+        let line = key_line(src, key);
+
+        // Rule 4: a free-form comment above the entry is enough to justify
+        // enabling a feature for every member.
+        match allow_above(src, line) {
+            Allow::Justified | Allow::Comment => None,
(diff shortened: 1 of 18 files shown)


.github/workflows/lint.yml
```diff
@@ -93,6 +94,8 @@ jobs:
             license_3rdparty:
               - 'Cargo.lock'
               - '**/Cargo.toml'
+            manifests:
+              - '**/Cargo.toml'
```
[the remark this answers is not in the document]
yannham: I agree, but this is a wider problem for ci-only crates. I think we should open a separate PR to run proper checks and lints on every ci-only crate, instead of having a solution specific to this linter.

.github/workflows/lint.yml
```diff
@@ -190,6 +193,35 @@ jobs:
...
+        id: rust-cache
+        uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # 2.8.1
+        with:
+          workspaces: .github/actions
+          cache-targets: true # cache build artifacts
+          shared-key: workspace-deps-lint
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+      - run: echo "rust-cache ${{ steps.rust-cache.outputs.cache-hit == 'true' && 'HIT' || 'MISS' }}"
+      # The helper crates under .github/actions pin their own stable toolchain.
+      - name: Build the dependency linter
```
iunanua, line 214: Instead of building it and using a cache entry (given that it is always at its size limit), have you considered to publish a release (a release on GH is enough, it's not necessary on crates.io) of the `workspace-deps-lint` binary and just download it.

(Although I've also found that building the binary takes about 26 seconds... I expected it to take longer.)

yannham: I haven't actually, since it's not going to change much, I like the idea! Do we do that already for other CI jobs? Would that be a release on the `libdatadog` repo, but just not a libdatadog release? Or should I just attach it to a libdatadog release as an artifact?

iunanua: I think there is no job doing it.
We did a test [here](https://github.com/DataDog/libdatadog/pull/1692/changes#diff-047610c611789abdd8558e7a2a9b053381df4fb087bde6bceb51c0347ab21faeR51-R65) to improve the `changed-crates` action.

It was released manually with gh cli (that's the ugly part).
```
gh release create $TAG --latest=false .github/actions/target/release/changed-crates#changed-crates-x86_64-unknown-linux-gnu --title "changed-crates ${VERSION}" --notes "Prebuilt binary for changed-crates action."
```

[Here](https://github.com/DataDog/libdatadog/releases/tag/ci-crates-v0.1.0) is the release.

iunanua: As it seems we are heading toward rust actions, I'm taking a look at https://github.com/taiki-e/upload-rust-binary-action

yannham: Ok. So right now, no other job is doing it. Do you think we should try to use a release-based workflow for this job only, _right now in this PR_, or is it reasonable to merge like this and have a more generic approach for all ci tooling? Or is the cache pollution unacceptable?

.github/actions/workspace-deps-lint/src/lint.rs
```diff
@@ -0,0 +1,505 @@
...
+pub const ALLOW_MARKER: &str = "allow(workspace-deps)";
+
+/// Marker that waives the features rule for the `[workspace.dependencies]`
+/// entry declared right below it: `# allow(workspace-deps-features):
+/// <justification>`.
+pub const ALLOW_FEATURES_MARKER: &str = "allow(workspace-deps-features)";
+
+/// The dependency tables a manifest can declare, both at the top level and
+/// under `[target.<cfg>]`.
+const DEP_TABLES: [&str; 3] = ["dependencies", "dev-dependencies", "build-dependencies"];
```
iunanua: It seems that `dev_dependencies` and `build_dependencies` are valid aliases for the 2021 edition.
I don't think they're very common, but just in case you want to include them...

.github/actions/workspace-deps-lint/src/lint.rs
```diff
@@ -0,0 +1,505 @@
...
+    else {
+        return Ok(Vec::new());
+    };
+
+    let violations = deps.iter().filter_map(|(name, _)| {
+        let (key, item) = deps.get_key_value(name)?;
+
+        // Path entries point at crates of this repository, not at external
+        // dependencies, so the feature policy does not apply to them.
+        if is_path_dependency(item) {
```
iunanua, line 99: What if one day we could decide to move all the `libdd-` dependency declarations also to workspace level? Should it be configurable?

.github/workflows/lint.yml
iunanua: No sorry, I was just thinking out loud :P
It's fine to merge it 👍

.github/actions/workspace-deps-lint/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+        read_and_lint(&root, &root.join("Cargo.toml"), lint_workspace_dependencies)?;
+    let mut manifests: Vec<PathBuf> = metadata
+        .packages
+        .iter()
+        .map(|package| PathBuf::from(package.manifest_path.as_std_path()))
+        .collect();
+
+    manifests.sort();
+
+    for manifest in &manifests {
```
iunanua, line 95: This loop checks all the member manifests unconditionally.
Thinking about checking only changed members, what if the action accepts a repeatable `-p`/`--package` argument (like other cargo tools) with cargo package names to check?

.github/workflows/lint.yml
iunanua: No sorry, I was just thinking out loud :P
It's fine to merge it 👍

yannham: No worries, it's still good to know right now :slightly_smiling_face:

.github/actions/workspace-deps-lint/src/lint.rs
yannham: Good addition, thanks

.github/actions/workspace-deps-lint/src/lint.rs
yannham: We will at some point, but I think it will be easy to just remove those lines and the `is_path_dependency` then. Making something generic for this switch (which will happen only once, I suppose, and then will not be reverted) sounds a bit overkill.

One risk is that we forget to do so and the lint doesn't enforce workspace deps at first, but making the script generic doesn't prevent this. When we do the switch we will have to toggle something somewhere to start linting workspace crates anyway.

.github/actions/workspace-deps-lint/src/main.rs
yannham: On one hand I expect this check to be rather fast, but I think it doesn't hurt to add such an argument indeed :+1:

iunanua (approved):

gyuheon0h (approved):

PR #2526, opened 2026-09-15
fix(gotter): skip relocations with non-zero addends

scottgerring:
# What does this PR do?
It doesn't make sense to try and hook symbols referenced by function relocations with a non-zero addend, because we are being told the target symbol should point some arbitrary offset into the function we are hooking. Likewise, we can't really patch REL relocations _at all_, because the addend is not recoverable once the dynamic linker has run, so we can't be sure we're not doing the former. 

Fortunately this should be academic - RELA relocations are used exclusively on AMD64 and ARM64, so this PR is effectively clarifying implementation rather than providing a functional change for regular workloads.


yannham (approved):
LGTM, besides the Codex comment that looks legit (although I haven't double checked the claim).

gyuheon0h (approved):

PR #2531, opened 2026-09-15
fix(otlp): use potential dedupe in otlp serialization path

paullegranddc:
# Motivation

Fixes APMSP-3876

the otlp serialization just iterated over meta/metrics keys. This is ok, as we dedup spans before calling the serialization call in the trace exporter.
Just to make sure a bug doesn't cause duplicated arguments, this PRs uses `as_deduped_map` instead of `iter`

Also don't call `drop_chunks` twice if client side stats are enabled


mabdinur (approved):

libdd-trace-utils/src/otlp_encoder/mapper.rs
```diff
@@ -209,7 +209,9 @@ fn collect_span_attributes<T: TraceData>(
             ProtoValue::StringValue(resource_name.to_string()),
         ));
     }
-    for (k, v) in span.meta.iter() {
+    let meta = span.meta.as_deduped_map();
```
yannham: I think we should use `defensive_dedup` instead. The semantics is exactly the same but we'll get a warning like "you should have deduped that before!".

yannham (approved):

libdd-trace-utils/src/span/vec_map.rs
```diff
@@ -56,7 +56,7 @@ impl<K, V> Default for VecMap<K, V> {
     fn default() -> Self {
         Self {
             data: Default::default(),
-            deduped: false,
+            deduped: true,
```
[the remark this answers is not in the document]
paullegranddc: Fixed

libdd-trace-utils/src/otlp_encoder/mapper.rs
paullegranddc: Thanks, I had forgotten this existed

PR #2534, opened 2026-09-16
fix(telemetry): deduplicate Configuration by name, not entire struct

r1viollet:
# What does this PR do?

Adds a `ConfigurationKey` type for deduplicating telemetry configurations by `name` only, matching how `Dependency` uses `DependencyKey`.

Previously, `Configuration` used the default blanket `Keyed<T>` implementation which keyed by the entire struct including `seq_id`. Since `seq_id` is monotonically increasing per call, every `add_configuration()` created a unique entry that bypassed deduplication.

# Motivation

This should reduce the amount of items we keep in memory.
This was found by accident in incident-60326. I doubt that it is related to the leak in the incident, though maybe still a relevant improvement?
The configuration items are bounded to 5k items either way.

# Additional Notes

I do not have functional knowledge on this, and would require someone to clarify if this is a relevant change.

# How to test the change?

Added regression test `test_configuration_deduplicates_by_name_not_seq_id` that verifies 50 inserts of the same config name with different seq_ids results in exactly 1 stored item.


juanjux (approved):

libdd-telemetry/src/data/payloads.rs
```diff
@@ -62,6 +62,22 @@ pub struct Configuration {
     pub seq_id: Option<u64>,
 }
 
+/// Key for deduplicating configurations in the telemetry store. Uses only `name` so that
+/// repeated reports of the same configuration (with different seq_id, value, or origin)
+/// update the existing entry rather than accumulating as separate items.
```
paullegranddc, line 67: This is not okay, reporting all of the versions of a configurations is a feature that is required and currently in use

paullegranddc (changes requested):

libdd-telemetry/src/data/payloads.rs
r1viollet: Thanks for confirming!

PR #2535, opened 2026-09-16
feat(data-pipeline)!: follow the full sql config in CSS obfuscation

Eldolfin:
# What does this PR do?
Pass the full obfuscation config exposed in the agent's /info to the sql obfuscator.

Now accept obfuscation_version 1 and 2 to enable obfuscation even when a custom config is set (see https://github.com/DataDog/datadog-agent/pull/54573).

# Motivation
More precise client-side stats for customers with a custom obfuscation config.

# Additional Notes

See https://datadoghq.atlassian.net/wiki/spaces/APM/pages/6701580544/PENDING+Stats+obfuscation+precisions

# How to test the change?

[APMSP-3720]

[APMSP-3720]: https://datadoghq.atlassian.net/browse/APMSP-3720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libdd-data-pipeline/src/trace_exporter/stats.rs
+++ b/libdd-data-pipeline/src/trace_exporter/stats.rs
@@ -264,13 +264,15 @@ fn update_obfuscation_config(
             client_side_stats.obfuscation_enabled && is_obfuscation_active(agent_info);
         // FIXME(APMSP-3720): there is more than this to obfuscation config
         let sql_obfuscation_mode = (|| {
-            agent_info
-                .info
-                .config
-                .as_ref()?
-                .obfuscation
-                .as_ref()?
-                .sql_obfuscation_mode
+            Some(
+                agent_info
+                    .info
+                    .config
+                    .as_ref()?
+                    .obfuscation
+                    .as_ref()?
+                    .sql_obfuscation_mode,
+            )
         })()
         .unwrap_or_default();
         client_side_stats
--- a/libdd-data-pipeline/src/agent_info/schema.rs
+++ b/libdd-data-pipeline/src/agent_info/schema.rs
@@ -97,7 +97,7 @@ pub struct ObfuscationConfig {
     pub elasticsearch: libdd_trace_obfuscation::json::JsonObfuscator,
     pub opensearch: libdd_trace_obfuscation::json::JsonObfuscator,
     pub mongodb: libdd_trace_obfuscation::json::JsonObfuscator,
-    pub tag_replace_rules: Option<Vec<ReplaceRule>>,
+    pub tag_replace_rules: Vec<ReplaceRule>,
 }
 
 impl AgentInfo {
--- a/libdd-data-pipeline/src/agent_info/schema.rs
+++ b/libdd-data-pipeline/src/agent_info/schema.rs
@@ -1,6 +1,11 @@
 // Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
 // SPDX-License-Identifier: Apache-2.0
 //! This module provides struct representing the info endpoint response
+use libdd_trace_obfuscation::{
+    json::JsonObfuscator,
+    obfuscation_config::{self, SqlObfuscationMode},
+    replacer::ReplaceRule,
+};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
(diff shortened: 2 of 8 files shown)


libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -263,24 +270,19 @@ fn update_obfuscation_config(
...
-        })()
-        .unwrap_or_default();
+        let obfuscation_config = (|| agent_info.info.config.as_ref()?.obfuscation.as_ref())();
+        let mut sql_obfuscation_config = obfuscation_config
+            .and_then(|cfg| cfg.sql.as_ref())
+            .cloned()
+            .unwrap_or_default();
+        sql_obfuscation_config.obfuscation_mode = obfuscation_config
+            .map(|cfg| cfg.sql_obfuscation_mode)
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
Eldolfin: We won't break existing fields in /info to the "legacy" `sql_obfuscation_mode` is there to stay

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -35,9 +35,12 @@ use web_time::SystemTime;
 
 pub(crate) const STATS_ENDPOINT: &str = "/v0.6/stats";
 
-/// The maximum obfuscation version this tracer supports.
+/// The obfuscation version this tracer supports.
 #[cfg(feature = "stats-obfuscation")]
 pub(crate) const SUPPORTED_OBFUSCATION_VERSION: u32 = 1;
+/// The obfuscation version this tracer supports, given that we follow the full obfuscation config
+#[cfg(feature = "stats-obfuscation")]
+pub(crate) const NEW_SUPPORTED_OBFUSCATION_VERSION: u32 = 2;
```
yannham: As an outsider, I don't find it really clear what are those two different versions. Does it just meant that the tracer supports both version? In that case would a single `SUPPORTED_VERSIONS = [1, 2]` make sense? Or do you need to distinguish both, and why?

Eldolfin: Yes, both version are supported but the "official" version we support and send back to the agent is still `1`. Version `2` is just a temporary workaround the agent sends us when it has a custom config but it's not a real obfuscation version.

yannham: Ok. I think I still find the naming confusing. Maybe the second one shouldn't be called supported but something like `CUSTOM_CONFIG_OBFUSCATION_VERSION` or something. And, similarly, explain in the comment that it doesn't have the same status as a "supported" version, but is rather a marker value. Those are just suggestions though

yannham (approved):

libdd-data-pipeline/src/trace_exporter/stats.rs
ichinaski: Agreed with the naming feedback here. The `NEW_` prefix is confusing.

libdd-data-pipeline/src/trace_exporter/stats.rs
```diff
@@ -35,9 +35,12 @@ use web_time::SystemTime;
...
 #[cfg(feature = "stats-obfuscation")]
 pub(crate) const SUPPORTED_OBFUSCATION_VERSION_STR: &str = "1";
```
ichinaski, line 51: Does this require an update when version `2` is used? We seem to be unconditionally using this version on the stats exporter header.

libdd-data-pipeline/src/trace_exporter/stats.rs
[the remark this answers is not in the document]
ichinaski: Reopening this just to comment on the same thing. Feel free to close again if you feel strongly about it @Eldolfin . It may be worth adding a small tweak here in case we decide to remove the legacy field, which will be redundant once the new object is propagated.

ichinaski (commented):
Consider adding some tests to cover the new behavior.

libdd-data-pipeline/src/trace_exporter/stats.rs
Eldolfin: Do you think we should send the header with version `2` when a custom config is used ? For me it didn't matter because:
- version `2` is not a real obfuscation version per se: the obfuscation still gives the same result
- from what I saw in the agent and in the spec: the obfuscation version value sent from the client is not used, only its presence is checked (unless the client send a version < 1 in which case a warning is printed agent-side)

I'm I missing something ?

ichinaski: That makes sense, I was just confused with the naming system on this file 👍

libdd-data-pipeline/src/trace_exporter/stats.rs
Eldolfin: I agree it was confusing, I renamed and explaining it 174908c

libdd-data-pipeline/src/trace_exporter/stats.rs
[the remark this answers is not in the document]
Eldolfin: Ok I thought it was a hard rule not to remove existing fields from /info, but given I added he field myself [5 months ago](https://github.com/DataDog/datadog-agent/pull/48857) we might be able to remove it actually

ichinaski (approved):

Eldolfin:
Waiting for 1 more approve on the [RFC](https://datadoghq.atlassian.net/wiki/spaces/APM/pages/6701580544/PENDING+Stats+obfuscation+precisions) then I'll merge this

PR #2537, opened 2026-09-16
Julio/split common

hoolioh:
# What does this PR do?

# Motivation

# Additional Notes

# How to test the change?


libdd-types/src/endpoint.rs
Eldolfin, line 1: I feel like `decode_uri_path_in_authority` and `parse_uri` don't belong in the types crate. They contain too much logic and add more dependencies to what should be a minimal crate.

IMO we should limit the logic in type crates to impl blocks that need to be here (accessors for sealed types and std trait impls)

libdd-types/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+#![cfg_attr(not(test), deny(clippy::expect_used))]
+#![cfg_attr(not(test), deny(clippy::todo))]
+#![cfg_attr(not(test), deny(clippy::unimplemented))]
+
+extern crate alloc;
+
+pub mod endpoint;
+pub mod tag;
+
+pub use endpoint::{decode_uri_path_in_authority, parse_uri, Endpoint, HttpRequestBuilder};
```
Eldolfin, line 14: Not something to fix in this PR but I'm not a fan of `pub use` to keep backward compatibility.

Having to rename a `use somecrate` into `use somecrate::submodule` is pretty easy, I would call that a light and acceptable breaking change.

Eldolfin (commented):
This seems like a reasonable split of `libdd-common`, the remaining types being logic heavy.

I was hopping to see fewer crates depending on `libdd-common` as a result of the refactor, aka having simplified dependency graph, but looking at the diff it's not the case.

You could use `cargo machete` to see unused dependency of `libdd-common` or ask your agent to check that if you want

PR #2538, opened 2026-09-16
feat(trace-exporter)!: add API to force flush stats on the trace exporter

paullegranddc:
# Motivation

CSS system-tests are marked as missing in dd-trace-rs. The reaon for this is that we cannot trigger a flush for stats without shutting done.

# Changes

The issue with the current code structure is that the trace exporter holds a reference to the stats exporter only through a generic worker handle which does not allow "sending orders".

In order to trigger a flush from the trace exporter, we need to hold a reference to the stats exporter alongside the worker handle.

This PR:
* Adds a new trait `FlushableStatsExporter` which we can use to implement flushing both for OTLPStats and Dd stats
* Implement the Worker trait on a Arc<StatsExporter> so we can hand-out multiople references to the exporter
* Store a Weak<dyn FlushableStatsExporter` alongside the worker handle.

The reaon the flusher is a Weak is that we don't want to prevent dropping the stats exporter from being dropped whenever we  turn stats on or off.


libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -728,6 +728,32 @@ impl<
...
 
+    /// Force an immediate flush of client-computed stats if stats computation is currently
+    /// enabled
+    pub async fn flush_client_side_stats_async(&self) -> bool {
+        let status = self.client_side_stats.status.load_full();
+        if let StatsComputationStatus::Enabled { flush_handle, .. } = &*status {
+            if let Some(exporter) = flush_handle.upgrade() {
+                exporter.force_flush().await;
+            }
+            true
```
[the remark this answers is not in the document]
paullegranddc: This is ok, force flushing is best effort. If a caller shutsdown the runtime they cannot expect that they're data is going to be sent successfully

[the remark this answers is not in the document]
ekump: minor: @paullegranddc Yes, but we probably shouldn't be returning true in that case, especially if we didn't even attempt to flush because the `upgrade()` returned `None`?

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -127,6 +127,43 @@ impl From<TracerMetadata> for StatsMetadata {
...
 }
+pub trait StatsCap: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static {}
+
+impl<T: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> StatsCap for T {}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+pub trait FlushableStatsExport: Debug + MaybeSend + Sync + 'static {
+    /// Flush all currently-buffered stats and send them.
+    /// Best-effort: errors are logged by the exporter and swallowed here.
```
ekump, line 138: Are we actually logging the error anywhere? I think for the DD path we already log transport errors, but not if the force flush fails for other reasons? And I don't think the OTLP path is logging at all?

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -728,6 +728,32 @@ impl<
...
+        } else {
+            false
+        }
+    }
+
+    /// Sync call over [`Self::flush_client_side_stats_async`]
+    /// SAFETY: panics inside an existing tokio context.
```
ekump, line 746: minor: I believe convention is for `SAFETY` to only be used for documenting the invariants that make an unsafe block sound. You probably want to use a [Panics](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure) section?

libdd-data-pipeline/tests/test_agentless_stats.rs
```diff
@@ -674,3 +674,66 @@ async fn test_agentless_stats_honors_additional_metric_tag_keys() {
...
+        exporter
+            .send_trace_chunks(
+                PooledChunks::unpooled(vec![vec![make_root_span(1, Some(1.0), 0)]]),
+                None,
+            )
+            .expect("send_trace_chunks failed");
+
+        // Force an immediate flush through the dyn handle, ahead of the periodic worker.
+        let flushed = exporter.flush_client_side_stats();
+        exporter.shutdown(None).expect("shutdown failed");
```
ekump, line 718: doesn't `shutdown` also flush? How do we know the manual flush was the one doing the flushing?

libdd-data-pipeline/src/otlp/metrics.rs
ekump, line 1: Should we have an integration test for OTLP similar to the DD pipeline?

ekump (approved):
Some minor comments. It would be cool if they could be addressed, but non-blocking. LGTM.

yannham:
Not reviewed yet, but naive question: could we instead of storing a generic worker handle, store a more specific concrete worker type that would provide a flush API, instead of having an additional `Weak` ref in a different field (but which is still logically tied to the worker)? IIRC, this is what we do for telemetry, where `TelemetryWorker` has a public API that let tracers explicitly `flush`, `shutdown`, etc. 

The worker should probably not become a mirror API for the exporter, but if it's just about exposing actions that the worker has to do itself already, it seems reasonable.

libdd-data-pipeline/src/otlp/metrics.rs
```diff
@@ -326,31 +327,50 @@ impl<C: HttpClientCapability + SleepCapability> OtlpStatsExporter<C> {
...
+        (Self(arc), weak)
+    }
+}
+
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
-impl<C: HttpClientCapability + SleepCapability + MaybeSend + Sync + 'static> Worker
-    for OtlpStatsExporter<C>
-{
+impl<C: StatsCap> FlushableStatsExport for OtlpStatsExporter<C> {
```
VianneyRuhlmann:
```suggestion
impl<C: StatsCap> FlushableStatsExporter for OtlpStatsExporter<C> {
```

libdd-data-pipeline/src/trace_exporter/mod.rs
```diff
@@ -728,6 +728,32 @@ impl<
         self.send_trace_chunks_inner(trace_chunks).await
     }
 
+    /// Force an immediate flush of client-computed stats if stats computation is currently
+    /// enabled
```
VianneyRuhlmann, line 732: nit: explain what the returned boolean means

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -127,6 +127,43 @@ impl From<TracerMetadata> for StatsMetadata {
...
+
+impl<Cap: StatsCap, Con: FlushableConcentrator + Debug + MaybeSend + Sync + 'static>
+    SharedStatsExporter<Cap, Con>
+{
+    pub fn wrap(exporter: StatsExporter<Cap, Con>) -> (Self, Weak<dyn FlushableStatsExport>) {
+        let arc = Arc::new(exporter);
+        let weak = Arc::downgrade(&arc) as Weak<dyn FlushableStatsExport>;
+        (Self(arc), weak)
+    }
+}
```
VianneyRuhlmann, line 166: Can this be made generic and reuse it for OtlpStatsExporter ?

VianneyRuhlmann (approved):

mabdinur (approved):

libdd-trace-stats/src/stats_exporter.rs
paullegranddc: Not really, I tried and it makes more code than otherwise

libdd-data-pipeline/tests/test_agentless_stats.rs
paullegranddc: Updated to capture and assert on received stats before shutdown

libdd-trace-stats/src/stats_exporter.rs
paullegranddc: Added and error log for the otlp export error

VianneyRuhlmann (approved):

libdd-data-pipeline/src/otlp/metrics.rs
paullegranddc: added tests

PR #2539, opened 2026-09-16
fix(spawn_worker): harden process creation

cataphract:
Functional changes:

* On Linux, weak-import `_Fork` and fall back to raw `SYS_fork` or `clone(SIGCHLD)` when it is unavailable. On supported macOS 12+ systems, use Darwin's fork-like `vfork()`. In earlier versions, `vfork()` suspends the parent and fully shares the memory space. See https://github.com/apple-oss-distributions/Libc/blob/Libc-1506.40.4/sys/fork.c#L111-L114
* Skip application `pthread_atfork` handlers for both the initial and second forks.
* Block all blockable signals around process creation and restore the exact original mask in the parent and ordinary child. On Linux, this includes libc-internal signals.
* Keep signals blocked through daemon setup and the second fork, reset every mutable disposition to `SIG_DFL`, call `setsid()` before that fork, and clear the final child's mask before its first exec attempt.
* Move session creation out of sidecar startup so detachment occurs before exec and before sidecar initialization. This ensures the process does not become a sessionleader.
* Use raw Linux `rt_sigaction` and `rt_sigprocmask` syscalls for post-fork signal setup, avoiding libc signal wrappers when the raw fallback may leave libc bookkeeping unrepaired.
* Restrict Unix process creation to Linux and macOS at compile time.
* Prepare arguments, environment, executable sources, temporary paths, descriptor scan bounds, and descriptor mappings before the first fork.
* Preserve non-UTF-8 bytes in prepared temporary-file paths and the `LD_PRELOAD` current-executable path.
* Route configured standard streams to fds 0-2, the optional passed fd to fd 3, and dependency memfds contiguously after it.
* Protect remapping sources from `dup2` collisions and treat remapping failures as fatal in the child.
* Use raw Linux `execveat`, `close`, and `ppoll` syscalls in the child, avoiding old glibc path formatting and musl AIO or cancellation state.
* Try to precompute a randomized disk-fallback path before forking. If available, create it with `O_EXCL|O_NOFOLLOW` only after `execveat` fails, populate it with raw syscalls, and apply executable permissions independently of the inherited umask.
* Keep a successfully materialized fallback through `execve`, and attempt to unlink a newly created file if population or fallback exec fails.
* Attempt to close other inherited child descriptors with Linux `close_range` or a bounded poll scan, without traversing procfs or devfs after forking.
* Clean up persisted temporary artifacts when preparation or the initial fork fails.
* On child setup or exec failure, unlink temporary artifacts, emit a static diagnostic using a raw Linux write syscall, and `_exit` without allocation or unwinding.


spawn_worker/src/unix/fork.rs
```diff
@@ -74,78 +35,322 @@ pub fn set_default_child_panic_handler() {
...
+        };
+        let last_signal = match libc::c_int::try_from(linux_raw_sys::general::_NSIG) {
+            Ok(signal) => signal,
+            Err(_) => return Err(libc::EINVAL),
+        };
+
+        let mut signal = 1;
+        while signal <= last_signal {
+            // These two can't have their dispositions changed
+            if signal != libc::SIGKILL && signal != libc::SIGSTOP {
```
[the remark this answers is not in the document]
cataphract: @codex The sidecar entrypoint never went through lang_start before either. To the extent the disposition of SIGPIPE was changed, this must have happened because the web server did so. If ignoring SIGPIPE is indeed necessary, this should be done in the sidecar entrypoint. Fix that

spawn_worker/src/fork_shim.c
```diff
@@ -0,0 +1,90 @@
...
+  // bookkeeping. The fallback cannot call that private function, so its child
+  // must use raw syscalls or wrappers known not to depend on unrepaired libc
+  // state until exec or _exit; spawn_worker enforces that stricter contract.
+
+  *error = result == -1 ? errno : 0;
+  return result;
+}
+
+#if defined(__APPLE__)
+static bool ddog_spawn_worker_has_fork_like_vfork(void) {
```
bwoebi: build.rs already checks minimum version. returning ENOTSUP is bad.

spawn_worker/src/unix/spawn.rs
```diff
@@ -492,285 +350,1025 @@ impl SpawnWorker {
...
-            unsafe { libc::dup2(fd.as_raw_fd(), 3) };
-            4..=i32::MAX
+        // TODO: A descriptor opened before the hard limit is lowered can remain
+        // above rlim_max and escape the fallback scan when close_range is unavailable.
+        let descriptor_count = if limit.rlim_max == libc::RLIM_INFINITY {
+            Self::system_fd_limit()?
         } else {
-            3..=i32::MAX
+            limit.rlim_max
         };
```
bwoebi: Unnecessary case to handle, really.

spawn_worker/src/unix/spawn.rs
```diff
@@ -492,285 +350,1025 @@ impl SpawnWorker {
...
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    unsafe fn close_fd_range(first: u32, last: u32, fallback_last: u32) {
+        unsafe {
+            Self::close_open_fds(first, last.min(fallback_last));
+        }
+    }
+
+    unsafe fn close_open_fds(first: u32, last: u32) {
```
bwoebi: So, a typical value for /proc/sys/fs/nr_open is 2^20. ... And the actual hard maximum is 2^31-1.
I don't want to emit multiple thousands or more ppoll() calls to do a single fork.

Tell me, what's the problem with walking /dev/fd or /proc/self/fd, which actually is accurate?
I don't mind correctness, but that's definitely at a significant worst case performance cost.

bwoebi: My proposal would be reverting the ChildFdPlan introduction and restoring the old code for fd closing. I find it honestly a lot less readable. Nicer abstracted away, but really not helping.

spawn_worker/src/unix/spawn.rs
```diff
@@ -383,29 +264,13 @@ impl SpawnWorker {
...
 
-            let fd = src_fd.try_clone()?;
-
-            // Strip any close on exec flag on this fd
-            let flags = nix::fcntl::fcntl(fd.as_raw_fd(), nix::fcntl::F_GETFD)?;
-            unsafe {
-                libc::fcntl(
-                    fd.as_raw_fd(),
-                    libc::F_SETFD,
-                    flags & !nix::libc::FD_CLOEXEC,
```
bwoebi, line 398: Is this, for some reason, no longer necessary?

cataphract: The reason the guarantee given by POSIX [`dup()`/`dup2()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/dup.html):

> Upon successful completion, if fildes is not equal to fildes2, the FD_CLOEXEC and FD_CLOFORK flags associated with fildes2 shall be cleared.

The new code guarantees that source and destination are never the same by duplicating all the source fds above the last dup2 destination. This is sufficient but upon closer inspection I can optimize this a bit more and only duplicate if source <= destination

spawn_worker/src/fork_shim.c
cataphract: forced a minimum deployment target of mac os 12

spawn_worker/src/unix/spawn.rs
cataphract: >Tell me, what's the problem with walking /dev/fd or /proc/self/fd, which actually is accurate?

The problem is that it has to be done in the child, otherwise it can race with another thread opening a file descriptor, and we're limited in what we can do in the forked children...

I'm gonna see if it can be done safely

bwoebi: But, it was already previously done in the child. I don't see the issue here? It was working, it's just a matter of restoring the code?

bwoebi: The only thing which it did, was allocating, which might not be perfectly safe. But that's just a matter of going down a layer of abstraction in that specific place when reading/converting the file names.

cataphract: I assume you're fine with systems with kernel < 5.9 (no close range) and where procfs is not mounted failing? In that case I can remove the poll() fallback altogether

bwoebi: Yes.

bwoebi:
I would also like to remind you to keep the code simple. I know, it's easily to build abstractions quickly, especially with AI, but it also makes it harder to reason about it for humans, in terms of performance and side-effects.

spawn_worker/src/unix/spawn.rs
cataphract: Done in https://github.com/DataDog/libdatadog/pull/2539/commits/de6edc961614c78a9345475dda941f4935a42be5

I also improved (in the next) the wait for the child to detect failures up to a failing execve call

PR #2545, opened 2026-09-17
feat(trace_utils)!: add mutable metadata

VianneyRuhlmann:
# What does this PR do?

Introduce a shared structure for mutable metadata. All the components hold a `MutableMetadataHandle` which stores an ArcSwap of the mutable fields. The SDK can update these values and the components must load the new value whenever they access them.

# Motivation

Allow runtime_id to be updated without recreating components. Other values will benefit from this feature e.g. process_tags, test_session_token.

# Breaking change
## What changed

| Struct | Removed fields | New field |
|---|---|---|
| `TracerMetadata` | `runtime_id`, `process_tags` | `mutable_metadata: MutableMetadataHandle` |
| `StatsMetadata` | `runtime_id`, `process_tags` | `mutable_metadata: MutableMetadataHandle` |
| `OtlpResourceInfo` | `runtime_id`, `process_tags` | `mutable_metadata: MutableMetadataHandle` |

Formats are unchanged: `runtime_id` is a runtime UUID; `process_tags` is a
comma-separated `key:value` string.

## Migrating construction

Replace the two removed fields with one `MutableMetadataHandle`. Initialize it from
a `MutableMetadata` snapshot:

```rust
// Before
let metadata = TracerMetadata {
    runtime_id: rt_id,
    process_tags: tags,
    ..Default::default()
};

// After
let mut snapshot = MutableMetadata::default();
snapshot.runtime_id = rt_id;
snapshot.process_tags = tags;
let metadata = TracerMetadata {
    mutable_metadata: MutableMetadataHandle::from(snapshot),
    ..Default::default()
};
```

## Migrating field reads

Values are no longer stored on the struct; read a snapshot through the handle at the
point of use:

```rust
// Before
let rt = metadata.runtime_id;
let tags = &metadata.process_tags;

// After
let snapshot = metadata.mutable_metadata.load();
let rt = &snapshot.runtime_id;
let tags = &snapshot.process_tags;
```
# Additional Notes

# How to test the change?


yannham (commented):
I haven't looked at yet, but I wonder how this will interact with metadata publication in `libdd_library_config_ffi::tracer_metadata::ddog_tracer_metadata_store` (which publishes the metadata to some shared memory, including the OTel process context). Which I think more or less assume that the metadata are immutable once published. Should re-publication be automatic (if that's even possible) upon modification? Or we will ask the tracers to re-publish manually if the want to update?

VianneyRuhlmann:
Currently the runtime_id already changes on a fork. I'm not sure how `libdd_library_config_ffi` handles it (the tracer probably has to call `set_metadata` again), but I think these are different concerns. This PR only make it possible to reuse a component when one of these values change without having to recreate the whole component.

libdd-common/src/mutable_metadata.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /// Set the process tags value
+    pub fn set_process_tags(&self, process_tags: String) {
+        self.0.rcu(|current| {
+            let mut next = (**current).clone();
+            next.process_tags = process_tags.clone();
+            next
+        });
+    }
```
Eldolfin: nit: what about using `self.update` for these two ? 🤓

libdd-common/src/mutable_metadata.rs
```diff
@@ -0,0 +1,207 @@
...
+/// - `runtime_id`: a runtime UUID,
+/// - `process_tags`: comma-separated `key:value` pairs, e.g. `"k1:v1,k2:v2"`.
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct MutableMetadata {
+    /// Runtime UUID. Uniquely identifies the runtime within the host process.
+    pub runtime_id: String,
+    /// Comma-separated `key:value` process tags of the instrumented application.
+    pub process_tags: String,
+}
```
Eldolfin: Have you considered having a `new` function instead of deriving `Default` so that we can fill runtime_id with a valid, random uuid by default ?

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -585,10 +588,19 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
     ///
     /// When set, this ID is reused for both OTLP trace exports and OTLP trace-metrics so that all
     /// signals can be correlated by the backend. If not set, a fresh UUID is generated.
+    ///
+    /// Ignored when a shared handle was set via
+    /// [`TraceExporterBuilder::set_mutable_metadata`].
```
Eldolfin: Might be worth adding a `warn!` in case `self.mutable_metadata.is_some()`

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -585,10 +588,19 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
...
     pub fn set_runtime_id(&mut self, id: &str) -> &mut Self {
         self.runtime_id = Some(id.to_owned());
         self
     }
+
+    /// Provide a shared, updatable [`MutableMetadataHandle`].
+    pub fn set_mutable_metadata(&mut self, handle: MutableMetadataHandle) -> &mut Self {
+        self.mutable_metadata = Some(handle);
```
Eldolfin: Might be worth adding a `warn!` in case `self.runtime_id.is_some()`

libdd-trace-stats/src/stats_exporter.rs
```diff
@@ -859,15 +859,42 @@ mod tests {
         mock.assert_calls_async(2).await;
     }
 
+    #[test]
+    fn mutable_metadata_updates_propagate_to_encoded_payloads() {
+        let handle = get_test_metadata().mutable_metadata;
+        let meta = StatsMetadata {
+            mutable_metadata: handle.clone(),
+            ..get_test_metadata()
+        };
```
Eldolfin: Seems equivalent ?
```suggestion
        let meta = get_test_metadata();
        let handle = meta.mutable_metadata.clone();
```

libdd-common/src/mutable_metadata.rs
VianneyRuhlmann: Yes but this means adding a new field becomes a breaking change since you need to update `new` signature

Eldolfin: I was thinking of a `new` function without argument that would fill runtime_id with a `Uuid::new_v4()` and leave `process_tags` empty. That way it would have to be intentional to have an empty runtime_id

libdd-common/src/mutable_metadata.rs
VianneyRuhlmann: These are more convenient for the most common use case which is to update a single field. `update`is provided to avoid doing the rcu cycle twice when updating multiple fields. Also update cannot be exposed through ffi since it requires a closure.

libdd-common/src/mutable_metadata.rs
VianneyRuhlmann: We rarely (almost never) want to use a generated uuid since we should use the same one as the app, I don't think it worth paying the cost of generating a uuid everytime since it's going to get dropped

libdd-common/src/mutable_metadata.rs
Eldolfin: I meant using `update` in the implementation of these functions, it's really nit because it would save something like 2 redundant lines

libdd-common/src/mutable_metadata.rs
Eldolfin: Makes sense thanks

libdd-data-pipeline/src/trace_exporter/builder.rs
```diff
@@ -871,35 +883,42 @@ impl<R: SharedRuntime> TraceExporterBuilder<R> {
                 ..Default::default()
             }));
 
-        let runtime_id = self
-            .runtime_id
-            .unwrap_or_else(|| uuid::Uuid::new_v4().to_string());
+        let mutable_metadata = match self.mutable_metadata {
+            Some(handle) => handle,
```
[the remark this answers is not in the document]
VianneyRuhlmann: Adressed in https://github.com/DataDog/libdatadog/pull/2552

PR #2548, opened 2026-09-18
feat(obfuscation)!: make json transforms caller-provided

### LINKED ISSUE #2541: Obfuscation: an empty SQL obfuscation result is reported as success where the Agent fails, and non-ASCII identifier support is untested
This was a finding from an Opus audit as I work on a trace pipeline where I am introducing libdatadog in agent-data-plane. Details below!

> AI
>
> Hello. I am working on the trace pipeline in Saluki (ADP), where we are porting the Agent's obfuscation behavior. I routinely ask cold, clean-room agents to review the port against the reference implementation, and on a JSON obfuscation PR (DataDog/saluki#2601) the review found two places where a SQL obfuscation success/failure decision diverges from the Agent.
>
> The findings were made against Saluki's own vendored port of the Agent's Go obfuscator, which is not this crate, so before filing I checked what `libdd-trace-obfuscation` does for each case. The first divergence exists in the crate; the second the crate already handles correctly, but nothing pins it.
>
> ## AI finding 3: two Ok/Err disagreements with the Agent's SQL obfuscator, now visible here and tracked nowhere
>
> Writing the Agent's failure message on `Err` (`json.rs:220-226`) is right, but it makes every
> Ok/Err disagreement between our `obfuscate_sql_string` and the Agent's show up in the body. I found
> two in differential testing, in opposite directions, and I could not find either in the acceptance
> lists of #2410, #2411 or #2412.
>
> **1. Empty result.** The Agent errors on an empty obfuscation result (`pkg/obfuscate/sql.go:450-451`,
> `errors.New("result is empty")`) and so writes the failure message; we return `Ok("")`:
>
> ```
> in:  {"query": ""}     go: {"query":"Datadog-agent failed to obfuscate..."}   rs: {"query":""}
> in:  {"query": "\t"}   go: {"query":"Datadog-agent failed to obfuscate..."}   rs: {"query":""}
> ```
>
> The PR body calls this out and points at #2410, but #2410's acceptance items are the `VALUES`
> grouping and the escape-dialect retry; nothing there would make an empty result an error. Could we
> add an acceptance item to #2410 (or file a small follow-up) so it does not fall through?
>
> **2. Non-ASCII identifiers.** `sql_tokenizer.rs:597-609` classifies identifier characters with
> `is_ascii_alphabetic`/`is_ascii_alphanumeric`, where the Agent uses `unicode.IsLetter`
> (`pkg/obfuscate/sql_tokenizer.go:914-920`). So a query with a non-ASCII identifier is a `LexError`
> for us and fine for the Agent:
>
> ```
> in:  {"query": "select * from café"}
> go:  {"query":"select * from café"}
> rs:  {"query":"Datadog-agent failed to obfuscate SQL string. Enable agent debug logs for more info."}
> ```
>
> A literal like `'José'` is fine, since it is inside quotes; it is identifiers, and any non-ASCII text
> outside a literal, that trip it. Before this PR those values came out as `"?"`; now a 79-character
> English sentence lands in the body and, for the exec-plan case, in the resource that stats aggregate
> on. The same gap already shows up as "Non-parsable SQL query" on the span path (`mod.rs:155-160`), so
> it is not this PR's doing, and the same lines also drop the Agent's `@` and `#` identifier characters.
>
> I don't think either of these should hold up this PR. Should they be tracked somewhere, though, since
> the SQL issues as written wouldn't catch them?
> AI
>
> Verified against `libdd-trace-obfuscation` at rev `8b0fa65092` (current `main` for `sql.rs`):
>
> - Empty result: `obfuscate_sql_string("\t")` and `obfuscate_sql_string("")` both return `""` as a success. `obfuscate_sql_opt` only returns `None` for input that is empty before obfuscation; a query that obfuscates to empty still comes back as `Some("")`. Through the JSON obfuscator (`json/mod.rs`, whose transformer is a plain `fn(&str) -> String`), a `{"query": ""}` body is written back as `{"query":""}`, where the Agent writes its failure message.
> - Non-ASCII identifiers: the crate already matches the Agent. `is_ident_start` and `is_ident_char` in `sql.rs` accept bytes above 127, and `@` and `#` continue identifiers, so `obfuscate_sql_string("select * from café")` returns `select * from café`, like the Agent. The vendored port classifies identifier characters with `is_ascii_alphabetic` and rejects the same query.
>
> Two asks. Should the crate expose the Agent's failure semantics for an empty obfuscation result, so a caller can distinguish "obfuscated to nothing" from a failure? This is the same missing error channel that keeps Saluki's JSON obfuscation a local port: the crate's `JsonObfuscator` cannot carry a `SqlObfuscationConfig` or report transformer failures (the PR body of DataDog/saluki#2601 says so explicitly). And could non-ASCII identifier support get a regression test, so the Go `unicode.IsLetter` behavior stays pinned against future refactors?

_Originally posted by @webern in https://github.com/DataDog/saluki/pull/2601#discussion_r4014714725_

webern:
## Human Summary

While working on DataDog/saluki#2601, I discovered that I couldn’t use libdatadog’s JSON obfuscator as written. Its transformer was a plain function pointer whose only argument was the JSON value. Unlike a closure, a function pointer cannot capture runtime state, so there was no way to pass Saluki’s dynamically loaded SQL obfuscation configuration without putting it in static or global state. Existing users can configure ordinary SQL obfuscation, but the existing JSON transformer API provides no normal way to pass those SQL settings when transforming SQL embedded in JSON. The provided transformer uses default settings.

In fixing this, I decided to take it a little further. Transform failures are now returned as structured information, scan failures use a typed error, and the hot-path API accepts caller-owned output and scratch buffers so repeated calls can reuse their allocations. Transformations return `Cow<str>`, which also avoids forcing an allocation when they can return borrowed text.

The coding agent brought in `thiserror` as a dependency. This is idiomatic and seems like a good call. The scanner's new const fns are `clippy::missing_const_for_fn`, which the workspace enables; they have no runtime effect. `Scanner::new` becoming restart is for buffer reuse. The parse-state stack now survives between passes.

# What does this PR do?

This separates JSON obfuscation policy from transformation behavior. `JsonObfuscatorConfig` now contains only serializable, comparable data. Callers provide transforms per call through generic closures that can capture runtime SQL configuration.

The allocating API is joined by `obfuscate_into`, which reuses caller-owned output and scratch buffers. Scratch capacity is observable and explicitly trimmable so callers can choose their memory-retention policy.

JSON scan errors and transform errors are reported separately. SQL obfuscation now reports an empty result as `SqlObfuscationError::EmptyResult`, and the crate exports the Agent's exact SQL failure replacement. This also pins non-ASCII identifier behavior and fixes the JSON unquoting fallback to retain the literal when unquoting fails.

# Motivation

The stored `fn(&str) -> String` transformer cannot capture runtime configuration or report errors. This prevents Agent-compatible configured SQL transformation inside JSON values.

DataDog/saluki#2601 currently duplicates libdatadog's JSON scanner to work around that API. A call-scoped closure lets Saluki use the shared scanner while retaining its configured SQL behavior, logging, and failure policy.

# Additional Notes

Transform callbacks return `Cow<str>`, allowing borrowed, static, or owned replacements. The callback is generic and allocation-free; no transformer trait or stored trait object is introduced.

A transform error replaces that value with `"?"`, records the error, and continues. Callers that need the Agent's SQL fallback can observe and log the SQL error inside the callback, then return `SQL_OBFUSCATION_FAILURE_REPLACEMENT`.

`JsonObfuscationScratch` retains capacity from the largest prior input until the caller trims or drops it. `retained_capacity` and `trim_to` make that policy explicit.

## Rebase onto #2490

#2490 moved `SqlObfuscateConfig`, `SqlObfuscationMode` and `DbmsKind` from `sql` into `obfuscation_config` and renamed the config struct to `SqlConfig`, and it began deserializing `JsonObfuscator` directly from the Agent's `/info` payload. Resolved as follows:

- Adopted `obfuscation_config::{DbmsKind, SqlConfig, SqlObfuscationMode}` everywhere, including the `obfuscate_with` doc example. No type is reintroduced in `sql`.
- `JsonObfuscatorConfig` keeps `#[serde(default)]` but not `deny_unknown_fields`: #2490 dropped it deliberately, and an `/info` payload from a newer Agent carries fields this struct has no counterpart for. A test pins that forward compatibility.
- #2490's hand-written `PartialEq for JsonObfuscatorConfig`, which existed only to skirt the uncomparable `transformer` field, is replaced by a derive now that the field is gone. The config also derives `Eq`.
- `obfuscate_resource_for_stats` and `obfuscate_pb_span` now consume the `Result` from `obfuscate_sql`: a resource that obfuscates to nothing is left as sent rather than blanked.

## Agent `/info` field names

Checked against the Agent rather than guessed. `pkg/trace/api/info.go` serves a *reduced* view:

```go
type reducedJSONObfuscationConfig struct {
	Enabled  bool     `json:"enabled"`
	KeepKeys []string `json:"keep_keys"`
}
...
oconf.Elasticsearch = reducedJSONObfuscationConfig{Enabled: o.ES.Enabled, KeepKeys: o.ES.KeepValues}
```

So `/info` sends `keep_keys`, which is already this crate's field name, and it does not report the transform set at all. The Agent's own obfuscation config (`pkg/obfuscate`, `apm_config.obfuscation.*`) calls the same two sets `keep_values` and `obfuscate_sql_values`. Rather than rename the Rust fields — `transform_keys` is no longer SQL-specific here, so `obfuscate_sql_values` would be a lie — both Agent spellings are accepted as `#[serde(alias = ...)]`, in the same style as #2490's PascalCase aliases. `obfuscation_config::tests::test_agent_json_obfuscation_field_names` pins all of it.

BREAKING CHANGE: `JsonObfuscatorConfig::transformer` and `JsonStringTransformer` are removed. JSON transforms move to `JsonObfuscator::obfuscate_with` or `JsonObfuscator::obfuscate_into`. SQL obfuscation functions now return `Result`.

# How to test the change?

The following pass on the rebased branch:

```bash
cargo +stable clippy -p libdd-trace-obfuscation --all-targets -- -D warnings
cargo +nightly-2026-07-26 fmt --all -- --check
cargo nextest run -p libdd-trace-obfuscation           # 387 passed
cargo test -p libdd-trace-obfuscation --doc
cargo check -p libdd-trace-stats -p libdd-data-pipeline-core \
  -p libdd-data-pipeline -p libdd-data-pipeline-ffi --all-targets
cargo nextest run -p libdd-data-pipeline -p libdd-trace-stats \
  -E '!test(tracing_integration_tests::)'              # 253 passed
```

The `tracing_integration_tests::` suite needs Docker and was not run locally. `Cargo.lock` gains only `thiserror`, which `LICENSE-3rdparty.csv` already covers, so no regeneration was needed. `cargo deny check` still reports pre-existing workspace advisory and license-policy failures unrelated to this diff.

# References

- Closes: #2541
- Related: #2490
- Related: DataDog/saluki#2601


libdd-trace-obfuscation/src/json/mod.rs
```diff
@@ -1,46 +1,222 @@
...
+
+/// How much memory a [`JsonObfuscationScratch`] is holding, and the limits
+/// [`JsonObfuscationScratch::trim_to`] shrinks it to.
+///
+/// Both counts are approximate: they describe the buffers' capacities, not the allocator's
+/// bookkeeping.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct ScratchCapacity {
+    /// Bytes the string buffer used to unescape transform values can hold.
```
webern: Reword this. It doesn't sound like a real sentence. I've read it several times and cannot understand the grammar.

libdd-trace-obfuscation/src/json/mod.rs
```diff
@@ -50,153 +226,353 @@ impl JsonObfuscator {
...
+    ///         }
+    ///     });
+    /// assert!(report.is_clean());
+    /// assert_eq!(out, r#"{"query":"select * from users where id = ?"}"#);
+    /// ```
+    pub fn obfuscate_with<F, E>(
+        &self,
+        input: &str,
+        transform: F,
+    ) -> (String, JsonObfuscationReport<E>)
```
webern: Should JsonObfuscationReport<E> be JsonObfuscationError<E> and this return type be: `-> Result<String, JsonObfuscationError>`?

libdd-trace-obfuscation/src/obfuscate.rs
```diff
@@ -141,9 +143,14 @@ pub fn obfuscate_pb_span(span: &mut pb::Span, config: &ObfuscationConfig) {
...
-            span.resource.clone_from(&obfuscated_query);
-            span.meta.insert(TAG_SQLQUERY.to_owned(), obfuscated_query);
+            // A resource that obfuscates to nothing is left as sent rather than blanked; see
+            // `obfuscate_sql`.
+            if let Ok(obfuscated_query) =
+                crate::sql::obfuscate_sql(&span.resource, &config.sql, dbms)
+            {
+                span.resource.clone_from(&obfuscated_query);
+                span.meta.insert(TAG_SQLQUERY.to_owned(), obfuscated_query);
+            }
```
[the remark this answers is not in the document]
webern: Fixed in https://github.com/DataDog/libdatadog/pull/2548/commits/4d4bb28a2f6d30fbae5bd4504061857664570f81

libdd-trace-obfuscation/src/json/mod.rs
```diff
@@ -50,153 +238,370 @@ impl JsonObfuscator {
...
+        // is passed on verbatim, which is what the Agent does when it cannot unquote.
+        let literal = self.input.get(self.literal.clone()).unwrap_or_default();
+        let value = match unescape(literal, self.unescaped) {
+            Unescaped::Borrowed(s) => s,
+            Unescaped::Buffered => self.unescaped.as_str(),
+        };
+
+        self.out.push('"');
+        match transform(value) {
+            Ok(replacement) => self.out.push_str(&replacement),
```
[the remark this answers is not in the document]
webern: Fixed in https://github.com/DataDog/libdatadog/pull/2548/commits/f183a033c7a0a1027822bfddc8a9746bd721e131

libdd-trace-obfuscation/src/sql.rs
```diff
@@ -2264,6 +2321,110 @@ fn normalize_plan_sql(s: &str) -> String {
...
+        super::obfuscate_sql(s, config, dbms).unwrap_or_default()
+    }
+
+    fn obfuscate_sql_string(s: &str) -> String {
+        super::obfuscate_sql_string(s).unwrap_or_default()
+    }
+
+    fn obfuscate_sql_string_normalized(s: &str) -> String {
+        super::obfuscate_sql_string_normalized(s).unwrap_or_default()
+    }
```
Eldolfin: I find these helpers a bit confusing, it would be clearer if other names were used (`obfuscate_sql_or_empty` etc...) or even better if the tests actually got refactored with the new return type

libdd-trace-obfuscation/src/sql.rs
```diff
@@ -2145,10 +2145,60 @@ fn collapse_limit_two_args(s: &str) -> String {
...
+/// (`pkg/trace/agent/obfuscate.go`), discarding the unobfuscated SQL rather than forwarding it.
+pub const SQL_NON_PARSABLE_REPLACEMENT: &str = "Non-parsable SQL query";
+
+/// Why a SQL string could not be obfuscated.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, thiserror::Error)]
+#[non_exhaustive]
+pub enum SqlObfuscationError {
+    /// Obfuscation produced nothing, so there is no query to report. The Agent fails the same way
+    /// (`errors.New("result is empty")` in `pkg/obfuscate/sql.go`); input that is empty or only
+    /// whitespace lands here.
```
Eldolfin: I think this is too much details for public docs: I'm not a fan of referencing other repo's code without a permalink even in regular comments.

libdd-trace-obfuscation/src/sql.rs
```diff
@@ -2145,10 +2145,60 @@ fn collapse_limit_two_args(s: &str) -> String {
...
+///
+/// This is the Agent's JSON SQL-transform marker only: `sqlObfuscationTransformer` in
+/// `pkg/obfuscate/json.go` returns this string, deliberately verbose, so the failure is visible to
+/// the user in the JSON value it replaces. It is not what the Agent writes for a failed SQL span
+/// resource or stats group; that is [`SQL_NON_PARSABLE_REPLACEMENT`].
+///
+/// The message is the Agent's, byte for byte, so the same input yields the same value from either
+/// implementation.
+pub const SQL_OBFUSCATION_FAILURE_REPLACEMENT: &str =
+    "Datadog-agent failed to obfuscate SQL string. Enable agent debug logs for more info.";
```
Eldolfin: I don't think this value belongs in this repo, libdatadog is first a tracer library and should stay independant from its consumers.

libdd-trace-obfuscation/src/obfuscate.rs
```diff
@@ -429,12 +436,24 @@ fn should_obfuscate_cc_key(key: &str, config: &ObfuscationConfig) -> bool {
...
+/// Non-empty SQL resources that tokenize to nothing, so obfuscation fails and the Agent's
+/// `Non-parsable SQL query` marker has to replace them. The comment-only cases carry text no test
+/// expects to see again, so a resource forwarded as sent instead of discarded fails loudly.
+#[cfg(test)]
+const UNOBFUSCATABLE_SQL: &[&str] = &[
+    "   ",
+    "\n\t \r\n",
+    "-- SELECT secret-hunter2-must-not-leak",
+    "/* SELECT secret-hunter2-must-not-leak */",
+];
```
Eldolfin: This can go in the tests module

libdd-trace-obfuscation/src/json/mod.rs
```diff
@@ -50,153 +238,403 @@ impl JsonObfuscator {
...
+    keep_depth: usize,
+    /// True while reading a key rather than a value.
+    key: bool,
+    /// True once the current value has been replaced, so a value spanning several characters is
+    /// replaced once rather than per character.
+    wiped: bool,
+    /// True while inside a value that is kept as sent.
+    keeping: bool,
+    /// True while reading a value that `transform` will rewrite.
+    transforming: bool,
```
Eldolfin: Having this many boolean in a parser always feels wrong to me. A enum would probably make more sense and prevent inconsistant states

Eldolfin (commented):
Looks good 👍🏽 

One thing I saw repeated multiple times in this PR is references of types and files from the agent repo which I think we should avoid because they will get stale and don't add much value IMO

libdd-trace-obfuscation/src/json/mod.rs
webern: Agreed. Refactored to use an enum in e5d1e5305

libdd-trace-obfuscation/src/obfuscate.rs
webern: Oops, fixed in ee4503c5c95abe34d7e576138a32a258426684f3

libdd-trace-obfuscation/src/sql.rs
webern: Agreed. Constant pet peeve with Opus is these little zero-value abstractions. Deleted them.

Fixed in https://github.com/DataDog/libdatadog/commit/fb62d3adc12ebe418522c376d6130436cc452848.

libdd-trace-obfuscation/src/sql.rs
webern: Makes sense. Leaked from the verification against the reference implementation.

Fixed in:
- https://github.com/DataDog/libdatadog/commit/40b39aeaff77c3fa1faa477dd485915b6d79885d (public documentation)
- https://github.com/DataDog/libdatadog/commit/b14e9ee48fee382637c2ddbbc1559e04a2af5359 (private comments)

libdd-trace-obfuscation/src/sql.rs
webern: Makes sense. Removed Agent references.

Fixed in:
- https://github.com/DataDog/libdatadog/commit/40b39aeaff77c3fa1faa477dd485915b6d79885d (generic failure message and public documentation)
- https://github.com/DataDog/libdatadog/commit/9afbe4e2acb97aa4e2788bf97a49f8f3d5bcba19 (test name and comment)

libdd-trace-obfuscation/src/obfuscate.rs
webern: Fixed

https://github.com/DataDog/libdatadog/commit/8f8e202c01e646257a4ff43ec7dad69565eb47a3

libdd-trace-obfuscation/src/json/mod.rs
webern: Yeah, good call. Renamed Pass to ParserState for clarity and replaced the bools with a ParserPhase enum.

Fixed in https://github.com/DataDog/libdatadog/commit/e5d1e5305d5c4b9b87853d4f6abb10c78f987645.

Eldolfin (approved):

PR #2551, opened 2026-09-18
feat(sidecar)!: Authenticate sidecar connections and shared memory

bwoebi:
This is quite far-reaching:
- connections get authenticated according to SO_PEERCREDS (or macos equivalent) - including caller checking on macos that socket ownership is correct
- strict checks against toctou by using a trusted directory for shared memory
- in process listener drops permissions if root to avoid being able to create files


datadog-sidecar-ffi/src/lib.rs
```diff
@@ -365,9 +363,18 @@ pub extern "C" fn ddog_sidecar_shutdown_master_listener() -> MaybeError {
     MaybeError::None
 }
 
+/// Remove the master listener's socket and lock file, for SAPIs that exit without running
+/// PHP's module shutdown - php-fpm's master calls `exit()` straight from `fpm_pctl_exit()`, so
+/// `ddog_sidecar_shutdown_master_listener` never runs there. Safe to call more than once, and a
+/// no-op in a process that did not bind them.
+#[no_mangle]
+pub extern "C" fn ddog_sidecar_reap_master_listener_files() {
+    MasterListener::reap_bound_files_at_exit();
```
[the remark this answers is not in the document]
bwoebi: not this crates convention.

datadog-sidecar/src/setup/mod.rs
```diff
@@ -35,6 +35,10 @@ pub trait Liaison: Sized {
     fn attempt_listen(&self) -> io::Result<Option<IpcServer>>;
     fn ipc_shared() -> Self;
     fn ipc_per_process() -> Self;
+
+    fn bound_files(&self) -> Vec<std::path::PathBuf> {
```
yannham: Do we really need to return an owned value here? Naively it looks like a `&[_]` would make more sense.

datadog-sidecar/src/setup/thread_listener.rs
```diff
@@ -18,21 +19,151 @@ use crate::setup::AbstractUnixSocketLiaison;
...
+///
+/// Tokio's blocking-pool threads are created from this one and so inherit the dropped
+/// credentials. The watchdog thread is started earlier and keeps what it had; it only samples
+/// memory usage and aborts, and creates nothing.
+fn drop_listener_thread_privileges(uid: u32, gid: u32) {
+    #[cfg(target_os = "linux")]
+    {
+        // SAFETY: none of these take pointers, and none can fail in a way that matters here.
+        let (euid, pid) = unsafe { (libc::geteuid(), libc::getpid()) };
+        let tid = unsafe { libc::syscall(libc::SYS_gettid) } as libc::pid_t;
```
yannham, line 55: Why not `libc::gettid()` ?

datadog-sidecar/src/setup/thread_listener.rs
```diff
@@ -18,21 +19,151 @@ use crate::setup::AbstractUnixSocketLiaison;
...
+                "Failed dropping sidecar thread {tid} to uid {uid}/gid {gid}: {err}; \
+                 continuing as uid {euid}"
+            );
+            return;
+        }
+        info!("Sidecar listener thread {tid} dropped from root to uid {uid}/gid {gid}");
+    }
+}
+
+/// Drop *this* thread if the served uid is already known.
```
yannham, line 125: ?

```suggestion
/// Drop *this* thread's privileges if the served uid is already known.
```

datadog-sidecar/src/setup/thread_listener.rs
```diff
@@ -18,21 +19,151 @@ use crate::setup::AbstractUnixSocketLiaison;
...
+///
+/// Registered as tokio's `on_thread_start`, because a current-thread runtime is not the only
+/// thread the sidecar ends up with: `spawn_blocking` has its own pool, created lazily. Those
+/// threads inherit the credentials of whichever thread spawned them, which is the dropped
```
yannham, line 129: Nit: "dropped thread" is a bit confusing as a terminology, because dropped sounds like the thread was killed. Maybe something like demoted/deprived?

datadog-sidecar/src/setup/thread_listener.rs
```diff
@@ -178,12 +352,29 @@ async fn accept_socket_loop_thread(
...
+                                        // seen; say so rather than let a second pool wonder why
+                                        // its remote config never arrives.
+                                        if let Some((served, _)) = SERVED_IDS.get() {
+                                            if *served != cred.uid {
+                                                warn!(
+                                                    "Peer pid {} runs as uid {}, but this listener serves uid {served}. Shared memory will not be readable by uid {}.",
+                                                    cred.pid, cred.uid, cred.uid
+                                                );
+                                            }
+                                        }
```
yannham, line 375:
```suggestion
                                        if matches!(SERVED_IDS.get(), Some((served,_)) if *served != cred.uid {
                                            warn!(
                                                "Peer pid {} runs as uid {}, but this listener serves uid {served}. Shared memory will not be readable by uid {}.",
                                                cred.pid, cred.uid, cred.uid
                                            );
                                        }
```

datadog-sidecar/src/setup/unix.rs
```diff
@@ -68,56 +201,93 @@ impl Liaison for SharedDirLiaison {
...
             fs::remove_file(&self.socket_path)?;
         }
-        Ok(Some(SeqpacketListener::bind(&self.socket_path)?))
+        let listener = SeqpacketListener::bind(&self.socket_path)?;
+        set_socket_mode(&self.socket_path, self.access.socket_mode());
+        Ok(Some(listener))
+    }
+
+    fn bound_files(&self) -> Vec<PathBuf> {
+        vec![self.socket_path.clone(), self.lock_path.clone()]
```
yannham, line 210: Should we return `Vec<&PathBuf>` instead?

datadog-sidecar/src/setup/unix.rs
```diff
@@ -230,6 +548,72 @@ mod tests {
 
     use super::Liaison;
 
+    /// A socket we created ourselves is exactly what the listener-owned check must accept.
+    /// The socket file outlives the process that bound it, so a listener pid that no longer
+    /// resolves leaves nothing to check the owner against - and the answer to "cannot verify"
+    /// is refuse, not proceed.
```
yannham, line 566:
```suggestion
    /// resolves leaves nothing to check the owner against.
```

datadog-sidecar/src/log.rs
```diff
@@ -672,23 +686,59 @@ mod tests {
...
-        assert_eq!(2, map.len());
-        assert_eq!(map[&Level::WARN], 2);
-        assert_eq!(map[&Level::ERROR], 1);
 
+        // This counter is process-global and other tests in this binary emit records, so exact
+        // counts are not something this test can assert - a stray record from a concurrently
+        // running test would land in one of the windows below. What it *can* assert, and what
+        // it is actually for, is per-level attribution, that a filtered-out level is never
+        // counted, and that collecting drains. The batch is sized so a stray record or two
+        // cannot change any of those conclusions.
```
yannham, line 699: Should we use serial tests instead, as done elsewhere in libdatadog, instead of relying on such strategies? Looks like potential future flakyness, in particular:

> The batch is sized so a stray record or two cannot change any of those conclusions

datadog-sidecar/src/tracer.rs
```diff
@@ -11,19 +11,44 @@ use std::ffi::CString;
...
+/// a missing limiter as "no rate limiting" rather than as a fatal error.
+pub static SHM_LIMITER: LazyLock<Option<Mutex<ManuallyDrop<ShmLimiterMemory<()>>>>> = LazyLock::new(
+    || match ShmLimiterMemory::create(shm_limiter_path()) {
+        Ok(memory) => {
+            unsafe { libc::atexit(drop_shm_limiter) };
+            Some(Mutex::new(ManuallyDrop::new(memory)))
+        }
+        Err(e) => {
+            error!(
+                "Could not create the shared rate limiter at {}: {e}. Continuing without                  rate limiting.",
```
yannham, line 30:
```suggestion
                "Could not create the shared rate limiter at {}: {e}. Continuing without rate limiting.",
```

datadog-sidecar/tests/connection_auth.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+impl Drop for Sidecar {
+    fn drop(&mut self) {
+        self.stop.notify_one();
+        // Detach instead of joining unconditionally. `main_loop` does not return once it has
+        // served a connection: its serve loop never observes the client's disconnect, which
+        // reproduces with authorization switched off entirely (`AuthPolicy::OsEnforced`), so it
+        // is not this feature's doing. Joining unconditionally would hang the test run rather
+        // than fail it; a detached thread dies with the test process.
```
yannham, line 86: I don't understand this comment. The thread below is joined?

libdd-ipc/src/platform/unix/mem_handle.rs
```diff
@@ -1,39 +1,105 @@
...
+///
+/// Only success is remembered. Latching a failure would mean that a host where somebody had
+/// squatted the name could never recover once cleaned up, short of a restart.
+///
+/// `creating` distinguishes the two callers, and it matters: [`private_dir::ensure`] may
+/// discard and recreate the directory to repair it, which would take the segments a writer is
+/// serving from with it. Only the path that is about to create may do that; a reader verifies
+/// and nothing more.
+fn check_fallback_dir(creating: bool) -> nix::Result<()> {
+    static VERIFIED: OnceLock<()> = OnceLock::new();
```
yannham, line 73: Should this just be an `AtomicBool` with relaxed operations? It looks a tad overkill to have a `OnceLock<()>`. Though it's arguably a not very important detail.

libdd-ipc/src/platform/unix/process.rs
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+    Some(ProcInfo {
+        uid: uid?,
+        gid: gid?,
+        ppid: ppid?,
+    })
+}
+
+#[cfg(not(target_os = "linux"))]
```
yannham, line 80: We should probably have a proper `target_os = "macos"` (or whatever it is) filter here, instead of a negative one.

spawn_worker/src/unix/spawn.rs
```diff
@@ -60,6 +60,17 @@ mod helper {
...
         }
+
+        /// Point an entry at a string this vector does not own.
+        ///
+        /// For post-`fork` callers, which must not allocate: the pointee only has to outlive the
+        /// `execve` that follows, so a buffer in the calling stack frame is fine. The previous
+        /// `CString` stays in `items`, keeping that slot's own allocation alive.
+        /// Only the Linux `FdExec` fallback needs this.
+        #[cfg(target_os = "linux")]
+        pub fn set_ptr(&mut self, index: usize, ptr: *const libc::c_char) {
```
yannham, line 71: Nit: technically speaking this should be `unsafe`, because providing a non-NULL terminated string is likely to cause UB somewhere down the line?

spawn_worker/src/unix/spawn.rs
```diff
@@ -521,49 +533,70 @@ impl SpawnWorker {
...
                     let tmpdir = if tmpdir.is_null() {
-                        b"/tmp"
+                        b"/tmp".as_slice()
                     } else {
                         CStr::from_ptr(tmpdir).to_bytes()
                     };
-                    if tmpdir.len() < 220 {
+                    // getrandom(2) - async signal safe and unguessable.
+                    let mut random = [0u8; 8];
+                    let got = libc::syscall(
```
yannham, line 550: why not use `libc::getrandom()`?

spawn_worker/src/unix/spawn.rs
```diff
@@ -521,49 +533,70 @@ impl SpawnWorker {
...
+                        const HEX: &[u8; 16] = b"0123456789abcdef";
                         temp_path[..tmpdir.len()].copy_from_slice(tmpdir);
                         let mut off = tmpdir.len();
                         let spawn_prefix = b"/dd-ipc-spawn_";
                         temp_path[off..off + spawn_prefix.len()].copy_from_slice(spawn_prefix);
                         off += spawn_prefix.len();
-                        for _ in 0..8 {
-                            temp_path[off] = fastrand::alphanumeric() as u8;
-                            off += 1;
+                        for byte in random {
```
yannham, line 564: I feel like this is defined 10 times through libdatadog 😅 and that we have a somehow reusable version somewhere? But maybe it's in a different crate though

yannham (approved):
It's hard to convince oneself that all the verification paths are correct given the size of the PR, but it's an improvement and looks correct from a distance.

datadog-sidecar/src/setup/thread_listener.rs
bwoebi: Exists only in glibc 2.30+.

spawn_worker/src/unix/spawn.rs
bwoebi: Because it requires glibc 2.25.

libdd-ipc/src/platform/unix/mem_handle.rs
bwoebi: Agree, it's overkill here.

datadog-sidecar/src/setup/thread_listener.rs
```diff
@@ -178,12 +352,29 @@ async fn accept_socket_loop_thread(
...
                                     }
-                                    drop(SHM_LIMITER.lock());
-                                });
+                                }
                                 handler(conn);
```
gyuheon0h, line 378: Please educate me if I am missing something, but we serve the connection here unconditionally? It sits outside the `if authorizer.authorize(&cred) == Decision::Allow` block so `Decision::Deny` has no effect?

datadog-sidecar/src/setup/thread_listener_windows.rs
```diff
@@ -89,15 +90,18 @@ impl MasterListener {
     }
 
     /// Check if the master listener is active for the given PID.
-    pub fn is_active(pid: i32) -> bool {
+    pub fn is_active() -> bool {
         let listener_mutex = MASTER_LISTENER.get_or_init(|| Mutex::new(None));
         if let Ok(listener_guard) = listener_mutex.lock() {
-            listener_guard.as_ref().is_some_and(|l| l.pid == pid)
+            listener_guard.is_some()
```
gyuheon0h, line 96: We no longer care if this process is actually the one that set up the listener?

datadog-sidecar/src/setup/thread_listener.rs
gyuheon0h: I would've expected something [like this path here](https://github.com/DataDog/libdatadog/blob/7f2c06097ca77291791a9521ebca13d11c0f632b/datadog-sidecar/src/entry.rs#L160-L169)

gyuheon0h (approved):
LGTM on crashtracker and general authentication standpoint -- your call on other integrations

spawn_worker/src/unix/spawn.rs
bwoebi: yeah, just write! with the right format...

